PR #17, opened 2018-06-06
Fix JSON encoding for malformed strings

kevinlebrun:
If for some reason the `json_encode` step fails, the encoder will build an invalid JSON output. As a fix I ignored all invalid spans from the JSON output.

It is something we want to log to inform the developer I think. It happened to me when I tried to send as tags all my SQL bindings. It turns out that my primary keys are binary encoded GUID.

ping @vlad-mh @palazzem

### DIFF
--- a/src/DDTrace/Encoders/Json.php
+++ b/src/DDTrace/Encoders/Json.php
@@ -12,14 +12,10 @@ final class Json implements Encoder
      */
     public function encodeTraces(array $traces)
     {
-        $notEmpty = function ($value) {
-            return !empty($value);
-        };
-
-        return '[' . implode(',', array_map(function ($trace) use ($notEmpty) {
+        return '[' . implode(',', array_map(function ($trace) {
             return '[' . implode(',', array_filter(array_map(function ($span) {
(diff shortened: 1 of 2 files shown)


jcchavezs:
+1

src/DDTrace/Encoders/Json.php
```diff
@@ -12,10 +12,14 @@
...
-            return '[' . implode(',', array_map(function ($span) {
+        $notEmpty = function ($value) {
+            return !empty($value);
+        };
+
+        return '[' . implode(',', array_map(function ($trace) use ($notEmpty) {
+            return '[' . implode(',', array_filter(array_map(function ($span) {
                 return $this->encodeSpan($span);
-            }, $trace)) . ']';
+            }, $trace), $notEmpty)) . ']';
```
jderusse: When `array_filter`'s callback parameter is not supplied, entries equals to `false` (after casted to bool) will be removed. Which is the case here  with empty string

kevinlebrun: Nice catch! Thank you.

PR #18, opened 2018-06-06
Fix ScopeManager fail to manage multiple scopes

kevinlebrun:
In its state the scope manager cannot manage more than one scope because `unset` will preserve the keys. I replaced `unset` by `array_splice` which does the same thing but recompute the array keys.

ping @vlad-mh @palazzem


jcchavezs:
+1

PR #20, opened 2018-06-15
Prevent curl to print the response in STDOUT

kevinlebrun:
When flushing traces, if everything work as expected, a "OK" can be present in the response; even in a `terminate` method for Laravel. In case of an API this will prevent the client to correctly parse the output.


jcchavezs:
Is this something that the Datadog agent returns or a curl thing?

kevinlebrun:
From what I seen it is `curl` that prints the response out. That's  why I added the `curl` option.

PR #23, opened 2018-06-28
Enhancement: Keep packages sorted in composer.json

localheinz:
This PR

* [x] keeps packages sorted in `composer.json`

Follows https://github.com/DataDog/php-datadogstatsd/pull/60.

💁‍♂️ For reference, see https://getcomposer.org/doc/06-config.md#sort-packages.


localheinz:
Looks like something is up with the CircleCi integration.

PR #24, opened 2018-07-03
Enhancement: Add configuration for phpunit

localheinz:
This PR

* [x] adds a configuration for `phpunit/phpunit`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,10 +50,10 @@
             "@test-unit",
             "@test-integration"
         ],
-        "test-unit": "phpunit ./tests/Unit",
+        "test-unit": "phpunit --colors=always --configuration=phpunit.xml --testsuite=unit",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -50,10 +50,10 @@
             "@test-unit",
             "@test-integration"
         ],
-        "test-unit": "./vendor/bin/phpunit ./tests/Unit",
+        "test-unit": "./vendor/bin/phpunit --colors=always --configuration=phpunit.xml --testsuite=unit",
```
localheinz: The `--colors=always` option ensures that we see colored output when we run

```
$ composer test-unit
```

localheinz:
@chuck 

Rebased to trigger the build again.

localheinz:
Thank you, @chuck!

PR #25, opened 2018-07-03
Fix: Composer resolves binaries itself

localheinz:
This PR

* [x] removes unnecessary prefixes to vendor binaries in `composer.json`

💁‍♂️ For reference, see https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands:

>**Note**: Before executing scripts, Composer's bin-dir is temporarily pushed on top of the PATH environment variable so that binaries of dependencies are easily accessible. In this example no matter if the `phpunit` binary is actually in `vendor/bin/phpunit` or `bin/phpunit` it will be found and executed.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,16 +50,16 @@
             "@test-unit",
             "@test-integration"
         ],
-        "test-unit": "./vendor/bin/phpunit ./tests/Unit",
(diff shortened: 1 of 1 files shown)


localheinz:
Thank you, @jcchavezs and @chuck!

PR #29, opened 2018-07-11
Fix: Use caret operator where applicable

localheinz:
This PR

* [x] uses the `^` operator instead of `~` and `*` for version constraints where applicable

💁‍♂️ For reference, see https://getcomposer.org/doc/articles/versions.md.


localheinz:
Thank you, @chuck!

PR #30, opened 2018-07-11
Enhancement: Use forward compatibility layer of phpunit/phpunit

localheinz:
This PR

* [x] uses the forward compatibility layer of `phpunit/phpunit`

💁‍♂️ This will make upgrading to newer versions of `phpunit/phpunit` (perhaps if support for PHP 5.6 is dropped) easier.


localheinz:
Thank you, @chuck!

PR #34, opened 2018-07-12
Set $logger for the Encoder class, and log errors encoding JSON

chuck:
Follow-up to #17 - pass the $logger to the Encoder, and if there's an error encoding JSON, log it at a debug level. If others like the LoggerAwareTrait pattern we could switch the Transport to use it as well. I'm not 100% happy with this since if you call Encoder methods yourself you aren't guaranteed to have a Logger, though. But passing $logger directly to the encodeTraces() method, while expedient, is almost certainly wrong.

### DIFF
--- a/src/DDTrace/Encoders/Json.php
+++ b/src/DDTrace/Encoders/Json.php
@@ -4,9 +4,13 @@ namespace DDTrace\Encoders;
 
 use DDTrace\Encoder;
 use DDTrace\Span;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
 
 final class Json implements Encoder
 {
+    use LoggerAwareTrait;
+
     /**
      * {@inheritdoc}
      */
@@ -35,6 +39,7 @@ final class Json implements Encoder
     {
         $json = json_encode($this->spanToArray($span));
         if (false === $json) {
+            $this->logger->debug("Failed to json-encode span: " . json_last_error_msg());
             return "";
         }
 
--- a/src/DDTrace/Encoders/Json.php
+++ b/src/DDTrace/Encoders/Json.php
@@ -4,12 +4,20 @@ namespace DDTrace\Encoders;
 
 use DDTrace\Encoder;
 use DDTrace\Span;
-use Psr\Log\LoggerAwareInterface;
-use Psr\Log\LoggerAwareTrait;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 
 final class Json implements Encoder
 {
-    use LoggerAwareTrait;
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    public function __construct(LoggerInterface $logger = null)
+    {
+        $this->logger = $logger ?: new NullLogger();
+    }
 
     /**
      * {@inheritdoc}
--- a/src/DDTrace/Encoders/Noop.php
+++ b/src/DDTrace/Encoders/Noop.php
@@ -3,9 +3,13 @@
 namespace DDTrace\Encoders;
 
 use DDTrace\Encoder;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
 
 final class Noop implements Encoder
 {
+    use LoggerAwareTrait;
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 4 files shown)


src/DDTrace/Encoders/Json.php
```diff
@@ -35,6 +39,7 @@ private function encodeSpan(Span $span)
     {
         $json = json_encode($this->spanToArray($span));
         if (false === $json) {
+            $this->logger->debug("Failed to json-encode span: " . json_last_error_msg());
```
pawelchcki: Is `$this->logger` always initialized at this point?

src/DDTrace/Encoders/Noop.php
```diff
@@ -3,9 +3,13 @@
 namespace DDTrace\Encoders;
 
 use DDTrace\Encoder;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
 
 final class Noop implements Encoder
 {
+    use LoggerAwareTrait;
```
pawelchcki: Do we need this trait in `Noop` class?

chuck:
I agree that you wouldn't want this in production, but it's the kind of thing that turning on debug logging is for.

Would it potentially make sense to try serializing without the tags, so that we just lose them instead of the whole span?

tests/Unit/Encoders/JsonTest.php
```diff
@@ -46,7 +47,15 @@ public function testEncodeIgnoreSpanWhenEncodingFails()
...
 
+        $logger = $this->prophesize(LoggerInterface::class);
+        $logger
+            ->debug(
+                'Failed to json-encode span: Malformed UTF-8 characters, possibly incorrectly encoded'
+            )
+            ->shouldBeCalled();
+
         $jsonEncoder = new Json();
+        $jsonEncoder->setLogger($logger->reveal());
```
pawelchcki: Could the logger be set in constructor one line above?

jcchavezs:
@chuck I don't think logs are meant to recover traces. I honestly would take advantage of the opportunity that DataDog also supports metrics.

Ping @palazzem

chuck:
I'm not thinking about this as recovering the trace, I'm thinking of it as giving someone developing an integration, or instrumenting their own code, a way to see why things are otherwise silently failing.

src/DDTrace/Encoders/Noop.php
chuck: The way things are currently set up, yes, because we call setLogger() on the encoder object from within the transport. Though the Noop Encoder is a bit of an odd duck anyway in terms of when you'd need it...

tests/Unit/Encoders/JsonTest.php
localheinz: I agree it would be better to allow injection via constructor.

How about

```diff
diff --git a/src/DDTrace/Encoders/Json.php b/src/DDTrace/Encoders/Json.php
index 01876e6..8b8c8f2 100644
--- a/src/DDTrace/Encoders/Json.php
+++ b/src/DDTrace/Encoders/Json.php
@@ -4,12 +4,19 @@ namespace DDTrace\Encoders;

 use DDTrace\Encoder;
 use DDTrace\Span;
-use Psr\Log\LoggerAwareInterface;
-use Psr\Log\LoggerAwareTrait;
+use Psr\Log;

 final class Json implements Encoder
 {
-    use LoggerAwareTrait;
+    /**
+     * @var Log\LoggerInterface
+     */
+    private $logger;
+
+    public function __construct(Log\LoggerInterface $logger = null)
+    {
+        $this->logger = $logger ?: new Log\NullLogger();
+    }

     /**
      * {@inheritdoc}
diff --git a/tests/Unit/Encoders/JsonTest.php b/tests/Unit/Encoders/JsonTest.php
index 5db16ef..1f46924 100644
--- a/tests/Unit/Encoders/JsonTest.php
+++ b/tests/Unit/Encoders/JsonTest.php
@@ -54,8 +54,7 @@ JSON;
             )
             ->shouldBeCalled();

-        $jsonEncoder = new Json();
-        $jsonEncoder->setLogger($logger->reveal());
+        $jsonEncoder = new Json($logger->reveal());
         $encodedTrace = $jsonEncoder->encodeTraces([[$span, $span]]);
         $this->assertEquals($expectedPayload, $encodedTrace);
     }
```

?

src/DDTrace/Encoders/Json.php
chuck: Taken care of with constructor injection

chuck:
Updated for constructor injection, which also means that to get these log message, you have to explicitly set a logger on the encoder. For me that's good to have for troubleshooting, but makes it pretty hard to get huge numbers of log lines without meaning to.

PR #35, opened 2018-07-12
Enhancement: Improve legibility of ScopeManager

localheinz:
This PR

* [x] improves the legibility of the `ScopeManager` a bit

### DIFF
--- a/src/DDTrace/ScopeManager.php
+++ b/src/DDTrace/ScopeManager.php
@@ -38,13 +38,12 @@ final class ScopeManager implements OpenTracingScopeManager
 
     public function deactivate(Scope $scope)
     {
-        $scopeLength = count($this->scopes);
+        $i = array_search($scope, $this->scopes, true);
 
-        for ($i = 0; $i < $scopeLength; $i++) {
-            if ($scope === $this->scopes[$i]) {
-                array_splice($this->scopes, $i, 1);
-                break;
-            }
+        if (false === $i) {
+            return;
         }
+
+        array_splice($this->scopes, $i, 1);
     }
 }


src/DDTrace/ScopeManager.php
```diff
@@ -33,18 +33,17 @@ public function getActive()
             return null;
         }
 
-        return $this->scopes[count($this->scopes) - 1];
+        return end($this->scopes);
```
localheinz: Apparently, we are interested in returning the last element, so maybe `end()` is better?

src/DDTrace/ScopeManager.php
```diff
@@ -33,18 +33,17 @@ public function getActive()
...
     }
 
     public function deactivate(Scope $scope)
     {
-        $scopeLength = count($this->scopes);
+        $i = array_search($scope, $this->scopes, true);
```
localheinz: We could use `array_search()` to find the index of `$scope` in `$this->scopes`, maybe better than looping?

src/DDTrace/ScopeManager.php
chuck: end() modifies the array's internal pointer, which may or may not surprise us somewhere else. I don't think the original line is unclear.

chuck:
@localheinz Can you update based on feedback, or should I work off of your branch?

localheinz:
@chuck 

Already amended a couple days ago - does this work? Please let me know if you prefer me to apply more changes. 

🤓

chuck:
Sorry, missed the update. Thanks!

localheinz:
Thank you, @chuck!

PR #36, opened 2018-07-12
Fix: Remove unnecessary continue statement

localheinz:
This PR

* [x] removes an unnecessary `continue` statement


src/DDTrace/Propagators/TextMap.php
```diff
@@ -41,8 +41,6 @@ public function extract($carrier)
             } elseif (strpos($key, self::DEFAULT_BAGGAGE_HEADER_PREFIX) === 0) {
                 $baggageItems[substr($key, strlen(self::DEFAULT_BAGGAGE_HEADER_PREFIX))] = $value;
             }
-
-            continue;
```
localheinz: It's already the end of the loop.

localheinz:
Thank you, @chuck!

PR #37, opened 2018-07-12
Fix: Use more appropriate assertions

localheinz:
This PR

* [x] uses more appropriate assertions

### DIFF
--- a/tests/Integration/Transport/HttpTest.php
+++ b/tests/Integration/Transport/HttpTest.php
@@ -97,6 +97,6 @@ final class HttpTest extends Framework\TestCase
         $httpTransport->send($traces);
         $output = ob_get_clean();
 
-        $this->assertEmpty($output);
+        $this->assertSame('', $output);
     }
 }
--- a/tests/Unit/Encoders/JsonTest.php
+++ b/tests/Unit/Encoders/JsonTest.php
@@ -34,7 +34,7 @@ JSON;
 
         $jsonEncoder = new Json($logger->reveal());
         $encodedTrace = $jsonEncoder->encodeTraces([[$span]]);
-        $this->assertEquals($expectedPayload, $encodedTrace);
+        $this->assertJsonStringEqualsJsonString($expectedPayload, $encodedTrace);
     }
 
     public function testEncodeIgnoreSpanWhenEncodingFails()


tests/Integration/Transport/HttpTest.php
```diff
@@ -97,6 +97,6 @@ public function testSilentlySendTraces()
         $httpTransport->send($traces);
         $output = ob_get_clean();
 
-        $this->assertEmpty($output);
+        $this->assertSame('', $output);
```
localheinz: Probably better to compare with an empty string, as 

```php
$this->assertEmpty('[]');
```

will pass.

tests/Unit/Encoders/JsonTest.php
```diff
@@ -28,6 +28,6 @@ public function testEncodeTracesSuccess()
         );
         $jsonEncoder = new Json();
         $encodedTrace = $jsonEncoder->encodeTraces([[$span]]);
-        $this->assertEquals($expectedPayload, $encodedTrace);
+        $this->assertJsonStringEqualsJsonString($expectedPayload, $encodedTrace);
```
localheinz: This assertion will also give a more helpful error message when it fails.

tests/Unit/SpanContextTest.php
```diff
@@ -35,7 +35,7 @@ public function testGetBaggageItemsReturnsExpectedValues()
             self::BAGGAGE_ITEM_KEY => self::BAGGAGE_ITEM_VALUE,
         ]);
 
-        $this->assertEquals(self::BAGGAGE_ITEM_VALUE, $context->getBaggageItem(self::BAGGAGE_ITEM_KEY));
+        $this->assertSame(self::BAGGAGE_ITEM_VALUE, $context->getBaggageItem(self::BAGGAGE_ITEM_KEY));
```
jcchavezs: Not sure about this change. `assertSame` gives me the feeling I am comparing objects and not the value of the string which is what we aim to test.

localheinz: @jcchavezs 

I think it depends on whether we care that a `string` value is returned, or would be fine, for example if an object were returned that could be casted to `string`. 

Not sure, what do you think?

Personally, I prefer to be as strict as possible, but I'm aware you know much more about this project than I do!

chuck: I agree with José here; being strict on types in tests when we just want a string-ish value could lead to brittleness later. Probably not in baggage items, but as a general philosophy I'd rather go that way. Kind of like accepting interfaces ... we want to get back something that acts like a string, it shouldn't be important what the actual implementation is.

tests/Integration/Transport/HttpTest.php
chuck: Could check strlen() also, but this is fine imo.

localheinz:
@chuck @jcchavezs 

No problem, will adjust later when I’m back at my computer.

PR #38, opened 2018-07-13
Fix: Remove dependency on guzzlehttp/psr7

localheinz:
This PR

* [x] removes the dependency on `guzzlehttp/psr7` and replaces it with `psr/http-message`


localheinz:
Thank you, @chuck!

PR #39, opened 2018-07-23
Fix tiny misconfiguration of phpcs, and add defaults in .phpcs.xml file

pawelchcki:
`phpcs` was used with both `Zend` and `PSR2` however only PSR2 was effectively used.

This PR instead uses PSR12 (a newer PSR2 with changes that are still compatible with our codebase).

In addition it adds my personal Favorites of disallowing `Todo` and `Fixme` comments.

With `.phpcs.xml` linting now also correctly works with VS Code extension PHP linter.

### DIFF
--- a/.phpcs.xml
+++ b/.phpcs.xml
@@ -1,17 +1,15 @@
 <?xml version="1.0"?>
-<ruleset name="DD">
-	<description>Generally-applicable sniffs for Datadog Tracing</description>
+<ruleset name="DD Trace PHP">
+    <description>Sniffs for DD Trace PHP</description>
 
-	<file>./</file>
+    <file>./</file>
 
-	<rule ref="PSR12" >
-	</rule>
+    <rule ref="PSR12"/>
 
-	<rule ref="Generic.Commenting.Todo"/>
-	<rule ref="Generic.Commenting.Fixme"/>
+    <rule ref="Generic.Commenting.Todo"/>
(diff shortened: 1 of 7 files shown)


.phpcs.xml
```diff
@@ -0,0 +1,17 @@
...
+	<description>Generally-applicable sniffs for Datadog Tracing</description>
+
+	<file>./</file>
+
+	<rule ref="PSR12" >
+	</rule>
+
+	<rule ref="Generic.Commenting.Todo"/>
+	<rule ref="Generic.Commenting.Fixme"/>
+    <rule ref="Generic.Commenting.DocComment.MissingShort">
```
localheinz: Maybe fix the indentation here?

.phpcs.xml
```diff
@@ -0,0 +1,17 @@
...
+	    <severity>0</severity>
+	</rule>
+
+	<exclude-pattern>*/vendor/*</exclude-pattern>
+</ruleset>
```
localheinz: How about adding a final newline?

Also see https://github.com/DataDog/dd-trace-php/pull/40.

.phpcs.xml
pawelchcki: Yeah having consistent indentation sounds like a good idea 👍 😄

.phpcs.xml
pawelchcki: #40 ❤️

PR #40, opened 2018-07-23
Enhancement: Add .editorconfig

localheinz:
This PR

* [x] adds an `.editorconfig`

💁‍♂️ I tried my best to match the standard existing in this repository. For reference, see https://editorconfig.org.


localheinz:
Thank you, @pawelchcki!

Issue #16, opened 2018-05-09
Passing objects within span tags


ryannutley:
Is this possible to do currently? Looking to pass an array/object into a span tag and view it within the trace attributes

jcchavezs:
Thanks for opening an issue! There is complexity around accepting other
things than scalar. In one hand there is a concern about how to serialise
it, in the other hand it is about cardinality, tags are suppose to be low
cardinality if possible. If you explain your use case we could understand
better the need for this.

> --

*José Carlos*

kevinlebrun:
I don't know if it will help you but if you flatten your data structures and join keys with a dot it will appear nicely in the UI. Here is an example with `setTag('bindings.0', 'xxx')`:
![screenshot 2018-06-08 18 39 19](https://user-images.githubusercontent.com/798082/41170147-688668f8-6b4b-11e8-8e1d-6081fee79da7.png)

jcchavezs:
This is interesting. OpenTracing spec [defines tag values as scalar](
https://github.com/opentracing/specification/blob/master/specification.md#set-a-span-tag)
so I would say the serialisation should be done by the instrumentation
(a.k.a developer)--

*José Carlos*

chuck:
@ryannutley Has your question been answered? I agree with @jcchavezs and @kevinlebrun that the spec is clear on tags being scalar. At the integration level, we might serialize down info ourselves in the way that Kevin suggested.

chuck:
I'm going to close this, but happy to continue conversation or re-open if needed.

PR #42, opened 2018-07-25
Feature/add error handling to set tag and logs

pawelchcki:


src/DDTrace/Span.php
```diff
@@ -310,7 +323,17 @@ public function getContext()
      */
     public function log(array $fields = [], $timestamp = null)
     {
-        throw new BadMethodCallException('not implemented');
+        foreach ($fields as $key => $value) {
+            if ($key === Tags\LOG_EVENT && $value === Tags\ERROR) {
+                $this->setError(true);
```
LotharSee: Not sure what `LOG_EVENT` are ; but are we certain they all are errors?

pawelchcki:
@LotharSee Thanks for review! 
I've added two addition test cases to properly test event handling. https://github.com/DataDog/dd-trace-php/pull/42/commits/5aee58216e8152f595f2eb242cfcfdac642ed71c

jcchavezs:
From OpenTracing perspective logs represent meaningful events in span
lifetime. This means that there could be many errors in the same span
whereas here it is just a tag that is being overriden everytime. I would
rather not implement it.--

*José Carlos*

jcchavezs:
I agree the exception should be removed if someone is going to use this in production. I have a suggestion tho. What if you use the fact that DD supports [JSON objects in tags](https://github.com/DataDog/dd-trace-php/issues/16#issuecomment-395818487) and do something like:

```php
logs.0.key=...
logs.0.value=...
logs.0.timestamp=...
```

That way, all logs (and not only error ones) are going to be keep.

Does it make sense?

PR #43, opened 2018-07-25
First cut at a Stdout transport for debugging

chuck:


src/DDTrace/Transport/Stdout.php
```diff
@@ -0,0 +1,35 @@
...
+
+    public function __construct(Encoder $encoder)
+    {
+        $this->encoder = $encoder;
+    }
+
+    public function send(array $traces)
+    {
+        var_dump($this->headers);
+        echo $this->encoder->encodeTraces($traces) . "\n";
```
kevinlebrun: Nit: I would use `PHP_EOL` instead of `\n`.

src/DDTrace/Transport/Stdout.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace DDTrace\Transport;
+
+use DDTrace\Encoder;
+use DDTrace\Transport;
+
+final class Stdout implements Transport
```
kevinlebrun: I like to have those kind of implementations for debugging purpose. It would have been very useful when I made my Laravel integration.

In my case I have some CLI scripts that generate traces and it will be very difficult to read debug traces with the standard output. I wonder if we can make a `Resource` transport that can accept Stdout / Stderr or any other resources.

Does it make sense?

chuck: Ah, I follow, and yes, that makes sense. I'm going to make all of the output valid json so that you can slice it or pretty print it with jq or standard tools. I think `Stream` makes sense then as the name (http://php.net/manual/en/book.stream.php).

chuck:
@kevinlebrun Updated to be a general stream transport, fully json, etc. Thanks for the feedback!

PR #44, opened 2018-07-26
Fix: Actually make use of anchor

localheinz:
This PR

* [x] actually makes use of a previously introduced anchor in `.circleci/config.yml`

Follows #41.


localheinz:
Thank you, @chuck!

PR #46, opened 2018-08-01
Fix: Improve use of anchors and references

localheinz:
This PR

* [x] attempts to improve the usages of anchors and references in the CircleCI configuration

💁‍♂️ Not sure if you prefer, I find the current usage a bit confusing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,36 +1,57 @@
 version: 2.0
 
 aliases:
+  - &CACHE_COMPOSER_KEY
+    key: 'betav1-composer-deps-{{ .Environment.CIRCLE_JOB }}-{{ checksum "composer.json" }}'
+
+  - &CACHE_NPM_KEY
+    key: 'betav1-lint-deps-{{ checksum "composer.json" }}'
+
+  - &IMAGE_DOCKER_DD_AGENT
+    image: datadog/docker-dd-agent
+    environment:
+      - DD_APM_ENABLED=true
+      - DD_BIND_HOST=0.0.0.0
+      - DD_API_KEY=invalid_key_but_this_is_fine
+
   - &STEP_COMPOSER_SELF_UPDATE
     run:
       name: Updating composer itself
       command: sudo composer self-update 1.6.5
-  - &docker_agent datadog/docker-dd-agent
-  - agent_env: &agent_env
-    - DD_APM_ENABLED=true
-    - DD_BIND_HOST=0.0.0.0
-    - DD_API_KEY=invalid_key_but_this_is_fine
-  - &load_composer_deps
-    key: 'betav1-composer-deps-{{ .Environment.CIRCLE_JOB }}-{{ checksum "composer.json" }}'
-  - &composer_install_deps
-    name: Composer installing dependencies
-    command: composer install -n
-  - &save_composer_deps
-    key: 'betav1-composer-deps-{{ .Environment.CIRCLE_JOB }}-{{ checksum "composer.json" }}'
-    paths:
-      - vendor/
-  - &test_results_path
-    path: test-results
-  - &php_test_defaults
+
+  - &STEP_COMPOSER_CACHE_RESTORE
+    restore_cache:
+      <<: *CACHE_COMPOSER_KEY
+
+  - &STEP_COMPOSER_CACHE_SAVE
+    save_cache:
+      <<: *CACHE_COMPOSER_KEY
+      paths:
+        - vendor/
+
+  - &STEP_COMPOSER_INSTALL
+    run:
+      name: Installing dependencies with composer
+      command: composer install -n
+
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,90 +1,114 @@
...
-    command: composer install -n
-  - &save_composer_deps
-    key: 'betav1-composer-deps-{{ .Environment.CIRCLE_JOB }}-{{ checksum "composer.json" }}'
-    paths:
-      - vendor/
-  - &test_results_path
-    path: test-results
-  - &php_test_defaults
+
+  - &CACHE_NPM_KEY
```
localheinz: Using upper-case names for anchors with consistent prefixes to make it easier to identify what they are for.

.circleci/config.yml
```diff
@@ -1,90 +1,114 @@
...
-      - restore_cache: &lint_deps_cache
-          key: 'betav1-lint-deps-{{ checksum "composer.json" }}'
-      - run: npm install eclint --no-package-lock --no-save
+      - <<: *STEP_COMPOSER_CACHE_RESTORE
+      - <<: *STEP_COMPOSER_INSTALL
+      - <<: *STEP_COMPOSER_CACHE_SAVE
+      - restore_cache:
+          <<: *CACHE_NPM_KEY
+      - run:
+          name: Installing dependencies with npm
```
localheinz: Consistently adding names to reveal the intent (if not obvious).

.circleci/config.yml
```diff
@@ -1,90 +1,114 @@
...
+    key: 'betav1-lint-deps-{{ checksum "composer.json" }}'
+
+  - &IMAGE_DOCKER_DD_AGENT
+    image: datadog/docker-dd-agent
+    environment:
+      - DD_APM_ENABLED=true
+      - DD_BIND_HOST=0.0.0.0
+      - DD_API_KEY=invalid_key_but_this_is_fine
+
+  - &STEP_COMPOSER_CACHE_RESTORE
```
localheinz: Extracting entire steps (which are used more often than once).

.circleci/config.yml
```diff
@@ -1,90 +1,114 @@
...
-    paths:
-      - vendor/
-  - &test_results_path
-    path: test-results
-  - &php_test_defaults
+
+  - &CACHE_NPM_KEY
+    key: 'betav1-lint-deps-{{ checksum "composer.json" }}'
+
+  - &IMAGE_DOCKER_DD_AGENT
```
localheinz: Extracting an entire object for the image to be used (so we only need to inject it, and do not need to repeat the environment separately).

.circleci/config.yml
```diff
@@ -1,90 +1,114 @@
...
           command: dockerize -wait tcp://127.0.0.1:8126 -timeout 1m
-      - run:
-          name: Run unit tests
-          command: composer test-unit -- --log-junit test-results/php-unit/results.xml
       - run:
           name: Run integration tests
           command: composer test-integration -- --log-junit test-results/php-integration/results.xml
-      - store_test_results: *test_results_path
-      - store_artifacts: *test_results_path
+      - run:
```
chuck: Was moving the unit tests after the integration tests intentional? Right now all of the tests are quite fast, but usually unit tests are faster, so I'd imagine we want them to run first so failures are faster (could probably also run them before starting the agent).

localheinz: @chuck 

No, that wasn’t intended. I’ll fix and let you know!

localheinz:
@chuck 

Unless I'm mistaken, the build failure is related to

* https://github.com/composer/composer/issues/7517

which already has a fix at 

* https://github.com/composer/xdebug-handler/pull/69

peter279k:
It seems that the CI build is broken and it should be fixed.

After checking the [CI build log](https://circleci.com/gh/DataDog/dd-trace-php/742?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), the PHPUnit use the unrecognized option `--ansi` when running the PHPUnit command.

I think it should remove that and run the command again 👍.

localheinz:
@chuck @peter279k 

I triggered a build again, should pass now.

Also see https://github.com/composer/composer/compare/1.7.0...1.7.1.

localheinz:
@chuck 

A specific version of composer can be installed by running

```
$ composer self-update <version>
```

but the bug is not fixed yet, see https://github.com/composer/xdebug-handler/pull/69.

chuck:
I see composer 1.7.1, and it doesn't have the issue. Will poke at CircleCI to see if they can update, or figure out how to install our own.

chuck:
@localheinz Apologies for the delay! Once #50 is merged, can you please rebase this? Should be good to go at that point.

localheinz:
@chuck 

Rebased!

localheinz:
Thank you, @chuck!

PR #47, opened 2018-08-01
Enhancement: Run phpstan on CircleCI

localheinz:
This PR

* [x] runs `phpstan` on CircleCi
* [x] fixes issues spotted on level 2

💁‍♂️ For reference, see https://github.com/phpstan/phpstan.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,6 +58,21 @@ jobs:
       - store_test_results: *test_results_path
       - store_artifacts: *test_results_path
 
+  "Static Analysis":
+    docker:
+      - image: circleci/php:7.2
+    steps:
+      - checkout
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -58,6 +58,21 @@ jobs:
...
+      - checkout
+      - restore_cache: *load_composer_deps
+      - run: *composer_install_deps
+      - save_cache: *save_composer_deps
+      - run:
+          name: Installing phpstan
+          command: composer require --dev phpstan/phpstan:~0.10.2
+      - run:
+          name: Running phpstan
+          command: vendor/bin/phpstan analyse --level=2 src
```
localheinz: If you like this we can increase the level (goes up to `7`, currently).

PR #48, opened 2018-08-01
Fix: Keep imports sorted

localheinz:
This PR

* [x] keeps imports sorted

💁‍♂️ I ran

```
$ composer global require friendsofphp/php-cs-fixer
```

and then

```
$ php-cs-fixer fix src --rules=ordered_imports
$ php-cs-fixer fix tests --rules=ordered_imports
```


localheinz:
Thank you, @chuck!

PR #51, opened 2018-08-15
Fix: Update phpstan/phpstan

localheinz:
This PR

* [x] updates `phpstan/phpstan`

💁‍♂️ For reference, see http://github.com/phpstan/phpstan/compare/0.10.2...0.10.3.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,7 +104,7 @@ jobs:
       - <<: *STEP_COMPOSER_CACHE_SAVE
       - run:
           name: Installing phpstan
-          command: composer require --dev phpstan/phpstan:~0.10.2
+          command: composer require --dev phpstan/phpstan:~0.10.3
       - run:
           name: Running phpstan
           command: vendor/bin/phpstan analyse --level=2 src


.circleci/config.yml
```diff
@@ -104,7 +104,7 @@ jobs:
       - <<: *STEP_COMPOSER_CACHE_SAVE
       - run:
           name: Installing phpstan
-          command: composer require --dev phpstan/phpstan:~0.10.2
+          command: composer require --dev phpstan/phpstan:~0.10.3
```
peter279k: How about setting the version is `^0.10`?
I think it cannot always set the current version and send PR.

localheinz: My change doesn't have an effect on the version currently installed, but it documents that we want to install at least `phpstan/phpstan:~0.10.3` - that's the primary purpose.

peter279k: Thank you for your explanation. The `phpstan` `0.10.3` version is the latest release version now.
I mean setting the `^0.10.3` version is better than `~0.10.3` version because it can clearly define the required version is `0.10.3+` at least.

chuck: Composer docs do say re: caret, "This is the recommended operator for maximum interoperability when writing library code." Both ways say 0.10.3+ at least though, and if I'm reading the docs right, with pre-1.0 code they're equivalent.

localheinz:
Thank you, @chuck!

PR #52, opened 2018-08-16
Enhancement: Update to latest version of composer

localheinz:
This PR

* [x] removes the version constraint to ensure we update to the latest version of `composer` itself on CircleCi

💁‍♂️ For reference, see https://github.com/composer/composer/compare/1.7.1...1.7.2.


localheinz:
Thank you, @chuck!

PR #54, opened 2018-08-30
Inherit the parent span's service unless explicitly overridden

chuck:
Otherwise you have to keep explicitly passing the service to not revert back to SAPI_NAME.

This feels much harder than it needed to be but StartSpanOptions is declared as final by the PHP OpenTracing class and doesn't provide any way to set a tag after instantiation either.

### DIFF
--- a/src/DDTrace/Tracer.php
+++ b/src/DDTrace/Tracer.php
@@ -5,6 +5,7 @@ namespace DDTrace;
 use DDTrace\Encoders\Json;
 use DDTrace\Propagators\Noop as NoopPropagator;
 use DDTrace\Propagators\TextMap;
+use DDTrace\Tags;
 use DDTrace\Transport\Http;
 use DDTrace\Transport\Noop as NoopTransport;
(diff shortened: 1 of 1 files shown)


jcchavezs:
The service name for tracer can be declared in https://github.com/DataDog/dd-trace-php/blob/master/src/DDTrace/Tracer.php#L44. Since a span should be created with the tracer it will get the service name from tracer and all child spans will do so. Inheriting from the span can become cumbersome as the tracer set that value too. WDYT @chuck ?

pawelchcki:
A Stacked approach is what we also do in other Integrations. And as @chuck  mentioned one service might encompass another.

jcchavezs:
I see. I thought the `service_name` was describing the `service` and not the component. Then it makes sense to do such a thing. Since this is something very specific from DD I think this is the right approach.

chuck:
@pawelchcki Can you approve this now that it has tests please? Thanks!

PR #55, opened 2018-08-31
C extension allowing overriding any method or function

pawelchcki:
This PR contains C extension written to allow overriding any method or function.

Currently it does so by injecting itself into processing every "execute function" opcode handled by PHP, and then checking if the function  (or method) to be executed was configured to be overridden.

If it was configured to be overridden, then configured overriding closure is executed instead of the original function.

This means that every function call - even if its not overridden by this extension -  has  an additional overhead of at least 1 hash lookup. For now this is necessary overhead, in future we could explore some optimizations like opcode rewriting or caching information within opcode itself.

This extension uses standard PHP extension facilities (`phpize`) to build and test the code - and it is able to check for memory leaks within tests using Valgrind.

CI configuration introduced in this PR, also runs tests and checks for leaks. It then pushes compiled extension via workspace to downstream builds, and configures composer in those builds to use compiled extension automatically. This way all composers tests already have access to this extension.


tests/ext/namespaces.phpt
```diff
@@ -0,0 +1,183 @@
+--TEST--
+Verify Multiple functions and methods will be instrumented successfully
```
cgilmour: Needs an update?

tests/ext/recursion.phpt
```diff
@@ -0,0 +1,165 @@
+--TEST--
+Verify Multiple functions and methods will be instrumented successfully
```
cgilmour: Needs an update?

tests/ext/simple_function_hook.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+Check used defined function can be overwritten and we're able to call the original
```
cgilmour: `user`? `overridden`?

src/ext/dispatch_setup.c
```diff
@@ -0,0 +1,124 @@
...
+static void php_execute(zend_execute_data *execute_data TSRMLS_DC) {
+    if (ddtrace_original_execute_ex) {
+        ddtrace_original_execute_ex(execute_data TSRMLS_CC);
+    } else execute_ex(execute_data TSRMLS_CC);
+}
+
+void ddtrace_dispatch_init(){
+    /**
+    * Replacing zend_execute_ex with anything other than original
+    * hanges some of the bevavior in PHP compilation and execution
```
willgittoes-dd: changes

src/ext/dispatch_compat_php7.c
```diff
@@ -0,0 +1,51 @@
...
+    ALLOC_HASHTABLE(class_lookup);
+    zend_hash_init(class_lookup, 8, NULL, ddtrace_class_lookup_free, 0);
+    zend_hash_update_ptr(&DDTRACE_G(class_lookup), clazz->name, class_lookup);
+
+    return class_lookup;
+}
+
+zend_bool ddtrace_dispatch_store(HashTable *lookup,  ddtrace_dispatch_t *dispatch_orig) {
+    ddtrace_dispatch_t *dispatch = pemalloc(sizeof(ddtrace_dispatch_t), lookup->u.flags & HASH_FLAG_PERSISTENT);
+    dispatch = pemalloc(sizeof(ddtrace_dispatch_t), lookup->u.flags & HASH_FLAG_PERSISTENT);
```
willgittoes-dd: Why twice?

src/ext/ddtrace.c
```diff
@@ -0,0 +1,171 @@
...
+
+static PHP_RINIT_FUNCTION(ddtrace)
+{
+    UNUSED(module_number);
+    UNUSED(type);
+
+#if defined(ZTS) && PHP_VERSION_ID >= 70000
+    ZEND_TSRMLS_CACHE_UPDATE();
+#endif
+    if (DDTRACE_G(disable)) {
```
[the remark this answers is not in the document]
pawelchcki: I think its safer to update `ZEND_TSRMLS_CACHE_UPDATE()` in both cases

src/ext/dispatch_compat_php5.c
```diff
@@ -0,0 +1,128 @@
...
+zend_bool ddtrace_dispatch_store(HashTable *lookup, ddtrace_dispatch_t *dispatch_orig) {
+    ddtrace_dispatch_t *dispatch = pemalloc(sizeof(ddtrace_dispatch_t), lookup->persistent);
+
+    memcpy(dispatch, dispatch_orig, sizeof(ddtrace_dispatch_t));
+    DD_PRINTF("%s, %d", Z_STRVAL_P(dispatch->function), Z_STRLEN_P(dispatch->function));
+
+    return zend_hash_update(lookup, Z_STRVAL_P(dispatch->function), Z_STRLEN_P(dispatch->function), &dispatch, sizeof(ddtrace_dispatch_t *), NULL) == SUCCESS;
+}
+
+// restore EX
```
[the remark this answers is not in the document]
pawelchcki: Well someone could include `.c` file :D but you're right I think we're safe here :)

tests/ext/simple_function_hook.phpt
pawelchcki: 👍

src/ext/dispatch_compat_php7.c
pawelchcki: Oh - thanks for finding it. I put that here as a canary to check if Valgrind actually could find leaks. But forgot about it.  Or I'm making excuses for myself for writing leaky code :).

I hoped Valgrind checks would have caught this but since the CI passes...

pawelchcki:
Thanks @willgittoes-dd and @cgilmour for review. I know there are a lot of PHP internals here, while I've tried to limit whats necessary to implement this - the PHP 5.6 part is especially scary as it parses some not well exposed internals (like arg stack). 

The 7.2 is much more nicer in that regard as the API there is much more prepared for the kind of modifications we're doing here.

In any case - I've addressed all your suggestions, and added `clang-format` check to the CI.

While PHP builtin Valgrind checks seems to show its all good, to be sure I've coerced this config of Valgrind locally to show more possible memory leaks and it looks that there are only false positives detected.

In separate PR I'd like to introduce some CI checks using `address sanitizer` to make double sure we're not missing anything nasty.

This PR should be now ready for another round of review - let me know what do you think about it.

bin/docker_test.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
chuck: What's this file for? Naive search I don't see it used anywhere.

pawelchcki:
Thanks @willgittoes-dd. 
I've added the comments to the `#endifs` of include guards. Also I had to adjust the `clang-format` to scan `.c` files and not `.cpp` :)

This PR is ready for more review /cc: @dd-caleb @willgittoes-dd @chuck

src/DDTrace/Scope.php
```diff
@@ -5,7 +5,7 @@
 use OpenTracing\Scope as OpenTracingScope;
 use OpenTracing\Span as OpenTracingSpan;
 
-final class Scope implements OpenTracingScope
+class Scope implements OpenTracingScope
```
chuck: What's this change for?

src/ext/ddtrace.c
```diff
@@ -0,0 +1,169 @@
...
+#else
+#define PHP5_UNUSED(...) /* unused unused */
+#endif
+
+ZEND_DECLARE_MODULE_GLOBALS(ddtrace)
+
+#define ddtrace_disabled_guard()                                                               \
+    do {                                                                                       \
+        if (DDTRACE_G(disable)) {                                                              \
+            zend_throw_exception_ex(spl_ce_RuntimeException, 0 TSRMLS_CC,                      \
```
chuck: What's the use case for disabling ddtrace where we want to interrupt execution with an exception? Or am I interpreting this incorrectly?

src/ext/ddtrace.c
```diff
@@ -0,0 +1,169 @@
...
+                                    "ddtrace is disabled by configuration (ddtrace.disable)"); \
+            return;                                                                            \
+        }                                                                                      \
+    } while (0)
+
+PHP_INI_BEGIN()
+STD_PHP_INI_ENTRY("ddtrace.disable", "0", PHP_INI_SYSTEM, OnUpdateBool, disable, zend_ddtrace_globals, ddtrace_globals)
```
chuck: I assume this is around as a quick way of making all calls a noop? Is it better than just removing the extension, since this is all it does?

src/ext/dispatch_setup.c
```diff
@@ -0,0 +1,125 @@
...
+
+void ddtrace_dispatch_init() {
+/**
+ * Replacing zend_execute_ex with anything other than original
+ * changes some of the bevavior in PHP compilation and execution
+ *
+ * e.g. it changes compilation of function calls to produce ZEND_DO_FCALL
+ * opcode instead of ZEND_DO_UCALL for user defined functions
+ *
+ * While this extension could be developed by using zend_execute_ex to hook
```
chuck: nit - remove While? or add a comma instead of period.

bin/docker_test.sh
pawelchcki: For now I'm using it run tests in docker locally. I'll drop it from this PR, I'll figure it out in another PR what to do with it.

src/DDTrace/Scope.php
pawelchcki: Sorry, I was investigating some errors I got when trying to run PHPUnit testcases for the first time. And its a leftover of this investigating. I'll revert it.
 Thanks for finding that!

src/ext/ddtrace.c
pawelchcki: Hm, this is a pattern in quite a few extensions I've seen that I followed without a second thought. 
Many PHP extensions can be compiled/statically linked into PHP itself then you need this switch to be able to enable/disable it. 

I think this extension should be able to be compiled into PHP as is, I remember adding some ifdefs to actually support it being a module.

chuck: Ah, that makes sense I guess. No problem leaving it.

src/ext/ddtrace.c
pawelchcki: It was meant to guard against accidentally using `dd_trace` and expecting it to work when its actually disabled. But after second thought if we silently NOOP this instead we would be able to avoid potential unwanted uncaught exceptions when someone simply disables the exception.

I'll change the implementation to not throw any exception and just do NOOP.

pawelchcki:
Thanks for review @chuck!

PR #56, opened 2018-09-14
Send system.pid with the root span to match python tracer

chuck:
Small step towards parity with other tracers


chuck:
@pawelchcki Added tests for parent and child.

PR #57, opened 2018-09-14
Set Datadog-Meta-Lang headers similar to python tracer

chuck:
Another small parity PR


jcchavezs:
There is a simpler approach for do assertions about headers or payloads in a request we used in `opencensus`: https://github.com/census-instrumentation/opencensus-php/pull/190.

Basically it launches a server with a handler function where you can set expectations. The library used is https://github.com/jcchavezs/httptest-php

What do you think @chuck @pawelchcki?

chuck:
Got my approach working; kind of similar to @jcchavezs's but more basic. Can look at that more tomorrow. @pawelchcki This at least gets things working and tested (along with fixing the pre-existing setHeader() bug).

tests/Integration/Transport/HttpTest.php
```diff
@@ -95,4 +97,56 @@ public function testSilentlySendTraces()
 
         $this->assertSame('', $output);
     }
+
+    public function testSendsMetaHeaders()
+    {
+        $process = new Process("exec php -S localhost:8500 -t " . __DIR__ . "/request_replayer");
```
pawelchcki: I think it would be good to move this to separate helper funcitons

tests/Integration/Transport/HttpTest.php
```diff
@@ -95,4 +97,56 @@ public function testSilentlySendTraces()
...
+        $httpTransport->send($traces);
+
+        $traceRequest = json_decode(file_get_contents('http://localhost:8500/replay'), true);
+
+        $this->assertEquals('php', $traceRequest['headers']['Datadog-Meta-Lang']);
+        $this->assertEquals(\PHP_VERSION, $traceRequest['headers']['Datadog-Meta-Lang-Version']);
+        $this->assertEquals(\PHP_SAPI, $traceRequest['headers']['Datadog-Meta-Lang-Interpreter']);
+        $this->assertEquals(Version\VERSION, $traceRequest['headers']['Datadog-Meta-Tracer-Version']);
+
+        $process->stop(0);
```
pawelchcki: This should probably also be abstracted away.

tests/Integration/Transport/HttpTest.php
```diff
@@ -95,4 +97,56 @@ public function testSilentlySendTraces()
...
+    }
+
+    public function testSetHeader()
+    {
+        $process = new Process("exec php -S localhost:8500 -t " . __DIR__ . "/request_replayer");
```
pawelchcki: Ditto

tests/Integration/Transport/HttpTest.php
```diff
@@ -95,4 +97,56 @@ public function testSilentlySendTraces()
...
+        $process->start();
+        usleep(100000);
+
+        $httpTransport = new Http(new Json(), null, [
+            'endpoint' => 'http://localhost:8500/test-trace',
```
pawelchcki: I feel like this URL hostname + port should also be abstracted away, so that all those magic strings and numbers would live in only one place.

tests/Integration/Transport/HttpTest.php
chuck: I didn't abstract the request details itself just yet - I can see getting there, but not sure it's necessary so far.

chuck:
Moved the process management and URLs into a separate helper class.

PR #61, opened 2018-10-04
Add lighter downloads when installing from composer

SammyK:
When you install `dd-trace-php` with composer, all the files in the repo (including the tests and c files) are copied into the `vendor/` directory.

screen shot 2018-10-04 at 10 18 09 am

This PR introduces a `.gitattributes` file to "blacklist" files that should not be downloaded when installing `dd-trace-php` with composer. This makes for a much lighter download.

screen shot 2018-10-04 at 10 30 36 am

One caveat that we'd need to discuss is that this change also impacts `git archive`:

```bash
git archive --format zip --output ../export.zip [branch-name]
```

In this example, `export.zip` would also not contain any of the blacklisted files. This also applies to downloading a zip of the repo from GitHub. So the only way to get the `ddtrace` extension that was just added in #55, would be to do a `git clone` which I'm not sure is the desired outcome.

For this reason, I'd love to get your thoughts on moving the `ddtrace` extension to its own separate repo if possible. :)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,14 @@
+tests               export-ignore
+.editorconfig       export-ignore
+.gitattributes      export-ignore
+.gitignore          export-ignore
+.phpcs.xml          export-ignore
+CONTRIBUTING.md     export-ignore
+phpunit.xml         export-ignore
+RELEASING.md        export-ignore
+
+# ext/ddtrace
+.circleci           export-ignore
+src/ext             export-ignore
+.clang-format       export-ignore
+config.m4           export-ignore


.gitattributes
```diff
@@ -0,0 +1,15 @@
+tests               export-ignore
+.editorconfig       export-ignore
+.gitattributes      export-ignore
+.gitignore          export-ignore
+.phpcs.xml          export-ignore
+CONTRIBUTING.md     export-ignore
+phpunit.xml         export-ignore
+README.md           export-ignore
```
pawelchcki: What do you think about allowing `README.md` to be installed along with PHP code?

SammyK: Good call. It never hurts to always distribute the docs! Fixed. :)

pawelchcki:
> In this example, export.zip would also not contain any of the blacklisted files. This also applies to downloading a zip of the repo from GitHub. So the only way to get the ddtrace extension that was just added in #55, would be to do a git clone which I'm not sure is the desired outcome.

We're still looking into the right way to distribute the extension itself. Right now a strong contender is building a lot of rpm/deb packages for many different platforms. PECL might also be an option - not sure how welcome we will be there.

> For this reason, I'd love to get your thoughts on moving the ddtrace extension to its own separate repo if possible. :)
Personally I like to keep relevant code close together - especially during early stages of development. Also it seems for now this setup is simpler WRT to PHPUnit test using the extension. At least until we'll start publishing the extension one way or another.

However I think eventually ddtrace extension will end up in a separate repo, there are a few compelling arguments to keep it separate - one of them you've already mentioned :)

PR #65, opened 2018-10-09
update README

chuck:
make language a bit less scary

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 PHP Tracer
 
-This is alpha software. We recommend using it staging.
+This is alpha software. We do not recommend using it in production yet.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,9 +4,9 @@
 [![OpenTracing Badge](https://img.shields.io/badge/OpenTracing-enabled-blue.svg)](http://opentracing.io)
 [![Minimum PHP Version](https://img.shields.io/badge/php-%3E%3D%205.6-8892BF.svg)](https://php.net/)
 
-Experimental PHP Tracer!
+PHP Tracer
 
-This project is experimental and under active development. Use it at your own risk.
+This is alpha software. We recommend using it staging.
```
rochdev: The wording feels a bit weird, like we recommend to only use it in staging. Maybe something simpler like "This is alpha software and is not recommended for production use."?

rochdev:
IMHO yes :)

PR #66, opened 2018-10-09
[integrations] add Datadog integrations to the base Tracer

chuck:
### Overview

This PR adds the following integration to provide an OOTB tracing experience:
* Eloquent
* Laravel
* Memcached
* PDO
* Redis client (`predis`)
* Symfony


src/DDTrace/Integrations/Memcached.php
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    public static function traceCas($memcached, $args)
+    {
+        $scope = GlobalTracer::get()->startActiveSpan('Memcached.cas');
+        $span = $scope->getSpan();
+        $span->setTag(Tags\SPAN_TYPE, Types\MEMCACHED);
+        $span->setTag(Tags\SERVICE_NAME, 'memcached');
+        $span->setTag('memcached.command', 'cas');
+        $span->setTag('memcached.cas_token', $args[0]);
```
pawelchcki: This tag will probably cause unbounded cardinality issue

src/DDTrace/Integrations/Memcached.php
```diff
@@ -0,0 +1,358 @@
...
+
+        $key = $args[1];
+        self::setServerTagsByKey($span, $memcached, $key);
+        $span->setTag('memcached.query', "cas $key");
```
pawelchcki: I think $key can be quite variable, and cause cardinality issues as well

src/DDTrace/Integrations/Memcached.php
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    public static function traceCasByKey($memcached, $args)
+    {
+        $scope = GlobalTracer::get()->startActiveSpan('Memcached.casByKey');
+        $span = $scope->getSpan();
+        $span->setTag(Tags\SPAN_TYPE, Types\MEMCACHED);
+        $span->setTag(Tags\SERVICE_NAME, 'memcached');
+        $span->setTag('memcached.command', 'casByKey');
+        $span->setTag('memcached.cas_token', $args[0]);
```
pawelchcki: cardinality ?

src/DDTrace/Integrations/PDO.php
```diff
@@ -0,0 +1,230 @@
...
+     */
+    public static function load()
+    {
+        if (!extension_loaded('ddtrace')) {
+            trigger_error('ddtrace extension required to load PDO integration.', E_USER_WARNING);
+            return;
+        }
+
+        // public PDO::__construct ( string $dsn [, string $username [, string $passwd [, array $options ]]] )
+        dd_trace('PDO', '__construct', function (...$args) {
```
pawelchcki: While using `PDO::_construct` to setup necessary data to be used later i.e. `PDO::storeConnectionParams($this, $args);` is fine. However I wonder if there is any added value from creating a span around a constructor. 
Is this constructor doing some IO ?

src/DDTrace/Integrations/Predis.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        // public mixed Predis\Client::executeCommand(CommandInterface $command)
+        dd_trace(Client::class, 'executeCommand', function ($command) {
+            $query = Predis::formatCommandArgs($command);
+            $scope = GlobalTracer::get()->startActiveSpan('Predis.executeCommand');
+            $span = $scope->getSpan();
+            $span->setTag(Tags\SPAN_TYPE, Types\REDIS);
+            $span->setTag(Tags\SERVICE_NAME, 'redis');
+            $span->setTag('redis.raw_command', $query);
```
pawelchcki: It might be worthwhile to sanitize this tag to avoid cardinality and data leak problems

src/DDTrace/Integrations/LaravelProvider.php
```diff
@@ -0,0 +1,137 @@
...
+        $scope->getSpan()->setTag(Tags\SERVICE_NAME, $this->getAppName());
+
+        // Name the scope when the route matches
+        $this->app['events']->listen(RouteMatched::class, function (RouteMatched $event) use ($scope) {
+            $span = $scope->getSpan();
+            $span->setResource($event->route->getActionName() . ' ' . Route::currentRouteName());
+            $span->setTag('laravel.route.name', Route::currentRouteName());
+            $span->setTag('laravel.route.action', $event->route->getActionName());
+            $span->setTag('http.method', $event->request->method());
+            $span->setTag('http.url', $event->request->path());
```
labbati: Why we store only the path (`/some/relative/patch`) in a tag named `http.url` instead of the `$event->request->url()` `https://example.com/some/relative/path`?

src/DDTrace/Integrations/Memcached.php
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Memcached::getServerByKey() /might/ return incorrect information if the
+     * distribution would be rebuilt on a real call (Memcached::get(),
+     * Memcached::getByKey(), and other commands that actually hit the server
+     * include logic to regenerate the distribution if a server has been ejected
+     * or if a timer expires; Memcached::getServerByKey() does not check for the
+     * distribution being rebuilt. Getting around that would likely be
+     * prohibitively expensive though.
+     */
+    public static function setServerTagsByKey($span, $memcached, $key)
```
labbati: Should this be private or is intended to be used by devs?

src/DDTrace/Integrations/PDO.php
```diff
@@ -0,0 +1,230 @@
...
+        // public PDOStatement PDO::query(string $query, int PDO::FETCH_COLUMN, int $colno)
+        // public PDOStatement PDO::query(string $query, int PDO::FETCH_CLASS, string $classname, array $ctorargs)
+        // public PDOStatement PDO::query(string $query, int PDO::FETCH_INFO, object $object)
+        // public int PDO::exec(string $query)
+        dd_trace('PDO', 'query', function (...$args) {
+            $scope = GlobalTracer::get()->startActiveSpan('PDO.query');
+            $span = $scope->getSpan();
+            $span->setTag(Tags\SPAN_TYPE, Types\SQL);
+            $span->setTag(Tags\SERVICE_NAME, 'PDO');
+            $span->setResource($args[0]);
```
labbati: Just for confirmation, can you confirm that  because the type is sql we will have the agent to "sanitize" it to both avoid data leaks and cardinality issues?

src/DDTrace/Integrations/PDO.php
```diff
@@ -0,0 +1,230 @@
...
+            PDO::setConnectionTags($this, $span);
+
+            try {
+                $result = $this->query(...$args[0]);
```
labbati: I think that here what we want to do is `$result = $this->query(...$args)` instead of only the first arg.

Otherwise we are not passing all the arguments:

```
function wrapper(...$args) {
    print_r("wrapper\n");
    print_r($args);
    query(...$args[0]);
}

function query(...$args) {
    print_r("query\n");
    print_r($args);
}

wrapper(1, 2, 3);

```

prints

```
wrapper
Array
(
    [0] => 1
    [1] => 2
    [2] => 3
)
PHP Warning:  Only arrays and Traversables can be unpacked in /Users/luca.abbati/projects/datadog/dd-trace-php/playground.php on line 7

Warning: Only arrays and Traversables can be unpacked in /Users/luca.abbati/projects/datadog/dd-trace-php/playground.php on line 7
query
Array
(
)

```

src/DDTrace/Integrations/PDO.php
```diff
@@ -0,0 +1,230 @@
...
+            }
+        });
+
+        // public PDOStatement PDO::prepare ( string $statement [, array $driver_options = array() ] )
+        dd_trace('PDO', 'prepare', function (...$args) {
+            $scope = GlobalTracer::get()->startActiveSpan('PDO.prepare');
+            $span = $scope->getSpan();
+            $span->setTag(Tags\SPAN_TYPE, Types\SQL);
+            $span->setTag(Tags\SERVICE_NAME, 'PDO');
+            $span->setResource($args[0]);
```
labbati: If is confirmed that the agent clears all the sql then this looks fine to me.

src/DDTrace/Integrations/PDO.php
pawelchcki: `$span->setTag(Tags\SPAN_TYPE, Types\SQL)` Ensures that this is sanitized in the Agent.

src/DDTrace/Integrations/PDO.php
chuck: Good catch, thanks!

src/DDTrace/Integrations/Memcached.php
chuck: I think I had it public when it was called directly from a closure and missing class scope. All of the calls are from within the integration class now though, so it can be private. Changed.

src/DDTrace/Integrations/LaravelProvider.php
chuck: I think I was following an example from dogweb traces here, but happy either way. I'll change it for now.

src/DDTrace/Integrations/LaravelProvider.php
```diff
@@ -0,0 +1,137 @@
...
+        // Creates a tracer with default transport and default encoders
+        $tracer = new Tracer(new Http(new Json()));
+
+        // Sets a global tracer (singleton). Also store it in the Laravel
+        // container for easy Laravel-specific use.
+        GlobalTracer::set($tracer);
+        $this->app->instance('datadog.tracer', $tracer);
+
+        // Trace middleware
+        dd_trace(Pipeline::class, 'through', function ($pipes) {
```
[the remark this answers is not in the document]
chuck: I defer to you on this; I read that guideline as, "don't do anything in register() that could possibly rely on another service". It's true that we're not guaranteed order of register() calls, but it does make sure that tracing is set up before anything starts happening in other providers' boot() calls.

If you still think we should change to boot() I can update it.

src/DDTrace/Integrations/LaravelProvider.php
```diff
@@ -0,0 +1,137 @@
...
+            return;
+        }
+
+        // Creates a tracer with default transport and default encoders
+        $tracer = new Tracer(new Http(new Json()));
+
+        // Sets a global tracer (singleton). Also store it in the Laravel
+        // container for easy Laravel-specific use.
+        GlobalTracer::set($tracer);
+        $this->app->instance('datadog.tracer', $tracer);
```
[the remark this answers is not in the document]
chuck: That makes sense; do you think there's a good reason to keep the datadog.tracer alias?

src/DDTrace/Integrations/LaravelProvider.php
```diff
@@ -0,0 +1,137 @@
...
+        if (php_sapi_name() == 'cli') {
+            return;
+        }
+
+        // Creates a tracer with default transport and default encoders
+        $tracer = new Tracer(new Http(new Json()));
+
+        // Sets a global tracer (singleton). Also store it in the Laravel
+        // container for easy Laravel-specific use.
+        GlobalTracer::set($tracer);
```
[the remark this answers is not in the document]
chuck: My approach here was to keep all of the integration code consistent, especially since the Laravel integration uses several others - PDO, Memcached, Predis, etc. - that can also be used outside of Laravel or from Symfony. So we need GlobalTracer or another solution that's consistent between each integration. I added the tracer to the Laravel container to make it easy to write extensions or add instrumentation inside of controllers or whatnot in a Laravel-ish way.

src/DDTrace/Integrations/Predis.php
chuck: Are these not sanitized in the agent?

src/DDTrace/Integrations/PDO.php
chuck: Yes, it actually connects when you call __construct, so this span captures connection time as well as potential connection errors:

screen shot 2018-10-10 at 1 06 59 pm

chuck:
Squashed the set of commits from when I was trying to get CI to work, meant to do that earlier. Won't force-push anymore without specific need.

src/DDTrace/Integrations/Predis.php
```diff
@@ -0,0 +1,120 @@
...
+            $span->setTag('redis.args_length', count($command->getArguments()));
+            $span->setResource($query);
```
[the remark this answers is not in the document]
chuck: Same question as on sanitizing, my naive understanding was that the agent takes care of this by quantizing, like with SQL queries. If that's not the case what do we want the resource to be here, just the actual command name?

chuck:
I believe I've responded to all of the feedback so far. Thanks!

src/DDTrace/Integrations/Predis.php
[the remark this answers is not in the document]
pawelchcki: Oh, yeah. Sorry for confusion, in Ruby a lot of those were still quantized manually. Looking at the code of https://github.com/DataDog/datadog-trace-agent/blob/master/obfuscate/redis.go
looks like as long as type is set correctly it will be quantized/obfuscated.

PR #74, opened 2018-10-18
Fix case sensitive lookup bug

pawelchcki:
Accidentally we were doing case sensitive lookup in the hash


tests/ext/case_insensitive_method_hook.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+Check if we can override method from a parent class usingcase insensiteve matching
```
chuck:
```suggestion
Check if we can override method from a parent class using case insensitive matching
```

chuck: One line spelling fix, otherwise looks good.

pawelchcki:
Thanks for review @chuck. Applied your suggestion - ready for another review round/ merge :)

PR #76, opened 2018-10-18
Initial version of installation guide

pawelchcki:


docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+```
+
+### Enabling tracing
+
+#### Laravel integration
+
+To enable Laravel integration we need to configure a new Provider in `config/app.php`
+
+```php
+   'providers' => array_merge(include(base_path('modules/system/providers.php')), [
```
labbati: I feel like we should avoid this `array_merge` detail, which may deviate attention from the line below. Even if you create a new Laravel app the providers definition is a plain array and I think that this is what most developers expect.

```suggestion
   'providers' => [
```

docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+
+For Symfony applications, add the bundle in app/AppKernel.php:
+
+
+```php
+    $bundles = [
+        // ...
+        new DDTrace\Integrations\SymfonyBundle(),
+        // ...
+        ];
```
labbati:
```suggestion
    ];
```

docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+   'providers' => array_merge(include(base_path('modules/system/providers.php')), [
+# .....
+     'DDTrace\Integrations\LaravelProvider',
+```
+
+Now your Laravel application should start sending traces to the Datadog agent running on localhost (in default configuration). The Datadog agent must have APM enabled; see https://docs.datadoghq.com/tracing/setup/ for instructions on installing and configuring the agent.
+
+#### Symfony integration
+
+For Symfony applications, add the bundle in app/AppKernel.php:
```
labbati: Looking at Symfony [docs for bundles](https://symfony.com/doc/current/bundles.html) seems like the file they are suggesting is the  `config/bundles.yml` (and this is where I used to config my bundles actually). Unless is there a specific reason (in which case we should mention it here), shouldn't we propose that approach. I mean, it still works, but maybe what we showing is a less common way to register bundles.

docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+
+
+```php
+    $bundles = [
+        // ...
+        new DDTrace\Integrations\SymfonyBundle(),
```
labbati: This comment is a sort of continuation of the previous one. In that case it would be the class reference and not an instance

docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+        // ...
+        ];
+```
+
+#### Adding tracing to a function or method
+
+To add spans for another function or method, you can use `dd_trace` to open a span before the code executes, close it when it’s done, set additional tags or errors on the span, and for more advanced usage, modify either the arguments or the return value.
+
+Here’s an example of tracing the `CustomDriver` class’s `doWork` method, including noting and re-throwing exceptions:
```
labbati: `noting`?

pawelchcki: "Note-ing" it threw me off as well, when first read it :)

docs/getting_started.md
pawelchcki: @chuck ^^ ?

labbati: 👍 give me a few minutes to setup an app locally.

docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+Datadog’s PHP extension (`ddtrace`) allows introspection of arbitrary PHP code.
+
+At this moment it is only distributed in source code form, and requires manual compilation.
+
+### Compiling and installing the extension manually
+
+```bash
+mkdir dd-trace
+cd dd-trace
+curl -L https://github.com/DataDog/dd-trace-php/archive/v0.2.2.tar.gz | tar x --strip-components=1
```
labbati: @pawelchcki and @chuck will also update this to `0.2.3` ack?

chuck: v0.2.3 ;)

docs/getting_started.md
labbati:
```suggestion
For Symfony applications, add the bundle in `config/bundles.php`:
```

docs/getting_started.md
```diff
@@ -0,0 +1,142 @@
...
+
+Now your Laravel application should start sending traces to the Datadog agent running on localhost (in default configuration). The Datadog agent must have APM enabled; see https://docs.datadoghq.com/tracing/setup/ for instructions on installing and configuring the agent.
+
+#### Symfony integration
+
+For Symfony applications, add the bundle in app/AppKernel.php:
+
+
+```php
+    $bundles = [
```
labbati:
```suggestion
    return [
```

docs/getting_started.md
labbati:
```suggestion
        DDTrace\Integrations\SymfonyBundle::class => ['all' => true],
```

docs/getting_started.md
labbati:
```suggestion
curl -L https://github.com/DataDog/dd-trace-php/archive/v0.2.3.tar.gz | tar x --strip-components=1
```

docs/getting_started.md
labbati:
```suggestion
Here’s an example of tracing the `CustomDriver` class’s `doWork` method, including reporting any exceptions as errors on the span and re-throwing them:
```

PR #77, opened 2018-10-19
Standardize import paths

pawelchcki:
Current import paths, caused the extension not to compile successfully on Alpine3.4 and Ubuntu16.04

This PR fixes that problem.


SammyK:
I was about to send a PR for this very thing! Glad you caught this too. :)

PR #81, opened 2018-10-22
Redis integration + integration tests basics

pawelchcki:
Initial implementation of docker compose

- experimental Makefile (probably could be 95% converted to composer scripts)


docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+      - redis_integration
+      - ddagent_integration
+    environment:
+      - REDIS_HOSTNAME=redis_integration
+      - DDAGENT_HOSTNAME=ddagent_integration
+    command: make -f DDMakefile install test_integration SUDO=sudo
+  '5.6': { <<: *base, image: 'circleci/php:5.6' }
+  '7.0': { <<: *base, image: 'circleci/php:7.0' }
+  '7.1': { <<: *base, image: 'circleci/php:7.1' }
+  redis_integration:
```
labbati: [Minor] does the `_integration` suffix add any specification to this service? In other words, there will be any `redis` machine?
I would just write `redis` but this is just a personal preference for no apparent reason.

src/DDTrace/Integrations/Predis.php
```diff
@@ -20,6 +20,18 @@ class Predis
      */
     private static $connections = [];
 
+    public static function unload()
```
labbati: This methods seems to be not called, did you maybe intend to call in the integration test's `tearDownAfterClass`?

docker-compose.yml
pawelchcki: My Idea was that `redis_integration` doesn't expose any ports. 
And then to have other `redis` to be used locally if need be.

I don't have a strong opinion about this. I'll change it to `redis`. And we'll see about exposing ports later

src/DDTrace/Integrations/Predis.php
pawelchcki: Ups sorry, this is a leftover from my work on unloading extensions from the data structures. Removing

tests/Integration/Integrations/PredisTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+namespace DDTrace\Tests\Integration\Integrations;
+
+use DDTrace\Encoders\Json;
+// use DDTrace\Tests\RequestReplayer;
+use DDTrace\Tracer;
+use DDTrace\Transport\Http;
+// use DDTrace\Version;
+use DDTrace\Integrations\Predis;
```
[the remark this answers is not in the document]
pawelchcki: Yeah sorting alphabetically sounds like fun :) (if it can be done automatically)

However I'd much prefer to have `php-cs-fixer`  do it. It should be possible, lets add that in separate PR, as it might trigger reformat in the rest of the code.

tests/Integration/Integrations/PredisTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+// use Predis\Client;
+use PHPUnit\Framework;
```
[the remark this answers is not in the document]
pawelchcki: No particular reason :D

docker-compose.yml
labbati: Oh that makes sense. Please leave it here!

tests/Integration/Integrations/PredisTest.php
```diff
@@ -0,0 +1,41 @@
...
+// use Prophecy\Argument;
+// use Psr\Log\LoggerInterface;
+use OpenTracing\GlobalTracer;
+
+
+final class PredisTest extends Framework\TestCase
+{
+    public static function setUpBeforeClass()
+    {
+        $tracer = new Tracer(new Http(new Json()));
```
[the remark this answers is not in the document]
pawelchcki: Agreed

pawelchcki:
Thanks for the review! @labbati

PR #84, opened 2018-10-24
Use test runners from docker library

labbati:


.circleci/config.yml
```diff
@@ -56,10 +56,23 @@ aliases:
...
+      - run:
+          name: testresults dir
+          command: mkdir test-results
       - run:
           name: export .env.circleci
           command: |
             echo "export $(cat .env.circleci | xargs)" >> $BASH_ENV
+      - run:
+          name: Run extension tests
+          command: TEST_PHP_JUNIT=test-results/simle_test.xml REPORT_EXIT_STATUS=1 make -f DDMakefile test_c
```
pawelchcki: Actually since we do not need to install the Valgrind anymore this step is no longer necessary. Previously I've done it twice to avoid having to install valgrind if the unit tests themselves are failing - which is an arguable optimization in the first place :).

`test_c_mem` will run tests with valgrind so we'll get both: the info about leaks, and test results.

DDMakefile
```diff
@@ -33,7 +33,7 @@ test_c_mem: $(SO_FILE)
 	$(MAKE) -C $(BUILD_DIR) test TESTS="-q --show-all -m"
 
 test_integration: install_ini
-	composer test
+	composer test -- $(PHPUNIT)
```
pawelchcki: 👍

scripts/install.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
pawelchcki: Something to think about later. 
Right now we have 3 places with build/run/test helpers `composer.json`(scripts), `DDMakefile`, and these scripts.

I think it would be great to limit it to at most two: I.e. any combination of `scripts + makefile`, or `scripts + composer.json`, or `composer.json + makefile`. Or maybe only `composer.json` if its powerful enough.

README.md
```diff
@@ -105,9 +105,38 @@ Before contributing to this open source project, read our [CONTRIBUTING.md](http
...
+
+
+Please note that the later is a wrapper around `phpunit`, so you can use all the common
+[options](https://phpunit.de/manual/5.7/en/textui.html#textui.clioptions) that you would with `phpunit`:
+
+    # Run a suite
+    $ ./scripts/test-tracer.sh --testsuite=unit
+
+    # Use a filter
+    $ ./scripts/test-tracer.sh --filter=Predis
```
pawelchcki: 👍

.circleci/config.yml
labbati: Removing it!

scripts/install.sh
labbati: Good idea! Let me try how it plays out putting everything in the composer!

pawelchcki:
Thanks, great stuff!

PR #87, opened 2018-10-25
Remove dd extension from required extensions in composer.

pawelchcki:
Makes using code instrumented with DD library a bit nicer to work on workstations where this extension is not installed.


composer.json
```diff
@@ -22,7 +22,6 @@
     "require": {
         "php": "^5.6 || ^7.0",
         "ext-curl": "*",
```
labbati: Do you think we should also remove the dependency on `ext-curl`?

PR #89, opened 2018-10-26
Add Laravel 5.7 and 4.2 sample apps

labbati:
This PR adds sample Laravel 4.2 and 5.7 apps to the repo to be used during tests. It has been created through `composer create-project laravel/laravel your-project-name x.x.*` command and a few folders have been removed.
In the future can be optimized, but for now we just push the entire default apps.

This PR just push the sample apps and not tests or shared code, so it can be easily merged. Work on tests will be done on a different branch.


pawelchcki:
Looks good! But...
it would be great to document the way those sample apps were generated, so that its easy to repeat the process if need be.
"Some" builds are failing :)

labbati:
Assigning back to @pawelchcki for final round of CR.

tests/Integration/Frameworks/README.md
```diff
@@ -0,0 +1,21 @@
+# Sample apps
```
pawelchcki: Great stuff!

PR #95, opened 2018-11-01
Symfony add service type and resource name to root span

pawelchcki:
Small tweaks to Symfony integration.

### DIFF
--- a/src/DDTrace/Integrations/SymfonyBundle.php
+++ b/src/DDTrace/Integrations/SymfonyBundle.php
@@ -40,7 +40,6 @@ class SymfonyBundle extends Bundle
             return;
         }
 
-
         // Creates a tracer with default transport and default propagators
         $tracer = new Tracer(new Http(new Json()));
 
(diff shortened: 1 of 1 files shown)


src/DDTrace/Integrations/SymfonyBundle.php
```diff
@@ -4,12 +4,14 @@
...
 use DDTrace\Encoders\Json;
 use DDTrace\Tags;
+use DDTrace\Types;
 use DDTrace\Tracer;
 use DDTrace\Transport\Http;
 use OpenTracing\GlobalTracer;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 use Symfony\Component\HttpKernel\HttpKernel;
 use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Bridge\Monolog\Processor\DebugProcessor;
```
chuck: Looks like this isn't used?

pawelchcki: Oh right, this is a leftover from testing I did to figure out how to access route name.

PR #100, opened 2018-11-07
Rename withTracer test method to isolateTracer

labbati:
**Blocked by #102**
This PR just rename the integration tests base method `withTracer` to `isolateTracer` to make naming more expressive. At the same time, after rebasing on master, we also adapted a few lines of code to be compatible with php 5.4.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -258,14 +258,17 @@ jobs:
           name: Test installing packages on target systems
           command: make -f dockerfiles/verify_packages/Makefile
 
+
 workflows:
   version: 2
   build_packages:
     jobs:
-      - "hold":
-          type: approval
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -258,14 +258,17 @@ jobs:
...
     jobs:
-      - "hold":
-          type: approval
       - "7.0 20151012": &BUILD_PACKAGE_FORKFLOW
-          requires: [ "hold" ]
+          filters:
+            tags:
+              only: /(^build$)|(^[v]?[0-9]+(\.[0-9]+)*$)/
+            branches:
+              ignore: /.*/
```
pawelchcki: Could we build the packages on the master branch as well?

labbati:
Assigning back to @pawelchcki as the requested fix has been implemented in the original branch that intended to remove the holds: #102

PR #103, opened 2018-11-07
Update composer.json

inverse:
Required in both JSON encoder and Stream transport.


labbati:
Hey @inverse, thanks again for this contribution! Merging it right now!

PR #112, opened 2018-11-12
Laravel: Prevent traces not sent when app name is empty

stayallive:
After a lot of head scratching and not reading the logs... (I know)

I noticed that when you leave the `config('app.name')` value empty in your app no traces are being sent since it needs to have a non-empty value. So I made sure the app name does return a non-empty value so traces won't be stopped by that :)

### DIFF
--- a/src/DDTrace/Integrations/LaravelProvider.php
+++ b/src/DDTrace/Integrations/LaravelProvider.php
@@ -135,12 +135,14 @@ class LaravelProvider extends ServiceProvider
 
     private function getAppName()
     {
+        $name = null;
+
         if (isset($_ENV['ddtrace_app_name'])) {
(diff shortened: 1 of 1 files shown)


stayallive:
Whoops, sorry about that :) 

Github web is not so nice as phpstorm is with cleaning that up after itself.

Should be fixed now 👍 

Great to hear about the improvements coming, will hold of on taking the integration apart for now until that's in place.

labbati:
Hey @stayallive 

> Github web is not so nice as phpstorm is with cleaning that up after itself.

eheh it happens to me all the time when not using an IDE

> Should the 5.4 compat also be fixed?

Not for now, we are going to fix all the 5.4 things at once in one single PR.

Merging it. Thanks!

PR #116, opened 2018-11-13
Fix laravel middleware class name bug

### LINKED ISSUE #115: Laravel authorize middleware not working
Hey,

I use the authorization via middleware of laravel (https://laravel.com/docs/5.7/authorization#authorizing-actions-using-policies).

In the Laravel provider ([line 61](https://github.com/DataDog/dd-trace-php/blob/4f9c95cef0afe38b112f67493ae51ef642f6e72a/src/DDTrace/Integrations/LaravelProvider.php#L61)) instead of a class I get `'Illuminate\\Auth\\Middleware\\Authorize:users'` which leads to this error: `unexpected parameter combination, expected (class, function, closure) or (function, closure)`.

Sh4d1:
Fixes #115 

Signed-off-by: Patrik Cyvoct <pcyvoct@assessfirst.com>

### DIFF
--- a/src/DDTrace/Integrations/LaravelProvider.php
+++ b/src/DDTrace/Integrations/LaravelProvider.php
@@ -58,7 +58,7 @@ class LaravelProvider extends ServiceProvider
             foreach ($pipes as $pipe) {
                 $pipeClass = explode(':', $pipe)[0];
 
-                dd_trace($pipe, 'handle', function (...$args) {
+                dd_trace($pipeClass, 'handle', function (...$args) {
                     $scope = GlobalTracer::get()->startActiveSpan('laravel.middleware');
                     $span = $scope->getSpan();
                     $span->setResource(get_class($this));


src/DDTrace/Integrations/LaravelProvider.php
```diff
@@ -58,7 +58,7 @@ public function register()
             foreach ($pipes as $pipe) {
                 $pipeClass = explode(':', $pipe)[0];
 
-                dd_trace($pipe, 'handle', function (...$args) {
+                dd_trace($pipeClass, 'handle', function (...$args) {
```
labbati: I may be missing something, but something is not clear to me! Based on #115 this callback receives `....\\Authorize:users`, so a callback. Here looks like that if a pipeline is a callback it is executed (https://github.com/laravel/framework/blob/621d91d802016ab4a64acc5c65f81cb9f5e5f779/src/Illuminate/Pipeline/Pipeline.php#L129).

So I believe that with this proposed approach we don't get the error anymore but we are not tracing the real pipe stop point.

Anyway, very good catch, if you are ok with that, we have already some work ready on a different branch and I will move that forward, does it make sense to you?

Sh4d1: Hmm didn't look at Pipeline.php. So the problem may be from the `dd_trace` function? 

Sounds good to me 😄

labbati: It may be! Did you have a chance to investigate at which point the callback `....\\Authorize:users` is sent by Laravel through the pipeline? Just to know how it is invoked, did not have a chance to look at it so far, still managing the various use cases based on the code I see in `Pipeline.php`

Sh4d1: I know where it comes from but not exactly how it ends up here. It's a middleware so I guess it's logic but don't actually know how. It's used like this:
```
Route::put('/blabla', function (Post $post) {
})->middleware('can:users');
```

labbati: Ok thanks, looking at your code and properly reading through the Laravel docs now I see the usage you were referring to (I had never seen the middleware with parameters notation before). So I am going to merge your change with another couple of changes that were necessary for similar issues and create a branch. Will you able to build and test that branch?

Sh4d1: Sounds good! (Though we are not tracing the real call yet right?) 

Yes will be able to build and test 😄

labbati:
Created PR #117 . @Sh4d1 if you have a chance to test it in you specific environment and provide us with some feedback it would be great! Thanks for you help!

PR #121, opened 2018-11-14
Move integrations to their own packages

labbati:
In order to better scale integrations development, we decided to move each integration to its own sub package.


labbati:
Hey @pawelchcki 

> One small optional question. Would it make sense to add information when its going to be deprecated? i.e. what version - 1.0? 0.5?

Honestly I can think of an exact version at the moment! What about adding it once we get a better grip?

labbati:
@pawelchcki just pushed a merge conflict resolution commit! Need a final CR

PR #124, opened 2018-11-16
Change register to boot for laravel integration

Sh4d1:
I'm having some issue when using register instead of boot ([something like this](https://stackoverflow.com/questions/33851821/laravel-5-1-class-cache-does-not-exist)). I've tested it, seems to run as before.

Signed-off-by: Patrik Cyvoct <pcyvoct@assessfirst.com>


Sh4d1:
@labbati seems good to me 👍

labbati:
Closing it as it will be fixed by #127

labbati:
@Sh4d1 just to let you know we just release 0.3.1 which brings the solution that you needed https://github.com/DataDog/dd-trace-php/releases/tag/0.3.1

PR #129, opened 2018-11-16
Symfony exception handling and meta tags for request

jkrnak:
Hook symfony's HttpKernel::handleException into the main span, in case of an exception now it will be added as an error to the main span.

Add request meta tags to the main request span:
- client ip
- request method
- content-type
- referrer
- uri (host, port, uri, query params)


src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
```diff
@@ -48,13 +52,18 @@ public function boot()
...
-        $symfony_request_span->setTag(Tags\SPAN_TYPE, Types\WEB_SERVLET);
+        $symfonyRequestSpan = $scope->getSpan();
+        $symfonyRequestSpan->setTag(Tags\SERVICE_NAME, $this->getAppName());
+        $symfonyRequestSpan->setTag(Tags\SPAN_TYPE, Types\WEB_SERVLET);
 
         // public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
-        dd_trace(HttpKernel::class, 'handle', function ($request, ...$args) use ($symfony_request_span) {
+        dd_trace(HttpKernel::class, 'handle', function ($request, ...$args) use ($symfonyRequestSpan) {
             $scope = GlobalTracer::get()->startActiveSpan('symfony.kernel.handle');
+            $symfonyRequestSpan->setTag('request.method', $request->getMethod());
```
jkrnak: I'm not sure if you guys have plans regarding meta tag names request attributes. Maybe you already have on other projects. If there are any guidelines on this, please let me know.

jkrnak:
About the failures, the variadic function args and the `::class` meta was not present in version earlier than 5.6. 
What's the minimum version of PHP you would like to support?

labbati:
We are going to enforce PHP 5.4+!

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
labbati: Hey! Thanks for the contribution, really appreciated!

> I'm not sure if you guys have plans regarding meta tag names request attributes

Do you mean how to to include request's query string parameters?

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
```diff
@@ -48,13 +52,18 @@ public function boot()
...
+            $symfonyRequestSpan->setTag('request.uri', $request->getUri());
+            $symfonyRequestSpan->setTag('request.content-type', $request->getContentType());
+            $symfonyRequestSpan->setTag('request.referrer', $request->headers->get('referer'));
+            $symfonyRequestSpan->setTag('request.client-ip', $request->getClientIp());
```
labbati: We would prefer to not store the client IP for now, mainly for privacy reasons. We can consider adding hooks, soon, to let users modify the request spans as they like, but for now please remove this line.

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
```diff
@@ -48,13 +52,18 @@ public function boot()
...
+        $symfonyRequestSpan = $scope->getSpan();
+        $symfonyRequestSpan->setTag(Tags\SERVICE_NAME, $this->getAppName());
+        $symfonyRequestSpan->setTag(Tags\SPAN_TYPE, Types\WEB_SERVLET);
 
         // public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
-        dd_trace(HttpKernel::class, 'handle', function ($request, ...$args) use ($symfony_request_span) {
+        dd_trace(HttpKernel::class, 'handle', function ($request, ...$args) use ($symfonyRequestSpan) {
             $scope = GlobalTracer::get()->startActiveSpan('symfony.kernel.handle');
+            $symfonyRequestSpan->setTag('request.method', $request->getMethod());
+            $symfonyRequestSpan->setTag('request.uri', $request->getUri());
```
labbati: Please see `Tags\HTTP_METHOD` and `Tags\HTTP_URL`

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
jkrnak: Query string, request method, uri, content-type etc... 
I guess not all will make sense to everyone, some people might don't want to see content-type but are more interested in some other custom headers to be added as meta to the trace.

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
jkrnak: Fair enough, I'll remove it

jkrnak:
Regarding php 5.4, do you want me to change the `::class` class resolution into strings for the fully qualified classname?

SammyK:
@jkrnak I'm guessing since this is a Symfony v4 integration which [requires a minimum of PHP 7.1](https://github.com/symfony/symfony/blob/4.0/composer.json#L19), we should be safe to use 7.1+ syntax, right @labbati? :)

labbati:
> @labbati Regarding php 5.4, do you want me to change the `::class` class resolution into strings for the fully qualified classname?

If you can do it on the symfony integration that would be great.

labbati:
> @jkrnak I'm guessing since this is a Symfony v4 integration which [requires a minimum of PHP 7.1](https://github.com/symfony/symfony/blob/4.0/composer.json#L19), we should be safe to use 7.1+ syntax, right @labbati? :)

Theoretically yes, but would it be the 5.4 lint check much more complex to configure? With exclude etc...Probably, in my opinion, we should just say "all the code should be 5.4 compliant", but I am open on it @SammyK.

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
labbati: > request method and uri

You were right to add.

> Query string

This could be a little bit more tricky, as it may leak some data. Hopefully nobody add "auth token" tot he query string, but I have seen them. So please hold on on this for now, we may want to provide some sort of configuration/hook to do that.

> content-type etc

For headers in general we have an RFC which is in final review stage to provide a consistent way across all the languages and integrations that we support to report http headers (and thus content-type among them). So for now I would prefer not to add the content type. While I cannot commit to a date, is will be something that we implement not far from now.

jkrnak: I removed the query string and content type.
HTTP meta now looks like this:
```
http { 
  method: GET
  url: http://localhost.lh:9091/ping
} 
```

jkrnak:
@labbati @SammyK I added the class names as string so we don't use the `::class` resolution.
Next problem:
> "finally" keyword (in exception handling) is not present in PHP version 5.4 or earlier

not sure how to get around this without ignoring this folder from PHP 5.4 linting :)

labbati:
> @labbati @SammyK I added the class names as string so we don't use the `::class` resolution.
> Next problem:
> 
> > "finally" keyword (in exception handling) is not present in PHP version 5.4 or earlier
> 
> not sure how to get around this without ignoring this folder from PHP 5.4 linting :)

Don't worry for now, we will fix all together. We will merge even if we have 5.4 lint failing. In the very next few days we will fix it! Working on the final CR today!

PR #132, opened 2018-11-19
Add key obfuscation to Memcached integration

SammyK:
Since Memcached keys can contain secrets and the keys aren't obfuscated by the dd-agent, this will obfuscate the keys before hitting the dd-agent. :)

### DIFF
--- a/src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
+++ b/src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
@@ -4,6 +4,7 @@ namespace DDTrace\Integrations\Memcached;
 
 use DDTrace\Tags;
 use DDTrace\Types;
+use DDTrace\Obfuscation;
 use OpenTracing\GlobalTracer;
 
 /**
@@ -207,7 +208,7 @@ class MemcachedIntegration
         if (!is_array($args[0])) {
             self::setServerTagsByKey($span, $memcached, $args[0]);
         }
-        $span->setTag('memcached.query', "$command " . self::obfuscateKeys($args[0]));
+        $span->setTag('memcached.query', "$command " . Obfuscation::toObfuscatedString($args[0]));
         $span->setResource($command);
 
         try {
@@ -230,7 +231,7 @@ class MemcachedIntegration
         $span->setTag('memcached.server_key', $args[0]);
         self::setServerTagsByKey($span, $memcached, $args[0]);
 
-        $span->setTag('memcached.query', "$command " . self::obfuscateKeys($args[1]));
+        $span->setTag('memcached.query', "$command " . Obfuscation::toObfuscatedString($args[1]));
         $span->setResource($command);
 
         try {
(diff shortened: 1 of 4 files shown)


src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
```diff
@@ -362,17 +356,16 @@ private static function setServerTagsByKey($span, $memcached, $key)
...
      * Given a callback arg, it extract the keys depending on scalar vs associative array.
      *
-     * @param $array
-     * @return array
+     * @param mixed $keys
+     * @param string $glue
+     * @return string
      */
-    private static function extractKeys($array)
+    private static function obfuscateKeys($keys, $glue = ' ')
```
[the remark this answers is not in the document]
SammyK: I like that idea! Do you mind if we remove the `Util` part of the namespace and just put it under `DDTrace\Obfuscation`? I'm also wandering if we just want to namespace a function instead of a static method like, `DDTrace\toObfuscatedString()`. Thoughts? :)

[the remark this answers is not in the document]
SammyK: Now that I think about it, maybe a static method on a class is better so we can add a `Obfuscation::REPLACEMENT` constant and possible other obfuscation-specific things. :)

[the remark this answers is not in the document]
labbati: I like to remove the `Util` part and I like to have a class with obfuscation related things 😄

SammyK:
Thanks @labbati! I've updated the PR based on your feedback. Let me know if you have other tweaks for me. :)

tests/Unit/ObfuscationTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
labbati: 🔝

tests/Integration/Integrations/Memcached/MemcachedTest.php
```diff
@@ -42,7 +43,7 @@ public function testAdd()
         $this->assertSpans($traces, [
             SpanAssertion::build('Memcached.add', 'memcached', 'memcached', 'add')
                 ->withExactTags(array_merge(self::baseTags(), [
-                    'memcached.query' => 'add key',
+                    'memcached.query' => 'add ' . Obfuscation::toObfuscatedString('key'),
```
labbati: Nit-pick 😄.
This is not a blocker for me and will approve the PR, but wouldn't you agree that we should write `'?'` here in place of `Obfuscation::toObfuscatedString('key')`? What if we change the glue at a certain point in the utility class? We would like to know that the user expectation is not matched anymore. 
I am always in favor of setting the "purest" expectation as possible in tests. But again, not a blocker to me!

SammyK: I agree! Although it technically makes the tests more brittle, I think that's what we want in this case. Also, the tests would be way easier to read. I'll revert this. :)

tests/Unit/ObfuscationTest.php
SammyK: Not sure I understand this comment. Hehe.

labbati: eheheh nothing specific, I love when we have specific classes/modules that can be independently unit tested 😄 so it was just excitement 😄

PR #133, opened 2018-11-19
[Integration] Add basic Guzzle v6 integration

SammyK:
This is still a work in progress, but I wanted to post the progress so far to make sure we're all on the same page.

There are a few specific things I'd like to point out:

1. When we're testing 3rd-party packages like this, do we want to add those to the main `composer.json` as a dev dependency? I'm thinking this will be an issue when we want to add a Guzzle v5 integration for example. So not really sure what we want to do there.
2. What are your thoughts on extracting the main tracing boilerplate into `GuzzleTracer.php`? I figured it could help reduce duplication and boilerplate.
3. Do we want to trace the other methods within the package or just the ones making HTTP calls? And if we're tracing other methods, which ones should we target?

Currently this only supports sync methods. So next up is to tackle the async methods. :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "opentracing/opentracing": "1.0.0-beta5"
     },
     "require-dev": {
+        "guzzlehttp/guzzle": "^6.3",
         "phpcompatibility/php-compatibility": "^9.0",
         "phpcompatibility/phpcompatibility-passwordcompat": "^1.0",
         "phpcompatibility/phpcompatibility-symfony": "*",
--- a/src/DDTrace/Integrations/Guzzle/v6/GuzzleTracer.php
+++ b/src/DDTrace/Integrations/Guzzle/v6/GuzzleTracer.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace DDTrace\Integrations\Guzzle\v6;
+
+use DDTrace\Tags;
+use DDTrace\Types;
+use GuzzleHttp\Client;
+use OpenTracing\GlobalTracer;
+
+class GuzzleTracer
+{
+    private $client;
+    private $command;
+    private $args;
+
+    private $scope;
+    private $span;
+
+    public function __construct(Client $client, $command, array $args)
+    {
+        $this->client = $client;
+        $this->command = $command;
+        $this->args = $args;
+
+        $this->scope = GlobalTracer::get()->startActiveSpan("GuzzleHttp\Client.$command");
+        $this->span = $this->scope->getSpan();
+        $this->span->setTag(Tags\SPAN_TYPE, Types\GUZZLE);
+        $this->span->setTag(Tags\SERVICE_NAME, 'guzzle');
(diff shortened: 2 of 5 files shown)


src/DDTrace/Integrations/Guzzle/v6/GuzzleTracer.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->client = $client;
+        $this->command = $command;
+        $this->args = $args;
+
+        $this->scope = GlobalTracer::get()->startActiveSpan("GuzzleHttp\Client.$command");
+        $this->span = $this->scope->getSpan();
+        $this->span->setTag(Tags\SPAN_TYPE, Types\GUZZLE);
+        $this->span->setTag(Tags\SERVICE_NAME, 'guzzle');
+        $this->span->setTag('guzzle.command', $command);
+        $this->span->setResource($command);
```
labbati: We are deprecating `$span->setResource(<value>);` in favor of `$span->setTag(Tags\RESOURCE_NAME, <value>);`.

src/DDTrace/Integrations/Guzzle/v6/GuzzleTracer.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function trace()
+    {
+        try {
+            return $this->client->{$this->command}(...$this->args);
+        } catch (\Exception $e) {
+            $this->span->setError($e);
+            throw $e;
+        } finally {
```
labbati: At this point we should go for php 5.4 compatibility (not using `finally`).

tests/Integration/Integrations/Guzzle/v6/GuzzleTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+    public static function setUpBeforeClass()
+    {
+        Integrations\Guzzle\v6\GuzzleIntegration::load();
+    }
+
+    protected function setUp()
+    {
+        parent::setUp();
+        $mock = new MockHandler([
```
labbati: ❤️

composer.json
```diff
@@ -34,6 +34,7 @@
         "opentracing/opentracing": "1.0.0-beta5"
     },
     "require-dev": {
+        "guzzlehttp/guzzle": "^6.3",
```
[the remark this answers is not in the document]
labbati: @SammyK now that I look at it, we should test 5.* version also. For now (not having yet a multi-version setup, I would only focus on 5.*)

labbati:
@SammyK is this still necessary or was superseded by the Guzzle 5 integration PR? :)

SammyK:
@labbati We'll still need to support it eventually since Guzzle v6 is the latest version (and we're probably going to see a Guzzle v7 soon now that [PSR-18](https://www.php-fig.org/psr/psr-18/) has passed.) But we can close this for now until we figure out how we want to support multiple versions of the same project. :)

Issue #119, opened 2018-11-14
High CPU usage


joeldodge79:
We're using Magento 1 and I decided to just throw this code into our index.php file as a first stab at seeing some data:

```
require_once('lib/vendor/autoload.php');
$tracer = new \DDTrace\Tracer(new \DDTrace\Transport\Http(new \DDTrace\Encoders\Json()));
\OpenTracing\GlobalTracer::set($tracer);
$scope = $tracer->startActiveSpan('ourorg.www.request');
$span = $scope->getSpan();
$span->setTag(\DDTrace\Tags\SERVICE_NAME, 'www');
$span->setTag(\DDTrace\Tags\SPAN_TYPE, \DDTrace\Types\WEB_SERVLET);
\DDTrace\Integrations\PDO::load();
register_shutdown_function(function () use ($scope) {
    $scope->close();
    \OpenTracing\GlobalTracer::get()->flush();
});
```

we got data in the datadog APM UI ... but the webserver slammed to 100% cpu and stopped responding. any suggestions?

joeldodge79:
using this extension: https://github.com/DataDog/dd-trace-php/releases/download/0.2.7/datadog-php-tracer-0.2.7_beta-1.x86_64.rpm

labbati:
Hi! @joeldodge79, sorry to hear that! Can you provide a little bit more detail about your environment, e.g. OS, php version, bits (I assume 64)? I will try to reproduce the issue in a docker + centos (I guessed from .rpm) environment with your code and will post the outcome here. Also bringing @pawelchcki in the loop (our C-extension master)

labbati:
Ok I did a very quick test and things looks good as I do not see any CPU anomaly, so must be something else.
I would ask you to share more info about your env so we can see if we can replicate. 
It may also be an issue specifically related to magento, and in that case we may need to postpone the investigation a little bit.

==============================================================================
Dumping here a few resources to speed up debug process:

I used the following `Dockerfile`

```
FROM naqoda/centos-apache-php:php71
RUN rpm -i https://github.com/DataDog/dd-trace-php/releases/download/0.2.7/datadog-php-tracer-0.2.7_beta-1.x86_64.rpm
```

Build:

```
docker build -t debug_centos .
```

Run through `docker-compose`:

```
  sample:
    image: "debug_centos"
    volumes:
      - .:/var/www/app/public_html
    environment:
      DD_AGENT_HOST: <YOUR_HOST_HERE>
    ports:
[...]

PR #135, opened 2018-11-20
[Fix] Memory leak with closures in PHP 7

SammyK:
The closures in PHP 7 aren't being freed. When running the following snippet of PHP:

```php
# foo.php
dd_trace('strlen', function (...$args) {
    // Do stuff
});
```

In debug mode, PHP generates the following warning:

```
[Tue Nov 20 08:34:45 2018]  Script:  '%s/foo.php'
%s/Zend/zend_closures.c(451) :  Freeing 0x%d (%d bytes), script=%s/foo.php
=== Total 1 memory leaks detected ===
```

This PR fixes this particular memory leak. This also might be tangentially related to issue #119, but that's just a guess. :)

### DIFF
--- a/src/ext/dispatch_compat_php7.c
+++ b/src/ext/dispatch_compat_php7.c
@@ -21,7 +21,10 @@ zend_function *ddtrace_function_get(const HashTable *table, zend_string *name) {
     return ptr;
 }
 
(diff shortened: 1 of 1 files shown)


clutchski:
worth an autmoated test if possible

SammyK:
Thanks @pawelchcki! :)

PR #136, opened 2018-11-20
Support for Eloquent 4.2

labbati:


labbati:
Pushed fixes for @SammyK review

PR #137, opened 2018-11-21
Make PHPStan available to contributors

SammyK:
Part of the build checks include static analysis but there's no way to run the analyzer locally before pushing up a PR without manually installing PHPStan globally and running it with the settings from CircleCI.

This PR includes PHPStan in the dev dependencies and adds a custom Composer command to run it with the same settings as CircleCI:

    $ composer static-analyze

As a side note - it might be worth revamping the CONTRIBUTING.md docs as bit. We could move all the testing/linting/container setup docs from the main README so that the README is less cluttered and easier for non-contributors to get dd-trace-php installed in their project. Does that sound reasonable? :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ aliases:
   - &STEP_COMPOSER_INSTALL
     run:
       name: Installing dependencies with composer
-      command: composer install --no-interaction --ignore-platform-reqs
+      command: composer install --no-interaction --ignore-platform-reqs --no-plugins
 
 
   - &STEP_PREPARE_TEST_RESULTS_DIR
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -55,7 +55,7 @@ aliases:
   - &STEP_COMPOSER_INSTALL
     run:
       name: Installing dependencies with composer
-      command: composer install -n
+      command: composer install --no-interaction --ignore-platform-reqs --no-plugins
```
pawelchcki: ❤️

pawelchcki: ❤️

SammyK:
I finally got the PHP < 7.1 builds to still pass even with the PHPStan requirement. :D 

All conflicts should be resolved and this should be ready for one last CR. :)

PR #139, opened 2018-11-22
Laravel 4.2 and 5.7 basic Integration Tests with multi-framework testing infrastructure

labbati:


tests/Integration/Frameworks/Laravel/Version_4_2/app/tests/Integration/CommonSpecsTest.php
```diff
@@ -0,0 +1,43 @@
...
+            }
+        });
+
+        $this->assertExpectedSpans($this, $traces, $spanExpectations);
+    }
+
+    public function provideSpecs()
+    {
+        $specsProvider = new CommonSpecsProvider();
+        return $specsProvider->provide(TestSpecs::all(), new Laravel4ExpectationsProvider());
```
pawelchcki: I feel that at this point it stops being easy to follow what is going on. 

I love how the Span "scenarios" are built, and group together the same behavior.

I just wonder if we could be a little bit more explicit So it would not require looking up 3 different files to understand behavior of single test.

To make this comment a bit more constructive:
- Laravel4ExpectationsProvider is great!
- GetSpecs is great!
- TestSpecs `priovide()` contents could probably be moved to explicit test cases:
-  the code inside ` if ($spec instanceof GetSpec)`  could be me moved to GetSpec class (?)

I.e. using custom assertion like:
`$this->assertRequestCreatesSpans(Specs::get("/error", 500), Specs::spans('A GET request with an exception'));`

then impl of `assertRequestCreatesSpans` could contain all the magic from provideSpecs() ?

My point is also to stay a bit closer to PHPUnit basic way of doing things - if possible.

Maybe I'm trying to simplify something thats just complex in nature. And maybe my counter proposal is more complicated.. Would like to know your thoughts :)

composer.json
```diff
@@ -78,6 +78,11 @@
             "echo \"Integration tests require the agent up and running. Use composer run-agent.\"",
             "phpunit --colors=always --configuration=phpunit.xml --testsuite=integration"
         ],
-        "test-unit": "phpunit --colors=always --configuration=phpunit.xml --testsuite=unit"
+        "test-unit": "phpunit --colors=always --configuration=phpunit.xml --testsuite=unit",
+
+        "laravel-42-update": "composer --working-dir=tests/Integration/Frameworks/Laravel/Version_4_2 update",
```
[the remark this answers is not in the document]
labbati: Eheheh I did not get the comment 😞 it it ok? 😄

tests/Integration/Frameworks/Laravel/Version_4_2/composer.json
```diff
@@ -15,8 +19,24 @@
 			"app/database/migrations",
 			"app/database/seeds",
 			"app/tests/TestCase.php"
+		],
+		"psr-4": {
+			"App\\": "app/",
+			"DDTrace\\": "../../../../../src/DDTrace/"
+		},
+		"files": [
+			"../../../../../src/DDTrace/Tags.php",
```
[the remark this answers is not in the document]
labbati: I could think of a few ways:
 - add to the composer.json a [repository path](https://getcomposer.org/doc/05-repositories.md#path) that points to the dd-trace-php root: but composer prevents me from doing it as the current folder is contained in the path I want to register. It seems just to be not possible.
 - this method (which I agree with you that is horrible)
 - more tricky (and brittle) approaches like copying relevant content to a 3rd location and register that as a repository path.

Amongs the evils, I found this a little bit better, but totally opened to change it. Any 4th idea? 😄

tests/Integration/Frameworks/Laravel/Version_4_2/app/tests/Integration/CommonSpecsTest.php
labbati: I appreciate the fact that you commented on this part because I believe it can be made simpler too.

> TestSpecs priovide() contents could probably be moved to explicit test cases

Could you please elaborate on this?

> the code inside if ($spec instanceof GetSpec) could be me moved to GetSpec class (?)

This could be a little bit tricky because different frameworks (and versions of the same framework) have often different ways to issue a `get/post/...` request.

> `$this->assertRequestCreatesSpans(Specs::get("/error", 500), Specs::spans('A GET request with an exception'));`
> then impl of assertRequestCreatesSpans could contain all the magic from provideSpecs() ?

In general I like the idea and I would like to point out a few things as I think about how to implement it:
 - Is my understanding that `$this->assertRequestCreatesSpans(...)` would be called in the specific test cases? So for example in `...\Laravel\Version_4_"\...\CommonTests`?
 - The idea was to define the scenarios in a common place, so we know that ALL the frameworks and ALL the versions of every framework have a test for that scenario. So if you like this approach, in the method above we would replace `Specs::get("/error", 500)` as the first argument with something like `CommonSpecs::get('A GET request with an exception')`. 
 - `Specs::spans('A GET request with an exception')` where would it leave?

I think that this suggestion BTW makes a lot of sense and we should talk through it more in depth.

composer.json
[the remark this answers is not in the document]
pawelchcki: It is OK

tests/Integration/Frameworks/Laravel/Version_4_2/app/tests/Integration/CommonSpecsTest.php
labbati: Rename to `TestScenario`

tests/Integration/Frameworks/Laravel/Version_4_2/app/tests/Integration/CommonSpecsTest.php
```diff
@@ -0,0 +1,43 @@
...
+use DDTrace\Tests\Integration\Common\TracerTestTrait;
+use DDTrace\Tests\Integration\Frameworks\Laravel\Laravel4ExpectationsProvider;
+use DDTrace\Tests\Integration\Frameworks\TestSpecs;
+use DDTrace\Tests\Integration\Frameworks\Util\CommonSpecsProvider;
+use DDTrace\Tests\Integration\Frameworks\Util\Request\GetSpec;
+use DDTrace\Tests\Integration\Frameworks\Util\Request\RequestSpec;
+use Tests\TestCase;
+
+
+class CommonSpecsTest extends TestCase
```
labbati: `CommonScenariosTest`

tests/Integration/Frameworks/Laravel/Version_4_2/app/tests/Integration/CommonSpecsTest.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    use TracerTestTrait, SpanAssertionTrait;
+
+    /**
+     * @dataProvider provideSpecs
+     * @param RequestSpec $spec
+     * @param array $spanExpectations
+     */
+    public function testSpecs(RequestSpec $spec, array $spanExpectations)
```
labbati: `testScenario`

tests/Integration/Frameworks/Laravel/Version_4_2/app/tests/Integration/CommonSpecsTest.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        $traces = $this->simulateWebRequestTracer(function() use ($spec) {
```
labbati: Add a few comments

labbati:
@pawelchcki Ready for next round of CR

composer.json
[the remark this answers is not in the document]
pawelchcki: :D

labbati:
🙇

Issue #140, opened 2018-11-23
NoopSpan doesn't have additional methods


funkjedi:
Integration works fine for web requests but my Laravel workers fail when running `artisan queue:work`

```
Symfony\Component\Debug\Exception\FatalThrowableError Call to undefined method OpenTracing\NoopSpan::setResource() 
    vendor/datadog/dd-trace/src/DDTrace/Integrations/PDO/PDOIntegration.php:42 PDO::DDTrace\Integrations\PDO\{closure}
    vendor/laravel/framework/src/Illuminate/Database/Connectors/Connector.php:68 Illuminate\Database\Connectors\Connector::createPdoConnection
    vendor/laravel/framework/src/Illuminate/Database/Connectors/Connector.php:44 Illuminate\Database\Connectors\Connector::createConnection
    vendor/laravel/framework/src/Illuminate/Database/Connectors/MySqlConnector.php:24 Illuminate\Database\Connectors\MySqlConnector::connect
```
The problem appears to be the tracer is trying to use [NoopSpan](https://github.com/opentracing/opentracing-php/blob/1.0.0-beta5/src/OpenTracing/NoopSpan.php) which doesn't have the extra methods implemented in `DDTrace\Span`.
https://github.com/DataDog/dd-trace-php/blob/690bd17583dcd14ca776fbc8f3c768f3c3704d51/src/DDTrace/Tracer.php#L97-L99
There probably needs to be a `OpenTracing\NoopSpan` subclass in DDTrace that implements all the additional methods that have been added.

Temporarily worked around by registering the DDTrace service provider from my AppServiceProvider only when not on CLI.
```php
/**
 * Register any application services.
 *
 * @return void
 */
public function register()
{
    if (extension_loaded('ddtrace') && php_sapi_name() !== 'cli') {
        $this->app->register(\DDTrace\Integrations\LaravelProvider::class);
    }
```

labbati:
Hi @funkjedi, thanks for the issue. Working on it very soon!

labbati:
Fixed by #141 in CR.

labbati:
Hi @funkjedi, 0.4.2 has been released: https://github.com/DataDog/dd-trace-php/releases/tag/0.4.2
Can you please confirm that this release fixes your issue?
Thanks for the patience!

labbati:
Hi @alkin , we are aware of the issue and we will be providing a fix in two steps:
[...]

PR #141, opened 2018-11-23
[core] Remove Span::setResource() which is not part of OpenTracing\Span interface

labbati:
The problem is that in specific conditions we may not have a `DDTrace\Span` and in place of it we get an `OpentTracing\NoopSpan` which just implement the interface `OpenTracing\Span` => exception.

### DIFF
--- a/src/DDTrace/Span.php
+++ b/src/DDTrace/Span.php
@@ -229,7 +229,7 @@ final class Span implements OpenTracingSpan
 
     /**
      * @deprecated
-     * @param $resource
+     * @param string $resource
      */
     public function setResource($resource)
     {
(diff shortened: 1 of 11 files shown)


labbati:
This also fixes #140

src/DDTrace/Span.php
```diff
@@ -227,9 +227,15 @@ public function getAllTags()
...
     }
 
+    /**
+     * @deprecated
+     * @param string $resource
+     */
     public function setResource($resource)
     {
-        $this->resource = (string)$resource;
+        error_log('DEPRECATED: Method "DDTrace\Span\setResource" will be removed soon, '
```
SammyK: I know we're using `error_log()` for the deprecated integration methods as well, but I'm curious why were using that instead of raising an `E_USER_DEPRECATED` error with `trigger_error()`.

```php
trigger_error('Method "DDTrace\Span\setResource" will be removed soon...', E_USER_DEPRECATED);
```

Sorry if I missed that discussion! :)

labbati: This is a good point! We should refactor all the deprecation warnings like this!

PR #143, opened 2018-11-23
SymfonyBundle: Use getenv() to get env variable

hinrik:
It is common on servers to not populate $_ENV, so `getenv()` is a more
reliable way to check an env var.

See: https://stackoverflow.com/questions/3780866/why-is-my-env-empty/27077452#27077452

### DIFF
--- a/src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
+++ b/src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
@@ -136,7 +136,7 @@ class SymfonyBundle extends Bundle
 
     private function getAppName()
     {
-        if (!empty(getenv('ddtrace_app_name'))) {
+        if ($appName = getenv('ddtrace_app_name')) {
             return getenv('ddtrace_app_name');
         } else {
             return 'symfony';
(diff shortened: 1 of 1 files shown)


src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
```diff
@@ -136,8 +136,8 @@ function (...$args) {
 
     private function getAppName()
     {
-        if (isset($_ENV['ddtrace_app_name'])) {
-            return $_ENV['ddtrace_app_name'];
+        if (!empty(getenv('ddtrace_app_name'))) {
```
labbati:
```suggestion
        if ($appName = getenv('ddtrace_app_name')) {
```

src/DDTrace/Integrations/Symfony/V4/SymfonyBundle.php
```diff
@@ -136,8 +136,8 @@ function (...$args) {
...
+            return getenv('ddtrace_app_name');
```
labbati:
```suggestion
            return $appName;
```

mariuscoto:
@labbati can we merge now ? When will the next version of `dd-trace-php` be released ?

PR #144, opened 2018-11-23
Laravel 5 publishable config file

funkjedi:
## Laravel 5 updates
- added publishable config
- bound tracer class as singleton in service container to allow tracer to be customized by integration

## General updates
- added `setTags()` method to tracer

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -99,7 +99,7 @@ To enable Laravel integration we need to configure a new Provider in `config/app
 ```php
     'providers' => [
 # .....
-      'DDTrace\Integrations\Laravel\V5\LaravelProvider',
+      \DDTrace\Integrations\Laravel\V5\LaravelProvider::class,
 ```
 
 Copy the package config to your local config with the publish command:
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -92,14 +92,20 @@ composer update
 
 ### Enabling tracing
 
-#### Laravel integration
+#### Laravel 5 integration
 
 To enable Laravel integration we need to configure a new Provider in `config/app.php`
 
 ```php
     'providers' => [
 # .....
-      'DDTrace\Integrations\LaravelProvider',
+      'DDTrace\Integrations\Laravel\V5\LaravelProvider',
+```
+
+Copy the package config to your local config with the publish command:
+
+```
+php artisan vendor:publish --provider="DDTrace\Integrations\Laravel\V5\LaravelProvider"
 ```
 
 Now your Laravel application should start sending traces to the Datadog agent running on localhost (in default configuration). The Datadog agent must have APM enabled; see https://docs.datadoghq.com/tracing/setup/ for instructions on installing and configuring the agent.
--- a/src/DDTrace/Integrations/Laravel/V5/ConfigRegistry.php
+++ b/src/DDTrace/Integrations/Laravel/V5/ConfigRegistry.php
@@ -1,0 +1,85 @@
+<?php
+
+namespace DDTrace\Integrations\Laravel\V5;
+
+use DDTrace\Configuration;
+use Illuminate\Config\Repository;
+
+/**
+ * Registry that holds configuration properties and that is able to recover
+ * configuration values from environment variables.
+ */
+class ConfigRegistry extends Configuration
+{
+    /** @var \Illuminate\Config\Repository */
+    private $repository;
+
+    public function __construct(Repository $repository)
+    {
+        $this->repository = $repository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function boolValue($key, $default)
+    {
+        return (bool) $this->getValueFromRepository($key, $default);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function inArray($key, $name)
+    {
+        $value = $this->getValueFromRepository($key);
+
+        if (is_string($value)) {
+            $value = explode(',', $value);
+            $value = array_map(function ($entry) {
+                return trim(strtolower($entry));
+            }, $value);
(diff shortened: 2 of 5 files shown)


labbati:
Hi @funkjedi thanks for the PR. We were actually talking of the proper way to allow configuring our integrations and this seems to be a very good start. We may consider, soon, to create classes around configs that can be hydrated in ways dependent on framework specific approaches. In this case, for Laravel, your approach seems to be very good and can be used, once we implement the config classes, as the source of config values.

I would like to receive other thoughts over this. @pawelchcki? @SammyK?

Lavoaster:
@labbati One item in this PR for changing this code from

```php
        if (!extension_loaded('ddtrace')) {
            trigger_error('ddtrace extension required to load Laravel integration.', E_USER_WARNING);
            return;
        }
```

to

```php
        if (!extension_loaded('ddtrace')) {
            return;
        }
```

Would you be willing to accept that change in a separate PR? It prevents us from installing this service provider in our app due to it killing the app in environments without dd-trace installed.

labbati:
> @labbati One item in this PR for changing this code from
> 
> ```
>         if (!extension_loaded('ddtrace')) {
>             trigger_error('ddtrace extension required to load Laravel integration.', E_USER_WARNING);
>             return;
>         }
> ```
> 
> to
> 
> ```
>         if (!extension_loaded('ddtrace')) {
>             return;
>         }
> ```
> 
> Would you be willing to accept that change in a separate PR? It prevents us from installing this service provider in our app due to it killing the app in environments without dd-trace installed.

Hey @Lavoaster, thanks for your comment. We are currently working toward making our integrations more configurable, but at the same time we understand which trigger warning (which in some envs may be converted to errors) is not the best on our side. At the same time is tracing is enabled and the ext is not installed we would like to notify the customer somehow.

Would it be ok for you if we replace

```
trigger_error('ddtrace extension required to load Laravel integration.', E_USER_WARNING);
```
with 
```
error_log('Warning: ddtrace extension is required to load Laravel integration.');
```
?

You will a lot of error_log, but it should still be ok while evaluating. Keep in mind that we are working to improve configuration and we will provide a global switch to disable tracing using env variables. 

If you are ok with that, it would be great if you can do it fo Laravel V4/V5 and Synfony. Not required for all other libraries.

Lavoaster:
@labbati In my opinion, preferably not. It's part of the troubleshooting stage of "Why isn't my app reporting data?", which would go to the sysadmin part of our company. 

If the error_log was put in, I would't want it bloating the log file during development. I'd end up removing the service provider from the app.php file and registering it manually in `AppServiceProvider` as such:

```php
public function register() {
    ...
    if (extension_loaded('ddtrace')) {
        $this->app->register(\DDTrace\Integrations\Laravel\V5::class)
    }
    ...
}
```

funkjedi:
@labbati if it was only triggering the error when enabled then I would concede the warning is not only 100% warranted but appropriate.
```php
if ($this->app['config']->get('ddtrace.enabled') && !extension_loaded('ddtrace')) {
    trigger_error('ddtrace extension required to load Laravel integration.', E_USER_WARNING);
    return;
}
```
However without the configuration updates in place there's currently no way to disable the tracer.

labbati:
Hey @funkjedi and @Lavoaster your objections totally make sense. In order to fix it at the root of the problem we merged to master the first chunk of configuration refactoring, so now you can both disable tracing completely `DD_TRACE_ENABLED=false` (default: `true`) or disable specific integrations `DD_INTEGRATIONS_DISABLED=curl,mysqli,...` (default to all enabled) #168 

What happens now is that if have the global switch enable and we do not find the extension, we trigger a warning which hints at how to fix it: `'Missing ddtrace extension. To disable tracing set env variable DD_TRACE_ENABLED=false'`.

If this ok for you? I guess so. You can already build it for master or wait for 0.7.0 which is due today (or tomorrow worst case).

@funkjedi at this point it would be great to integrate this PR with the new configuration system. If/when you have some spare cycles we can discuss of it, if you are interested of course.

Lavoaster:
Ah excellent update, thanks @labbati.

funkjedi:
@labbati I've updated to use the new configuration system. 

One comment on the configuration system implementation. Should the replace method not accept the `DDTrace\Configuration\Registry` interface instead of `DDTrace\Configuration\AbstractConfiguration`? 
https://github.com/DataDog/dd-trace-php/blob/329e2903f8ba95182c7d2856343a65ef577b8192/src/DDTrace/Configuration/AbstractConfiguration.php#L64
I've worked around it for now by just subclassing `DDTrace\Configuration` instead of just implementing the interface.

docs/getting_started.md
```diff
@@ -92,14 +92,20 @@ composer update
...
-#### Laravel integration
+#### Laravel 5 integration
 
 To enable Laravel integration we need to configure a new Provider in `config/app.php`
 
 ```php
     'providers' => [
 # .....
-      'DDTrace\Integrations\LaravelProvider',
+      'DDTrace\Integrations\Laravel\V5\LaravelProvider',
```
Lavoaster: It's probably worth updating this to conform with a normal laravel's `app.php`. 

```suggestion
      \DDTrace\Integrations\Laravel\V5\LaravelProvider::class,
```

labbati:
Hey @funkjedi and @Lavoaster [0.7.0](https://github.com/DataDog/dd-trace-php/releases/tag/0.7.0) is out which brings integrations disabling.
@funkjedi I commit to look at the PR on monday as I'l be PTO tomorrow.
Please, in the meantime, let me know if this release helps with the issues mentioned above.

krakjoe:
I'm closing this PR as it would seem stale, and the feedback loop appeared broken.

If I'm wrong to do that, please open a new PR, and accept my apologies in advance :)

PR #145, opened 2018-11-26
Distributed tracing, no sampling priority

labbati:


src/DDTrace/StartSpanOptionsFactory.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Creates an instance of StartSpanOptions making sure that if DD specific distributed tracing headers exist,
+     * then the \OpenTracing\Span that is about to be started will get the proper reference to the remote Span.
+     *
+     * @param OTTracer $tracer
+     * @param array $options
+     * @param array $headers An associative array containing header names and values.
+     * @return StartSpanOptions
+     */
+    public static function createForWebRequest(OTTracer $tracer, $options = [], $headers = [])
```
SammyK:
```suggestion
    public static function createForWebRequest(OTTracer $tracer, array $options = [], array $headers = [])
```
:)

tests/Unit/Http/HeadersTest.php
```diff
@@ -0,0 +1,34 @@
...
+final class HeadersTest extends Framework\TestCase
+{
+    public function test_headersMapToColonSeparatedValues_emptyArray()
+    {
+        $this->assertEmpty(Headers::headersMapToColonSeparatedValues([]));
+    }
+
+    public function test_headersMapToColonSeparatedValues_null()
+    {
+        $this->assertEmpty(Headers::headersMapToColonSeparatedValues(null));
```
SammyK: Is there a reason we want to accept `null` here as opposed to `[]`? :)

tests/Unit/Http/HeadersTest.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace DDTrace\Tests\Unit\Http;
+
+use DDTrace\Http\Headers;
+use PHPUnit\Framework;
+
+
+final class HeadersTest extends Framework\TestCase
+{
+    public function test_headersMapToColonSeparatedValues_emptyArray()
```
SammyK: I'm OK with snake case for test names, but it seems a bit harder to read with both snake case and camel case combined. Could we change this to be all snake case or all camel case? :)

src/DDTrace/Http/Headers.php
```diff
@@ -0,0 +1,29 @@
...
+    /**
+     * Provided an associative array as input ['key' => 'value] , it returns an indexed array where each value is
+     * 'key: value'.
+     *
+     * @param string[] $headers
+     * @return string[]
+     */
+    public static function headersMapToColonSeparatedValues($headers)
+    {
+        $headers = $headers ?: [];
```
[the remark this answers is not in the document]
labbati: to handle the null case, but I will remove it as you (correctly) asked my to enforce `array` type on method signature

tests/Unit/Http/HeadersTest.php
labbati: You know? For consistency with all the other tests that we have, let's revert back to camel case!

tests/Unit/Http/HeadersTest.php
labbati: Removing it (as we changed the code in the original function) :)

src/DDTrace/Propagators/TextMap.php
```diff
@@ -49,4 +49,24 @@ public function extract($carrier)
...
     }
+
+    /**
+     * A utility function to mitigate differences between how headers are provided by various web frameworks.
+     * E.g. in both the cases that follow, this method would return 'application/json':
+     *   1) as array of values: ['content-type' => ['application/json']]
+     *   2) as string value: ['content-type0 => 'application/json']
+     *
+     * @param $value
+     * @return mixed|null
```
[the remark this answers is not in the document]
SammyK: Were you able to tackle this one? :)

PR #148, opened 2018-11-27
[Integration] Add Guzzle v5 integration

SammyK:
This adds the Guzzle integration for v5. I also created an abstract `Integration` class that takes on the majority of the boilerplate for the integration without using instances of objects like in #133. If we'd rather not do this change on this PR, let me know and I'll refactor it out. Then we can just try this abstraction in a separate PR. :)

### DIFF
--- a/src/DDTrace/Integrations/Guzzle/V5/GuzzleIntegration.php
+++ b/src/DDTrace/Integrations/Guzzle/V5/GuzzleIntegration.php
@@ -6,6 +6,7 @@ use DDTrace\Tags;
 use DDTrace\Types;
 use OpenTracing\Span;
 use DDTrace\Integrations\Integration;
+use GuzzleHttp\Message\ResponseInterface;
 
 class GuzzleIntegration extends Integration
 {
@@ -15,13 +16,19 @@ class GuzzleIntegration extends Integration
     {
         self::traceMethod('send', function (Span $span, array $args) {
             $span->setTag('http.method', $args[0]->getMethod());
+            $span->setTag('http.url', $args[0]->getUrl());
+        }, function (Span $span, $response) {
+            if (!$response instanceof ResponseInterface) {
+                return;
+            }
+            $span->setTag('http.status_code', $response->getStatusCode());
         });
     }
 
     public static function setDefaultTags(Span $span, $method)
     {
         parent::setDefaultTags($span, $method);
-        $span->setTag(Tags\SPAN_TYPE, Types\GUZZLE);
+        $span->setTag(Tags\SPAN_TYPE, Types\HTTP_CLIENT);
         $span->setTag(Tags\SERVICE_NAME, 'guzzle');
     }
 }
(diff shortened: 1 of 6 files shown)


tests/Integration/Integrations/Guzzle/v5/GuzzleIntegrationTest.php
```diff
@@ -0,0 +1,73 @@
...
+    public function testSend()
+    {
+        $traces = $this->isolateTracer(function () {
+            $request = new Request('put', 'http://example.com');
+            $this->client->send($request);
+        });
+        $this->assertSpans($traces, [
+            SpanAssertion::build('GuzzleHttp\Client.send', 'guzzle', 'guzzle', 'send')
+                ->withExactTags([
+                    'http.method' => 'PUT',
```
labbati: Can we add info about the url and status_code?
https://github.com/DataDog/architecture/blob/willgittoes-dd/apm-priority-sampling/rfcs/apm-priority-sampling/rfc.md

src/DDTrace/Integrations/Guzzle/v5/GuzzleIntegration.php
```diff
@@ -0,0 +1,27 @@
+<?php
```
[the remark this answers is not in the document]
SammyK: Nice catch! :)

src/DDTrace/Integrations/Integration.php
```diff
@@ -0,0 +1,66 @@
...
+        if (!extension_loaded('ddtrace')) {
+            trigger_error('The ddtrace extension is required to trace ' . static::CLASS_NAME, E_USER_WARNING);
+            return false;
+        }
+        if (!class_exists(static::CLASS_NAME)) {
+            trigger_error(static::CLASS_NAME . ' is not loaded and cannot be traced', E_USER_WARNING);
+            return false;
+        }
+        static::loadIntegration();
+        return true;
```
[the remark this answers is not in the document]
SammyK: This is used to check whether or not to skip the tests with this line:

```php
if(!GuzzleIntegration::load()) {
    self::markTestSkipped('Guzzle required to run tests.');
}
```

tests/Integration/Integrations/Guzzle/v5/GuzzleIntegrationTest.php
SammyK: Good call! :)

src/DDTrace/Integrations/Guzzle/V5/GuzzleIntegration.php
```diff
@@ -0,0 +1,34 @@
...
+
+class GuzzleIntegration extends Integration
+{
+    const CLASS_NAME = 'GuzzleHttp\Client';
+
+    protected static function loadIntegration()
+    {
+        self::traceMethod('send', function (Span $span, array $args) {
+            $span->setTag('http.method', $args[0]->getMethod());
+            $span->setTag('http.url', $args[0]->getUrl());
```
labbati: This I forgot to mention: you can use `DDTrace\Http\Urls::sanitize()` to sanitize the url before putting it in the tag.

PR #150, opened 2018-11-28
Add timeouts to Curl Http client

pawelchcki:
Fail fast if there are problems connecting to the agent.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,7 +278,7 @@ jobs:
           command: composer test-integration -- --log-junit test-results/php-integration/results.xml
       - run:
           name: Run << parameters.integration_testsuite >> integration test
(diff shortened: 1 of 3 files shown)


pawelchcki:
Just commit a change that broke some functionality - but all tests have passed. 
Marking it as WIP, until we add some integration tests for Http class

pawelchcki:
We do now have web integration tests that respect these timeouts and if timeout is not increased tests do fail. 
This means that this Http class is now indirectly tested. I think this PR is now ready for another review. 

/cc: @SammyK

pawelchcki:
> ❤️ Love it! Just needs CHANGELOG update and should be good to go! :)

🤦‍♂ I think this PR predated our PR template :D

pawelchcki:
Thanks for the review @SammyK 🙇

PR #154, opened 2018-11-29
Labbati/elastic search

labbati:
### Description

Added Elasticsearch v1.x integration. Al public methods from the client and the various name spaces are traced. In addition to that, http requests to the elastic search cluster are also tracked.

### Readiness checklist
- [ ] [docs/changelog.md](Changelog entry) added, if necessary
- [ x ] Tests added for this feature/bug

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,6 +34,7 @@
         "opentracing/opentracing": "1.0.0-beta5"
     },
     "require-dev": {
+        "elasticsearch/elasticsearch": "1.2.*",
         "guzzlehttp/guzzle": "^5.0",
         "mockery/mockery": "*",
         "phpcompatibility/php-compatibility": "^9.0",
@@ -41,6 +42,7 @@
         "phpcompatibility/phpcompatibility-symfony": "*",
         "phpstan/phpstan": "^0.10.5",
         "phpunit/phpunit": "^5.7.19",
+        "pimple/pimple": "~2.0",
         "predis/predis": "^1.1",
         "squizlabs/php_codesniffer": "^3.3.0",
         "symfony/process": "*"
--- a/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
+++ b/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
@@ -1,13 +1,15 @@
 <?php
 
-namespace DDTrace\Integrations\ElasticSearch;
+namespace DDTrace\Integrations\ElasticSearch\V1;
 
 use DDTrace\Span;
 use DDTrace\Tags;
 use DDTrace\Types;
 use OpenTracing\GlobalTracer;
 
-
+/**
+ * ElasticSearch driver v1 Integration
+ */
 class ElasticSearchIntegration
 {
     const DEFAULT_SERVICE_NAME = 'elasticsearch';
@@ -27,13 +29,21 @@ class ElasticSearchIntegration
         self::traceClientMethod('__construct');
         self::traceClientMethod('count');
         self::traceClientMethod('delete');
-        self::traceClientMethod('exist');
+        self::traceClientMethod('exists');
         self::traceClientMethod('explain');
         self::traceClientMethod('get');
         self::traceClientMethod('index');
         self::traceClientMethod('scroll');
         self::traceClientMethod('search');
         self::traceClientMethod('update');
+
+        // Serializers
+        self::traceMethod('Elasticsearch\Serializers\ArrayToJSONSerializer', 'serialize');
+        self::traceMethod('Elasticsearch\Serializers\ArrayToJSONSerializer', 'deserialize');
+        self::traceMethod('Elasticsearch\Serializers\EverythingToJSONSerializer', 'serialize');
+        self::traceMethod('Elasticsearch\Serializers\EverythingToJSONSerializer', 'deserialize');
+        self::traceMethod('Elasticsearch\Serializers\SmartSerializer', 'serialize');
+        self::traceMethod('Elasticsearch\Serializers\SmartSerializer', 'deserialize');
 
         // IndicesNamespace operations
         self::traceNamespaceMethod('IndicesNamespace', 'analyze');
@@ -99,7 +109,6 @@ class ElasticSearchIntegration
         self::traceNamespaceMethod('SnapshotNamespace', 'status');
 
         // ClusterNamespace operations
-        self::traceNamespaceMethod('ClusterNamespace', 'get');
         self::traceNamespaceMethod('ClusterNamespace', 'getSettings');
         self::traceNamespaceMethod('ClusterNamespace', 'health');
         self::traceNamespaceMethod('ClusterNamespace', 'pendingTasks');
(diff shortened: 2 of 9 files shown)


src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
```diff
@@ -0,0 +1,328 @@
...
+    public static function traceClientMethod($name)
+    {
+        $class = 'Elasticsearch\Client';
+        if (!method_exists($class, $name)) {
+            return;
+        }
+
+        dd_trace($class, $name, function () use ($name) {
+            $args = func_get_args();
+            list($params) = extract($args);
```
SammyK: The `extract()` function returns an `int`, not an `array` so I'm not sure what this line is doing. If we remove it and replace `$params` with `$args` on line 188, the tests still pass for me. :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
```diff
@@ -0,0 +1,328 @@
...
+    public static function traceNamespaceMethod($namespace, $name)
+    {
+        $class = 'Elasticsearch\Namespaces\\' . $namespace;
+        if (!method_exists($class, $name)) {
+            return;
+        }
+
+        dd_trace($class, $name, function () use ($namespace, $name) {
+            $args = func_get_args();
+            list($params) = extract($args);
```
SammyK: Same thing here and replacing `$params` with `$args` on line 283. :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
```diff
@@ -0,0 +1,328 @@
...
+            if ($thrown) {
+                throw $thrown;
+            }
+
+            return $result;
+        });
+    }
+
+    /**
+     * @param $methodName
```
SammyK: Just in case :)
```suggestion
     * @param string $methodName
```

tests/Integration/Common/IntegrationTestCase.php
```diff
@@ -33,4 +33,15 @@ public function assertSpans($traces, $expectedSpans)
     {
         $this->assertExpectedSpans($this, $traces, $expectedSpans);
     }
+
+    /**
+     * Checks that the provide span exists in the provided traces and matches expectations.
+     *
+     * @param $traces
```
SammyK: Not sure if this is the correct type declaration here, but I think it should be:
```suggestion
     * @param Span[][] $traces
```

tests/Integration/Common/SpanAssertionTrait.php
```diff
@@ -16,4 +17,26 @@ public function assertExpectedSpans($testCase, $traces, $expectedSpans)
     {
         (new SpanChecker($testCase))->assertSpans($traces, $expectedSpans);
     }
+
+    /**
+     * Checks that the provide span exists in the provided traces and matches expectations.
+     *
+     * @param TestCase $testCase
+     * @param $traces
```
SammyK: Would this be the same here too?
```suggestion
     * @param Span[][] $traces
```

tests/Integration/Common/SpanAssertionTrait.php
```diff
@@ -16,4 +17,26 @@ public function assertExpectedSpans($testCase, $traces, $expectedSpans)
     {
         (new SpanChecker($testCase))->assertSpans($traces, $expectedSpans);
     }
+
+    /**
+     * Checks that the provide span exists in the provided traces and matches expectations.
+     *
+     * @param TestCase $testCase
```
SammyK: We need to import this type declaration above:
```php
use PHPUnit\Framework\TestCase;
```

src/DDTrace/Integrations/Laravel/V4/LaravelProvider.php
```diff
@@ -93,6 +94,7 @@ public function boot()
         }
 
         PDOIntegration::load();
+        ElasticSearchIntegration::load();
```
SammyK: As stated above, we might want to wrap these with `class_exists('Elasticsearch\Client')` checks. :)

SammyK: Also - it just occurred to me that someone might be using the Laravel integration but a different version of `Elasticsearch\Client` than v1. Not sure how we want to check that off the top of my head. :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
```diff
@@ -0,0 +1,328 @@
...
+
+    public static function load()
+    {
+        if (!extension_loaded('ddtrace')) {
+            trigger_error('The ddtrace extension is required to instrument Eloquent', E_USER_WARNING);
+            return;
+        }
+
+        if (!class_exists('Elasticsearch\Client')) {
+            return;
```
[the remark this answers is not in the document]
labbati: This is actually something I will add to the list of the things to discuss. Until we have integrations that can be turned off by configuration, we should not rise the warning in my opinion. Once we have configurable integration, then we should. Does it make sense to you.

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
```diff
@@ -0,0 +1,328 @@
...
+                // Some endpoints can throw exception during getURI() if some parameters are missing, wso
+                // make sure that the uri is read within the try-catch-finally block.
+                $span->setTag(Tags\RESOURCE_NAME, 'performRequest');
+                $span->setTag(Tags\ELASTICSEARCH_URL, $this->getURI());
+                $span->setTag(Tags\ELASTICSEARCH_METHOD, $this->getMethod());
+                if (is_array($this->params)) {
+                    $span->setTag(Tags\ELASTICSEARCH_PARAMS, json_encode($this->params));
+                }
+                if ($this->getMethod() === 'GET' && $body = $this->getBody()) {
+                    $span->setTag(Tags\ELASTICSEARCH_BODY, json_encode($this->getBody()));
```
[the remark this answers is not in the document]
labbati: It is done by the trace agent: https://github.com/DataDog/datadog-trace-agent/blob/master/datadog.example.yaml#L108

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
labbati: very good catch! I think that I actually should convert to
`list($params) = $args` as different methods can have other stuff into the method signature.
Thoughts? :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
labbati: Same as above

src/DDTrace/Integrations/Laravel/V4/LaravelProvider.php
labbati: This is also something that we need to discuss. Ideally, in the spirit of simplifying onboarding, we should detect ourself the version. In this case I added some guards to protect us from crashes: e.g.
```        $class = 'Elasticsearch\Namespaces\\' . $namespace;
        if (!method_exists($class, $name)) {
            return;
        }```

In my mind each integration should have a `DDTrace\Integrations\ElasticSearchIntegration::load()` which either guess from `xxxx_exist` or read from configuration and delegates the actuall loading logic to `DDTrace\Integrations\Vx\ElasticSearchIntegration::load()`. But this comes just from the top of my head.

File: composer.json [hunk 10]
```diff
@@ -34,13 +34,15 @@
...
     "require-dev": {
+        "elasticsearch/elasticsearch": "1.2.*",
         "guzzlehttp/guzzle": "^5.0",
         "mockery/mockery": "*",
         "phpcompatibility/php-compatibility": "^9.0",
         "phpcompatibility/phpcompatibility-passwordcompat": "^1.0",
         "phpcompatibility/phpcompatibility-symfony": "*",
         "phpstan/phpstan": "^0.10.5",
         "phpunit/phpunit": "^5.7.19",
+        "pimple/pimple": "~2.0",
```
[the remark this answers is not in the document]
labbati: It was a left over. Good catch

tests/Integration/Integrations/Elasticsearch/V1/ElasticSearchIntegrationTest.php
```diff
@@ -0,0 +1,518 @@
...
+
+    public function testExplain()
+    {
+        $client = $this->client();
+        $client->index([
+            'id' => 1,
+            'index' => $this->envSpecificIndexName(),
+            'type' => 'my_type',
+            'body' => ['my' => 'body'],
+        ]);
```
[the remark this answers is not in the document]
labbati: this was of huuuuuge help. Thanks

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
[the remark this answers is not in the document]
SammyK: Sounds good! Looking forward to discussing it! :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
SammyK: Sounds good! If all we need is the first argument, we could even do this and remove `$args` altogether:

```php
list($params) = func_get_args();
```

PR #160, opened 2018-11-30
Possibility to enable/disable distributed tracing and priority sampling

labbati:
### Description

This PR adds the possibility to enable and disable distributed tracing and priority sampling

### Readiness checklist
- [x] [docs/changelog.md](Changelog entry) added, if necessary

### DIFF
--- a/src/DDTrace/Configuration.php
+++ b/src/DDTrace/Configuration.php
@@ -16,7 +16,7 @@ class Configuration extends AbstractConfiguration
      */
     public function isDistributedTracingEnabled()
     {
-        return $this->registry->boolValue('distributed.tracing', true);
+        return $this->boolValue('distributed.tracing', true);
     }
 
     /**
@@ -27,6 +27,6 @@ class Configuration extends AbstractConfiguration
     public function isPrioritySamplingEnabled()
     {
         return $this->isDistributedTracingEnabled()
-            && $this->registry->boolValue('priority.sampling', true);
+            && $this->boolValue('priority.sampling', true);
     }
 }
--- a/src/DDTrace/Configuration.php
+++ b/src/DDTrace/Configuration.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace DDTrace;
+
+use DDTrace\Configuration\AbstractConfiguration;
+
+/**
+ * DDTrace global configuration object.
+ */
+class Configuration extends AbstractConfiguration
+{
+    /**
+     * Whether or not distributed tracing is enabled globally.
+     *
+     * @return bool
+     */
+    public function isDistributedTracingEnabled()
+    {
+        return $this->registry->boolValue('distributed.tracing', true);
+    }
+
+    /**
+     * Whether or not priority sampling is enabled globally.
+     *
+     * @return bool
+     */
+    public function isPrioritySamplingEnabled()
+    {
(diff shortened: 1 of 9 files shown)


src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace DDTrace\Configuration;
+
+/**
+ * Registry that holds configuration properties and that is able to recover configuration values from environment
+ * variables.
+ */
+class EnvVariableRegistry
```
pawelchcki: Would it make sense to extract interface of this?

src/DDTrace/Configuration.php
```diff
@@ -0,0 +1,32 @@
...
+ * DDTrace global configuration object.
+ */
+class Configuration extends AbstractConfiguration
+{
+    /**
+     * Whether or not distributed tracing is enabled globally.
+     *
+     * @return bool
+     */
+    public function isDistributedTracingEnabled()
```
pawelchcki: I'm stumped here a bit. one hand I really like that we're using statically analyzable code to fetch configuration.

OTOH, adding a method for every configuration option - I feel like it might be a bit of an overkill in the long run.

WDYT if AbstractConfiguration implemented also `VariableRegistry` interface. 
So that boolValue could be reimplemented here.
And we could use this configuration like so:

`$globalConfiguration->boolValue('distributed.tracing');`

Hmm. This also means that we would have split the configuration defaults from the implementation. Maybe to a `DefaultVariableRegistry` ?

I might be optimizing a bit for the future here, so just thinking out loud. Otherwise I really like this approach!

src/DDTrace/Configuration/EnvVariableRegistry.php
labbati: Absolutely, let's do it

src/DDTrace/Configuration.php
labbati: Ehehehe the solution that you proposed it much more php-style, so I would be totally for that.
To be honest, I really like to have statically analysable methods and I don't think that they will be dozens.
So I still gives my vote to methods 😄 but totally opened to change it if you guys prefer 😄

pawelchcki: I have a feeling we'll end up with both, and it should be cool too. I.e. the big things like `isDistrubutedTracingEnabled()` is really good case for methods approach.

The small things like:
`agent.connection.timeout`

labbati: TBH I am not a big fan of allowing both 😄 but I also acknowledge that the string approach is more in line with PHP approach. So let me add both the access mechanism...but I hope that everybody will use the explicit method call 😄.

labbati:
@pawelchcki I applied changes from your comments, it would be great if you have some time to take it a look for second round of code review.

src/DDTrace/Configuration.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        return $this->registry->boolValue('distributed.tracing', true);
```
pawelchcki: Can you use `$this->boolValue(...`  here ?

labbati: eheheh very good catch!!! 👍 thanks

PR #161, opened 2018-12-03
Guzzle and Curl enabling for Laravel integrations

labbati:
…

### Readiness checklist


src/DDTrace/Integrations/Laravel/V5/LaravelProvider.php
```diff
@@ -163,6 +163,7 @@ public function boot()
         });
 
         // Enable extension integrations
+        CurlIntegration::load();
         ElasticSearchIntegration::load();
         EloquentIntegration::load();
         if (class_exists('Memcached')) {
```
labbati: Not enabling Guzzle for Laravel 5 because there is an higher probability, compared to Laravel 4, that Guzzle v6 is required instead of the supported v5.

PR #166, opened 2018-12-04
MongoDB legacy integration

SammyK:
### Description

Adds an integration for the [legacy mongo extension](http://php.net/manual/en/book.mongo.php). So far this traces the following classes:

* `MongoClient`
* `MongoDB`
* `MongoCollection`

If you think we need to trace more than this, let me know. Also, this tags the BSON `MongoID` where it finds it and I'm not sure if we want to trace the ID or not. Just let me know! :)

### Readiness checklist


SammyK:
It looks like the CI tests are passing because the mongo extension [I tried to pull in for 5.6](https://github.com/DataDog/docker-library/pull/21) isn't working so it's skipping them. Thoughts? :)

.circleci/config.yml
```diff
@@ -40,6 +40,14 @@ aliases:
       - MYSQL_USER=test
```
labbati: Can we please add a change log entry? :)

.circleci/config.yml
```diff
@@ -40,6 +40,14 @@ aliases:
...
       - MYSQL_DATABASE=test
 
+  - &IMAGE_DOCKER_MONGODB
+    image: bitnami/mongodb:4.0
```
labbati: Any specific reason why we are not using `image: circleci/mongo:4.0`? https://circleci.com/docs/2.0/circleci-images/#mongodb

docker-compose.yml
```diff
@@ -47,6 +48,15 @@ services:
   elasticsearch2_integration:
     image: "elasticsearch:2"
 
+  mongodb_integration:
+    image: "bitnami/mongodb:4.0"
```
labbati: If we use circleci images in CI then we can also use them here.

src/DDTrace/Integrations/Mongo/MongoClientIntegration.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
labbati: Great work overall, it would be great if we could add some comments to the various methods and classes. :)

src/DDTrace/Integrations/Mongo/MongoClientIntegration.php
```diff
@@ -0,0 +1,59 @@
...
+
```
labbati: Given that this is a legacy driver, then I would move it upfront to a Vx folder because we know we will have the new one.

src/DDTrace/Integrations/Mongo/MongoIntegration.php
```diff
@@ -0,0 +1,28 @@
...
+use DDTrace\Integrations\Integration;
+
+final class MongoIntegration
+{
+    public static function load()
+    {
+        if (!extension_loaded('mongo')) {
+            return;
+        }
+        MongoClientIntegration::load();
```
labbati: We should add this at least to Laravel4 integration. We can hold on the other ones until we have a way to disable specific integrations.

src/DDTrace/Integrations/Mongo/MongoClientIntegration.php
SammyK: Since this is a PECL extension, I based the folder names off of the name of the extension rather than the version number since the version numbers would collide between different extensions. The legacy extension is called [mongo](http://php.net/mongo) and the new one is called [mongodb](http://php.net/mongodb). So I figured we could just put the new extension under the `src/DDTrace/Integrations/MongoDB` folder. Would that work for you? :)

SammyK:
I've updated the PR based on the feedback - thanks @labbati! :) The PHP 5.6 docker image still doesn't seem to pull in the latest version with the mongo extension installed. Any thoughts on getting that to work? :)

src/DDTrace/Integrations/Mongo/MongoClientIntegration.php
labbati: Perfect

labbati: 🔝

labbati:
It looks great, just one thing from the top of my head @SammyK. Did you try to send some real traces tot he UI? How many "levels" of tracing you see? Just to make sure that we are not tracing too much. E.g. in the case of the methods that we trace are mutually exclusive, so the traces only go no more than 2/3 levels. Just curious if you played with it.

src/DDTrace/Integrations/Laravel/V4/LaravelProvider.php
```diff
@@ -99,6 +100,7 @@ public function boot()
 
         PDOIntegration::load();
         ElasticSearchIntegration::load();
+        MongoClientIntegration::load();
```
labbati: Please rebase on master and you will see you have to add this to the list of autoloaded integrations

tests/Integration/Integrations/Mongo/MongoTest.php
```diff
@@ -0,0 +1,705 @@
...
+{
+    const HOST = 'mongodb_integration';
+    const PORT = '27017';
+    const USER = 'test';
+    const PASSWORD = 'test';
+    const DATABASE = 'admin';
+
+    public static function setUpBeforeClass()
+    {
+        if (!extension_loaded('mongo')) {
```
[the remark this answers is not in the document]
SammyK: The main issue with making it always required is that `ext/mongo` only works on PHP 5. So I could change this check to just skip the tests if the PHP version is >= 7.0 if you think that'll have the same effect. :)

SammyK:
I think I came up with an elegant solution to our skip-tests woes. I [created a new `tests/IntegrationPHP5`](https://github.com/DataDog/dd-trace-php/pull/166/commits/00bc29c18e59020fd8fd8a8f00af14d688aff291) directory that only gets run on PHP 5 builds. Now when you run the tests on PHP 7, you don't get a huge list of skipped tests. :)

I also refactored this based on the new auto loading feature. :)

SammyK:
Thanks for all the help with this @labbati! Here's a screenshot of the trace we did together:

screen shot 2018-12-05 at 10 46 32 am

src/DDTrace/Integrations/Mongo/MongoClientIntegration.php
```diff
@@ -0,0 +1,66 @@
...
+        self::traceMethod('getHosts');
+        self::traceMethod('getReadPreference');
+        self::traceMethod('getWriteConcern');
+        self::traceMethod('listDBs');
+        self::traceMethod('setWriteConcern');
+    }
+
+    // If the `db` option isn't provided via the constructor, we extract
+    // the database name from the DSN string if it exists.
+    private static function extractDatabaseNameFromDsn($dsn)
```
labbati: Sorry for not noticing earlier, should it be a docblock comment?

src/DDTrace/Integrations/Mongo/MongoCollectionIntegration.php
```diff
@@ -0,0 +1,110 @@
...
+        self::traceMethod('getWriteConcern');
+        self::traceMethod('group');
+        self::traceMethod('insert');
+        self::traceMethod('parallelCollectionScan');
+        self::traceMethod('save');
+        self::traceMethod('setWriteConcern');
+        self::traceMethod('validate');
+    }
+
+    public static function setDefaultTags(Span $span, $method)
```
labbati: Dockblock missing :(

src/DDTrace/Integrations/Mongo/MongoDBIntegration.php
```diff
@@ -0,0 +1,75 @@
...
+        self::traceMethod('getProfilingLevel');
+        self::traceMethod('getReadPreference');
+        self::traceMethod('getWriteConcern');
+        self::traceMethod('lastError');
+        self::traceMethod('listCollections');
+        self::traceMethod('repair');
+        self::traceMethod('setWriteConcern');
+    }
+
+    public static function setDefaultTags(Span $span, $method)
```
labbati: Dockblock missing :(

src/DDTrace/Integrations/Mongo/MongoClientIntegration.php
SammyK: Thanks for pointing that out! I'm so used to coding for PHP 7.2 that I don't use DocBlocks nearly as much as I used to. Hehe. I'll add DocBlocks to the ones you pointed out! :)

PR #167, opened 2018-12-04
Propagate distributed tracing

labbati:
### Description

This pool request add distributed tracing headers propagation for guzzle and curl integration

### Readiness checklist


src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -50,5 +57,52 @@ public static function load()
...
+        if (!Configuration::instance()->isDistributedTracingEnabled()) {
+            return;
+        }
+
+        $currentHttpHeaders = ArrayKVStore::getForResource($ch, 'http_headers', []);
+        if (is_array($currentHttpHeaders)) {
+            $tracer = GlobalTracer::get();
+            $context = $tracer->getActiveSpan()->getContext();
+            $carrier = Headers::colonSeparatedValuesToHeadersMap($currentHttpHeaders);
+            $tracer->inject($context, HTTP_HEADERS, $carrier);
```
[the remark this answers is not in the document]
labbati: I think I may not follow here.
The problem I had is that different libraries handle headers differently. So for example, in the case of duplicate headers you may receive `['headers1: value1, value2']` for curl and `['header1' => ['value1', 'value2']]` in Guzzle. So would an additional propagator resolve this problem? :(

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -50,5 +57,52 @@ public static function load()
...
+
+            curl_setopt($ch, CURLOPT_HTTPHEADER, Headers::headersMapToColonSeparatedValues($carrier));
```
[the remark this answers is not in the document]
labbati: this is a good idea, let me see how I can change the implementation.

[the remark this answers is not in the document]
labbati: Ahhh now I remember, it has to be in the exec method to have the proper span activated (for parent span id).

src/DDTrace/Integrations/Curl/CurlIntegration.php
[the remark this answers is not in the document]
pawelchcki: Then possible two additional propagators might be in order. 
My point is however that in case of curl we can expect values like:
`[['header' => 'value1"], ['header' => 'value2']]` so converting the parameters first to associative array
means that we will lose `['header' => 'value']` when converting back array of tuples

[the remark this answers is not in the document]
labbati: > My point is however that in case of curl we can expect values like
Ah ok I did not know this. Looking at them it looked like I only could get them in the form `['header: value1, value2']`, but let me check.

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -50,5 +55,56 @@ public static function load()
...
+    {
+        if (!Configuration::instance()->isDistributedTracingEnabled()) {
+            return;
+        }
+
+        $httpHeaders = ArrayKVStore::getForResource($ch, Formats\CURL_HTTP_HEADERS, []);
+        if (is_array($httpHeaders)) {
+            $tracer = GlobalTracer::get();
+            $context = $tracer->getActiveSpan()->getContext();
+            $tracer->inject($context, Formats\CURL_HTTP_HEADERS, $httpHeaders);
```
pawelchcki: ❤️  😄

src/DDTrace/Util/ArrayKVStore.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace DDTrace\Util;
+
+/**
+ * A key value store that stores metadata into an array. If you have an object that you can use as a carrier, then
+ * prefer ObjectKVStore as is provides a better performance. Use this if you do not have an object you can use as
+ * a carrier.
+ */
+class ArrayKVStore
```
pawelchcki: 👍

src/DDTrace/Util/ArrayKVStore.php
```diff
@@ -0,0 +1,97 @@
...
+    private static function getResourceKey($resource)
+    {
+        // Converting to integer a resource results in the "unique resource number assigned to the resource by PHP at
+        // runtime":
+        //   - http://php.net/manual/en/language.types.integer.php#language.types.integer.casting
+        // Resource ids are guaranteed to be unique per script execution:
+        //   - http://www.php.net/manual/en/language.types.string.php#language.types.string.casting
+        return intval($resource);
+    }
+}
```
pawelchcki: This class is 💯

tests/Unit/Util/ArrayKVStoreTest.php
```diff
@@ -0,0 +1,52 @@
...
+
+final class ArrayKVStoreTest extends TestCase
+{
+    private $resource1;
+    private $resource2;
+
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->resource1 = fopen('php://memory', 'r');
```
pawelchcki: TIL :D

tests/Integration/Frameworks/Laravel/Version_4_2/composer.json
```diff
@@ -25,6 +25,7 @@
 			"DDTrace\\": "../../../../../src/DDTrace/"
 		},
 		"files": [
+			"../../../../../src/DDTrace/Formats.php",
```
pawelchcki: 😢 would be great to find some approach - not requiring this.
But this is out of scope of this PR.

labbati: I know :(

PR #168, opened 2018-12-04
Let autoload all integrations and disable specific ones

labbati:
### Description

This PR adds the possibility to have a unique and safe way to load all the libraries (not frameworks) integrations and to disable specific ones by means of env variable.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Integrations/IntegrationsLoader.php
+++ b/src/DDTrace/Integrations/IntegrationsLoader.php
@@ -21,7 +21,7 @@ class IntegrationsLoader
         CurlIntegration::NAME => '\DDTrace\Integrations\Curl\CurlIntegration',
         ElasticSearchIntegration::NAME => '\DDTrace\Integrations\ElasticSearch\V1\ElasticSearchIntegration',
         EloquentIntegration::NAME => '\DDTrace\Integrations\Eloquent\EloquentIntegration',
-        GuzzleIntegration::NAME => '\DDTrace\Integrations\Guzzle\GuzzleIntegration',
+        GuzzleIntegration::NAME => '\DDTrace\Integrations\Guzzle\V5\GuzzleIntegration',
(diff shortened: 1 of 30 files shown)


src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -0,0 +1,53 @@
...
+    ];
+
+    /**
+     * Loads all the enabled library integrations.
+     */
+    public static function load()
+    {
+        $globalConfig = Configuration::get();
+
+        if (!$globalConfig->isEnabled()) {
```
pawelchcki: ❤️

src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -0,0 +1,53 @@
...
+            return;
+        }
+
+        if (!extension_loaded('ddtrace')) {
+            error_log('Missing ddtrace extension. To disable tracing set env variable DD_TRACE_ENABLED=false');
```
pawelchcki: would - since we allow now for things to be disabled `$globalConfig->isEnabled()` - would it make sense to bring `E_USER_WARNING` back instead of `error_log?

Sort of makes sense i think, especially since we're adding instructions how to disable this via DD_TRACE_ENABLED=false
WDYT ?

src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -0,0 +1,53 @@
...
+            return;
+        }
+
+        foreach (self::LIBRARIES as $name => $class) {
+            if ($globalConfig->isIntegrationEnabled($name)) {
+                call_user_func([$class, 'load']);
```
pawelchcki: ❤️

src/DDTrace/Integrations/IntegrationsLoader.php
labbati: This is a good idea actually, let me do it very quick!

PR #169, opened 2018-12-05
Add priority sampling

labbati:
### Description

This PR adds priority sampling management to the game. In the process, some refactoring to the way we handle distributed tracing in general was necessary.

### Readiness checklist


src/DDTrace/Transport/Http.php
```diff
@@ -62,19 +66,24 @@ private function configure($config)
     {
         $host = getenv(self::AGENT_HOST_ENV) ?: self::DEFAULT_AGENT_HOST;
         $port = getenv(self::TRACE_AGENT_PORT_ENV) ?: self::DEFAULT_TRACE_AGENT_PORT;
-        $path = self::DEFAULT_TRACE_AGENT_PATH;
-        $endpoint = "http://${host}:${port}${path}";
+        $defaultEndpoint = "http://${host}:${port}" . self::DEFAULT_TRACE_AGENT_PATH;
+        $prioritySamlpingEndpoint = "http://${host}:${port}" . self::PRIORITY_SAMPLING_TRACE_AGENT_PATH;
```
pawelchcki: I know this is probably not very bad for performance. 
But it worries me a bit that we're doing some computations every request, that are not normally used during the request and after.

Overall would be great to defer most of the transport setup to the very end when we need to actually send  (but this is probably out of the scope for this PR).

src/DDTrace/SpanContext.php
```diff
@@ -113,4 +156,22 @@ private static function nextId()
     {
         return bin2hex(random_bytes(8));
     }
+
+    /**
+     * @return bool
+     */
+    public function isDistributedTracingActivationContext()
```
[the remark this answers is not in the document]
labbati: Added tot he property description of this is ok with you

src/DDTrace/Transport/Http.php
labbati: This is a good point and I like the idea. If you agree I will do for the parts that I touched in this PR and will postpone a wider refactoring of the Http transport (as the same applies also to other config params) to a later stage.

src/DDTrace/Transport/Http.php
```diff
@@ -128,4 +137,30 @@ private function sendRequest($url, array $headers, $body)
...
+
+    /**
+     * Returns whether or not we should send these traces to the priority sampling aware trace agent endpoint.
+     * This approach could be optimized in the future if we refactor how traces are organized in parent/child relations
+     * but this would be out of scope at the moment.
+     *
+     * @param array $traces
+     * @return bool
+     */
+    private function isPrioritySamplingEndpoint(array $traces)
```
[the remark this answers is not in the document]
labbati: totally 👍

src/DDTrace/Transport/Http.php
labbati: As discussed, it is risky as it introduces potential breaking changes for users, so I would defer it to a later http transport refactoring

labbati:
@pawelchcki I did some manual testing and look good. It would be great to have your latest CR round. Thanks a lot!

PR #173, opened 2018-12-06
Fix compatibility with PCS extension

pawelchcki:
### Readiness checklist


SammyK:
So I've been digging deeper into this and I'm thinking it's an incompatibility issue that the PCS extension has with PHP 7.2.

The latest version of the PCS extension (1.3.3) was [released on Jan 20, 2017](https://pecl.php.net/package/pcs) and PHP 7.2 wasn't [released until Nov 30, 2017](http://www.php.net/ChangeLog-7.php#7.2.0) so PHP 7.2 didn't come out until almost a year after the latest version of the PCS extension.

On a clean build of PHP 7.2 with debug mode & ZTS:

```bash
$ php -v
PHP 7.2.14-dev (cli) (built: Dec  6 2018 16:28:54) ( ZTS DEBUG )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.14-dev, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.6.1, Copyright (c) 2002-2018, by Derick Rethans
```

If I try to list the installed extensions after installing pcs (and no ddtrace at all), I get the following error:

```bash
$ php -m
Assertion failed: ((&ht->nInternalPointer != pos || (ht)->gc.refcount == 1) || ((ht)->u.flags & (1<<6))), function zend_hash_internal_pointer_reset_ex, file %s/Zend/zend_hash.c, line 2052.
Abort trap: 6
```

Thoughts? :)

pawelchcki:
The PCS extension seems to be abandoned at this point :/
However it does look like the fault was on our side where we were setting up the wrapping of opcodes on MINIT hook while the datastructures were actually initialized only in RINIT.

So if some extension managed to sneak in compiled PHP code before the our RINIT hook was run it resulted in the exception we were seeing. 

I've added some code to make sure we also initialize the datastructures on MINIT allowing any arbitrary PHP code to be run between our MINIT and RINIT .

Also added `99-` prefix to the ddtrace.ini file when installing the package, to make sure we're the last module loaded and follow the convention of adding numerals to control sorting

PR #175, opened 2018-12-07
Request hook implementation

pawelchcki:
### Description
Initial mplementation of request hook, allowing hooking arbitrary php in the request initialization pipeline.

Its implementation is based on `spl_autoload` method.

~It currently only works with php 7.x (tested with 7.2). But the code for 5.x is quite similar - also already been implemented, but not yet tested.~

It introduces new configuration option `ddtrace.request_init_hook` that can be used to configure the path to a file that is to be executed during request.

### Readiness checklist
- [x] Reorganize the code (move PHP5 and 7 specific implementations to separate files)
- [x] Add some sanity checking - file exists, is readable etc
- [x] ~Do some benchmarks to measure the impact~
- [x] ~Log errors, i.e. if file is configured to be loaded, log if we cannot load it.~


src/ext/ddtrace.c
```diff
@@ -89,9 +95,103 @@ static PHP_RINIT_FUNCTION(ddtrace) {
     zend_hash_init(&DDTRACE_G(class_lookup), 8, NULL, (dtor_func_t)table_dtor, 0);
     zend_hash_init(&DDTRACE_G(function_lookup), 8, NULL, (dtor_func_t)ddtrace_class_lookup_free, 0);
 
+    zend_string *filename = zend_string_init(DDTRACE_G(request_init_hook), strlen(DDTRACE_G(request_init_hook)), 0);
+
+    if (filename && ZSTR_LEN(filename)!= 0){
+        run_a_file(filename);
+    }
```
SammyK: Just need a little cleanup here. :)
```suggestion
    }
    zend_string_release(filename);
```

SammyK:
One other thing we might want to consider is making use of the [`auto_prepend_file` ini setting](http://php.net/manual/en/ini.core.php#ini.auto-prepend-file). :)

pawelchcki:
> One other thing we might want to consider is making use of the [`auto_prepend_file` ini setting](http://php.net/manual/en/ini.core.php#ini.auto-prepend-file). :)

@SammyK Interesting, does this method stack values, or does it override previously set value?

pawelchcki:
@labbati rebased but looks like dev is behind in history vs master

pawelchcki:
@labbati https://github.com/DataDog/dd-trace-php/pull/233

PR #187, opened 2018-12-14
Predis integration supporting constructor options as an object

raffaellopaletta:
### Description

Bug fix on predis integration class (avoid exception Cannot use object of type Predis\Configuration\Options as array)


src/DDTrace/Integrations/Predis/PredisIntegration.php
```diff
@@ -158,8 +158,11 @@ public static function storeConnectionParams($predis, $args)
 
         if (isset($args[1])) {
             $options = $args[1];
-            if (isset($options['parameters']) && isset($options['parameters']['database'])) {
-                $tags['out.redis_db'] = $options['parameters']['database'];
+
+            $parameters = $options->__get('parameters');
```
labbati: Thanks for noticing this. Actually it seems to me that options can be provided both as an array and as an object instance: https://github.com/nrk/predis/blob/111d100ee389d624036b46b35ed0c9ac59c71313/src/Client.php#L71

So what do you think about doing something like:

```
if (is_array($options)) {
    $parameters = isset($options['parameters']) ? $options['parameters'] : [];
} elseif ($options instanceof OptionsInterface) {
    $parameters = $options->__get('parameters') ?: [];
}
```

If you like the idea, would you please add a test case for when we have options as object?
https://github.com/DataDog/dd-trace-php/blob/455e89ac54528eb7129e193ddc096bbf67171074/tests/Integration/Integrations/Predis/PredisTest.php#L42

labbati:
Also, please rebase/pull master as we have changed a few things in that integration due to php 5.4 compatibility

src/DDTrace/Integrations/Predis/PredisIntegration.php
raffaellopaletta: Hi Luca, it seems a good solution, I like it! I'll add the test case asap, thank you.

tests/Integration/Integrations/Predis/PredisTest.php
```diff
@@ -48,6 +49,19 @@ public function testPredisConstruct()
             $this->assertNotNull($client);
         });
```
labbati: In general we prefer (and try to enforce) tests that only test one specific thing. This way it is clear what we may "miss" of we, at a certain point", remove that test. 😄 

In this case I would create a new test instead of editing the current one. Specifically:

 - I would rename `testPredisConstruct` to `testPredisConstructOptionsAsArray`
 - Create an exact same replica `testPredisConstructOptionsAsObject` which contains the part that you added to the existing test.

This way we know exactly what we miss if we remove one of the two.

raffaellopaletta: Of Course, I'll do it soon

raffaellopaletta: Done!

labbati:
Noooo we have conflicts with master. Thanks for the updates!

Please pull `master` and resolve the following conflicts, then I will approve right away.

In `PredisIntegration.php` expected final `use` statements are:
```
use DDTrace\Tags;
use DDTrace\Types;
use DDTrace\Util\TryCatchFinally;
use OpenTracing\GlobalTracer;
use Predis\Configuration\OptionsInterface;
use Predis\Pipeline\Pipeline;
```

In `PredisTest.php` expected final `use` statements are:
```
use DDTrace\Integrations\IntegrationsLoader;
use DDTrace\Tests\Common\IntegrationTestCase;
use DDTrace\Tests\Common\SpanAssertion;
use Predis\Configuration\Options;
use Predis\Response\Status;
```

There is still something broken on CI but it is not related to your work so I will figure out what it is.
Thanks again for your patience!

labbati:
Hey @raffaellopaletta thank for working on it! I can merge if you are cool with that!

PR #188, opened 2018-12-14
Verify packages for manually installed php on centos

pawelchcki:
### Readiness checklist


.circleci/config.yml
```diff
@@ -459,19 +459,19 @@ jobs:
...
   build_packages:
     jobs:
       - "5.6 20131106": &BUILD_PACKAGE_FORKFLOW
-          filters:
-            tags:
-              only: /(^build$)|(^[v]?[0-9]+(\.[0-9]+)*$)/
-            branches:
-              # Always build on master
-              ignore: /^(?!master$).*/
+          filters: {}
```
labbati: is this intended?

pawelchcki: No - its just temporary - to unblock the CI. I'll remove it before merging

package/post-install.sh
```diff
@@ -8,15 +8,81 @@ INI_FILE_NAME='99-ddtrace.ini'
...
+    println
+    println "    extension=${EXTENSION_FILE_PATH}"
+    println
+    println "Note that your PHP API version must match the extension's API version"
+    println "PHP API version can be found using following command"
+    println
+    println "    php -i | grep 'PHP API'"
+    println
+
+    exit 0 # exit - but do not fail the installtion
```
labbati: the whole meaning of this is that we still want to succeed so user can go and add it manually. is it my interpretation correct?

pawelchcki: Yes - exactly. If someone has a very custom setup it still should be possible to install the package correctly.

PR #193, opened 2018-12-17
Remove OpenTracing dependency

SammyK:
### Description

This PR moves the `opentracing/opentracing` dependency to the dev dependencies so that it is no longer required to use ddtrace. The cleanest way I found to do this was to put all the OT stuff under the `DDTrace\OpenTracing` namespace so we can slowly refactor all the OT stuff out over time.

Although all the tests pass and there is an [upgrade guide](https://github.com/DataDog/dd-trace-php/compare/sammyk/remove-opentracing?expand=1#diff-240b0af6ce02e96f9535f637ede52bbe), there are a few issues to address to make real-world upgrades as seamless as possible:

#### The tracer singleton

The original idea was for `DDTrace\GlobalTracer::get()` be backwards compatible with `OpenTracing\GlobalTracer::set()` such that `DDTrace\GlobalTracer::get()` would just fall back to `OpenTracing\GlobalTracer::get()` if there wasn't a singleton tracer set with the DDTrace one (basically so that all the old code snippets from the docs would still work after this PR.)

The problem is that `DDTrace\Tracer` now implements `DDTrace\OpenTracing\Tracer` instead of `OpenTracing\Tracer`, so it is no longer compatible with `OpenTracing\GlobalTracer::set()` which expects an instance of `OpenTracing\Tracer`.

There are two ways I can think of to address this:

1. Make people have to update references from `OpenTracing\GlobalTracer::set()` to `DDTrace\GlobalTracer::set()`. This is less than ideal since it's a BC break.
2. Keep `DDTrace\Tracer` as is (an implementation of `OpenTracing\Tracer`) and create a new class `DDTrace\DDTracer` that implements `DDTrace\OpenTracing\Tracer`. We could mark `DDTrace\Tracer` as deprecated and we can remove it for the stable release.

Option 2 would give users the simplest upgrade path since it wouldn't require them updating any of their code while the tracer is still in beta.

I'd love to hear your thoughts! :)

### Readiness checklist


src/DDTrace/GlobalTracer.php
```diff
@@ -0,0 +1,52 @@
...
+    public static function get()
+    {
+        if (null !== self::$instance) {
+            return self::$instance;
+        }
+        if (!class_exists('\OpenTracing\GlobalTracer')) {
+            return self::$instance = NoopTracer::create();
+        }
+        $tracer = \OpenTracing\GlobalTracer::get();
+        if (!$tracer instanceof \OpenTracing\NoopTracer) {
```
labbati: I think we should never return an opentracing tracer (even a noop), as the `NoopSpan` returned by it would not be compatible with our integrations. :)

labbati:
Hey @SammyK great job, a lot of interesting stuff. I'd like to discuss a few details which are not perfectly clear to me (related to removing opentracing dependency in general, not to the code)

labbati:
Hey @SammyK, I was looking at the new code. It seems very good stuff to me 😄 

> That in turn ended up adding a lot more methods to the interfaces under `DDTrace\Contracts`.

Looking at the  `DDTrace\Contracts` folder, it seems to me that the number of interfaces/methods is totally legit. And by the way those contracts are not strictly required, we have them because our original implementation was, per-se, a pure 100% OT implementation and so it makes things much more easier to transition away from it while still fully supporting OT 😄 

> I'm not sure this is 100% the route we want to go, but the OT interfaces don't seem sufficient for everything we need them to do.

Absolutely open to other approaches, if you see any code/architecture smell feel free to propose other approaches. This is the very central part of our architecture, so we want to get it right 😄 

In the end, our end goal, if for users to be able to do `OpenTracing\GlobalTracer::get()` in their own code and be able to use that tracer as an OT tracer. 

So, at least for me, so far so (very) good 😄

SammyK:
Thanks for the review @labbati! :)

> This is the very central part of our architecture, so we want to get it right 😄

Agreed! I'd love to really spend the time to get this right - especially since it introduces a breaking change. :)

> In the end, our end goal, if for users to be able to do OpenTracing\GlobalTracer::get() in their own code and be able to use that tracer as an OT tracer.

I think we're getting pretty close to that! :)

SammyK:
Thanks for the very detailed feedback @labbati! I do like Option 2, but I want to make sure I understand with code examples. Say for example someone is already using the [Jaeger Bindings for OpenTracing](https://github.com/jonahgeorge/jaeger-client-php) package and they want to be able to try out ddtrace without having to change much code.

Given the snippet of setup code from their getting-started doc:

```php
use Jaeger\Config;
use OpenTracing\GlobalTracer;

$config = new Config([/* ... */], 'your-app-name');
$config->initializeTracer();

$scope = $tracer->startActiveSpan('TestSpan', []);
$scope->close();

$tracer->flush();
```

Once they got an instance of the OT tracer, they simply set the DD tracer singleton and load the integrations like so:

```php
\DDTrace\GlobalTracer::set($tracer);
\DDTrace\Integrations\IntegrationsLoader::load();
```

Does that sound about right?

If they don't change any of their other code, their OT tracer will be doing the `flush()`ing at the end. So wouldn't that require we make the OT tracer the storage mechanism in this case? And if that is the case, how do we access things like tags on OT spans for testing (since the OT interface has a `setTag()` but not a `getTag()`).

Or did I just miss what Option 2 was about? Hehe. :)

labbati:
Hey @SammyK, totally opened to have a conversation about the approach and I love this thread and talking upfront about possible things we may have not considered.

About Opt#2 and the example you posted:

This part would not be required anymore as we are in charge of doing it.
```
use Jaeger\Config;
use OpenTracing\GlobalTracer;

$config = new Config([/* ... */], 'your-app-name');
$config->initializeTracer();
```

Users can continue to do the usage nelow in their code, so we provide support for OpenTracing:

```
$tracer = OpenTracing\GlobalTracer::get();

$scope = $tracer->startActiveSpan('TestSpan', []);
$scope->close();

$tracer->flush();
```

The following part is generally not required to be done by users, as we make sure we create and set the global tracer

```
// Not typically required
\DDTrace\GlobalTracer::set($tracer);
\DDTrace\Integrations\IntegrationsLoader::load();
```

> If they don't change any of their other code, their OT tracer will be doing the `flush()`ing at the end. So wouldn't that require we make the OT tracer the storage mechanism in this case? And if that is the case, how do we access things like tags on OT spans for testing (since the OT interface has a `setTag()` but not a `getTag()`).

Good point. The main aspect of Opt#2 is the the `OTTracerWrapper` (returned by `OpenTracing\GlobalTracer::get()`) is our wrapper which contains our tracer and implements `OpenTracing\Tracer`. So when this is `flush()`ed it just forward the flush to the underlying `DDTrace\Tracer::flush()` which does the real job.

Does it make sense?

src/DDTrace/Encoder.php
```diff
@@ -2,10 +2,12 @@
 
 namespace DDTrace;
 
+use DDTrace\Contracts\Span as SpanInterface;
```
labbati: Not required any change, but I noticed that we do the `use DDTrace\Contracts\Span as SpanInterface;` quite often.
Given that we are probably going to use the interfaces more than concrete (hopefully 😄) what do you think about adopting the following convention for the future?

```
// If only contract used
use DDTrace\Contracts\Span;

// If only concrete used
use DDTrace\Span;

// If both used
use DDTrace\Contracts\Span;
use DDTrace\Span as DDSpan;
```
?

src/DDTrace/GlobalTracer.php
```diff
@@ -0,0 +1,53 @@
...
+     * should call GlobalTracer::set as early as possible in bootstrap, prior to
+     * start a new span. Prior to calling GlobalTracer::set, any Spans started
+     * via the `Tracer::startActiveSpan` (etc) globals are noops.
+     *
+     * @param TracerInterface $tracer
+     */
+    public static function set(TracerInterface $tracer)
+    {
+        self::$instance = $tracer;
+        if (class_exists('\OpenTracing\GlobalTracer')) {
```
labbati: ❤️

src/DDTrace/GlobalTracer.php
```diff
@@ -0,0 +1,53 @@
...
+     * implementation that drops all data handed to it.
+     *
+     * @return TracerInterface
+     */
+    public static function get()
+    {
+        if (null !== self::$instance) {
+            return self::$instance;
+        }
+        return self::$instance = NoopTracer::create();
```
labbati: Here our NoopTracer and OT's NoopTracer are not in sync, but I don't see a way to "sync" them and it should not even be a problem at all.

src/DDTrace/ID.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace DDTrace;
+
+final class ID
```
labbati: ❤️

src/DDTrace/OpenTracer/Scope.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $this->scope->close();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSpan()
+    {
+        if (isset($this->span)) {
```
labbati: Do we need this optimization? Just to play safe I would always ask the scope for the current span, one less thing to keep in sync :)

src/DDTrace/OpenTracer/ScopeManager.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->scopeManager = $scopeManager;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function activate(OTSpan $span, $finishSpanOnClose = true)
+    {
+        return $this->scopeManager->activate(
+            Span::toDDSpan($span),
```
labbati: I may need some guidance here because I am not sure this would work! Typically this method will not be called by the user, it will be called by our `DDTrace\OpenTracer\Tracer::startActiveSpan()`. If this is true then we don't need to "copy" the Span as it is risky and will not be kept in sync (also looking at the actual implementation looks like we do not transfer tags?), because in that case the OTSpan that we receive would already be our wrapped span, so not need to copy into our own.

There is a corner case here, when users provide as with their own `OpenTracing\Span` implementation, which we should honor. I think that in that case what we should do is to wrap it into another wrapper which implements our interface, but still use it as storage, as this way we don't have to keep it in sync.

WDYT?

src/DDTrace/OpenTracer/SpanContext.php
```diff
@@ -0,0 +1,78 @@
...
+        return $this->context;
+    }
+
+    /**
+     * Converts an OpenTracing SpanContext instance to a DD one
+     *
+     * @param OTSpanContext $otContext
+     * @return DDSpanContext
+     */
+    public static function toDDSpanContext(OTSpanContext $otContext)
```
labbati: As I mentioned above, there may be cases when this "copying" vs "wrapping" may bite us back. Let's just make some more reasoning about it.

src/DDTrace/OpenTracer/Tracer.php
```diff
@@ -0,0 +1,119 @@
...
+            $this->tracer->getScopeManager()
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getActiveSpan()
+    {
+        return $this->tracer->getActiveSpan();
```
labbati: Shouldn't we be wrapping the returned object (it is a scope if I am correct).

src/DDTrace/OpenTracer/Scope.php
SammyK: My thought on this was that every time you call `getSpan()` here, you'll always get the same instance of the wrapper (just like the unwrapped version). Without this check, the internal call to `$this->scope->getSpan()` will always be the same instance, but the wrapper instance that is returned will always be different.

It's not a big deal either way and I can totally remove the check if you like the other way better. :)

src/DDTrace/OpenTracer/ScopeManager.php
SammyK: You're totally right - this won't work. Lol. Nice catch! There's another place or two where I do this with other OT objects. This was part of one of my failed attempts to wrap everything. I'm going to change this to check if the object is an instance of our OT implementation like I did [in the tracer](https://github.com/DataDog/dd-trace-php/pull/193/files#diff-e20592803012a46049e8d5df3097b7f7R65) and if so we can unwrap it. Otherwise, we'll fall back to another wrapper like you suggested. :)

src/DDTrace/OpenTracer/Tracer.php
SammyK: Nice catch! I figured there were one or two other unwrapped versions of things that I missed. :)

composer.json
```diff
@@ -22,15 +22,12 @@
...
-        "opentracing/opentracing": "1.0.0-beta5@dev"
-    },
-    "provide": {
-        "opentracing/opentracing": "1.0.0-beta5"
+        "ext-json": "*"
     },
     "require-dev": {
         "g1a/composer-test-scenarios": "~3.0",
         "mockery/mockery": "*",
+        "opentracing/opentracing": "1.0.0-beta5",
```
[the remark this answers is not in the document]
SammyK: I love this idea. Can we tackle this in a separate PR? :)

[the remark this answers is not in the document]
labbati: Sure of course!

src/DDTrace/OpenTracer/Scope.php
labbati: This actually makes a lot of sense. Thanks for clarifying!

src/DDTrace/OpenTracer/Tracer.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @param TracerInterface|null $tracer
+     */
+    public function __construct(TracerInterface $tracer = null)
+    {
+        $this->tracer = $tracer ?: self::make();
+    }
+
+    public static function make(Transport $transport = null, array $propagators = null, array $config = [])
```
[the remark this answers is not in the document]
labbati: Oh no no! Perfect I was just curious, please leave like here, I like it.

SammyK:
Thanks again @labbati for all the help on this massive PR! 😄 

All the tests have been added and the docs have been updated. There are still a few outstanding issues that I'd love to tackle in separate PR's if possible:

- **Too many interface methods:** I'm not 100% happy with all the methods I added to the interfaces that live under `DDTrace\Contracts`. It just smells funny to me, but if you all are OK with it, I'm OK with it. But if we feel like breaking that up a bit more, let's discuss. :)
- **OpenTracing-implementation edge case:** There is an edge case that @labbati brought up when a user injects their own OT instances into our OT instances (namely via `Tracer::inject()`, the `child_of` item in `Tracer::extract()`, and `ScopeManager::activate()`). Right now we inject the unwrapped version if our OT implementation is detected. Otherwise it tries to recreate a new instance of the object, but since we don't have access to critical data via the OT interfaces (like trace & span ID's), this potentially breaks the functionality. We should discuss a wrapping strategy to better handle this edge case. :)
- **Dev dependency:** As @labbati pointed out, we should remove the `opentracing/opentracing` dependency from `require-dev` and put it into `scenarios` to be consistent with other 3rd-party-dependent tests. Although I'm also OK with keeping it as is to insure every time we run `composer test`, it also runs the OT tests. I kind of like that OT is still "tightly coupled" to the dev process as part of the core. But I'm also open to moving it. :)

Other than that - I think this can be wrapped up! 😃

PR #203, opened 2018-12-21
Allow testing of multiple library versions

labbati:
### Description

When we integrate with different versions of the same library we need a structured way to define the various test scenarios for the several versions. @inverse mentioned this excellent library from @greg-1-anderson that seems to be the perfect tool for us: https://github.com/g1a/composer-test-scenarios
This PR aims at integrating this tool into our code base!

### Readiness checklist


CONTRIBUTING.md
```diff
@@ -74,10 +74,28 @@ $ composer test
...
+$ composer test -- --testsuite=integration
+```
+
+Testing integrations with libraries requires an additional step, as there are different scenarios where you want to test
+a specific integration. You can find available scenarios in `composer.json` at key `extras.scenarios`.
+
+As an example, in order to run Guzzle tests with Guzzle v5 library, run:
+
+    # Only the first time, to create all the different scenarios
+    $ composer scenario:update
```
SammyK: After running this command, the `LICENSE` file gets updated. Here's the diff:

```diff
diff --git a/LICENSE b/LICENSE
index be5bf07..1e187aa 100644
--- a/LICENSE
+++ b/LICENSE
@@ -27,3 +27,16 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+DEPENDENCY LICENSES:
+
+Name                       Version      License
+ircmaxell/password-compat  v1.0.4       MIT
+opentracing/opentracing    1.0.0-beta5  MIT
+paragonie/random_compat    v9.99.99     MIT
+psr/http-message           1.0.1        MIT
+psr/log                    1.1.0        MIT
+symfony/polyfill-php55     v1.10.0      MIT
+symfony/polyfill-php56     v1.10.0      MIT
+symfony/polyfill-php70     v1.10.0      MIT
+symfony/polyfill-util      v1.10.0      MIT
```

labbati: Thanks for noticing it, let me look at it!

labbati:
@SammyK 

> My biggest question is around running the entire test suite. Right now this PR changes the behavior of $ composer test which now runs only a subset of the tests:

The biggest problem here is that while circleci is PHP version aware, while composer is not. Some tests, e.g. `Mongo`, can be executed only on 5.6, while others can only be executed on 7.x, for more recent versions of the libraries. From a "stability" point of view this is not an issue (because of course CI has the last word) but still you are absolutely right about the fact that we should have a way to run the entire suite locally. Let me see how things are when moving all the definitions to composer. If we decide that it is not clear enough we can always roll back!

labbati:
@SammyK I tried to move all the various env specific definitions from circleci's config to composer. I think it makes a lot of sense. What do you think about it?

.gitignore
```diff
@@ -1,5 +1,6 @@
 **/vendor/
 composer.lock
+.scenarios.lock/
```
greg-1-anderson: If you want to use composer-test-scenarios to do highest/lowest testing (I recommend this, even for libraries), then you should commit composer.lock and .scenarios.lock. If you only want to do highest testing, then what you're doing here is fine.

composer.json
```diff
@@ -92,5 +171,27 @@
...
+                }
+            },
+            "predis1": {
+                "require": {
+                    "predis/predis": "^1.1"
+                }
+            }
+        },
+        "scenario-options": {
+            "dependency-licenses": false
```
greg-1-anderson: Maybe this should be the default. I like keeping my dependency licenses listed and up-to-date in the LICENSE file, but this feature is orthogonal to the primary purpose of composer-test-scenarios.

labbati: > Maybe this should be the default.

It probably would make sense, as you suggested this is orthogonal to the purpose of this library. Very nice feature though, and the fact that it is on/off configurable denoted a very forward-thinking approach!

composer.json
```diff
@@ -80,9 +78,90 @@
...
+            "@test-integration",
+            "@test-integrations-72"
+        ],
         "test-integration": [
             "echo \"Integration tests require the agent up and running. Use composer run-agent.\"",
-            "phpunit --colors=always --configuration=phpunit.xml --testsuite=integration"
+            "@composer test -- --testsuite=integration"
+        ],
+        "test-integrations-56": [
+            "@composer scenario:update",
```
inverse: Are these lists of commands going to be the same across each integration?

labbati: Right now only mongo tests are different, I expect that as we add more and more versions (soon) they will diverge quite a bit.

inverse: :+1:

SammyK:
Great work @labbati! :) I really like how I can run all the tests with one command.

One thing that I think we should consider going forward is how long it takes to run all the tests now. Considering that the tests suite takes several minutes to run and that time will certainly increase with every integration we add, this will eventually become a big issue for development going forward. I'm not sure of an obvious solution right now but I think we should definitely be thinking of some ways to speed this up with caching or something. :)

labbati:
Ops sorry @SammyK I had missed this comment before merging.
I absolutely agree with you. I don't have a precise idea at the moment. If I remember correctly we used https://github.com/paratestphp/paratest with my previous team to run tests in parallel, but be ready to some refactoring because, as an example, tests that uses the same mysql server may be overlapping and cause each other failure, and so on.
But we probably will need to find a solution. It is also true that as we move forward, work will probably shift (in quantity) from core to specific integrations, and for them we have specific command to run test individually. But let's start thinking about it.

chuck:
Setting up individual pipelines for integrations in circleci config is also an option - doesn't the python integration do something like that?

labbati:
@chuck this is absolutely an option. Actually, I thought that the comment from @SammyK was related the local env. In the end I never felt that having "builds" in CI lasting for a few minutes were a big problem. But in that case pipelines are surely a good approach! 👍

PR #206, opened 2019-01-02
Removal of external dependencies to support auto-instrumentation

labbati:
### Description

This PR removes external dependencies from the ddtrace package toward supporting a simpler usage of this library outside of composer. When providing auto-instrumentation, we are going to provide a bundled version of the source code and removing unnecessary dependencies will definitely simplify bundling.

### Readiness checklist


composer.json
```diff
@@ -23,12 +23,7 @@
         "php": "^5.6 || ^7.0",
         "ext-curl": "*",
         "ext-json": "*",
-        "opentracing/opentracing": "1.0.0-beta5@dev",
-        "psr/http-message": "^1.0",
-        "psr/log": "^1.0.0",
-        "symfony/polyfill-php55": "^1.9",
-        "symfony/polyfill-php56": "^1.9",
-        "symfony/polyfill-php70": "^1.8.0"
+        "opentracing/opentracing": "1.0.0-beta5@dev"
```
pawelchcki: https://github.com/DataDog/dd-trace-php/pull/206/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L40-L43 Should probably also be removed

labbati:
Hey @SammyK , thanks for the review!

> Since this PR removes that interface, I would now have to write a logger wrapper just so I can inject it into the tracer which could be a give-up point for people who are trying to troubleshoot when the tracer fails silently. Just my 2 cents! :)

Wouldn't the provided wrapper work? https://github.com/DataDog/dd-trace-php/pull/206/files#diff-245d8af1436aa36db09e3a87bac7fc15R1
In my original intent it was to be used in such cases.

SammyK:
Hey @labbati!

>Wouldn't the provided wrapper work?

Yes it would! :D But I guess I'm focused the reasoning behind removing `psr/log` from the `composer.json` but still reference the `\Psr\Log\LoggerInterface`. :)

composer.json
labbati: 👍 removed, thanks

labbati:
Re-opening after rebasing

labbati:
Thanks @SammyK . And BTW once we merge OT refactoring this can be merged right away 😄 so it won't be too far, be patient ahahah

PR #210, opened 2019-01-04
Add PECL config

SammyK:
### Description

This PR adds the necessary metadata to build a package for [datadog_trace on PECL](https://pecl.php.net/package/datadog_trace).

To build, make sure any new files are added to the `package.xml` and validate the XML and then package it:

```bash
$ pear package-validate package.xml && pear package
```

Since ddtrace is still in beta, you can install it from PECL with the following:

```bash
$ sudo pecl install datadog_trace-beta
```

Once GA hits, you'll be able to just do:

```bash
$ sudo pecl install datadog_trace
```

### Readiness checklist

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -25,7 +25,51 @@
     </notes>
     <contents>
         <dir name="/">
+            <dir name="src">
+                <dir name="ext">
+                    <file name="compat_zend_string.c" role="src" />
+                    <file name="compat_zend_string.h" role="src" />
+                    <file name="ddtrace.c" role="src" />
+                    <file name="ddtrace.h" role="src" />
+                    <file name="debug.h" role="src" />
+                    <file name="dispatch.c" role="src" />
+                    <file name="dispatch.h" role="src" />
+                    <file name="dispatch_compat.h" role="src" />
+                    <file name="dispatch_compat_php5.c" role="src" />
+                    <file name="dispatch_compat_php5.h" role="src" />
+                    <file name="dispatch_compat_php7.c" role="src" />
+                    <file name="dispatch_setup.c" role="src" />
+                    <file name="version.h" role="src" />
+                </dir>
+            </dir>
+            <dir name="tests">
+                <dir name="ext">
+                    <file name="access_modifier_method_access_hook.phpt" role="test" />
+                    <file name="access_modifier_property_access_hook.phpt" role="test" />
+                    <file name="case_insensitive_method_hook.phpt" role="test" />
+                    <file name="closure_accessing_outside_variables.phpt" role="test" />
+                    <file name="closure_set_inside_object_methods.phpt" role="test" />
+                    <file name="method_returning_array.phpt" role="test" />
+                    <file name="multiple_instrumentations.phpt" role="test" />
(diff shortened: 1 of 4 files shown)


package.xml
```diff
@@ -0,0 +1,93 @@
...
+    <version>
+        <release>0.8.1</release>
+        <api>0.8.1</api>
+    </version>
+    <stability>
+        <release>beta</release>
+        <api>beta</api>
+    </stability>
+    <license uri="https://github.com/DataDog/dd-trace-php/blob/master/LICENSE">BSD 3-Clause</license>
+    <notes>
```
labbati: I think those were added here unintentionally?

package.xml
```diff
@@ -0,0 +1,93 @@
...
+                    <file name="dispatch.h" role="src" />
+                    <file name="dispatch_compat.h" role="src" />
+                    <file name="dispatch_compat_php5.c" role="src" />
+                    <file name="dispatch_compat_php5.h" role="src" />
+                    <file name="dispatch_compat_php7.c" role="src" />
+                    <file name="dispatch_setup.c" role="src" />
+                    <file name="version.h" role="src" />
+                </dir>
+            </dir>
+            <dir name="tests">
```
labbati: Is it a practice to include tests in pear packages? (asking) 😄

labbati: Yep apparently it is: https://github.com/phpredis/phpredis/blob/develop/package.xml#L113

package.xml
```diff
@@ -0,0 +1,93 @@
...
+            </php>
+            <pearinstaller>
+                <min>1.4.0</min>
+            </pearinstaller>
+            <extension>
+                <name>json</name>
+            </extension>
+        </required>
+    </dependencies>
+    <providesextension>ddtrace</providesextension>
```
labbati: I checked a couple of extensions and seems that typically `<name/>` and `<providesextension/>` match. I am wondering what are the cons/issues we face if we provide the name `datadog_trace` also here.

package.xml
remicollet: Definitively YES.
Test are useful (at least) for downstream packager

package.xml
```diff
@@ -0,0 +1,93 @@
...
+            <file name="CHANGELOG.md" role="doc" />
+            <file name="config.m4" role="src" />
+            <file name="LICENSE" role="doc" />
+            <file name="README.md" role="doc" />
+        </dir>
+    </contents>
+    <dependencies>
+        <required>
+            <php>
+                <min>5.6</min>
```
remicollet: can also set `<max>7.2.99</min>` as you known 7.3 is not supported

package.xml
remicollet: some other extension have different name, this is not a blocker (e.g. pecl_http / http, apcu_bc / apc, ...)

package.xml
SammyK: Thanks! I just updated that until we can get PHP 7.3 support rocking. :)

package.xml
labbati: Thanks @remicollet

PR #211, opened 2019-01-04
Track of integration loadings to avoid re-loading unnecessary ones

labbati:
### Readiness checklist

### DIFF
--- a/src/DDTrace/Integrations/Integration.php
+++ b/src/DDTrace/Integrations/Integration.php
@@ -8,15 +8,22 @@ use OpenTracing\GlobalTracer;
 
 abstract class Integration
 {
+    // Possible statuses for the concrete
+    const NOT_LOADED = 0;
+    const LOADED = 1;
+    const NOT_AVAILABLE = 2;
+
     const CLASS_NAME = '';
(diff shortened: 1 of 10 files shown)


src/DDTrace/Integrations/Integration.php
```diff
@@ -8,15 +8,25 @@
...
+    const LOADED = 1;
+    const NOT_AVAILABLE = 2;
+
     const CLASS_NAME = '';
 
     public static function load()
     {
         if (!class_exists(static::CLASS_NAME)) {
-            return;
+            return Integration::NOT_LOADED;
```
SammyK: Just a little tweak :)

```suggestion
            return self::NOT_LOADED;
```

src/DDTrace/Integrations/Integration.php
```diff
@@ -8,15 +8,25 @@
...
         }
         // See comment on the commented out abstract function definition.
         static::loadIntegration();
+
+        return Integration::LOADED;
```
SammyK: Here too :)

```suggestion
        return self::LOADED;
```

Issue #212, opened 2019-01-04
PHP 7.3 compatibility


remicollet:
Build fails.

```
In file included from /opt/remi/php73/root/usr/include/php/Zend/zend.h:29,
                 from /opt/remi/php73/root/usr/include/php/main/php.h:33,
                 from /dev/shm/BUILD/php73-php-pecl-datadog-trace-0.8.1/NTS/src/ext/dispatch_compat_php7.c:1:
/dev/shm/BUILD/php73-php-pecl-datadog-trace-0.8.1/NTS/src/ext/dispatch_compat_php7.c: In function 'ddtrace_dispatch_store':
/dev/shm/BUILD/php73-php-pecl-datadog-trace-0.8.1/NTS/src/ext/dispatch_compat_php7.c:46:91: error: 'HASH_FLAG_PERSISTENT' undeclared (first use in this function); did you mean 'HASH_FLAG_CONSISTENCY'?
     ddtrace_dispatch_t *dispatch = pemalloc(sizeof(ddtrace_dispatch_t), lookup->u.flags & HASH_FLAG_PERSISTENT);
                                                                                           ^~~~~~~~~~~~~~~~~~~~
/opt/remi/php73/root/usr/include/php/Zend/zend_alloc.h:193:38: note: in definition of macro 'pemalloc'
 #define pemalloc(size, persistent) ((persistent)?__zend_malloc(size):emalloc(size))
                                      ^~~~~~~~~~
/dev/shm/BUILD/php73-php-pecl-datadog-trace-0.8.1/NTS/src/ext/dispatch_compat_php7.c:46:91: note: each undeclared identifier is reported only once for each function it appears in
     ddtrace_dispatch_t *dispatch = pemalloc(sizeof(ddtrace_dispatch_t), lookup->u.flags & HASH_FLAG_PERSISTENT);
                                                                                           ^~~~~~~~~~~~~~~~~~~~
/opt/remi/php73/root/usr/include/php/Zend/zend_alloc.h:193:38: note: in definition of macro 'pemalloc'
 #define pemalloc(size, persistent) ((persistent)?__zend_malloc(size):emalloc(size))

```

SammyK:
Thanks for reporting this @remicollet! We're working on adding 7.3 support. :)

pawelchcki:
7.3 is compiling successfully 🎉 thanks @SammyK for fix  and @remicollet for reporting

PR #215, opened 2019-01-04
Fix compile-time error on PHP 7.3

SammyK:
### Description

In PHP 7.3 there was a [small BC](https://github.com/php/php-src/blob/1c850bfcca97e0456a716a1889b5cf6c5e477cd8/UPGRADING.INTERNALS#L77-L78) that affected ddtrace which was pointed out in #212. This PR doesn't officially offer PHP 7.3 support (we still need to add the proper Docker containers, etc), but ddtrace will at least compile for 7.3 now. I didn't want to make too many changes that would cause more work for @pawelchcki's open work. :)

### Readiness checklist


remicollet:
+1

PR #218, opened 2019-01-07
fix for 7.3

remicollet:
From UPGRADING.INTERNALS

>   d. HASH_FLAG_PERSISTENT renamed into IS_ARRAY_PERSISTENT and moved into
>      GC_FLAGS (to be consistent with IS_STR_PERSISTENT).
>


remicollet:
looks like I miss PR #215

SammyK:
Thanks for the PR anyway! Glad to see we had an almost identical patch! :D

PR #221, opened 2019-01-07
Bump version to 0.9.0

SammyK:
### Description

Version bump for `0.9.0` release.

### Readiness checklist
- ~[ ] Tests added for this feature/bug~

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -10,9 +10,9 @@
         <email>sammyk@php.net</email>
         <active>yes</active>
     </lead>
-    <date>2018-12-14</date>
+    <date>2019-01-07</date>
     <version>
-        <release>0.8.1</release>
+        <release>0.9.0</release>
         <api>0.8.1</api>
     </version>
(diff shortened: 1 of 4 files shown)


package.xml
```diff
@@ -10,9 +10,9 @@
         <email>sammyk@php.net</email>
         <active>yes</active>
     </lead>
-    <date>2018-12-14</date>
+    <date>2019-01-07</date>
     <version>
-        <release>0.8.1</release>
+        <release>0.9.0</release>
         <api>0.8.1</api>
```
labbati: Shouldn't we also update the API version?

SammyK: Yeah, I didn't change this one since none of the existing public API's changed but we did at least add a new public method or two. So I'll go ahead and bump that as well. But looking at other extensions in PECL, they only up the `<api>` tag if the API has changed. I think it technically did in our case do I'll update it. :)

PR #222, opened 2019-01-07
Add `dd_trace_reset` function, and stop throwing exceptions when overriding missing methods 

pawelchcki:
### Description

- `dd_trace_reset` simply resets the lookup hashes to initial state, allowing removing of existing overrides - mostly usable for tests

- Exception throwing behavior is controlled by `ddtrace.ignore_missing_overridables` ini setting that defaults to `true`

### Readiness checklist

### DIFF
--- a/src/ext/ddtrace.h
+++ b/src/ext/ddtrace.h
@@ -6,7 +6,6 @@ extern zend_module_entry ddtrace_module_entry;
 ZEND_BEGIN_MODULE_GLOBALS(ddtrace)
 zend_bool disable;
 zend_bool ignore_missing_overridables;
-zend_bool lookup;
 HashTable class_lookup;
 HashTable function_lookup;
 ZEND_END_MODULE_GLOBALS(ddtrace)
--- a/src/ext/ddtrace.h
+++ b/src/ext/ddtrace.h
@@ -5,6 +5,8 @@ extern zend_module_entry ddtrace_module_entry;
 
 ZEND_BEGIN_MODULE_GLOBALS(ddtrace)
 zend_bool disable;
+zend_bool ignore_missing_overridables;
+zend_bool lookup;
 HashTable class_lookup;
 HashTable function_lookup;
 ZEND_END_MODULE_GLOBALS(ddtrace)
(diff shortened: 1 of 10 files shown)


src/ext/ddtrace.h
```diff
@@ -5,6 +5,8 @@ extern zend_module_entry ddtrace_module_entry;
 
 ZEND_BEGIN_MODULE_GLOBALS(ddtrace)
 zend_bool disable;
+zend_bool ignore_missing_overridables;
+zend_bool lookup;
```
labbati: How is this global variable `lookup` used? I only see usages of `class_lookup ` and `function_lookup `.

tests/ext/reset_configured_overrides.phpt
```diff
@@ -0,0 +1,55 @@
+--TEST--
```
labbati: Do you think that makes sense to add a test to make sure that the exception is not thrown? Something like

```
--TEST--
Check tracing of non existing classes, methods and functions.
--FILE--
<?php

class ClassWithNoMethods
{
}

dd_trace("ClassNotExist", "method" ,function(){});
dd_trace("ClassWithNoMethods", "method", function(){});
dd_trace("function_not_exist", function(){});

?>
--EXPECT--

```

src/ext/ddtrace.h
pawelchcki: Whoops, that wasn't supposed to go into this PR.

tests/ext/reset_configured_overrides.phpt
pawelchcki: Ops I missed two files when submitting this PR. I'll do better next time. One of those files checked exactly what, you've been proposing.

tests/ext/silently_check_if_overridable_exists.phpt
```diff
@@ -1,13 +1,38 @@
 --TEST--
-Toggle checking if overridable method/function exists or not
+Do not throw exceptions when veryfying if class/method and function existst.
```
labbati: existststststst 😄

PR #225, opened 2019-01-09
One-liner API to enable tracing

SammyK:
### Description

This is a WIP PR to start a discussion around what an API might look like to enable the tracer with one line. I'm using a "documentation first" approach to make sure we make a solid API before going for the implementation. Essentially my proposal to kick things off is to add this line to early script execution:

```php
\DDTrace\Tracer::init('My Application');
```

I've documented the features in more detail in the PR. Feel free to post any thoughts! :)

### Readiness checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,35 +82,21 @@ Once the `ddtrace` extension and Composer package is installed, you can start tr
 
 ### Manual instrumentation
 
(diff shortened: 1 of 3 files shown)


SammyK:
Oops! I thought I closed this already. Since auto-instrumentation pretty much covers what this PR was all about, I'm going to close this. :)

SammyK:
I will say that we do need a way to inject a config array into the tracer so maybe we can discuss possible options for that. :)

PR #227, opened 2019-01-10
add TryCatchFinally::executeAnyMethod to avoid not being able to call protected methods

pawelchcki:
### Readiness checklist

### DIFF
--- a/src/DDTrace/Util/TryCatchFinally.php
+++ b/src/DDTrace/Util/TryCatchFinally.php
@@ -62,9 +62,12 @@ class TryCatchFinally
         $result = null;
         /** @var SpanInterface $span */
         $span = $scope->getSpan();
-        $instance_call_user_func_array = (function ($fn, $args) {
+        $call_user_func_array = function ($fn, $args) {
             return call_user_func_array($fn, $args);
-        })->bindTo($instance, $callingScope);
+        };
+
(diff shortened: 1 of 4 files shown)


src/DDTrace/Util/TryCatchFinally.php
```diff
@@ -45,6 +45,47 @@ public static function executePublicMethod(Scope $scope, $instance, $method, arr
...
+        /** @var SpanInterface $span */
+        $span = $scope->getSpan();
+        $call_user_func_array = function ($fn, $args) {
+            return call_user_func_array($fn, $args);
+        };
+
+        $instance_call_user_func_array->bindTo($instance, $callingScope);
+
+        try {
+            $result = $instance_call_user_func_array([$instance, $method], $args);
```
SammyK: The only way I could get this to work was by doing something like this:
```suggestion
            $result = call_user_func(\Closure::bind(
                function () use ($instance, $method, $args) {
                    return call_user_func_array([$instance, $method], $args);
                },
                null,
                $instance
            ));
```
Then we can kill the callback and the `bindTo()` lines above. We can also remove the `$callingScope` param. :)

pawelchcki: Thanks - Good call! :)

PR #235, opened 2019-01-15
Improve request replayer returning traces

labbati:
### Readiness checklist

### DIFF
--- a/tests/request_replayer/index.php
+++ b/tests/request_replayer/index.php
@@ -1,7 +1,7 @@
 <?php
 
-$defaultRequestLog = '/tmp/php_request_replayer_' . getmypid() . '.json';
-$requestLog = getenv('DD_REQUEST_DUMPER_FILE') ? : $defaultRequestLog;
+$filename = basename(getenv('DD_REQUEST_DUMPER_FILE') ?: 'php_request_replayer_' . getmypid() . '.json');
+$requestLog = sys_get_temp_dir() . '/' . $filename;
 
 if (php_sapi_name() == 'cli-server') {
     if ($_SERVER['REQUEST_URI'] == '/replay') {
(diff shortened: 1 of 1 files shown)


tests/request_replayer/index.php
```diff
@@ -1,15 +1,21 @@
 <?php
 
-$requestLog = '/tmp/php_request_replayer_' . getmypid() . '.json';
+$defaultRequestLog = '/tmp/php_request_replayer_' . getmypid() . '.json';
```
SammyK: This will work on systems with temp dirs in other locations. :)
```suggestion
$requestLog = tempnam(sys_get_temp_dir(), 'php_request_replayer_');
```

tests/request_replayer/index.php
```diff
@@ -1,15 +1,21 @@
...
+$requestLog = getenv('DD_REQUEST_DUMPER_FILE') ? : $defaultRequestLog;
```
SammyK: And then we can change this such that `DD_REQUEST_DUMPER_FILE` is just the file name jailed to the system temp dir. :)
```suggestion
if (getenv('DD_REQUEST_DUMPER_FILE')) {
    $requestLog = sys_get_temp_dir() . '/' . basename(getenv('DD_REQUEST_DUMPER_FILE'));
}
```

tests/request_replayer/index.php
SammyK: Actually, we don't want to create the file here if we don't need to (which `tempnam()` does) so this is better. :)
```suggestion
$requestLog = sys_get_temp_dir() . '/php_request_replayer_' . getmypid() . '.json';
```

labbati:
> Looks good! Just a tweak to make it a bit more secure and cross-platform. :)

That was a good call, thanks for catching that. Please look if the changes I made address the issue!

PR #241, opened 2019-01-17
Add root span/scope tracking

SammyK:
### Description

This PR adds a way to track the root span so that it can be modified later in the script execution. This is important for integrations that don't have a way of hooking into the request/routing/etc early on in the script execution when the tracer is inited.

It works just like `Tracer::startActiveSpan()`, but it saves the `Scope` instance on the tracer to be accessed later by `Tracer::getRootScope()`.

```php
$tracer->startRootSpan('my_root_span');
// ...do lots of stuff...
$scope = $tracer->getRootScope();
// ..add tags and whatnot
```

### Readiness checklist

### DIFF
--- a/src/DDTrace/Contracts/Tracer.php
+++ b/src/DDTrace/Contracts/Tracer.php
@@ -129,7 +129,7 @@ interface Tracer
     public function getPrioritySampling();
 
     /**
-     * @return Scope
+     * @return Scope|null
      */
     public function getRootSpan();
 }
--- a/src/DDTrace/Contracts/Tracer.php
+++ b/src/DDTrace/Contracts/Tracer.php
@@ -127,4 +127,9 @@ interface Tracer
      * @return mixed
      */
     public function getPrioritySampling();
+
+    /**
+     * @return Scope
+     */
+    public function getRootSpan();
 }
--- a/src/DDTrace/Contracts/Tracer.php
+++ b/src/DDTrace/Contracts/Tracer.php
@@ -129,11 +129,14 @@ interface Tracer
     public function getPrioritySampling();
 
     /**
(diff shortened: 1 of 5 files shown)


src/DDTrace/Contracts/Tracer.php
```diff
@@ -127,4 +127,9 @@ public function flush();
      * @return mixed
      */
     public function getPrioritySampling();
+
+    /**
+     * @return Scope|null
+     */
+    public function getRootSpan();
```
labbati: Is  there a  reason why  we don't want to add `startRootSpan()` method to the contract?  :)

src/DDTrace/Tracer.php
```diff
@@ -63,6 +64,11 @@
      */
     private $scopeManager;
 
+    /**
+     * @var ScopeInterface
+     */
+    private $rootSpan;
```
labbati: Being it a scope, shouldn't it be named `$rootScope`, and so the getter name also should change.
What  do you think?  :)

tests/Unit/TracerTest.php
```diff
@@ -195,4 +195,17 @@ public function testUnfinishedSpansCanBeFinishedOnFlush()
         $this->assertSame('root', $sent[0][0]->getOperationName());
         $this->assertSame('child', $sent[0][1]->getOperationName());
     }
+
+    public function testSpanStartedAtRootCanBeAccessedLater()
+    {
+        $tracer = new Tracer(new NoopTransport());
+        $span = $tracer->startRootSpan(self::OPERATION_NAME);
```
labbati: If you approve the suggestions above, then this should be `$scope`. I think that  we should refactor that  part returning the span that let you access the scope, but this is very in the future

PR #260, opened 2019-01-24
Configurable sampler

labbati:
### Description

This PR will allow user to configure the sampling rate through an environmental variable `DD_SAMPLING_RATE` which can be set as a floating point number between `0.0` (always reject) and `1.0` (always keep).

### Readiness checklist

Special kudos to @SammyK in the PR header to have spotted the 32/64 bits issue 😄


src/DDTrace/Sampling/ConfigurableSampler.php
```diff
@@ -0,0 +1,34 @@
...
+        if ($rate === 1.0) {
+            return PrioritySampling::AUTO_KEEP;
+        } elseif ($rate === 0.0) {
+            return PrioritySampling::AUTO_REJECT;
+        }
+
+        // NOTE: we do not apply the knuth hashing algorithm here and this is fine for the sole purpose of setting the
+        // sampling priority. This would not work though if we were using this value for client sampling, as the value
+        // and statistical distribution should exactly match the one used by the agent.
+        $shouldKeep = (int) $span->getSpanId() <= $rate * PHP_INT_MAX;
```
SammyK: It just occurred to me that for the trace and span ID generation, we're assuming [`mt_rand()` can spit out 64-bit ints](https://github.com/DataDog/dd-trace-php/blob/5bb3432483280d4e4e1354b0b6987d00f9484dbb/src/DDTrace/ID.php#L17), when in fact [it's maxed out at 32-bit](https://github.com/php/php-src/blob/67e0138c0dfd966624223911a0821f6c294ad1c6/ext/standard/php_mt_rand.h#L30). And that would skew this rate calculation. For now I'm guessing we need to do something like this:
```suggestion
        $shouldKeep = (int) $span->getSpanId() <= $rate * mt_getrandmax();
```
And we should probably update `DDTrace\ID::generate()` to use `mt_getrandmax()` instead of `PHP_INT_MAX` as well. If we wanted 64-bit ints, we could switch to `random_int()`, but it's slower since it's designed for use in crypto. Also it's only PHP 7+ so that might not be viable. Or do you have another suggestion? :)

src/DDTrace/Sampling/ConfigurableSampler.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace DDTrace\Sampling;
+
+use DDTrace\Configuration;
+use DDTrace\Contracts\Span;
+
+/**
+ * A sampler configurable using global a global configuration parameter.
+ */
+class ConfigurableSampler implements Sampler
```
SammyK: Just to lock things down a bit. :)
```suggestion
final class ConfigurableSampler implements Sampler
```

src/DDTrace/Tracer.php
```diff
@@ -101,7 +101,7 @@ public function reset()
     {
         $this->scopeManager = new ScopeManager();
         $this->globalConfig = Configuration::get();
-        $this->sampler = new AlwaysKeepSampler();
+        $this->sampler = new ConfigurableSampler();
```
SammyK: ❤️

src/DDTrace/Sampling/ConfigurableSampler.php
labbati: Very good catch! For the current purpose I think that the change you suggested makes sense. as alternative suggestions, looking around seems that combining 2 32 bits random numbers could serve our usage. Let's investigate this in a different PR!

labbati:
@SammyK up again for second round of CR

src/DDTrace/ID.php
```diff
@@ -14,6 +19,19 @@ public static function generate()
          * will do. And since all integers in PHP are signed, an int
          * between 1 & PHP_INT_MAX will be 63-bit.
          */
-        return (string) mt_rand(1, PHP_INT_MAX);
+        $first31bits = mt_rand(0, mt_getrandmax()) << 32;
+        $second31bits = mt_rand(0, mt_getrandmax()) << 1;
+        $random1or0 = mt_rand(0, 1);
+        return (string) ($first31bits | $second31bits | $random1or0);
```
SammyK: Nice workaround @labbati! For the future PR, we'll look into possibly supporting 32-bit architectures as well. :)

src/DDTrace/ID.php
```diff
@@ -14,6 +19,19 @@ public static function generate()
...
+    }
+
+    /**
+     * The expected max value for the generated id.
+     *
+     * @return int
+     */
+    public static function getMaxId()
+    {
+        return (mt_getrandmax() << 32) | (mt_getrandmax() << 1) | 1;
```
SammyK: I'm curious why this exists. Is it just to show that the return value is the same as `PHP_INT_MAX` in the tests? I don't see it being used anywhere else in the tracer and that test might break on 32-bit boxes. Just double checking. :)

src/DDTrace/Sampling/ConfigurableSampler.php
```diff
@@ -24,10 +24,16 @@ public function getPrioritySampling(Span $span)
...
         // sampling priority. This would not work though if we were using this value for client sampling, as the value
         // and statistical distribution should exactly match the one used by the agent.
-        $shouldKeep = (int) $span->getSpanId() <= $rate * mt_getrandmax();
+        //
+        // NOTE 2: as SammyK correctly pointed out, we use `mt_rand()` to generate IDs which is topped to 31 bits, not
+        // 63. In order to compensate for this we concatenate multiple generators which cause a major biasing of the
+        // algorithm. To compensate this we use here a `mt_rand` generate value to decide how to set the priority
+        // sampling. When we will introduce client sampling we will have to implement the real and final knuth hashing
+        // function.
+        $shouldKeep = mt_rand(1, mt_getrandmax()) <= $rate * mt_getrandmax();
```
SammyK: ❤️

PR #261, opened 2019-01-24
Debug mode for logging

labbati:
### Description

This PR add a new setting `DD_TRACE_DEBUG` which defaults to `false` and when it is turned on it dumps a number of useful info to the logger. For now we do not implement logging levels as it will come later on.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Log/InterpolateTrait.php
+++ b/src/DDTrace/Log/InterpolateTrait.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace DDTrace\Log;
+
+/**
+ * Provides methods to interpolate log messages.
+ */
+trait InterpolateTrait
+{
+    /**
+     * Interpolates context values into the message placeholders. Example code from:
+     * https://www.php-fig.org/psr/psr-3/
+     *
+     * @param string $message
+     * @param array $context
+     * @return string
+     */
+    public function interpolate($message, array $context = [])
+    {
+        // build a replacement array with braces around the context keys
+        $replace = array();
+        foreach ($context as $key => $val) {
+            // check that the value can be casted to string
+            if (!is_array($val) && (!is_object($val) || method_exists($val, '__toString'))) {
+                $replace['{' . $key . '}'] = $val;
+            }
+        }
+
+        // interpolate replacement values into the message and return
+        return strtr($message, $replace);
+    }
+}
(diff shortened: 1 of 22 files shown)


src/DDTrace/Transport/Http.php
```diff
@@ -7,12 +7,15 @@
 use DDTrace\Encoder;
 use DDTrace\Log\Logger;
 use DDTrace\Log\LoggerInterface;
+use DDTrace\Log\LoggingTrait;
 use DDTrace\Sampling\PrioritySampling;
 use DDTrace\Transport;
 use DDTrace\GlobalTracer;
 
 final class Http implements Transport
```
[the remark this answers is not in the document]
labbati: good idea, let me do it (probably it make sense to directly using the LoggingTrait which does all the things) 👍 .

src/DDTrace/Log/InterpolateTrait.php
```diff
@@ -0,0 +1,32 @@
...
+        $replace = array();
+        foreach ($context as $key => $val) {
+            // check that the value can be casted to string
+            if (!is_array($val) && (!is_object($val) || method_exists($val, '__toString'))) {
+                $replace['{' . $key . '}'] = $val;
+            }
+        }
+
+        // interpolate replacement values into the message and return
+        return strtr($message, $replace);
```
SammyK: What are your thoughts on adding a `[ddtrace] ` prefix to all the logs to easily differentiate the entries from other messages? This would be similar to how the [Ruby integration](https://docs.datadoghq.com/tracing/advanced_usage/?tab=ruby#logging) does it. :)
```suggestion
        return '[ddtrace] ' . strtr($message, $replace);
```

labbati: Love it, but I would keep this out of interpolate function, will move to the message building part. Thanks 👍

src/DDTrace/Transport/Http.php
```diff
@@ -41,17 +39,11 @@ final class Http implements Transport
      */
     private $config;
 
-    /**
-     * @var LoggerInterface
-     */
-    private $logger;
-
-    public function __construct(Encoder $encoder, LoggerInterface $logger = null, array $config = [])
+    public function __construct(Encoder $encoder, array $config = [])
```
SammyK: ❤️

PR #262, opened 2019-01-25
Auto-instrumentation in Symfony 3.4 and PHP 5.6

labbati:
### Readiness checklist


bridge/dd_wrap_autoloader.php
```diff
@@ -12,16 +12,25 @@
 
 $dd_autoload_called = false;
```
SammyK: Since these variables get dumped into the global scope even with the namespacing, can we wrap everything from this line down with a self-calling function?

```php
call_user_func(function () {
    $dd_autoload_called = false;
    // ....
});
```

That will prevent these variables from showing up in user's apps to potentially cause unexpected behavior. :)

screen shot 2019-01-25 at 12 53 18 pm

bridge/functions.php
```diff
@@ -15,3 +15,20 @@ function dd_tracing_enabled()
...
     }
 }
+
+/**
+ * Checks if any of the provided classes exists.
+ *
+ * @param string[] $sentinelClasses
+ * @return bool
+ */
+function any_class_exists(array $sentinelClasses)
```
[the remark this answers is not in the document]
SammyK: Also if we add the namespaces, we'll need to update line 10 of **dd_init.php** to this:

```php
if (!\DDTrace\dd_tracing_enabled()) {
```

docker-compose.yml
```diff
@@ -110,3 +110,33 @@ services:
...
     command: sh -c "curl --output index.php https://raw.githubusercontent.com/DataDog/dd-trace-php/master/tests/request_replayer/index.php -q && php -S 0.0.0.0:80 index.php"
+
+  agent:
+    image: datadog/agent:latest
+    volumes:
+    - /var/run/docker.sock:/var/run/docker.sock:ro
+    - /proc/:/host/proc/:ro
+    - /sys/fs/cgroup/:/host/sys/fs/cgroup:ro
+    environment:
+    - DD_API_KEY=${DATADOG_API_KEY}
```
SammyK: Was this intentional to duplicate the agent container `ddagent_integration` here and pull the API key from the env vars instead of using the value `invalid_key_but_its_ok`? If so, cool, but wanted to double check. :)

local-build.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+rm -rf extensions
+mkdir -p extensions
+rm -rf build/packages
+mkdir -p build/packages
+
+function build_version() {
```
SammyK: ❤️

bridge/dd_wrap_autoloader.php
```diff
@@ -12,16 +12,25 @@
```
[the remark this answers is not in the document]
labbati: If it is fine for you, for consistency I would add `DDTrace\Bridge` as int he other files.

docker-compose.yml
labbati: This can be used together with `apache_56_server` container to send real traces to our backend. I can drop this no problem (or move to a different file). But it may be useful to have some real server sending traces to our backend.

bridge/dd_wrap_autoloader.php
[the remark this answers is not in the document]
SammyK: Even better! :)

docker-compose.yml
SammyK: Perfect! Just wanted to double check. :)

PR #264, opened 2019-01-25
Unsupported & custom framework traces

SammyK:
### Description

This PR ensures that unsupported and custom frameworks still get traced out of the box.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Configuration/EnvVariableRegistry.php
+++ b/src/DDTrace/Configuration/EnvVariableRegistry.php
@@ -80,8 +80,8 @@ class EnvVariableRegistry implements Registry
     public function floatValue($key, $default, $min = null, $max = null)
     {
         if (!isset($this->registry[$key])) {
-            $value = getenv($this->convertKeyToEnvVariableName($key));
-            $value = trim(strtolower($value));
+            $value = self::get($key);
+            $value = strtolower($value);
             if (is_numeric($value)) {
                 $floatValue = (float)$value;
             } else {
--- a/src/DDTrace/Configuration/EnvVariableRegistry.php
+++ b/src/DDTrace/Configuration/EnvVariableRegistry.php
@@ -19,20 +19,51 @@ class EnvVariableRegistry implements Registry
     }
 
     /**
+     * Return an env variable that starts with "DD_".
+     *
+     * @param string $key
+     * @return string|null
+     */
+    public static function get($key)
+    {
+        $value = getenv(self::convertKeyToEnvVariableName($key));
+        if (false === $value) {
+            return null;
+        }
+        return trim($value);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function stringValue($key, $default)
+    {
+        return isset($this->registry[$key])
+            ? $this->registry[$key]
+            : $this->registry[$key] = (string) self::get($key);
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function boolValue($key, $default)
     {
-        if (!isset($this->registry[$key])) {
-            $value = getenv($this->convertKeyToEnvVariableName($key));
-            $value = trim(strtolower($value));
-            if ($value === '1' || $value === 'true') {
-                $this->registry[$key] = true;
-            } elseif ($value === '0' || $value === 'false') {
-                $this->registry[$key] = false;
-            } else {
-                $this->registry[$key] = $default;
-            }
+        if (isset($this->registry[$key])) {
+            return $this->registry[$key];
+        }
+
+        $value = self::get($key);
+        if (null === $value) {
+            return $default;
+        }
+
+        $value = strtolower($value);
+        if ($value === '1' || $value === 'true') {
+            $this->registry[$key] = true;
+        } elseif ($value === '0' || $value === 'false') {
+            $this->registry[$key] = false;
+        } else {
+            return $default;
         }
 
         return $this->registry[$key];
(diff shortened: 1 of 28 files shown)


src/DDTrace/Integrations/Laravel/V4/LaravelProvider.php
```diff
@@ -45,30 +42,21 @@ public function register()
             return;
         }
 
-        $appName = $this->getAppName();
+        $appName = self::getAppName();
+        Bootstrap::tracerAndIntegrations();
```
labbati: Starting a conversation here as from a functional standpoint this is perfectly legit :)

I am not sure that we should invoke `Bootstrap::tracerAndIntegrations();` here. I believe we should assume that if we are here the tracer has already been bootstrapped, being it automatically by the auto-instrumentation or manually by the user calling `Bootstrap::tracerAndIntegrations();` at the very begin of the script.

Invoking this script has a cost as integrations not previously loaded are tried again.

WDYT? (Note here and in the providers/bundles below)

tests/Integrations/Custom/Autoloaded/CommonScenariosTest.php
```diff
@@ -0,0 +1,78 @@
...
+    public function provideSpecs()
+    {
+        return $this->buildDataProvider(
+            [
+                'A simple GET request returning a string' => [
+                    SpanAssertion::build(
+                        'web.request',
+                        'web.request',
+                        'web',
+                        'web.request'
```
labbati: Not required for this PR, but we should probably provide an hook to allow users to set proper resource names (depending on their routing mechanism). Otherwise they will see in the UI one and only one resource for each and every request :)

SammyK: I agree! Like you said, maybe not for this PR, but if we provided an out-of-the-box low-cardinality URL-to-resource-name converter, do you think that would be useful? For example, it would convert URL's from `/user?id=123` to `/user?id=?` or something like that. Thoughts? :)

src/DDTrace/Integrations/Laravel/V4/LaravelProvider.php
labbati: My point is that we should do nothing here :)
Did you add it because somehow we got to that point and no root span was created or this is justa  safety net?

tests/Integrations/Custom/Autoloaded/CommonScenariosTest.php
labbati: Let's discuss for the future as well, we may also think about some nice integration with the agent, which has all the relevant info to do that for all the language integrations :)

src/DDTrace/Integrations/Laravel/V4/LaravelProvider.php
SammyK: Ah gotcha! Yeah, I was thinking safety net, but maybe we don't need it since maybe we drop support for manual instrumentation in the future? :)

src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -18,21 +18,52 @@ public function __construct()
...
+        if (false === $value) {
+            return null;
+        }
+        return trim($value);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function stringValue($key, $default)
```
labbati: Should we honor `$default`? :)

src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -18,21 +18,52 @@ public function __construct()
...
-            } else {
-                $this->registry[$key] = $default;
-            }
+        if (isset($this->registry[$key])) {
+            return $this->registry[$key];
+        }
+
+        $value = self::get($key);
+        if (null === $value) {
+            return $default;
```
labbati: [non blocking] Shouldn't we store the result in `$this->registry[$key]` and remember for the next time? Keep in mind that this is never supposed to be called by users (which may provide different defaults), it is always called by a configuration instance. We may even considering to protect somehow this class.

src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -18,21 +18,52 @@ public function __construct()
...
+        }
+
+        $value = strtolower($value);
+        if ($value === '1' || $value === 'true') {
+            $this->registry[$key] = true;
+        } elseif ($value === '0' || $value === 'false') {
+            $this->registry[$key] = false;
+        } else {
+            return $default;
```
labbati: Same as above :)

src/DDTrace/Configuration/EnvVariableRegistry.php
SammyK: Ha! Good catch! I totally missed that one! :D

src/DDTrace/Configuration/EnvVariableRegistry.php
SammyK: I can totally change it back to that. My thought was that saving the fallback value in the registry  causes this behavior:

```php
var_dump($config->getString('some_unset_val', 'foo')); // "foo"
// ...and then later on...
var_dump($config->getString('some_unset_val', 'bar')); // "foo" (still)
```

I feel like this could be the cause of esoteric bugs later on. :D But if that's the desired behavior, I'm totally down for changing it back!

PR #273, opened 2019-01-30
Support for global tags

labbati:
### Description

This PR adds support for global tags that will be set in each span. They can be defined through the env `DD_TRACE_GLOBAL_TAGS=key1:valu1,key2:value2`

### Readiness checklist

### DIFF
--- a/src/DDTrace/Configuration/EnvVariableRegistry.php
+++ b/src/DDTrace/Configuration/EnvVariableRegistry.php
@@ -103,6 +103,48 @@ class EnvVariableRegistry implements Registry
     }
 
     /**
+     * Given a string like 'key1:value1,key2:value2', it returns an associative array
+     * ['key1'=> 'value1', 'key2'=> 'value2']
+     *
+     * @param string $key
+     * @return string[]
+     */
+    public function associativeStringArrayValue($key)
+    {
+        $default = [];
+
+        if (!isset($this->registry[$key])) {
+            $value = self::get($key);
+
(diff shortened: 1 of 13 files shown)


pawelchcki:
@SammyK this code is ready for another review round :) I addressed your suggestions, and merged in latest master

src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -102,6 +102,49 @@ public function floatValue($key, $default, $min = null, $max = null)
...
+    {
+        if (isset($this->registry[$key])) {
+            return $this->registry[$key];
+        }
+
+        $default = [];
+        $value = self::get($key);
+
+        if (null === $value) {
+            return $this->registry[$key] = $default;
```
SammyK: My bad for not catching this in the first CR! But this shouldn't save the default value:
```suggestion
            return $default;
```
Otherwise we'll get some weird behavior like this:
```php
var_dump($config->getString('some_unset_val', 'foo')); // "foo"
// ...and then later on...
var_dump($config->getString('some_unset_val', 'bar')); // "foo" (still)
```
I totally missed that in the first CR! Doh! :D

pawelchcki: Ohh nice catch!!!

SammyK:
Perfect! Nice idea to add a test for that @pawelchcki! :)

PR #277, opened 2019-02-01
Detect installed SAPIs and link INI file to each one

SammyK:
### Description

Currently, the post-install script only enables ddtrace for the CLI SAPI. This PR is a first draft for enabling the extension for all the installed SAPI's. It works by detecting a common convention of scoping INI files to specific SAPI's via the following directory structure:

`/<PREFIX>/<SAPI>/conf.d`

This is also the convention used by the popular package repo `ppa:ondrej/php`.

Whereas this is only convention, the PR won't be able to detect SAPI's with other custom config directories but we can improve the SAPI-detection logic in future PR's. :)

### Readiness checklist


package/post-install.sh
```diff
@@ -26,28 +26,49 @@ ddtrace.request_init_hook=${EXTENSION_AUTO_INSTRUMENTATION_FILE}
 EOF
 }
 
-function install_conf_d_file() {
+function install_conf_d_files() {
     INI_FILE_PATH="${EXTENSION_CFG_DIR}/$INI_FILE_NAME"
 
-    println "Creating ddtrace.ini"
-    println "\n"
+    if [[ ! -f $INI_FILE_PATH ]]; then
```
pawelchcki: This file is by convention autogenerated - and we need to be able to recreate it when e.g. file paths would update (or PHP version changes in between installations).

pawelchcki: Allowing custom configuration should be possible though, but I think it could be done in other place. E.g. via adding another file like `99-z-ddtrace.ini` in php installtion which . would be able to  override specific options

package/post-install.sh
```diff
@@ -26,28 +26,49 @@ ddtrace.request_init_hook=${EXTENSION_AUTO_INSTRUMENTATION_FILE}
...
+        println "Creating ddtrace.ini"
+        println "\n"
 
-    append_configuration_to_file "${INI_FILE_PATH}"
+        append_configuration_to_file "${INI_FILE_PATH}"
```
pawelchcki: seems like wrong tab/space creeped here

package/post-install.sh
pawelchcki: Would it make sense to extract this to a function like "generate_config_file()"?

package/post-install.sh
```diff
@@ -26,28 +26,49 @@ ddtrace.request_init_hook=${EXTENSION_AUTO_INSTRUMENTATION_FILE}
...
+    fi
+
+    for SAPI_CFG_DIR in ${SAPI_CONFIG_DIRS[@]}
+    do
+        println "Found SAPI config directory: ${SAPI_CFG_DIR}"
 
-    println "Linking ddtrace.ini to ${PHP_DDTRACE_INI}"
-    test -f "${PHP_DDTRACE_INI}" && rm "${PHP_DDTRACE_INI}"
-    ln -s "$INI_FILE_PATH" "${PHP_DDTRACE_INI}"
+        PHP_DDTRACE_INI="$SAPI_CFG_DIR/$INI_FILE_NAME"
```
pawelchcki: this could be extracted to separate fuinciton allowing execution like:
```bash
ensure_file_linked $INI_FILE_PATH "$PHP_CFG_DIR/$INI_FILE_NAME"
ensure_file_linked $INI_FILE_PATH "$PHP_FPM_CFG_DIR/$INI_FILE_NAME"
```
etc.

package/post-install.sh
SammyK: It think the diff is weird looking, but that line is within an `if` block now. In my IDE it looks correct. Let me know if isn't show showing up correctly for you! :)

PR #292, opened 2019-02-11
Logging more span's info when in debug mode

labbati:
### Description

We had one single error report from a customer that span encoding went out of memory. In order to answer the following questions we need to add a few info to our logs (only in debug mode).

1) is out of memory caused by too many spans?
2) is out of memory caused by a very large method’s argument added to a manually created span?
3) is out of memory caused by a very large method’s argument added to an automatically created span?
4) is memory leak/recursion problem in our encoder?

At the same time calculating some of the info above may be computationally expansive. For this reason this PR also partially refactor the logging engine to provide the log level information as a boolean.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Encoders/Json.php
+++ b/src/DDTrace/Encoders/Json.php
@@ -94,15 +94,15 @@ final class Json implements Encoder
         }
 
         self:self::logDebug(
-        "Encoding span '{id}' op: '{operation}' serv: '{service}' res: '{resource}' type '{type}'",
-        [
-            'id' => $span->getSpanId(),
-            'operation' => $span->getOperationName(),
-            'service' => $span->getService(),
-            'resource' => $span->getResource(),
-            'type' => $span->getType(),
-        ]
-    );
+            "Encoding span '{id}' op: '{operation}' serv: '{service}' res: '{resource}' type '{type}'",
+            [
+                'id' => $span->getSpanId(),
+                'operation' => $span->getOperationName(),
+                'service' => $span->getService(),
+                'resource' => $span->getResource(),
+                'type' => $span->getType(),
+            ]
+        );
         self::logDebug('Tags for span {id} \'tag:chars_count\' are: {lengths}', [
             'id' => $span->getSpanId(),
             'lengths' => implode(',', $lengths),
--- a/src/DDTrace/Encoders/Json.php
+++ b/src/DDTrace/Encoders/Json.php
@@ -84,7 +84,7 @@ final class Json implements Encoder
      */
     private function logSpanDetailsIfDebug(Span $span)
     {
-        if (!self::isLogLevelActive(LogLevel::DEBUG)) {
+        if (!self::isLogDebugActive()) {
             return;
         }
 
@@ -93,7 +93,7 @@ final class Json implements Encoder
             $lengths[] = "$tagName:" . strlen($tagValue);
         }
 
-        self:self::logDebug(
+        self::logDebug(
             "Encoding span '{id}' op: '{operation}' serv: '{service}' res: '{resource}' type '{type}'",
             [
                 'id' => $span->getSpanId(),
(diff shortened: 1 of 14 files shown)


src/DDTrace/Log/AbstractLogger.php
```diff
@@ -0,0 +1,34 @@
...
+     */
+    private $enabledLevels = [];
+
+    /**
+     * @param string $level
+     */
+    public function __construct($level)
+    {
+        // At the moment we only support debug or off.
+        foreach (LogLevel::all() as $knownLevel) {
```
pawelchcki: This looks to me a bit like overoptimization. Can multiple levels be active at the same time? 

Just asking since it looks like it required quite a bit of refactoring to make it work this way.

pawelchcki: Ok seems like I misunderstood intended use.

pawelchcki: I think if the method was implemented fully, I wouldn't misunderstood it so easily.

@labbati WDYT about implementation like this:

```php
# all() have all levels ordered from highest to lowest
# all preceding levels to given $level will be marked as enabled
foreach (LogLevel::all() as $knownLevel) {
    if (!$found) {
        if ($level === $knownLevel){
            $found = true;
        }
        $this->enabledLevels[$knownLevel] = true;
    } else {
        $this->enabledLevels[$knownLevel] = false;
    }
}
```

labbati: Thanks @pawelchcki you are totally right, I think that implementing just what we needed this time was a real mistake 😄

labbati: One simple note. If you don't mind I would invert the logic. With the code proposed here, if the user insert a wrong value (e.g. `my_level`) then I would have `debug` level activate while I would like to have logging off.

So something like

```
        # all() have all levels ordered from highest to lowest
        # all preceding levels to given $level will be marked as enabled
        $found = false;
        foreach (array_reverse(LogLevel::all()) as $knownLevel) {
            $found = $found || ($level === $knownLevel);
            $this->enabledLevels[$knownLevel] = $found || false;
        }
```

pawelchcki: Yea makes total sense!
Just one tidbit I would rename the `$found` to `$enabled`
Also Not sure if I understand this construct `$found || false` :D

src/DDTrace/Encoders/Json.php
```diff
@@ -72,6 +77,38 @@ private function encodeSpan(Span $span, Tracer $tracer)
         ], $json);
     }
 
+    /**
+     * Logs a Span's detailed info.
+     *
+     * @param Span $span
+     */
+    private function logSpanDetailsIfDebug(Span $span)
```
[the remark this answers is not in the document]
labbati: I like the idea of having a shared method specifically about the debug, less about callbacks (I am not a super fan of them).

Given also that they seems a little bit slower:

```php
require_once __DIR__ . '/benchmark.php';

$bm = new benchmark;

function whenDebugEnabled($callback)
{
    if ('value' === 'value') {
        return $callback();
    }
}

function ifDebugEnabled()
{
    return 'value' === 'value';
}

$with_callback = function () {
    whenDebugEnabled(function () { return true; });
};

$if_check = function () {
    if (ifDebugEnabled()) {
        return true;
    }
};

$bm->time_this('if check', $if_check);
$bm->time_this('callback', $with_callback);

$bm->summary();


PHP Version: 7.2.13

            if check | callback
          +---------------------+
 if check |      N/A |  147.31% |
 callback |   67.88% |      N/A |
          +---------------------+

 if check = 501,997 iterations per second
 callback = 340,765 iterations per second
```

What about a deal in between? 😄 

```php
if (self::isLogDebugEnabled()) {
    // ... do things
}
```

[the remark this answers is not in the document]
pawelchcki: ^^ its too simple - we need to overcomplicate this ;)
- read - oh, right, this is actually how it should be written :) lets do it

PR #295, opened 2019-02-12
Fix 5.4 internal functions being called twice

pawelchcki:
### Readiness checklist

### DIFF
--- a/tests/Integrations/Guzzle/V5/GuzzleIntegrationTest.php
+++ b/tests/Integrations/Guzzle/V5/GuzzleIntegrationTest.php
@@ -12,6 +12,7 @@ use GuzzleHttp\Ring\Client\MockHandler;
 use DDTrace\Tests\Common\SpanAssertion;
 use DDTrace\Tests\Common\IntegrationTestCase;
 use DDTrace\GlobalTracer;
+use DDTrace\Util\Versions;
 
 final class GuzzleIntegrationTest extends IntegrationTestCase
 {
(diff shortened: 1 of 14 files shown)


tests/Integrations/Guzzle/V5/GuzzleIntegrationTest.php
```diff
@@ -130,7 +131,14 @@ public function testDistributedTracingIsPropagated()
         // trace is: some_operation
         $this->assertSame($traces[0][0]->getContext()->getSpanId(), $found['headers']['X-Datadog-Trace-Id']);
         // parent is: curl_exec, used under the hood
-        $this->assertSame($traces[0][2]->getContext()->getSpanId(), $found['headers']['X-Datadog-Parent-Id']);
+
+        if (Versions::phpVersionMatches('5.4')) {
+            // in 5.4 curl_exec is not included in the trace due to being run through `call_func_array`
```
labbati: If we know of a specific method with predefined arguments combinations, we can avoid the "auto-magic" instrumentation and call only that specific method with

```
if (count($args) == 1) {
  $this->my_method($arg[0]);
} elseif (count($args) == 2) {
  $this->my_method($arg[0], $arg[1]);
}
```

so we can avoid using `call_user_func_array()`. If this can help in this case.

pawelchcki: hmm thats actually only my guess , but it seems to me that guzzle is doing something that actually hides the real `curl_exec` 

I haven't investigated it further than that though :|
Might be good to dig a bit deeper into why curl exec is not present in this trace. But this can be done in separate PR

PR #299, opened 2019-02-15
Update images and enable leak detection, and split tests in CI to Unit, Integration and Web

pawelchcki:
### Description

Use new Alpine based Debug images for most tests to ensure repeatable results. 

### Readiness checklist
- ~[ ] [Changelog entry](docs/changelog.md) added, if necessary~


.circleci/config.yml
```diff
@@ -256,200 +248,66 @@ jobs:
...
-      - <<: *IMAGE_DOCKER_DD_AGENT
-      - <<: *IMAGE_DOCKER_ELASTICSEARCH2
-      - <<: *IMAGE_DOCKER_HTTPBIN
-      - <<: *IMAGE_DOCKER_REDIS
-      - <<: *IMAGE_DOCKER_MEMCHACED
-      - <<: *IMAGE_DOCKER_MYSQL
-      - <<: *IMAGE_DOCKER_MONGODB
-      - <<: *IMAGE_DOCKER_REQUEST_REPLAYER
+    environment:
+      DDAGENT_HOSTNAME: 127.0.0.1
```
labbati: Shouldn't this be `DD_AGENT_HOST`

.circleci/config.yml
```diff
@@ -547,45 +405,109 @@ workflows:
...
+      - integration_tests:
+          name: "Intg tests 56"
+          integration_testsuite: "test-integrations-56"
+          image_tag: "ddtrace_alpine_php-5.6-debug"
+      - integration_tests:
+          name: "Intg tests 70"
+          integration_testsuite: "test-integrations-70"
+          image_tag: "ddtrace_alpine_php-7.0-debug"
+      - integration_tests:
+          name: "Intg tests 71"
```
labbati: Please don't hate me 😞 what about `s/Intg/Integrations/g`

.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 2.0
+version: 2.1
```
labbati: I think we ended up eating the `composer test-integration` part (different from `composer test-integrations-XX`)

.circleci/config.yml
pawelchcki: I agree with all my :heart: but  
I shortened it to make PHP version actually visible in Circle Ci UI. Which makes long names essentially useless

pawelchcki: image

pawelchcki: See `Static Analy...`
These changes actually started with `Integration tests 54` and even `Web integration tests 54` etc. 
But this turned into a bunch of 
`Integration tes...`
`Integration tes...`
`Integration tes...`
`Integration tes...`
`Integration tes...`
`Web integratio...`
`Web integratio...`
`Web integratio...`
`Web integratio...`
`Web integratio...`

I would have start with with version number but. The name of a job cannot start with a number in `2.1`.

Maybe something like `PHP 54 Integration tests` could work WDYT ?

pawelchcki: Hmm actually I like that more than current short forms. Let me quickly change.

pawelchcki: WDYT? 
image

.circleci/config.yml
pawelchcki: Its there but its not immediately obvious. I remove the yaml reference in lieu of normal `run` devfinition to make it more obvious

.circleci/config.yml
labbati: I sort of hate that I asked you to change that 😄! I am ok with this, but feel free to go back to the initial proposal or to abbreviate Integrations again if you prefer 👍

pawelchcki: You were right, I didn't like them either. But with naming I just usually try to use >first< thing that came to my mind,  turns out 2nd idea was much better :D

PR #300, opened 2019-02-15
Print backtrace on segfault

pawelchcki:
### Description

This PR adds optional signal handler to catch `SIGSEGV` signal (caused by segmentation fault) and print backtrace to stderr (then exit).

This signal handler is enabled via `ddtrace.debug_mode` ini bool flag, and is disabled by default to avoid any potential conflicts in PHP runtime configuration.

After backtrace is written to stderr the runtime is killed via `exit(1)` ensuring it will not be running in unspecified state.

### Readiness checklist

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -13,6 +13,7 @@
 #include "compat_zend_string.h"
 #include "ddtrace.h"
 #include "debug.h"
+#include "debug_backtrace.h"
 #include "dispatch.h"
 #include "dispatch_compat.h"
 #include "request_hooks.h"
@@ -67,6 +68,7 @@ static PHP_MINIT_FUNCTION(ddtrace) {
     if (DDTRACE_G(disable)) {
         return SUCCESS;
     }
+    ddtrace_install_backtrace_handler();
 
     ddtrace_dispatch_init(TSRMLS_C);
     ddtrace_dispatch_inject(TSRMLS_C);
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -56,7 +56,7 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
                   zend_ddtrace_globals, ddtrace_globals)
 STD_PHP_INI_ENTRY("ddtrace.ignore_missing_overridables", "1", PHP_INI_SYSTEM, OnUpdateBool, ignore_missing_overridables,
                   zend_ddtrace_globals, ddtrace_globals)
-STD_PHP_INI_ENTRY("ddtrace.debug_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, debug_mode, zend_ddtrace_globals,
+STD_PHP_INI_ENTRY("ddtrace.log_backtrace", "0", PHP_INI_SYSTEM, OnUpdateBool, log_backtrace, zend_ddtrace_globals,
                   ddtrace_globals)
 PHP_INI_END()
 
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -56,6 +56,8 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
                   zend_ddtrace_globals, ddtrace_globals)
 STD_PHP_INI_ENTRY("ddtrace.ignore_missing_overridables", "1", PHP_INI_SYSTEM, OnUpdateBool, ignore_missing_overridables,
                   zend_ddtrace_globals, ddtrace_globals)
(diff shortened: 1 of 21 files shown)


pawelchcki:
/cc: @willgittoes-dd @cgilmour - We're adding here a signal handler to printout backtraces and just want to make sure we're not going to blow up anything (this will be disabled by default and only enabled in `debugmode`, so this should be safe'ish).

pawelchcki:
Also obligatory: Εδώ είναι Σπάρτη :)

pawelchcki:
/cc: @SammyK and @labbati wdyt about enabling this when `DD_TRACE_DEBUG=true` is set ?

src/ext/debug_backtrace.c
```diff
@@ -0,0 +1,39 @@
+#include <execinfo.h>
```
SammyK: We'll need to add `debug_backtrace.c`, `debug_backtrace.h` and the two new test files to `package.xml` for PECL. :)

src/ext/ddtrace.c
```diff
@@ -55,6 +56,8 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
                   zend_ddtrace_globals, ddtrace_globals)
 STD_PHP_INI_ENTRY("ddtrace.ignore_missing_overridables", "1", PHP_INI_SYSTEM, OnUpdateBool, ignore_missing_overridables,
                   zend_ddtrace_globals, ddtrace_globals)
+STD_PHP_INI_ENTRY("ddtrace.debug_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, debug_mode, zend_ddtrace_globals,
```
SammyK: I love this! ❤️My only concern is that this will introduce two ways to get into debug mode and each of them mean something different. So I'm curious if we want to favor this INI setting over the current `DD_TRACE_DEBUG=true` method. Maybe @labbati has some thoughts on this? :)

src/ext/debug_backtrace.c
```diff
@@ -0,0 +1,39 @@
...
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "ddtrace.h"
+#include "debug_backtrace.h"
+
+ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
+
+void ddtrace_backtrace_handler(int sig) {
+    fprintf(stderr, "Datadog PHP Trace extension (DEBUG MODE)\n");
```
SammyK: What are your thoughts on using `php_log_err()` for these lines? :)
```suggestion
    php_log_err("Datadog PHP Trace extension (DEBUG MODE)");
```

tests/ext/segfault_backtrace_enabled.phpt
```diff
@@ -0,0 +1,16 @@
+--TEST--
+Dump backtrace when segmentation fault signal is raised and config enables it
+--INI--
+ddtrace.debug_mode=1
+--FILE--
+<?php
+posix_kill(posix_getpid(), 11); // boom
```
SammyK: ❤️💣💥

src/ext/ddtrace.c
willgittoes-dd: Maybe https://github.com/DataDog/architecture/pull/295 is interesting? I think you can have all of config file, command line args, and environment variables. And they just have a defined priority if they don't match.

labbati: I like the idea from @willgittoes-dd (command line args are probably not relevant for php use case but the rest is).
To answer @SammyK, even is they share the same name, I am not sure that mean the same thing 😃. Debug mode (as intended as before this PR) is "print useful info as the app is served" without changing how the underlying extension works, while in this case we are changing the way the underlying extension is working and I would keep the two configs separated. What about changing the name to `ddtrace.dump_backtrace`?

As will said, as soon as we implement all the config options, this can be sourced in different ways.

src/ext/debug_backtrace.c
pawelchcki: 👍 good idea

src/ext/debug_backtrace.c
```diff
@@ -0,0 +1,39 @@
...
+    fprintf(stderr, "Received Signal %d\n", sig);
+    void *array[MAX_STACK_SIZE];
+    size_t size = backtrace(array, MAX_STACK_SIZE);
+
+    fprintf(stderr, "Note: Backtrace below might be incomplete and have wrong entries due to optimized runtime\n");
+    fprintf(stderr, "Backtrace:\n");
+    backtrace_symbols_fd(array, size, STDERR_FILENO);
+    fflush(stderr);
+
+    exit(0);
```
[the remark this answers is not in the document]
pawelchcki: `exit sig` sounds good 👍

src/ext/ddtrace.c
pawelchcki: @labbati @SammyK how about `ddtrace.log_backtrace` nad DD_LOG_BACKTRACE env/sapi var ?

PR #303, opened 2019-02-16
Fix callable scope after tracing

### LINKED ISSUE #224: Invocation context seems not to be properly passed when invoking traced functions
Tested in the php 7.2 test runner. This is probably a different version of #149 

Invocation context seems not to be properly passed when invoking traced functions.

```
class A
{
    private function private_method()
    {
    }

    public function public_method()
    {
        spl_autoload_register(__CLASS__.'::private_method');
    }
}

dd_trace('spl_autoload_register', function () {
    return call_user_func_array('spl_autoload_register', func_get_args());
});

$a = new A();
$a->public_method();

$n = new \Not\Exist();
```

Results in 
```
Fatal error: Uncaught LogicException: Function 'A::private_method' not callable (cannot access private method A::private_method()) in /home/circleci/app/playground.php on line 16

LogicException: Function 'A::private_method' not callable (cannot access private method A::private_method()) in /home/circleci/app/playground.php on line 16

Call Stack:
    0.0010     393032   1. {main}() /home/circleci/app/playground.php:0
    0.0010     393752   2. A->public_method() /home/circleci/app/playground.php:20
    0.0010     394072   3. {closure:/home/circleci/app/playground.php:15-17}() /home/circleci/app/playground.php:11
    0.0011     394072   4. spl_autoload_register() /home/circleci/app/playground.php:16
```
While the expected output is
```
Fatal error: Uncaught Error: Class 'Not\Exist' not found in /home/circleci/app/playground.php on line 22

Error: Class 'Not\Exist' not found in /home/circleci/app/playground.php on line 22

Call Stack:
    0.0009     393032   1. {main}() /home/circleci/app/playground.php:0
```

SammyK:
### Description

This is a WIP to fix #224. It's not fixed yet, but I wanted to brain-dump my findings so far.

The main thing I'm noticing when running this test through the debugger is the change of [the `ctx` param passed to `curl_write_header()`](https://github.com/php/php-src/blob/623911f993f39ebbe75abe2771fc89faf6b15b9b/ext/curl/interface.c#L1752) when tracing `curl_exec()`.

I focused on the scope of `curl_write_header()`. Before tracing, the only vars that change in between two invocations of `curl_exec()` are `t` and `data`.

screen shot 2019-02-15 at 6 04 33 pm

After tracing, we also see `ctx` (and therefore `ch`) change as well between two invocations.

screen shot 2019-02-15 at 6 03 40 pm

I'll keep digging, but if anyone sees anything I'm missing, feel free to chime in! :)

The test makes a live call to Datadog right now just to illustrate this bug in as "real" of a context as possible. We'll refactor this out once we get a patch in place.

### Readiness checklist


SammyK:
After doing a lot of playing around with passing around different calling scopes/contexts, I eventually realized that essentially what I was trying to do at the extension level was something like this:

```php
class Foo
{
    private static function privateRegister()
    {
    }
}

$a = new Foo();

$register = function() {
    spl_autoload_register('Foo::privateRegister');
};
$register->bindTo($a);
$register();
```

But that code doesn't work as I expected. It still generates the same error:

```
Function 'Foo::privateRegister' not callable (cannot access private method Foo::privateRegister())
```

Then I realized that the proposed `dd_trace_invoke_original()` function to replace the  `call_user_func*()` calls in #284 would actually fix this issue as well. So since that would ultimately solve two big issues in one patch, I'm going to shift gears to focus on getting `dd_trace_invoke_original()` rocking. :)

labbati:
Happy to hear that we reduced the two issues to the same root cause! It makes sense to me now that I think about it! Good progress on this, thanks @SammyK!

SammyK:
Another update on this. I've been focusing on wrapping up #284 which was going to include a patch for this issue, but after some roadblocks with this issue that were holding up the other fix, we decided to release the PR's separately.

Also, the original code I posted above is actually incorrect, it should be this:

```php
class Foo
{
    private static function privateRegister()
    {
    }
}

$a = new Foo();

$register = function() {
    spl_autoload_register('Foo::privateRegister');
};
$foo = $register->bindTo($a, 'Foo');
$foo();
```

Or as @hollodotme pointed out:

```php
class Foo
{
    private static function privateRegister()
    {
    }
}

$a = new Foo();

$register = function() {
    spl_autoload_register('Foo::privateRegister');
};
$register->call($a);
```

And those work as expected. So that's the approach we're attempting at the C level after wrapping up the other PR. :)

src/ext/ddtrace.h
```diff
@@ -14,6 +14,13 @@ HashTable class_lookup;
 HashTable function_lookup;
 zend_bool log_backtrace;
 zend_function *current_fbc;
+#if PHP_VERSION_ID < 70000
+zval *original_this;
```
pawelchcki: Looks like there is a family of `originals` now. Would it make sense to refactor all `call_original()` data to a separate a separate struct ? 

It still could be a non pointer field in this struct to simplify stuff. 
I just feel that would be nice to start wrangling some of wild code we've accrued, and separating concerns sounds like a good start.

WDYT ?

SammyK: Great idea! Done. :)

PR #314, opened 2019-02-20
Fix accessing freed memory when instrumentation code un/instrumented itself 

pawelchcki:
### Description
In rare cases where instrumented code have actually removed/changed existing instrumentation of its own execution a freed pointer was accessed when removing recursion guard. 

### Readiness checklist

### DIFF
--- a/tests/ext/used_dispatch_shouldn_t_be_freed.phpt
+++ b/tests/ext/used_dispatch_shouldn_t_be_freed.phpt
@@ -1,0 +1,22 @@
+--TEST--
+xxx
+--FILE--
+<?php
+function test($a){
+    dd_trace("test", function($a){
+        return 'NEW HOOK ' . test($a);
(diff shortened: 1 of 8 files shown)


tests/ext/used_dispatch_shouldn_t_be_freed.phpt
```diff
@@ -0,0 +1,22 @@
+--TEST--
+xxx
```
SammyK: Looks like we're missing a little bit more of a description here. 😆

pawelchcki: 🤦‍♂️ 
Thanks for spotting that :D

pawelchcki:
thanks @SammyK for review. I added the missing test description 🤦‍♂️ 
Ready for another review round :D

PR #319, opened 2019-02-22
Loading of integrations before knowing if the library will be actually used

labbati:
## Problem
A user reached to as reporting a similar error in two of their instrumented applications: the browser would render the text below instead of the regular page.

```
Not found [OpenTracing\GlobalTracer]
```

## Investigation
After an in-dept investigation we found the root cause of the issue.
The user implemented an autoloader that does not follow the general requirement of the `spl_autoload_register` approach, which should not throw an exception if a class is not found in order for multiple class loaders to be chained together. 

While we are lenient with respect to this situation (we would not crash), the problem is that we are not lenient to `die()/exit()` when the class is not found.

Example

```
    public static function load($class) 
    {
        try {
            .... try to load, throws exception if not found ...
            return $class;
        } catch (Exception $e) {
            exit("Not found [". $class ."]");
        }
    }
```
While this is somehow not common, it made us realise that:

1. we should be robust enough to support this usage scenarios which may be common especially in legacy applications.
2. we are too invasive, as when, as an example, we do (very often) `class_exists('Some\Class)` we fire the autoloading mechanism, forcing the app to load classes that in the app flow would have been loaded later on otherwise.

## Proposed solution

We should never use `class_exist`, `method_exists` and `defined` with classed and constants other that the one in the `DDTrace` namespace, for which we provide our own autoloader.

In order to do that our integrations should be refactored after #325 in order to always registering methods to be traced.

A major refactoring was necessary in order to achieve this.

### Readiness checklist


src/DDTrace/Bootstrap.php
```diff
@@ -29,6 +29,7 @@ public static function tracerOnce()
         self::$bootstrapped = true;
         self::resetTracer();
         self::initRootSpan();
+        self::registerOpenTracing();
```
SammyK: ❤️

src/DDTrace/Integrations/Guzzle/GuzzleIntegration.php
```diff
@@ -3,30 +3,159 @@
...
+            $self->setUrlTag($span, $request);
+            $span->setTag(Tag::RESOURCE_NAME, $method);
+        };
+    }
+
+    /**
+     * @param Span $span
+     * @param mixed $request
+     */
+    protected function setUrlTag(Span $span, $request)
```
SammyK: Just a nit-pick, but these can be marked `private` since it's a `final` class. :)

tests/ext/trace_method_that_will_exist_later.phpt
```diff
@@ -0,0 +1,25 @@
+--TEST--
+Check that a function can be traced before it exists.
```
SammyK: This file was renamed to **tests/ext/trace_method_or_function_that_will_exist_later.phpt** in #325 so it can be deleted. :)

PR #351, opened 2019-03-11
Add Laravel 5.8 compatibility

labbati:
### Description

Prior to Laravel 5.8 'events' was available in container during first app()->bind()
because aliases were already registered.

With Laravel 5.8 this behavior changed, resulting in an error due to missing 'events' alias
in container.

Based on our latest changes that allow us to trace classes and methods that no longer exists,
we refactored the part of code which binds to events to directly trace event's constructor methods.

### Readiness checklist


src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
```diff
@@ -33,41 +33,34 @@ public function load()
...
-                    $span->setTag('laravel.route.name', Route::currentRouteName());
-                    $span->setTag('laravel.route.action', $event->route->getActionName());
-                    $span->setTag('http.url', $event->request->url());
-                    $span->setTag('http.method', $event->request->method());
-                }
-            );
 
-            $laravelApp['events']->listen(
-                'Illuminate\Foundation\Http\Events\RequestHandled',
-                function (RequestHandled $event) use ($self) {
```
SammyK: Since these references were removed, we can also remove the imports from above. :)

```php
use Illuminate\Foundation\Application;
use Illuminate\Foundation\Http\Events\RequestHandled;
use Illuminate\Routing\Events\RouteMatched;
```

tests/Frameworks/Laravel/Version_5_8/resources/views/simple_view.blade.php
```diff
@@ -0,0 +1,8 @@
+<!doctype html>
+<html lang="{{ app()->getLocale() }}">
+    <head>
+    </head>
+    <body>
+        Hi!
```
SammyK: 👋

labbati: Easy fix, pushed a commit for that 👍

src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
labbati: Easy fix 😄 pushed a commit for that 👍

PR #360, opened 2019-03-13
Integration aware spans

labbati:
### Description

As a preparatory work to support Trace Analytics configuration from the PHP tracer, we opted for making Spans integration aware. With this in mind we can enable batch post processing and be totally aware of the specific integration, and consequently integration's configuration, that generated such span.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
+++ b/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
@@ -12,26 +12,10 @@ use DDTrace\Type;
 /**
  * ElasticSearch driver v1 Integration
  */
-class ElasticSearchIntegration extends AbstractIntegration
+final class ElasticSearchIntegration extends AbstractIntegration
 {
     const NAME = 'elasticsearch';
     const DEFAULT_SERVICE_NAME = 'elasticsearch';
-
-    /**
-     * @var self
-     */
-    private static $instance;
-
-    /**
-     * @return self
-     */
-    public static function getInstance()
-    {
-        if (null === self::$instance) {
-            self::$instance = new self();
-        }
-        return self::$instance;
-    }
 
     /**
      * @return string The integration name.
--- a/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
+++ b/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
@@ -18,6 +18,22 @@ class ElasticSearchIntegration extends AbstractIntegration
     const DEFAULT_SERVICE_NAME = 'elasticsearch';
 
     /**
+     * @var self
+     */
+    private static $instance;
+
+    /**
+     * @return self
+     */
+    public static function getInstance()
+    {
+        if (null === self::$instance) {
+            self::$instance = new self();
+        }
+        return self::$instance;
+    }
+
+    /**
      * @return string The integration name.
      */
     public function getName()
--- a/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
+++ b/src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
@@ -12,10 +12,26 @@ use DDTrace\Type;
 /**
  * ElasticSearch driver v1 Integration
  */
-final class ElasticSearchIntegration extends AbstractIntegration
+class ElasticSearchIntegration extends AbstractIntegration
 {
     const NAME = 'elasticsearch';
     const DEFAULT_SERVICE_NAME = 'elasticsearch';
+
+    /**
+     * @var self
+     */
+    private static $instance;
+
+    /**
+     * @return self
+     */
+    public static function getInstance()
+    {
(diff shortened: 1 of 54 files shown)


src/DDTrace/Integrations/Mongo/MongoCollectionIntegration.php
```diff
@@ -19,17 +19,20 @@ protected static function loadIntegration()
 
         // MongoCollection::__construct ( MongoDB $db , string $name )
         self::traceMethod('__construct', function (Span $span, array $args) {
+            $span->setIntegration(MongoIntegration::getInstance());
```
pawelchcki: Would it be possible to override self::traceMethod to automatically call ->setIntegration to every span ?

labbati: Very good idea, but I will push to the end of the signature to avoid breaking changes if a user implemented an integration relying on this utility method.

tests/Common/SpanIntegrationChecker.php
```diff
@@ -0,0 +1,76 @@
...
+
+namespace DDTrace\Tests\Common;
+
+use DDTrace\Contracts\Span;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Checks that spans are assigned the proper integration.
+ */
+class SpanIntegrationChecker
```
SammyK: Probably doesn't matter, but just to lock this down a bit. :)

```suggestion
final class SpanIntegrationChecker
```

tests/Common/SpanIntegrationChecker.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    /**
+     * Returns the known matching patterns operation ==> integration class.
+     */
+    public function defineIntegrationsByPattern()
```
SammyK: This can be static and private. :)
```suggestion
    private static function defineIntegrationsByPattern()
```

tests/Common/SpanIntegrationChecker.php
```diff
@@ -0,0 +1,76 @@
...
+        ];
+    }
+
+    /**
+     * Checks that $span belongs to the proper integration.
+     *
+     * @param TestCase $test
+     * @param Span $span
+     */
+    public function checkIntegration(TestCase $test, Span $span)
```
SammyK: This could be static. :)
```suggestion
    public static function checkIntegration(TestCase $test, Span $span)
```

tests/Integrations/Symfony/V2_3/RouteNameTest.php
```diff
@@ -13,6 +13,13 @@ protected static function getAppIndexScript()
         return __DIR__ . '/../../../Frameworks/Symfony/Version_2_3/web/app.php';
     }
 
+    protected static function getEnvs()
```
SammyK: Since this is used for several of the tests, we could possibly extract this to a trait if you think that would be better. :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchIntegration.php
```diff
@@ -17,6 +17,22 @@ class ElasticSearchIntegration extends AbstractIntegration
...
 
+    /**
+     * @var self
+     */
+    private static $instance;
+
+    /**
+     * @return self
+     */
+    public static function getInstance()
```
SammyK: Would it make sense to move this up to the abstract `Integration` class since nearly all the integrations need it? Also - what do you think about making the `Integration` class constructor private to help enforce singleton? :)

labbati: Sounds good! Originally I did this because classes are not marked as final and making the constructor protect the constructor meant breaking things for users that possibly extended them. At this point I am both moving to parent class instantiation and making classes final. 👍

tests/Common/SpanIntegrationChecker.php
labbati: If you don't mind (no problem on changing my mind) I'd prefer to keep this classes with non-static methods unless we have to do it for functional/performance reasons (and of course this is not the case for testing purposes).

The point is that it happened more often that I had to inject a service at a certain point than I needed to access a class when I could not create an instance for performance reasons 😄 

Again totally opened to do this if you don't agree.

tests/Common/SpanIntegrationChecker.php
labbati: Same as above ^^

tests/Integrations/Symfony/V2_3/RouteNameTest.php
labbati: Good idea. BTW given that these tests all are web framework tests, what if I can move directly there instead of a separate trait? In the end ALL web framework tests have to be tested with this env, WDYT?

labbati:
@SammyK I actually reverted the "move singleton to the abstract class" as we want per integration singleton, one singleton for all the integrations 😄

labbati:
All tests pass but they are not reported: https://circleci.com/workflow-run/684df80f-1c6e-488c-9cdf-9db0ccefb4e7

Force merging it

Issue #346, opened 2019-03-08
WoD with Laravel 5.8


stayallive:
I have been trying to get some useful debug info but as far as I can tell the Laravel integration breaks with Laravel 5.8 and dumps the following (truncated error) to the screen.

```
2019/03/08 20:51:15 [error] 24207#24207: *1 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught ReflectionException: Class events does not exist in /home/forge/cert.chief.app/releases/20190308204600/vendor/laravel/framework/src/Illuminate/Container/Container.php:794
Stack trace:
#0 /home/forge/cert.chief.app/releases/20190308204600/vendor/laravel/framework/src/Illuminate/Container/Container.php(794): ReflectionClass->__construct('events')
#1 /home/forge/cert.chief.app/releases/20190308204600/vendor/laravel/framework/src/Illuminate/Container/Container.php(671): Illuminate\Container\Container->build('events')
#2 /home/forge/cert.chief.app/releases/20190308204600/vendor/laravel/framework/src/Illuminate/Container/Container.php(619): Illuminate\Container\Container->resolve('events', Array)
#3 /home/forge/cert.chief.app/releases/20190308204600/vendor/laravel/framework/src/Illuminate/Foundation/Application.php(757): Illuminate\Container\Container->make('events', Array)
#4 /home/forge/cert.chief.app/releases/20190308204600/vendor/laravel/framework/src/Illuminate/Conta" while reading response header from upstream, client: 83.84.103.115, server: cert.chief.app, request: "GET / HTTP/2.0", upstream: "fastcgi://unix:/var/run/php/php7.2-fpm.sock:", host: "cert.chief.app"
```

I have been trying to get some more useful debug in info but got nil to show.

Disabled the Laravel integration with the following in nginx until the cause is found.

```
fastcgi_param DD_INTEGRATIONS_DISABLED laravel;
```

Might be nice to also figure out how to not dump stack traces to the screen in production.

ArondeParon:
We experienced this same issue as well today. Took us quite some time to figure out why this was going wrong, but it's definitely the APM that is causing our application to break.

![image](https://user-images.githubusercontent.com/7697/54120993-94919f80-43f9-11e9-9adb-db351affbd6f.png)

stayallive:
The way of "safely" getting the user ID is a probable source of errors.

See: https://3v4l.org/fgakS, it cannot be catched :)

[...]

PR #363, opened 2019-03-14
Fix user ID access in Laravel integration

SammyK:
### Description

Fixes the remaining issue in #346 where access to the user ID was generating errors when the user model was null.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
+++ b/src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
@@ -10,6 +10,7 @@ use DDTrace\Scope;
 use DDTrace\Tag;
 use DDTrace\Type;
 use DDTrace\Util\TryCatchFinally;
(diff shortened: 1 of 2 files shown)


src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
```diff
@@ -52,8 +52,10 @@ public function load()
 
             $span = $self->rootScope->getSpan();
             try {
-                $user = auth()->user()->id;
-                $span->setTag('laravel.user', strlen($user) ? $user : '-');
+                $user = auth()->user();
+                if (null !== $user) {
```
labbati: This is not required if starting from 5.0 we always have that same interface. Do you think it would be safer if is conditionally do this operation only if `is_a(... interface ...)`?

SammyK: Good point! Fixed. :)

PR #367, opened 2019-03-18
Trace Analytics Client Configuration

labbati:
### Readiness checklist


src/DDTrace/Processing/TraceAnalyticsProcessor.php
```diff
@@ -0,0 +1,42 @@
...
+            return;
+        }
+
+        // If a trace analytics tag has already been set, then we honor it.
+        if ($span->hasTag(Tag::ANALYTICS_KEY)) {
+            return;
+        }
+
+        $integration = $span->getIntegration();
+        if ($integration->isTraceAnalyticsEnabled()) {
```
SammyK: Since this the return value is nullable. :)

```suggestion
        if (null !== $integration && $integration->isTraceAnalyticsEnabled()) {
```

src/DDTrace/Processing/TraceAnalyticsProcessor.php
```diff
@@ -0,0 +1,42 @@
...
+final class TraceAnalyticsProcessor
+{
+    /**
+     * Process the span adding the trace analytics client config option when appropriate.
+     *
+     * @param Span $span
+     */
+    public function process(Span $span)
+    {
+        // We only consider spans that are marked as trace analytics candidates, otherwise the customer bill would
```
SammyK: ❤️

src/DDTrace/Span.php
```diff
@@ -241,6 +256,60 @@ public function getAllTags()
...
+     */
+    public function getMetrics()
+    {
+        return $this->metrics;
+    }
+
+    /**
+     * @return string[] The known metrics names
+     */
+    private static function getMetricsNames()
```
SammyK: 🔝

src/DDTrace/Processing/TraceAnalyticsProcessor.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace DDTrace\Processing;
```
SammyK: I'm curious if we should consider naming this `namespace DDTrace\Filters` to be in line with other tracers? Maybe filter functionality is something we do in a separate PR and refactor `TraceAnalyticsProcessor` into a filter. Maybe we can add an `@TODO` comment? Or we could just mention it in the backlog. :)

tests/Common/SpanChecker.php
```diff
@@ -138,4 +151,25 @@ public function flattenTraces($traces)
...
+
+    /**
+     * PHP < 5.6 does not offer a way to filter only specific elements in an array by key.
+     *
+     * @param array $associative
+     * @param string[] $allowedKeys
+     * @param bool $include
+     * @return array
+     */
+    private function filterArrayByKey54($associative, $allowedKeys, $include = true)
```
SammyK: Without reading the docblock comment, the "54" in the method name is a bit confusing (I saw the usage before the definition so that why I was confused). 😄 But also since this affects PHP 5.5 as well. And we can also make it static if we want to. :)

```suggestion
    private static function filterArrayByKey($associative, $allowedKeys, $include = true)
```

Maybe we do a check for the PHP version and use `array_filter()` with the supported constants and fall back to this code for older versions? Or maybe we just keep the method body as is and not worry about it. :)

SammyK:
Also - I just realized that we need to rebase the dev branch from master since there will likely be a few merge conflicts with this one. :)

src/DDTrace/Processing/TraceAnalyticsProcessor.php
labbati: This is a good discussion! If you don't mind I'd prefer to add it into the comment because I feel like `xFilter` would be a sort of misleading at this stage. There is no mechanism to filter out spans. Of course this can be translated to filter once we put this mechanism in place. Being it final it should be no risky at all. Would you agree?

src/DDTrace/Processing/TraceAnalyticsProcessor.php
labbati: Adding as a safety net, but we already check for the integration not being `null` on line 28 :)

tests/Common/SpanChecker.php
labbati: Good catch! Renaming an moving to static! If you don't mind I would not move to double implementation (pre and post 5.5) as I think it'd add some noise. What do you think?

labbati:
hey @SammyK thanks for the awesome review! Back to you with some changes/ansswers 😄

labbati:
Force merging as this is green on https://circleci.com/workflow-run/4a26c85d-0988-4416-a6ae-5659a48f311e but success is not reported yet!

PR #370, opened 2019-03-19
Reimplement blacklist as simple list to avoid problems with missing symbols between modules

pawelchcki:
### Description

Before module blacklist was implement using pcre regexp which caused problems in some installations which differ in installed library versions which changed load order which caused the problem to surface.

This change avoids using php's pcre regexps.

### Readiness checklist

### DIFF
--- a/src/ext/request_hooks.c
+++ b/src/ext/request_hooks.c
@@ -6,30 +6,22 @@
 #include <Zend/zend.h>
 #include <Zend/zend_compile.h>
 #include <php_main.h>
-
-#if PHP_VERSION_ID >= 70000
-#include <php/ext/pcre/php_pcre.h>
-#else
-#include <ext/pcre/php_pcre.h>
-#endif
+#include <string.h>
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
 #if PHP_VERSION_ID < 70000
 int dd_no_blacklisted_modules(TSRMLS_D) {
-    char *blacklist_regexp = DDTRACE_G(internal_blacklisted_modules_regexp);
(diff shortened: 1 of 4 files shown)


pawelchcki:
Lint test seem to fail but its unrelated to changes in this PR.
@SammyK could you review this PR ?

src/ext/request_hooks.c
```diff
@@ -6,30 +6,22 @@
...
+    int rv = 1;
     HashPosition pos;
 
-    if ((pce = pcre_get_compiled_regex(blacklist_regexp, &re_extra, &re_options TSRMLS_CC)) != NULL) {
+    if (blacklist) {
         zend_hash_internal_pointer_reset_ex(&module_registry, &pos);
 
         while (zend_hash_get_current_data_ex(&module_registry, (void *)&module, &pos) != FAILURE) {
-            if (!pcre_exec(pce, re_extra, module->name, strlen(module->name), 0, re_options, NULL, 0)) {
+            if (module && module->name && strstr(blacklist, module->name) != NULL) {
```
SammyK: We don't have to change this right now, but I believe this will match a bit too liberally using `strstr()`. For example if I had:

```
ddtrace.internal_blacklisted_modules_list=mongodb
```

And I had the `mongo` extension installed, this would also blacklist the `mongo` extension, right? I don't think it'll be an issue right now for the limited subset we're shooting for, but something to think about for the future; especially for short ext names like `ui` which would blacklist: q*ui*ckhash, Mosq*ui*tto, q*ui*ckhash, and u*ui*d from PECL for example. :)

pawelchcki: I agree with whole heart with your points 👍 
Lets followup with another PR, to make this behavior more predictable

PR #373, opened 2019-03-19
Add configure flag to enable debug builds

SammyK:
### Description

This PR makes it easy to enable a debug build of ddtrace using the `--enable-ddtrace-debug` config flag. If this flag is present during the configure step, ddtrace will be built with debug symbols and debug messages will be exposed. Additionally, the `dd_trace_noop()` function is moved to debug-only builds. An added advantage of this is that it makes it easy to check if debug mode is enabled in userspace (for example if we need to check for debug mode in a test in the future).

### Readiness checklist

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -1,4 +1,5 @@
 PHP_ARG_ENABLE(ddtrace, whether to enable Datadog tracing support,[  --enable-ddtrace   Enable Datadog training support])
+PHP_ARG_ENABLE(ddtrace-debug, whether ddtrace should be built in debug mode,[  --enable-ddtrace-debug Build Datadog tracing in debug mode],, no)
 
 PHP_ARG_WITH(ddtrace-sanitize, whether to enable AddressSanitizer for ddtrace,[  --with-ddtrace-sanitize Build Datadog tracing with AddressSanitizer support], no, no)
 
@@ -8,6 +9,11 @@ if test "$PHP_DDTRACE" != "no"; then
 	  EXTRA_CFLAGS="-fsanitize=address -fno-omit-frame-pointer"
 	  PHP_SUBST(EXTRA_LDFLAGS)
     PHP_SUBST(EXTRA_CFLAGS)
+  fi
+
+  if test "$PHP_DDTRACE_DEBUG" != "no"; then
+    AC_DEFINE(DDTRACE_DEBUG, 1, [ ])
+    STD_CFLAGS="-g -O0 -Wall"
   fi
 
   PHP_NEW_EXTENSION(ddtrace, src/ext/ddtrace.c src/ext/dispatch_setup.c src/ext/dispatch.c src/ext/request_hooks.c src/ext/compat_zend_string.c src/ext/dispatch_compat_php5.c src/ext/dispatch_compat_php7.c src/ext/backtrace.c src/ext/logging.c src/ext/env_config.c, $ext_shared,, -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file - [read more
 
 ## [Unreleased]
 
+### Changed
+- Removed `dd_trace_noop()` from non-debug builds of ddtrace #373
+
+### Added
+- Configure flag `--enable-ddtrace-debug` to build ddtrace with debug symbols and expose debug messages #373
+
 ## [0.16.0]
 
 ### Changed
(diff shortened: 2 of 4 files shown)


config.m4
```diff
@@ -1,4 +1,5 @@
 PHP_ARG_ENABLE(ddtrace, whether to enable Datadog tracing support,[  --enable-ddtrace   Enable Datadog training support])
+PHP_ARG_ENABLE(ddtrace-debug, whether ddtrace should be built in debug mode,[  --enable-ddtrace-debug Build Datadog tracing in debug mode],, no)
```
pawelchcki: Do you expect our clients to enable debug mode through pecl install ?

config.m4
```diff
@@ -10,6 +11,11 @@ if test "$PHP_DDTRACE" != "no"; then
     PHP_SUBST(EXTRA_CFLAGS)
   fi
 
+  if test "$PHP_DDTRACE_DEBUG" != "no"; then
+    AC_DEFINE(DDTRACE_DEBUG, 1, [ ])
+    STD_CFLAGS="-g -O0 -Wall"
```
pawelchcki: I believe most of debug mode enabling will happen when the extension is checked out. As such its probably more useful to have the debug mode in main Makefile since its used more often, and also allows to easily switch builds between versions due to out of tree builds.

pawelchcki: TBH I'm not a big fan the enable/disable flags of pecl (they are very hard to automate when installing etc.)

config.m4
SammyK: I think they do but I don't have any data to back that up. :D The main thing is that it's quite common for extensions to provide a way to enable debug mode via a config flag (i.e., `--enable-apcu-debug` for APCu, `--enable-xdebug-dev` for Xdebug, etc) So it's just something that seems to be "the PHP way" to do it. :)

config.m4
SammyK: > I believe most of debug mode enabling will happen when the extension is checked out. 

Agreed.

> As such its probably more useful to have the debug mode in main Makefile since its used more often

We don't have any documentation around the Makefile so when an extension developer does a git checkout, they will likely run the common extension building commands (`$ phpize && ./configure && make && sudo make install`) which blows away the packaged Makefile anyway. :) What do you think about migrating away from the packaged Makefile by moving some features to the config script (like debug mode) and other features to a makefile fragment (like extra make targets). That way we follow more conventional extension development practices.  What are your thoughts? :)

SammyK:
Thanks for the CR @pawelchcki! :)

1. I'm cool with moving `dd_trace_noop()` back into non-debug builds. :)
2. One way the swoole extension gets away with this is by having two flags `--enable-debug` (for debug symbols) & `--enable-debug-log` (for debug messages). Do you think it would make sense to do something similar?
3. True - I'm curious to hear your thoughts on my suggestion to migrate away from a version-controlled Makefile. :)

config.m4
pawelchcki: I think `phpize` is broken in that it overwrites Makefile and doesn't support out of folder builds 
> We don't have any documentation around the Makefile 
 we can add some :)

We did previously use another Makefile that wasn't the best experience to always type `make -f DDMakefile test_c` - esp since simple make without `-f  DDMakefile` was almost never used

> What do you think about migrating away from the packaged Makefile by moving some features to the config script (like debug mode) and other features to a makefile fragment (like extra make targets). That way we follow more conventional extension development practices. What are your thoughts? :) 

Makefile has good builtin dependency management which both scripts and composer doesn't have.

For most of the tasks within the current Makefile its quite crucial to have dependency management to have speedy builds.

The main purpose of current Makefile is to allow executing things like:
```
docker-compose run 7.2 make test_c
docker-compose run 7.1 make test_c
docker-compose run 7.0 make test_c
docker-compose run 5.6 make test_c
docker-compose run 5.4 make test_c
```

And  don't have to care about phpize leftovers between versions and have fast recompilation on changes

But this gives me idea that I think adding a script like `bin/make` which would:
```bash
#!/bin/bash
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"

make -f $DIR/../Makefile.something  "$@"
```

Would probably solve most problems with adding `-f DDMakefile` to every make invocation.

pawelchcki:
I think we're onto something awesome here!

>One way the swoole extension gets away with this is by having two flags --enable-debug (for debug symbols) & --enable-debug-log (for debug messages). Do you think it would make sense to do something similar?

Sounds just like the fix we need!

morrisonlevi:
Given that we have `make debug all` and such, I don't think we need a configure flag. The other discussion about the DD_PRINTF and logging stuff makes it seem like it should be a different PR. How's that sound? Feel okay closing this PR?

SammyK:
Absolutely makes sense. :)

PR #375, opened 2019-03-20
Fix issue where error traces don't appear in "Total Errors" panel

SammyK:
### Description

Error traces do not appear under the "Total Errors" panel in the APM UI unless `error.type` is set. This PR ensures that `error.type` is set with 500-level HTTP responses.

### Readiness checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file - [read more
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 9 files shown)


SammyK:
Thanks for the review @labbati! The linting will fail for all PR's until code sniffer releases a fix for [a known bug](https://github.com/squizlabs/PHP_CodeSniffer/issues/2450) that affects the PHP tracer. They said it should be patched in the next release. :)

Issue #192, opened 2018-12-17
User extractor - Error when the id is a protected property


Lavoaster:
We use Doctrine within our Laravel Project, and not eloquent. So we don't expose our object properties, however, the code for extracting users appears to rely on accessing the id property directly, as apposed to going through the `\Illuminate\Contracts\Auth\Authenticatable` interface, which is required anyway.

You could modify the code in place to the following, and get rid of the try/catch.

```php
$user = auth()->user();
// You probably would want to import the class below
if ($user && $user instanceof \Illuminate\Contracts\Auth\Authenticatable) {
    $span->setTag('laravel.user', strlen($user->getAuthIdentifierName()) > 0 ? $user->getAuthIdentifierName() : '-');
}
```

tatianajiselle:
hey -- please see this PR for this ticket; https://github.com/DataDog/dd-trace-php/pull/376

SammyK:
Hey @Lavoaster! Thanks again for reporting this issue. This patch was applied in version `0.16.0` via #363.

PR #376, opened 2019-03-21
update user extractor to use authenticatable class

tatianajiselle:
### Description

https://github.com/DataDog/dd-trace-php/issues/192

Just opening a PR for the suggested change from this issue...

### DIFF
--- a/src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
+++ b/src/DDTrace/Integrations/Laravel/V5/LaravelIntegrationLoader.php
@@ -52,12 +52,13 @@ class LaravelIntegrationLoader
             $args = func_get_args();
 
             $span = $self->rootScope->getSpan();
-            try {
-                $user = auth()->user();
-                if ($user instanceof Authenticatable) {
-                    $span->setTag('laravel.user', $user->getAuthIdentifier());
-                }
-            } catch (\Exception $e) {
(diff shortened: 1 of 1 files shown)


labbati:
Looking better at the code and the interface definition I propose a different change, sorry for me changing my mind.

`$user && $user instanceof ...` seems to be redundant because while `null instance of A\Class` would throw a fatal, `$a = null; $a instanceof A\Class` would not as $a is not a constant, compared to null.

Moreover, while your approach of removing the try catch is awesome in a standard environment, we may prefer to take less risks when interacting with frameworks. Let's keep it.

At this point basically, I think that the existing approach (which just recently stopped accessing `$user->id` directly`, should work in your case.

This was released in 0.16.0.

SammyK:
Hey @tatianajiselle! It's good to see you out here in GitHub land! :D Thank you so much for the PR. I'm curious to hear more details of what this change addresses outside of the #363 PR? I could have totally missed something on that. :)

Also - the `getAuthIdentifierName()` method seems to [return the name of the ID column in the database](https://github.com/laravel/framework/blob/5.2/src/Illuminate/Auth/GenericUser.php#L34) and not the user name. So I think we'll want to stick with `getAuthIdentifier()` for now.

I'm going to close this for now, but if I missed something feel free to chime in and I'll reopen. And thanks again for opening the PR! :)

PR #378, opened 2019-03-22
Vendor in MessagePack

SammyK:
### Description

This PR adds a new function `dd_trace_serialize_msgpack()` which takes a simple array (composed of types `array`, `string`, `int`, `float`, `bool` or `null`) and returns a binary string ~~encoded by [**msgpack-c**](https://github.com/msgpack/msgpack-c)~~ (statically linked).

**Edit:** After more research, it appears as if [**ludocode/mpack**](https://github.com/ludocode/mpack) is a better choice to vendor in since it is a 100% C implementation that is designed to be embedded just like we need for ddtrace. :)

Even though MessagePack is now the default encoder, the JSON encoder still exists and can be implemented by setting the environment variable `DD_TRACE_ENCODER=json`.

### Readiness checklist


Makefile
```diff
@@ -78,8 +79,26 @@ debug:
...
 
+clang_find_files_to_lint:
+	@find ./ -type f \
+	-path ./tmp -prune -o \
+	-path ./vendor -prune -o \
+	-path ./tests -prune -o \
+	-path ./src/ext/mpack -prune -o \
+	-iname "*.h" -o -iname "*.c"
+
+clang_format_check:
```
pawelchcki: ❤️

pawelchcki: Thanks for implementing it, the check internally should produce more readable output too

src/DDTrace/Transport/Http.php
```diff
@@ -128,7 +131,10 @@ private function sendRequest($url, array $headers, $body)
         }
 
         if ($statusCode !== 200) {
-            self::logError('Reporting of spans failed, status code {code}', ['code' => $statusCode]);
+            self::logError(
+                'Reporting of spans failed, status code {code}: {response}',
+                ['code' => $statusCode, 'response' => $response]
```
pawelchcki: ❤️

src/ext/dispatch.c
```diff
@@ -437,7 +437,7 @@ static int update_opcode_leave(zend_execute_data *execute_data TSRMLS_DC) {
     EX(called_scope) = DECODE_CTOR(EX(called_scope));
 
     zend_arg_types_stack_3_pop(&EG(arg_types_stack), &EX(called_scope), &EX(current_object), &EX(fbc));
-    zend_vm_stack_clear_multiple(TSRMLS_CC);
+    zend_vm_stack_clear_multiple(TSRMLS_C);
```
pawelchcki: 👍 😅

src/ext/serializer.c
```diff
@@ -0,0 +1,130 @@
...
+
+    if (is_assoc) {
+        mpack_finish_map(writer);
+    } else {
+        mpack_finish_array(writer);
+    }
+}
+#endif
+
+static void ddtrace_zval_to_writer(mpack_writer_t *writer, zval *trace) /* {{{ */
```
pawelchcki: Variable named `trace` makes me think there might be something trace specific, while this is simply zval serializer.

`msgpack_write_zval` also would help with readability imho.

To avoid very long function names, I mostly tried to add `ddtrace_` prefix to non static functions, (with non static functions the probability of symbol name conflic exits, so the prefix needs to be there).

src/ext/serializer.c
```diff
@@ -0,0 +1,130 @@
...
+        case IS_TRUE:
+        case IS_FALSE:
+            mpack_write_bool(writer, Z_TYPE_P(trace) == IS_TRUE);
+            break;
+        case IS_STRING:
+            mpack_write_cstr(writer, ZSTR_VAL(Z_STR_P(trace)));
+            break;
+#endif
+        default: {
+            // @TODO Error message
```
pawelchcki: Probably a good place to either throw an error or at least log it.
Since we're serializing here a ordinary PHP array with only basic values (or other arrays of the same) and if its used by us it should never contain unexpected objects.

src/ext/serializer.c
```diff
@@ -0,0 +1,130 @@
...
+            mpack_write_cstr(writer, "unknown type");
+        } break;
+    }
+}
+
+int ddtrace_serialize_trace(zval *trace, zval *retval) {
```
pawelchcki: This function is more generic wdyt about naming it `ddtrace_serialize_simple_array` ?

src/ext/serializer.c
```diff
@@ -0,0 +1,130 @@
+#include <php.h>
+#include <stdio.h>
+#include "mpack/mpack.h"
+
+static void ddtrace_zval_to_writer(mpack_writer_t *writer, zval *trace);
+
+#if PHP_VERSION_ID < 70000
+static void ddtrace_hash_table_to_writer(mpack_writer_t *writer, HashTable *ht) /* {{{ */
```
pawelchcki: similarliy to suggestion bellow:

the "ddtrace_" prefix is unnecessary, and at least to my ears `write_hash_table` sounds a bit nicer WRT intent of the function.

src/ext/ddtrace.c
```diff
@@ -48,6 +50,130 @@
...
+ZEND_BEGIN_ARG_INFO(arginfo_tracer_get_safe_root_scope, 0)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO(arginfo_tracer_as_array, 0)
+ZEND_END_ARG_INFO()
+/* }}} */
+
+/* {{{ Function table */
+static const zend_function_entry php_ddtrace_span_context_funcs_interface[] = {
+    PHP_ABSTRACT_ME(SpanContext, getBaggageItem, arginfo_span_context_get_baggage_item) PHP_ABSTRACT_ME(
```
pawelchcki: Is the SpanContext used from within C code ? It looks to me like the `msgpack` serialization is quite generic and doesn't rely on it. Or did I miss something ?

My point is, is that if its not used from C level, then I wonder what os the reason for this functionality here

src/ext/ddtrace.c
```diff
@@ -48,6 +50,130 @@
 #define PHP7_UNUSED(...) UNUSED(__VA_ARGS__)
 #endif
 
+/* {{{ arginfo */
```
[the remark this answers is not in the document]
SammyK: Thanks for the CR! I'm thinking of reverting moving any of the interfaces to the C level and changing the `dd_trace_serialize_trace()` API to accept a plain-old PHP array instead of an instance of the tracer. What do you think? :)

src/ext/ddtrace.c
SammyK: I started to move all the interfaces down to the C level, but realized that I was probably a Bad Idea™. :) I think I'm going to revert these altogether.

src/DDTrace/Transport/Http.php
```diff
@@ -68,10 +68,13 @@ private function configure($config)
...
-    public function send(array $traces)
+    /**
+     * {@inheritdoc}
+     */
+    public function send(Tracer $tracer)
     {
-        $tracesPayload = $this->encoder->encodeTraces($traces);
-        self::logDebug('About to send to the agent {count} traces', ['count' => count($traces)]);
+        $tracesPayload = $this->encoder->encodeTraces($tracer);
+        self::logDebug('About to send trace(s) to the agent');
```
labbati: One thing I don't get is how we replicate the old behavior. Before if we had "unfinished" traces not sent to the agent. How can we replicate this?

src/DDTrace/Encoders/MessagePack.php
```diff
@@ -0,0 +1,33 @@
...
+        }
+        return $messagePack;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getContentType()
+    {
+        return 'application/msgpack';
```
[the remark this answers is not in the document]
SammyK: Yeah, I looked for it too and it's not in the spec. 😆 I just went with what the Agent looks for: https://github.com/DataDog/datadog-agent/blob/355a34d610bd1554572d7733454ac4af3acd89cd/pkg/trace/api/api.go#L420

src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -176,6 +176,12 @@ public static function load()
         self::get()->loadAll();
     }
 
+    public static function reload()
```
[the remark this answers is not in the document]
SammyK: Ah! Nice catch. It was left over when I was trying to reset the config state for the Curl tests. Removed. :)

src/DDTrace/Transport/Http.php
SammyK: I'm not sure which old functionality you're referring to with the unfinished spans count, but if you think we should be able to count the traces publicly, we could have `DDTrace\Contracts\Tracer` [implement `Countable`](https://www.php.net/countable) if you think that'd be a good idea. :)

SammyK: FTR, the main reason I didn't go this route is to not introduce another BC, but if you think it's important enough, we should totally add it now since the major version is still `0`. :)

src/ext/mpack/mpack.h
```diff
@@ -0,0 +1,7151 @@
+/**
```
pawelchcki: Would it make sense to also bundle other filfes from the amalgamation package?
```
AUTHORS.md
CHANGELOG.md
LICENSE
README.md
``` <- Those file would be a good reference to have even if the author doesn't require them to be included

File: src/ext/serializer.c [hunk 14]
```diff
@@ -0,0 +1,153 @@
...
+    }
+    // finish writing
+    if (mpack_writer_destroy(&writer) != mpack_ok) {
+        free(data);
+        return 0;
+    }
+#if PHP_VERSION_ID < 70000
+    ZVAL_STRINGL(retval, data, size, 1);
+#else
+    ZVAL_STRINGL(retval, data, size);
```
pawelchcki: This operation copies string value over. Which for a brief while doubles the memory use.
Would it be possible to create empty zend_string at the beginning of this function, and pass pointers to it char* and size ?

src/ext/mpack/mpack.h
SammyK: Great point! I'll pull these in! :)

labbati:
@SammyK 👋 can we proceed with the second round of CR or still changes to be pushed? Ideally if we can merge it by EOD today it would be great.

src/ext/serializer.c
SammyK: Unless I misunderstood your suggestion, I don't think this is possible to do. I spent a lot of time digging for a solution and when I came up empty, I asked on Room 11 and [Niki said it wasn't possible](https://chat.stackoverflow.com/transcript/message/45874560#45874560). Or did I just misunderstand what you were suggesting? :)

pawelchcki: Looks like with "growable" approach its indeed not possible due to how PHP stores zend_string :( 

Thanks for looking into it!

src/DDTrace/Transport/Http.php
labbati: Nevermind. This is done in `Tracer:getTracesAsArray()` 👍

PR #388, opened 2019-04-05
Add cap to payloads sent to the Agent

SammyK:
### Description

The Agent has a [10MB payload cap](https://github.com/DataDog/datadog-agent/blob/355a34d610bd1554572d7733454ac4af3acd89cd/pkg/trace/api/api.go#L31) which isn't an issue for us right now, but when/if we start buffering traces sent to the Agent, this will be an issue to account for. The main reason is that the Agent will return a successful 200 response for payloads over the 10MB limit and will drop all the traces quietly.

### Readiness checklist

### DIFF
--- a/src/DDTrace/Transport/Http.php
+++ b/src/DDTrace/Transport/Http.php
@@ -16,6 +16,9 @@ final class Http implements Transport
 
     // Env variables to configure trace agent. They will be moved to a configuration class once we implement it.
     const AGENT_HOST_ENV = 'DD_AGENT_HOST';
+    // The Agent has a payload cap of 10MB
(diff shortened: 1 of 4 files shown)


src/DDTrace/Transport/Http.php
```diff
@@ -96,14 +99,23 @@ public function getConfig()
 
     private function sendRequest($url, array $headers, $body)
     {
+        $bodySize = strlen($body);
```
labbati: Question: how is this working with msgpack binary protocol we are about to merge?

SammyK: Very good thing to check! The `strlen()` [returns the number of bytes in the string](https://www.php.net/strlen), not the number of characters like `mb_string()` so it is safe to use in this context since we're measuring total payload size vs number of characters in the payload. :)

PR #398, opened 2019-04-10
Improve span handling performance (setTag and serialization)

pawelchcki:
### Description

Use direct value access instead of getters and setters to improve the performance of span handling.

Additionally:
- ~cache EnvVariableRegistry variable conversion result~
- improve p50 setTag performance by using associative array to filter out edge cases. 

The performance measuring env graphs seem to be happy:
image

### Readiness checklist


src/DDTrace/Encoders/SpanEncoder.php
```diff
@@ -15,43 +16,43 @@ final class SpanEncoder
      * @param Span $span
      * @return array
      */
-    public static function encode(Span $span)
+    public static function encode(SpanData $span)
```
SammyK: ❤️

src/DDTrace/Encoders/SpanEncoder.php
```diff
@@ -15,43 +16,43 @@ final class SpanEncoder
...
 
-        if ($span->getType() !== null) {
-            $arraySpan['type'] = $span->getType();
+        if ($span->type !== null) {
+            $arraySpan['type'] = $span->type;
         }
 
-        if ($span->isFinished()) {
-            $arraySpan['duration'] = (int) ($span->getDuration() . '000');
+        if ($span->duration != null) { // is span finished ?
```
SammyK: I'd do a strict check here. :)

```suggestion
        if ($span->duration !== null) { // is span finished ?
```

src/DDTrace/Processing/TraceAnalyticsProcessor.php
```diff
@@ -13,33 +13,48 @@
  */
 final class TraceAnalyticsProcessor
 {
+    /**
+    * @param bool|float $value
+    */
+    public static function normalizeAnalyticsValue(&$array, $value)
```
SammyK: I think I'd call this `$metrics` and update the DocBlock. :)

```suggestion
    public static function normalizeAnalyticsValue(&$metrics, $value)
```

src/DDTrace/Span.php
```diff
@@ -4,139 +4,71 @@
 
 use DDTrace\Integrations\Integration;
 use DDTrace\Contracts\Span as SpanInterface;
```
SammyK: We can remove this :)

src/DDTrace/Span.php
```diff
@@ -4,139 +4,71 @@
...
+use DDTrace\Data\Span as SpanData;
+
 use DDTrace\Contracts\SpanContext as SpanContextInterface;
```
SammyK: We can remove this line as well as the new line above it :)

src/DDTrace/Span.php
```diff
@@ -4,139 +4,71 @@
...
 use DDTrace\Exceptions\InvalidSpanArgument;
+use DDTrace\SpanContext as SpanContext;
```
SammyK: This is redundant so we can remove this as well :)

src/DDTrace/Span.php
```diff
@@ -4,139 +4,71 @@
...
-    /**
-     * @var Integration
-     */
-    private $integration = null;
-
-    /**
-     * @var bool Whether or not this trace can be even considered for trace analytics automatic configuration.
-     */
-    private $isTraceAnalyticsCandidate = false;
+    private static $METRIC_NAMES = [ Tag::ANALYTICS_KEY => true ];
```
SammyK: Just a nitpick. I know these kind of act more like constants, hence the name in ALL CAPS and snake_case, but I'd still have this follow how other class property names using camelCase. And I'd do the same for `$SPECIAL_TAGS` below. Again - that's just my opinion, you don't have to change anything. :)

src/DDTrace/Span.php
```diff
@@ -192,48 +124,49 @@ public function getDuration()
...
-
         if ($key !== (string)$key) {
             throw InvalidSpanArgument::forTagKey($key);
         }
 
-        if ($key === Tag::ERROR) {
-            $this->setError($value);
-            return;
-        }
+        if (array_key_exists($key, self::$SPECIAL_TAGS)) {
```
SammyK: 🔝

src/DDTrace/SpanContext.php
```diff
@@ -174,6 +126,13 @@ public function getAllBaggageItems()
 
     public function isEqual(SpanContextInterface $spanContext)
     {
+        if (is_subclass_of($spanContext, "DDTrace\Data\SpanContext")) {
```
SammyK: We can remove a function invocation here. :)

```suggestion
        if ($spanContext instanceof SpanContextData) {
```

src/DDTrace/SpanContext.php
```diff
@@ -174,6 +126,13 @@ public function getAllBaggageItems()
...
+            return $this->traceId == $spanContext->traceId
```
SammyK: Just because PHP's type coercion can get squirrely. :)

```suggestion
            return $this->traceId === $spanContext->traceId
```

src/DDTrace/Tracer.php
```diff
@@ -19,7 +19,10 @@
 use DDTrace\Contracts\Scope as ScopeInterface;
 use DDTrace\Contracts\Span as SpanInterface;
 use DDTrace\Contracts\SpanContext as SpanContextInterface;
+use DDTrace\Data\Span as SpanData;
+use DDTrace\Data\SpanContext as SpanContextData;
```
SammyK: We can remove these two lines and the `use DDTrace\Time;` line below. :)

src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -8,6 +8,11 @@
  */
 class EnvVariableRegistry implements Registry
 {
+    /**
+     * @var array
+     */
+    private $keyTranslator = [];
```
labbati: I am actually quite surprised that this brings a noticeable performance improvement. The reason for saying this is that this registry is a very low level one that is only called once per key (at least in theory). In fact, the upstream classes cache the results. Not saying we should not so it (actually some values are not cached on purpose) but still I am surprised :)

src/DDTrace/Data/Span.php
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Resource is a query to a service. A web application might use
+     * resources like "/user/{user_id}". A sql database might use resources
+     * like "select * from user where id = ?".
+     *
+     * You can track thousands of resources (not millions or billions) so
+     * prefer normalized resources like "/user/{id}" to "/user/123".
+     *
+     * Resources should only be set on an app's top level spans.
```
labbati: > Resources should only be set on an app's top level spans.

Not sure about this sentence, there are non-top level spans that are perfectly fit for a resource, e.q. `sql.query`. Wouldn't you agree?

src/DDTrace/Data/Span.php
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @var string
+     */
+    public $resource;
+
+    /**
+     * Service is the name of the process doing a particular job. Some
+     * examples might be "user-database" or "datadog-web-app". Services
+     * will be inherited from parents, so only set this in your app's
+     * top level span.
```
labbati: Again I am not sure that this is the spirit of the service. Why should a PDO span inherit the service name `payments-api` instead of having the service `pdo`?

src/DDTrace/Data/SpanContext.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace DDTrace\Data;
+
+use DDTrace\Contracts\SpanContext as SpanContextInterface;
+
+abstract class SpanContext implements SpanContextInterface
+{
+    /**
+     * The unique integer (64-bit unsigned) ID of the trace containing this span.
+     * It is stored in hexadecimal representation.
```
labbati: This is unsigned 63 bits: [link](https://github.com/DataDog/dd-trace-php/blob/1665e91a9ed9717f2c5906c9a11e27981620a862/src/DDTrace/ID.php#L12).

Also, I don't think that this is stored as hexadecimal. The [generate id](https://github.com/DataDog/dd-trace-php/blob/1665e91a9ed9717f2c5906c9a11e27981620a862/src/DDTrace/ID.php#L12) is the string representation of a 63-bit unsigned integer number `"1234567"` and it is [stored as it is](https://github.com/DataDog/dd-trace-php/blob/1665e91a9ed9717f2c5906c9a11e27981620a862/src/DDTrace/SpanContext.php#L37)

src/DDTrace/Data/SpanContext.php
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @var string
+     */
+    public $traceId;
+
+    /**
+     * The span integer (64-bit unsigned) ID.
```
labbati: Same as above

src/DDTrace/Data/SpanContext.php
```diff
@@ -0,0 +1,70 @@
...
+     * It is stored in hexadecimal representation.
+     *
+     * @var string
+     */
+    public $spanId;
+
+    /**
+     * The span integer ID of the parent span.
+     * It is stored in hexadecimal representation.
```
labbati: Same as above

src/DDTrace/Encoders/SpanEncoder.php
```diff
@@ -15,43 +16,43 @@ final class SpanEncoder
      * @param Span $span
```
labbati: This is `SpanData` now

src/DDTrace/Configuration/EnvVariableRegistry.php
```diff
@@ -35,7 +40,15 @@ public function __construct($prefix = 'DD_')
...
-        $value = getenv($this->convertKeyToEnvVariableName($key));
+        if (array_key_exists($key, $this->keyTranslator)) {
+            $key = $this->keyTranslator[$key];
+        } else {
+            $new_key = $this->convertKeyToEnvVariableName($key);
+            $this->keyTranslator[$key] = $new_key;
+            $key = $new_key;
+        }
+
+        $value = getenv($key);
```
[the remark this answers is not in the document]
pawelchcki: Agreed its much cleaner . Seems my performance hat squeezed my head maybe a little too tight :)
And I wrote it like that to avoid extra array lookup :D.
However it does look like my version produces 5 more opcodes to process. I think going with your approach with one extra array lookup is less expensive than processing more opcodes. And cleaner PHP too :)

src/DDTrace/Processing/TraceAnalyticsProcessor.php
```diff
@@ -13,33 +13,48 @@
...
+    {
+        if (true === $value || null === $value) {
```
labbati: Why should we convert `null` to `100%`? If the reason is, as how I understand it, to [remove this](https://github.com/DataDog/dd-trace-php/pull/398/files#diff-7952f110894c1eceb5170b6255a3550dL42) then I believe we should instead set the metric over there and remove this risky `null === $value` which looks a little bit counter intuitive to me. WDYT?

src/DDTrace/Configuration/EnvVariableRegistry.php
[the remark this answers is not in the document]
pawelchcki: Actually Just noticed this key translation is actually cached already :) by all the (bool|string|...l)Value thats actually used to access the values. 

Probably the only way to speed it up would be to use hashmap thats shared between requests.

src/DDTrace/Span.php
pawelchcki: Conventions are there to be broken 🙈 
Yeah it was a `const` but 5.4 made me change it to this :(.

I'll follow your suggestion though

src/DDTrace/Configuration/EnvVariableRegistry.php
pawelchcki: Indeed. I've seen on the Profiler it being shown quite high. TBF I haven't checked this after changes if it improved the performance significantly. I've since now removed it when I noticed the boolValue et. al does indeed cache it. https://github.com/DataDog/dd-trace-php/pull/398#discussion_r274365118

src/DDTrace/Data/Span.php
pawelchcki: Agreed!

src/DDTrace/Data/Span.php
pawelchcki: 👍

src/DDTrace/Data/SpanContext.php
pawelchcki: 👍

src/DDTrace/Processing/TraceAnalyticsProcessor.php
pawelchcki: I don't know ;) I only copy pasted it here :D 

Let me take a look :)

pawelchcki: I agree the `null === $value` looks weird, let me remvoe it and see what breaks.

pawelchcki:
Thanks @labbati and @SammyK for the reviews!
lets :shipit:

PR #401, opened 2019-04-12
increase composer memory limit to infinite

pawelchcki:
### Readiness checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,6 +171,7 @@ executors:
   with_integrations:
     environment:
       DDAGENT_HOSTNAME: 127.0.0.1
+      COMPOSER_MEMORY_LIMIT: -1 # disable composer memory limit completely
     parameters:
       tag:
         type: string


.circleci/config.yml
```diff
@@ -171,6 +171,7 @@ executors:
   with_integrations:
     environment:
       DDAGENT_HOSTNAME: 127.0.0.1
+      COMPOSER_MEMORY_LIMIT: -1 # disable composer memory limit completely
```
labbati: I am probably being over concerned here, but given that we are also running our tests via composer wouldn't safer to just change the memory limit on the composer update command? This way all the other invocations involving composer can keep failing in case of excessive memory.

Probably this is useless though, is it exceed the default value of 128M, then probably will exceed also the physical limit of 1.5 GB.

pawelchcki: Composer spawns its dependencies as subprocesses. And seems to not pass along the memory limit change.
I just checked by modifying `phpunit` to only run `phpinfo(); die();`
And the output is this:

```
circleci@b987d7790a8c:~/app$ COMPOSER_MEMORY_LIMIT=-1 composer test | grep memory
> phpunit --colors=always --configuration=phpunit.xml
memory_limit => 128M => 128M
Collecting memory statistics => No
```

labbati: Awesome, then this is the perfect solution. Thanks for verifying this 👍

pawelchcki: So If I understood your concerns correctly here - we should still get info if any of our tests cross the default limit of 128M

PR #404, opened 2019-04-13
Fix original curl headers overwritten

### LINKED ISSUE #365: CURL headers are overwritten in v0.15.1
Hola. Wanted to report this bug which blocks us actually.
Using php 7.1

When I make curl request, datadog seems to intercept headers and rewrite them to its own
In our case its host that is important

Screen Shot 2019-03-15 at 15 32 41

SammyK:
### Description

This PR fixes #365 where the original cURL headers set with `curl_setopt_array()` were overwritten by the tracer. Thanks so much to @labbati for [finding the repro](https://github.com/DataDog/dd-trace-php/tree/curl/missing-headers/tests/ManualTesting/CurlMissingHttpHeaders)! :)

I'll include more details on Monday. :)

### Readiness checklist


pawelchcki:
Awesome stuff @SammyK - thanks for fixing this one!

PR #409, opened 2019-04-17
Force tracing of whole trace via special Span tags (manual.keep and manual.drop)

pawelchcki:
### Readiness checklist


pawelchcki:
note: setTag currently performs a lot of string operations. Running simple benchmarks to see if "string" based dispatch would be faster than a host of ifs.

pawelchcki:
Benchmarks still shows a host of ifs is much faster than string based function dispatch. even in worst case (last tag on the list), it has only comparable performance and still no clear performance advantage

PR #417, opened 2019-04-30
Implementation of `limited` tracer state in which most spans should not be created 

pawelchcki:
### Description

Tracer will enter `limited()` state after creating 1000 (configurable number) spans. In limited state most integrations should stop creating new spans while maintaining all other operation like distributed tracing, etc. 

Note to reviewers:
I've changed the concept of `ready()` to its own reverse `limited()` somehow it felt more natural this way.

It still means a ton of Ifs in the code needed to be added. And I added them for all integrations + tests.

~Trying to explore some concepts how to best implement optional traces. This skeleton PR is my best idea so far :D (it's probably not objectively best so open to all and every feedback) 
/cc: @SammyK @labbati~

~I've considered changing existing api to return null spans or adding higher level abstractions. And this felt the cleanest.~

### Readiness checklist

### DIFF
--- a/src/DDTrace/Integrations/Curl/CurlIntegration.php
+++ b/src/DDTrace/Integrations/Curl/CurlIntegration.php
@@ -44,9 +44,9 @@ class CurlIntegration extends Integration
 
         dd_trace('curl_exec', function ($ch) use ($integration) {
             $tracer = GlobalTracer::get();
-            CurlIntegration::injectDistributedTracingHeaders($ch);
-
             if ($tracer->limited()) {
+                CurlIntegration::injectDistributedTracingHeaders($ch);
+
                 return dd_trace_forward_call();
             }
 
@@ -55,6 +55,7 @@ class CurlIntegration extends Integration
             $span->setTraceAnalyticsCandidate();
             $span->setTag(Tag::SERVICE_NAME, 'curl');
             $span->setTag(Tag::SPAN_TYPE, Type::HTTP_CLIENT);
+            CurlIntegration::injectDistributedTracingHeaders($ch);
 
             $result = dd_trace_forward_call();
             if ($result === false && $span instanceof Span) {
@@ -118,10 +119,13 @@ class CurlIntegration extends Integration
         $httpHeaders = ArrayKVStore::getForResource($ch, Format::CURL_HTTP_HEADERS, []);
         if (is_array($httpHeaders)) {
             $tracer = GlobalTracer::get();
-            $context = $tracer->getActiveSpan()->getContext();
-            $tracer->inject($context, Format::CURL_HTTP_HEADERS, $httpHeaders);
+            $activeSpan = $tracer->getActiveSpan();
+            if ($activeSpan !== null) {
+                $context = $activeSpan->getContext();
+                $tracer->inject($context, Format::CURL_HTTP_HEADERS, $httpHeaders);
 
-            curl_setopt($ch, CURLOPT_HTTPHEADER, $httpHeaders);
+                curl_setopt($ch, CURLOPT_HTTPHEADER, $httpHeaders);
+            }
         }
     }
 }
--- a/src/DDTrace/Integrations/Curl/CurlIntegration.php
+++ b/src/DDTrace/Integrations/Curl/CurlIntegration.php
@@ -44,13 +44,17 @@ class CurlIntegration extends Integration
 
         dd_trace('curl_exec', function ($ch) use ($integration) {
             $tracer = GlobalTracer::get();
+            CurlIntegration::injectDistributedTracingHeaders($ch);
+
+            if ($tracer->limited()) {
+                return dd_trace_forward_call();
+            }
+
             $scope = $tracer->startIntegrationScopeAndSpan($integration, 'curl_exec');
             $span = $scope->getSpan();
             $span->setTraceAnalyticsCandidate();
             $span->setTag(Tag::SERVICE_NAME, 'curl');
             $span->setTag(Tag::SPAN_TYPE, Type::HTTP_CLIENT);
-
-            CurlIntegration::injectDistributedTracingHeaders($ch);
 
             $result = dd_trace_forward_call();
             if ($result === false && $span instanceof Span) {
--- a/src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
+++ b/src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
@@ -60,64 +60,111 @@ class MemcachedIntegration extends Integration
 
         // bool Memcached::add ( string $key , mixed $value [, int $expiration ] )
         dd_trace('Memcached', 'add', function () {
+            if (GlobalTracer::get()->limited()) {
+                return dd_trace_forward_call();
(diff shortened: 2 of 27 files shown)


SammyK:
I like where the idea is going! So would `$tracer->ready()` check against a hard limit on number of spans/span size? :)

pawelchcki:
> I like where the idea is going! So would `$tracer->ready()` check against a hard limit on number of spans/span size? :)

Yes, although that is somewhat implementation specific :) i.e. it could mean other things as well. Here its mostly meant that the tracer is fully ready to create more spans. 

I might change the `ready()` to something else before the final PR, I'm not 100% happy with it.

Thanks for looking into it @SammyK I'll proceed with this concept and lets lets how it will turn out in the final PR.

SammyK:
Sounds good! So if I understand correctly, given the current circumstances (max spans/size), whenever `$tracer->ready()` returns `false` for the first time, it will never return `true` for the rest of the request. I'm curious if for now if we wanted to do something like...

```php
if (!$tracer->ready()) {
    dd_trace_disable_in_request();
    return dd_trace_forward_call();
}
```

...to disable span creation for the rest of the request. And then once we have a use case in the future where `$tracer->ready()` returns `true` after it has returned `false` at least once, then we can come up with a check for every span creation. Did I understand that correctly? :)

pawelchcki:
> ...to disable span creation for the rest of the request. And then once we have a use case in the future where `$tracer->ready()` returns `true` after it has returned `false` at least once, then we can come up with a check for every span creation. Did I understand that correctly? :)

at this moment with span limit for current trace/request  only I think this statement its true. Once its goes `false` it never returns to `true`. I presume though it might not be the case in future.

As for `dd_trace_disable_in_request()`, we do need to still trace curl calls to be able to add distributed tracing headers :/ - I was thinking too about using this but it seems its not possible to do so in its current form.

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -44,12 +44,17 @@ public static function load()
 
         dd_trace('curl_exec', function ($ch) use ($integration) {
             $tracer = GlobalTracer::get();
+            if ($tracer->limited()) {
+                CurlIntegration::injectDistributedTracingHeaders($ch);
```
labbati: What do you think of moving `CurlIntegration::injectDistributedTracingHeaders($ch);` as the first line so we can avoid repetition between this call and the one on line 58?

src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
```diff
@@ -60,70 +60,123 @@ public static function load()
 
         // bool Memcached::add ( string $key , mixed $value [, int $expiration ] )
         dd_trace('Memcached', 'add', function () {
+            if (GlobalTracer::get()->limited()) {
+                return dd_trace_forward_call();
```
labbati: Not sure if `dd_trace_forward_call()` would work when called from `traceCommand` or `traceCommandByKey`. If yes then we may want to move this check over there? If not we should create a task to make sure that this can be done in the future as it seems pretty useful. WDYT?

labbati: Ok probably

> I've considered changing existing api to return null spans

seems the answer to my question 😄

pawelchcki: Actually when I first implemented it it was called like that. But there is a limitation of dd_trace_forward_call() that it only works from within the closure.

Overall I think we could change the traceCommandByKey API it a bit to make that possible.
For now more pragmatic way was to just copy paste this tidbit.  In general however I would see us move away from using `dd_trace` directly.

src/DDTrace/Integrations/Curl/CurlIntegration.php
pawelchcki: I considered that, and looking at implementation of:
```
   CurlIntegration::injectDistributedTracingHeaders($ch);
```

It takes current active span. Which means this needs to be called in two different places.

labbati: Good point! 👍 resolving then

src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
labbati: How does this wrk then? https://github.com/DataDog/dd-trace-php/pull/417/files#diff-5ed8e1b32015797d9cf19b5ce8f354b9R53

labbati: How does this work then? https://github.com/DataDog/dd-trace-php/pull/417/files#diff-5ed8e1b32015797d9cf19b5ce8f354b9R53

pawelchcki: @labbati because of 
https://github.com/DataDog/dd-trace-php/pull/417/files#diff-5ed8e1b32015797d9cf19b5ce8f354b9R43 :D (i.e. it runs directly within dd_trace closure)

labbati: I meant the underlying root cause......and the underlying root cause is.....ME :palmface:

pawelchcki:
>Overall I have one question (I should probably have asked earlier on). Can we think of any pro and cons in doing the check internally in the tracer's startSpan() and returning:
>* a real span (registered in the tracer) when below the limit
>* a noop span (not registered in the tracer) when above the limit

In short my conclusion was:
Noop spans, trade in performance for nicer code quality and backwards compatibility.

Null spans trade in some performance, and backward compatibility for somewhat nicer code. I.e. the check for `if $span === null` would happen a little bit later, and its more specific. I also did not feel like adding yet another method to create spans (we should actually remove most of them :D ).

`->limited()` felt like trading in some code quality (due to repetition, which could be solved by adding an abstraction over more `dd_trace` uses), for best performance, and backwards compatibility (nothing changes without adding check for `->limited()` 

Since our focus was on performance the `Noop::span` did not get much airtime. The `null` span - someone once said that `null` is a billion dollar mistake :).  So ~`->ready()`~/ `->limited()` won

src/DDTrace/Integrations/Memcached/MemcachedIntegration.php
pawelchcki: :) Overall I think we could make dd_trace_forward_call call work from whatever context, but its a C level change, and I'm worried about unexpected consequences. 
And I liked our hands to be tied a little here, due to performance, we shouldn't do a lot of calls/allocations/objects from within the closure. 

Also it makes stacktraces make more sense if everything is happening within dd_trace.

I think following existing examples of `tracePublicMethod` and some functional programming concepts we could make nice idempotent functions passed to dd_trace flat and uber-fast.

labbati: 👍 totally agree.

pawelchcki:
@labbati fixed the 5.4 tests in guzzle. 
Feel free to give it another look once all tests pass

Issue #332, opened 2019-02-27
PDO integration throws exception when using Propel


mruoss:
Propel wraps `PDO` with its own class `PdoConnection`. `PdoConnection` again wraps some calls like `exec()` with its own function and returns an object instead of an int (what `PDO::exec()` returns). 

see 
https://github.com/propelorm/Propel2/blob/master/src/Propel/Runtime/Connection/PdoConnection.php#L115

Now the DDTrace PDO integration traces calls to `PDO::exec()` sets the tag `db.rowcount` to the result of the call. Since the result of `PdoConnection::exec()` returns an object, this step throws an exception because the value passed to `setTag` is an object, not an int.

Is it possible for DDTrace to wrap only the calls to PDO::exec, not the one to `PdoConnection::exec()`?

labbati:
Hi @mruoss and sorry for the late reply. I think this is connected to an issue we have with `parent::` invocation in some specific circumstances and [here it is](https://github.com/propelorm/Propel2/blob/6c4996fc130c7555d82d9dca04396c6f3eeef306/src/Propel/Runtime/Connection/PdoConnection.php#L117).

We are working hard at the moment on this as our top priority and it should be released in the next path release.

We will update this issue once released.

SammyK:
Hey @mruoss! I just wanted to give an update on this. I was able to reproduce your issue and it seems that it is a separate issue from the `parent::` bug. We'll certainly try to get a fix for this on the radar. :)

@pawelchcki & @labbati - Here's a small repro that illustrates the underlying issue:

```php
<?php

class FakePDO
{
    public function exec()
    {
        return 42;
    }
}

class MyPDO extends FakePDO
{
    public function exec()
    {
        $obj = new \stdClass;
        $obj->rows = parent::exec();
        return $obj;
    }
}

dd_trace('FakePDO', 'exec', function() {
    var_dump($this->exec());
});

(new MyPDO())->exec();
```

Since the subclass essentially changes the return type of `exec()` from `int` to `stdClass`, this causes an issue in the integrations (specifically when [setting a tag value](https://github.com/DataDog/dd-trace-php/blob/a8de1d4238a773f03576c8e33fbaf9aebdd64b13/src/DDTrace/Span.php#L186)) when we assume the return values can be cast to strings. So I'm curious your thoughts - perhaps if the value sent to `setTag()` is an object we could check for the existence of the `__toString()` magic method? :)

labbati:
[...]

PR #423, opened 2019-04-30
Ignore objects sent to Span::setTag()

SammyK:
### Description

In cases where a subclass changes the return value of a known method, the `$value` sent to `Span::setTag()` can be an object that has no `__toString` magic method. This causes an error when trying to cast the value to a string. This is the root cause of #332.

The following repro illustrates what can happen within a traced integration:

```php
class FakePDO
{
    public function exec()
    {
        return 42;
    }
}

class MyPDO extends FakePDO
{
    public function exec()
    {
        $obj = new \stdClass;
        $obj->rows = parent::exec();
        return $obj;
    }
}

dd_trace('FakePDO', 'exec', function() {
    var_dump($this->exec()); // Assumes return value is an int
});

(new MyPDO())->exec();
```

### Readiness checklist

### DIFF
--- a/src/DDTrace/Span.php
+++ b/src/DDTrace/Span.php
@@ -134,6 +134,11 @@ final class Span extends SpanData
         if ($key !== (string)$key) {
             throw InvalidSpanArgument::forTagKey($key);
         }
+        // Since sub classes can change the return value of a known method,
+        // we quietly ignore values that could cause errors when converting to string
+        if (is_object($value) && $key !== Tag::ERROR) {
+            return;
+        }
 
         if (array_key_exists($key, self::$specialTags)) {
             if ($key === Tag::ERROR) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file - [read more
 ### Fixed
 - `parent::` keyword not honored from a subclass when forwarding a call from a tracing closure #284
 - Private and protected callable strings not resolved properly from a tracing closure #303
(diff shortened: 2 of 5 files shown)


src/DDTrace/Span.php
```diff
@@ -134,6 +134,11 @@ public function setTag($key, $value, $setIfFinished = false)
         if ($key !== (string)$key) {
             throw InvalidSpanArgument::forTagKey($key);
         }
+        // Since sub classes can change the return value of a known method,
+        // we quietly ignore values that could cause errors when converting to string
+        if (is_object($value) && $key !== Tag::ERROR) {
```
pawelchcki: On second thought - would it make sense to check if object has `__toString` method ? and only ingore it if does not ?

SammyK: That's what [I was suggesting](https://github.com/DataDog/dd-trace-php/issues/332#issuecomment-487464713) too! :)

But @labbati made a good point about potentially encountering huge strings with, for example, giant ORM models and similar objects. If we really wanted to, we could do a check for string size and truncate it if it's too big. But I'm also thinking of a scenario where the tracer encounters 1,000+ huge models that get converted to a string for the trace. Then we have a ton of string juggling & memory consumption gong on. And since none of our officially supported integrations return objects that we want to use as tags (with the exception of `Exception`'s), we'd be committing to support a feature which might make it even harder to debug segfaults in very specific edge cases.

What are your thoughts? :)

pawelchcki: Yeah that makes a lot of sense. Thanks for clarifying

PR #428, opened 2019-05-06
Ph redis integration

obi11235:
### Description

Initial draft of auto instrumentation integration for the redis php module https://pecl.php.net/package/redis

### Readiness checklist


obi11235:
I have not updated the tests at all, but they may not pass at all if the php module https://pecl.php.net/package/redis is not installed.

obi11235:
Any recommendations on a path forward?

labbati:
Hey @obi11235 thanks so much for opening this PR. It is great to see contributions coming from the community. We will look into this asap, possibly within this week.

obi11235:
Thanks, The other thing i noticed but have not dug into yet, (i'm setting our our environment so that i can use my branch instead of the apt provided php code), is that when using multi() and exec() i don't see any of the calls between captured. it may be called like:

```
$ret = $redis->multi()
    ->set('key1', 'val1')
    ->get('key1')
    ->set('key2', 'val2')
    ->get('key2')
    ->exec();
```

but i've not yet dug into our code to see if that is the case or not.

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+        return self::NAME;
+    }
+
+    /**
+     * Static method to add instrumentation to the Redis library
+     */
+    public static function load()
+    {
+        if (!extension_loaded('redis')) {
+            // `curl` extension is not loaded, if it does not exists we can return this integration as
```
SammyK:
```suggestion
            // `redis` extension is not loaded, if it does not exists we can return this integration as
```

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+const CMD_MAX_LEN = 1000;
+
+class RedisIntegration extends Integration
+{
+    const NAME = 'redis';
+
+    /**
+     * @var array
+     */
+    private static $connections = [];
```
SammyK: I think we can remove this as it does not appear to be used anywhere. :)

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+
+namespace DDTrace\Integrations\Redis;
+
+use DDTrace\Integrations\Integration;
+use DDTrace\Tag;
+use DDTrace\Type;
+use DDTrace\GlobalTracer;
+use DDTrace\Util\TryCatchFinally;
+
+const VALUE_PLACEHOLDER = "?";
```
SammyK: I don't see where these constants are being used anywhere. I think they can be removed. :)

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+            );
+            $span = $scope->getSpan();
+            $span->setTag(Tag::SPAN_TYPE, Type::CACHE);
+            $span->setTag(Tag::SERVICE_NAME, 'redis');
+            $span->setTag(Tag::RESOURCE_NAME, 'Redis.connect');
+
+
+            $thrown = null;
+            try {
+                $result = call_user_func_array([$this, 'connect'], $args);
```
SammyK: We have a function for forwarding calls in a more stable manner which we can use here. :)

```suggestion
                $result = dd_trace_forward_call();
```

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+            } catch (\Exception $e) {
+                RedisIntegration::setErrorOnException($span, $e);
```
SammyK: This method does not appear to exist. :)

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+                $thrown = $e;
+            }
+            $scope->close();
+
+            if ($thrown) {
+                throw $thrown;
+            }
+
+            return $result;
```
SammyK: If we rely on the **try_catch_finally.php** file we have, we can delete all the code after the last `setTag()` above and replace this return with:

```suggestion
            return include __DIR__ . '/../../try_catch_finally.php';
```

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+        });
+
+        dd_trace('Redis', 'getSet', function () {
```
SammyK: We typically don't trace all the possible methods from an extension. Instead we try to only trace methods that perform I/O operations or are otherwise "slow" relative to the other methods. Would you be able to whittle this list of methods down to just the ones that are typically slower for this extension? :)

src/DDTrace/Integrations/Redis/RedisIntegration.php
```diff
@@ -0,0 +1,835 @@
...
+        $span->setTag('redis.command', $command);
+
+        self::setServerTags($span, $redis);
+
+        $span->setTag(Tag::RESOURCE_NAME, $command);
+
+        return TryCatchFinally::executePublicMethod($scope, $redis, $command, $args);
+    }
+
+    private static function setServerTags($span, $redis)
```
SammyK: Since this is only being used in one place above, I would suggest moving these two `setTag()` calls to line `823` and remove this method. :)

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file - [read more
 
 ### Fixed
 - Error when a subclassed integration returns an object that cannot be cast as a string #423
+- initial redis auto instrumentation integration (https://pecl.php.net/package/redis)
```
SammyK: I would move this under the `## Added` header and change the copy to the following:

```suggestion
- Support for [ext/redis](https://pecl.php.net/package/redis) #428
```

nesl247:
Hi, is there any update on this?

nesl247:
Looks like #948 replaces this.

morrisonlevi:
As of version 0.48 we have Predis and PhpRedis integrations built-in.

PR #432, opened 2019-05-07
Use DD_TRACE_APP_NAME instead of DD_TRACE_APP_NAME to configure service name

labbati:
This is done for consistency with other tracers, e.g. Java tracer
configuration.


tests/Unit/ConfigurationTest.php
```diff
@@ -100,6 +100,26 @@ public function testAppNameFallbackPriorities()
...
+        putenv('DD_SERVICE_NAME=my_app');
+        $this->assertSame('my_app', Configuration::get()->appName('my_app'));
+    }
+
+    public function testServiceNameHasPrecedenceOverDeprecatedMethods()
+    {
+        Configuration::clear();
+
+        putenv('DD_SERVICE_NAME=my_app');
+        putenv('DD_TRACE_APP_NAME=wrong_app');
```
SammyK: ❤️

src/DDTrace/Configuration.php
```diff
@@ -125,10 +125,18 @@ public function getGlobalTags()
      */
     public function appName($default = '')
     {
+        // Using the env `DD_SERVICE_NAME` for consistency with other tracers.
+        $appName = $this->stringValue('service.name');
+        if ($appName) {
+            return $appName;
+        }
+
+        // This is deprecated and will be removed in a future release
```
SammyK: Thoughts on logging a deprecation notice if this env var is set? :)

tests/Unit/ConfigurationTest.php
```diff
@@ -100,6 +100,26 @@ public function testAppNameFallbackPriorities()
         $this->assertSame('bar_app', Configuration::get()->appName());
     }
 
+    public function testServiceName()
+    {
+        Configuration::clear();
```
SammyK: I think we need to reset the fallback env var before this test so that other test env vars don't leak into this one. :)

```suggestion
        putenv('DD_TRACE_APP_NAME');
        Configuration::clear();
```

src/DDTrace/Configuration.php
labbati: If you agree with this approach, I would suggest to add a huge header in the release notes for users to be notified and in 2/3 release to issue a deprecation warning to reduce the risk that logs will be flooded with this error logs. What do you think?

tests/Unit/ConfigurationTest.php
labbati: Good catch 👍

src/DDTrace/Configuration.php
SammyK: I was thinking of debug level log, but now that I think about it, no one will probably notice unless they are debugging on staging or locally. :) We could just focus on documenting it well here and in the docs if you think that's a good strategy.

tests/Unit/ConfigurationTest.php
labbati: @SammyK does the change ^^ work for you?

SammyK: @labbati Perrrrrfect! 🐈

labbati:
This should be good to go now! Let's 🤞

PR #436, opened 2019-05-10
CakePHP v2 integration

SammyK:
### Description

This PR adds tracing support for CakePHP v2 and also includes support for Cake Console.

### Readiness checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -115,13 +115,13 @@ services:
       - "8126:8126"
 
   request_replayer:
-    image: php:7.2
+    image: php:7.3
     working_dir: /var/www
+    volumes:
+      - ./tests/request_replayer:/var/www
     ports:
-      - 9000:80
-    environment:
-      DD_REQUEST_DUMPER_FILE: dump.json
-    command: sh -c "curl --output index.php https://raw.githubusercontent.com/DataDog/dd-trace-php/master/tests/request_replayer/index.php -q && php -S 0.0.0.0:80 index.php"
+      - 9090:80
+    command: sh -c "php -S 0.0.0.0:80 index.php"
 
   agent:
     image: datadog/agent:latest
--- a/src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
+++ b/src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
@@ -67,8 +67,9 @@ class CakePHPIntegration extends Integration
         // Web bootstrap
         dd_trace('Dispatcher', '__construct', $initCakeV2);
         // CLI bootstrap
+        // Once auto-instrumentation has been added for non-autoloaded projects,
+        // uncomment the following line and delete the if block below it
         //dd_trace('ShellDispatcher', 'run', $initCakeV2);
-        // Temporary workaround until we fix request_init_hook for non-autoloaded projects
         if ('cli' === PHP_SAPI) {
             dd_trace('App', 'init', $initCakeV2);
         }
--- a/src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
+++ b/src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace DDTrace\Integrations\CakePHP;
+
+use DDTrace\Configuration;
+use DDTrace\Integrations\CakePHP\V2\CakePHPIntegrationLoader;
+use DDTrace\Integrations\Integration;
+
+/**
+ * The base Laravel integration which delegates loading to the appropriate integration version.
+ */
+class CakePHPIntegration extends Integration
+{
+    const NAME = 'cakephp';
+
+    /**
+     * @var self
+     */
+    private static $instance;
+
+    /**
+     * @return self
+     */
+    public static function getInstance()
+    {
+        if (null === self::$instance) {
+            self::$instance = new self();
+        }
+        return self::$instance;
+    }
+
+    /**
+     * @return string The integration name.
+     */
+    public function getName()
+    {
+        return self::NAME;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function requiresExplicitTraceAnalyticsEnabling()
+    {
+        return false;
+    }
+
+    /**
+     * Loads the integration.
+     *
+     * @return int
+     */
+    public static function load()
+    {
+        if (!self::shouldLoad(self::NAME)) {
+            return self::NOT_AVAILABLE;
+        }
+
+        $integration = self::getInstance();
+
+        // CakePHP v2.x - we don't need to check for v3 since it does not have \Dispatcher or \ShellDispatcher
+        $initCakeV2 = function () use ($integration) {
+            $loader = new CakePHPIntegrationLoader();
+            $loader->load($integration);
+            return dd_trace_forward_call();
+        };
+        // Web bootstrap
+        dd_trace('Dispatcher', '__construct', $initCakeV2);
+        // CLI bootstrap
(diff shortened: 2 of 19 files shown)


docker-compose.yml
```diff
@@ -115,13 +115,13 @@ services:
...
     working_dir: /var/www
+    volumes:
+      - ./tests/request_replayer:/var/www
     ports:
-      - 9000:80
-    environment:
-      DD_REQUEST_DUMPER_FILE: dump.json
-    command: sh -c "curl --output index.php https://raw.githubusercontent.com/DataDog/dd-trace-php/master/tests/request_replayer/index.php -q && php -S 0.0.0.0:80 index.php"
+      - 9090:80
+    command: sh -c "php -S 0.0.0.0:80 index.php"
```
pawelchcki: ❤️

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
```diff
@@ -0,0 +1,89 @@
...
+        // CakePHP v2.x - we don't need to check for v3 since it does not have \Dispatcher or \ShellDispatcher
+        $initCakeV2 = function () use ($integration) {
+            $loader = new CakePHPIntegrationLoader();
+            $loader->load($integration);
+            return dd_trace_forward_call();
+        };
+        // Web bootstrap
+        dd_trace('Dispatcher', '__construct', $initCakeV2);
+        // CLI bootstrap
+        //dd_trace('ShellDispatcher', 'run', $initCakeV2);
```
pawelchcki: Should this be still commented out ?

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
```diff
@@ -0,0 +1,89 @@
...
+        // Temporary workaround until we fix request_init_hook for non-autoloaded projects
```
pawelchcki: This temporary might be quite permanent :) 
I would drop that comment for now, but of course, we should discuss alternatives :D

src/DDTrace/Integrations/CakePHP/V2/CakePHPIntegrationLoader.php
```diff
@@ -0,0 +1,103 @@
...
+     * @var Span
+     */
+    public $rootSpan;
+
+    public function load(CakePHPIntegration $integration)
+    {
+        // Very strange workaround to get the integration to load in tests
+        // We need to find this bug in the CLI SAPI's built-in web server
+        // Once the bug is fixed we can remove this line block
+        if ('true' === getenv('DD_TEST_INTEGRATION')) {
```
pawelchcki: 🤔

src/DDTrace/Integrations/CakePHP/V2/CakePHPIntegrationLoader.php
```diff
@@ -0,0 +1,103 @@
...
+            echo ' ';
+        }
+        $this->rootSpan = GlobalTracer::get()->getRootScope()->getSpan();
+        // Overwrite the default web integration
+        $this->rootSpan->setIntegration($integration);
+        $this->rootSpan->setTraceAnalyticsCandidate();
+        $this->rootSpan->overwriteOperationName('cakephp.request');
+        $this->rootSpan->setTag(Tag::SERVICE_NAME, CakePHPIntegration::getAppName());
+
+        $self = $this;
```
pawelchcki: I know this is a pattern we do already use. But It looks to me like the readability might be improved if its renamed to `$integration` instead of `$self`- WDYT?

tests/Integrations/CLI/CakePHP/V2_8/CommonScenariosTest.php
```diff
@@ -0,0 +1,78 @@
...
+        ]);
+    }
+
+    public function testCommandWithArgument()
+    {
+        $traces = $this->getTracesFromCommand('command_list');
+
+        $this->assertSpans($traces, [
+            SpanAssertion::build(
+                'cakephp.console',
```
pawelchcki: Do we want to trace cake console by default?

tests/Integrations/CakePHP/V2_8/CommonScenariosTest.php
```diff
@@ -0,0 +1,111 @@
...
+                        'Errors/index.ctp'
+                    )->withExactTags([
+                        'cakephp.view' => 'Errors/index.ctp',
+                        'integration.name' => 'cakephp',
+                    ]),
+                ],
+            ]
+        );
+    }
+}
```
pawelchcki: 👍 I like this test very clear whats going on !

tests/request_replayer/index.php
```diff
@@ -1,30 +1,48 @@
...
 
-if (php_sapi_name() == 'cli-server') {
-    if ($_SERVER['REQUEST_URI'] == '/replay') {
-        if (file_exists($requestLog)) {
-            $value = file_get_contents($requestLog);
-            error_log("Returning value from $requestLog: " . print_r($value, true));
-            echo $value;
-        } else {
-            error_log("No value to replay in $requestLog");
+switch ($_SERVER['REQUEST_URI']) {
```
pawelchcki: Going bare ;)

tests/request_replayer/index.php
```diff
@@ -1,30 +1,48 @@
...
+    default:
         $value = json_encode([
             'uri' => $_SERVER['REQUEST_URI'],
             'headers' => getallheaders(),
             'body' => file_get_contents('php://input'),
         ]);
-        error_log("Dumping data in $requestLog: $value");
-        file_put_contents($requestLog, $value);
-    }
+        file_put_contents(REQUEST_LOG_FILE, $value);
```
pawelchcki: 👍

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
SammyK: Yes, those are the real entry points we need to trace in order to hook into early script execution. We'll need to comment those out once we have a plan in place for auto-instrumentation for non-autoloaded projects. :)

tests/Integrations/CLI/CakePHP/V2_8/CommonScenariosTest.php
SammyK: All CLI integrations are disabled by default but this will be enabled by setting `DD_TRACE_CLI_ENABLED=true`. :)

src/DDTrace/Integrations/CakePHP/V2/CakePHPIntegrationLoader.php
SammyK: Nice catch! `$integration` is already being used in that scope so I changed it to `$loader`. :)

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
SammyK: Yeah, that comment isn't super clear. I'd like to keep something there as a note to change it once we get auto-instrumentation rocking for non-autoloaded projects. So I've updated the comment to be a bit more clear. Does that work for you? :)

pawelchcki: What I meant more is that - I've seen a lot of comments in the code where `temporary` was still there 4+ years later. 

In itself saying something is temporary without having a good plan and timeline of getting rid of this temporary is just asking for a case like this. 

Its more a philosophical issue than engineering one. though :)

SammyK: Very good point! 😄

PR #437, opened 2019-05-13
Limit span creation based on memory limit

pawelchcki:
### Description

This PR uses `zend_memory_usage()` and  `memory_limit` setting (via `PG(memory_limit)` ) to calculate the memory use ceiling that will limit span creation when its reached. 

By default its set to 80% of memory limit. When memory limit is not set the span creation is unlimited.

The memory limit can be set separately from PHP usual limit, via env variable `DD_TRACE_MEMORY_LIMIT` which allows values like:
- 100M
- 10G
- 10% (which will be calculated from PHPs memory limit)

### Readiness checklist


src/ext/ddtrace_extra.c
```diff
@@ -0,0 +1,118 @@
...
+
+    if (DDTRACE_G(disable)) {
+        RETURN_BOOL(0);
+    }
+
+    RETURN_BOOL(1);
+}
+
+static zend_long get_memory_limit(TSRMLS_D) {
+    char *raw_memory_limit = ddtrace_get_c_string_config("DD_MEMORY_LIMIT");
```
SammyK: Is there any reason why we're using the `DD_` prefix here instead of the `DD_TRACE_` prefix? I don't know if we narrowed down a distinction so just wanted to double check. :)

src/ext/ddtrace_extra.c
```diff
@@ -0,0 +1,118 @@
+#ifdef HAVE_CONFIG_H
```
SammyK: What is the motivation for moving these functions out of **ddtrace.c**? The name of the file **ddtrace_extra.c** isn't obvious to me what it contains and all other PHP extensions I've seen in the wild keep all the PHP API stuff in their "dedicated to PHP" file. So unless there's really good reason to break this out, I'm -1 for this separate file. Just my 2 cents. But if we decide to move forward with this, we'll need to add this to the **package.xml** file. :)

src/ext/compatibility.h
```diff
@@ -0,0 +1,34 @@
+#ifndef DD_COMPATIBILITY_H
```
SammyK: ❤️ I like moving this stuff to its own file and the file name make total sense. Nice work! We just gotta add this to the **package.xml** file. :)

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file - [read more
 
 ## [Unreleased]
 
+### Added
+- Tracer limited mode, stopping span creation when memory use raises to 80% of current PHP memory limit #437
```
SammyK: Can we update this to mention the new function names and new env var to configure it? :)

tests/ext/check_memory_under_limit_default.phpt
```diff
@@ -0,0 +1,11 @@
+--TEST--
```
SammyK: These tests are 🔝! We just gotta add all of them to the **package.xml** file. :)

src/ext/ddtrace_extra.c
pawelchcki: Motivation is to not to have everything in one file :) 
There are many extensions that separate PHP things to different files (like uopz), php itself is doing this as well. 

My rationale here is to cutout most things not directly related to tracing to separate file. The `_extra` is only meant as a conversation starter :) - I'm entirely open to different naming/ separation schemes.

If you feel that its more important to keep everything in one file thats cool too. I Just didn't want to add yet another thing to file that contain so many components of the extension already.

CHANGELOG.md
pawelchcki: I was thinking we shouldn't advertise this at this moment too much. But if you feel its important to do so then I'll add these.

tests/ext/check_memory_under_limit_default.phpt
pawelchcki: Everytime I added new file I thought, remember to add it to package.xml - remember to add it to package.xml. 
Then I forgot :D

src/ext/ddtrace_extra.c
pawelchcki: Oh I haven't thought too much about it. `DD_TRACE_` does feel more canonical

src/ext/ddtrace_extra.c
SammyK: For sure! I'm all about separating things out into smaller files to make things much easier to maintain. My comment is more about how we break it up. To use the example of **ext/uopz**, I love how [the main **uopz.c** file](https://github.com/krakjoe/uopz/blob/master/uopz.c) has just the PHP API stuff in it - INI settings, PHP info, PHP_FUNCTION's, and ZPP. All the implementation details are in separate files.

That's what I'm advocating here - just keeping all the parts tying the extension into PHP in the **ddtrace.c** file so there's one "source of truth" for all the PHP API stuff. And then we make sure we've separating any specific implementation logic in separate files. What do you think? :)

CHANGELOG.md
SammyK: I see your point on this. I think the env var should definitely be documented but putting the functions here I'll leave up to you. For me, I always like to have a heads up if an extension is going to put more unnamespaced function definitions in the root symbol table - just to be aware of it. But like you said, we might not want to put that out there quite yet until we nail down a hard API for everything. :)

tests/ext/check_memory_under_limit_default.phpt
SammyK: Haha! I'm the same way! I think to myself: this time I won't forget. Then I totally forget. 😆

src/ext/ddtrace_extra.c
SammyK: Yeah, me neither so it got me thinking about why we're choosing one over the other... I think I hear an RFC idea... *jots down TODO*. :)

src/ext/ddtrace_extra.c
pawelchcki: Indeed, my memory was a bit fuzzy there it seems :D lets keep it in one file then - for now at least :)  its still quite manageable at this moment

CHANGELOG.md
pawelchcki: oh but all our functions are namespaced ;)  Using good old C type namespacing of prefixing everything with `dd_trace`. So clashes with existing functions should be minimal. Do you think us adding function to main namespace might have other sideffects ?

SammyK: Hehe. Yes, we are doing some good old-fashioned PHP-style namespacing. 😄I just like to be explicit in the documentation - especially when adding new symbols to the symbol table. That's all. 😺I'm OK leaving this out until we nail down a solid API.

Makefile
```diff
@@ -131,9 +131,10 @@ packages: .apk .rpm .deb .tar.gz
 	tar -zcf packages.tar.gz $(PACKAGES_BUILD_DIR)
 
 verify_pecl_file_definitions:
-	@for i in src/ext/*.c src/ext/*.h tests/ext/*.php*; do\
-		k=$$(basename $$i); \
-		grep -q $$k package.xml || ( echo missing $$k && exit 1); \
+	@for i in $(notdir $(C_FILES) $(TEST_FILES)); do\
```
SammyK: ❤️ Thanks for fixing this!

src/ext/memory_limit.c
```diff
@@ -0,0 +1,52 @@
+#ifdef HAVE_CONFIG_H
```
SammyK: I love how you've broken this up - this file name makes total sense! 🔝

PR #440, opened 2019-05-16
IPC based Circuit breaker allowing all currently running PHP processes to stop sending data to the agent 

pawelchcki:
### Description
This also persists the circuit breaker state between restarts so that if the PHP process is recycled it will inherit the CB state.

Its using `shm_open` to allocate shared memory object (on linux its an actual file in /dev/shm) base on path (path is built from `dd_trace_+VERSION_STRING`).

Then that shared memory object is mapped to local address space and accessed locally via atomic C11 functions, guaranteeing consistency when setting /incrementing/changing a value.

See:
https://en.wikipedia.org/wiki/Circuit_breaker_design_pattern

### Readiness checklist


src/ext/vendor_stdatomic.h
```diff
@@ -0,0 +1,325 @@
+/*
```
SammyK: Since this is the second lib we've vendored in, we should probably start considering were we put 3rd-party code. I put the MessagePack files in `src/ext/mpack` but I'm not 100% happy with that location. And dropping in a `vendor` folder could get confusing with Composer. So I'm open to ideas, but I'd love to consolidate all the vendered code. :)

Also - could we append this line to the `.gitattributes` to have this auto-collapse in diffs:

```
src/ext/vendor_stdatomic.h*     linguist-generated=true
```

If we throw all the vendored code into one directory, we can just always have it auto-collapse every time we add/update:

```
src/ext/vendored_code_here/*     linguist-generated=true
```

Anyway - I'd love to hear your thoughts. :)

src/DDTrace/Transport/Http.php
```diff
@@ -82,6 +82,11 @@ private function configure($config)
      */
     public function send(Tracer $tracer)
     {
+        if (function_exists('dd_tracer_circuit_breaker_can_try') && !dd_tracer_circuit_breaker_can_try()) {
```
SammyK: I'm guessing the `function_exists()` checks are just in case manually instrumented code from Composer gets out of sync with the installed extension? :)

src/ext/ddtrace.c
```diff
@@ -327,11 +328,64 @@ static PHP_FUNCTION(dd_trace_check_memory_under_limit) {
...
+static PHP_FUNCTION(dd_tracer_circuit_breaker_register_success) {
+    PHP5_UNUSED(return_value_used, this_ptr, return_value_ptr, ht TSRMLS_CC);
+    PHP7_UNUSED(execute_data);
+
+    dd_tracer_circuit_breaker_register_success();
+
+    RETURN_BOOL(1);
+}
+
+static PHP_FUNCTION(dd_tracer_circuit_breaker_open) {
```
SammyK: As far as I can tell, the only functions that are being used in userland are `dd_tracer_circuit_breaker_can_try()`, `dd_tracer_circuit_breaker_register_error()`, and `dd_tracer_circuit_breaker_register_success()`. So I'm curious if the other functions are just exposing an API for testing purposes? If so, I'm wondering if there's another solution that doesn't require adding additional functions in userland. Maybe only exposing them when compiling in debug mode or something? What are your thoughts? :)

tests/ext/circuit_breaker_retry_time.phpt
```diff
@@ -0,0 +1,52 @@
...
+putenv('DD_TRACE_AGENT_ATTEMPT_RETRY_TIME_MSEC=garbage0');
+echo 'garbage time set CAN_RETRY ' . (dd_tracer_circuit_breaker_can_try() ? 'true' : 'false') . PHP_EOL; // 5000msec default gets reinstated
+
+// we can shorten the time again
+putenv('DD_TRACE_AGENT_ATTEMPT_RETRY_TIME_MSEC=0');
+echo 'min time set CAN_RETRY ' . (dd_tracer_circuit_breaker_can_try() ? 'true' : 'false') . PHP_EOL; // should be able to retry
+
+putenv('DD_TRACE_AGENT_ATTEMPT_RETRY_TIME_MSEC=1000');
+echo '1 second retry set CAN_RETRY ' . (dd_tracer_circuit_breaker_can_try() ? 'true' : 'false') . PHP_EOL; // 1 second not yet passed so no retry for you
+sleep(1);
```
SammyK: Could we use `usleep()` here with milliseconds (and also on line 40) to have a shorter test run time? :)

package.xml
```diff
@@ -82,6 +85,9 @@
                     <file name="case_insensitive_method_hook.phpt" role="test" />
                     <file name="closure_accessing_outside_variables.phpt" role="test" />
                     <file name="closure_set_inside_object_methods.phpt" role="test" />
+                    <file name="circuit_breaker_low_level.phpt" role="test" />
```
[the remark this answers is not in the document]
pawelchcki: Now we have working linter check 😈

src/DDTrace/Transport/Http.php
pawelchcki: Actually yes :) 
I think we need to address this somehow more generally the more functions we'll add. For now since these are called only once per request its relatively fine - although definitely not pretty :|

src/ext/vendor_stdatomic.h
pawelchcki: Need to vendor in this file surprised me abit. Hopefully we'll be able to get rid of this once we'll be able to upgrade GCC in 7.3 build images 
`stdatomic.h` should be part of C11 standard but GCC lagged "a bit" in implementing it :(

pawelchcki: Basically I hope to get rid of `vendor_stdatomic.h` "soon" so the sorest it looks the higher chance we'll remember to remove it.

overall I don't mind having `vendor` folder. But I worry about being able to set include paths correctly so that the files don't need to be modified when they reference themselves under specifc paths.  "Practicality beats purity" etc :)

tests/ext/circuit_breaker_retry_time.phpt
pawelchcki: Hmm makes sense. I was a bit worried about creating a flaky test but reducing the timeouts a bit should be possible.

src/ext/ddtrace.c
pawelchcki: I wonder whats the downside keeping them in?
I would prefer to test as close to the final binary as possible so functions only exposed in tests would mean that we have two different binaries - which in turn could foster theoretical edge-cases which wouldn't be detected in tests.

Also those functions provide more access to the Circuit breaker and could be used to extend/ enhance the functioning a bit in the future.

src/ext/circuit_breaker.c
```diff
@@ -0,0 +1,137 @@
...
+    uint64_t current_time = current_timestamp_monotonic_usec();
+
+    return (last_failure_timestamp + get_retry_time_usec()) <= current_time;
+}
+
+void dd_tracer_circuit_breaker_register_error() {
+    prepare_cb();
+
+    atomic_fetch_add(&dd_trace_circuit_breaker->consecutive_failures, 1);
+    atomic_fetch_add(&dd_trace_circuit_breaker->total_failures, 1);
```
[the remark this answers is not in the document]
pawelchcki: I left in in since this is the data I would love to collect with runtime metrics. But right now it doesn't serve any purpose.

However I think actually might be good idea to add debug mode, where this data could be exposed.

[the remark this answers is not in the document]
SammyK: I ❤️ the idea of putting that at debug level. :)

src/ext/ddtrace.c
SammyK: > I wonder whats the downside keeping them in?

I'm just thinking of the commitment to ongoing maintenance and not exposing any API's that aren't necessary so we have fewer issues if we need to change or remove them. But I'm not saying we shouldn't add them, I'm just wondering if we have sufficient reason for adding them and are OK with any tradeoffs for adding them. So if you're confident, I'm confident too. :)

> could be used to extend/ enhance the functioning a bit in the future.

I personally don't like adding code for possible future features since the domain could change and we have to commit to maintaining the future code for an indeterminate amount of time. In my experience, most times I've added future code, the future code is never used. :D 

> I would prefer to test as close to the final binary as possible

I agree - I think "to ensure stability" is a pretty good reason to expose these. I would prefer a method to ensure stability that doesn't expose any more public API's, but if there's really no other viable alternative, I'm on board with this! :)

pawelchcki: Excellent points Sammy!

I think I could remove a few functions though. Will probably be able to drop `_close` and `_open`. Probably `_is_closed` to since its functionality is implemented by a new `_info` .

src/DDTrace/Transport/Http.php
```diff
@@ -82,6 +82,17 @@ private function configure($config)
      */
     public function send(Tracer $tracer)
     {
+        if ($this->isLogDebugActive() && function_exists('dd_tracer_circuit_breaker_info')) {
+            self::logDebug('circuit breaker status: closed => {closed}, total_failures => {total_failures},'
+            . 'consecutive_failures => {consecutive_failures}, opened_timestamp => {opened_timestamp}, '
+            . 'last_failure_timestamp=> {last_failure_timestamp}', dd_tracer_circuit_breaker_info());
```
SammyK: ❤️

.gitattributes
```diff
@@ -11,3 +11,4 @@ RELEASING.md        export-ignore
 .circleci           export-ignore
 .clang-format       export-ignore
 src/ext/mpack/*     linguist-generated=true
+src/ext/vendor_stdatomic.h     linguist-generated=true
```
SammyK: 🔝

PR #441, opened 2019-05-16
Append hostname to root span

pawelchcki:
### Description

a Very simple PR :) adds hostname to a root span when enabled (only at very last moment to avoid people being able to easily overwrite this hostname).

### Readiness checklist

### DIFF
--- a/tests/Unit/TracerTest.php
+++ b/tests/Unit/TracerTest.php
@@ -178,7 +178,7 @@ final class TracerTest extends BaseTestCase
 
     public function testUnfinishedSpansCanBeFinishedOnFlush()
     {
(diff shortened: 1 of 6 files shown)


tests/Unit/TracerTest.php
```diff
@@ -212,7 +238,7 @@ public function testIfNoRootScopeExistsItWillBeNull()
 
     public function testHonorGlobalTags()
     {
-        Configuration::replace(\Mockery::mock('\DDTrace\Configuration', [
+        Configuration::replace(\Mockery::mock(Configuration::get(), [
```
labbati: ❤️

PR #442, opened 2019-05-16
Normalized URL as the resource name

SammyK:
### Description

This PR adds URL normalization so that that they can be used as a low-cardinality resource name. The default substitution patterns scan for UUID's, hexadecimal hashes, and int's. We can certainly add more to that list.

The `DD_TRACE_RESOURCE_URI_MAPPING` environment variable can be set to a CSV list of URL-to-resource-name mapping rules that can contain `*` and `$*` wildcards.

- The `*` wildcard will match one or more characters to be replaced with `?`
- The `$*` wildcard will match one or more characters without any replacement

For example, given `DD_TRACE_RESOURCE_URI_MAPPING=/user/*,/city/$*/*`:

| URL                     | Normalized    |
| ---------------- | ------------- |
| `/user/sammyk` | `/user/?`        |
| `/city/foo/123`    | `/city/foo/?`  |
| `/city/bar/123`    | `/city/bar/?`  |

### Readiness checklist


src/DDTrace/Bootstrap.php
```diff
@@ -112,6 +110,12 @@ private static function initRootSpan()
             $span->setTag(Tag::HTTP_URL, $_SERVER['REQUEST_URI']);
             // Status code defaults to 200, will be later on changed when http_response_code will be called
             $span->setTag(Tag::HTTP_STATUS_CODE, 200);
+            // Normalized URL as the resource name
+            $normalizer = new Urls(explode(',', getenv('DD_TRACE_RESOURCE_URI_MAPPING')));
```
labbati: A couple of comments here, if they make sense to you:

- While I see our conversation about simplifying the way we read configurations, wouldn't you see more strategic to not use `getenv` here? What I mean is that you may just want to move this `getenv()` call to a method in `Config` and not use all the underlying infrastructure.
- I was evaluating pros and cons of doing this here or as a post processing operation. Doing it here means that devs cannot, as an example, set this env in their `index.php` or the likes. In the end we need that info only at the very end, just before we send the trace. Wouldn't you think it is a good idea to push this setting to the very end?

src/DDTrace/Configuration/WildcardToRegex.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace DDTrace\Configuration;
```
labbati: I love this class, I wonder why it is in the `Configuration` namespace though? :)

src/DDTrace/Http/Urls.php
```diff
@@ -2,12 +2,37 @@
...
 /**
  * A utility class that provides methods to work on urls
  */
 class Urls
 {
+    private static $defaultPatterns = [
+        // UUID's
+        '|\b([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89AB][0-9a-f]{3}-[0-9a-f]{12})\b|i',
+        // 16-512 bit hex hashes
+        '|\b([0-9a-f]{4,128})\b|i',
```
labbati: To be honest I would not add this. Isn't it too risky to have false positives with a 4-128 range? WDYT?

src/DDTrace/Http/Urls.php
```diff
@@ -2,12 +2,37 @@
...
 
-
 /**
  * A utility class that provides methods to work on urls
  */
 class Urls
 {
+    private static $defaultPatterns = [
+        // UUID's
+        '|\b([0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89AB][0-9a-f]{3}-[0-9a-f]{12})\b|i',
```
labbati: This is just based on my previous experience, can we make the `-` optional? We slightly increase the risk to get false positives here but it is limited to only hex strings with one exact number of chars, which is perfectly acceptable in my opinion.

I really appreciate the detailed regex considering `[89AB]` from the spec. I would ask you though:
- to add a comment to the [section](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format) in wikipedia explaining it as it surprised me :) and I had to go through a few google link to find it 😄 
- for consistency with the other parts of the expression can we make it lowercase, the pattern in `i`nsensitive so it should make no difference.

src/DDTrace/Http/Urls.php
```diff
@@ -2,12 +2,37 @@
...
+    /**
+     * Inject URL replacement patterns using '*' and '$*' as wildcards
+     * The '*' wildcard will match one or more characters to be replaced with '?'
+     * The '$*' wildcard will match one or more characters without replacement
+     *
+     * @param string[] $patternsWithWildcards
+     */
+    public function __construct(array $patternsWithWildcards = [])
+    {
+        $this->replacementPatterns = array_map(
```
labbati: [Optional]: this is a matter of styles and this comment is totally optional. I love to `Ctrl+click` around to see who is using a method. While I know that UTs would detect something wrong here, I still prefer the plain old `for` loop. This is personal here :) don't hate me 😄

tests/Unit/Configuration/WildcardToRegexTest.php
```diff
@@ -0,0 +1,29 @@
...
+     */
+    public function testWildcardStringsCanConvertToRegex($withWildcards, array $regexAndReplacement)
+    {
+        $this->assertSame($regexAndReplacement, WildcardToRegex::convert($withWildcards));
+    }
+
+    public function wildcardToRegexExamples()
+    {
+        return [
+            ['/foo/*/bar',                  ['|^/foo/.+/bar$|', '/foo/?/bar']],
```
labbati: What happens to `^/foo/.+/bar$` with the pattern `/foo/123/legitimate/bar`?

src/DDTrace/Configuration/WildcardToRegex.php
SammyK: Very good point. Naming is hard. 😄

What do you think about adding an `Obfuscation` namespace? We could also move [the existing obfuscation stuff](https://github.com/DataDog/dd-trace-php/blob/master/src/DDTrace/Obfuscation.php) to that namespace in a separate PR. And possibly organize that code a bit better. What do you think? :)

src/DDTrace/Bootstrap.php
SammyK: Great points!

> What I mean is that you may just want to move this getenv() call to a method in Config and not use all the underlying infrastructure.

I'm certainly open to moving the `getenv()` to an abstraction. :) What do you see as the main advantage for abstracting this one-time env access?

> Wouldn't you think it is a good idea to push this setting to the very end?

This is a great idea! I'll move it to right before the flush... maybe we should start a discussion about possibly adding span filters in a future PR? :)

src/DDTrace/Http/Urls.php
SammyK: From a security standpoint, I'd rather err on the side of over-obfuscation, but I totally see your point as well.

I was thinking about valid words that are 4+ characters that contain only a-f (like *cafe*). One way around erroneous obfuscation would be curating a whitelist of valid words that are 4+ and a-f. Perhaps we could eventually offer a feature to add words to the whitelist in the event there is a proper noun or a made-up word that's important (like *affeca*). There are also i18n considerations to support other languages.

In the end, a configurable whitelist that's also very fast (maybe making use of a Trie data structure or something) would be a bit of an undertaking, but it might be worth the tradeoff to ensure better security.

If you're OK with it, I'd like to ping someone from the security department to weigh in on this one. :)

src/DDTrace/Http/Urls.php
SammyK: > This is just based on my previous experience, can we make the - optional?

Good call! There's [a test for this](https://github.com/DataDog/dd-trace-php/pull/442/files#diff-0ebbc6ef112d2cd18640240b6f8557a7R49) and it is obfuscated by the regex that catches "16-512 bit hex hashes". But since we may end up changing/removing that regex based on your other comment above, I'll change the regex to make the dashes optional. :)

> to add a comment to the section in wikipedia explaining it

Good idea! I'll add that. :)

> for consistency with the other parts of the expression can we make it lowercase

Another good idea. I'll change that. :)

tests/Unit/Configuration/WildcardToRegexTest.php
SammyK: The wildcards are [greedy by default](https://github.com/DataDog/dd-trace-php/pull/442/files#diff-0ebbc6ef112d2cd18640240b6f8557a7R82). Do you think we should add a non-greedy wildcard like `*?`? :)

src/DDTrace/Http/Urls.php
SammyK: Lol - I don't mind refactoring this to a `foreach` loop since we aren't able to support `ClassName::class`. :)

src/DDTrace/Http/Urls.php
SammyK: For now I've narrowed the scope of this to start at 32-bit hashes since an 8-character word with only a-f is much less likely to occur. :)

src/DDTrace/Integrations/Lumen/V5/LumenIntegrationLoader.php
```diff
@@ -34,7 +34,7 @@ public function load()
 
         dd_trace('Laravel\Lumen\Application', 'dispatch', function () use ($span) {
             $response = dd_trace_forward_call();
-            $resourceName = 'unnamed_route';
+            $resourceName = null;
```
labbati: Look at who is sneaking in 😄 👍

tests/Unit/Configuration/WildcardToRegexTest.php
```diff
@@ -0,0 +1,29 @@
+<?php
```
labbati: Shuold we move this to Obfuscation namespace as well?

tests/Unit/Configuration/WildcardToRegexTest.php
labbati: Let's leave like this, and users can add their own specific paths. Also, if we receive feedback about it then we can iterate on it.

src/DDTrace/Tracer.php
```diff
@@ -399,6 +404,25 @@ private function addHostnameToRootSpan()
         }
     }
 
+    private function addUrlAsResourceNameToRootSpan()
+    {
```
labbati: I think the only thing we miss is to make this OFF by default. This in order to prevent unexpected huge cardinality (and so bill) issues out of the blu.

tests/Unit/Configuration/WildcardToRegexTest.php
SammyK: Doh! 🤦‍♂ Nice catch! 😄

PR #449, opened 2019-05-23
Replace MT19937 with MT19937-64

SammyK:
### Description

This PR reduces the number of function calls we need to generate the unsigned 63-bit int strings we use as span ID's by adding a new function called `dd_trace_generate_id()`. The function implements [MT19937-64](http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/VERSIONS/C-LANG/mt19937-64.c) with one bit shifted off.

The PRNG can be seeded via an env variable ~~`DD_TRACE_PRNG_SEED`~~ `DD_TRACE_DEBUG_PRNG_SEED`, but if no seed is provided, the PRNG will be seeded via PHP's PRNG seed generator.

### Readiness checklist


tests/ext/dd_trace_generate_id.phpt
```diff
@@ -0,0 +1,21 @@
+--TEST--
+dd_trace_generate_id() Generates a 63-bit unsigned int as a string
+--ENV--
+DD_TRACE_PRNG_SEED=42
```
pawelchcki: ❤️

PR #452, opened 2019-05-27
fix test for upcomint 7.4

remicollet:
From UPGRADING file

> 
>   . "fn" is now a reserved keyword. In particular it can no longer be used as a
>     function or class name. It can still be used as a method or class constant
>     name.
>


pawelchcki:
Thanks for this awesome PR @remicollet !

PR #459, opened 2019-05-31
Group by hostname for curl and Guzzle requests

SammyK:
### Description

This PR adds a feature to set hostname of a URL as the service name for curl and Guzzle requests when `DD_TRACE_HTTP_CLIENT_SPLIT_BY_DOMAIN=true` (e.g. `example.com` and `example.com`)

### Readiness checklist


src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -64,6 +64,9 @@ public static function load()
 
             $info = curl_getinfo($ch);
             $sanitizedUrl = Urls::sanitize($info['url']);
+            if ($globalConfig->isHttpClientSplitByDomain()) {
```
labbati: What's the motivation for keeping `$span->setTag(Tag::SERVICE_NAME, 'curl');`? In case something in between fails we want to at least set `curl` as the service name?

src/DDTrace/Integrations/Guzzle/GuzzleIntegration.php
```diff
@@ -138,10 +139,16 @@ private function buildLimitTracerCallback()
      */
     private function setUrlTag(Span $span, $request)
     {
+        $url = null;
```
labbati: TBH I don't particularly like that a method `setUrlTag` sets both the url tag and overwrite the service name.

What do you think to:

1. Add a `getUrl` that applies the logic to extract the url
2. In `buildPreCallback()` we get the url and if it is bot null we set both the url tag and the service tag?

pawelchcki: @labbati great suggestion, applied !

src/DDTrace/Integrations/Curl/CurlIntegration.php
pawelchcki: fixed, thanks for spotting it.

PR #477, opened 2019-06-13
Only override execute on compilation step

pawelchcki:
### Description

Overriding `zend_execute_ex` causes tail recursive calls to fail due to different codepaths taken by the opcode parsers based on `zend_execute_ex == execute_ex` check. This makes each subsequent call to tailrecursive function actually be put on stack which causes os error to be raised due to stack overflow.

### Readiness checklist

### DIFF
--- a/src/ext/dispatch_setup.c
+++ b/src/ext/dispatch_setup.c
@@ -18,13 +18,35 @@ user_opcode_handler_t ddtrace_old_icall_handler;
 user_opcode_handler_t ddtrace_old_fcall_by_name_handler;
 
 #if PHP_VERSION_ID >= 70000
-void (*ddtrace_original_execute_ex)(zend_execute_data *TSRMLS_DC);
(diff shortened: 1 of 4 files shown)


src/ext/dispatch_setup.c
```diff
@@ -18,19 +18,60 @@ user_opcode_handler_t ddtrace_old_icall_handler;
...
 }
+
+#if PHP_VERSION_ID >= 70300
+static int (*ddtrace_original_post_startup_cb)(void);
+
+static int php_post_startup_cb(void) {
+    if (ddtrace_original_post_startup_cb) {
+        int (*callback)(void) = ddtrace_original_post_startup_cb;
+        ddtrace_original_post_startup_cb =
+            NULL;  // guard against possible recursion if callback is calling zend_post_startup_cb directly
```
SammyK: ❤️

PR #478, opened 2019-06-13
Fix backtrace handler being run more than once, and inform about max stacktrace size reached

pawelchcki:
### Description
In cases of stack overflow exception the backtrace handler can be run more than one time. This PR fixes that case and informs user when max stacktrace size (to be printed) have been reached indicating a possible stackoverflow type error.

### Readiness checklist

### DIFF
--- a/src/ext/backtrace.c
+++ b/src/ext/backtrace.c
@@ -14,13 +14,27 @@
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
+struct _backtrace_data_t {
+    BOOL_T handler_already_run;
+    BOOL_T handler_installed;
+};
+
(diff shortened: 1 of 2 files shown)


src/ext/backtrace.c
```diff
@@ -44,10 +58,8 @@ void ddtrace_install_backtrace_handler(TSRMLS_D) {
 
     static int handler_installed = 0;
     if (!handler_installed) {
```
SammyK: Shouldn't this be:

```suggestion
    if (!backtrace_globals.handler_installed) {
```

And deleting the `static int handler_installed = 0;` statement? :)

pawelchcki:
Thanks for review Sammy!

Issue #315, opened 2019-02-21
Predis Integration exception


ynuyasha:
Hi,
if i enable ddtrace module, my application breaks with this error:

Recoverable Error: Object of class Predis\Connection\Aggregate\MasterSlaveReplication could not be converted to string (File: /opt/datadog-php/dd-trace-sources/src/DDTrace/Integrations/Predis/PredisIntegration.php - Line: 171)

labbati:
Hi @ynuyasha 👋 , thanks for reporting this. Not sure if you are interested in it, but if you like to contribute this would be a very good first issue. Otherwise, we will work on it as soon as possible.

In the meantime, you can still use all the other integrations and disable only Predis `DD_INTEGRATIONS_DISABLED=false` (see [docs](https://docs.datadoghq.com/tracing/languages/php/#configuration))

bwoebi:
This has been fully fixed by #574 and then as part of #1094.

PR #479, opened 2019-06-14
Handle MasterSlaveReplication connections

liam-wiltshire:
At the moment, DD causes errors in some situations when using predis/predis, when predis is using master-slave replication. This was discussed in #315 

There may be a couple of points of discussion around this:
- What should happen when `Client::getConnection()` returns a `MasterSlaveReplication` connection, but the connection hasn't been used yet, so `MasterSlaveReplication::connection` is empty. In our use case (and thus in this PR) we ask it to return the master, but this might not be preferred behaviour?

- I looked at writing a test to cover this, but I was struggling to get my head around how your tests work, and I couldn't actually create a breaking test in the first place to fix! If someone can point me in the right direction @SammyK @pawelchcki the  I'll take another look at getting a test to play nice!!

### DIFF
--- a/src/DDTrace/Integrations/Predis/PredisIntegration.php
+++ b/src/DDTrace/Integrations/Predis/PredisIntegration.php
@@ -225,7 +225,7 @@ class PredisIntegration extends Integration
             $connection = $predis->getConnection();
 
             if ($connection instanceof \Predis\Connection\Aggregate\MasterSlaveReplication) {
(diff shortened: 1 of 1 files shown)


liam-wiltshire:
Yeah, so I might have forgot PHP 5 - will replace the ?? :)

krakjoe:
I'm closing this as it no longer looks applicable, and would appear to be stale.

If I'm wrong to do that, please open a new PR, and accept my apologies in advance.

PR #486, opened 2019-06-21
Backport memoized configuration API

pawelchcki:
### Description

This PR fixes race condition when reading configuration from within writer context leading to possible crashes due to zend_mm_ corruption.

New configuration API implements getters that follow memoization pattern, to load and store configuration data. Getters are implemented using X-Macros, via declarative "DSL" to define configuration options and its defaults without extra boilerplate necessary.

Thanks to this approach it should be extremely fast. 

### Readiness checklist


src/ext/configuration.h
```diff
@@ -0,0 +1,24 @@
+#ifndef DD_CONFIGURATION_H
+#define DD_CONFIGURATION_H
+struct ddtrace_memoized_configuration_t;
+extern struct ddtrace_memoized_configuration_t ddtrace_memoized_configuration;
+
+void ddtrace_initialize_config();
+void ddtrace_reload_config();
+
+#define DD_CONFIGURATION                                                                        \
```
SammyK: ❤️ So much cleaner :)

src/ext/configuration_php_iface.c
```diff
@@ -0,0 +1,116 @@
...
+        return 0;
+    }
+
+    if (snprintf(envname, envname_length + sizeof('\0'), ID_TO_ENV_PREFIX "%s", id) <= 0) {
+        free(envname);
+        return 0;
+    }
+
+    char *cptr = envname;
+    while (*cptr && (size_t)(cptr - envname) < envname_length) {
```
SammyK: 🔝  This is a great way to be compatible with the existing config naming - nice catch!

Ideally in the future it would be great to move away from the Java-style config names so we don't have the extra overhead of converting them and it makes the config vars easier to search for them in the codebase. What do you think? :)

src/ext/ddtrace.c
```diff
@@ -226,6 +233,26 @@ static PHP_FUNCTION(dd_trace_forward_call) {
 #endif
 }
 
+static PHP_FUNCTION(dd_trace_cfg) {
```
SammyK: What are your thoughts on renaming this to:

```suggestion
static PHP_FUNCTION(dd_trace_env_config) {
```

src/ext/ddtrace.c
```diff
@@ -226,6 +233,26 @@ static PHP_FUNCTION(dd_trace_forward_call) {
...
+    PHP5_UNUSED(return_value_used, this_ptr, return_value_ptr, ht TSRMLS_CC);
+    PHP7_UNUSED(execute_data);
+    zval *env_name = NULL;
+
+    if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "z", &env_name) != SUCCESS) {
+        if (DDTRACE_G(strict_mode)) {
+            zend_throw_exception_ex(spl_ce_InvalidArgumentException, 0 TSRMLS_CC,
+                                    "unexpected parameter. the function name must be provided");
```
SammyK: Just a small tweak :)

```suggestion
                                    "unexpected parameter. the environment variable name must be provided");
```

src/ext/env_config.c
```diff
@@ -1,17 +1,32 @@
...
     if (tmp) {
-        env = estrdup(tmp);
+        env = ddtrace_strdup(tmp);
     } else {
+        // reading sapi_getenv from within writer thread can and will lead to undefined behaviour
+        if (ddtrace_in_writer_thread()) {
+            return NULL;
+        }
+
+        TSRMLS_FETCH();
```
SammyK: This macro adds quite a bit of overhead which might be one of the the reasons why config has been a bit slow. I know this would make the API's really ugly, but perhaps we should consider adding the `TSRMLS_DC` macros to the these API's. What are your thoughts on this? :)

src/ext/configuration.h
```diff
@@ -0,0 +1,24 @@
...
+    CHAR(get_dd_agent_host, "DD_AGENT_HOST", "localhost")                                       \
+    INT(get_dd_trace_agent_port, "DD_TRACE_AGENT_PORT", 8126)                                   \
+    BOOL(get_dd_trace_agent_debug_verbose_curl, "DD_TRACE_AGENT_DEBUG_VERBOSE_CURL", FALSE)     \
+    BOOL(get_dd_trace_debug_curl_output, "DD_TRACE_DEBUG_CURL_OUTPUT", FALSE)                   \
+    BOOL(get_dd_trace_send_traces_via_thread, "DD_TRACE_BETA_SEND_TRACES_VIA_THREAD", FALSE)    \
+    CHAR(get_dd_trace_memory_limit, "DD_TRACE_MEMORY_LIMIT", NULL)                              \
+    INT(get_dd_trace_agent_flush_interval, "DD_TRACE_AGENT_FLUSH_INTERVAL", 5000)               \
+    INT(get_dd_trace_agent_flush_after_n_requests, "DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS", 10) \
+    INT(get_dd_trace_agent_timeout, "DD_TRACE_AGENT_TIMEOUT", 500)                              \
+    INT(get_dd_trace_agent_connect_timeout, "DEFAULT_AGENT_CONNECT_TIMEOUT", 100)
```
SammyK: I know this isn't a comprehensive list of all the env vars the tracer uses, but perhaps it's worthwhile to replace the existing `getenv()` calls in userland with `dd_trace_cfg()` for these env vars? For example in the [`Http` class](https://github.com/DataDog/dd-trace-php/blob/ba07549dec65113ff2e61d3cd9b40d393395d0da/src/DDTrace/Transport/Http.php#L66-L78). What are your thoughts? :)

tests/ext/read_c_configuration.phpt
```diff
@@ -0,0 +1,54 @@
...
+echo dd_trace_cfg("DD_TRACE_DEBUG_CURL_OUTPUT") ? 'TRUE' : 'FALSE';
+echo PHP_EOL;
+echo dd_trace_cfg("DD_TRACE_BETA_SEND_TRACES_VIA_THREAD") ? 'TRUE' : 'FALSE';
+echo PHP_EOL;
+echo dd_trace_cfg("DD_TRACE_MEMORY_LIMIT");
+echo PHP_EOL;
+echo dd_trace_cfg("DD_NON_EXISTING_ENTRY") === NULL ? 'NULL' : 'NOT_NULL' ;
+echo PHP_EOL;
+
+// this format is for compatibility with existing configuration registry
```
SammyK: ❤️

bridge/dd_init.php
```diff
@@ -7,6 +7,8 @@
 
 require_once __DIR__ . '/functions.php';
 
+// trigger configuration reload to memoize values of all configuration options as set by environment variables
+function_exists('dd_trace_internal_fn') && dd_trace_internal_fn('ddtrace_reload_config');
```
SammyK: Do we need this here since it's being called in **bridge/dd_wrap_autoloader.php**? I might be missing the context for this though. :)

src/ext/compatibility.h
```diff
@@ -46,7 +46,9 @@ typedef int32_t zend_long;
 #if PHP_VERSION_ID < 70000
 #define ZVAL_VARARG_PARAM(list, arg_num) (*list[arg_num])
 #define IS_TRUE_P(x) (Z_TYPE_P(x) == IS_BOOL && Z_LVAL_P(x) == 1)
+#define COMPAT_RETVAL_STRING(c) RETVAL_STRING(c, 1)
```
SammyK: 🔝

src/ext/configuration_php_iface.c
pawelchcki: Yeah, I sort of like improved readability of "agent.host" over "DD_AGENT_HOST". But grep-ability :D Is I think more important if someone first looks at the code and tries to find whats up.
And grep-ability of `DD_AGENT_HOST` is much better.  

I'm fine with either solution :). I've implemented this conversion shim to make sure we're able to support both at this moment but I'm very ok with phasing out "agent.host".

src/ext/ddtrace.c
pawelchcki: Its whole 7 characters more to type :D But `dd_trace_env_config` looks good to me 👍

src/ext/ddtrace.c
pawelchcki: Too much copy/paster from me 😄 Thanks for spotting that

src/ext/env_config.c
pawelchcki: 👍 Agreed, I'll tweak the api, also this is not thread safe, so I'll probably make this a mandatory parameters, and cause the background thread to only allow read only access to config.

bridge/dd_init.php
pawelchcki: I assumed someone could require `dd_init` directly too - just to be safe I've added `ddtrace_reload_config` call here to make sure its not surprising for our users if their `putenv` config stopped working.

pawelchcki:
Thanks for review @SammyK!
I was finally able to resolve all the compilation problems, and  I had to update most environment configuration loaded from C level to make sure ZTS compilation goes smoothly - instead of fixing ZTS invocations everywhere I converted most to use new API.

Also I've decided to make getters pure getters without memoization. The memoization of data is done on `MINIT` or on `ddtrace_configuration_reload` internal call.

PR #488, opened 2019-06-25
Request init hook version sync

SammyK:
### Description

This PR does version checks in the request init hook to ensure that the version of the extension matches the version in userland code. If the versions don't match, then the tracer does not run. This prevents the extension and PHP code from getting out of sync (which can happen after upgrading the tracer and before restarting the SAPI.)

The extension version is exposed via a new userland constant `DD_TRACE_VERSION`.

One thing to keep in mind. Until we remove `Tracer::VERSION`, we need to update the version number in three places for version bumps:

1. **src/ext/version.h**
2. **src/DDTrace/Tracer.php** (`Tracer::VERSION`; deprecated in this PR)
3. **src/DDTrace/version.php** (new; accessed via `Tracer::version()`)

### Readiness checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,8 @@ BUILD_DIR := tmp/build_$(BUILD_SUFFIX)
 SO_FILE := $(BUILD_DIR)/modules/ddtrace.so
 WALL_FLAGS := -Wall -Wextra
 CFLAGS := -O2 $(WALL_FLAGS)
-VERSION:=$(shell cat src/DDTrace/Tracer.php | grep VERSION | awk '{print $$NF}' | cut -d\' -f2)
-VERSION_WITHOUT_SUFFIX:=$(shell cat src/DDTrace/Tracer.php | grep VERSION | awk '{print $$NF}' | cut -d\' -f2 | cut -d- -f1)
+VERSION:=$(shell cat src/DDTrace/version.php | grep return | awk '{print $$2}' | cut -d\' -f2)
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -143,6 +143,7 @@ verify_version:
 	@grep -q "<release>$(VERSION_WITHOUT_SUFFIX)</release>" package.xml || (echo package.xml release version missmatch && exit 1)
 	@grep -q "<api>$(VERSION_WITHOUT_SUFFIX)</api>" package.xml || (echo package.xml api version missmatch && exit 1)
 	@grep -q "#define PHP_DDTRACE_VERSION \"$(VERSION)" src/ext/version.h || (echo src/ext/version.h Version missmatch && exit 1)
+	@grep -q "const VERSION = '$(VERSION)" src/DDTrace/Tracer.php || (echo src/DDTrace/Tracer.php Version missmatch && exit 1)
```
pawelchcki: ❤️

src/DDTrace/Tracer.php
```diff
@@ -522,4 +530,15 @@ public function getSafeRootSpan()
...
+
+    /**
+     * @return string
+     */
+    public static function version()
+    {
+        if (empty(self::$version)) {
+            self::$version = include __DIR__ . '/version.php';
+        }
+        return self::$version;
```
pawelchcki: 👍

PR #490, opened 2019-07-01
Prefix invalid hostnames as service names

SammyK:
### Description

The Agent [expects tags](https://github.com/DataDog/datadog-agent/blob/5da007b59095381240223f59b29916207c553c39/pkg/trace/api/normalizer.go#L296-L317) to begin with a nonnumeric unicode character. With the `DD_TRACE_HTTP_CLIENT_SPLIT_BY_DOMAIN=true` feature added in #459, service names (sent to the Agent via a tag) can begin with a number (IP addresses & hostnames that begin with a number).

For example, making a request to `2guysaroundtheworld.com` with `DD_TRACE_HTTP_CLIENT_SPLIT_BY_DOMAIN=true` results in an incorrect service name in the UI:

Screen Shot 2019-07-01 at 10 04 17 AM

~~This PR ensures that any hostname that begins with a number will be prefixed with a valid tag character, `:`. So the previous example becomes:~~ After some discussion, all hostnames are now automatically prefixed with `host-`.

Screen Shot 2019-07-01 at 10 04 55 AM

Service names that consist of IP addresses [are ignored by the Agent](https://github.com/DataDog/datadog-agent/blob/5da007b59095381240223f59b29916207c553c39/pkg/trace/api/normalizer.go#L336-L338) so the traces only appear in the UI as a partial trace or no trace at all. This PR ensures that service names that are IP addresses are prefixed with a valid character.

Screen Shot 2019-07-01 at 10 05 48 AM

### Readiness checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file - [read more
 - Edge case where the extension version and userland version can get out of sync #488
 
 ### Changed
-- Prefix invalid hostnames as service names with `:` #490
+- Prefix hostnames as service names with `host-` to ensure compatibility with the Agent #490
 
 ## [0.28.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ All notable changes to this project will be documented in this file - [read more
 ### Fixed
 - Edge case where the extension version and userland version can get out of sync #488
 
+### Changed
+- Prefix invalid hostnames as service names with `:` #490
+
 ## [0.28.1]
 
 ### Fixed
--- a/src/DDTrace/Http/Urls.php
+++ b/src/DDTrace/Http/Urls.php
@@ -60,6 +60,22 @@ class Urls
     }
 
     /**
+     * Metadata keys must start with [a-zA-Z:] so IP addresses,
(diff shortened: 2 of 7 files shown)


SammyK:
This is ready for a CR, but PHP 7.1 & 7.3 web integration tests are failing with:

> The process has been signaled with signal "9".

So it looks like we're hitting a memory limit with CircleCI. 😬

SammyK:
@labbati Nice catch on the metrics names! I'm curious if we should just automatically prepend `host-` to all service names by default; even for valid hostnames. I'm thinking that would keep things consistent across services and make the service map a bit more consistent.

Also, (for a separate PR later) what are your thoughts on offering a bit more sanitization on custom tag/metadata names that ensure that they conform to metric name conventions? :)

labbati:
I am fine with `host-`

brettlangdon:
After talking with Manu the reason we don't allow a starting with a number is because of metrics tagging, the service name gets converted to a metrics tag at some point which is why we have the same restriction.

brettlangdon:
I think a generic prefix could make sense here, `host-`

labbati:
Thanks Brett

SammyK:
Thanks @labbati and @brettlangdon! I've updated the PR to prepend `host-` to all hostnames used as service names. :)

PR #491, opened 2019-07-01
Sandboxing: Append tracing closures

SammyK:
**TL;DR:** Exceptions and errors generated from within a tracing closure are ignored. The original call is forwarded at the C level _before_ and independently of a tracing closure.

### The current issue

When functions and methods are traced with `dd_trace()`, the original call is forwarded from _within_ the tracing closure via `dd_trace_forward_call()`. This prevents us from ignoring any errors raised (often due to invalid argument types) or exceptions thrown from the tracing closures since ignoring errors and exceptions will also bubble down to the original call and change the original call behavior.

Another tradeoff to forwarding the call from within the tracing closure occurs when exceptions are thrown in the original call. In that case, the ddtrace tracing closure appears in the stack trace which is often mistaken as the source of the exception.

### The solution

This PR addresses these issues by forwarding the original call at the C level and then executing the tracing closure _after_ the original call in a sandboxed environment. We now have the following safeguards in place:

1. All exceptions and errors are ignored from tracing closures so that the original call is unaffected.
2. The tracing closure no longer appears in the backtrace of an exception thrown in the original call to reduce confusion when debugging.
3. Tracing closures are not executed when there is an uncaught exception in the original call.
4. The original call remains the first to receive any arguments allowing errors/exceptions to be handled by the original call instead of by the tracing closure.

Spans are also available at the C level now which offers a light-weight alternative to the existing open-close span cycle in userland.

The new API's described below work alongside the existing implementation of `dd_trace()`. This will allow us to migrate integrations one at a time until we can move them all to the new "append tracing closures."

### Tracing Functions

Functions are traced with `dd_trace_function()` and the tracing closure is executed after the original call is forwarded.

```php
<?php

use DDTrace\SpanData;

function addNums($a, $b) {
	$sum = $a + $b;
	printf("%d + %d = %d\n", $a, $b, $sum);
	return $sum;
}

dd_trace_function(
	'addNums',
	function(SpanData $span, $args, $retval) {
		echo "Traced\n";
	}
);

var_dump(addNums(2, 8));
// 2 + 8 = 10
// Traced
// int(10)
```

### Tracing Methods

Methods are traced with `dd_trace_method()` and the tracing closure is executed after the original call is forwarded.

```php
<?php

use DDTrace\SpanData;

class Calc {
	public function addNums($a, $b) {
		$sum = $a + $b;
		printf("%d + %d = %d\n", $a, $b, $sum);
		return $sum;
	}
}

dd_trace_method(
	'Calc', 'addNums',
	function(SpanData $span, $args, $retval) {
		echo "Traced\n";
	}
);

$calc = new Calc();
var_dump($calc->addNums(2, 8));
// 2 + 8 = 10
// Traced
// int(10)
```

### Tracing Closure Arguments

The tracing closure provides three arguments:

```php
function(DDTrace\SpanData $span, array $args, mixed $retval);
```

1. An instance of `DDTrace\SpanData` which is the API to write span information
2. An array of arguments, `$args`, from the original call (similar to [`func_get_args()`](https://www.php.net/func_get_args))
3. The return value, `$retval`, of the original call

#### Arg 1: `DDTrace\SpanData $span`

The `DDTrace\SpanData` instance contains [the same span information that the Agent expects](https://docs.datadoghq.com/api/?lang=python#send-traces). A few exceptions are `trace_id`, `span_id`, `parent_id`, `start`, and `duration` which are set at the C level and not exposed to userland via the `DDTrace\SpanData` object.

> **Note:** The `error` field will be added to the C level in a followup PR that focuses on error handling.

| Property | Type | Description |
| --- | --- | --- |
| `SpanData::$name` | `string` | The span name |
| `SpanData::$resource` | `string` | The resource you are tracing |
| `SpanData::$service` | `string` | The service you are tracing |
| `SpanData::$type` | `string` | _(Optional)_ The type of request which can be set to: **web**, **db**, **cache**, or **custom** |
| `SpanData::$meta` | `string[]` | _(Optional)_ An array of key-value span metadata; keys and values must be strings |
| `SpanData::$metrics` | `float[]` | _(Optional)_ An array of key-value span metrics; keys must be strings and values must be floats |

```php
<?php

use DDTrace\SpanData;

dd_trace_function(
	'mt_rand',
	function(SpanData $span, $args, $retval) {
		$span->name = 'mt_rand';
		$span->resource = 'mt_rand';
		$span->service = 'php';
		// The following are optional
		$span->type = 'web';
		$span->meta = [
			'rand.range' => $args[0] . ' - ' . $args[1],
			'rand.value' => $retval,
		];
		$span->metrics = [
			'_sampling_priority_v1' => 0.9,
		];
	}
);
```

#### Arg 2: `array $args`

The second argument to the tracing closure is an array of the original arguments. One thing to keep in mind: since the tracing closure is executed _after_ the original call, any arguments passed by reference could be a different value when they reach the tracing closure.

```php
<?php

use DDTrace\SpanData;

function argsByRef(&$a) {
	return ++$a;
}

dd_trace_function(
	'argsByRef',
	function(SpanData $span, $args, $retval) {
		var_dump($args);
	}
);

$foo = 10;
var_dump(argsByRef($foo));
// array(1) {
//   [0]=>
//   int(11)
// }
// int(11)
```

When types are declared in the function header, the arguments will be passed to the tracing closure as they were _before_ any type coercion by the type declaration.

```php
<?php

use DDTrace\SpanData;

function typeCoercion(string $a) {
	var_dump($a);
}

dd_trace_function(
	'typeCoercion',
	function(SpanData $span, $args, $retval) {
		var_dump($args[0]);
	}
);

$foo = 10;
typeCoercion($foo);
// string(2) "10"
// int(10)
```

#### Arg 3: `mixed $retval`

The final argument passed to the tracing closure is the return value of the original call. Functions or methods that declare a `void` return type or ones that don't return at all will have a `$retval` of `NULL` in the tracing closure.

```php
<?php

use DDTrace\SpanData;

function message(): void {
	echo "Hello!\n";
}

dd_trace_function(
	'message',
	function(SpanData $span, $args, $retval) {
		echo "Traced\n";
		var_dump($retval);
	}
);

var_dump(message());
// Hello!
// Traced
// NULL
// NULL
```

### Future considerations

We can make the tracing closure an optional parameter by using reasonable defaults on the span:

* `SpanData::$name` -> fully-qualified function or method name: `functionName` or `ClassName::methodName`
* `SpanData::$resource` -> same value as `SpanData::$name` (or fully-qualified function/method name if `SpanData::$name` is custom)
* `SpanData::$service` -> the value of the `DD_SERVICE_NAME` env var

There are two pieces of span information, `type` and `service`, that are usually static and could be exposed to userland by adding two new arguments when registering a tracing closure.

A possible API could look like the following:

```php
dd_trace_function(
	'myFunction',
	NULL, // Tracing closure is optional saving overhead
	'foo-service', // The service name
	DD_TRACE_TYPE_WEB, // One of four supported types
);
```

In the above example, if `foo-service` was not passed, the service would default to `DD_SERVICE_NAME`. If `DD_TRACE_TYPE_WEB` was not passed, no type would be sent to the Agent allowing the the value to default to **custom**.

### Readiness checklist


src/ext/dispatch_compat_php7.c
```diff
@@ -71,7 +71,9 @@ zend_bool ddtrace_dispatch_store(HashTable *lookup, ddtrace_dispatch_t *dispatch
     return zend_hash_update_ptr(lookup, Z_STR(dispatch->function_name), dispatch) != NULL;
 }
 
-void ddtrace_forward_call(zend_execute_data *execute_data, zval *return_value TSRMLS_DC) {
+// This function is used by dd_trace_forward_call() from userland and can be removed
```
pawelchcki: With this sandboxed function will it be possible to inject distributed headers in requests made via e.g. Guzzle ?
```
$client->request('GET', '/foo.js', [
    'headers'        => ['Accept-Encoding' => 'gzip'],
    'decode_content' => false
]);
```

Also what will happen if function parameters are modified during function execution and we will be interested in their initial contents ? 

My point is - its not certain we will be able to remove `dd_trace` soon. So as mentioned in another comment I think its fine to point the relationship to a certain code via function name e.g. 

`ddtrace_wrapper_forward_call_from_userland`. And rename some of the old behavior functions to contain the common root `wrapper`. This way if/when we will remove them - we should be able to easily identify all the dependencies.

tests/ext/dd_trace_method.phpt
```diff
@@ -0,0 +1,174 @@
...
+        echo "Foo::bar()\n";
+        return [
+            'thoughts' => $thoughts,
+            'first' => isset($bar[0]) ? $bar[0] : '(none)',
+            'rand' => mt_rand(42, 999)
+        ];
+    }
+}
+
+var_dump(dd_trace_method('Test', 'testFoo', function (SpanData $span) {
```
pawelchcki: I like the distinction between dd_trace function and method 👍 Its a good change :)

src/ext/ddtrace.c
```diff
@@ -224,10 +283,80 @@ static PHP_FUNCTION(dd_trace) {
         }
     }
 
-    zend_bool rv = ddtrace_trace(class_name, function, callable TSRMLS_CC);
+    zend_bool rv = ddtrace_trace(class_name, function, callable, 0 TSRMLS_CC);
+    RETURN_BOOL(rv);
+}
+
+static PHP_FUNCTION(dd_trace_method) {
```
pawelchcki: I was thinking recently that we don't use Namespaces. Would it make users more happy if we did not "pollute" global namespace ? At very least would it make sense to explicitly define functions in root namespace ?
Its a good opportunity to consider it since we're introduction quite a few new functions .
🤔 `DDTrace/trace_method` vs `dd_trace_method`

pawelchcki: Also what about php 5.3 :duck:`s

src/ext/dispatch.h
```diff
@@ -9,12 +9,13 @@
 
 typedef struct _ddtrace_dispatch_t {
     zval callable, function_name;
+    zend_bool append;  // Remove after removing `dd_trace()`
```
pawelchcki: Bikeshedding comment warning: 
WDYT about calling it something similar  to`postprocess_function_result` -> append looks a bit ambiguous.

src/ext/ddtrace.c
```diff
@@ -540,34 +669,58 @@ static PHP_FUNCTION(dd_trace_internal_fn) {
...
-    RETURN_STR(dd_trace_generate_id());
-#else
-    char buf[20];
-    dd_trace_generate_id(buf);
 #if PHP_VERSION_ID >= 70000
-    RETURN_STRING(buf);
+#define RETURN_SPAN_ID(id_fn)                            \
+    do {                                                 \
+        char buf[20];                                    \
+        php_sprintf(buf, "%" PRIu64, (id_fn)(TSRMLS_C)); \
```
pawelchcki: Its fine in this instance since the id is of known max lenght. But my spidey senses do tingle when not using snprintf with set buffer length.
Also I think we could safely use vanilla (non php) `snprintf/sprintf` here, WDYT ?

src/ext/ddtrace.c
```diff
@@ -540,34 +669,58 @@ static PHP_FUNCTION(dd_trace_internal_fn) {
...
+        RETURN_STRING(buf);                              \
+    } while (0);
 #else
-    RETURN_STRING(buf, 1);
-#endif
+#define RETURN_SPAN_ID(id_fn)                            \
+    do {                                                 \
+        char buf[20];                                    \
+        php_sprintf(buf, "%" PRIu64, (id_fn)(TSRMLS_C)); \
+        RETURN_STRING(buf, 1);                           \
```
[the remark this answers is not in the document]
SammyK: It means to duplicate the string so we can copy it into `return_value`. :)

src/ext/dispatch_compat_php5.c
```diff
@@ -173,7 +175,9 @@ static int get_args(zval *args, zend_execute_data *ex) {
     return 1;
 }
 
-void ddtrace_forward_call(zend_execute_data *execute_data, zval *return_value TSRMLS_DC) {
+// This function is used by dd_trace_forward_call() from userland and can be removed
+// when we remove dd_trace() and dd_trace_forward_call() from userland.
```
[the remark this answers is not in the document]
SammyK: Good idea. I thought renaming it to `ddtrace_forward_call_from_userland()` would help indicate that but if you have a better naming suggestion, I can rename it. :)

src/ext/dispatch_compat_php5.c
```diff
@@ -281,4 +285,87 @@ BOOL_T ddtrace_should_trace_call(zend_execute_data *execute_data, zend_function
 
     return TRUE;
 }
+
+/**
+ * trace.c
```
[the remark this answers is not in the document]
SammyK: Ah! I wanted to ask you about this. Essentially, as we start thinking of how we want to separate the code for each major PHP version, this would be related to the code in **trace.c**. So for example if we had folders this could be put in **php5/trace.c** and **php7/trace.c**.

We probably don't want to figure out how to separate the code for this PR, but perhaps you, me and @morrisonlevi could have a chat about it sometime soon? :)

src/ext/dispatch_compat_php5.c
```diff
@@ -281,4 +285,87 @@ BOOL_T ddtrace_should_trace_call(zend_execute_data *execute_data, zend_function
...
+    }
+    args[1] = &user_args;
+    // Arg 2: mixed $retval
+    args[2] = &user_retval;
+
+    fci.param_count = 3;
+    fci.params = args;
+    fci.retval_ptr_ptr = &retval_ptr;
+    if (zend_call_function(&fci, &fcc TSRMLS_CC) == FAILURE) {
+        // TODO Log error
```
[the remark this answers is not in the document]
SammyK: What would your recommendation be on this one? We have a "one logger to rule them all" in the backlog, but for now, I'm not sure which method I should use here. :)

src/ext/random.c
```diff
@@ -5,24 +5,63 @@
...
 
-#if PHP_VERSION_ID >= 70200
-// zend_strpprintf() wasn't exposed until PHP 7.2
-zend_string *dd_trace_generate_id() { return zend_strpprintf(0, "%llu", generate_id()); }
-#else
-void dd_trace_generate_id(char* buf) { php_sprintf(buf, "%llu", generate_id()); }
-#endif
+uint64_t ddtrace_pop_span_id(TSRMLS_D) {
+    if (DDTRACE_G(span_ids_top) == NULL) {
+        return 0;
```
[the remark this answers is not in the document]
SammyK: Yes, it's used to indicate a root span. Essentially if there's no active span and we get `0` here, we know that the span we're creating is a root span. This is helpful since we don't know if C or PHP has the root span. :)

src/ext/serializer.c
```diff
@@ -171,3 +171,62 @@ int ddtrace_serialize_simple_array(zval *trace, zval *retval TSRMLS_DC) {
...
+
+    ADD_ELEMENT_IF_PROP_TYPE("name", IS_STRING);
+    ADD_ELEMENT_IF_PROP_TYPE("resource", IS_STRING);
+    ADD_ELEMENT_IF_PROP_TYPE("service", IS_STRING);
+    ADD_ELEMENT_IF_PROP_TYPE("type", IS_STRING);
+    ADD_ELEMENT_IF_PROP_TYPE("meta", IS_ARRAY);
+    ADD_ELEMENT_IF_PROP_TYPE("metrics", IS_ARRAY);
+
+    if (span->exception) {
+        // TODO Serialize exception
```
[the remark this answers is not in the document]
SammyK: This is part of @morrisonlevi's PR :)

src/ext/trace.c
```diff
@@ -0,0 +1,101 @@
...
+        EG(error_reporting) = orig_error_reporting;
+        // If the tracing closure threw an exception, ignore it to not impact the original call
+        if (EG(exception)) {
+            // TODO Log the exception
+            zend_clear_exception(TSRMLS_C);
+        }
+    }
+
+#if PHP_VERSION_ID < 50500
+    (void)opline;  // TODO Make work on PHP 5.4
```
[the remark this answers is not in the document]
SammyK: This is part of @morrisonlevi's `zend_execute_internal()` PR :)

src/ext/ddtrace.c
```diff
@@ -56,6 +73,42 @@ ZEND_END_ARG_INFO()
...
+    zend_class_entry ce_span_data;
+    INIT_NS_CLASS_ENTRY(ce_span_data, "DDTrace", "SpanData", NULL);
+#if PHP_VERSION_ID < 70000
+    ddtrace_ce_span_data = zend_register_internal_class_ex(&ce_span_data, NULL, NULL TSRMLS_CC);
+#else
+    ddtrace_ce_span_data = zend_register_internal_class_ex(&ce_span_data, NULL);
+#endif
+
+    // TODO Add the rest of these
+    // TODO Generate from array of meta keys
```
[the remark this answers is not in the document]
SammyK: Since the class constants aren't used at all, I'm going to remove them for now. We can add them in a PR down the road if we want to. :)

src/ext/ddtrace.c
```diff
@@ -56,6 +73,42 @@ ZEND_END_ARG_INFO()
...
+static void register_span_data_ce(TSRMLS_D) {
+    zend_class_entry ce_span_data;
+    INIT_NS_CLASS_ENTRY(ce_span_data, "DDTrace", "SpanData", NULL);
+#if PHP_VERSION_ID < 70000
+    ddtrace_ce_span_data = zend_register_internal_class_ex(&ce_span_data, NULL, NULL TSRMLS_CC);
+#else
+    ddtrace_ce_span_data = zend_register_internal_class_ex(&ce_span_data, NULL);
+#endif
+
+    // TODO Add the rest of these
```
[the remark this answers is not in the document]
SammyK: Same for these constants - I'm going to remove them for now. :)

src/ext/dispatch_compat_php7.c
SammyK: > With this sandboxed function will it be possible to inject distributed headers in requests made via e.g. Guzzle ?

The sandboxed closures won't allow us to inject headers, but since Guzzle uses curl under the hood, we can provide a C-level function handler override for `curl_exec` to inject any headers.

> Also what will happen if function parameters are modified during function execution and we will be interested in their initial contents ?

Args passed by reference will be in their post-modified state in the tracing closure. Args passed by value will be in the original state before any type coercion. (See the PR description under the **Arg 2: array $args** header.)

> And rename some of the old behavior functions to contain the common root wrapper. This way if/when we will remove them - we should be able to easily identify all the dependencies.

I like it! :)

src/ext/ddtrace.c
SammyK: I'm totally down for using namespaces. :) I would prefer we update all the API's at once if we're going to do it vs having some functions under the `DDTrace/` namespace and others with a `dd_trace_` prefix. But I'm totally +1 on the namespace idea for a future PR. :)

src/ext/dispatch.h
SammyK: 🤔 I like disambiguating things. 💛 What about `run_as_postprocess`? Or some other suggestion? :)

src/ext/ddtrace.c
SammyK: 🕷 👍 😆

src/ext/span.c
```diff
@@ -0,0 +1,94 @@
...
+    object_init_ex(stack->span_data, ddtrace_ce_span_data);
+
+    // Peek at the active span ID before we push a new one onto the stack
+    stack->parent_id = ddtrace_peek_span_id(TSRMLS_C);
+    stack->span_id = ddtrace_push_span_id(TSRMLS_C);
+    // Set the trace_id last so we have ID's on the stack
+    stack->trace_id = DDTRACE_G(root_span_id);
+    stack->duration = 0;
+    stack->exception = NULL;
+    stack->start = _get_nanoseconds();
```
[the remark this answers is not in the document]
SammyK: Wow! Nice catch! 🔝 If you don't mind checking the refactored version to make sure all is well. 😸

src/ext/dispatch_compat_php5.c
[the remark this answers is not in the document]
pawelchcki: I think logging to ddtrace_log_errf() is ok - which for now  itself is just a wrapper for: php_log_err
those checks fail on complete failure, so it seems ok to log it there.

src/ext/dispatch_compat_php7.c
pawelchcki: >Args passed by reference will be in their post-modified state in the tracing closure. Args passed by value will be in the original state before any type coercion. (See the PR description under the Arg 2: array $args header.)
>The sandboxed closures won't allow us to inject headers, but since Guzzle uses curl under the hood, we can provide a C-level function handler override for curl_exec to inject any headers.

My point is that because of these issues + some unforeseen things, we might not be able to remove dd_trace - not in a foreseeable future anyway.

I'm just pointing that this comment makes some definite statements, that might or might not be true in the future. Its better to have that discussion elsewhere than in source code.

additionally comments are not right place to track items that need doing issues and cards work much better in that regard.

src/ext/ddtrace.c
pawelchcki: Just keep in mind, that we do need to have some backwards compatibilty, since those are publicly accessible functions, and we **can not break the contract** from version to version.

Simply introducing a namespace at this point you would avoid the need to also transition these new functions without declaring them twice during transition period. 
But ok, it was only an optional suggestion.

src/ext/dispatch.h
pawelchcki: `run_as_postprocess` sounds good!

src/ext/trace.c
```diff
@@ -0,0 +1,101 @@
...
+#include "span.h"
+
+/* Move these to a header if dispatch.c still needs it */
+#if PHP_VERSION_ID >= 70100
+#define RETURN_VALUE_USED(opline) ((opline)->result_type != IS_UNUSED)
+#else
+#define RETURN_VALUE_USED(opline) (!((opline)->result_type & EXT_TYPE_UNUSED))
+#endif
+
+/* Why did we redef this? */
```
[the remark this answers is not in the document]
pawelchcki: Just to clarify, if it compiles and runs all tests without this redef, then this redef can be removed: )

src/ext/dispatch_compat_php7.c
SammyK: Very good points! Thanks @pawelchcki! :)

src/ext/trace.c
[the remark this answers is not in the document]
SammyK: I had to add this redef to get it to compile on PHP 5. We'll probably need to dig deeper into refactoring this a bit later. :)

src/ext/random.c
```diff
@@ -5,24 +5,63 @@
...
+        DDTRACE_G(span_ids_top) = stack->next;
+        efree(stack);
+    }
+}
+
+uint64_t ddtrace_push_span_id(TSRMLS_D) {
+    ddtrace_span_ids_t *stack = ecalloc(1, sizeof(ddtrace_span_ids_t));
     // We shift one bit to get 63-bit
-    return (long long)(genrand64_int64() >> 1);
+    stack->id = (uint64_t)(genrand64_int64() >> 1);
```
pawelchcki: Looking at this function again in relation to another previouse comment. Seems to me that we need to reroll or handle the case when genrand64_int64 returns 0. 

Since `0` is handled as a special value elsewhere we need to to at least add 1 or something along the lines of 

```c
id = (uint64_t)((genrand64_int64()-1) >> 1) + 1);
```

To avoid those one in trillion chances of something going wrong :D

SammyK: Brilliant! 🔝

PR #493, opened 2019-07-03
Improve synchronous flushing on PHP vm shutdown

pawelchcki:
### Description

This PR improves flushing on VM shutdown. it reintroduces safe at_exit shutdown hook to run when php vm exits without unloading modules.

And its add function to restart thread after process has been forked.

Additionally new configuration option `DD_TRACE_SHUTDOWN_TIMEOUT` to ensure PHP will shutdown in a controlled manner when data needs to be flushed to the agent on the Process shutdown.

### Readiness checklist


.circleci/config.yml
```diff
@@ -303,6 +303,33 @@ jobs:
...
+      - <<: *STEP_EXPORT_CI_ENV
+      - <<: *STEP_WAIT_AGENT
+      - run: |
+          set -xe
+          curl -q -o hyperfine.tgz -L https://github.com/sharkdp/hyperfine/releases/download/v1.6.0/hyperfine-v1.6.0-x86_64-unknown-linux-musl.tar.gz
+          tar xf hyperfine.tgz --strip-component=1
+          php -m | grep ddtrace
+          sudo rm /usr/local/etc/php/conf.d/ddtrace.ini
+          php -m | grep ddtrace || true
+          ./hyperfine -s basic 'php -m' 'php -m -dextension=ddtrace.so' --min-runs 10000 --export-markdown hyperfine.md
```
SammyK: It would be interesting to see if we could run these with **tests/ext/dd_trace_forward_call_for_functions.phpt** or some other script. And for the second command, we could add `-dddtrace.request_init_hook=<path/to>/bridge/dd_wrap_autoloader.php` to include the `request_init_hook` in the measurement.

This would be an improvement in a separate PR, but I love that you added this! :)

src/DDTrace/Tracer.php
```diff
@@ -406,7 +395,9 @@ public function getTracesAsArray()
             }
 
             $tracesToBeSent[] = $traceToBeSent;
-            unset($this->traces[$traceToBeSent[0]['trace_id']]);
+            if (!(empty($traceToBeSent) || empty($this->traces))) {
```
SammyK: Since `unset()` is a language construct, it will not raise a warning for unset vars so we don't need to check for `empty($this->traces)`. But for the offset, we can check that with `isset()`. :)

```suggestion
            if (isset($traceToBeSent[0]['trace_id'])) {
```

src/ext/coms.c
```diff
@@ -6,13 +6,15 @@
...
 #include "coms.h"
+#include "coms_curl.h"
 #include "env_config.h"
 #include "mpack.h"
 #include "vendor_stdatomic.h"
 
 typedef uint32_t group_id_t;
 
 #define GROUP_ID_PROCESSED (1UL << 31UL)
+#define MEMORY_PRESSURE_LEVEL (0.8)
```
SammyK: Is there a reason why we're going for 80% vs some other percentage? Maybe we could add a comment to clarify? :)

.circleci/config.yml
pawelchcki: Definately agree - lets  expand on this in more PRs!

src/ext/coms.c
pawelchcki: I've made this value configurable, and added comment clarifying the arbitrary nature of it :D

src/ext/coms_curl.c
```diff
@@ -74,135 +97,274 @@ inline static struct timespec deadline_in_ms(uint32_t ms) {
...
-#else  //__clang__
-static struct _writer_loop_data_t global_writer = {.mutex = PTHREAD_MUTEX_INITIALIZER,
-                                                   .condition = PTHREAD_COND_INITIALIZER};
-#endif
+static void (*ptr_at_exit_callback)(void) = 0;
 
-inline static struct _writer_loop_data_t *get_writer() { return &global_writer; }
+static void at_exit_callback() {
+    ddtrace_coms_flush_shutdown_writer_synchronous();
+    fprintf(stderr, "X");
```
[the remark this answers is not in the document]
pawelchcki: Whoops :D removed :) Thanks for spotting it!

src/ext/coms_curl.c
```diff
@@ -74,135 +97,274 @@ inline static struct timespec deadline_in_ms(uint32_t ms) {
...
 
-        curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
-        curl_easy_setopt(curl, CURLOPT_UPLOAD, 1);
-        curl_easy_setopt(curl, CURLOPT_INFILESIZE, 10);
-        curl_easy_setopt(curl, CURLOPT_VERBOSE, 0L);
+        curl_easy_setopt(writer->curl, CURLOPT_READFUNCTION, ddtrace_coms_read_callback);
+        curl_easy_setopt(writer->curl, CURLOPT_WRITEFUNCTION, dummy_write_callback);
+    }
+
+    CURL *curl = writer->curl;
```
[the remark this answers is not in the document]
pawelchcki: Agree! cleaned it up

src/ext/ddtrace.c
```diff
@@ -410,37 +418,15 @@ static PHP_FUNCTION(dd_tracer_circuit_breaker_info) {
...
     }
-    zval *group_id = NULL, *trace_array = NULL;
+    zval *trace_array = NULL;
 
-    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "za", &group_id, &trace_array) ==
-            FAILURE ||
-        Z_TYPE_P(group_id) != IS_LONG) {
+    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "a", &trace_array) == FAILURE) {
         if (DDTRACE_G(strict_mode)) {
             zend_throw_exception_ex(spl_ce_InvalidArgumentException, 0 TSRMLS_CC, "Expected group id and an array");
```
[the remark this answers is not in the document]
SammyK: Not a big deal, but I think this is the only change missing. :)

PR #496, opened 2019-07-10
Increase resource_class for web integration jobs and fix clang-format-6.0 being removed from image

pawelchcki:
### Description

Some of our tests started to exhaust all available memory when installing code through composer.
This PR fixes that

Additionally clang-format stopped working due to being removed from Debian repos. - newer version needed to reformat the files putting module header at the top, which caused some compilation configurations to break

### Readiness checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -351,14 +351,14 @@ jobs:
       - run:
           shell: /bin/bash -eo pipefail -l
           name: Build extension basic .so
-          command: make all CFLAGS="-std=c11 -O2 -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
+          command: make all CFLAGS="-std=gnu11 -O2 -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
       - run:
           name: Copy extension basic .so
           command: cp tmp/build_extension/.libs/ddtrace.so extensions/ddtrace-<< parameters.so_suffix >>.so
       - run:
           shell: /bin/bash -eo pipefail -l
           name: Build extension debug .so
-          command: make clean all CFLAGS="-std=c11 -g -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
+          command: make clean all CFLAGS="-std=gnu11 -g -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
       - run:
           name: Copy extension basic .so
           command: cp tmp/build_extension/.libs/ddtrace.so extensions/ddtrace-<< parameters.so_suffix >>-debug.so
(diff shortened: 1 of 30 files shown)


src/ext/ddtrace.c
```diff
@@ -349,19 +350,19 @@ static PHP_FUNCTION(dd_trace_dd_get_memory_limit) {
...
+    RETURN_LONG(ddtrace_get_memory_limit(TSRMLS_C));
 }
 
 /* {{{ proto bool dd_trace_check_memory_under_limit() */
 static PHP_FUNCTION(dd_trace_check_memory_under_limit) {
     PHP5_UNUSED(return_value_used, this_ptr, return_value_ptr, ht);
     PHP7_UNUSED(execute_data);
 
-    static zend_long limit = -1;
+    static int64_t limit = -1;
```
morrisonlevi: Is there a reason to avoid zend_long here? I think PHP publishes `Zend/zend_long.h`, so we can just use `zend_long` from PHP for our default integer types and we wouldn't have to worry about integer type mismatches at all.

pawelchcki: @morrisonlevi good point - looking at https://github.com/DataDog/dd-trace-php/pull/496/files#diff-044fd8e3f6c94a464fabd5a901bbcf44L43 it seems that we had to override it for compatibility with 5.x - having looked at 5.4 I can only confirm that only `zend_long64` is defined there.

morrisonlevi: Prior to version 7, PHP used "long", which means it was only 32 bit on Windows but 64 bit in most other places. Here's the definition for a zval in PHP 5.4, for example: https://github.com/php/php-src/blob/PHP-5.4/Zend/zend.h#L307-L316.

.circleci/config.yml
```diff
@@ -340,14 +351,14 @@ jobs:
...
-          command: make all CFLAGS="-std=c11 -O2 -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
+          command: make all CFLAGS="-std=gnu11 -O2 -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
       - run:
           name: Copy extension basic .so
           command: cp tmp/build_extension/.libs/ddtrace.so extensions/ddtrace-<< parameters.so_suffix >>.so
       - run:
           shell: /bin/bash -eo pipefail -l
           name: Build extension debug .so
-          command: make clean all CFLAGS="-std=c11 -g -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
+          command: make clean all CFLAGS="-std=gnu11 -g -Wall -Wextra <<# parameters.catch_warnings >> -Werror <</ parameters.catch_warnings >>" ECHO_ARG="-e"
```
morrisonlevi: Is c11 -> gnu11 related to these changes?

pawelchcki: This change was made because changing ordering has broken how `macros.h` is loaded causing this to stop compiling on older GCCs .

src/ext/ddtrace.c
pawelchcki: At the moment we don't support 32 bit - however casting from 32 bit to 64 bit on a signed integer should "still work"(tm)

pawelchcki:
Thanks for reviews!

PR #497, opened 2019-07-11
Prepare files for new release process

morrisonlevi:
The version in these files is now 'nightly'. The package.xml file has placeholders for `${date}`, `${notes}`, and `${version}` which will need to get replaced at release time as well. I picked the `${}` syntax as it seems to be used by a few other projects that use placeholders in XML, such as Maven, so maybe there will be available tools to help us make releases if needed.

The new release process should be more in line with how other the other tracers are doing releases.

### Readiness checklist


package.xml
```diff
@@ -10,10 +10,10 @@
...
-    <date>2019-07-02</date>
+    <date>${date}</date>
     <version>
-        <release>0.29.0</release>
-        <api>0.29.0</api>
+        <release>${version}</release>
+        <api>${version}</api>
     </version>
     <stability>
         <release>stable</release>
```
labbati: Can we update also `<notes></notes>` a few lines down  this line please? :)

labbati: @morrisonlevi just a reminder about this suggestion of replacing the content of `<note></note>` with `${note}`.

.circleci/config.yml
```diff
@@ -412,7 +412,7 @@ workflows:
               only: /(^build$)|(^[v]?[0-9]+(\.[0-9]+)*$)/
             branches:
               # Always build on master
-              only: /^.*(master|build).*$/
+              only: /^.*(master|[0-9]+.[0-9]+.[0-9]+).*$/
```
labbati: Now that I am looking into it, I wonder why we decided to add the leading and trailing `.*`. Maybe @pawelchcki or @SammyK have a better memory. Unless any specific reason let's remove it.

pawelchcki: Thanks for spotting this @labbati , I'm using `build` suffix to a branch name to trigger all the package builds whenever I'm testing changes that could affect that step.

If possible I'd like to retain that ability, or have some other way to trigger those builds.

src/DDTrace/Tracer.php
```diff
@@ -29,7 +29,7 @@ final class Tracer implements TracerInterface
     /**
      * @deprecated Use Tracer::version() instead
      */
-    const VERSION = '0.29.0'; // Update ./version.php too
+    const VERSION = 'nightly'; // Update ./version.php too
```
pawelchcki: I like the idea of nightly builds. 
One question: how would we differentiate between different nightly builds across time? 

I.e. a `nightly` version from january will be completely different than `nigthly` from December. 
I was thinking just adding a `nightly` suffix to last released version could work  e.g.  `0.29.0-nigthly`  or maybe some suffix like git sha would as well ?  WDYT?

.circleci/config.yml
labbati: Ah sure I think that this was only there from back in the days when we wanted to trigger builds any time. Then @morrisonlevi in the end the final version will be something like

```
only: /^.*(master|build|[0-9]+.[0-9]+.[0-9]+).*$/
```

morrisonlevi:
I couldn't find `<note>`; I assumed you meant `<notes>` and pushed that. Let me know if that's wrong there.

Not sure what to do about "nightly". I see the value in knowing something more than that, but if `0.29.0` has already been released, then `0.29.0-nightly` seems weird as the stuff after `-` is (almost?) always prelease info. Perhaps bump one of the numbers and then add `-dev`? e.g. `0.30.0-dev`.

package.xml
labbati: @morrisonlevi sorry I meant `<notes></notes>`

src/DDTrace/Tracer.php
labbati: So we had some back and forth on this. And if you  agree this what can be done,  considering that from now on we will no be releasing from master but from a release branch.

1. We update version to `nightly`
2. On a different PR we will create a simple  script to be run in CI/GH action that every time a PR is merged to master it updates the version to be `nightly-20190708-150000` to mark nightly built at 8/7/2019 at 3PM  UTC.

Everybody agree?

PR #499, opened 2019-07-17
Tests from PHP

pawelchcki:
### Description

Add php test suite to test suite

### Readiness checklist
- [ ] ~[Changelog entry](docs/changelog.md) added, if necessary~


.circleci/config.yml
```diff
@@ -276,6 +290,55 @@ jobs:
...
+      - <<: *STEP_EXPORT_CI_ENV
+      - run:
+          name: Run tests
+          command: |
+            docker-php-source extract
+            if [[ ! "<<parameters.xfail_list>>" == "none" ]]; then
+              cp "<<parameters.xfail_list>>" /usr/src/php/xfail_tests.list
+              (
+                cd /usr/src/php
+                cat xfail_tests.list | xargs -n 1 sed -i -e $'s/\(--FILE.*--\)/--XFAIL--\\\nvar_dump missmatch\\\n\\1/g'
```
SammyK: 🔝

dockerfiles/ci/Dockerfile_buster
```diff
@@ -0,0 +1,444 @@
+ARG VERSION=v7_3
```
SammyK: Do we need to [include the license](https://github.com/docker-library/php/blob/master/LICENSE) with these files? :)

dockerfiles/ci/xfail_tests/5.4.list
```diff
@@ -0,0 +1,244 @@
+tests/run-test/test005.phpt
```
SammyK: How were these files generated? Do we have a method for keeping them up to date? :)

pawelchcki: Those were generated after manually looking at the output of failed tests. And comparing them to a baseline. 
E.g. 
SSL tests have failed with extension being turned off.
Some tests failed because `var_dump` output has changed because of new objects in the object stack.

Also yesterday while disabling those I noticed that there are some (quite a few) tests in 7.1 and 7.2 that I missed previously that seem to need more investigation.

About updating them - the end goal I believe should be that we make them all pass (var_dump ones probably for the last). 

Right now they can serve as a baseline and a safeguard against regressions.

dockerfiles/ci/Dockerfile_buster
pawelchcki: Good point, we have to start adding those elsewher. I think the original file didn't have the copyright notice. But we need to mark the copyrights somewhere. I'll add that to the PR

pawelchcki: https://github.com/DataDog/dd-trace-php/pull/516

PR #500, opened 2019-07-17
Cleanup for July 2019

morrisonlevi:
### Description

These are minor issues that I noticed while familiarizing myself with the extension. I left every change as an individual commit in case we need to back single changes out during review.

### Readiness checklist

### DIFF
--- a/src/ext/dispatch.c
+++ b/src/ext/dispatch.c
@@ -232,6 +232,11 @@ static int is_anonymous_closure(zend_function *fbc, ddtrace_lookup_data_t *looku
     if (!(fbc->common.fn_flags & ZEND_ACC_CLOSURE) || !lookup->function_name) {
         return 0;
     }
+
+    /* This checks for a "{closure}" prefix, not a complete string. PHP adds
+     * null characters to the closure name to separate different parts, which
+     * is why this works at all. */
+    /* todo: why do we do this check at all? */
 #if PHP_VERSION_ID < 70000
     if (lookup->function_name_length == 0) {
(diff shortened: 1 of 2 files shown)


src/ext/dispatch.c
```diff
@@ -557,7 +536,8 @@ void ddtrace_class_lookup_release(ddtrace_dispatch_t *dispatch) {
         efree(dispatch);
     }
 }
-int find_method(zend_class_entry *ce, zval *name, zend_function **function) {
+
+static int _find_method(zend_class_entry *ce, zval *name, zend_function **function) {
```
pawelchcki: Really like the `_undersocore` prefix for static methods!

src/ext/dispatch.c
```diff
@@ -60,18 +45,9 @@ static ddtrace_dispatch_t *lookup_dispatch(const HashTable *lookup, ddtrace_look
...
 #else
 static ddtrace_dispatch_t *lookup_dispatch(const HashTable *lookup, ddtrace_lookup_data_t *lookup_data) {
-    zend_string *to_free = NULL, *key = lookup_data->function_name;
-
-    if (!ddtrace_is_all_lower(key)) {
-        key = zend_string_tolower(key);
-        to_free = key;
-    }
-
+    zend_string *key = zend_string_tolower(lookup_data->function_name);
```
pawelchcki: Followed the code of `zend_string_tolower_ex` Looks like the case of lowercased bas string is already handled well there. Great find!

PR #505, opened 2019-07-23
Address segmentation fault caused by accessing lookup data once the extension data was freed

pawelchcki:
### Description

In some cases PHP code could be run after extension destructor has run, causing a segmentation fault when accessing lookup data from within PHP code. This PR changes the code to use pointers to hashtables and sets them to NULL to denote they're not to be used. Additional checks have been added to lookup code to make sure we check the status of those pointers before accessing them. 

Additionally this PR includes fixes for:
- open basedir request init hook limitation that caused the request init hook not to be loaded when basedir restriction was in place causing fatal errors.
- fixes our autoloader overrides errors found when testing against PHP tests. 

### Readiness checklist


bridge/functions.php
```diff
@@ -35,13 +35,19 @@ function dd_tracing_enabled()
 /**
  * Checks if any of the provided classes exists.
  *
- * @param string[] $sentinelClasses
+ * @param string[]|string $sentinelClasses
  * @return bool
  */
-function any_class_exists(array $sentinelClasses)
+function any_class_exists($sentinelClasses)
```
SammyK: I can't find where `any_class_exists()` is being invoked from. Where is that? :)

Also - instead of changing the method signature, could we just invoke it differently with `any_class_exists(['FooClass'])`?

bridge/functions.php
```diff
@@ -112,9 +118,13 @@ function dd_wrap_autoloader()
     $composerAutogeneratedClass = null;
 
     dd_trace('spl_autoload_register', function () use (&$composerAutogeneratedClass) {
-
         $originalAutoloaderRegistered = dd_trace_forward_call();
-        list($loader) = func_get_args();
+        $args = func_get_args();
+        if (sizeof($args) == 0) {
```
SammyK: Nice change. Just a nit-pick - could we use the non-alias version of this function and do strict comparison? :)

```suggestion
        if (count($args) === 0) {
```

src/ext/ddtrace.c
```diff
@@ -70,7 +69,6 @@ static PHP_MINIT_FUNCTION(ddtrace) {
     ddtrace_initialize_config(TSRMLS_C);
 
     ddtrace_install_backtrace_handler();
-    ddtrace_dispatch_init(TSRMLS_C);
```
SammyK: 😱 Very nice catch! ❤️

src/ext/dispatch.c
```diff
@@ -474,7 +474,8 @@ int default_dispatch(zend_execute_data *execute_data TSRMLS_DC) {
 
 int ddtrace_wrap_fcall(zend_execute_data *execute_data TSRMLS_DC) {
     DD_PRINTF("OPCODE: %s", zend_get_opcode_name(EX(opline)->opcode));
-    if (DDTRACE_G(disable) || DDTRACE_G(disable_in_current_request)) {
+    if (DDTRACE_G(disable) || DDTRACE_G(disable_in_current_request) || DDTRACE_G(class_lookup) == NULL ||
```
SammyK: 🔝

src/ext/request_hooks.c
```diff
@@ -126,8 +132,9 @@ int dd_execute_php_file(const char *filename TSRMLS_DC) {
     zend_file_handle file_handle;
     zend_op_array *new_op_array;
     zval result;
-    int ret;
-
+    int ret, rv = FALSE;
+    char *original_open_basedir = PG(open_basedir);
+    PG(open_basedir) = NULL;
```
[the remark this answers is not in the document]
pawelchcki: Its all thanks to PHP tests :)

bridge/functions.php
pawelchcki: Indeed these seem to be from older code, probably it usage was removed recently. Will double check the tests results maybe its possible I "fixed" wrong function 😮

src/ext/ddtrace.c
pawelchcki: Found it while debugging why phpdbg fails surprised me it haven't failed sooner. Or maybe it did 😨

bridge/functions.php
pawelchcki: Oh I misread the error - assumed that it was related to the other autoloader errror - however the failing tests was simply executing every defined function 😮

pawelchcki:
Thanks @SammyK for the review, I've addressed all suggestions. Should be ready for next review round.

PR #506, opened 2019-07-23
WIP: Run request init hook on first function call instead of request initializer

pawelchcki:
### Description

Running PHP code in module request initialization can lead to undefined behaviour in PHP. This PR moves that execution to first executed function.

### Readiness checklist


morrisonlevi:
Closing as we moved to `auto_prepend_file` instead. We plan to eventually move all this down to the extension.

PR #511, opened 2019-07-25
Fix in-tree builds

morrisonlevi:
This allows checking out the repo into ext/ddtrace of PHP's source, running the PHP's buildconf script, configuring with --enable-ddtrace, and it will build ddtrace as part of PHP. This is useful for development reasons. For instance, we can use `run-tests.php` with its `--asan` flag, which does not work correctly for out-of-tree builds.

Note that there are several ZTS fixes. I am unsure why the out of tree builds did not trigger these issues, but they were definitely required for in tree builds to work.

I haven't added any tests for ensuring the in-tree build works yet, but it has worked locally on 5.6 and 7.4 branches.


morrisonlevi:
Going to close this and open new PRs for different bits.

PR #517, opened 2019-07-30
WIP: generate configuration.csv

pawelchcki:
### Readiness checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 **/vendor/
 tmp/build_extension/
 composer.lock
+docs/generate/c_config_csv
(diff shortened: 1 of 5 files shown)


morrisonlevi:
The idea here is great. At one point we discussed using something like this to have every environment variable also controllable by an ini file; generating those from a single source similar to what's going on here would be great.

krakjoe:
I would have argued at the time that we should not own documentation problems in source code in this way, but that's kinda beside the point now.

I'm going to close this as it would seem stale and I think we moved on from here.

If I'm wrong to do that please open a new PR, and accept my apologies in advance :)

PR #520, opened 2019-08-02
Default request_init_hook path is computed based on extension directory location.

pawelchcki:
### Description

This PR changes how `request_init_hook` is handled by default.
Previously if `request_init_hook` was not specified or it was pointing to a missing file it wasn't loaded.
Now we will attempt to find `dd_wrap_autoloader.php` in a few locations:
if `ddtrace*.so` file is located in `/opt/datadog-php/extensions/` we will attempt to load the file from `/opt/datadog-php/dd-trace-sources/bridge/dd_wrap_autoloader.php`

If `ddtrace*.so` path contains `tmp/build_extension/modules` we will attempt to load this file from 
`/../../../bridge/dd_wrap_autoloader.php` relative to `so` file location.

in other cases we will attempt to load the file from
`/../dd-trace-sources/bridge/dd_wrap_autoloader.php` relative to `so` file location.

### Readiness checklist


src/ext/request_hooks.c
```diff
@@ -68,6 +68,54 @@ int dd_no_blacklisted_modules(TSRMLS_D) {
...
+        return FALSE;
+    } else {
+        return memcmp(with, str, with_len) == 0;
+    }
+}
+
+static inline void _resolve_path(const char *path, size_t path_len, const char *token, char *resolved_path) {
+    size_t length = path_len + strlen(token) + 1;
+    char *tmp_path = calloc(length, 1);
+    int last_slash = strrchr(path, '/') - path;
```
labbati: [Optional] I'd rename it to `last_slash_index`.

src/ext/request_hooks.c
```diff
@@ -68,6 +68,54 @@ int dd_no_blacklisted_modules(TSRMLS_D) {
...
+
+    if (_starts_with("/opt/datadog-php/extensions/", path, path_len)) {
+        strcpy(resolved_path, "/opt/datadog-php/dd-trace-sources/bridge/dd_wrap_autoloader.php");
+    } else if (strstr(path, "tmp/build_extension/modules") != NULL) {
+        _resolve_path(path, path_len, "/../../../bridge/dd_wrap_autoloader.php", resolved_path);
+    } else {
+        _resolve_path(path, path_len, "/../dd-trace-sources/bridge/dd_wrap_autoloader.php", resolved_path);
+    }
+
+    if (access(resolved_path, F_OK) != -1) {
```
labbati: [Optional] I noticed a strange behavior of `realpath` when folders somewhere in the path do not exist (i.e. we get `errno = EEXIST`), where the resulting buffer gets written up to the existing path. While probably here would not be the case, it might be safer to check that `resolved_path` actually ends with `dd_wrap_autoloader.php` also.

src/ext/ddtrace.c
```diff
@@ -157,9 +157,16 @@ static PHP_RINIT_FUNCTION(ddtrace) {
     ddtrace_init_span_stacks(TSRMLS_C);
     ddtrace_coms_on_pid_change();
 
-    if (DDTRACE_G(request_init_hook)) {
-        DD_PRINTF("%s", DDTRACE_G(request_init_hook));
+    if (DDTRACE_G(request_init_hook) && access(DDTRACE_G(request_init_hook), F_OK) == 0) {
+        DD_PRINTF("Loading request init hook: %s", DDTRACE_G(request_init_hook));
```
labbati: I think it would be useful to differentiate the two error messages a little bit more.

E.g.:
- when request init hook is specified: `Loading request init hook from user-defined location: %s`
- when request init hook is not specified: `Loading request init hook from computed default location: %s`

src/ext/request_hooks.c
```diff
@@ -68,6 +68,54 @@ int dd_no_blacklisted_modules(TSRMLS_D) {
...
+    if (dladdr(ddtrace_fetch_computed_request_init_hook_path, &dl_info) == -1) {
+        return NULL;
+    }
+    char *path = (char *)dl_info.dli_fname;
+
+    if (access(path, F_OK) == -1) {
+        return NULL;
+    }
+    size_t path_len = strlen(path);
+    char *resolved_path = calloc(PATH_MAX + 1, 1);
```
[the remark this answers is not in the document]
pawelchcki: IDK, depends on how we approach this. 
In general the approach I take is that if a function returns a pointer - the callee responsibility to free it. But before that function exits the pointer is its own, if it decides to free it and return NULL thats ok too, because no ownership is transfer this way.

I think its a somewhat common pattern, at least its been used a lot here.
But from a POV of counting mallocs and free's  (eyeballing if everything what was allocated is freed) your approach makes much more sense if both are in the same place.

one thing I don't like here, is that we need to guess if the data needs to be freed using free or efree 🤔.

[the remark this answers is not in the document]
pawelchcki: I moved all the code to request_hook.c
there started to be way too many things in the RINIT. Also most importantly it hid a bug(s) when RINIT hook would finish early when IonCube module was found.

morrisonlevi:
Our PECL package does something like this, but the whole concept of a request init hook will be going away once migration to C is complete. We are officially starting that work this quarter so I'm going to close this out.

PR #522, opened 2019-08-06
Add `X-Datadog-Trace-Count` header to C Curl requests

pawelchcki:
### Description
Add `X-Datadog-Trace-Count` to Curl headers in on C level. 

This PR includes also a benchmarking script sends a stream of Traces (each containing 50 spans) to a local tracer.

### Readiness checklist


src/ext/coms.c
```diff
@@ -497,16 +497,19 @@ void ddtrace_msgpack_group_stack_by_id(ddtrace_coms_stack_t *stack, struct _grou
 void *ddtrace_init_read_userdata(ddtrace_coms_stack_t *stack) {
     size_t total_bytes = atomic_load(&stack->bytes_written);
 
-    struct _grouped_stack_t *readstack_ptr = malloc(sizeof(struct _grouped_stack_t));
-    struct _grouped_stack_t readstack = {.position = 0, .total_bytes = total_bytes};
+    struct _grouped_stack_t *readstack = calloc(sizeof(struct _grouped_stack_t), 1);
```
morrisonlevi: Any particular reason this switched from `malloc(x) to `calloc(x, 1)`?

pawelchcki: to initialize all values to 0, before it was set somewhat manually.

tests/benchmarks/array_spans_stress.php
```diff
@@ -0,0 +1,76 @@
...
+putenv('DD_TRACE_CLI_ENABLED=true');
+putenv('DD_SPANS_LIMIT=9999999');
+putenv('DD_AGENT_HOST=ddagent_integration');
+putenv('DD_TRACE_DEBUG_CURL_OUTPUT=0');
+putenv("DD_TRACE_DEBUG_CURL_OUTPUT=0");
+putenv('DD_TRACE_BETA_SEND_TRACES_VIA_THREAD=1');
+putenv('DD_TRACE_AGENT_TIMEOUT=500');
+putenv('DD_TRACE_AGENT_CONNECT_TIMEOUT=500');
+
+function_exists('dd_trace_internal_fn') && dd_trace_internal_fn('ddtrace_reload_config');
```
morrisonlevi: Is there a reason `dd_trace_internal_fn` would not exist?

docker-compose.yml
```diff
@@ -124,16 +124,6 @@ services:
       - 9090:80
     command: sh -c "php -S 0.0.0.0:80 index.php"
 
-  agent:
```
morrisonlevi: I use this `agent` target when I need to start up an agent to send traces to Datadog so I can look around in the UI. Is there any particular reason to drop this?

tests/benchmarks/array_spans_stress.php
morrisonlevi: I noticed later in the test it isn't guarded. They should either all be guarded, or none of them should, I think.

morrisonlevi:
I am working on the background sender and will incorporate the work from this PR. Thanks, Pawel!

PR #524, opened 2019-08-07
Refactor opcode handler entry

SammyK:
### Description

In order to wrap up #491, we needed to refactor the entry into the opcode handler `ddtrace_wrap_fcall()`. The check for "should we trace this?" has been bubbled up to `ddtrace_wrap_fcall()` instead of being spread around to several function calls. There were also a few functions that we inlined which allowed us to remove extra code.

This will allow us (we think) to forward the original call at the C level and append the tracing closure cleanly for #491.

Thanks @morrisonlevi for pair programming this one with me. :)

### Readiness checklist
- ~~[ ] [Changelog entry](docs/changelog.md) added, if necessary~~

### DIFF
--- a/src/ext/dispatch.c
+++ b/src/ext/dispatch.c
@@ -30,30 +30,17 @@ ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 #define EX(x) ((execute_data)->x)
 #endif
 
-#if PHP_VERSION_ID < 70000
-static ddtrace_dispatch_t *lookup_dispatch(const HashTable *lookup, ddtrace_lookup_data_t *lookup_data) {
-    if (lookup_data->function_name_length == 0) {
-        lookup_data->function_name_length = strlen(lookup_data->function_name);
-    }
-
-    char *key = zend_str_tolower_dup(lookup_data->function_name, lookup_data->function_name_length);
+static ddtrace_dispatch_t *find_function_dispatch(const HashTable *lookup, zval *fname) {
+    char *key = zend_str_tolower_dup(Z_STRVAL_P(fname), Z_STRLEN_P(fname));
     ddtrace_dispatch_t *dispatch = NULL;
-    dispatch = zend_hash_str_find_ptr(lookup, key, lookup_data->function_name_length);
+    dispatch = zend_hash_str_find_ptr(lookup, key, Z_STRLEN_P(fname));
 
     efree(key);
     return dispatch;
 }
-#else
-static ddtrace_dispatch_t *lookup_dispatch(const HashTable *lookup, ddtrace_lookup_data_t *lookup_data) {
-    zend_string *key = zend_string_tolower(lookup_data->function_name);
-    ddtrace_dispatch_t *dispatch = zend_hash_find_ptr(lookup, key);
-    zend_string_release(key);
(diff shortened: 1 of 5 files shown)


src/ext/dispatch.c
```diff
@@ -213,215 +199,97 @@ static void execute_fcall(ddtrace_dispatch_t *dispatch, zval *this, zend_execute
     Z_DELREF(closure);
 }
 
-static int is_anonymous_closure(zend_function *fbc, ddtrace_lookup_data_t *lookup) {
```
pawelchcki: 👍 great stuff and getting rid of cargoculted checks! 
For long time I assumed that odd `{closure}` function can slip without ZEND_ACC_CLOSURE being properly set.
Thanks to your PR and our testsuite just verified that assumption to be false - by adding an assertion on the name and running tests: https://circleci.com/workflow-run/48f8d03a-cde4-45b7-9a3a-a9d88909b264
Good stuff!

src/ext/dispatch.c
```diff
@@ -457,7 +325,7 @@ static int update_opcode_leave(zend_execute_data *execute_data TSRMLS_DC) {
     return ZEND_USER_OPCODE_LEAVE;
 }
 
-int default_dispatch(zend_execute_data *execute_data TSRMLS_DC) {
+static int _default_dispatch(zend_execute_data *execute_data TSRMLS_DC) {
```
pawelchcki: 👍

src/ext/dispatch_compat_php5.c
```diff
@@ -227,4 +227,57 @@ void ddtrace_forward_call(zend_execute_data *execute_data, zval *return_value TS
...
+        return CACHED_PTR(opline->op1.literal->cache_slot);
+    } else if (EXPECTED(zend_hash_quick_find(EG(function_table), Z_STRVAL_P(fname), Z_STRLEN_P(fname) + 1,
+                                             Z_HASH_P(fname), (void **)&fbc) == SUCCESS)) {
+        return fbc;
+    } else {
+        return NULL;
+    }
+}
+
+int ddtrace_should_trace_call(zend_execute_data *execute_data, zend_function **fbc,
```
pawelchcki: Nice, much cleaner than the previous approach!
One suggestion - some time ago BOOL_T define/type was introduced and started to be used in new code. If you feel this approach has merits feel free to use BOOL_T and FALSE, TRUE defines. for functions expected to return true and false values only.

src/ext/dispatch_compat_php7.c
```diff
@@ -123,4 +123,38 @@ void ddtrace_forward_call(zend_execute_data *execute_data, zval *return_value TS
 
     zval_ptr_dtor(&fname);
 }
+
+int ddtrace_should_trace_call(zend_execute_data *execute_data, zend_function **fbc,
```
pawelchcki: Same comment as in php5 counterpart: Nice self contained function!
Consider using BOOL_T since this is expected to use boolean return values only.

src/ext/dispatch_compat_php5.c
SammyK: Thanks @pawelchcki! I like the idea of using `BOOL_T` - done! :)

Issue #510, opened 2019-07-24
PDO: Prepared statements not released/deallocated


The-Sorce:
Prepared SQL statements are automatically released by PDO on the database connection (i.e. on MySQL using `DEALLOCATE PREPARE`) when the last reference to the `PDOStatement` is unset. Under certain circumstances, dd-trace-php prevents this from happening.

For persistent/long-lived PHP processes, this will cause unexpected behavior in terms of leftover prepared statements on the database server. This can be observed directly on the database with commands such as `SHOW GLOBAL STATUS LIKE 'prepared_stmt_count';`. In specific test cases I saw this issue triggering `1461 Can't create more than max_prepared_stmt_count statements` database errors.

Environment prerequisites:
- PHP (tested with versions 5.6.40 as well as PHP 7.3.6)
- dd-trace-php (tested with version 0.29.0)
- MySQL/MariaDB database server (version N/A)
- Composer autoloader (without any specific dependencies, i.e. a freshly initialized Composer project is enough)
- PDO:s emulation of prepared statements must be disabled (i.e. the PDO option `PDO::ATTR_EMULATE_PREPARES` must be set to false)

Please see the following PHP sample code with which this issue can be recreated:
```
<?php

// A Composer autoloader (please provide your own) is for some unknown reason required to recreate the issue.
require_once __DIR__ . '/vendor/autoload.php';

// Please provide your own DB credentials. Also, modify the SQL statements below (specifically any_table, foo, bar) into
// some valid select in your specific environment.

$host = '127.0.0.1';
$port = 3306;
$dbname = 'test_db';
$username = 'mysql';
$passwd = 'secret';

$dsn = 'mysql:host='.$host.';port='.$port.';dbname='.$dbname;
$options = [PDO::ATTR_EMULATE_PREPARES => false]; // Important! PDO's preparation emulation is unaffected by this issue.
$pdo = new PDO($dsn, $username, $passwd, $options);

// Example 1: The prepared statement should be deallocated when the $prep variable is unset.
$prep = $pdo->prepare("SELECT * FROM any_table WHERE foo = :foo");
$prep->execute(['foo' => 'bar']);
unset($prep); // The last reference to the PDOStatement instance is removed

// Example 2: The prepared statement should be deallocated immediately after execution, as no reference to the
// PDOStatement instance is available anywhere.
$pdo->prepare("SELECT * FROM any_table WHERE foo = :foo")->execute(['foo' => 'bar']);

[...]

Issue #451, opened 2019-05-24
Causing PDO errors in our application


jonners99:
We are evaluating on a dev server, we have noticed an increased number of exceptions since installing dd-trace-php. All like the following: 

`[24-May-2019 10:43:52 Europe/London] PHP Fatal Error: (See Admin Error Log for full details) PDOException: SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.
{"message":"Uncaught exception","context":{"exception":{"class":"PDOException","message":"SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.","code":"HY000","file":"/opt/datadog-php/dd-trace-sources/src/DDTrace/Integrations/PDO/PDOIntegration.php:269"}},"level":500,"level_name":"CRITICAL","channel":"exception","datetime":{"date":"2019-05-24 10:43:52.530095","timezone_type":3,"timezone":"Europe/London"},"extra":[]}`

For now I have disabled monitoring of PDO, which has resolved the issue. But obviously this means we are not receiving PDO data.

PHP 5.6, MySQL 5.5, CentOS7

labbati:
Hi @jonners99 sorry for the issue. We have been made aware of a similar issue and based on an early investigation we recently performed it may be related to one of two scenarios if you are using the default 'buffered' mode. Stored procedures and queries against sphinx. Can you please provide more info about your use case?

Ideally it would be great if you can provide a code snippet that reproduces the query causing the issue so we can run/test in our local env.

In the meantime, if you still are interested in evaluating our tracer, feel free to disable PDO integration: `DD_INTEGRATIONS_DISABLED=pdo`

ananth-iyer:
@labbati Can I set this environment variable in PHP code? `DD_INTEGRATIONS_DISABLED=pdo`

labbati:
No it is too late then. Please follow these instructions: https://docs.datadoghq.com/tracing/setup/php/#configuration

carflo:
+1 
[...]

PR #536, opened 2019-08-19
Release This as soon as wrapping closure is finished and scope is left

pawelchcki:
### Description

After passing This to a closure, we were not releasing it in correct time when it was requested by the compiled opcodes. 

This PR releases This and closure object when first is requested, and when not needed anymore.

This fixes any code that expects the destructor to be called as soon as possible to free up resources (db connections etc).

### Related Issues:
https://github.com/DataDog/dd-trace-php/issues/510
https://github.com/DataDog/dd-trace-php/issues/451

### Readiness checklist

### Reviewer checklist
- [x] Appropriate labels assigned
- [x] Milestone set

### DIFF
--- a/src/ext/compatibility.h
+++ b/src/ext/compatibility.h
@@ -39,6 +39,11 @@
 #define PHP7_UNUSED(...) UNUSED(__VA_ARGS__)
 #endif
 
+#if PHP_VERSION_ID >= 70000 && PHP_VERSION_ID < 70300
+#define GC_ADDREF(x) (++GC_REFCOUNT(x))
+#define GC_DELREF(x) (--GC_REFCOUNT(x))
+#endif
+
 #if PHP_VERSION_ID < 70000
 #define ZVAL_VARARG_PARAM(list, arg_num) (*list[arg_num])
 #define IS_TRUE_P(x) (Z_TYPE_P(x) == IS_BOOL && Z_LVAL_P(x) == 1)
--- a/src/ext/dispatch.c
+++ b/src/ext/dispatch.c
@@ -8,6 +8,7 @@
 #include <ext/spl/spl_exceptions.h>
 
 #include "compat_zend_string.h"
+#include "compatibility.h"
 #include "ddtrace.h"
 #include "debug.h"
 #include "dispatch_compat.h"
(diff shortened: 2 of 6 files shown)


src/ext/dispatch.c
```diff
@@ -378,7 +387,15 @@ int ddtrace_wrap_fcall(zend_execute_data *execute_data TSRMLS_DC) {
 #endif
 
     wrap_and_run(execute_data, dispatch TSRMLS_CC);
-
+#if PHP_VERSION_ID >= 70000
+    if (DDTRACE_G(original_context).this) {
+#if PHP_VERSION_ID < 70300
```
morrisonlevi: Maybe we should backport GC_DELREF to earlier PHP versions?

pawelchcki: I've seen both approaches in 7.3, but yeah, backporting GC_DELREF macro seems like a good way to make the code a bit cleaner.

morrisonlevi: The GC_ADDREF macro would get used in at least a few places too. Let's leave that for another PR.

pawelchcki:
@morrisonlevi Thanks for looking into it! I added a `phpt` test to test this fix and added the macros for GC ADD/DEL REF.

.gitignore
```diff
@@ -6,7 +6,7 @@ composer.lock
 .libs
 .deps
 **.code-workspace
-Makefile*
+/Makefile*
```
morrisonlevi: What's the point of this change?

src/ext/compatibility.h
```diff
@@ -39,6 +39,11 @@
 #define PHP7_UNUSED(...) UNUSED(__VA_ARGS__)
 #endif
 
+#if PHP_VERSION_ID < 70300
```
morrisonlevi: This should apply only to 7.0-7.2. With the way it is written atm it will also apply to PHP 5.

.gitignore
pawelchcki: Its not related to this PR. But previous rule also caught Makefile it wasn't supposed to. I'll remove this, since its also present in Laravel PR and sneaked in here.

src/ext/compatibility.h
pawelchcki: True, actually I had other defines for 5.6 but they didn't have desired effect in 5.x. I'll make this `if` as you've suggested.

pawelchcki:
Thanks for the review!

PR #538, opened 2019-08-20
Sandboxed integration: PDO

SammyK:
### Description

This PR adds a sandboxed integration for PDO which is enabled by default (but disabled on < PHP 5.6 since the sandbox API is not supported on those versions.)

The sandboxed integration can be disabled by setting `DD_TRACE_SANDBOX_ENABLED=false` which will force the tracer to fall back to the original PDO integration.

One thing to note - the sandboxed integrations do not have support for a limited tracer but that feature can be added in a separate PR.

### Readiness checklist

### Reviewer checklist
- [ ] Appropriate labels assigned
- [ ] Milestone set

### DIFF
--- a/src/DDTrace/Configuration.php
+++ b/src/DDTrace/Configuration.php
@@ -143,7 +143,11 @@ class Configuration extends AbstractConfiguration
      */
     public function isSandboxEnabled()
     {
-        return $this->boolValue('trace.sandbox.enabled', false);
+        // Sandbox API not available on < PHP 5.6 yet
+        if (PHP_VERSION_ID < 50600) {
+            return false;
+        }
+        return $this->boolValue('trace.sandbox.enabled', true);
     }
 
     /**
--- a/src/DDTrace/Configuration.php
+++ b/src/DDTrace/Configuration.php
@@ -137,6 +137,16 @@ class Configuration extends AbstractConfiguration
     }
 
     /**
+     * Whether or not sandboxed tracing closures are enabled.
+     *
+     * @return bool
+     */
+    public function isSandboxEnabled()
+    {
+        return $this->boolValue('trace.sandbox.enabled', false);
+    }
+
+    /**
      * The name of the application.
      *
      * @param string $default
--- a/src/DDTrace/Integrations/PDO/PDOIntegration.php
+++ b/src/DDTrace/Integrations/PDO/PDOIntegration.php
@@ -387,14 +387,15 @@ class PDOIntegration extends Integration
 
     public static function storeConnectionParams($pdo, array $constructorArgs)
     {
+        $hash = is_object($pdo) ? spl_object_hash($pdo) : '';
         if (count($constructorArgs) > 0) {
             $tags = self::parseDsn($constructorArgs[0]);
             if (isset($constructorArgs[1])) {
                 $tags['db.user'] = $constructorArgs[1];
             }
-            self::$connections[spl_object_hash($pdo)] = $tags;
+            self::$connections[$hash] = $tags;
         } else {
-            self::$connections[spl_object_hash($pdo)] = array();
+            self::$connections[$hash] = array();
         }
     }
 
@@ -404,15 +405,15 @@ class PDOIntegration extends Integration
             // When an error occurs 'FALSE' will be returned in place of the statement.
             return;
         }
-        $pdoHash = spl_object_hash($pdo);
+        $pdoHash = is_object($pdo) ? spl_object_hash($pdo) : '';
         if (isset(self::$connections[$pdoHash])) {
-            self::$statements[spl_object_hash($stmt)] = $pdoHash;
+            self::$statements[is_object($stmt) ? spl_object_hash($stmt) : ''] = $pdoHash;
         }
     }
 
     public static function setConnectionTags($pdo, $span)
     {
-        $hash = spl_object_hash($pdo);
+        $hash = is_object($pdo) ? spl_object_hash($pdo) : '';
         if (!isset(self::$connections[$hash])) {
             return;
         }
@@ -424,7 +425,7 @@ class PDOIntegration extends Integration
 
     public static function setStatementTags($stmt, $span)
     {
-        $stmtHash = spl_object_hash($stmt);
+        $stmtHash = is_object($stmt) ? spl_object_hash($stmt) : '';
         if (!isset(self::$statements[$stmtHash])) {
             return;
         }
--- a/src/DDTrace/Integrations/PDO/PDOIntegration.php
+++ b/src/DDTrace/Integrations/PDO/PDOIntegration.php
@@ -84,11 +84,15 @@ class PDOIntegration extends Integration
             $thrown = null;
             try {
                 dd_trace_forward_call();
-                PDOIntegration::storeConnectionParams($this, func_get_args());
                 PDOIntegration::detectError($span, $this);
             } catch (\Exception $e) {
                 PDOIntegration::setErrorOnException($span, $e);
                 $thrown = $e;
+            }
+
+            if (!$thrown || $thrown instanceof \PDOException) {
+                PDOIntegration::storeConnectionParams($this, func_get_args());
+                PDOIntegration::setConnectionTags($this, $span);
             }
 
             $scope->close();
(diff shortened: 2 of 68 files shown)


tests/Common/SpanIntegrationChecker.php
```diff
@@ -16,6 +17,10 @@ final class SpanIntegrationChecker
     public function defineIntegrationsByPattern()
     {
         // <regex pattern> => <integration class>
+        $pdoIntegration = 'DDTrace\Integrations\PDO\PDOIntegration';
```
morrisonlevi: Consider ternary instead?

src/DDTrace/Configuration.php
```diff
@@ -150,6 +150,20 @@ public function isHttpClientSplitByDomain()
         return $this->boolValue('trace.http.client.split.by.domain', false);
     }
 
+    /**
+     * Whether or not sandboxed tracing closures are enabled.
+     *
+     * @return bool
+     */
+    public function isSandboxEnabled()
```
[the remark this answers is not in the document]
SammyK: The postprocess name was only for the [C-level dispatcher flag](https://github.com/DataDog/dd-trace-php/blob/master/src/ext/dispatch.c#L350). I'm happy to rename if we want to discuss options. We're calling it sandbox everywhere else in the codebase so we'd have to consider renaming it everywhere else "sandbox" is referenced as well. :)

src/ext/serializer.c
```diff
@@ -224,6 +224,12 @@ static void _serialize_meta(zval *el, ddtrace_span_t *span TSRMLS_DC) {
     }
 
     _serialize_exception(el, meta, span TSRMLS_CC);
+    if (!span->exception && zend_hash_exists(Z_ARRVAL_P(meta), "error.msg", sizeof("error.msg"))) {
+        add_assoc_long(el, "error", 1);
```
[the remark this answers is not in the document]
SammyK: It is, but this check only happens when there was no exception thrown. This allows non-exception errors in integrations from userland to set `$span->meta['error.msg'] = "Foo error";` and the tracer will automatically flag the span as an error span at this point. :)

src/ext/serializer.c
```diff
@@ -263,22 +269,31 @@ static void _serialize_exception(zval *el, zval *meta, ddtrace_span_t *span) {
...
 }
 
 void _serialize_meta(zval *el, ddtrace_span_t *span) {
     zval meta_zv, *meta = _read_span_property(span->span_data, "meta", sizeof("meta") - 1);
 
-    if (!meta || Z_TYPE_P(meta) != IS_ARRAY) {
-        meta = &meta_zv;
-        array_init(meta);
+    if (meta && Z_TYPE_P(meta) == IS_ARRAY) {
+        ZVAL_DUP(&meta_zv, meta);
```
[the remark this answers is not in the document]
SammyK: If you don't use `ZVAL_DUP()`, PHP 7 complains about a COW violation. Don't have a 🐮 PHP 7! 😄

tests/Common/SpanIntegrationChecker.php
SammyK: Ternary will make us check `Configuration::get()->isSandboxEnabled()` for every integration we add sanboxing for. I chose this way so that we can just group all the sandboxed integrations into one `if` block. :)

tests/Integrations/PDO/PDOSandboxedTest.php
```diff
@@ -0,0 +1,398 @@
...
+                // phpcs:enable
+                ->withExactTags($this->baseTags()),
+            SpanAssertion::build('PDO.prepare', 'PDO', 'sql', $query)
+                ->withExactTags($this->baseTags()),
+            SpanAssertion::exists('PDO.__construct'),
+        ]);
+    }
+
+    // Add limited tracer to sandboxed closures
+    /*
```
pawelchcki: This test / and its functionality needs to be implemented . Since PDO can create tens of thousands Spans easily, this mode ensures that we stop creating spans when configured limit is reached. In future the limit might be increased but in general it should be there to safeguard against excessive resource consumption.

tests/Integrations/PDO/PDOSandboxedTest.php
```diff
@@ -0,0 +1,398 @@
...
+            unset($stmt);
+
+            $pdo->prepare($query)->execute([2]);
+
+            $closedCount = $pdo->query("SHOW SESSION STATUS LIKE 'Com_stmt_close'")->fetchColumn(1);
+
+            $this->assertEquals(2, $closedCount);
+        });
+    }
+
```
pawelchcki: Some new tests are missing from this file:
https://github.com/DataDog/dd-trace-php/blob/master/tests/Integrations/PDO/PDOTest.php#L242
and
https://github.com/DataDog/dd-trace-php/blob/master/tests/Integrations/PDO/PDOTest.php#L258

tests/Integrations/PDO/PDOTest.php
```diff
@@ -16,7 +16,8 @@ final class PDOTest extends IntegrationTestCase
     public static function setUpBeforeClass()
     {
         parent::setUpBeforeClass();
-        IntegrationsLoader::load();
+        putenv('DD_TRACE_SANDBOX_ENABLED=false');
```
pawelchcki: I think ideally when porting integrations to sandboxing we should be using the exact same test suite for new and old implementation.

I don't mind refactoring - I think the new testsuite looks really great, but if possible both should share as much code as possible. Sometimes new tests or minute differences can slip with two separate test cases which might increase a chance of a regression.

WDYT about using the same test suite for both variants ? Simple inheritance could be used to make two identical sets.

src/DDTrace/Integrations/SandboxedIntegration.php
```diff
@@ -0,0 +1,40 @@
...
+
+use DDTrace\SpanData;
+use DDTrace\Tag;
+
+abstract class SandboxedIntegration extends Integration
+{
+    /**
+     * @var self
+     */
+    private static $instance;
```
labbati: Unfortunately this won't work. Being a static instance the first sandboxed integration class will define the one and one instance. So I suggest to remove this instance thing and just call the constructor in `IntegrationLoader`. WDYT?

src/DDTrace/Integrations/PDO/PDOIntegration.php
```diff
@@ -394,15 +405,15 @@ public static function storeStatementFromConnection($pdo, $stmt)
             // When an error occurs 'FALSE' will be returned in place of the statement.
             return;
         }
-        $pdoHash = spl_object_hash($pdo);
+        $pdoHash = is_object($pdo) ? spl_object_hash($pdo) : '';
```
morrisonlevi: The pattern `is_object($x) ? spl_object_hash($x) : ''` is used in at least 3 places; should we make this a helper method?

src/ext/trace.c
```diff
@@ -58,10 +58,17 @@ void ddtrace_trace_dispatch(ddtrace_dispatch_t *dispatch, zend_function *fbc,
 
     BOOL_T keep_span = TRUE;
     if (fcall_status == SUCCESS && Z_TYPE(dispatch->callable) == IS_OBJECT) {
+        zend_error_handling error_handling;
         int orig_error_reporting = EG(error_reporting);
         EG(error_reporting) = 0;
+#if PHP_VERSION_ID < 70000
+        zend_replace_error_handling(EH_SUPPRESS, NULL, &error_handling TSRMLS_CC);
+#else
+        zend_replace_error_handling(EH_THROW, NULL, &error_handling TSRMLS_CC);
```
morrisonlevi: No need for `TSRMLS_CC` on the PHP 7+ branch.

src/DDTrace/Integrations/PDO/PDOSandboxedIntegration.php
```diff
@@ -0,0 +1,245 @@
...
+            return;
+        }
+
+        $class = strtoupper(substr($errorCode, 0, 2));
+        if (in_array($class, ['00', '01', 'IM'], true)) {
+            // Not an error
+            return;
+        }
+        $errorInfo = $pdoOrStatement->errorInfo();
+        $span->meta[Tag::ERROR_MSG] = 'SQL error: ' . $errorCode . '. Driver error: ' . $errorInfo[1];
```
morrisonlevi: What is the resulting interaction between automatic exception handling and setting this meta element?

SammyK: Auto-exception handling takes precedence. If no exception is thrown, a span can be marked as an error span if `error.msg` is set from userland. [Related comment](https://github.com/DataDog/dd-trace-php/pull/538/files#r321329487). :)

src/DDTrace/Integrations/PDO/PDOIntegration.php
SammyK: Good point. I'd prefer to keep it as-is since we're already making quite a few static method calls in this integration. I also don't want to make too many changes to the existing integration. But I can totally be convinced if you really think it needs it. :)

PR #546, opened 2019-08-27
[Sandbox] Bind called object to tracing closure

SammyK:
### Description

A little oversight from #491; the sandboxed tracing closures need to access the original call's object. This PR binds `$this` to the tracing closure for non-static method calls.

This PR also ensures that static tracing closures aren't executed for non-static methods. A debug message will be emitted in this case.

The main inspiration for this check is due to [a bug in PHP 5](https://3v4l.org/Rgo87) where closures are marked as static when they are declared within a static method context. Since our integrations are loaded from a static method, `Integration::load()`, PR #538 will ensure the sandboxed integrations are loaded from a non-static method to ensure consistency on PHP 5.

*Edit:* I also went ahead and included a little refactor to prevent redefining `EX()` everywhere. I hope that's OK. :)

### Readiness checklist
- ~~[ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.~~

### Reviewer checklist
- [ ] Milestone is set.


SammyK:
@morrisonlevi Nice catch on the `$this` dtor'ing! It's due to VM cleanup but in a weird spot. I refactored it in f7e0c83 with comments so that it's hopefully more clear. :)

PR #548, opened 2019-08-29
Attach exceptions to spans

morrisonlevi:
### Description

This will attach exceptions to open spans that were traced through the new `dd_trace_function`/`dd_trace_method` API.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/dispatch_compat_php5.c
+++ b/src/ext/dispatch_compat_php5.c
@@ -16,6 +16,14 @@
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace)
 
+#if !defined(ZVAL_COPY_VALUE)
+#define ZVAL_COPY_VALUE(z, v)      \
+    do {                           \
+        (z)->value = (v)->value;   \
+        Z_TYPE_P(z) = Z_TYPE_P(v); \
+    } while (0)
+#endif
+
 static zend_always_inline void **vm_stack_push_args_with_copy(int count TSRMLS_DC) /* {{{ */
 {
     zend_vm_stack p = EG(argument_stack);
@@ -381,5 +389,13 @@ void ddtrace_execute_tracing_closure(zval *callable, zval *span_data, zend_execu
         zval_ptr_dtor(&retval_ptr);
     }
     zend_fcall_info_args_clear(&fci, 0);
+}
+
+void ddtrace_span_attach_exception(ddtrace_span_t *span, zval *exception) {
+    if (exception) {
+        MAKE_STD_ZVAL(span->exception);
+        ZVAL_COPY_VALUE(span->exception, exception);
+        zval_copy_ctor(span->exception);
+    }
 }
 #endif  // PHP 5
--- a/src/ext/serializer.c
+++ b/src/ext/serializer.c
@@ -1,10 +1,12 @@
 #include <Zend/zend.h>
 #include <Zend/zend_exceptions.h>
+#include <Zend/zend_interfaces.h>
 #include <php.h>
 
 #include <ext/spl/spl_exceptions.h>
 
 #include "ddtrace.h"
+#include "dispatch_compat.h"
 #include "mpack/mpack.h"
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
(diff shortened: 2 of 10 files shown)


src/ext/span.c
```diff
@@ -21,8 +21,12 @@ static void _free_span(ddtrace_span_t *span) {
     ddtrace_zval_ptr_dtor(span->span_data);
     efree(span->span_data);
     if (span->exception) {
+#if PHP_VERSION_ID < 70000
         ddtrace_zval_ptr_dtor(span->exception);
```
morrisonlevi: On PHP 5, `ddtrace_zval_ptr_dtor` evaluates to `zval_dtor`, which will not free the zval itself, so theoretically this would leak. I need to switch t his to `zval_ptr_dtor` and then it should be good.

src/ext/span.h
```diff
@@ -6,9 +6,13 @@
 
 #include "compatibility.h"
 
-typedef struct _ddtrace_span_t {
+typedef struct ddtrace_span_t {
```
pawelchcki: 👍

src/ext/serializer.c
```diff
@@ -173,32 +175,125 @@ int ddtrace_serialize_simple_array(zval *trace, zval *retval TSRMLS_DC) {
...
-            ZVAL_COPY(&value, prop);                                                                                 \
-            add_assoc_zval(el, (name), &value);                                                                      \
-        }                                                                                                            \
-    } while (0);
+static zval *_read_span_property(zval *span_data, const char *name, size_t name_len) {
+    zval rv;
+    return zend_read_property(ddtrace_ce_span_data, span_data, name, name_len, 1, &rv);
+}
+
+static void _add_assoc_zval(zval *el, const char *name, zval *prop) {
```
pawelchcki: I found the distinction based between `_add_assoc_zval` and `add_assoc_zval` not immediately obvious. Would it make sense to rename `_add_assoc_zval` to something more descriptive to its internals - e.g.  `_add_assoc_zval_with_copy` ?

src/ext/dispatch_compat_php5.c
```diff
@@ -382,4 +390,12 @@ void ddtrace_execute_tracing_closure(zval *callable, zval *span_data, zend_execu
     }
     zend_fcall_info_args_clear(&fci, 0);
 }
+
+void ddtrace_span_attach_exception(ddtrace_span_t *span, zval *exception) {
+    if (exception) {
```
pawelchcki: would it make sense to check  if `Z_TYPE_P(tmp) != IS_NULL)`

src/ext/serializer.c
```diff
@@ -173,32 +175,125 @@ int ddtrace_serialize_simple_array(zval *trace, zval *retval TSRMLS_DC) {
...
+    INIT_PZVAL_COPY(value, prop);
+    zval_copy_ctor(value);
+    add_assoc_zval(el, name, value);
+}
+
+void _serialize_exception(zval *el, zval *meta, ddtrace_span_t *span TSRMLS_DC) {
+    zval *exception, *msg, *stack;
+    zend_uint class_name_len;
+    const char *class_name;
+    int dup;
```
pawelchcki: What value this is set to ?

src/ext/serializer.c
```diff
@@ -173,32 +175,125 @@ int ddtrace_serialize_simple_array(zval *trace, zval *retval TSRMLS_DC) {
...
+
+static void _add_assoc_zval(zval *el, const char *name, zval *prop) {
+    zval value;
+    ZVAL_COPY(&value, prop);
+    add_assoc_zval(el, (name), &value);
+}
+
+void _serialize_exception(zval *el, zval *meta, ddtrace_span_t *span) {
+    zval exception, name, msg, stack;
+    ZVAL_OBJ(&exception, span->exception);
```
pawelchcki: Would it make sense to check if `span->exception` is`null` here as in both `span->exception == NULL && Z_TYPE_P(tmp) == IS_NULL)`?

src/ext/serializer.c
```diff
@@ -173,32 +175,125 @@ int ddtrace_serialize_simple_array(zval *trace, zval *retval TSRMLS_DC) {
...
+    zval meta_zv, *meta = _read_span_property(span->span_data, "meta", sizeof("meta") - 1);
+
+    if (!meta || Z_TYPE_P(meta) != IS_ARRAY) {
+        meta = &meta_zv;
+        array_init(meta);
+    } else {
+        Z_ADDREF_P(meta);
+    }
+
+    if (span->exception) {
```
pawelchcki: Elsewhere in the code (`ddtrace_span_attach_exception` and `free_*`) I've seen this null check delegated to called method. Would it make sense to do the same here for consistency ?

src/ext/serializer.c
```diff
@@ -173,32 +175,125 @@ int ddtrace_serialize_simple_array(zval *trace, zval *retval TSRMLS_DC) {
...
+    zval_copy_ctor(value);
+    add_assoc_zval(el, name, value);
+}
+
+void _serialize_exception(zval *el, zval *meta, ddtrace_span_t *span TSRMLS_DC) {
+    zval *exception, *msg, *stack;
+    zend_uint class_name_len;
+    const char *class_name;
+    int dup;
+    exception = span->exception;
```
pawelchcki: Same comment as for `ver > 70000`.  Would it make sense to add null check here ?

src/ext/dispatch_compat_php5.c
morrisonlevi: IIRC it's always a null pointer or an object. I can look again to be sure.

src/ext/serializer.c
morrisonlevi: It's set on line 199; I forward declare my local variables out of habit in php-src since that was a requirement for a long time (not even sure if it's still a requirement -- IIRC Windows toolchain was last holdout).

src/ext/serializer.c
morrisonlevi: On PHP 7, the exception is an object pointer, not a zval, so it's definitely not needed on 7.

src/ext/serializer.c
morrisonlevi: Good idea.

src/ext/serializer.c
pawelchcki: Ah thanks for clarification

src/ext/serializer.c
pawelchcki: :| Missed it - thanks. 
In our case we don't need to forward declare - usually I'm fine with either. 
Here the dup escaped my brain parser - so it might indicate, either broken parser, or visibility problem.

Not to bikeshed too much - I think `needs_duplication` might be more descriptive name. Especially when its a result from function named `zend_get_object_classname`. 

Both work though, so lets treat  this as optional suggestion.

morrisonlevi: I moved definition of `dup` to its initialization, which should help.

src/ext/serializer.c
morrisonlevi: I pushed this into the function.

PR #549, opened 2019-08-29
[Sandbox] Pass original call exception to tracing closure

SammyK:
### Description

This PR adds a fourth parameter to sandboxed tracing closures that contains the exception thrown from the original call. If no exception was thrown, the argument passed in will be `null`.

```php
dd_trace_function('myFunc', function (SpanData $span, array $args, $retval, $ex) {
    if ($ex instanceof FooException) {
        echo "FooException thrown!\n";
    }
});
```

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/ext/dispatch_compat_php7.c
+++ b/src/ext/dispatch_compat_php7.c
@@ -226,12 +226,17 @@ void ddtrace_copy_function_args(zend_execute_data *execute_data, zval *user_args
 }
 
 void ddtrace_execute_tracing_closure(zval *callable, zval *span_data, zend_execute_data *execute_data, zval *user_args,
-                                     zval *user_retval) {
+                                     zval *user_retval, zend_object *exception) {
     zend_fcall_info fci = {0};
     zend_fcall_info_cache fcc = {0};
     zval rv;
     INIT_ZVAL(rv);
-    zval args[3];
+    zval args[4];
+    zval exception_arg;
+    ZVAL_UNDEF(&exception_arg);
+    if (exception) {
+        ZVAL_OBJ(&exception_arg, exception);
+    }
     zval *this = ddtrace_this(execute_data);
 
     if (zend_fcall_info_init(callable, 0, &fci, &fcc, NULL, NULL) == FAILURE) {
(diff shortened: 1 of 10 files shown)


src/ext/dispatch_compat_php7.c
```diff
@@ -256,7 +262,10 @@ void ddtrace_execute_tracing_closure(zval *callable, zval *span_data, zend_execu
     // Arg 2: mixed $retval
     ZVAL_COPY(&args[2], user_retval);
 
-    fci.param_count = 3;
+    // Arg 3: Exception|null $exception
+    ZVAL_COPY(&args[3], &exception_arg);
```
pawelchcki: ❤️

tests/ext/sandbox/exceptions_are_passed_to_the_tracing_closure.phpt
```diff
@@ -0,0 +1,56 @@
...
+}
+
+dd_trace_function('testExceptionIsNull', function (SpanData $span, array $args, $retval, $ex) {
+    $span->name = 'TestNull';
+    var_dump($ex === null);
+});
+
+dd_trace_function('testExceptionIsPassed', function (SpanData $span, array $args, $retval, $ex) {
+    $span->name = 'TestEx';
+    var_dump($ex instanceof Exception);
```
pawelchcki: ❤️

tests/ext/sandbox/exceptions_are_passed_to_the_tracing_closure.phpt
```diff
@@ -0,0 +1,56 @@
...
+});
```
pawelchcki: Would it make sense to add test that throws an exception inside dd_trace_function. At the same time when exception is passed to the hook function, then asserting that the outer exception is the only one that gets "uncaught"?

Also I wonder about scenario like this:

```
class SubException extends Exception { }

function a(){
  throw new SubException('Oops!');
}

dd_trace_function('a', function( $s, $args, $r, $ex) {
    throw new Exception("!");
});

try {
  a();
catch (SubException $ex) {
   echo "Caught SubException" ;
}
```
With no uncaught exception reported.

Also now just thinking what would happen when someone evil or by mistake does something like this:
```
dd_trace_function('a', function( $s, $args, $r, $ex) {
    throw $ex;
});
```
Its evil, but would be nice to catch this error in other ways than crash and burn :)

PR #552, opened 2019-09-02
Container tagging

labbati:
### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Util/ContainerInfo.php
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @return string|null
+     */
+    public function getContainerId()
+    {
+        if (!file_exists($this->cgroupProcFile)) {
+            return null;
+        }
+
+        $lines = file($this->cgroupProcFile);
```
pawelchcki: what will happen if the file is not user readable ?

pawelchcki: Also What will happen if file is 100MBs long ?

labbati: Very good points, both:

> what will happen if the file is not user readable ?

Let me add a try-catch block.

> Also What will happen if file is 100MBs long ?

Not sure this is very realistic for this file specifically, but still we just a little bit of complexity we can sleep much better :) Are you cool with using [`fgets`](https://www.php.net/manual/en/function.fgets.php)? We now break if one single line is 100MB instead of the while file.

Again we are being extra-defensive here but I agree with that. On the other hand, probably going beyond this (unless you have some other idea) might be possibly overkill.

src/DDTrace/Util/ContainerInfo.php
```diff
@@ -0,0 +1,59 @@
...
+
+        $file = null;
+        try {
+            $file = fopen($this->cgroupProcFile, 'r');
+            while(!feof($file))  {
+                $line = fgets($file);
+                $matches = array();
+                preg_match(self::LINE_RE, trim($line), $matches);
+                if (count($matches) > 3) {
+                    return $matches[3];
```
pawelchcki: :+1 I Like the early return

src/DDTrace/Util/ContainerInfo.php
pawelchcki: `fgets` looks good!
Optionally we could add a fgets read limit. But I think current implementation already might be enough since if that file is corrupted then it means the whole system is completely broken. 

One thing missing I think - would be to add some test to check what happens if we access a read protected file - if possible

labbati: > One thing missing I think - would be to add some test to check what happens if we access a read protected file - if possible

Very good point, it might take a while to figure it out how btw :)

tests/Unit/Util/ContainerInfoTest.php
```diff
@@ -0,0 +1,218 @@
...
+        $this->createCGroupProcFileWithLines([
+            // phpcs:disable
+            '         ',
+            '',
+            // phpcs:enable
+        ]);
+        $this->assertNull($this->containerInfo->getContainerId());
+    }
+
+    public function testFileNoReadPermissions()
```
labbati: @pawelchcki added test for no read permisisons

src/DDTrace/Util/ContainerInfo.php
labbati: @pawelchcki added test `testFileNoReadPermissions`

labbati:
Merging this as comments from @pawelchcki about the missing test when file is not readable was added.

PR #562, opened 2019-09-09
Migrate elasticsearch integration to sandboxed API

labbati:
### Description

Migrate ElasticSearch integration to new sandboxed api

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchSandboxedIntegration.php
```diff
@@ -0,0 +1,227 @@
...
+
+                $span->name = "Elasticsearch.Client.$name";
+
+                if ($isTraceAnalyticsCandidate) {
+                    $integration->addTraceAnalyticsIfEnabled($span);
+                }
+
+                $span->service = ElasticSearchSandboxedIntegration::NAME;
+                $span->type = Type::ELASTICSEARCH;
+                $span->resource = ElasticSearchCommon::buildResourceName($name, $params);
```
SammyK: Bikeshed warning!  ⚠️ 😆 I think this might be a bit more readable than that `$params` temp var above:

```suggestion
                $span->resource = ElasticSearchCommon::buildResourceName($name, isset($args[0]) ? $args[0] : []);
```

I can see this is in the original integration so no worries if you want to keep it as-is. But if you wanna change it, it also exists in `traceNamespaceMethod()` below. :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchSandboxedIntegration.php
```diff
@@ -0,0 +1,227 @@
...
+        );
+    }
+
+    /**
+     * @param string $class
+     * @param string $name
+     */
+    public function traceSimpleMethod($class, $name)
+    {
+        dd_trace_method($class, $name, function (SpanData $span, $args) use ($class, $name) {
```
SammyK: `$args` can be dropped here too :)

tests/Integrations/Elasticsearch/V1/ElasticSearchSandboxedIntegrationTest.php
```diff
@@ -0,0 +1,562 @@
...
+use DDTrace\Tests\Common\IntegrationTestCase;
+use DDTrace\Tests\Common\SpanAssertion;
+use Elasticsearch\Client;
+
+/**
+ * Tests for Elasticsearch Client. We test specifically only most commonly used tests, for the other tests we just make
+ * sure that if a non existing method is provided, that for example does not exists for the used client version
+ * the integration does not throw an exception.
+ */
+class ElasticSearchSandboxedIntegrationTest extends IntegrationTestCase
```
SammyK: I know this is a preview PR, so I'm just assuming you're going to eventually extend this from `ElasticSearchIntegrationTest`, right? :)

src/DDTrace/Integrations/ElasticSearch/V1/ElasticSearchSandboxedIntegration.php
labbati: :palm_face: 😄 nice suggestion

tests/Integrations/Elasticsearch/V1/ElasticSearchSandboxedIntegrationTest.php
labbati: Actually here the rule `IS_SANDBOX=true|false` does not apply as for most of the tests are parent child so the order is the same for sandboxed and not sandboxed.

PR #563, opened 2019-09-09
[Sandbox] Limited tracer

SammyK:
### Description

This PR adds limited tracer functionality to the sandbox API.

### Readiness checklist
- ~~[ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.~~

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
If the tracer is in limited mode, is there anything this callback can do that is worthwhile? Basically, should we keep this in userland or should we run this check in the extension before calling the closure?

SammyK:
> If the tracer is in limited mode, is there anything this callback can do that is worthwhile?

Nope. :)

> should we keep this in userland or should we run this check in the extension before calling the closure?

Ideally we'd do this check at the C-level before even running the tracing closure. The main issue (that we should totally solve when we have more time dedicated to it) is marking certain traces as "limited" when registered. For example, adding another argument to `dd_trace_function()` and `dd_trace_method()`:

```php
dd_trace_method('Foo', 'bar', function () {
    /* */
}, DD_TRACE_IS_LIMITED);
```

But doing this "locks in" our API before we can figure out how we want to pass in other often-static info like "type" and "service". So this PR just offers a "step in the right direction" without making any major API changes that might block future scope. And we can certainly improve it as we land on the best course of action for a static-info API.

Great question - I'm totally tracking with you! :)

src/DDTrace/Tracer.php
```diff
@@ -333,7 +315,7 @@ public function flush()
         if (self::isLogDebugActive()) {
             self::logDebug('Flushing {count} traces, {spanCount} spans', [
                 'count' => $this->getTracesCount(),
-                'spanCount' => $this->getSpanCount(),
+                'spanCount' => dd_trace_closed_spans_count(),
```
morrisonlevi: Does this only count the internally closed spans?

src/ext/memory_limit.c
```diff
@@ -51,3 +51,16 @@ int64_t ddtrace_get_memory_limit(TSRMLS_D) {
 
     return limit;
 }
+
+BOOL_T ddtrace_check_memory_under_limit(TSRMLS_D) {
+    static int64_t limit = -1;
+    static zend_bool fetched_limit = 0;
+    if (!fetched_limit) {  // cache get_memory_limit() result to make this function blazing fast
```
morrisonlevi: This isn't threadsafe, correct?

src/DDTrace/Tracer.php
SammyK: It counts both internal and userland spans.

src/ext/memory_limit.c
SammyK: Not explicitly. This is code that was [moved from **ddtrace.c**](https://github.com/DataDog/dd-trace-php/blob/master/src/ext/ddtrace.c#L480-L496).

PR #570, opened 2019-09-12
WordPress 4.x integration

SammyK:
### Description

This PR add support for WordPress 4.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


tests/Common/SpanAssertion.php
```diff
@@ -97,6 +97,9 @@ public static function forOperation($name, $error = false, $onlyCheckExistence =
     public function setError($errorType = null, $errorMessage = null, $exceptionThrown = false)
     {
         $this->hasError = true;
+        if (!is_array($this->exactTags)) {
+            $this->exactTags = [];
+        }
```
morrisonlevi: Can you show me an example of why this is needed?

tests/Frameworks/WordPress/Version_4_8/wp-content/plugins/datadog/datadog.php
```diff
@@ -0,0 +1,26 @@
...
+ * Text Domain: datadog-sample
+*/
+
+function datadog_parse_request($wp) {
+	if (!isset($wp->query_vars['name'])) {
+		return;
+	}
+	if ('simple' === $wp->query_vars['name']) {
+		echo "Simple text endpoint\n";
+        exit;
```
morrisonlevi: Is that the idiomatic WordPress way to do this simple endpoint?

tests/Frameworks/WordPress/Version_4_8/wp-content/themes/twentyseventeen/functions.php
```diff
@@ -565,3 +565,5 @@ function twentyseventeen_front_page_template( $template ) {
  * SVG icons functions and filters.
  */
 require get_parent_theme_file_path( '/inc/icon-functions.php' );
+
+remove_filter('template_redirect', 'redirect_canonical');
```
morrisonlevi: What's the purpose of this one?

tests/Common/SpanAssertion.php
SammyK: The [default for `$this->exactTags`](https://github.com/DataDog/dd-trace-php/blob/8093450efd81fc32f76e109536c226f6ac0c621a/tests/Common/SpanAssertion.php#L15) is a string. So if you don't use `withExactTags()` before calling `setError()`, then `$this->exactTags` will be a string and will raise a warning in the tests. This is needed since the WordPress tests check for errors but not exact tags on the error span.

tests/Frameworks/WordPress/Version_4_8/wp-content/themes/twentyseventeen/functions.php
SammyK: By default WordPress redirects URL's that don't end with a slash to the slashed version. Our web tests test for endpoints sans trailing slashes `/simple`,  `/simple_view`, and  `/error` which were all returning 301 redirects without this.

tests/Frameworks/WordPress/Version_4_8/wp-content/plugins/datadog/datadog.php
SammyK: It looks like there are a number of options for adding custom endpoints in WP. A quick Google search unearthed [this resource](https://torquemag.io/2017/07/four-strategies-adding-custom-endpoints-wordpress-site/). It appears we're doing it the "old school" way here. :)

The custom endpoints for [the REST API](https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/) would be nice, but the endpoints are all prefixed with `/wp-json`, so it wouldn't work for our tests without an additional workaround. But I'm totally open to some other method of adding custom endpoints to test if you think that's a good route to take. :)

Also the `/simple` test endpoint is an actual custom WP page that is in the DB dump for the tests.

morrisonlevi: Okay. We should at least fix the formatting issue here ;)

SammyK: Lol - nice catch. Fixed. I went ahead and added a comment above the `remove_filter()` line you pointed out as well. :)

NickStallman:
Is there a particular reason why this Integration is only enabled for Wordpress 4 when 5 has been out for about a year?
Wordpress 5 should be virtually identical and still work perfectly as-is (unless additional hooks are wanted for Gutenberg). It shouldn't need a separate integration to support it.

I'll be doing some testing with Wordpress 5 this week.

SammyK:
@NickStallman Thanks for the feedback! We targeted WordPress 4 first due to demand. WordPress 5 is on our radar as well as improved WordPress support overall. :)

PR #571, opened 2019-09-13
Correctly treat class names as case insensitive

morrisonlevi:
### Description

In PHP, class and method names are case insensitive. We already handled method names correctly. This PR adds support for case insensitive class names too.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/dispatch.c
+++ b/src/ext/dispatch.c
@@ -43,14 +43,16 @@ static ddtrace_dispatch_t *find_method_dispatch(const zend_class_entry *class, z
     }
     HashTable *class_lookup = NULL;
 
+    // downcase the class name as they are case insensitive
 #if PHP_VERSION_ID < 70000
-    const char *class_name = NULL;
-    size_t class_name_length = 0;
-    class_name = class->name;
-    class_name_length = class->name_length;
+    size_t class_name_length = class->name_length;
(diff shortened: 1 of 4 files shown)


src/ext/dispatch.c
```diff
@@ -43,14 +43,16 @@ static ddtrace_dispatch_t *find_method_dispatch(const zend_class_entry *class, z
...
     HashTable *class_lookup = NULL;
 
+    // downcase the class name as they are case insensitive
 #if PHP_VERSION_ID < 70000
-    const char *class_name = NULL;
-    size_t class_name_length = 0;
-    class_name = class->name;
-    class_name_length = class->name_length;
+    size_t class_name_length = class->name_length;
+    char *class_name = zend_str_tolower_dup(class->name, class_name_length);
```
labbati: Should we measure the impact of lowering every call? What do you think? Will it be noticeable?

morrisonlevi: It's just a guess, but I think it will be okay. I'm not sure we can really avoid it -- the case insensitivity is PHP's behavior, which I'm sure we'll need to obey. In cases like CodeIgniter they definitely use the insensitivity as a feature, which is how I discovered this bug.

PR #575, opened 2019-09-15
Add filesystem integration

RaoulMeyer:
### Description

This adds an integration for the native `file_get_contents` and `file_put_contents` functions. In a follow up, I want to also add spans to `fopen`/`fread`/`fwrite`/`fclose`. 

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Type.php
+++ b/src/DDTrace/Type.php
@@ -4,6 +4,7 @@ namespace DDTrace;
 
 class Type
 {
+    const FILESYSTEM = 'filesystem';
     const CACHE = 'cache';
     const HTTP_CLIENT = 'http';
     const WEB_SERVLET = 'web';
--- a/bridge/dd_require_all.php
+++ b/bridge/dd_require_all.php
@@ -72,6 +72,7 @@ require __DIR__ . '/../src/DDTrace/Integrations/Eloquent/EloquentSandboxedIntegr
 require __DIR__ . '/../src/DDTrace/Integrations/Memcached/MemcachedIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/Memcached/MemcachedSandboxedIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/Curl/CurlIntegration.php';
+require __DIR__ . '/../src/DDTrace/Integrations/Filesystem/FilesystemIntegration.php';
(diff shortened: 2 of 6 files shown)


RaoulMeyer:
Looks like a lot of framework tests now fail, because they do filesystem things and assert on an exact list of spans. What is the desired way to fix this? Seems to me like integration tests should not be affected by changes in other integrations, but that might be intended.

SammyK:
Thanks for the integration PR @RaoulMeyer! I'm really happy you tackled some of the core IO functions. We'll get this reviewed this week and get back with you on any feedback. :)

SammyK:
Thanks again for the fantastic integration PR, @RaoulMeyer!

There are two few points I'd like to get @labbati and @morrisonlevi's feedback on:

0. For PHP standard lib integrations - should we prefix the integration names with `PhpStd` or something similar?
1. @RaoulMeyer had an excellent point about this affecting the other integrations. Do we want to have this integration disabled by default or perhaps just explicitly disable it for the integration tests that have file I/O operations? Or some other idea?

@RaoulMeyer: We'll need a little bit of extra time to sort the implementation details. Until then, since we've been doing a lot of merging to master recently, there's a conflict on this PR now. Would you mind resolving that? :)

RaoulMeyer:
@SammyK Thanks! I rebased the PR, as soon as you've decided on the details I'll this PR update accordingly.

morrisonlevi:
Once we merge in https://github.com/DataDog/dd-trace-php/pull/585 then we can "assert a hierarchical graph of spans with parent child relations", which may be the way we can make things like this more robust in the future. I haven't thoroughly reviewed the PR to see if how it handles extra spans, though, so maybe that won't work out.

src/DDTrace/Type.php
```diff
@@ -4,6 +4,7 @@
 
 class Type
 {
+    const FILESYSTEM = 'filesystem';
```
morrisonlevi: Given that these functions can do more than filesystems, I think we should not name it 'filesystem'. I've been thinking more about it; these are really stream operations. Maybe we call it 'phpstream' or something?

krakjoe:
While enthusiasm was initially high, this no longer fits into our road map and the conversation went stale long ago.

We're grateful for the time you spent here, and we apologize for the long delay in feedback, but I'm going to close this PR as we won't be moving forward in this direction.

Thanks again.

PR #576, opened 2019-09-16
Fix static keyword handling

SammyK:
### Description

This PR fixes an edge case the breaks usage of `new static()` for traced methods in both the sandbox and original API's.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/dispatch_compat_php5.c
+++ b/src/ext/dispatch_compat_php5.c
@@ -301,7 +301,11 @@ int ddtrace_forward_call(zend_execute_data *execute_data, zend_function *fbc, zv
     fcc.function_handler = fbc;
     fcc.object_ptr = ddtrace_this(execute_data);
     fcc.calling_scope = fbc->common.scope;  // EG(scope);
-    fcc.called_scope = fcc.object_ptr ? Z_OBJCE_P(fcc.object_ptr) : fbc->common.scope;
+#if PHP_VERSION_ID < 50500
+    fcc.called_scope = EX(called_scope);
+#else
+    fcc.called_scope = EX(call) ? EX(call)->called_scope : NULL;
+#endif
 
     ddtrace_setup_fcall(execute_data, &fci, &retval_ptr TSRMLS_CC);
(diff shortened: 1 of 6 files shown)


src/ext/dispatch_compat_php5.c
```diff
@@ -209,7 +209,13 @@ void ddtrace_wrapper_forward_call_from_userland(zend_execute_data *execute_data,
     fcc.function_handler = DDTRACE_G(original_context).fbc;
     fcc.object_ptr = DDTRACE_G(original_context).this;
     fcc.calling_scope = DDTRACE_G(original_context).calling_ce;
-    fcc.called_scope = fcc.object_ptr ? Z_OBJCE_P(fcc.object_ptr) : DDTRACE_G(original_context).fbc->common.scope;
+#if PHP_VERSION_ID < 50500
```
morrisonlevi: I assume you got inspiration from php-src somewhere? If so, we should leave references.

SammyK: Nope - just many months of stepping through the VM and knowing where this stuff exists in the call frame. 😆 Is there something in particular you think should be referenced?

morrisonlevi: Not that I know of. I did a search using OpenGrok and found [this snippet in `zend_call_method`](https://lxr.room11.org/xref/php-src%405.6/Zend/zend_interfaces.c#86-95); maybe it should look more like that on PHP 5? I'm slightly concerned we haven't added enough variety to our tests.

SammyK: I'm absolutely +1 with you on not having enough core test coverage on traced functions and methods.

I think one possible solution to that would be running the core php-src phpt tests for functions and methods but with tracing on all of the calls. I've been pondering a way we could use reflection or something to inject `dd_trace_*()` functions at the top of every `--FILE--` section for all the called functions & methods in the test and making sure the test still passes.

I don't think this is a blocker for this PR, but something we'll certainly address very soon! :)

PR #585, opened 2019-09-20
Add possibility to assert a hierarchical graph of spans with parent child relations  

labbati:
### Description

Add the possibility to test a flame graph (with parent child relations) instead of only flat list of spans.

In order to demonstrate how it works, the duplicate elastic search integration test suite has been migrated to the new `assertFlameGraph()` function removing code duplication.

### Readiness checklist
- ~[ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.~

### Reviewer checklist
- [ ] Milestone is set.


tests/Integrations/Elasticsearch/V1/ElasticSearchSandboxedIntegrationTest.php
```diff
@@ -2,270 +2,16 @@
...
-                    // No results, scroll cursor is empty.  You've exported all the data
-                    break;
-                }
-            }
-        });
-
-        $this->assertEmpty($traces);
-    }
 
     public function testScroll()
```
labbati: An edge case (multiple spans same operation and resource name at the same level) not handled yet. For now we just replicate that single one test and in the future we will improve.

tests/Common/SpanChecker.php
```diff
@@ -2,13 +2,122 @@
 
 namespace DDTrace\Tests\Common;
 
+use DDTrace\Contracts\Span;
```
SammyK: This doesn't seem to be referenced anywhere. :)

tests/Common/SpanChecker.php
```diff
@@ -2,13 +2,122 @@
...
+    {
+        $flattenTraces = $this->flattenTraces($traces);
+        $actualGraph = $this->buildSpansGraph($flattenTraces);
+        foreach ($expectedFlameGraph as $oneTrace) {
+            $this->assertNode($actualGraph, $oneTrace);
+        }
+    }
+
+    /**
+     * @param [] $spansByFingerPrint
```
SammyK: Just a little nitpick :D 

```suggestion
     * @param array $graph
```

tests/Common/SpanChecker.php
```diff
@@ -2,13 +2,122 @@
...
+     * @return array
+     */
+    private function buildSpansGraph(array $flatSpans)
+    {
+        $byId = [];
+        foreach ($flatSpans as $span) {
+            $byId[$span['span_id']] = ['span' => $span, 'children' => []];
+        }
+
+        while (true) {
```
SammyK: Could we change this to a do/while?

```php
do {
// ...
} while (count($byId) !== $lastCount);
```

PR #606, opened 2019-10-07
Fix CodeIgniter v2 integration init hook

morrisonlevi:
### Description

When working with a real agent we have an issue with dd_trace_method
being used to set up the global tracer and company. Using the old
dd_trace API avoids this issue since it doesn't attempt to make any
spans.

At present, I'm not sure how to test it through CI. Will sleep on it.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/bridge/functions.php
+++ b/bridge/functions.php
@@ -91,10 +91,17 @@ function dd_wrap_autoloader()
 
     /* CodeIgniter v2 does not use an autoloader. Tracing the CI_Hooks
      * constructor let's us set up the world because it is called very early
-     * in CodeIgniter's startup process before we need to trace anything. */
-    dd_trace_method('CI_Hooks', '__construct', function () {
+     * in CodeIgniter's startup process before we need to trace anything.
+     *
+     * Note that this hook cannot use dd_trace_method, since dd_trace_method
+     * will attempt to create a span before everything we need to make spans
+     * has been set up, it puts us in a bad state.
+     */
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Good thinking, Luca. I haven't tested it, but the thinking was that it should be benign.

bridge/functions.php
```diff
@@ -92,9 +92,9 @@ function dd_wrap_autoloader()
     /* CodeIgniter v2 does not use an autoloader. Tracing the CI_Hooks
      * constructor let's us set up the world because it is called very early
      * in CodeIgniter's startup process before we need to trace anything. */
-    dd_trace_method('CI_Hooks', '__construct', function () {
+    dd_trace('CI_Hooks', '__construct', function () {
         require __DIR__ . '/dd_init.php';
```
SammyK: To address @labbati excellent point, can we add the following line after this `require` to disable the autoloader hook?

```php
dd_untrace('spl_autoload_register');
```

Even though the bootstrapper will only load the tracer once after requiring **dd_init.php** again, this will at least remove any unnecessary extra bootsrapping. :)

morrisonlevi:
I verified manually that using CodeIgniter with composer is fine. This uncovered a minor issue with ZendFramework's integration loader not reporting the right status: https://github.com/DataDog/dd-trace-php/pull/609.

PR #608, opened 2019-10-09
[POC] Drop spans without orphans

SammyK:
### Description

**DO NOT MERGE** :)

This is just a POC showing how to save child spans from becoming orphans when dropping a parent span from the sandbox API. This implementation would slow things down especially in limited-tracing mode. There are a number of ways to optimize this, but as @morrisonlevi pointed out, avoiding span dropping in the first place is a cleaner approach.

The two main areas where the tracer drops spans in the sandbox API are:

0. Call hooks that load integrations but do not need to create spans.
1. Dropping the span when the tracer is in limited tracing mode which has the potential to orphan a huge number of valid spans.

I'd love to have a discussion for figuring out an API for each of these requirements. :)

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -140,6 +140,7 @@
                         <file name="dd_trace_method_works_with_dd_trace.phpt" role="test" />
                         <file name="dd_trace_push_span_id.phpt" role="test" />
                         <file name="drop_spans.phpt" role="test" />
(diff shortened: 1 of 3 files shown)


SammyK:
A fix for this in on the radar so I'm closing this POC PR. :)

PR #615, opened 2019-10-15
Support x-datadog-origin header (Synthetics support)

SammyK:
### Description

This PR will check for a distributed trace origin via the `x-datadog-origin` HTTP header. If it exists, the top-level span metadata is appended with `_dd.origin` which is set to the value of the `x-datadog-origin` HTTP header. The value is also propagated to other traces via the same HTTP header.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Propagators/TextMap.php
```diff
@@ -67,6 +70,7 @@ public function extract($carrier)
 
         $spanContext = new SpanContext($traceId, $spanId, null, $baggageItems, true);
         $this->extractPrioritySampling($spanContext, $carrier);
+        $this->extractOrigin($spanContext, $carrier);
```
brettlangdon: extracting the origin is only valid if we were able to successfully extract a trace id.

SammyK: Thanks for clarifying! This line only gets run in this case as [on line 67-69](https://github.com/DataDog/dd-trace-php/pull/615/files#diff-bca28826d2bcd463c28ea6562b776fccR67-R69) there is a null-check on the trace ID. :)

PR #617, opened 2019-10-15
Close internally opened spans when DD_AUTOFINISH_SPANS is on

morrisonlevi:
### Description

With this PR, internally created spans (e.g. ones from sandboxing APIs such as `dd_trace_method`) now respect the environment variable DD_AUTOFINISH_SPANS.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -57,6 +57,7 @@ ZEND_ARG_INFO(0, function_name)
 ZEND_ARG_INFO(0, tracing_closure)
 ZEND_END_ARG_INFO()
 
+// also used by ddtrace_close_open_spans
 ZEND_BEGIN_ARG_INFO_EX(arginfo_dd_trace_serialize_closed_spans, 0, 0, 0)
 ZEND_END_ARG_INFO()
 
@@ -339,6 +340,12 @@ static PHP_FUNCTION(dd_trace_serialize_closed_spans) {
     PHP5_UNUSED(return_value_used, this_ptr, return_value_ptr, ht);
     PHP7_UNUSED(execute_data);
     ddtrace_serialize_closed_spans(return_value TSRMLS_CC);
+}
+
+static PHP_FUNCTION(ddtrace_close_open_spans) {
+    PHP5_UNUSED(return_value_used, this_ptr, return_value_ptr, ht);
+    PHP7_UNUSED(execute_data);
+    ddtrace_close_open_spans(TSRMLS_C);
 }
 
 // Invoke the function/method from the original context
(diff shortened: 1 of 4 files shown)


src/ext/ddtrace.c
```diff
@@ -175,6 +175,11 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
         return SUCCESS;
     }
 
+    // todo: default DD_AUTOFINISH_SPANS to on for both userland and internally made spans
+    if (ddtrace_get_bool_config("DD_AUTOFINISH_SPANS", 0 TSRMLS_CC)) {
+        ddtrace_close_open_spans(TSRMLS_C);
```
SammyK: So far so great! :)

We might need to expose this function and env var check to userland since RSHUTDOWN is too late to close the spans. The tracer flushes in a `register_shutdown_function()` closure which runs [first in the shutdown process](https://github.com/php/php-src/blob/db233501ff9d56765ef4a870b777a643c2136711/main/main.c#L1832-L1835) so when [RSHUTDOWN is called](https://github.com/php/php-src/blob/db233501ff9d56765ef4a870b777a643c2136711/main/main.c#L1864-L1867), the tracer has already flushed.

src/ext/span.c
```diff
@@ -131,3 +131,14 @@ void ddtrace_serialize_closed_spans(zval *serialized TSRMLS_DC) {
...
 }
+
+void ddtrace_close_open_spans(TSRMLS_D) {
+    ddtrace_span_t *span = DDTRACE_G(open_spans_top);
+    while (span != NULL) {
+        ddtrace_close_span(TSRMLS_C);
+        span = DDTRACE_G(open_spans_top);
+    }
+
+    // todo: isn't open_spans_count always 0?
```
SammyK: The `DDTRACE_G(open_spans_count)` state is handled by `ddtrace_pop_span_id()` which is called from `ddtrace_close_span()` so we don't need to manually reset it to `0` here. :)

src/ext/ddtrace.c
morrisonlevi: Yeah, I discovered this a bit later. I wonder if we ought to do it the other way, though: move our `register_shutdown_function` stuff into `RSHUTDOWN`? This would be more resilient in the face of someone using `exit` in a shutdown function, for instance. What do you think?

SammyK: I'm totally for this as we start moving things to the C level for sure. Right now about 90% of the flushing logic is in userland so we might need to keep it there for a while. Also - `register_shutdown_function()` closures are still called on `exit` so we wouldn't be necessarily gaining an edge there. But overall, I'm totally on board for eventually adding flush-on-RSHUTDOWN logic. :)

morrisonlevi: Ah, I wasn't clear. In the event a shutdown handler calls exit then the remainder won't get called. It's a small edge, definitely. I was thinking about literally calling the PHP code we call in the shudown handler today from RSHUTDOWN. Basically, instead of an anonymous function we give it a name and then just call that from RSHUTDOWN. Thoughts?

SammyK: Ah! You are the edge-case master. Your original thought was clear - I just wasn't clear reading it. :)

I like the idea of flushing in userland on RSHUTDOWN. I've never tried running userland in RSHUTDOWN before and it's after the output buffer dies so I'm wondering if we might have to deal with some edge cases there. But I really love the idea! :) Are you thinking of exposing a userland function similar to `register_shutdown_function()` to register a closure that is called in RSHUTDOWN or something?

morrisonlevi: Today we have this in `DDTrace\Bootstrap::tracerOnce`:

```php
    register_shutdown_function(function () {
        dd_trace_disable_in_request(); //disable function tracing to speedup shutdown

        $scopeManager = $tracer->getScopeManager();
        $scopeManager->close();
        $tracer->flush();
    });
```

What if we took that anonymous function and put it in a static method, like `DDTrace\Bootstrap::shutdown_handler` or something, and then in RSHUTDOWN we did the internals equivalents of `class_exists('DDTrace\Bootstrap', $autoload = false)` and `method_exists('DDTrace\Bootstrap', 'shutdown_handler')` and if it is there then we call it?

SammyK: Even better! :)

SammyK: Also - the output buffer dies after RSHUTDOWN, not before as I stated earlier. So that's good! (I was looking at the post-RSHUTDOWN hook.)

morrisonlevi: Current PR status: `Bootstrap::shutdown` is called from RSHUTDOWN on PHP 7 and on PHP 5 there is still `register_shutdown_handler` until I get a few more minutes to do the internals for it.

morrisonlevi:
Closing. This approach does not run the closure callback, and since required properties will be missing these spans will get dropped.

PR #621, opened 2019-10-17
Health metrics: track segmentation faults

morrisonlevi:
### Description

This installs an alternate stack and sigaction for SIGSEGV, which should trigger when there is a segmentation fault.

Depends on Add minimal Dogstatsd client #628.

### Readiness checklist
- [x] Report the segfault to health metrics.

### Reviewer checklist
- [x] Milestone is set.


src/ext/ddtrace.c
```diff
@@ -92,6 +99,39 @@ static void register_span_data_ce(TSRMLS_D) {
...
+#define METRICS_CONST_TAGS "lang:php,lang_version:" PHP_VERSION ",tracer_version:" PHP_DDTRACE_VERSION
+
+static void ddtrace_sigsegv_handler(int sig) {
+    TSRMLS_FETCH();
+    char *dogstatsd_host = get_dd_agent_host();
+    char *dogstatsd_port = get_dd_dogstatsd_port();
+
+    ddtrace_log_errf("Segmentation fault");
+
+    // todo: extract buffer, host, port, and client to module global, so this is just a send
```
SammyK: Is this something we can do for this PR or should we wait and tackle later?

src/ext/ddtrace.c
```diff
@@ -102,10 +142,27 @@ static PHP_MINIT_FUNCTION(ddtrace) {
...
+     * stack overflows.
+     */
+    if (health_metrics_enabled && (ddtrace_altstack.ss_sp = malloc(SIGSTKSZ))) {
+        ddtrace_altstack.ss_size = SIGSTKSZ;
+        ddtrace_altstack.ss_flags = 0;
+        if (sigaltstack(&ddtrace_altstack, NULL) == 0) {
+            ddtrace_sigaction.sa_flags = SA_ONSTACK;
+            ddtrace_sigaction.sa_handler = ddtrace_sigsegv_handler;
+            sigemptyset(&ddtrace_sigaction.sa_mask);
+            sigaction(SIGSEGV, &ddtrace_sigaction, NULL);
```
SammyK: I'm wondering if we should split all this into a separate file and just expose functions for MINIT and MSHUTDOWN. Since [backtrace.c already ties into a signal handler](https://github.com/DataDog/dd-trace-php/blob/master/src/ext/backtrace.c), it might make sense to add this there, or create a separate place for it. It would be nice to have a single place for signal handlers in general so that when we fuzz the extension, it would be simple to disable all signal handlers from one place. Any ideas on how we could organize something like that? :)

tests/Metrics/SigSegV/SigSegVTest.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $log = __DIR__ . '/../../Frameworks/Custom/Version_Not_Autoloaded/error.log';
+        self::assertFileNotExists($log);
+
+        $spec = GetSpec::create('sigsegv', '/');
+        $this->call($spec);
+
+        self::assertFileExists($log);
+        $contents = \file_get_contents($log);
+        self::assertRegExp("/.*Segmentation fault\n.*sigsegv health metric sent/", $contents);
```
SammyK: ❤️

src/dogstatsd/README.md
```diff
@@ -0,0 +1,85 @@
+# A dogstatsd client written in C
```
SammyK: Your docs and comments are the 💣.com

config.m4
```diff
@@ -10,7 +10,34 @@ if test "$PHP_DDTRACE" != "no"; then
     PHP_SUBST(EXTRA_CFLAGS)
   fi
 
-  PHP_NEW_EXTENSION(ddtrace, src/ext/ddtrace.c src/ext/memory_limit.c src/ext/configuration.c src/ext/configuration_php_iface.c src/ext/circuit_breaker.c src/ext/dispatch_setup.c src/ext/dispatch.c src/ext/third-party/mt19937-64.c src/ext/random.c src/ext/coms.c src/ext/coms_curl.c src/ext/coms_debug.c src/ext/request_hooks.c src/ext/compat_string.c src/ext/dispatch_compat_php5.c src/ext/dispatch_compat_php7.c src/ext/backtrace.c src/ext/logging.c src/ext/env_config.c src/ext/serializer.c src/ext/span.c src/ext/trace.c src/ext/mpack/mpack.c, $ext_shared,, -DZEND_ENABLE_STATIC_TSRMLS_CACHE=1 -Wall -std=gnu11)
+  dnl ddtrace.c comes first, then everything else alphabetically
```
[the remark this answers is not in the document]
morrisonlevi: It's just what I've seen most projects do; the "main" file comes first.

src/ext/configuration.h
```diff
@@ -12,10 +12,12 @@ void ddtrace_config_shutdown(void);
 
 #define DD_CONFIGURATION                                                                                  \
     CHAR(get_dd_agent_host, "DD_AGENT_HOST", "localhost")                                                 \
+    CHAR(get_dd_dogstatsd_port, "DD_DOGSTATSD_PORT", "8125")                                              \
     INT(get_dd_trace_agent_port, "DD_TRACE_AGENT_PORT", 8126)                                             \
     BOOL(get_dd_trace_debug, "DD_TRACE_DEBUG", FALSE)                                                     \
     BOOL(get_dd_trace_agent_debug_verbose_curl, "DD_TRACE_AGENT_DEBUG_VERBOSE_CURL", FALSE)               \
     BOOL(get_dd_trace_debug_curl_output, "DD_TRACE_DEBUG_CURL_OUTPUT", FALSE)                             \
+    BOOL(get_dd_trace_heath_metrics_enabled, "DD_TRACE_HEALTH_METRICS_ENABLED", FALSE)                    \
```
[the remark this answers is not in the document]
morrisonlevi: No, it's is supposed to be disabled by default for now.

src/ext/ddtrace.c
morrisonlevi: We can do this now, or later. I just didn't want to stall an important feature on this extraction when at the moment it will have no effect on code whatsoever until we have other metrics.

src/ext/ddtrace.c
morrisonlevi: I've don't think seen anything that sets `DD_LOG_BACKTRACE` before, or at least didn't notice it. We should refactor this before merge. For both features to work we'll have to.

morrisonlevi: I don't think we have tests for the existing backtrace printer, which is why my tests pass without issues.

PR #624, opened 2019-10-23
Add Yii 2 integration

morrisonlevi:
### Description

Adds support for Module and Controller runAction, View::renderFile, and adds some metadata to the root span.

Tests are for PHP 5.4, 5.6, and 7.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -205,6 +205,7 @@
             "@cakephp-28-test",
             "@laravel-42-update",
             "@laravel-42-test",
+            "@yii-2-suite",
             "@zend-framework-1-test",
             "@custom-framework-autoloaded-suite"
         ],
@@ -220,6 +221,7 @@
             "@symfony-30-suite",
             "@symfony-33-suite",
             "@symfony-34-suite",
+            "@yii-2-suite",
             "@zend-framework-1-test",
             "@custom-framework-autoloaded-suite"
         ],
--- a/composer.json
+++ b/composer.json
@@ -235,6 +235,7 @@
             "@symfony-33-suite",
             "@symfony-34-suite",
             "@wordpress-48-test",
+            "@yii-2-suite",
             "@zend-framework-1-test",
             "@custom-framework-autoloaded-suite"
         ],
@@ -262,6 +263,7 @@
             "@symfony-40-suite",
             "@symfony-42-suite",
             "@wordpress-48-test",
+            "@yii-2-suite",
             "@zend-framework-1-test",
             "@custom-framework-autoloaded-suite"
         ],
@@ -292,6 +294,7 @@
             "@symfony-40-suite",
             "@symfony-42-suite",
             "@wordpress-48-test",
+            "@yii-2-suite",
             "@zend-framework-1-test",
             "@custom-framework-autoloaded-suite"
         ],
@@ -312,6 +315,7 @@
             "@symfony-40-suite",
             "@symfony-42-suite",
             "@wordpress-48-test",
+            "@yii-2-suite",
             "@zend-framework-1-test",
             "@custom-framework-autoloaded-suite"
         ],
@@ -398,6 +402,10 @@
         "symfony-42-suite": [ "@symfony-42-update", "@symfony-42-test" ],
 
         "wordpress-48-test": "@composer test -- tests/Integrations/WordPress/V4_8",
+
+        "yii-2-update": "@composer --working-dir=tests/Frameworks/Yii/Version_2_0_26 update",
+        "yii-2-test": "@composer test -- tests/Integrations/Yii/V2_0_26",
+        "yii-2-suite": [ "@yii-2-update", "@yii-2-test"],
 
         "zend-framework-1-test": "@composer test -- tests/Integrations/ZendFramework/V1",
 
--- a/src/DDTrace/Integrations/IntegrationsLoader.php
+++ b/src/DDTrace/Integrations/IntegrationsLoader.php
@@ -24,6 +24,7 @@ use DDTrace\Integrations\Slim\SlimIntegration;
 use DDTrace\Integrations\Symfony\SymfonyIntegration;
 use DDTrace\Integrations\Web\WebIntegration;
 use DDTrace\Integrations\WordPress\WordPressSandboxedIntegration;
+use DDTrace\Integrations\Yii\YiiSandboxedIntegration;
 use DDTrace\Integrations\ZendFramework\ZendFrameworkIntegration;
 use DDTrace\Log\LoggingTrait;
 
@@ -91,6 +92,8 @@ class IntegrationsLoader
                 '\DDTrace\Integrations\PDO\PDOSandboxedIntegration';
             $this->integrations[WordPressSandboxedIntegration::NAME] =
                 '\DDTrace\Integrations\WordPress\WordPressSandboxedIntegration';
+            $this->integrations[YiiSandboxedIntegration::NAME] =
+                '\DDTrace\Integrations\Yii\YiiSandboxedIntegration';
         }
     }
 
--- a/src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
+++ b/src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
@@ -5,7 +5,6 @@ namespace DDTrace\Integrations\Yii\V2;
 use DDTrace\Configuration;
 use DDTrace\Contracts\Scope;
 use DDTrace\GlobalTracer;
-use DDTrace\Http\Urls;
 use DDTrace\Integrations\Yii\YiiSandboxedIntegration;
 use DDTrace\Integrations\Integration;
 use DDTrace\SpanData;
@@ -28,21 +27,13 @@ class YiiIntegrationLoader
         $root->setTraceAnalyticsCandidate();
         $service = Configuration::get()->appName(YiiSandboxedIntegration::NAME);
 
-        // This will also attach app.endpoint info to the root span
         \dd_trace_method(
             'yii\web\Application',
             'run',
-            function (SpanData $span) use ($root, $service) {
+            function (SpanData $span) use ($service) {
                 $span->name = $span->resource = \get_class($this) . '.run';
                 $span->type = Type::WEB_SERVLET;
                 $span->service = $service;
-
-                if (isset($this->controller->action->actionMethod)) {
-                    $controller = \get_class($this->controller);
-                    $endpoint = "{$controller}::{$this->controller->action->actionMethod}";
-                    $root->setTag("app.endpoint", $endpoint);
-                    $root->setTag(Tag::HTTP_URL, Url::base(true) . Url::current());
-                }
             }
         );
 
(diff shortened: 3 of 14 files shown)


src/DDTrace/Integrations/Yii/YiiSandboxedIntegration.php
```diff
@@ -0,0 +1,67 @@
...
+        if (!self::shouldLoad(self::NAME)) {
+            return self::NOT_AVAILABLE;
+        }
+
+        $integration = self::getInstance();
+
+        // This happens somewhat early in the setup, though there may be a better candidate
+        \dd_trace_method('yii\di\Container', '__construct', function () use ($integration) {
+            $version = \Yii::getVersion();
+            $majorVersion = \substr($version, 0, 2);
```
labbati: If Yii follows semver, you could use utilities in `\DDTrace\Util\Versions::versionMatches('2', \Yii::getVersion())`

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $scope = GlobalTracer::get()->getRootScope();
+        if (!$scope instanceof Scope) {
+            return Integration::NOT_LOADED;
+        }
+        $root = $scope->getSpan();
+        // Overwrite the default web integration
+        $root->setIntegration($integration);
+        $root->setTraceAnalyticsCandidate();
+        $root->overwriteOperationName('yii.request');
```
labbati: I think that operation name should still be `web.request`, here and for the future integrations. The reason why I am saying this is based on the following scenario.

We do not support a framework, so given the url to resource name will be `name: web.request, resource: GET /some/path`. When we start to support a framework it becomes `name: yii.request, resource: GET /some/path`. I don't see a benefit for this. What about adding a tag `app.framework: yii` to the root span and leave `web.request`?

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,114 @@
...
+        $service = Configuration::get()->appName(YiiSandboxedIntegration::NAME);
```
labbati: This is quite risky actually because we can end up with having an empty service if the user did not set any service name, causing traces to be dropped. I would suggest to go into `Configuration::get()->appName()` and set a default to `'php-web-app'`.

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,114 @@
...
+            'renderFile',
+            function (SpanData $span, $args) use ($service) {
+                $span->name = \get_class($this) . '.renderFile';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $service;
+                $span->resource = isset($args[0]) && \is_string($args[0]) ? $args[0] : $span->name;
+            }
+        );
+
+        /* I'm hoping to be able to get app.route.path and proper roote resource name out of this somehow:
```
labbati: typo `roote`

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,114 @@
...
+            return false;
+        });
+        */
+
+        $root->setTag(Tag::SERVICE_NAME, $service);
+        if ('cli' !== PHP_SAPI) {
+            $normalizer = new Urls(explode(',', getenv('DD_TRACE_RESOURCE_URI_MAPPING')));
+            $root->setTag(
+                Tag::RESOURCE_NAME,
+                $_SERVER['REQUEST_METHOD'] . ' ' . $normalizer->normalize($_SERVER['REQUEST_URI']),
```
labbati: I noticed your commented code above about extracting path info. Do you think it is worth to do it? If not that what you are doing here is done automatically for you in `\DDTrace\Tracer::addUrlAsResourceNameToRootSpan()` at flush time. In general we should do it here in the integration ONLY if we can add more precise info based on the `Route` object. Typically this should be feasible.

src/DDTrace/Integrations/Yii/YiiSandboxedIntegration.php
morrisonlevi: They do not follow semver, but I suspect this will work on all v2 releases anyway.  Think I should switch?

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
morrisonlevi: There's already `integration.name` set on it, with the value of `yii`. I guess we could add `app.framework.name=yii` and `app.framework.version=\Yii::getVersion()` if you think customers will think that's helpful. Otherwise I'm fine with `integration.name`.

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
morrisonlevi: The parameter to `Configuration::get()->appName` is a default in the case `DD_SERVICE_NAME` (or one of the other deprecated env variables) is not set or is set but empty. So worst case the service is `yii`.

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
morrisonlevi: So far I can extract what the variables should be, so in the case of `/homes/new-york/new-york/manhattan` I can extract `['state' => 'new-york', 'city' => 'new-york', 'neighborhood' => 'manhattan']`, but not where in the URL those values should get plugged. This is why I dropped it, but I thought the commented code is worth keeping around for when we try for V1.

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
labbati: > I guess we could add app.framework.name=yii and app.framework.version=\Yii::getVersion() if you think customers will think that's helpful
I did not remember of `integration.name`. So forget about it.

What do you think about `web.request` instead?

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
morrisonlevi: I've added support for parameterized routes, both `:` and `?`, in the latest commit. Ready for re-review.

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,119 @@
...
+            'renderFile',
+            function (SpanData $span, $args) use ($service) {
+                $span->name = \get_class($this) . '.renderFile';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $service;
+                $span->resource = isset($args[0]) && \is_string($args[0]) ? $args[0] : $span->name;
+            }
+        );
+
+        /* I'm hoping to be able to get app.route.path and proper root resource name out of this somehow:
```
SammyK: Can this bit be removed now that route names have been implemented? :)

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,119 @@
...
+         * application, such as in the event of an unhandled exception.
+         */
+        \dd_trace_method(
+            'yii\base\Module',
+            'runAction',
+            function (SpanData $span, $args) use ($service) {
+                $span->name = \get_class($this) . '.runAction';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $service;
+                $span->resource = isset($args[0]) && \is_string($args[0]) ? $args[0] : $span->name;
```
SammyK: ❤️ the safety checks.

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
```diff
@@ -0,0 +1,119 @@
...
+                if ($root->getTag('app.endpoint') === null && isset($this->action->actionMethod)) {
+                    $controller = \get_class($this);
+                    $endpoint = "{$controller}::{$this->action->actionMethod}";
+                    $root->setTag("app.endpoint", $endpoint);
+                    $root->setTag(Tag::HTTP_URL, Url::base(true) . Url::current());
+                }
+
+                if ($root->getTag('app.route.path') === null) {
+                    $route = $this->module->requestedRoute;
+                    $named_params = [$route];
```
SammyK: Total nitpick. But the tracer uses camelCase for variable names in userland. :)

composer.json
```diff
@@ -399,6 +403,10 @@
 
         "wordpress-48-test": "@composer test -- tests/Integrations/WordPress/V4_8",
 
+        "yii-2-update": "@composer --working-dir=tests/Frameworks/Yii/Version_2_0_26 update",
```
SammyK: On the folder name, is this guaranteed to be version `2.0.26` after running composer update? We only lock the other frameworks to the minor version so that BC's in patch versions can be detected (which has happened a few times). :)

tests/Integrations/Yii/V2_0_26/ParameterizedRouteTest.php
```diff
@@ -0,0 +1,79 @@
...
+
+use DDTrace\Tag;
+use DDTrace\Tests\Common\SpanAssertion;
+use DDTrace\Tests\Common\SpanAssertionTrait;
+use DDTrace\Tests\Common\TracerTestTrait;
+use DDTrace\Tests\Common\WebFrameworkTestCase;
+use DDTrace\Tests\Frameworks\Util\Request\GetSpec;
+use DDTrace\Type;
+
+class ParameterizedRouteTest extends WebFrameworkTestCase
```
SammyK: ❤️

src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -89,6 +90,8 @@ public function __construct(array $integrations)
                 '\DDTrace\Integrations\Memcached\MemcachedSandboxedIntegration';
             $this->integrations[PDOSandboxedIntegration::NAME] =
                 '\DDTrace\Integrations\PDO\PDOSandboxedIntegration';
+            $this->integrations[YiiSandboxedIntegration::NAME] =
```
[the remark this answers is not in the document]
morrisonlevi: Ahahaha, it was meant to be alphabetized, I just goofed it in this location :)

composer.json
morrisonlevi: On it!

morrisonlevi: According to this, it's like Yii 2.0.14+:

> https://github.com/yiisoft/yii2/blob/master/ROADMAP.md

src/DDTrace/Integrations/Yii/V2/YiiIntegrationLoader.php
morrisonlevi: I'm fine with `web.request` instead of `yii.request`. @SammyK , what do you think?

SammyK: I think @labbati makes a good point. I'm cool with keeping it `web.request` :)

PR #642, opened 2019-11-05
Curl Integration: Add all available information from curl_getinfo as span tags

jstanton617:
### Description

We're already calling curl_getinfo to get the URL and status code. Rather than throwing away the rest of the array, add it as tags to the span

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Curl/CurlIntegration.php
+++ b/src/DDTrace/Integrations/Curl/CurlIntegration.php
@@ -98,7 +98,7 @@ class CurlIntegration extends Integration
             foreach ($info as $key => $val) {
                 // Datadog doesn't support arrays in tags
                 if (!is_array($val)) {
-                  // Datadog sets durations in nanoseconds - convert from seconds
+                    // Datadog sets durations in nanoseconds - convert from seconds
                     if (substr_compare($key, '_time', -5) === 0) {
                         $val = $val * 1000000000;
                     }
--- a/src/DDTrace/Integrations/Curl/CurlIntegration.php
+++ b/src/DDTrace/Integrations/Curl/CurlIntegration.php
@@ -78,31 +78,31 @@ class CurlIntegration extends Integration
 
             $span->setTag(Tag::HTTP_URL, $sanitizedUrl);
 
-            tagFromCurlInfo($span, $info, Tag::HTTP_STATUS_CODE, 'http_code');
+            CurlIntegration::tagFromCurlInfo($span, $info, Tag::HTTP_STATUS_CODE, 'http_code');
 
             // Datadog sets durations in nanoseconds - convert from seconds
             $span->setTag('duration', $info['total_time'] * 1000000000);
             unset($info['duration']);
 
-            tagFromCurlInfo($span, $info, 'network.client.ip', 'local_ip');
-            tagFromCurlInfo($span, $info, 'network.client.port', 'local_port');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.client.ip', 'local_ip');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.client.port', 'local_port');
 
-            tagFromCurlInfo($span, $info, 'network.destination.ip', 'primary_ip');
-            tagFromCurlInfo($span, $info, 'network.destination.port', 'primary_port');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.destination.ip', 'primary_ip');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.destination.port', 'primary_port');
 
-            tagFromCurlInfo($span, $info, 'network.bytes_read', 'size_download');
-            tagFromCurlInfo($span, $info, 'network.bytes_written', 'size_upload');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.bytes_read', 'size_download');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.bytes_written', 'size_upload');
 
 
             // Add the rest to a 'curl.' object
             foreach ($info as $key => $val) {
-                // Datadog doesn't support arrays in tags 
-                if(!is_array($val)) {
+                // Datadog doesn't support arrays in tags
+                if (!is_array($val)) {
                   // Datadog sets durations in nanoseconds - convert from seconds
-                  if(substr_compare($key, '_time', -5) === 0) {
-                     $val = $val * 1000000000;
-                  }
-                  $span->setTag('curl.' . $key, $val);
+                    if (substr_compare($key, '_time', -5) === 0) {
+                        $val = $val * 1000000000;
+                    }
+                    $span->setTag('curl.' . $key, $val);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


jstanton617:
This is failing tests as there doesn't appear to be a way to specify only some tags, and not `withExactTags`

Any suggestions for a way forward?

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -63,14 +63,48 @@ public static function load()
...
+            // Add the rest to a 'curl.' object
+            foreach ($info as $key => $val) {
+                // Datadog doesn't support arrays in tags 
+                if(!is_array($val)) {
+                  // Datadog sets durations in nanoseconds - convert from seconds
+                  if(substr_compare($key, '_time', -5) === 0) {
+                     $val = $val * 1000000000;
+                  }
+                  $span->setTag('curl.' . $key, $val);
+                }
```
morrisonlevi: The linter is complaining about this bit. It wants a space after `if` and before `(`, and it wants four space indents (looks like you indented with 3 spaces).

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -148,3 +182,16 @@ public static function injectDistributedTracingHeaders($ch)
...
+/**
+ * @param span $span
+ * @param tagName $tagName
+ * @param info $info
+ */
+ function tagFromCurlInfo(&$span, &$info, $tagName, $curlInfoOpt)
+ {
+     if (array_key_exists($curlInfoOpt, $info)) {
+        $span->setTag($tagName, $info[$curlInfoOpt]);
+        unset($info[$curlInfoOpt]);
```
morrisonlevi: Style nit: I think it's complaining about the indent size here too (message is less clear).

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -148,3 +182,16 @@ public static function injectDistributedTracingHeaders($ch)
         }
     }
 }
+
+/**
+ * @param span $span
+ * @param tagName $tagName
+ * @param info $info
+ */
+ function tagFromCurlInfo(&$span, &$info, $tagName, $curlInfoOpt)
```
morrisonlevi: This should be a private method on the `CurlIntegration` class so we don't introduce a global symbol that could clash with user code. The span variable should not be passed by reference.

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -63,14 +63,48 @@ public static function load()
...
+            unset($info['duration']);
+
+            tagFromCurlInfo($span, $info, 'network.client.ip', 'local_ip');
+            tagFromCurlInfo($span, $info, 'network.client.port', 'local_port');
+
+            tagFromCurlInfo($span, $info, 'network.destination.ip', 'primary_ip');
+            tagFromCurlInfo($span, $info, 'network.destination.port', 'primary_port');
+
+            tagFromCurlInfo($span, $info, 'network.bytes_read', 'size_download');
+            tagFromCurlInfo($span, $info, 'network.bytes_written', 'size_upload');
```
morrisonlevi: These look great!

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -63,14 +63,48 @@ public static function load()
...
+
+
+            // Add the rest to a 'curl.' object
```
morrisonlevi: Is the info available stable across the PHP versions we support? I'm a bit nervous about blindly adding this info.

draco2003:
> Thank you for the PR. Adding the standard `network.*` tags looks great. Great job.
> 
> I left comments inline. As mentioned in one of them, I am worried about blindly adding all curlinfo data. Is this driven by a particular need?
> 
> As for the tests, I think we'll just have to update them. This is also partly why I am worried about blindly adding all the curl flags; will it be stable enough across the PHP versions to write tests for those tags?

Based on the change log on this page ( https://www.php.net/manual/en/function.curl-getinfo.php ) it has been stable since php 5.5.0

The primary usecase for us is the timing related info to be able to break down the curl_exec into the sub components to identify issues:
CURLINFO_TOTAL_TIME - Total transaction time in seconds for last transfer
CURLINFO_NAMELOOKUP_TIME - Time in seconds until name resolving was complete
CURLINFO_CONNECT_TIME - Time in seconds it took to establish the connection
CURLINFO_PRETRANSFER_TIME - Time in seconds from start until just before file transfer begins
CURLINFO_STARTTRANSFER_TIME - Time in seconds until the first byte is about to be transferred

Though having the additional information like 
CURLINFO_PRIMARY_IP
CURLINFO_REDIRECT_COUNT
etc.. would all be useful as well, but could live without if causing a problem quantity wise, testing or otherwise.

src/DDTrace/Integrations/Curl/CurlIntegration.php
jstanton617: Lints clean now

src/DDTrace/Integrations/Curl/CurlIntegration.php
jstanton617: Made this class scoped, but making it private caused the tests to fail with 

`Error: Call to private method DDTrace\Integrations\Curl\CurlIntegration::tagFromCurlInfo() from context 'DDTrace\Tests\Integrations\Curl\CurlIntegrationTest'`

src/DDTrace/Integrations/Curl/CurlIntegration.php
jstanton617: As Dan said below, it does seem pretty stable. My thinking was to not throw away anything that we're gathering, and that it would probably vary over time. 

A nice feature of the way we're calling `curl_getinfo` is that:

> Note that private data is not included in the associative array and must be retrieved individually with the CURLINFO_PRIVATE option.

from https://www.php.net/manual/en/function.curl-getinfo.php

So it should be safe to add whatever we get

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -63,14 +63,48 @@ public static function load()
...
+
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.bytes_read', 'size_download');
+            CurlIntegration::tagFromCurlInfo($span, $info, 'network.bytes_written', 'size_upload');
+
+
+            // Add the rest to a 'curl.' object
+            foreach ($info as $key => $val) {
+                // Datadog doesn't support arrays in tags
+                if (!is_array($val)) {
+                  // Datadog sets durations in nanoseconds - convert from seconds
```
morrisonlevi: Style nit: comment needs one more space for indenting. I'm pretty sure that's the last style issue the linter picked up.

src/DDTrace/Integrations/Curl/CurlIntegration.php
jstanton617: Any updates on this PR?

morrisonlevi:
We're good to go from a product point of view -- we just need to make sure our tests work across the board. So the next step is getting the existing tests working.

draco2003:
Just wanted to nudge this to see if someone is still working on this to fix the tests, etc.. ?

SammyK:
@jstanton617 @draco2003 We'll be wrapping the tests up for this soon and we have it slated for the next release. :)

draco2003:
> @jstanton617 @draco2003 We'll be wrapping the tests up for this soon and we have it slated for the next release. :)

Great, thank you!

morrisonlevi:
@draco2003 FYI, this was included in version 0.37.0.

draco2003:
Perfect, thank you!

PR #645, opened 2019-11-08
Fix distributed trace ID handling

SammyK:
### Description

Distributed traces serialize over the wire using HTTP headers `x-datadog-trace-id` and `x-datadog-parent-id`.

Due to the way the tracer supports both internal and userland spans, neither of these propagated ID's were being synced between the sandbox and legacy API's properly and would thus break distributed traces. This issue wasn't caught by the tests since there wasn't a full integration test for distributed traces.

This PR:
- Adds integration tests for distributed traces including Synthetics requests
- Properly handles `x-datadog-trace-id` and `x-datadog-parent-id` for internal and userland spans
- Improves validation and handling of span ID's
- Fixes a small edge-case where the sampling priority would erroneously be set to the default value when the `x-datadog-sampling-priority` header was set to `0`.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,7 @@
         "test-all-54": [
             "@test-unit",
             "@test-auto-instrumentation",
+            "@test-distributed-tracing",
             "@test-integration",
             "@test-integrations-54",
             "@test-web-54"
@@ -67,6 +68,7 @@
         "test-all-56": [
             "@test-unit",
             "@test-auto-instrumentation",
+            "@test-distributed-tracing",
             "@test-integration",
             "@test-integrations-56",
             "@test-web-56"
@@ -74,6 +76,7 @@
         "test-all-70": [
             "@test-unit",
             "@test-auto-instrumentation",
+            "@test-distributed-tracing",
             "@test-integration",
             "@test-integrations-70",
             "@test-web-70"
@@ -81,6 +84,7 @@
         "test-all-71": [
             "@test-unit",
             "@test-auto-instrumentation",
+            "@test-distributed-tracing",
             "@test-integration",
             "@test-integrations-71",
             "@test-web-71"
@@ -88,6 +92,7 @@
         "test-all-72": [
             "@test-unit",
             "@test-auto-instrumentation",
+            "@test-distributed-tracing",
             "@test-integration",
             "@test-integrations-72",
             "@test-web-72"
@@ -328,6 +333,8 @@
         "test-unit": "phpunit --colors=always --configuration=phpunit.xml --testsuite=unit",
 
         "test-auto-instrumentation": "phpunit --colors=always --configuration=phpunit.xml --testsuite=auto-instrumentation",
+
+        "test-distributed-tracing": "phpunit --colors=always --configuration=phpunit.xml --testsuite=distributed-tracing",
 
         "cakephp-28-update": "@composer --working-dir=tests/Frameworks/CakePHP/Version_2_8 update",
         "cakephp-28-test": "@composer test -- --testsuite=cakephp-28-test",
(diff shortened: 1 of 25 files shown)


src/ext/ddtrace.c
```diff
@@ -692,24 +718,23 @@ static PHP_FUNCTION(dd_trace_tracer_is_limited) {
...
-                                PHP_FE(dd_trace_coms_trigger_writer_flush, NULL)
-                                    PHP_FE(dd_trace_buffer_span, arginfo_dd_trace_buffer_span)
-                                        PHP_FE(dd_trace_internal_fn, NULL)
-                                            PHP_FE(dd_trace_serialize_msgpack, arginfo_dd_trace_serialize_msgpack)
-                                                PHP_FE(dd_trace_push_span_id, NULL) PHP_FE(dd_trace_pop_span_id, NULL)
-                                                    PHP_FE(dd_trace_peek_span_id, NULL)
-                                                        PHP_FALIAS(dd_trace_generate_id, dd_trace_push_span_id, NULL)
-                                                            PHP_FE(dd_trace_closed_spans_count, NULL)
-                                                                PHP_FE(dd_trace_tracer_is_limited, NULL) ZEND_FE_END};
+    PHP_FE(dd_trace, NULL) PHP_FE(dd_trace_method, arginfo_dd_trace_method) PHP_FE(
```
morrisonlevi: Okay, we need to do something about this in the future -- this is impossible to review. These macros appear to be stable, so we can write our own, I think. Again, not for now, but just pointing it out here because I can't review this part at all.

src/ext/random.c
```diff
@@ -19,32 +21,71 @@ void ddtrace_seed_prng(TSRMLS_D) {
...
+#else
+    int i = 0;
+#endif
+    for (; i < Z_STRLEN_P(zid); i++) {
+        if (id[i] < '0' || id[i] > '9') {
+            return 0U;
+        }
+    }
+    uint64_t uid = (uint64_t)strtoull(id, NULL, 10);
+    return (uid && errno != ERANGE) ? uid : 0U;
```
morrisonlevi: You cannot inspect errno unless there was an error, so if there wasn't an error this code may misbehave.

tests/DistributedTracing/DistributedTraceTest.php
```diff
@@ -0,0 +1,100 @@
...
+    // Synthetics requests have "0" as the parent ID
+    public function testDistributedTraceWithNoParent()
+    {
+        $traces = $this->tracesFromWebRequest(function () {
+            $spec = new RequestSpec(
+                __FUNCTION__,
+                'GET',
+                '/index.php',
+                [
+                    'x-datadog-trace-id: 6017420907356617206',
```
morrisonlevi: Why this specific value? I assume it will overflow a 64 bit signed int but other than that it looks arbitrary.

composer.json
```diff
@@ -60,34 +60,39 @@
...
             "@test-integrations-71",
             "@test-web-71"
         ],
         "test-all-72": [
             "@test-unit",
             "@test-auto-instrumentation",
+            "@test-distributed-tracing",
             "@test-integration",
             "@test-integrations-72",
             "@test-web-72"
```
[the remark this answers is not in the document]
SammyK: They don't run in CI from `test-all-*`, but they run the same way the `test-auto-instrumentation` tests do [in the CI config](https://github.com/DataDog/dd-trace-php/pull/645/files#diff-1d37e48f9ceff6d8030570cd36286a61R401-R403).

src/ext/ddtrace.c
SammyK: Totally agree. I think the macros are stable all the way back to PHP 4 IIRC. :)

tests/DistributedTracing/DistributedTraceTest.php
SammyK: It doesn't overflow, it's just a real ID generated by Synthetics I captured during a Synthetics test. :)

src/ext/random.c
SammyK: Nice catch!

PR #650, opened 2019-11-13
Fix CodeIgniter namespace definition

labbati:
### Description

This PR fixes a bug in the namespace definition of the CodeIgniter integrations.

An application is impacted by this bug is all the following conditions apply:
- the application uses composer.
- ddtrace manual tracing is used (if your application only uses automatic instrumentation is not impacted).
- the application explicitly declare a dependency in composer instead of using classes provided by our extension.

What happens is that the when declaring a dependency our library in `composer.json` composer register an autoloader that place itself before our `OptionalDepsAutoloader` and `RequiredDepsAutoloader`, as it uses the `$prepend = true` flag when invoking `spl_autoload_register`.

As a result we have the following chain of autoloaders, that are attempted in this order:

1. `\Composer\Autoload\ClassLoader`
2. `\DDTrace\Bridge\OptionalDepsAutoloader`
3. `\DDTrace\Bridge\RequiredDepsAutoloader`

Now assume we are in the phase of request init hook execution, the class `\DDTrace\Bootstrap` is loaded, composer's autoloader has a definition for it and it loads the class, so step 2 and 3 are not executed. While loading `Bootstrap` a number of other classes get loaded, e.g. `DDTrace\Log\LoggingTrait`. Let me say again that so far autoloaders in position 1. and 2. have not been used.

Now assume that a request for `DDTrace\Integrations\CodeIgniter\V2_2\CodeIgniterSandboxedIntegration` class gets in. Because the folder `src/...../V2_2/` does not exist (the name was `.../V2/...`) the composer's autoloader cannot load it and will delegate the task to autoloader in position 2. Our autoloaders load ALL the files assigned to them by a brute force `require` so it happens that a class, e.g. `DDTrace\Log\LoggingTrait` is loaded twice (now and before by the composer's autoloader), causing a class redefinition fatal error.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/CodeIgniter/V2/CodeIgniterSandboxedIntegration.php
+++ b/src/DDTrace/Integrations/CodeIgniter/V2/CodeIgniterSandboxedIntegration.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace DDTrace\Integrations\CodeIgniter\V2_2;
+namespace DDTrace\Integrations\CodeIgniter\V2;
 
 use DDTrace\Configuration;
 use DDTrace\Contracts\Span;
--- a/src/DDTrace/Integrations/IntegrationsLoader.php
+++ b/src/DDTrace/Integrations/IntegrationsLoader.php
@@ -4,7 +4,7 @@ namespace DDTrace\Integrations;
 
 use DDTrace\Configuration;
 use DDTrace\Integrations\CakePHP\CakePHPIntegration;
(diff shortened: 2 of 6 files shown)


labbati:
@morrisonlevi 

> These changes look fine to me, but I don't understand what this fixes.

I added a more detailed explanation in the PR description. Thank for pointing that out, reading again the description I realized it is quite poor.

PR #652, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
### Description

Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix src --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Bootstrap.php
+++ b/src/DDTrace/Bootstrap.php
@@ -30,7 +30,7 @@ final class Bootstrap
         self::initRootSpan($tracer);
         self::registerOpenTracing();
 
-        register_shutdown_function(function () {
+        \register_shutdown_function(function () {
             dd_trace_disable_in_request(); //disable function tracing to speedup shutdown
 
             $tracer = GlobalTracer::get();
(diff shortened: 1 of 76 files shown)


draco2003:
FYI, my fixer run seems to have been overly agreesive. Will submit a new PR with significantly less changes since it should only be add the \ for optimizable functions. 

Also an alternative syntax is the :
`use function is_array;`

Let me know if that is peferrable to the inline \is_array syntax.

morrisonlevi:
I think the PR is fine as it is. I don't like `use function xxx;` up at the top, and I think only doing the optimizable ones doesn't make sense.

Of course, other team members are free to chime in :)

morrisonlevi:
If you were to update to the latest master and re-run the script it could probably be merged right away this time -- we've fixed the image access issue you hit here. Would you be interested in doing that, or should we consider this closed?

morrisonlevi:
Closed since it's pretty far out of date. Sorry that you happened to open it at a time that our builds were broken and we couldn't merge it right away.

For anyone else who reads this: we can't do "use function is_array" because we need to support PHP 5.4 and PHP 5.5 and they don't support that syntax; any future PRs like this would have to use a fully qualified function name.

PR #655, opened 2019-11-18
Migrate Zend framework integration to sandboxed api

labbati:
### Description

This PR implements a sandboxed integration for the Zend Framework and enables it by default.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/bridge/dd_require_all.php
+++ b/bridge/dd_require_all.php
@@ -62,7 +62,6 @@ require __DIR__ . '/../src/DDTrace/Integrations/Integration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/SandboxedIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/CodeIgniter/V2/CodeIgniterSandboxedIntegration.php';
-require __DIR__ . '/../src/DDTrace/Integrations/ZendFramework/ZendFrameworkIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/Web/WebIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/Predis/PredisIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/IntegrationsLoader.php';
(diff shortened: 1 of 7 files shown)


bridge/dd_require_all.php
```diff
@@ -94,6 +93,8 @@
 require __DIR__ . '/../src/DDTrace/Integrations/Yii/YiiSandboxedIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/WordPress/WordPressSandboxedIntegration.php';
 require __DIR__ . '/../src/DDTrace/Integrations/WordPress/V4/WordPressIntegrationLoader.php';
+require __DIR__ . '/../src/DDTrace/Integrations/ZendFramework/ZendFrameworkIntegration.php';
+require __DIR__ . '/../src/DDTrace/Integrations/ZendFramework/ZendFrameworkSandboxedIntegration.php';
```
morrisonlevi: Thanks for moving it alphabetically.

src/DDTrace/Integrations/ZendFramework/ZendFrameworkSandboxedIntegration.php
```diff
@@ -0,0 +1,132 @@
...
+                    $rootSpan->setResource($controller . '@' . $action . ' ' . $route);
+                    $rootSpan->setTag(Tag::HTTP_METHOD, $request->getMethod());
+                    $rootSpan->setTag(
+                        Tag::HTTP_URL,
+                        $request->getScheme() . '://' .
+                        $request->getHttpHost() .
+                        $request->getRequestUri()
+                    );
+                } catch (\Exception $e) {
+                }
```
morrisonlevi: Do you know of any exceptions that might be thrown here?
@SammyK, if an exception gets thrown then the span gets dropped, right? We shouldn't have to do a try-catch in this instance?

labbati: The problem with this is that if an exception is thrown we do not return `false`. Hence an almost empty span is generated, not having basic info including name/service/resource.

This error would cause the entire trace to be dropped.

labbati: Do you think it might make sense to add a field like `->transient` to a `SpanData` object that can be set as the first thing and causes the span to be dropped in place of the return false approach? cc @morrisonlevi @SammyK

SammyK: Dropping the span from the tracing closure it problematic since it can [create orphan spans](https://github.com/DataDog/dd-trace-php/pull/608). For hooks like this, we need a separate API that just does the pre or post hooking without any span creation. I created a card for this. :)

PR #658, opened 2019-11-19
Add polyfill.m4 to support autoconf < 2.64

morrisonlevi:
### Description

With the recent addition of `ax_execinfo` for detecting support for `<execinfo.h>` support, this caused our builds to fail on CentOS 6. The version of autoconf on CentOS 6 is 2.63, and the `ax_execinfo` macro uses `m4_ifnblank` which was added in 2.64.

This adds a polyfill for `m4_inblank` so that it will work with autoconf < 2.64, such as CentOS 6.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -666,14 +666,16 @@ workflows:
           requires: [ "Framework tests" ]
           framework_target: laravel_no_ddtrace
           name: "Laravel baseline testsuite"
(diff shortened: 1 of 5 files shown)


labbati:
This requires tests on master to be fixed before being merged. Assigning to @SammyK for review in the meantime.

PR #659, opened 2019-11-19
Refactor opcode hooks

SammyK:
### Description

Ticket: https://datadoghq.atlassian.net/browse/APMPHP-42

This PR moves the opcode hooks to the new **engine_hooks.c** file and removes the previous handlers out of the extension globals and into true globals since they do not need to be thread safe in this context. It also includes a fix for forwarding the `ZEND_DO_ICALL` and `ZEND_DO_UCALL` handlers to the proper previous opcode handlers on PHP 7.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -120,8 +120,7 @@ static PHP_MINIT_FUNCTION(ddtrace) {
 
     register_span_data_ce(TSRMLS_C);
 
-    ddtrace_opcode_minit();
-    ddtrace_compile_minit();
+    ddtrace_engine_hooks_minit();
 
     ddtrace_coms_initialize();
     ddtrace_coms_setup_atexit_hook();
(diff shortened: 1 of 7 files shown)


src/ext/ddtrace.c
```diff
@@ -120,7 +120,7 @@ static PHP_MINIT_FUNCTION(ddtrace) {
 
     register_span_data_ce(TSRMLS_C);
 
-    ddtrace_dispatch_inject(TSRMLS_C);
+    ddtrace_opcode_minit();
     ddtrace_compile_minit();
```
morrisonlevi: What do you think about exposing just `ddtrace_engine_hooks_minit` and `_mshutdown` instead of having `_minit` for opcodes, compile, etc individually?

SammyK: I love that idea. Done! :)

src/ext/engine_hooks.h
```diff
@@ -4,12 +4,11 @@
 #include <php.h>
 #include <stdint.h>
```
morrisonlevi: I don't think anything needs stdint in this header anymore, right?

Issue #656, opened 2019-11-19
Memcached doesn't work as expected using ddtrace-20180731.so


salexi:
Later Later Edit: i cannot say if it worked in the previous of ddtrace versions. It is possible that php failed to load the extension when i changed the version and flavour.

Later edit: ddtrace-20180731-zts.so -doesn't manifest the issue;

With ddtrace-20180731.so there is one issue with memcached, that is not present using ddtrace-20170718.so.

Tested on php 7.1.x and 7.3.x with laravel 5.8 and 6.x

The following code 
```
$memcached = new \Memcached();
$m->addServer('i_don_t_have_memcached_server', 11211);
$m->get('foo'));
$m->get('foo'));
var_dump($m->getResultCode());  // returns 47 which is correct

require __DIR__.'/../vendor/autoload.php';

$memcached = new \Memcached();
$m->addServer('i_don_t_have_memcached_server', 11211);
$m->get('foo'));
$m->get('foo'));
var_dump($m->getResultCode()); // returns 0 which is incorrect 
                               // since no memcached server is available
```

The issue since to be present only when ran in server environment (php-fpm or php development integrated server).

If we ran the file via `php -f public/index.php` it returns correctly 47 and 47

labbati:
Hi @salexi thanks for opening this issue and for the details provided. Looking at the snippet this can be related to #622.

salexi:
@labbati 
Do you guys have any ideea what's causing this false status codes (and messages) and if you do, is there any rough estimation till you guys sort it out ?

labbati:
Hi, @salexi, the fix for the memcached status code is in our current sprint so hopefully we will deliver by end of next week. You can still use the tracer but unfortunately you have to turn off memcached `DD_INTEGRATIONS_DISABLED=memcached`. We'll post updates here.

SammyK:
Hey @salexi! Thank you so much for your repro to help debug this issue. We've identified the issue (it's the same as #622) and should be rolling out a fix for it sometime within the next two weeks. :)

salexi:
> Hey @salexi! Thank you so much for your repro to help debug this issue. We've identified the issue (it's the same as #622) and should be rolling out a fix for it sometime within the next two weeks. :)

Hehe, no problem, glad i could help, and that the issue will be fixed promptly.

labbati:
[...]

PR #663, opened 2019-11-25
[Memcached] Prevent mutation of the result code

### LINKED ISSUE #622: "Error deleting memcache key" after upgrading to 0.33.0
Hello,

Earlier today we upgraded from 0.30.2 to 0.33.0 to get Synthetics support but found we started seeing errors deleting memcache keys. All instances upgraded to this version started reporting "Error deleting memcache key" errors. Setting and getting memcache keys seems OK. I was able to delete the memcache keys manually using memcrm tool from the same instance, so problem isn't the ElastiCache nodes or network. I disabled the memcached integration by adding it to DD_INTEGRATIONS_DISABLED in Apache and errors went away on all nodes.

Please let me know if you need any additional information from me to troubleshoot.

SammyK:
### Description

This PR fixes #622 and #656.

The Memcached integration was calling `Memcached::getServerByKey()` after the original traced call in order to append the server info to the span metadata. That method mutates the result code returned from `Memcached::getResultCode()` which was causing unexpected behavior when there was an error returned from the original call.

The Memcached integration now uses `Memcached::getServerList()` in order to append the server metadata which does not mutate the result code.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Memcached/MemcachedSandboxedIntegration.php
+++ b/src/DDTrace/Integrations/Memcached/MemcachedSandboxedIntegration.php
@@ -107,7 +107,7 @@ class MemcachedSandboxedIntegration extends SandboxedIntegration
             $integration->setCommonData($span, 'cas');
             $span->meta['memcached.cas_token'] = $args[0];
             $span->meta['memcached.query'] = 'cas ?';
-            $integration->setServerTagsByKey($span, $this, $args[1]);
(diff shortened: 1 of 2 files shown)


SammyK:
@morrisonlevi Yep! If we had a sandboxed pre-hook, that would be better :)

PR #665, opened 2019-11-26
Tag "env" in symfony integrations as env variable

raffaellopaletta:
### Description

It's now possible to pass the tag "env" in symfony integrations as environment variable

### DIFF
--- a/src/DDTrace/Integrations/Symfony/V3/SymfonyBundle.php
+++ b/src/DDTrace/Integrations/Symfony/V3/SymfonyBundle.php
@@ -59,6 +59,7 @@ class SymfonyBundle extends Bundle
         $symfonyRequestSpan->setIntegration(SymfonyIntegration::getInstance());
         $symfonyRequestSpan->setTraceAnalyticsCandidate();
         $symfonyRequestSpan->setTag(Tag::SERVICE_NAME, $appName);
+        $symfonyRequestSpan->setTag(Tag::ENV, $this->getAppEnv());
         $request = null;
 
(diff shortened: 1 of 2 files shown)


labbati:
Hi @raffaellopaletta and thanks for opening this PR. Unfortunately I think I cannot merge it at the moment, as if we introduce such a thing like env configuration then we should do for all the integrations and not only for symfony.

There are two workarounds that you can put in place in the meantime though, it is up to you which one you prefer:
1. You can set on the agent (not the php tracer) the env `DD_APM_ENV=my-env`.
2. You can use on the PHP tracer global tags: `DD_TRACE_GLOBAL_TAGS=env:my-env` (see [docs](https://docs.datadoghq.com/tracing/setup/php/#environment-variable-configuration)).

raffaellopaletta:
Hi @labbati, ok thanks! I'll investigate on workaround 2, because in my architecture the workaround number 1  would not work. I could have only one agent and two app running in two different containers, one for prod and one for staging, so I need a dynamic way to send tag env info. At the moment I'll continue to use my fork. Thanks again!

labbati:
Hi @raffaellopaletta, then WA2 is what you are looking for

labbati:
@raffaellopaletta did the workaround work for you? Just making sure, as unfortunately I either need to close this PR or we should work to make this PR more generic.

labbati:
@raffaellopaletta I will close this PR  for now. If you are interested in making this feature more generic I'll be happy to  help and  to consult with other teams to agree on  a name and  one  common approach. Thank you so much for the work on this, though.

PR #666, opened 2019-11-27
Speed up test execution in PHP 7.0, 7.1, 7.2, 7.3

labbati:
### Description

While we were debugging the issues we lately had in CI, the fact that the test suite is so slow to execute considerably slowed us down. This contributed to a delay in release. In addition to that a release often takes almost half a day because at different stages we need to build and run tests and slowest test suites take ~30 mins to complete.

This PR is a quick attempt at having builds of ~10 mins top. There are different steps we can take (e.g. smartly leveraging CircleCI cache for different versions), but for now this is a few hours fix that gets the job done.

Key aspects:

- We added a script that call composer updates on ALL the composer files in our frameworks test suites (the slowest) and cache the result for both the `vendor` folder and `composer.lock`.
- At every run the cache is restore and composer update.
- We considered the difference between using composer install vs composer update, update adds just a little bit of overhead when the vendor folder is present and it has the advantage of testing the actual edge dependencies as they would be installed by users

Why this approach? This approach provides the following benefit:
- We always execute `composer update`, so even if we forget to update the images for 1 year we are not risking to run our tests with old versions. The only effect would be for web test execution to be slower.
- Just caching the composer folder (which was the more natural way initially), wasn't providing the best possible speed up. Some examples for Laravel 5.8:
  - [2mins 4secs] `composer update` starting from scratch.
  - [1mins 10secs] `composer update` with all cached dependencies in composer's cache starting from empty `vendor` folder and no `composer.lock`.
  - [28secs]  `composer update` linking to an existing (in this case cached at image build time) `vendor` and `composer.lock` files.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
This still has build failures. Looks like maybe 7.0 and 7.3 web integration tests are still using an `apk` command even though they aren't on alpine anymore?

tooling/composer-cache
```diff
@@ -0,0 +1,96 @@
...
+set -e
+
+COMPOSER_WARMUP=~/.dd-trace-php-composer-warmup
+
+function usage {
+    echo "Usage:"
+    echo "  ./composer-update-all.sh cache"
+    echo "  ./composer-update-all.sh link-cache"
+    echo "  ./composer-update-all.sh unlink-cache"
+    echo
```
morrisonlevi: This says `composer-update-all`, but the file is named `composer-cache`; usually these match.

tooling/composer-cache
```diff
@@ -0,0 +1,96 @@
...
+    exit 1
+fi
+
+# Cache all the composer.lock files and vendor folders for future quick reuse
+cache() {
+    echo "Ensuring clean warmup dir '${COMPOSER_WARMUP}' exists (it might take a while)."
+    rm -rf ${COMPOSER_WARMUP}
+    mkdir -p ${COMPOSER_WARMUP}
+
+    for composer_json in $(find tests/Frameworks/ -name composer.json | grep -v "/vendor" | grep -v "CakePHP")
```
morrisonlevi: Why do we ignore CakePHP?

labbati: It has some strange behavior that I need to investigate further and I need to de-prioritize at the moment. I created a separate task for this if you are fine with that. So we can merge this PR and be much faster with build time.

I do not want to push this too much, so if you believe that I should better look into this let's discuss it.

dockerfiles/development/Dockerfile_54
```diff
@@ -0,0 +1,124 @@
+FROM ubuntu:14.04
```
labbati: Mostly copied from our ci repo.

dockerfiles/development/Dockerfile_56
```diff
@@ -0,0 +1,29 @@
+FROM circleci/php:5.6-cli-stretch
```
labbati: Mostly copied from our ci repo.

.circleci/config.yml
```diff
@@ -841,29 +841,29 @@ workflows:
           requires: [ 'Code Checkout' ]
           name: "PHP 70 Web integration tests"
           integration_testsuite: "test-web-70"
-          docker_image: "datadog/docker-library:ddtrace_alpine_php-7.0-debug"
-          lib_curl_command: sudo apk update ; sudo apk add curl-dev
+          docker_image: "datadog/dd-trace-ci:php-dev-7.0"
```
SammyK: Just making a documentation note for the future here: these images switch the Linux flavor from Alpine to Debian.

PR #667, opened 2019-11-27
Migrate Symfony framework integration to sandboxed api

labbati:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/IntegrationsLoader.php
+++ b/src/DDTrace/Integrations/IntegrationsLoader.php
@@ -23,6 +23,7 @@ use DDTrace\Integrations\PDO\PDOSandboxedIntegration;
 use DDTrace\Integrations\Predis\PredisIntegration;
 use DDTrace\Integrations\Slim\SlimIntegration;
 use DDTrace\Integrations\Symfony\SymfonyIntegration;
+use DDTrace\Integrations\Symfony\SymfonySandboxedIntegration;
 use DDTrace\Integrations\Web\WebIntegration;
 use DDTrace\Integrations\WordPress\WordPressSandboxedIntegration;
 use DDTrace\Integrations\Yii\YiiSandboxedIntegration;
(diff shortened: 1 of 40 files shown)


src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -94,6 +95,10 @@ public function __construct(array $integrations)
                 '\DDTrace\Integrations\Mysqli\MysqliSandboxedIntegration';
             $this->integrations[PDOSandboxedIntegration::NAME] =
                 '\DDTrace\Integrations\PDO\PDOSandboxedIntegration';
+            // Symfony integration sandboxing is disabled until we support spans auto-closing on
```
morrisonlevi: Do we know specifically why we need auto-closing spans here? I have forgotten if it was ever pointed out.

src/DDTrace/Integrations/Symfony/SymfonySandboxedIntegration.php
```diff
@@ -0,0 +1,255 @@
...
+        );
+
+        dd_trace_method(
+            'Symfony\Component\EventDispatcher\EventDispatcher',
+            'dispatch',
+            function (SpanData $span, $args) use (&$appName, &$symfonyRequestSpan, $integration, &$isSymfony2) {
+                if ($isSymfony2) {
+                    // Disabled for symfony 2
+                    return false;
+                }
```
morrisonlevi: All these symfony2 checks make me think we need to get a `dd_trace_untrace_method` or something so we can disable these once we know the Symfony version -- not necessary for this PR, of course.

tests/Frameworks/Symfony/Version_4_0/config/packages/validator.yaml
```diff
@@ -1,3 +1,2 @@
 framework:
-    validation:
```
morrisonlevi: What was this for, and why has it changed?

src/DDTrace/Integrations/IntegrationsLoader.php
labbati: Some users we have interacted with in the past are exiting in controllers.

src/DDTrace/Integrations/Symfony/SymfonySandboxedIntegration.php
labbati: This might be a good idea, actually.

tests/Frameworks/Symfony/Version_4_0/config/packages/validator.yaml
labbati: this is some internal validation module for symfony that we are not using. the configuration interface changed and given that we are not using it I think we should only use defaults

PR #679, opened 2019-12-07
Close open sandboxed spans on exit on PHP 7

morrisonlevi:
### Description
When a script calls `exit` the spans opened through the sandbox API will now get closed. This PR targets PHP 7 only.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/tests/Integrations/WordPress/V4_8/CommonScenariosTest.php
+++ b/tests/Integrations/WordPress/V4_8/CommonScenariosTest.php
@@ -46,6 +46,50 @@ final class CommonScenariosTest extends WebFrameworkTestCase
 
     public function provideSpecs()
     {
+        $children = [
+            SpanAssertion::exists(
+                'wpdb.query',
+                "SELECT option_value FROM wp_options WHERE option_name = 'WPLANG' LIMIT 1"
+            )->withChildren([
+                SpanAssertion::exists('mysqli_query'),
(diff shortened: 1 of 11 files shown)


tests/Integrations/WordPress/V4_8/CommonScenariosTest.php
```diff
@@ -46,6 +46,50 @@ public function testScenario(RequestSpec $spec, array $spanExpectations)
...
+                        AND post_type IN ('page','attachment')"
+                    )->withChildren([
+                        SpanAssertion::exists('mysqli_query'),
+                    ]),
+                ]),
+            ]),
+        ];
+
+        // In PHP 5, we don't yet support auto-closing sandbox spans with exit()
+        $children = \array_merge($children, \PHP_MAJOR_VERSION > 5 ? $exit_children : []);
```
SammyK: Nice solution. 🔝

morrisonlevi:
Do you think we need to improve integrations that use return values prior to merging in this PR?

SammyK:
That’s a valid point. I think it’s safe to merge as a malformed span will just get dropped at the Agent level so it shouldn’t negatively impact the experience. But certainly something we should address in another PR.

@labbati Would you like to make the final call on Levi's point? :)

PR #686, opened 2019-12-17
Add support for tracing without limits

labbati:
### Description

This PR adds a new capability to the PHP tracer: supporting trace without limits.

In order to achieve this we are doing 2 things with this PR:

- always send all the traces to the agent (the 1000 span/req hard limit still applies)
- provide a way to configure sampling rules for specific service/name combinations
- lazily evaluate priority sampling at the last possible time, which is:
  - when we are sending the trace to the agent OR
  - when we are injecting priority sampling into a carrier for distributed tracing

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Configuration.php
+++ b/src/DDTrace/Configuration.php
@@ -13,6 +13,13 @@ class Configuration extends AbstractConfiguration
     use LoggingTrait;
 
     /**
+     * Parsing sampling rules might be expensive so we cache values the first time we parse them.
+     *
+     * @var array
+     */
+    private $samplingRulesCache;
+
+    /**
      * Whether or not tracing is enabled.
      *
      * @return bool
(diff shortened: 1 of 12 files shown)


src/DDTrace/Configuration.php
```diff
@@ -87,6 +94,55 @@ public function getSamplingRate()
...
+     * - example: DD_TRACE_SAMPLING_RULES=[]
+     *        --> sample rate is 100%
+     * - example: DD_TRACE_SAMPLING_RULES=[{"sample_rate": 0.2}]
+     *        --> sample rate is 20%
+     * - example: DD_TRACE_SAMPLING_RULES=[{"service": "a.*", "name": "b", "sample_rate": 0.1}, {"sample_rate": 0.2}]
+     *        --> sample rate is 20% except for spans of service starting with 'a' and with name 'b' where rate is 10%
+     *
+     * Note that 'service' and 'name' is optional when when omitted the '*' pattern is assumed.
+     *
+     * @return float
```
morrisonlevi: This actually returns an array, right? Not a float?

tests/Integrations/Curl/CurlIntegrationTest.php
```diff
@@ -417,20 +417,22 @@ public function testAppendHostnameToServiceName()
 
     public function testHttpHeadersIsCorrectlySetAgain()
     {
-        $found = [];
+        $this->inRootSpan(function () use (&$found) {
```
morrisonlevi: `$found` doesn't exist outside this scope, so I don't think `use (&$found)` should be there at all.

src/DDTrace/Configuration.php
labbati: good catch 👍

tests/Integrations/Curl/CurlIntegrationTest.php
labbati: that is a leftover from a previous configuration, very good catch 👍

PR #689, opened 2019-12-18
Fix limited tracing

SammyK:
### Description

This PR removes the need to detect limited-tracing mode using `dd_trace_tracer_is_limited()` calls in tracing closures. This allows the tracer to stop tracing much earlier in the execution when the tracer is in limited mode; this in turn provides a much more optimized experience in limited mode.

Currently, to ensure `foo()` will not be instrumented in limited mode, an integration would make a call to `dd_trace_tracer_is_limited()` at the top of the tracing closure: 

```php
dd_trace_function('foo', function (\DDTrace\SpanData $span) {
    if (dd_trace_tracer_is_limited()) {
        return false;
    }
    // ...
});
```

With this PR, when limited mode is active, no calls will be instrumented unless `instrument_when_limited` is set to `1` from a new declarative API. To ensure a function or method is still instrumented even in limited tracing mode, pass a config array in place of a tracing closure to `dd_trace_function()` or `dd_trace_method()`.

For functions:

```php
dd_trace_function('foo', [
    'instrument_when_limited' => 1,
    'posthook' => function (\DDTrace\SpanData $span) {
        // ...
    }
]);
```

For methods:

```php
dd_trace_method('MyClass', 'myMethod', [
    'instrument_when_limited' => 1,
    'posthook' => function (\DDTrace\SpanData $span) {
        // ...
    }
]);
```

The only two supported config array keys in this PR are `posthook` (the tracing closure) and `instrument_when_limited` (signifies a call should be instrumented even when the tracer is in limited mode). More options can be added in the future to support a more declarative approach to integrations.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/logging.c
+++ b/src/ext/logging.c
@@ -2,13 +2,19 @@
 
 #include <php.h>
 
-void _ddtrace_log_errf(TSRMLS_FC const char *format, ...) {
+void ddtrace_log_err(char *message) {
+    TSRMLS_FETCH();
+    php_log_err(message TSRMLS_CC);
+}
+
+void _ddtrace_log_errf(const char *format, ...) {
     va_list args;
     char *buffer;
 
     va_start(args, format);
     vspprintf(&buffer, 0, format, args);
-    ddtrace_log_err(buffer);
+    TSRMLS_FETCH();
+    php_log_err(buffer TSRMLS_CC);
 
     efree(buffer);
     va_end(args);
--- a/src/ext/logging.c
+++ b/src/ext/logging.c
@@ -2,10 +2,7 @@
 
 #include <php.h>
 
-void ddtrace_log_err(char *message) {
-    TSRMLS_FETCH();
-    php_log_err(message TSRMLS_CC);
-}
+extern inline void ddtrace_log_err(char *message);
 
 void _ddtrace_log_errf(const char *format, ...) {
     va_list args;
@@ -13,8 +10,7 @@ void _ddtrace_log_errf(const char *format, ...) {
 
     va_start(args, format);
     vspprintf(&buffer, 0, format, args);
-    TSRMLS_FETCH();
-    php_log_err(buffer TSRMLS_CC);
+    ddtrace_log_err(buffer);
 
     efree(buffer);
     va_end(args);
--- a/src/ext/logging.h
+++ b/src/ext/logging.h
@@ -4,19 +4,18 @@
 
 #include "configuration.h"
 
-#if defined(ZTS) && PHP_VERSION_ID < 70000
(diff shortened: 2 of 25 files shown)


src/ext/logging.c
```diff
@@ -2,13 +2,19 @@
...
+
+void _ddtrace_log_errf(const char *format, ...) {
     va_list args;
     char *buffer;
 
     va_start(args, format);
     vspprintf(&buffer, 0, format, args);
-    ddtrace_log_err(buffer);
+    TSRMLS_FETCH();
+    php_log_err(buffer TSRMLS_CC);
```
morrisonlevi: Why not just call `ddtrace_log_err`?

src/ext/logging.c
```diff
@@ -2,13 +2,19 @@
 
 #include <php.h>
 
-void _ddtrace_log_errf(TSRMLS_FC const char *format, ...) {
+void ddtrace_log_err(char *message) {
```
morrisonlevi: I would put this in the header as `inline` (not static inline) and then here in the .c file put the `extern inline` declaration. I've done this is some places already; search for `extern inline` for examples.

src/ext/span.c
```diff
@@ -5,6 +5,7 @@
 #include <unistd.h>
 
 #include "ddtrace.h"
+#include "random.h"
```
morrisonlevi: Why include random.h?

SammyK: This is for the `ddtrace_free_span_id_stack()` call.

src/ext/logging.c
SammyK: Nice catch! 👍

src/ext/logging.c
SammyK: Excellent feedback. Updated! :)

src/ext/logging.h
```diff
@@ -4,19 +4,21 @@
 
 #include "configuration.h"
 
-#if defined(ZTS) && PHP_VERSION_ID < 70000
-#define TSRMLS_FC TSRMLS_D,
-#define ddtrace_log_errf(...) _ddtrace_log_errf(TSRMLS_C, __VA_ARGS__)
-#else
-#define TSRMLS_FC
 #define ddtrace_log_errf(...) _ddtrace_log_errf(__VA_ARGS__)
```
morrisonlevi: Why do we have this macro that just forwards to one of our own functions?

SammyK: It made sense when we had separate API's under the hood for each PHP version, but now that we've refactored it quite a bit, we can certainly get rid of it. :)

PR #700, opened 2019-12-30
Move Laravel to sanboxed api

labbati:
### Description

Moving laravel to the sanboxed api

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/DDTrace/Integrations/Laravel/LaravelSandboxedIntegration.php
```diff
@@ -0,0 +1,206 @@
...
+
+        \dd_trace_method(
+            'Illuminate\Routing\Route',
+            'run',
+            function (SpanData $span) use ($integration) {
+                $span->name = 'laravel.action';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $integration->getServiceName();
+                $span->resource = $this->uri;
+                $span->meta['integration.name'] = LaravelSandboxedIntegration::NAME;
```
SammyK: The `integration.name` tag is [only available for testing when `DD_TEST_INTEGRATION=1`](https://github.com/DataDog/dd-trace-php/blob/9662c09db74666ad4417f711a9875f788998e793/src/DDTrace/Encoders/SpanEncoder.php#L74-L82). Since internal spans do not have the concept of an integration yet, we might have to treat this in a similar manner as [setting a trace analytics candidate](https://github.com/DataDog/dd-trace-php/blob/9662c09db74666ad4417f711a9875f788998e793/src/DDTrace/Integrations/SandboxedIntegration.php#L17-L23). Or do you have other thoughts on what we could do here? :)

tooling/bin/composer-update-all
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env php
+
+<?php
+
+// Usage
```
SammyK: 👍

.circleci/config.yml
```diff
@@ -55,7 +55,7 @@ aliases:
 
   - &STEP_ATTACH_WORKSPACE
     attach_workspace:
-      at: ~/datadog
+      at: ~/app
```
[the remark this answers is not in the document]
labbati: Good point. The motivation for this is that we run tests locally in `/home/circleci/app` and in CI in `/home/circleci/datadog` which makes harder to verify things such as absolute path (for example I had to do that for a resource that was the abs path of a view in the disk).

Let me revert back this change a find a different solution.

src/DDTrace/Integrations/Laravel/LaravelSandboxedIntegration.php
```diff
@@ -0,0 +1,206 @@
...
+use DDTrace\Configuration;
+use DDTrace\GlobalTracer;
+use DDTrace\SpanData;
+use DDTrace\Integrations\SandboxedIntegration;
+use DDTrace\Scope;
+use DDTrace\Tag;
+use DDTrace\Type;
+
+/**
+ * The base Laravel integration which delegates loading to the appropriate integration version.
```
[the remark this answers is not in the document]
labbati: This is a good point. Theoretically this should be not a problem thanks to the sandboxed api but let me make the version check here and then we can have a broader conversation about what to do in such cases. Thanks for pointing that out.

src/DDTrace/Integrations/Laravel/LaravelSandboxedIntegration.php
labbati: very good idea. I like it 👍

src/DDTrace/Integrations/Laravel/LaravelSandboxedIntegration.php
```diff
@@ -0,0 +1,206 @@
...
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $integration->getServiceName();
+                if (isset($args[0]) && \is_string($args[0])) {
+                    $span->resource = $args[0];
+                }
+                $span->meta['integration.name'] = LaravelSandboxedIntegration::NAME;
+            }
+        );
+
+        // Using the legacy API only for this method because of orphaned spans in sandboxed api when `return false` is
```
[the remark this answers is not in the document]
labbati: I think in this case this is a very good idea, as `Illuminate\Foundation\ProviderRepository::load` could potentially be some info we want to give. E.g. if a lot of providers are provided than it could even be a source of slowness.

More in general, I would go with your suggestion if:
- it makes sense to trace a specific method
- the method is 'risky', e.g. accessing `$this` or `$args` or the returned value or the likes.
While I would vote to temporarily use `dd_trace` if tracing a specific method is pointless AND there are no risks in the callback code (in this case we are not interacting at all with laravel).

src/DDTrace/Tracer.php
```diff
@@ -383,6 +383,17 @@ public function getTracesAsArray()
         }
 
         $internalSpans = dd_trace_serialize_closed_spans();
+
+        // Setting global tags on internal spans, if any
```
SammyK: Should we remove this (and the test for it) since it's part of a separate PR? :)

labbati: actually I pulled in that branch, so once we merge that one this will disappear.

PR #703, opened 2019-12-31
Move dogstatsd client init to rinit

morrisonlevi:
### Description

This moves the client init to rinit and shutdown to rshutdown. This is in preparation for using the client in more places than only the sigsegv health metric, such as a heartbeat metric or tracking tracing closure exceptions.

It also does some minor cleanup, such as always emitting "Segmentation fault" (it was accidentally only emitting that message if health metrics are enabled) and allowing the caller of the dogstatsd client to control error messages.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/dogstatsd/test/client.cc
+++ b/src/dogstatsd/test/client.cc
@@ -33,7 +33,7 @@ dogstatsd_server dogstatsd_server_make() {
 
 void dogstatsd_server_listen(dogstatsd_server *server, dogstatsd_client *client,
                              const char *expected_string) {
-  socklen_t client_addr_size = sizeof(client->address);
+  socklen_t client_addr_size = sizeof(struct addrinfo);
   // 60 bytes for the IP header
   // 8 bytes for the UDP overhead
   int buffer_len = client->msg_buffer_len + 60 + 8;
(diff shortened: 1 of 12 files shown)


src/dogstatsd/test/client.cc
```diff
@@ -66,9 +66,10 @@ void _test_tags(const char *tags, const char *const_tags, const char *expect) {
   getnameinfo((sockaddr *)&server.addr, server.addr.sin_len, host, NI_MAXHOST,
               port, NI_MAXSERV, NI_NUMERICHOST | NI_NUMERICSERV);
 
-  dogstatsd_client client =
-      dogstatsd_client_ctor(host, port, buf, len, const_tags);
-  REQUIRE(!dogstatsd_client_is_default_client(&client));
+  struct addrinfo *addrs = nullptr;
```
SammyK: It looks like `nullptr` was [added in GCC 4.6](https://gcc.gnu.org/projects/cxx-status.html). Do all the supported OS's have at least this version? We upgrade [GCC in our CentOS containers](https://github.com/DataDog/docker-library/blob/master/dd-trace-php/Dockerfile_centos#L11-L13), but do we know if the officially supported GCC version in CentOS 6 is 4.6 or greater? Or is this even an issue at all since it's just in the tests?

src/ext/dogstatsd_client.c
```diff
@@ -0,0 +1,57 @@
...
+
+    if (health_metrics_enabled) {
+        host = get_dd_agent_host();
+        port = get_dd_dogstatsd_port();
+        buffer = malloc(DOGSTATSD_CLIENT_RECOMMENDED_MAX_MESSAGE_SIZE);
+        size_t len = DOGSTATSD_CLIENT_RECOMMENDED_MAX_MESSAGE_SIZE;
+
+        struct addrinfo *addrs;
+        int err;
+        if ((err = dogstatsd_client_getaddrinfo(&addrs, host, port))) {
```
SammyK: 🔝 The error handling refactoring looks great!

src/dogstatsd/test/client.cc
morrisonlevi: I use a few C++ 11 features, like closures. I think all our operating systems support C++11 except native CentOS 6/7, and we use software collections there so no worries.

morrisonlevi: This does remind me that as our client gets more advanced we probably do need to start running its standalone tests on every OS.

src/ext/dogstatsd_client.c
morrisonlevi: Thanks!

PR #716, opened 2020-01-09
Set global tags in internal spans

labbati:
### Description

Global tags were not set on internal spans generated by the sandboxed API. This PR adds such tags when the final traces is generated for serialization.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/DDTrace/Tracer.php
+++ b/src/DDTrace/Tracer.php
@@ -389,6 +389,9 @@ final class Tracer implements TracerInterface
         if ($globalTags) {
             foreach ($internalSpans as &$internalSpan) {
                 foreach ($globalTags as $globalTagName => $globalTagValue) {
+                    if (isset($internalSpan['meta'][$globalTagName])) {
+                        continue;
+                    }
                     $internalSpan['meta'][$globalTagName] = $globalTagValue;
                 }
             }
--- a/src/DDTrace/Tracer.php
+++ b/src/DDTrace/Tracer.php
@@ -384,7 +384,7 @@ final class Tracer implements TracerInterface
 
         $internalSpans = dd_trace_serialize_closed_spans();
 
-        // Setting global tags on internal tags, if any
+        // Setting global tags on internal spans, if any
         $globalTags = $this->globalConfig->getGlobalTags();
         if ($globalTags) {
             foreach ($internalSpans as &$internalSpan) {
--- a/src/DDTrace/Tracer.php
+++ b/src/DDTrace/Tracer.php
@@ -383,6 +383,17 @@ final class Tracer implements TracerInterface
         }
 
         $internalSpans = dd_trace_serialize_closed_spans();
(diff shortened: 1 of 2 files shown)


tests/Integration/TracerTest.php
```diff
@@ -0,0 +1,65 @@
...
+            ]);
+            $scope->getSpan()->finish();
+        });
+
+        $this->assertSame('custom.name', $traces[0][0]['name']);
+        $this->assertSame('local', $traces[0][0]['meta']['local_tag']);
+        $this->assertSame('global', $traces[0][0]['meta']['global_tag']);
+    }
+
+    public function testGlobalTagsArePresentOnInternalSpansByFlushTime()
```
SammyK: I can see the test passed for this one on PHP 5.4. 😮But we should probably [skip this test on PHP 5.4](https://github.com/DataDog/dd-trace-php/blob/69b684121a4ddf67b79c5b24d452cfd7ba24e8e2/tests/Unit/TracerTest.php#L275-L278) since it does not support the sandbox API and might cause this test to become flaky.

labbati: How is it possible that this test passes?

src/DDTrace/Tracer.php
```diff
@@ -383,6 +383,17 @@ public function getTracesAsArray()
         }
 
         $internalSpans = dd_trace_serialize_closed_spans();
+
+        // Setting global tags on internal spans, if any
+        $globalTags = $this->globalConfig->getGlobalTags();
+        if ($globalTags) {
+            foreach ($internalSpans as &$internalSpan) {
+                foreach ($globalTags as $globalTagName => $globalTagValue) {
+                    $internalSpan['meta'][$globalTagName] = $globalTagValue;
```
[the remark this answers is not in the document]
labbati: I am not sure from the product perspective which one is the most appropriate, but let's make it consistent with the legacy API at least, good catch 👍

tests/Integration/TracerTest.php
labbati: Ah wait we [check for it PHP](https://github.com/DataDog/dd-trace-php/blob/116c107921376f6a3a1f2aea7fe4bbd00a9bd213/src/DDTrace/Configuration.php#L202), the way we do not use sandboxing in 5.4 is not loading sandboxed integrations in the loader. Now it makes sense.

SammyK: True, but since we're using `dd_trace_method()` directly, this test was still being run in PHP 5.4. Since the `dummyMethod()` doesn't have arguments or return a value, it will work as expected in PHP 5.4. But if there were arguments or a return value, things would not work as expected. In fact... I need to submit a PR to not expose `dd_trace_method()` and `dd_trace_function()` in PHP 5.4 until we can add support for that... 🤔

PR #719, opened 2020-01-10
PHP 5.4: Remove sandbox API

SammyK:
### Description

Since the sandbox API is not supported on PHP 5.4, this PR removes the functions `dd_trace_function()` and `dd_trace_method()` on PHP 5.4 (and technically PHP 5.5 since that is not supported at all).

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -180,6 +180,7 @@
                         <file name="safe_to_string_metadata.phpt" role="test" />
                         <file name="safe_to_string_metadata_drops_invalid_keys.phpt" role="test" />
                         <file name="safe_to_string_properties.phpt" role="test" />
+                        <file name="sandbox_api_not_available_on_unsupported_versions.phpt" role="test" />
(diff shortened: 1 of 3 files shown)


morrisonlevi:
Apparently this leaks!

SammyK:
Yeah. I guess I'm not too surprised since all fatal errors in PHP leak. I'll dig into it and see if it's something we need to patch in `_dd_compile_file` or if it's one we can safely add a suppression for.

morrisonlevi:
You can switch to something like `var_dump(function_exists('dd_trace_method'));` but first double-check those leaks are just the regular fatal error leaks.

SammyK:
Thanks for the suggestion. I updated the test to use `function_exists()`. I also confirmed that the leak is a PHP fatal-error leak and not one from ddtrace.

PR #724, opened 2020-01-15
Update coms code style to C99

morrisonlevi:
### Description

Update code style to C99 and fix small potential bug in `ddtrace_coms_buffer_data` with a return-type mismatch. This potential bug is unlikely to have any effect in practice, and would be isolated to 
 disabled-by-default BETA background sender.

Code style changes include:
- Add void parameters to functions that take no args
- Convert BOOL_T->bool, TRUE->true, FALSE->false. Note that C99's bool will do a conversion from integral types; the BOOL_T will not and be vulnerable to truncation issues (which is basically what the small potential bug was).
- Remove `inline` from static functions in source files
- Remove `static`  from inline functions in header files
- Rename a few macros to be consistent with the rest of the repo
- Refactor `ddtrace_coms_buffer_data` and `ddtrace_init_read_userdata`

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/coms.c
+++ b/src/ext/coms.c
@@ -2,16 +2,17 @@
 
 #include <errno.h>
 #include <stddef.h>
-#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
 #include "coms_curl.h"
 #include "configuration.h"
-#include "env_config.h"
 #include "mpack.h"
 #include "vendor_stdatomic.h"
+
+extern inline bool ddtrace_coms_is_stack_unused(ddtrace_coms_stack_t *stack);
+extern inline bool ddtrace_coms_is_stack_free(ddtrace_coms_stack_t *stack);
 
 typedef uint32_t group_id_t;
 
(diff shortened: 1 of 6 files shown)


src/ext/coms.c
```diff
@@ -218,46 +219,42 @@ BOOL_T ddtrace_coms_rotate_stack(BOOL_T attempt_allocate_new) {
...
         size = strlen(data);
+        if (size == 0) {
+            return false;
+        }
     }
 
-    if (size == 0) {
-        return FALSE;
-    }
-    BOOL_T store_result = store_data(group_id, data, size);
```
morrisonlevi: This is where the potential bug was. `BOOL_T` is a `uint8_t` and `store_group` returns a `uint32_t`.

src/ext/coms_debug.h
```diff
@@ -1,9 +1,10 @@
-#ifndef DD_COMS_TEST_H
-#define DD_COMS_TEST_H
+#ifndef DD_COMS_DEBUG_H
```
SammyK: Nice catch! :)

PR #738, opened 2020-01-29
Update background sender

morrisonlevi:
### Description

This updates the background sender to re-use headers that have already been gathered by PHP, and also uses the PHP encoder instead of serializing msgpack on its own. As a side effect, the whole trace is buffered at once. This means in the future we can significantly simplify the background sender code.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/coms.c
+++ b/src/ext/coms.c
@@ -265,6 +265,8 @@ struct _grouped_stack_t {
 };
 
 static size_t write_array_header(char *buffer, size_t buffer_size, size_t position, uint32_t array_size) {
+    // todo: fix background sender's model now that we send complete traces
+    return 0;
     size_t free_space = buffer_size - position;
     char *data = buffer + position;
     if (array_size < 16) {
@@ -508,6 +510,11 @@ void ddtrace_deinit_read_userdata(void *userdata) {
         free(data->dest_data);
     }
     free(userdata);
+}
+
+size_t ddtrace_read_userdata_get_total_groups(void *userdata) {
+    struct _grouped_stack_t *data = userdata;
+    return data->total_groups;
 }
 
 ddtrace_coms_stack_t *ddtrace_coms_attempt_acquire_stack(void) {
--- a/src/ext/coms.c
+++ b/src/ext/coms.c
@@ -282,8 +282,6 @@ struct _grouped_stack_t {
 };
 
 static size_t write_array_header(char *buffer, size_t buffer_size, size_t position, uint32_t array_size) {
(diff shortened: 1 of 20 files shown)


morrisonlevi:
I'm adding a DO NOT MERGE label - now that I think about it there is a memory leak that wasn't caught in any of our tests. I need to get that tested and make sure we catch that leak before fixing it.

morrisonlevi:
Okay, I've added a test for the memory leak and fixed it. If the tests pass, this should be ready to go!

src/ext/coms.c
```diff
@@ -69,10 +69,14 @@ static uint32_t store_data(group_id_t group_id, const char *src, size_t size) {
     return 0;
 }
 
-static ddtrace_coms_stack_t *new_stack(void) {
+static ddtrace_coms_stack_t *new_stack(size_t min_size) {
+    size_t size = DDTRACE_COMS_STACK_SIZE;
+    while (min_size > size) {
```
labbati: Can we add a check to make sure that we do not add more then, 5MB?

morrisonlevi: Of course; I just didn't get to it before lunch and wanted to have something for you to test in the meantime.

src/ext/coms_debug.c
```diff
@@ -47,12 +47,12 @@ uint32_t ddtrace_coms_test_writers(void) {
 }
 
 uint32_t ddtrace_coms_test_consumer(void) {
-    if (!ddtrace_coms_rotate_stack(true)) {
+    if (!ddtrace_coms_rotate_stack(true, atomic_load(&ddtrace_coms_globals.stack_size))) {
         printf("error rotating stacks");
```
SammyK: I know this is old code, since this is a new code path, we probably want to replace all `printf()`'s with `ddtrace_log_debugf()` / `ddtrace_log_debug()` to ensure errors with the background sender do not result in 500 errors.

tests/ext/dd_trace_send_traces_via_thread.phpt
```diff
@@ -0,0 +1,26 @@
...
+$headers = [
+    'Datadog-Meta-Lang' => 'php',
+];
+
+// payload = [[]]
+$payload = "\x91\x90";
+
+
+dd_trace_send_traces_via_thread($url, $headers, $payload);
+
```
SammyK: Nit: Can we add an `echo 'Done';` here to ensure a `zend_bailout` didn't occur? :)

PR #757, opened 2020-02-14
Put compile_file hook behind a feature flag, default on

morrisonlevi:
### Description

Our `compile_file` hook will delay compiling files a little bit by measuring how long it takes. If those files are prone to SIGBUS then this hook will exacerbate those issues by keeping them open slightly longer.

This flag defaults to on as most customers do not have this issue, but those who do have it can turn it off by setting `DD_TRACE_MEASURE_COMPILE_TIME=false` in their environment.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/engine_hooks.c
+++ b/src/ext/engine_hooks.c
@@ -3,6 +3,7 @@
 #include <php.h>
 #include <time.h>
 
+#include "configuration.h"
 #include "ddtrace.h"
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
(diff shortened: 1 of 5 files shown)


src/ext/engine_hooks.c
```diff
@@ -41,12 +42,14 @@ static zend_op_array *_dd_compile_file(zend_file_handle *file_handle, int type T
 }
 
 static void _compile_minit(void) {
-    _prev_compile_file = zend_compile_file;
-    zend_compile_file = _dd_compile_file;
+    if (get_dd_trace_measure_compile_time()) {
```
SammyK: I'm wondering if we should keep the `zend_compile_file` hook, but just feature-flag the sys calls. As it stands right now, a user will have to restart FPM/Apache after changing `DD_TRACE_MEASURE_COMPILE_TIME`.

morrisonlevi: That's a fair point. At the same time, having the hook will still take longer than not at all; for customers experiencing SIGBUS minimizing this time is important.

morrisonlevi: What if we flag the inside and outside? This way customers moving from having it ON (default) can move back and forth until restart, but after starting with it OFF it will not have a meaningful effect until restart (will report 0 time for all compiles).

SammyK: I like it! :)

morrisonlevi: The check turns out to do a lock/unlock, so it won't fair any better performance wise by doing it on the inside. So I think as written is the best option.

I think this just needs a test and it should be good to go.

PR #762, opened 2020-02-19
Always return unaltered VM dispatch (PHP 7)

SammyK:
### Description

This PR addresses part two of a two-part refactor of the opcode handlers. The the first part was merged in #754. This PR changes the way the tracer hooks into the engine for instrumented calls for PHP 7. PHP 5 will be addressed in a separate PR.

Before this PR, instrumented calls would be forwarded from within the custom handlers for `ZEND_DO_FCALL` and friends. This required the handlers to return `ZEND_USER_OPCODE_LEAVE` to the VM which would prevent any downstream neighboring extensions from reliably hooking the function call opcodes.

This PR will start the span and run any prehook tracking closures in the function-call handlers and immediately return `ZEND_USER_OPCODE_DISPATCH` to the VM to resume normal execution. The span is stopped by hooking return opcodes. Exceptions are tracked via the `ZEND_HANDLE_EXCEPTION` pseudo opcode. Since all the opcode handlers allow the VM to continue its expected path unaltered, downstream neighboring extensions can reliably hook the function call opcodes for every call.

### Nice-Neighbor Framework Results

This change was tested on the nice-neighbor framework to ensure that the new design allowed neighboring extensions to function without adverse effects from the tracer.

#### dd_fcall_dump

The fake `dd_fcall_dump` extension prints all function-call related opcodes. This is a nice neighbor that runs any upstream neighboring handlers before running its own. If it detects a change in the VM state (if the upstream handlers return anything other than `ZEND_USER_OPCODE_DISPATCH`) it will not run its own handler.

Before this change, when `dd_fcall_dump` is loaded _after_ `ddtrace`, the following output is produced when running [drop_spans.phpt](https://github.com/DataDog/dd-trace-php/blob/383e70e/tests/ext/sandbox/drop_spans.phpt):

```
OPCODE (ZEND_DO_ICALL -> date_default_timezone_set)
OPCODE (ZEND_DO_ICALL -> dd_trace_function)
OPCODE (ZEND_DO_ICALL -> dd_trace_method)
Traced array_sum
OPCODE (ZEND_RETURN)
VM state change detected. Cannot handle OPCODE (ZEND_INIT_FCALL)
Traced array_sum
OPCODE (ZEND_RETURN)
VM state change detected. Cannot handle OPCODE (ZEND_NEW)
Traced DateTime
OPCODE (ZEND_DO_FCALL -> DateTime::format)
OPCODE (ZEND_RETURN)
VM state change detected. Cannot handle OPCODE (ZEND_FREE)
Traced DateTime
OPCODE (ZEND_DO_FCALL -> DateTime::format)
OPCODE (ZEND_RETURN)
VM state change detected. Cannot handle OPCODE (ZEND_FREE)
OPCODE (ZEND_DO_ICALL -> dd_trace_serialize_closed_spans)
OPCODE (ZEND_DO_ICALL -> array_map)
DateTime: 2000-01-01
OPCODE (ZEND_RETURN)
ArraySum: 6
OPCODE (ZEND_RETURN)
```

The "**VM state change detected**" messages are from `dd_fcall_dump`. This reveals two fundamental issues.

1. For instrumented calls, `ddtrace` is altering the VM state (returning `ZEND_USER_OPCODE_LEAVE` from the opcode handler), and `dd_fcall_dump` is unable to reliably hook those opcodes.
2. By design, extensions that alter the VM state and return `ZEND_USER_OPCODE_LEAVE` must manually increment the opline `EX(opline)++`. However, neighboring extension's opcode handlers are still run for the original opcode with the `execute_data` mutated. For example, in the message above "**VM state change detected. Cannot handle OPCODE (ZEND_INIT_FCALL)**", that message was generated by the `ZEND_DO_ICALL` opcode handler. The `dd_fcall_dump` extension does not hook `ZEND_INIT_FCALL`, but since `ddtrace` mutated the VM state, the `ZEND_INIT_FCALL` was being passed to its `ZEND_DO_ICALL` opcode handler.

After this change, the output reflects `ddtrace` being a nice neighbor by not altering the VM state and allowing `dd_fcall_dump` to successfully hook the intended opcodes with the proper handlers.

```
OPCODE (ZEND_DO_FCALL -> date_default_timezone_set)
OPCODE (ZEND_DO_FCALL -> dd_trace_function)
OPCODE (ZEND_DO_FCALL -> dd_trace_method)
OPCODE (ZEND_DO_FCALL -> array_sum)
Traced array_sum
OPCODE (ZEND_RETURN)
OPCODE (ZEND_DO_FCALL -> array_sum)
Traced array_sum
OPCODE (ZEND_RETURN)
OPCODE (ZEND_DO_FCALL -> DateTime::__construct)
Traced DateTime
OPCODE (ZEND_DO_FCALL -> DateTime::format)
OPCODE (ZEND_RETURN)
OPCODE (ZEND_DO_FCALL -> DateTime::__construct)
Traced DateTime
OPCODE (ZEND_DO_FCALL -> DateTime::format)
OPCODE (ZEND_RETURN)
OPCODE (ZEND_DO_FCALL -> dd_trace_serialize_closed_spans)
OPCODE (ZEND_DO_FCALL -> array_map)
DateTime: 2000-01-01
OPCODE (ZEND_RETURN)
ArraySum: 6
OPCODE (ZEND_RETURN)
```

You might also have noticed that all the `ZEND_DO_ICALL` opcodes are now being emitted as `ZEND_DO_FCALL`. This is a direct result of the new design overriding `zend_execute_internal` which [changes the opcode that the compiler emits](https://github.com/php/php-src/blob/c5159b3/Zend/zend_compile.c#L3100-L3123).

This design does not hook `zend_execute_ex` in order to avoid the stack limitations that that override introduces. This means userland functions will continue to emit `ZEND_DO_UCALL` as illustrated by running [exit_and_drop_span.phpt](https://github.com/DataDog/dd-trace-php/blob/31b7251/tests/ext/sandbox/exit_and_drop_span.phpt).

```
OPCODE (ZEND_DO_FCALL -> dd_trace_function)
OPCODE (ZEND_DO_UCALL -> foo)
foo()
Dropping span
OPCODE (ZEND_RETURN)
OPCODE (ZEND_EXIT)
```

#### Argument Mutation

The primary side effect of this new design relates to the mutability of arguments within an instrumented call. With this change, the tracing closure has access to the same arguments passed to the original call. If the original call mutates an argument, the posthook tracing closure will receive the mutated argument. This is the expected behavior of arguments in PHP 7.

The best way to illustrate this is with an example:

```php
function foo($a) {
    var_dump(func_get_args());
    $a = 'Dogs';
    var_dump(func_get_args());
}

foo('Cats');
```

On PHP 7, this outputs: 

```
array(1) {
  [0]=>
  string(4) "Cats"
}
array(1) {
  [0]=>
  string(4) "Dogs"
}
```

This illustrates that arguments in PHP 7 are mutable in a sense. If a posthook tracing closure were to access argument `$a`, it would be set to "Dogs". Before this change, the tracing closure would receive `$a` in its state before mutation and would be set to "Cats". If an argument needs to be accessed before mutation, the tracing closure can be run as a prehook to access the arguments before the original call.

So given:

```php
function foo($a) {
    $a = 'Dogs';
}
```

Posthook example:

```php
dd_trace_function('foo', function ($span, array $args) {
    var_dump($args[0]);
});

foo('Cats');

// string(4) "Dogs"
```

Prehook example:

```php
dd_trace_function('foo', [
    'prehook' => function ($span, array $args) {
        var_dump($args[0]);
    }
]);

foo('Cats');

// string(4) "Cats"
```

A few integrations have been changed to prehooks on PHP 7 to circumvent argument mutation. The Symfony integration was changed preceding this PR: #786.

#### Generators

Another side effect of the new design impacts function calls that return a generator with `yield` and `yield from`. A current limitation of the existing design is that generators are only traced until the first `yield`. Subsequent `yield`'s in an iteration are not instrumented. The following code illustrates this limitation:

```php
function getResults() {
    yield 1;
    yield 2;
    yield 3;
}

dd_trace_function('getResults', function() {});

$generator = getResults();
foreach ($generator as $value) {
    echo $value . PHP_EOL;
}

// Instruments getResults() one time, not three times
```

The same limitation exists in this new design, but now the return value is passed to the tracing closure differently. Before this change, the return value is provided to the tracing closure as an instance of `Generator`. With the new design, the return values are passed to the tracing closure directly from the opline (before the `ZEND_YIELD` opcode has run.) Therefore the return value provided to the tracing closure is now the _actual value_ provided by the `yield` statement, not an instance of `Generator`. The following code demonstrates this behavior change:

```php
function getResults() {
    yield 1;
    yield 2;
    yield 3;
}

dd_trace_function('getResults', function($s, $a, $retval) {
    var_dump($retval);
});

$generator = getResults();
foreach ($generator as $value) {
    echo $value . PHP_EOL;
}
```

Existing behavior output:

```
object(Generator)#3 (0) {
}
1
2
3
```

New behavior output:

```
int(1)
1
2
3
```

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/php7/engine_hooks.c
+++ b/src/ext/php7/engine_hooks.c
@@ -75,31 +75,65 @@ static bool _dd_should_trace_call(zend_execute_data *call, zend_function *fbc, d
     return true;
 }
 
-static void _dd_copy_function_args(zend_execute_data *call, zval *user_args) {
-    uint32_t i;
+#define _DD_TRACE_COPY_NULLABLE_ARG(q)      \
+    {                                       \
+        if (Z_TYPE_INFO_P(q) != IS_UNDEF) { \
+            ZVAL_DEREF(q);                  \
+            if (Z_OPT_REFCOUNTED_P(q)) {    \
+                Z_ADDREF_P(q);              \
+            }                               \
+        } else {                            \
+            q = &EG(uninitialized_zval);    \
+        }                                   \
+        ZEND_HASH_FILL_ADD(q);              \
+    }
+
+static void _dd_copy_function_args(zend_execute_data *call, zval *user_args, bool extra_args_moved) {
+    uint32_t i, first_extra_arg, extra_arg_count;
     zval *p, *q;
     uint32_t arg_count = ZEND_CALL_NUM_ARGS(call);
 
     // @see https://github.com/php/php-src/blob/PHP-7.0/Zend/zend_builtin_functions.c#L506-L562
     array_init_size(user_args, arg_count);
     if (arg_count) {
+        first_extra_arg = call->func->op_array.num_args;
+        bool has_extra_args = first_extra_arg > 0 && arg_count > first_extra_arg;
+
         zend_hash_real_init(Z_ARRVAL_P(user_args), 1);
         ZEND_HASH_FILL_PACKED(Z_ARRVAL_P(user_args)) {
             i = 0;
             p = ZEND_CALL_ARG(call, 1);
-            while (i < arg_count) {
-                q = p;
-                if (EXPECTED(Z_TYPE_INFO_P(q) != IS_UNDEF)) {
-                    ZVAL_DEREF(q);
-                    if (Z_OPT_REFCOUNTED_P(q)) {
-                        Z_ADDREF_P(q);
+            if (has_extra_args) {
+                if (extra_args_moved) {
+                    while (i < first_extra_arg) {
+                        q = p;
+                        _DD_TRACE_COPY_NULLABLE_ARG(q);
+                        p++;
+                        i++;
+                    }
+                    if (call->func->type != ZEND_INTERNAL_FUNCTION) {
+                        p = ZEND_CALL_VAR_NUM(call, call->func->op_array.last_var + call->func->op_array.T);
                     }
                 } else {
-                    q = &EG(uninitialized_zval);
+                    i = arg_count - first_extra_arg;
                 }
-                ZEND_HASH_FILL_ADD(q);
+            }
+            while (i < arg_count) {
+                q = p;
+                _DD_TRACE_COPY_NULLABLE_ARG(q);
                 p++;
                 i++;
+            }
+            /* If we are copying arguments before i_init_func_execute_data() has run, the extra agruments
+               have not yet been moved to a separate array. */
+            if (has_extra_args && !extra_args_moved) {
+                p = ZEND_CALL_VAR_NUM(call, first_extra_arg);
+                extra_arg_count = arg_count - first_extra_arg;
+                while (extra_arg_count--) {
+                    q = p;
+                    _DD_TRACE_COPY_NULLABLE_ARG(q);
+                    p++;
+                }
             }
         }
         ZEND_HASH_FILL_END();
@@ -220,7 +254,7 @@ static bool _dd_call_sandboxed_tracing_closure(ddtrace_span_t *span, zval *user_
         return true;
     }
 
-    _dd_copy_function_args(call, &user_args);
+    _dd_copy_function_args(call, &user_args, dispatch->options & DDTRACE_DISPATCH_POSTHOOK);
     if (EG(exception)) {
         exception = EG(exception);
         EG(exception) = NULL;
--- a/src/ext/php7/engine_hooks.c
+++ b/src/ext/php7/engine_hooks.c
@@ -3,6 +3,7 @@
 #include <Zend/zend_closures.h>
 #include <Zend/zend_compile.h>
 #include <Zend/zend_exceptions.h>
+#include <Zend/zend_generators.h>
 #include <Zend/zend_interfaces.h>
 #include <stdbool.h>
 
@@ -27,6 +28,10 @@ static user_opcode_handler_t _prev_fcall_handler;
 static user_opcode_handler_t _prev_fcall_by_name_handler;
 static user_opcode_handler_t _prev_return_handler;
 static user_opcode_handler_t _prev_return_by_ref_handler;
+#if PHP_VERSION_ID >= 70100
+static user_opcode_handler_t _prev_yield_handler;
+static user_opcode_handler_t _prev_yield_from_handler;
+#endif
 static user_opcode_handler_t _prev_handle_exception_handler;
 static user_opcode_handler_t _prev_exit_handler;
 
@@ -468,6 +473,19 @@ static int _dd_begin_fcall_handler(zend_execute_data *execute_data) {
         dispatch->options & (DDTRACE_DISPATCH_POSTHOOK | DDTRACE_DISPATCH_PREHOOK)) {
         return vm_retval;
     }
+
+#if PHP_VERSION_ID < 70100
+    /*
+    For PHP < 7.1: The current execute_data gets replaced in the DO_FCALL handler and freed shortly
+    afterward, so there is no way to track the execute_data that is allocated for a generator.
+    */
+    if ((current_fbc->common.fn_flags & ZEND_ACC_GENERATOR) != 0 &&
+        dispatch->options & (DDTRACE_DISPATCH_POSTHOOK | DDTRACE_DISPATCH_PREHOOK)) {
+        ddtrace_log_debug("Cannot instrument generators for PHP versions < 7.1");
+        return vm_retval;
+    }
+#endif
+
     ddtrace_dispatch_copy(dispatch);  // protecting against dispatch being freed during php code execution
     dispatch->busy = 1;               // guard against recursion, catching only topmost execution
 
@@ -557,6 +575,52 @@ static int _dd_return_by_ref_handler(zend_execute_data *execute_data) {
     _dd_do_return_handler(execute_data);
     return _prev_return_by_ref_handler ? _prev_return_by_ref_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
 }
+
+#if PHP_VERSION_ID >= 70100
+static void _dd_do_yield_handler(zend_execute_data *execute_data) {
+    ddtrace_span_t *span = DDTRACE_G(open_spans_top);
+    /*
+    Generators store their execute data on the heap and we lose the address to the original call
+    so we grab the original address from the executor globals.
+    */
+    zend_execute_data *orig_ex = (zend_execute_data *)EG(vm_stack_top);
+    if (span && span->call == orig_ex) {
+        zval rv;
+        zval *retval = NULL;
+        span->call = execute_data;
+        switch (EX(opline)->op1_type) {
+            case IS_CONST:
+#if PHP_VERSION_ID >= 70300
+                retval = RT_CONSTANT(EX(opline), EX(opline)->op1);
+#else
+                retval = EX_CONSTANT(EX(opline)->op1);
+#endif
+                break;
+            case IS_TMP_VAR:
+            case IS_VAR:
+            case IS_CV:
+                retval = EX_VAR(EX(opline)->op1.var);
+                break;
+                /* IS_UNUSED is NULL */
+        }
+        if (!retval || Z_TYPE_INFO_P(retval) == IS_UNDEF) {
+            ZVAL_NULL(&rv);
+            retval = &rv;
+        }
+        _dd_end_span(span, retval);
+    }
+}
+
+static int _dd_yield_handler(zend_execute_data *execute_data) {
+    _dd_do_yield_handler(execute_data);
+    return _prev_yield_handler ? _prev_yield_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+}
+
+static int _dd_yield_from_handler(zend_execute_data *execute_data) {
+    _dd_do_yield_handler(execute_data);
+    return _prev_yield_from_handler ? _prev_yield_from_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+}
+#endif
 
 #if PHP_MINOR_VERSION == 0
 static zend_op *_dd_get_next_catch_block(zend_execute_data *execute_data, zend_op *opline) {
@@ -700,6 +764,12 @@ void ddtrace_opcode_minit(void) {
     zend_set_user_opcode_handler(ZEND_RETURN, _dd_return_handler);
     _prev_return_by_ref_handler = zend_get_user_opcode_handler(ZEND_RETURN_BY_REF);
     zend_set_user_opcode_handler(ZEND_RETURN_BY_REF, _dd_return_by_ref_handler);
+#if PHP_VERSION_ID >= 70100
+    _prev_yield_handler = zend_get_user_opcode_handler(ZEND_YIELD);
+    zend_set_user_opcode_handler(ZEND_YIELD, _dd_yield_handler);
+    _prev_yield_from_handler = zend_get_user_opcode_handler(ZEND_YIELD_FROM);
+    zend_set_user_opcode_handler(ZEND_YIELD_FROM, _dd_yield_from_handler);
+#endif
     _prev_handle_exception_handler = zend_get_user_opcode_handler(ZEND_HANDLE_EXCEPTION);
     zend_set_user_opcode_handler(ZEND_HANDLE_EXCEPTION, _dd_handle_exception_handler);
     _prev_exit_handler = zend_get_user_opcode_handler(ZEND_EXIT);
@@ -714,6 +784,10 @@ void ddtrace_opcode_mshutdown(void) {
 
     zend_set_user_opcode_handler(ZEND_RETURN, NULL);
     zend_set_user_opcode_handler(ZEND_RETURN_BY_REF, NULL);
+#if PHP_VERSION_ID >= 70100
+    zend_set_user_opcode_handler(ZEND_YIELD, NULL);
+    zend_set_user_opcode_handler(ZEND_YIELD_FROM, NULL);
+#endif
     zend_set_user_opcode_handler(ZEND_EXIT, NULL);
 }
 
--- a/src/ext/php7/engine_hooks.c
+++ b/src/ext/php7/engine_hooks.c
@@ -448,30 +448,15 @@ static int _dd_opcode_default_dispatch(zend_execute_data *execute_data) {
     return ZEND_USER_OPCODE_DISPATCH;
 }
 
-static int _dd_begin_fcall_handler(zend_execute_data *execute_data) {
-    zend_function *current_fbc = EX(call)->func;
-    if (!current_fbc) {
-        return _dd_opcode_default_dispatch(execute_data);
-    }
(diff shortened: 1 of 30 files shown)


src/ext/php7/engine_hooks.c
```diff
@@ -547,3 +558,52 @@ void ddtrace_opcode_mshutdown(void) {
...
 }
+
+void ddtrace_execute_internal_minit(void) {
+    _prev_execute_internal = zend_execute_internal;
+    zend_execute_internal = _dd_execute_internal;
+}
+
+void ddtrace_execute_internal_mshutdown(void) { zend_execute_internal = _prev_execute_internal; }
+
+static void _dd_call_original_internal(zend_execute_data *execute_data, zval *return_value) {
```
morrisonlevi: I recommend having another function pointer and point it at either the previous execute internal hook or execute_internal; no reason to keep checking that on every call.

SammyK: Brilliant!

src/ext/php7/engine_hooks.c
```diff
@@ -547,3 +559,45 @@ void ddtrace_opcode_mshutdown(void) {
...
     zend_set_user_opcode_handler(ZEND_EXIT, NULL);
 }
+
+void ddtrace_execute_internal_minit(void) {
+    _prev_execute_internal = zend_execute_internal;
+    zend_execute_internal = _dd_execute_internal;
+    _dd_call_original_internal = !_prev_execute_internal ? execute_internal : _prev_execute_internal;
+}
+
+void ddtrace_execute_internal_mshutdown(void) { zend_execute_internal = _prev_execute_internal; }
```
morrisonlevi: Think through this with me: we only need one of `_prev_execute_internal` or `_dd_call_original_internal`, I think. If there is a non-null `zend_execute_internal` then we can set prev to that, otherwise set it to `execute_internal`. The only other reason we use this pointer is to restore it in shutdown, which can be done conditionally if the pointer is equal to `execute_internal` or not. Right?

SammyK: Good point. I think that'll be cleaner. :)

morrisonlevi: I think so too. Having two pointers always felt a bit weird before, but in all the times I've played with this I only realized this now lol.

SammyK: Yeah, it hadn't occurred to me either! :D

src/ext/php7/engine_hooks.c
```diff
@@ -545,5 +535,49 @@ void ddtrace_opcode_mshutdown(void) {
...
+    zend_function *current_fbc = EX(func);
+    if (!current_fbc) {
+        _prev_execute_internal(execute_data, return_value);
+        return;
+    }
+    ddtrace_dispatch_t *dispatch = NULL;
+    if (!_dd_should_trace_call(execute_data, current_fbc, &dispatch)) {
+        _prev_execute_internal(execute_data, return_value);
+        return;
+    }
```
morrisonlevi: What do you think about pushing the null check on the fbc into `_dd_should_trace_call`? This simplifies the caller and makes the call more robust, and I think this is the only caller of this function anyway. What do you think?

src/ext/php7/engine_hooks.c
```diff
@@ -466,44 +408,54 @@ int ddtrace_wrap_fcall(zend_execute_data *execute_data) {
...
-        DDTRACE_G(original_context).calling_fbc = current_fbc->common.scope ? current_fbc : execute_data->func;
+        ddtrace_open_span(EX(call), dispatch);
+        return ZEND_USER_OPCODE_DISPATCH;
+    }
 
-        zval *this = _dd_this(EX(call));
+    // Store original context for forwarding the call from userland
+    zend_function *previous_fbc = DDTRACE_G(original_context).fbc;
+    DDTRACE_G(original_context).fbc = current_fbc;
+    zend_function *previous_calling_fbc = DDTRACE_G(original_context).calling_fbc;
```
morrisonlevi: I didn't understand this despite the comment, sorry. Can you explain this more?

SammyK: This is ancient legacy code that stores parts of the execute_data for use with `dd_trace_forward_call()`. It's only green here because I removed the `else` block that surrounded it. :)

src/ext/php7/engine_hooks.c
SammyK: That's where it used to be, but I had to move it out because in the function call hander, FBC lives at `EX(call)->func` and in the internal function hook it lives at `EX(func)`.

src/ext/php7/engine_hooks.c
```diff
@@ -73,17 +75,35 @@ static bool _dd_should_trace_call(zend_execute_data *execute_data, zend_function
...
+                while (i < first_extra_arg) {
+                    q = p;
+                    if (EXPECTED(Z_TYPE_INFO_P(q) != IS_UNDEF)) {
+                        ZVAL_DEREF(q);
+                        if (Z_OPT_REFCOUNTED_P(q)) { 
+                            Z_ADDREF_P(q);
+                        }
+                    } else {
+                        q = &EG(uninitialized_zval);
+                    }
```
morrisonlevi: Presumably `EG(uninitialized_zval)` is `IS_UNDEF`; is that right? If so, I think this can be reduced.
```suggestion
                    ZVAL_DEREF(q);
                    Z_TRY_ADDREF_P(q);
```

src/ext/php7/engine_hooks.c
```diff
@@ -73,17 +75,35 @@ static bool _dd_should_trace_call(zend_execute_data *execute_data, zend_function
 }
 
 static void _dd_copy_function_args(zend_execute_data *call, zval *user_args) {
-    uint32_t i;
+    uint32_t i, first_extra_arg;
     zval *p, *q;
     uint32_t arg_count = ZEND_CALL_NUM_ARGS(call);
 
     // @see https://github.com/php/php-src/blob/PHP-7.0/Zend/zend_builtin_functions.c#L506-L562
```
morrisonlevi: I wonder why the builtin we looked at doesn't handle variadics.

SammyK: The builtin does handle variadics, but we removed it for our case. We didn't need it because we used to forward the call with `zend_call_function()` which apparently doesn't store the variadic args in the same way as an unaltered call. This is just restoring the args copying back to the original `func_get_args()` behavior.

src/ext/php7/engine_hooks.c
SammyK: Strangely enough, `EG(uninitialized_zval)` is [initialized to null](https://heap.space/xref/PHP-7.4/Zend/zend_execute_API.c?r=98deece6#128). I think that's for BC.

morrisonlevi:
It's going to take a while to comb over. In the meantime, can you think of a way to instrument `ZEND_DO_ICALL` instead of using `zend_instrument_internal` without the downsides of a noisy neighbor?

SammyK:
Unfortunately not. Internal function's don't emit a `ZEND_RETURN` so there is no way to hook the end of the function call to close the span. In a future PR, internal functions will have their own custom handlers but for now we have to hook `zend_execute_internal` to know when the span ends.

src/ext/php7/engine_hooks.c
```diff
@@ -479,10 +448,13 @@ static int _dd_opcode_default_dispatch(zend_execute_data *execute_data) {
...
+    zend_function *current_fbc = EX(call)->func;
+    if (!current_fbc) {
+        return _dd_opcode_default_dispatch(execute_data);
+    }
     ddtrace_dispatch_t *dispatch = NULL;
-    if (!_dd_should_trace_call(execute_data, &current_fbc, &dispatch)) {
+    if (!_dd_should_trace_call(EX(call), current_fbc, &dispatch)) {
         return _dd_opcode_default_dispatch(execute_data);
     }
     int vm_retval = _dd_opcode_default_dispatch(execute_data);
```
morrisonlevi, line 460: Let's see if we can rework this bit so that `_dd_opcode_default_dispatch` only shows up in this section of code once.

morrisonlevi: For unaltered dispatch, we should call our neighbor _at the end_. Since we don't mutate any VM state, we are free to do our own thing and then call them -- this way they have less chance of screwing up what we want to do.

For legacy dispatch, we need to call the neighbor first or not at all.

src/ext/php7/engine_hooks.c
```diff
@@ -479,80 +448,362 @@ static int _dd_opcode_default_dispatch(zend_execute_data *execute_data) {
...
 
+static int _dd_do_icall_handler(zend_execute_data *execute_data) {
+    // Assume that a neighboring extension could have incremented the opline before forwarding the handler
+    if (ZEND_DO_ICALL != EX(opline)->opcode) {
+        return _prev_icall_handler ? _prev_icall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
+    ddtrace_dispatch_t *dispatch = NULL;
+    if (!EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
+        return _prev_icall_handler ? _prev_icall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
```
morrisonlevi, line 544: I think the following change is equivalent. If so, we can apply the same technique to all of the call handlers.
```suggestion
    ddtrace_dispatch_t *dispatch = NULL;
    if (ZEND_DO_ICALL != EX(opline)->opcode || !EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
        return _prev_icall_handler ? _prev_icall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
    }
```

src/ext/php7/engine_hooks.c
```diff
@@ -479,80 +448,362 @@ static int _dd_opcode_default_dispatch(zend_execute_data *execute_data) {
...
+
+static int _dd_do_ucall_handler(zend_execute_data *execute_data) {
+    // Assume that a neighboring extension could have incremented the opline before forwarding the handler
+    if (ZEND_DO_UCALL != EX(opline)->opcode) {
+        return _prev_ucall_handler ? _prev_ucall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
+    ddtrace_dispatch_t *dispatch = NULL;
+    if (!EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
+        return _prev_ucall_handler ? _prev_ucall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
```
morrisonlevi, line 560:
```suggestion
    ddtrace_dispatch_t *dispatch = NULL;
    if (ZEND_DO_UCALL != EX(opline)->opcode || !EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
        return _prev_ucall_handler ? _prev_ucall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
    }
```

src/ext/php7/engine_hooks.c
```diff
@@ -479,80 +448,362 @@ static int _dd_opcode_default_dispatch(zend_execute_data *execute_data) {
...
+
+static int _dd_do_fcall_handler(zend_execute_data *execute_data) {
+    // Assume that a neighboring extension could have incremented the opline before forwarding the handler
+    if (ZEND_DO_FCALL != EX(opline)->opcode) {
+        return _prev_fcall_handler ? _prev_fcall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
+    ddtrace_dispatch_t *dispatch = NULL;
+    if (!EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
+        return _prev_fcall_handler ? _prev_fcall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
```
morrisonlevi, line 576:
```suggestion
    ddtrace_dispatch_t *dispatch = NULL;
    if (ZEND_DO_FCALL != EX(opline)->opcode || !EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
        return _prev_fcall_handler ? _prev_fcall_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
    }
```

src/ext/php7/engine_hooks.c
```diff
@@ -479,80 +448,362 @@ static int _dd_opcode_default_dispatch(zend_execute_data *execute_data) {
...
+
+static int _dd_do_fcall_by_name_handler(zend_execute_data *execute_data) {
+    // Assume that a neighboring extension could have incremented the opline before forwarding the handler
+    if (ZEND_DO_FCALL_BY_NAME != EX(opline)->opcode) {
+        return _prev_fcall_by_name_handler ? _prev_fcall_by_name_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
+    ddtrace_dispatch_t *dispatch = NULL;
+    if (!EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
+        return _prev_fcall_by_name_handler ? _prev_fcall_by_name_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
+    }
```
morrisonlevi, line 592: Probably need to clang-format this one.
```suggestion
    ddtrace_dispatch_t *dispatch = NULL;
    if (ZEND_DO_FCALL_BY_NAME != EX(opline)->opcode || !EX(call)->func || !_dd_should_trace_call(EX(call), EX(call)->func, &dispatch)) {
        return _prev_fcall_by_name_handler ? _prev_fcall_by_name_handler(execute_data) : ZEND_USER_OPCODE_DISPATCH;
    }
```

src/ext/php7/engine_hooks.c
```diff
@@ -578,5 +841,63 @@ void ddtrace_opcode_mshutdown(void) {
...
+    zend_function *current_fbc = EX(func);
+    if (!current_fbc) {
+        _prev_execute_internal(execute_data, return_value);
+        return;
+    }
+    ddtrace_dispatch_t *dispatch = NULL;
+    if (!_dd_should_trace_call(execute_data, current_fbc, &dispatch)) {
+        _prev_execute_internal(execute_data, return_value);
+        return;
+    }
```
morrisonlevi, line 872: Same technique.
```suggestion
    ddtrace_dispatch_t *dispatch = NULL;
    if (!current_fbc || !_dd_should_trace_call(execute_data, current_fbc, &dispatch)) {
        _prev_execute_internal(execute_data, return_value);
        return;
    }
```

src/ext/php7/engine_hooks.c
SammyK: Excellent suggestion! I've refactored them with your suggestion. I also added the opcode check that we discussed yesterday to the other handlers so that we have the same safety net across the board.

PR #773, opened 2020-02-27
Separate background sender timeout env vars from userland

morrisonlevi:
### Description

Separate background sender timeout env vars from userland. This allows them to be controlled independently, which is important because they both be in use at the same time. The background sender has higher default values than the PHP components.

Also create `DD_TRACE_BGS_ENABLED` to replace `DD_TRACE_BETA_SEND_TRACES_VIA_THREAD`
for when the feature goes into GA.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/coms.c
+++ b/src/ext/coms.c
@@ -619,12 +619,16 @@ void ddtrace_coms_curl_shutdown(void) {
     }
 }
 
+static long _dd_max_long(long a, long b) { return a >= b ? a : b; }
+
 static void _dd_curl_set_timeout(CURL *curl) {
-    curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS, get_dd_trace_agent_timeout());
+    long timeout = _dd_max_long(get_dd_trace_bgs_timeout(), get_dd_trace_agent_timeout());
+    curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS, timeout);
 }
 
 static void _dd_curl_set_connect_timeout(CURL *curl) {
-    curl_easy_setopt(curl, CURLOPT_CONNECTTIMEOUT_MS, get_dd_trace_agent_connect_timeout());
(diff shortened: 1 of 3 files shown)


src/ext/coms.c
```diff
@@ -619,12 +619,16 @@ void ddtrace_coms_curl_shutdown(void) {
     }
 }
 
+static long _dd_max_long(long a, long b) { return a >= b ? a : b; }
+
 static void _dd_curl_set_timeout(CURL *curl) {
-    curl_easy_setopt(curl, CURLOPT_TIMEOUT_MS, get_dd_trace_agent_timeout());
+    long timeout = _dd_max_long(get_dd_trace_bgs_timeout(), get_dd_trace_agent_timeout());
```
SammyK: Just to make sure we're on the same page with how this would work. If I have `DD_TRACE_BGS_TIMEOUT=500` and `DD_TRACE_AGENT_TIMEOUT=500`, and I change `DD_TRACE_AGENT_TIMEOUT=900`, this will impact both the userland Agent timeout and the BGS timeout. However, if I were to instead only change `DD_TRACE_BGS_TIMEOUT=900`, the BGS timeout will increase but the userland Agent connection timeout will be unaffected. Is that the idea? Essentially the BGS timeouts can never be less than the Agent timeouts, but they can be greater. That feels pretty intuitive to me.

morrisonlevi: Correct. If the user was experiencing timeout issues in the PHP HTTP transport component and raised this setting, then the background sender will need at least the same amount of time and perhaps more, as it may send more than one trace at a time, unlike the PHP component. Plus, increasing the background sender timeout has fewer side affects than the userland timeout because it doesn't block the PHP thread from finishing its request. So that's why it's strictly larger than the userland timeout; does that make sense?

SammyK: Indeed! Very nice solution. :)

PR #774, opened 2020-02-28
Build alpine packages on alpine and add e2e package verification

labbati:
### Description

This PR changes the way we build and verify alpine images. Notably:

- we were building alpine in debian/centos envs, we instead now build alpine extension in alpine itself, with muslc and tyical env.
- we were verifying alpine images in `php:X.X-alpine`, which have curl and other required libs installed, we are now instead verifying images in raw alpine images, so that we make sure that `libexecinfo` and `curl` are correctly marked as dependencies by the apk package.
- we are now doing e2e testing, meaning from clean alpine image to one trace received by the agent.

In the process a few changes were made to how we build and verify Alpine. The long term goal is to have build + verify as part of the same process, which is reproduced in exactly the same way both locally (dev machine) and in CI.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


Makefile
```diff
@@ -132,7 +132,7 @@ FPM_FILES=extensions/=$(EXT_DIR)/extensions \
 	docs=$(EXT_DIR)/docs README.md=$(EXT_DIR)/docs/README.md UPGRADE-0.10.md=$(EXT_DIR)/docs/UPGRADE-0.10.md\
 	src=$(EXT_DIR)/dd-trace-sources \
 	bridge=$(EXT_DIR)/dd-trace-sources
-FPM_OPTS=$(FPM_INFO_OPTS) $(FPM_DIR_OPTS) --after-install=package/post-install.sh --depends="php > 7"
+FPM_OPTS=$(FPM_INFO_OPTS) $(FPM_DIR_OPTS) --after-install=package/post-install.sh
```
labbati: Does anyone remember why this dependency on `php > 7`? It does not make a lot of sense to me here.

Makefile
```diff
@@ -144,7 +144,7 @@ $(PACKAGES_BUILD_DIR):
 .rpm: $(PACKAGES_BUILD_DIR)
 	fpm -p $(PACKAGES_BUILD_DIR) -t rpm $(FPM_OPTS) $(FPM_FILES)
 .apk: $(PACKAGES_BUILD_DIR)
-	fpm -p $(PACKAGES_BUILD_DIR) -t apk $(FPM_OPTS) --depends=libc6-compat --depends=bash --depends=libexecinfo $(FPM_FILES)
+	fpm -p $(PACKAGES_BUILD_DIR) -t apk $(FPM_OPTS) --depends=libc6-compat --depends=bash --depends=curl --depends=libexecinfo $(FPM_FILES)
```
labbati: I added `curl` as a dependency but I have mixed feelings about it, for the following reasons:
1. if users installed php from a package 99.99% it is build `--with-curl` => libcurl is installed
2. if the customer built php in the machine and has `--with-curl` => libcurl-dev is installed
3. if the customer built/installed php without curl, we would install the dependency but we cannot work in any case, so I might leave it here as a "reference" is anyone inspect the package, but I believe is pretty much useless.

dockerfiles/build-extension/fpm_apk_pax_header.patch
```diff
@@ -0,0 +1,29 @@
+From 18221cceb06860395120b56cca7e270539b74073 Mon Sep 17 00:00:00 2001
```
labbati: This file is a clone of the patch we already have in use, moved here from our docker-library repo.

dockerfiles/build-extension/Dockerfile_fpm
```diff
@@ -0,0 +1,23 @@
+# Copied from:
```
labbati: This file is (roughly) a clone of the patch we already have in use, moved here from our docker-library repo.

Makefile
morrisonlevi: For #3: in the future I think all curl communication will be done at the C level, so while today they probably will not work without PHP's curl, in the future it will.

labbati: I think that this alone is a very good selling point. Thanks for pointing that out.

dockerfiles/build-extension/Dockerfile_alpine
```diff
@@ -0,0 +1,45 @@
...
+    bison \
+    git \
+    curl-dev \
+    re2c \
+    libxml2-dev \
+    libexecinfo-dev
+
+RUN git checkout PHP-${php_version}
+RUN ./buildconf
+RUN ./configure --prefix=/php-install --without-sqlite3 --without-pdo-sqlite --with-curl
```
labbati: @SammyK @morrisonlevi any specific flag you would set in the machine that we use to build alpine images?

morrisonlevi: I wouldn't use `--enable-debug` because that changes the API, but I would make sure that the debugging symbols are there. Currently I'm debugging a segfault on Alpine that only exists with our APK, and the PHP apk doesn't have a `-dbg` version :'(

I think it's default to have symbols and Alpine `strip`s them, but just making it explicit because it's a super annoying thing.

labbati: Do you think this should be part of the images we use to build alpine or we should have dev images specific to alpine?

Makefile
SammyK: Maybe @pawelchcki would know?

dockerfiles/build-extension/Dockerfile_alpine
morrisonlevi: I think we need debugging symbols everywhere, you never know when you are going to get a sigsegv so I'd much rather have them available.

Here are the flags that the alpine container I am in used:

```
--build=x86_64-linux-musl \
--with-config-file-path=/usr/local/etc/php \
--with-config-file-scan-dir=/usr/local/etc/php/conf.d \
--enable-option-checking=fatal \
--with-mhash \
--enable-ftp \
--enable-mbstring \
--enable-mysqlnd \
--with-password-argon2 \
--with-sodium=shared \
--with-pdo-sqlite=/usr \
--with-sqlite3=/usr \
--with-curl \
--with-libedit \
--with-openssl \
--with-zlib \
build_alias=x86_64-linux-musl
```
I think we probably want to have `libedit` or `readline` so we can do `php -a`. Probably want mbstring and zlib because I think composer requires them (or is it PhpUnit?).

SammyK: Is there a reason why we can't (or shouldn't) use our existing Alpine images for this? We started moving those from `docker-library` to `dd-trace-ci` starting with `datadog/dd-trace-ci:php-7.4-debug-alpine-3.11`. https://github.com/DataDog/dd-trace-ci/blob/master/php/alpine/Dockerfile

morrisonlevi: Do the existing Alpine images have debugging symbols for PHP? If not, that alone is enough for me to vote to remake them.

dockerfiles/build-extension/Dockerfile_fpm
SammyK: Since we've been consolidating all the Dockerfiles to https://github.com/DataDog/dd-trace-ci/tree/master/php, can we move these Dockerfiles there too? I think we started a conversion about this a while ago but we didn't come to any official conclusion yet, so maybe it's time to strike that conversation up again. :D

morrisonlevi: Most recently I think we talked about bringing them back into _this_ repo, didn't we?

dockerfiles/build-extension/Dockerfile_alpine
labbati: That's a good question. What was the original motivation behind those images? I wonder if it might be a little bit difficult to follow whats going on there vs having the simplest image possible (from the point of view of installed dependencies) dedicated to the build.

dockerfiles/compile-extension/Dockerfile_alpine
```diff
@@ -0,0 +1,27 @@
...
+ADD ./install-php /usr/bin/install-php
+
+ARG php_version
+ARG php_sha
+ARG php_api
+
+ENV SRC_DIR=/usr/src
+ENV PHP_VERSION=${php_version}
+ENV PHP_API=${php_api}
+ENV PHP_URL=https://secure.php.net/get/php-${php_version}.tar.xz/from/this/mirror
```
morrisonlevi, line 16: I believe the preferred URL scheme changed to:
```suggestion
ENV PHP_URL=https://www.php.net/distributions/php-${php_api}.tar.xz
```

dockerfiles/compile-extension/Makefile
```diff
@@ -0,0 +1,17 @@
...
+# # 	docker build -t $(ALPINE_BASE_IMAGE) -f Dockerfile_alpine_base .
+
+# build_alpine_%:
+# 	docker build -t $(ALPINE_IMAGE)-$(*) --build-arg php_version=$(PHP_$(*)_VERSION) --build-arg php_sha=$(PHP_$(*)_SHA) -f Dockerfile_alpine .
+
+# shell_alpine_%: build_alpine_%
+# 	docker run --rm -ti $(ALPINE_IMAGE)-$(*) sh
+
+# install_alpine_%: build_alpine_%
+# 	docker run --rm -ti $(ALPINE_IMAGE)-$(*) install-php
```
morrisonlevi, line 17: Should we remove this file?

dockerfiles/compile-extension/build-dd-trace-php
```diff
@@ -0,0 +1,16 @@
...
+apk add --no-cache \
+    autoconf \
+    bash \
+    g++ \
+    gcc \
+    libexecinfo-dev \
+    make \
+
+make clean dist_clean
+make all CFLAGS="-std=gnu11 -O2 -Wall -Wextra -Werror" ECHO_ARG="-e"
```
morrisonlevi, line 14: We should build with debug symbols. If desired we can always strip them later using `strip --strip-debug`.
```suggestion
make all CFLAGS="-std=gnu11 -O2 -g -Wall -Wextra -Werror" ECHO_ARG="-e"
```

dockerfiles/compile-extension/docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+
+services:
+
+  7.3-alpine:
+    image: datadog/dd-trace-ci:php-compile-extension-alpine-7.3
+    build:
+      context: .
+      dockerfile: Dockerfile_alpine
+      args:
+        php_version: 7.3.2
```
morrisonlevi, line 11: This is quite old; is there a specific reason for using it instead of the latest (7.3.23 atm)?

morrisonlevi (commented):
Looking good. Have a few suggestions.

dockerfiles/compile-extension/Dockerfile_alpine
labbati: it is `${php_version}` so I am not accepting this specific suggestion and push directly the change, instead

dockerfiles/compile-extension/docker-compose.yml
labbati: yep it was there from back in the days, updating it, thanks.

Makefile
labbati: I think we can mark this as resolved as we now do e2e testing and this proved to be not used. Moreover, this is not correct as we support PHP < 7.

src/ext/php5_4/dispatch.c
```diff
@@ -58,30 +58,3 @@ zend_bool ddtrace_dispatch_store(HashTable *lookup, ddtrace_dispatch_t *dispatch
     return zend_hash_update(lookup, Z_STRVAL(dispatch->function_name), Z_STRLEN(dispatch->function_name), &dispatch,
                             sizeof(ddtrace_dispatch_t *), NULL) == SUCCESS;
 }
-
```
labbati, line 61: This caused an error due to `-Werror`. It seems not to be used and it is safe to be removed.

package/post-install.sh
```diff
@@ -11,17 +11,25 @@ CUSTOM_INI_FILE_NAME='ddtrace-custom.ini'
 
 PATH="${PATH}:/usr/local/bin"
 
-if [[ -z "$DD_TRACE_PHP_BIN" ]]; then
-    DD_TRACE_PHP_BIN=$(command -v php)
```
labbati, line 15: Wasn't able to understand which sort of problem `command -v` was causing on alpine. Moving to `$(which php || true)` but taking note of it to look into it asap

src/ext/php5_4/dispatch.c
morrisonlevi: Nice! I thought I already removed it, but turns out this is just a similar function in a similar place. Good riddance :)

package/post-install.sh
```diff
@@ -11,17 +11,25 @@ CUSTOM_INI_FILE_NAME='ddtrace-custom.ini'
...
-    # In case when 1) users import large ini files (e.g. browsercap.ini) and 2) they set memory_limit using an env
-    # variable we would fail because of the memory limit reached. In order to avoid this we set a sane memory limit
-    # that should cause no problem in any machine when our tracer is installed.
-    $DD_TRACE_PHP_BIN -d memory_limit=128M $*
-}
+# In case of .apk post-install hooks the script has no access the set of exported ENVS.
+# When 1) users import large ini files (e.g. browsercap.ini) and 2) they set memory_limit using an env
+# variable we would fail because of the memory limit reached. In order to avoid this we set a sane memory limit
+# that should cause no problem in any machine when our tracer is installed.
+RUN_PHP="$DD_TRACE_PHP_BIN -d memory_limit=128M"
```
labbati, line 32: This is an addition I was planning to do from a long time as it was reported by a user.

SammyK (approved):
LGTM! Thanks for getting this all sorted @labbati. I'd like see if @morrisonlevi has any final thoughts before merge.

package/post-install.sh
morrisonlevi: What addition is this? Are you referring to quoting it?

labbati: No this is a left-over/merge from a branch I am about to close.  Let me merge the two changes.

package/post-install.sh
```diff
@@ -11,19 +11,30 @@ CUSTOM_INI_FILE_NAME='ddtrace-custom.ini'
...
-function invoke_php(){
+function invoke_php() {
     # In case of .apk post-install hooks the script has no access the set of exported ENVS.
-    # In case when 1) users import large ini files (e.g. browsercap.ini) and 2) they set memory_limit using an env
+    # When 1) users import large ini files (e.g. browsercap.ini) and 2) they set memory_limit using an env
     # variable we would fail because of the memory limit reached. In order to avoid this we set a sane memory limit
     # that should cause no problem in any machine when our tracer is installed.
-    $DD_TRACE_PHP_BIN -d memory_limit=128M $*
+    # alias invoke_php="$DD_TRACE_PHP_BIN -d memory_limit=128M"
+    $DD_TRACE_PHP_BIN -d memory_limit=128M "$*"
```
labbati, line 34: @morrisonlevi it turned out that escaping the `$*` was enough to have it properly working in all the cases below. So I pretty much restored the old pattern.

morrisonlevi (approved):

PR #784, opened 2020-03-03
Prehook: run tracing closures before the original call (PHP 7)

SammyK:
### Description

This PR allows tracing closures to be run _before_ the original call by using the `prehook` option.

```php
dd_trace_function('foo', [
    'prehook' => function (\DDTrace\SpanData $span, array $args) {
        // ...
    }
]);
```

This feature is only supported on PHP 7.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/php7/engine_hooks.c
+++ b/src/ext/php7/engine_hooks.c
@@ -317,7 +317,7 @@ static void _dd_end_span(ddtrace_span_t *span, zval *user_retval) {
 static bool _dd_trace_dispatch(ddtrace_dispatch_t *dispatch, zend_function *fbc, zend_execute_data *execute_data) {
     ddtrace_span_t *span = ddtrace_open_span(EX(call), dispatch);
 
-    if (dispatch->options & DDTRACE_DISPATCH_PREHOOK && _dd_call_sandboxed_tracing_closure(span, NULL) == false) {
+    if ((dispatch->options & DDTRACE_DISPATCH_PREHOOK) && _dd_call_sandboxed_tracing_closure(span, NULL) == false) {
         ddtrace_drop_span();
         return false;
     }
--- a/src/ext/php7/engine_hooks.c
+++ b/src/ext/php7/engine_hooks.c
@@ -164,7 +164,7 @@ static bool _dd_execute_tracing_closure(zval *callable, zval *span_data, zend_ex
     }
     zval *this = _dd_this(call);
 
-    if (!callable || !span_data || !user_args || !user_retval) {
+    if (!callable || !span_data || !user_args) {
         if (get_dd_trace_debug()) {
             const char *fname = ZSTR_VAL(call->func->common.function_name);
             ddtrace_log_errf("Tracing closure could not be run for %s() because it is in an invalid state", fname);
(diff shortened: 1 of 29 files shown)


src/ext/php7/engine_hooks.c
```diff
@@ -301,20 +294,47 @@ static void _dd_end_span(ddtrace_span_t *span, zval *user_retval) {
...
+        ddtrace_close_span();
+    } else {
+        ddtrace_drop_span();
+    }
+}
+
+static bool _dd_trace_dispatch(ddtrace_dispatch_t *dispatch, zend_function *fbc, zend_execute_data *execute_data) {
+    ddtrace_span_t *span = ddtrace_open_span(EX(call), dispatch);
+
+    if (dispatch->options & DDTRACE_DISPATCH_PREHOOK && _dd_call_sandboxed_tracing_closure(span, NULL) == false) {
```
morrisonlevi: I've been bitten by bitwise operator precedence in the past; do you mind if we group this?
```suggestion
    if ((dispatch->options & DDTRACE_DISPATCH_PREHOOK) && _dd_call_sandboxed_tracing_closure(span, NULL) == false) {
```

tests/ext/sandbox-prehook/close_on_exit.phpt
```diff
@@ -0,0 +1,45 @@
...
+<?php if (PHP_VERSION_ID < 70000) die('skip: Prehook not supported on PHP 5'); ?>
+--FILE--
+<?php
+use DDTrace\SpanData;
+
+register_shutdown_function(function () {
+    $spans = dd_trace_serialize_closed_spans();
+    array_map(
+        function($span) {
+            echo @$span['name'], PHP_EOL;
```
morrisonlevi: What's the reason for suppressing this?
```suggestion
            echo $span['name'], PHP_EOL;
```

tests/ext/sandbox-prehook/dd_trace_function_userland.phpt
```diff
@@ -0,0 +1,41 @@
...
+use DDTrace\SpanData;
+
+var_dump(dd_trace_function('filter_to_array', ['prehook' => function (SpanData $span) {
+    $span->name = 'filter_to_array';
+}]));
+
+function filter_to_array($fn, $input) {
+    $output = array();
+    foreach ($input as $x) {
+        if (call_user_func($fn, $x)) {
```
morrisonlevi: Are we calling `call_user_func` array because of some issue on older PHP versions or for some other specific reason? On newer ones it can definitely be called directly: 
```suggestion
        if ($fn($x)) {
```

tests/ext/sandbox-prehook/close_on_exit.phpt
SammyK: Not sure of the original reason, but this is copied over as-is from the original test: https://github.com/DataDog/dd-trace-php/blob/master/tests/ext/sandbox/close-on-exit.phpt#L13 I like your suggestion though. :)

tests/ext/sandbox-prehook/dd_trace_function_userland.phpt
SammyK: This is another test that was copied as-is from the original: https://github.com/DataDog/dd-trace-php/blob/master/tests/ext/sandbox/dd_trace_function_userland.phpt#L16 But I'm 👍 for the change! :)

tests/ext/sandbox-prehook/exit_and_drop_span.phpt
```diff
@@ -0,0 +1,23 @@
+--TEST--
+[Prehook Regression] Exit gracefully handles a dropped span
```
morrisonlevi: This doesn't actually test that the span gets dropped though, does it?

SammyK: No, that's tested with `tests/ext/sandbox-prehook/drop_spans.phpt`. This test makes sure that a dropped span with `exit` does not segfault (aded in #765).

PR #790, opened 2020-03-11
Enable full-feature pecl installs 

morrisonlevi:
### Description

In the past, installing our tracer via pecl would only install the extension bits. It would not install the PHP sources nor install the ini files.

This PR will install PHP files and .ini files. When the feature is complete, full installation will be accomplished with something like:

```bash
# installs the extension and PHP sources
pecl install datadog_trace

# install the .ini files (and any other postinstall steps we need)
pear run-scripts pecl/datadog_trace  
```

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
This approach is no longer needed. Sammy landed a few PRs for version 0.48.0 that supersede this PR.

PR #792, opened 2020-03-12
Set a sane memory limit when running php from inside post-install hook

labbati:
### Description

Scenario:
- User installs a very large .ini file (e.g. for browsercap) of ~20MB.
- At the same time it sets in php.ini memory_limit=${PHP_INI_MEMORY_LIMIT} and in the Dockerfile ENV PHP_INI_MEMORY_LIMIT=128M.
- When `apk add` runs the post-install hook that we use to do some operations at installation time, it does not pass the entire set of exported environment var.
  Instead, it only pass an artificially created PATH, that's it. Any value that is exported before installing the script is not visible from inside the post-install script.
  (From [apk source code](https://git.alpinelinux.org/apk-tools/tree/src/database.c#n1918) function 'apk_db_run_script' which is the one responsible for invoking the post install).
- As a consequence PHP_INI_MEMORY_LIMIT=128M is not visible and the above is the same as executin `php -d memory_limit=${SOME_ENV_UNKNOWN} -i` which results in out of memory (I assume because it is read as `0`).

The proposed approach for now is to set a sane default memory limit of 128M, keeping in mind that this issue, as an example, also breaks the selection of the customer php binary for alpine. We have a task to fix it.

### Readiness checklist
- ~[ ] Tests added for this feature/bug.~ This was manually tested.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/package/post-install.sh
+++ b/package/post-install.sh
@@ -15,6 +15,14 @@ if [[ -z "$DD_TRACE_PHP_BIN" ]]; then
     DD_TRACE_PHP_BIN=$(command -v php)
 fi
 
+function invoke_php(){
+    # In case of .apk post-install hooks the script has no access the set of exported ENVS.
+    # In case when 1) users import large ini files (e.g. browsercap.ini) and 2) they set memory_limit using an env
+    # variable we would fail because of the memory limit reached. In order to avoid this we set a sane memory limit
+    # that should cause no problem in any machine when our tracer is installed.
+    $DD_TRACE_PHP_BIN -d memory_limit=128M $*
+}
+
 function println(){
     echo -e '###' "$@"
 }
@@ -125,7 +133,7 @@ function fail_print_and_exit() {
 }
 
 function verify_installation() {
-    $DD_TRACE_PHP_BIN -m | grep ddtrace && \
+    invoke_php -m | grep ddtrace && \
         println "Extension enabled successfully" || \
         fail_print_and_exit
 }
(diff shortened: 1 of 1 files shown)


package/post-install.sh
```diff
@@ -139,12 +147,12 @@ println
 println "Logging $DD_TRACE_PHP_BIN -i to a file"
 println
 
-$DD_TRACE_PHP_BIN -i > "$EXTENSION_LOGS_DIR/php-info.log"
+invoke_php -i > "$EXTENSION_LOGS_DIR/php-info.log"
 
-PHP_VERSION=$($DD_TRACE_PHP_BIN -i | awk '/^PHP[ \t]+API[ \t]+=>/ { print $NF }')
-PHP_CFG_DIR=$($DD_TRACE_PHP_BIN -i | grep 'Scan this dir for additional .ini files =>' | sed -e 's/Scan this dir for additional .ini files =>//g' | head -n 1 | awk '{print $1}')
+PHP_VERSION=$(invoke_php -i | awk '/^PHP[ \t]+API[ \t]+=>/ { print $NF }')
+PHP_CFG_DIR=$(invoke_php -i | grep 'Scan this dir for additional .ini files =>' | sed -e 's/Scan this dir for additional .ini files =>//g' | head -n 1 | awk '{print $1}')
```
morrisonlevi: Unrelated to this change, I noticed this chain of commands can be simplified:
 
```suggestion
PHP_CFG_DIR=$(invoke_php -i | awk -F"=>" '/Scan this dir for additional .ini files/ { sub(/^[ \t]/, "", $2) ; print $2 }')
```
A quick partial breakdown for those who don't know awk:
  - Regular expressions are delimited with `//`
  - The parameter `-F"=>"` sets the field separator to be `=>`, so the line gets broken by that instead of spaces.
  - `sub(/[^ \t]/, "", $2)` will do a substitution on the variable `$2`, in this case replace leading spaces and tabs with nothing (aka ltrim)

We use awk on the line above this, so we have `awk` available.

labbati: Nice idea. Added as part of: APMPHP-210 once we improve verification.

PR #795, opened 2020-03-16
Add Span checker tests for tag asserts

ajessu:
### Description

No new functionality here, feature or bug. Just tests added.

As part of working on #794, got curious about the SpanChecker related to the code that is checking for span tags on many tests.

Since the SpanChecker itself had no tests, created a few to explore the functionality. Pushing them upstream, as they might be helpful to continue testing this class, which is itself heavily used for testing.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
Thank you for the contribution. In this case, I don't think anything in CircleCI executed this test, though. It looks like everything in that common folder are for base classes, traits, etc, that get used by other tests, and aren't tests themselves. Because of that, they aren't run in CI anywhere. I'm not sure where to put it.

ajessu:
> Because of that, they aren't run in CI anywhere. I'm not sure where to put it.

Thanks for double checking on that. I was running them directly and missed checking on this.

I was also unsure where this would go, and even though slightly out of place as well, the next best place I can think of is: `tests/Unit/Common/`

Pushing it there in case that works better. More than open to any other suggestion.

ajessu:
Moved the test to `tests/Unit/Tests/Common/SpanCheckerTest.php`

It still feels a bit off place, but I can either use that, or track a new directory within the common directory inside `tests`, which is where these utility classes live. That last option might feel less out of place.

I'm more than fine with either option.

PR #798, opened 2020-03-17
Uri to resource name ON by default

labbati:
### Description

This PR enable the feature request URI by default. In addition to that a few changes were necessary to test real web request with different env variables.

### Readiness checklist
- [*] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.
- [*] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Tracer.php
+++ b/src/DDTrace/Tracer.php
@@ -353,6 +353,7 @@ final class Tracer implements TracerInterface
         foreach ($this->traces as $trace) {
             $traceToBeSent = [];
             foreach ($trace as $span) {
+                // If resource is empty, we normalize it the the operation name.
                 if ($span->getResource() === null) {
                     $span->setResource($span->getOperationName());
                 }
@@ -387,6 +388,7 @@ final class Tracer implements TracerInterface
         $globalTags = $this->globalConfig->getGlobalTags();
         if ($globalTags) {
             foreach ($internalSpans as &$internalSpan) {
+                // If resource is empty, we normalize it the the operation name.
                 if ($internalSpan['resource'] === null) {
                     $internalSpan['resource'] === $internalSpan['name'];
                 }
@@ -426,7 +428,7 @@ final class Tracer implements TracerInterface
             return;
         }
         $span = $scope->getSpan();
-        if ('web.request' !== $span->getResource()) {
+        if (null !== $span->getResource()) {
             return;
         }
         // Normalized URL as the resource name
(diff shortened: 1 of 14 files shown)


src/DDTrace/Tracer.php
```diff
@@ -384,6 +388,10 @@ public function getTracesAsArray()
         $globalTags = $this->globalConfig->getGlobalTags();
         if ($globalTags) {
             foreach ($internalSpans as &$internalSpan) {
+                // If resource is empty, we normalize it the the operation name.
+                if ($internalSpan['resource'] === null) {
+                    $internalSpan['resource'] === $internalSpan['name'];
```
SammyK: I'm assuming this is supposed to be an assignment, right?

tests/Integration/TracerTest.php
```diff
@@ -70,4 +72,121 @@ public function testGlobalTagsArePresentOnInternalSpansByFlushTime()
     public function dummyMethod()
     {
     }
+
+    /**
+     * When resource is set through tracer's $config object, it should be honored for CLI
```
SammyK: The extra comments are 🔝

src/DDTrace/Tracer.php
labbati: Well yes, and interestingly enough this hasn't been caught by tests, this means at least one more test is required or the exiting ones have to be modified. Thanks for noticing

labbati: [Fix](https://github.com/DataDog/dd-trace-php/pull/798/commits/4f1f873a0ef5e38f5172025cba30d043cda28800) to this comment.

I was just wondering....why you need a test suite when you have @SammyK doing your code reviews? 😎

labbati:
The failing test is the usual flaky on 5.4

tests/Integration/TracerTest.php
```diff
@@ -21,6 +22,7 @@ protected function setUp()
     protected function tearUp()
     {
         \putenv('DD_TRACE_GLOBAL_TAGS');
+        \putenv('DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED');
         parent::tearDown();
     }
```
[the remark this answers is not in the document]
SammyK: Yes, `tearUp` should be `tearDown` for sure. Good catch. :)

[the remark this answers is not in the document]
labbati: Good catch 😄! The test is not failing because in this case I was doing a `putenv` at the begin of relevant tests that do override that value. Nonetheless, it is a good habit to left the env clean after every test.

PR #800, opened 2020-03-20
Enable background sender in tests

morrisonlevi:
### Description

Building on PR #796, this enables the background sender (BGS) across the entire testsuite. Since the BGS works only with msgpack, we change the default encoding too.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/compatibility.h
+++ b/src/ext/compatibility.h
@@ -3,6 +3,7 @@
 
 #include <TSRM/TSRM.h>
 #include <Zend/zend.h>
+#include <php_version.h>
 
 #define UNUSED_1(x) (void)(x)
 #define UNUSED_2(x, y) \
--- a/tests/Common/WebFrameworkTestCase.php
+++ b/tests/Common/WebFrameworkTestCase.php
@@ -13,6 +13,8 @@ abstract class WebFrameworkTestCase extends IntegrationTestCase
 {
     use CommonScenariosDataProviderTrait;
 
+    const FLUSH_INTERVAL_MS = 333;
+
     const PORT = 9999;
 
     /**
@@ -50,10 +52,10 @@ abstract class WebFrameworkTestCase extends IntegrationTestCase
     {
         $envs = [
             'DD_TEST_INTEGRATION' => 'true',
-            'DD_TRACE_ENCODER' => 'json',
-            'DD_TRACE_AGENT_TIMEOUT' => '10000',
-            'DD_TRACE_AGENT_CONNECT_TIMEOUT' => '10000',
             'DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED' => 'true',
+            'DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS' => 1,
+            // Short flush interval by default or our tests will take all day
+            'DD_TRACE_AGENT_FLUSH_INTERVAL' => static::FLUSH_INTERVAL_MS,
         ];
 
         if (!self::isSandboxed()) {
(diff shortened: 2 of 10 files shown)


tests/Common/WebFrameworkTestCase.php
```diff
@@ -50,10 +50,18 @@ protected static function getEnvs()
     {
         $envs = [
             'DD_TEST_INTEGRATION' => 'true',
-            'DD_TRACE_ENCODER' => 'json',
+            'DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED' => 'true',
+
+            // Settings for the background sender (BGS)
+            'DD_TRACE_BGS_ENABLED' => true,
```
labbati: I wonder if we should only include settings that are required, so we test a scenario as close as possible to our users' OOTB experience.

Here I would leave: `DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS` and `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED` only (with the latter to be removed with next PR into master :) )

morrisonlevi: I think we'll want to decrease the flush interval drastically or else the test suite will take forever to run -- since is fully async we basically have to wait for 2 interval cycles. So maybe defaults for everything except:
- `DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS`
- `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED`
- `DD_TRACE_AGENT_FLUSH_INTERVAL`
What do you think? /cc @SammyK

tests/Integrations/Custom/Autoloaded/TraceSearchConfigTest.php
```diff
@@ -37,7 +37,7 @@ public function testScenario()
                     'web.request',
                     'web.request',
                     'web',
-                    'web.request'
+                    'GET /simple'
```
morrisonlevi: This should be `GET /simple`, right?

labbati: With `'DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED' => 'true'` yes.

morrisonlevi: I'm not sure why I would have had to change this then -- I didn't modify anything related to that.

src/ext/compatibility.h
```diff
@@ -3,6 +3,7 @@
 
 #include <TSRM/TSRM.h>
 #include <Zend/zend.h>
+#include <php_version.h>
```
SammyK: Why do we need this include?

tests/ext/read_c_configuration.phpt
```diff
@@ -6,6 +6,7 @@ DD_TRACE_MEMORY_LIMIT=9999
 --FILE--
 
 <?php
+$bgs_should_be_enabled = PHP_VERSION_ID >= 50600;
```
SammyK: Nit: camelCase. :)

src/ext/compatibility.h
morrisonlevi: This is what adds the `PHP_VERSION_ID` definition. My guess is that most files which include this one also include `<php.h>` or some other file previously, so in the past we didn't have issues. However, from this particular file there isn't such an include, so we need the proper header.

tests/ext/read_c_configuration.phpt
morrisonlevi: For some reason that's hard for me to remember when switching from C to PHP >.<

morrisonlevi: Fixed in acfb83efed.

tests/Integrations/Custom/Autoloaded/TraceSearchConfigTest.php
SammyK: @morrisonlevi Did you discover this locally? This test does not appear to be running on CI which would explain why it hasn't shown up before: https://github.com/DataDog/dd-trace-php/blob/master/phpunit.xml#L47

morrisonlevi: Yes, I ran it locally.

PR #804, opened 2020-03-26
Let curl calculate Content-Length header value when sending requests to the agent

labbati:
### Description

We were calculating and setting the content-length header when not using the background sender.

This was probably the source of errors `unexpected EOF` in the trace agents logs.

While we were not able to deterministically assert this, here are a few facts:
- Literature seems to relegate that error to Content-Length > actualy payload size
- I was able to recreate from curl
- I was able to recreate after hours-long random spans/trace generation using msgpack
- In long runs, I was never able to recreate this error using json
- If Content-Length is not provided, it is automatically added by the underlying curl lib
- I assume curl is better at calculating the payload compared to us using PHP's strlen on a binary var
- That explixit Content-Length header was there from forever (one of the first commits) and there is no memory or documentation on why it was there
- It works perfectly without it

### Readiness checklist
- [*] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.
- [*] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/playground/start-server.sh
+++ b/playground/start-server.sh
@@ -1,10 +1,14 @@
 #!/usr/bin/env bash
 
-SCRIPT_DIR=$(realpath $(dirname "$0"))
+set -e
+
+SCRIPT_DIR="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 PROJECT_ROOT=$(dirname "$SCRIPT_DIR")
 REQUEST_INIT_HOOK="$PROJECT_ROOT/bridge/dd_wrap_autoloader.php"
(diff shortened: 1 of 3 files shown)


labbati:
I think it is worth keeping it for two reasons:
- we have to check in any case if we are beyond a certain limit
- we do not now if a different way to calculate length is more reliable, as we are not able to reproduce it
- I also dumped ALL json for a thousands of requests and replayed ALL of them in msgpack around the time I had the EOF error. The error was not reproduced, so it might be a different combination / sorting of fields, I guess, or it can be a completely different cause.

So, I don't have a better (and tested to be better) alternative.

PR #809, opened 2020-03-31
Sandbox the Guzzle integration

morrisonlevi:
### Description

This adds sandboxing to the Guzzle integration using a posthook.

Note that the Guzzle 6 distributed tracing did not work, but curl adds the distributed headers anyway, so in practice everything is fine. I am still verifying if Guzzle 5 has the same issue.

I mention this because the current iteration does not attempt to send these in Guzzle anymore because we can't use a pre- and post- hook simultaneously -- not yet, anyway. Since we have to use a post-hook to get the HTTP response status and curl already adds the distributed tracing headers, this may be okay even on Guzzle 5.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Guzzle/GuzzleSandboxedIntegration.php
+++ b/src/DDTrace/Integrations/Guzzle/GuzzleSandboxedIntegration.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace DDTrace\Integrations\Guzzle;
+
+use DDTrace\Configuration;
+use DDTrace\GlobalTracer;
+use DDTrace\Http\Urls;
+use DDTrace\Integrations\SandboxedIntegration;
+use DDTrace\SpanData;
+use DDTrace\Tag;
+use DDTrace\Type;
+
+class GuzzleSandboxedIntegration extends SandboxedIntegration
+{
+
+    const NAME = 'guzzle';
+
+    public function getName()
+    {
+        return self::NAME;
+    }
+
+    public function init()
+    {
+        if (!Configuration::get()->isIntegrationEnabled(self::NAME)) {
+            return SandboxedIntegration::NOT_LOADED;
+        }
+
+        $tracer = GlobalTracer::get();
+        $rootScope = $tracer->getRootScope();
+        if (!$rootScope) {
+            return SandboxedIntegration::NOT_LOADED;
+        }
+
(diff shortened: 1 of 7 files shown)


src/DDTrace/Integrations/Guzzle/GuzzleSandboxedIntegration.php
```diff
@@ -0,0 +1,118 @@
...
+                $span->type = Type::HTTP_CLIENT;
+
+                if (isset($args[0])) {
+                    $integration->addRequestInfo($span, $args[0]);
+                }
+                if (isset($retval)) {
+                    $response = $retval;
+                    if (\is_a($response, 'GuzzleHttp\Promise\PromiseInterface')) {
+                        /** @var \GuzzleHttp\Promise\PromiseInterface $response */
+                        $response->then(function (\Psr\Http\Message\ResponseInterface $response) use ($span) {
```
SammyK, line 89: Confirming that this appears to be non-blocking. 🔝

tests/Integrations/Guzzle/V6/GuzzleIntegrationTest.php
```diff
@@ -125,8 +141,18 @@ public function testDistributedTracingIsPropagated()
 
         // trace is: custom
         $this->assertSame($traces[0][0]['span_id'], (int) $found['headers']['X-Datadog-Trace-Id']);
+
         // parent is: curl_exec, used under the hood
-        $this->assertSame($traces[0][2]['span_id'], (int) $found['headers']['X-Datadog-Parent-Id']);
+        $curl_exec = null;
```
SammyK, line 146: Total nit: 🐫 And same for the v5 test. 😉

morrisonlevi: Obviously this was another case of me forgetting this style piece, but in this case 'curl_exec' is the operation name -- wouldn't it make more sense to match that?

SammyK: Yeah, that's why I approved the PR with this bit being optional. My personal preference to to keep all vars with a consistent style, but given our past conversations about naming, I figured you'd take exception here. Overall I don't have a strong preference either way. Just follow your heart. 😄

PR #813, opened 2020-04-06
Sandbox predis integration

labbati:
### Description

Sandbox the predis integration, with this we have done migrating library integrations other the curl (in progress).

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Integrations/Predis/PredisSandboxedIntegration.php
```diff
@@ -0,0 +1,217 @@
...
+                }
+                $commands = $args[1];
+                $span->meta['redis.pipeline_length'] = count($commands);
+            },
+        ]);
+
+        return SandboxedIntegration::LOADED;
+    }
+
+    /**
```
labbati, line 122: I decided to duplicate the functions here instead of extracting as:

- I propose to remove the non-sandboxed integrations very soon
- only 2 out of 3 are exactly the same
- the overhead of loading yet one class when we know that the non sandboxed integration will go soon
- we still have tests (they are shared) to make sure that they do not go out of sync

tests/Integrations/Predis/PredisTest.php
```diff
@@ -167,12 +168,18 @@ public function testPredisPipeline()
             $this->assertInstanceOf('Predis\Response\Status', $responseFlush);
         });
 
-        $this->assertSpans($traces, [
+        if (Versions::phpVersionMatches('5') && static::IS_SANDBOX) {
```
morrisonlevi, line 171: The previous code seemed to work on PHP 5; why is PHP 5 + sandbox special cased?

labbati: Because it was counting the number of items in the queue before forwarding the real call.
If you do the posthook (as the prehook is not supported in php 5 for the sandboxed api) by then is already empty.

PR #814, opened 2020-04-07
Sandbox curl (PHP 7)

morrisonlevi:
### Description

This sandboxes curl on PHP 7. It has two parts:
  1. The CurlSandboxIntegration, which does a posthook on `curl_exec` to make a span and add meta.
  2. The C extension instruments `curl_exec`, `curl_copy_handle`, `curl_close`, `curl_setopt`, and `curl_setopt_array` to gather headers and to inject distributed tracing headers.

Some calls are not sandboxed yet, and there are a few other todo comments in the code.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/ext/ddtrace.c
```diff
@@ -46,6 +48,41 @@ STD_PHP_INI_BOOLEAN("ddtrace.strict_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, st
                     ddtrace_globals)
 PHP_INI_END()
 
+static int ddtrace_startup(struct _zend_extension *extension) {
+    PHP5_UNUSED(extension);
+    PHP7_UNUSED(extension);
+
+    ddtrace_curl_handlers_startup();
```
SammyK, line 55: Given we make `_dd_ext_curl_loaded` atomic, would having a one-time RINIT feature achieve this same functionality without having to run as a Zend extension hybrid?

src/ext/ddtrace.c
```diff
@@ -1000,10 +1054,20 @@ static const zend_function_entry ddtrace_functions[] = {
...
+zend_module_entry ddtrace_module_entry = {STANDARD_MODULE_HEADER,
+                                          PHP_DDTRACE_EXTNAME,
+                                          ddtrace_functions,
+                                          PHP_MINIT(ddtrace),
+                                          PHP_MSHUTDOWN(ddtrace),
+                                          PHP_RINIT(ddtrace),
+                                          PHP_RSHUTDOWN(ddtrace),
+                                          PHP_MINFO(ddtrace),
+                                          PHP_DDTRACE_VERSION,
+                                          PHP_MODULE_GLOBALS(ddtrace),
```
SammyK, line 1066: 👍

tests/Integrations/Curl/CurlIntegrationTest.php
```diff
@@ -46,6 +46,7 @@ public function tearDown()
     {
         parent::tearDown();
         putenv('DD_CURL_ANALYTICS_ENABLED');
+        putenv('DD_TRACE_HTTP_CLIENT_SPLIT_BY_DOMAIN');
```
SammyK, line 49: I don't see where this impacted the tests. Why was this added?

src/ext/php7/curl_handlers.c
```diff
@@ -0,0 +1,612 @@
...
+ZEND_TLS zend_function *_dd_ArrayKVStore_putForResource_fe = NULL;
+ZEND_TLS zend_function *_dd_ArrayKVStore_getForResource_fe = NULL;
+ZEND_TLS zend_function *_dd_ArrayKVStore_deleteResource_fe = NULL;
+ZEND_TLS zend_function *_dd_GlobalTracer_get_fe = NULL;
+ZEND_TLS zend_function *_dd_GlobalTracer_inject_fe = NULL;
+ZEND_TLS zend_function *_dd_SpanContext_ctor = NULL;
+ZEND_TLS bool _dd_curl_integration_loaded = false;
+
+// Do not pass things like "parent", "self", "static" -- fully qualified names only!
+static zend_class_entry *_dd_lookup_ce(const char *str, size_t len, bool autoload) {
```
SammyK, line 50: May I recommend removing `bool autoload` from the function signature? I think we should always call `zend_lookup_class_ex()` with `ZEND_FETCH_CLASS_NO_AUTOLOAD`. The reason is that later on we might add something that accidentally triggers the autoloader with this API which could have adverse effects. I think if we need autoloading in the future it should have a different function signature and be named in such a way that makes it explicit that we are invoking the autoloader on purpose. Something like `_dd_dangerously_lookup_ce_with_autoloading_enabled()` or something. 😆

src/ext/php7/curl_handlers.c
```diff
@@ -0,0 +1,612 @@
...
+ * @param obj May be null
+ * @param ce Must not be null
+ * @param fn_proxy The result of the method lookup is cached here, unless NULL
+ * @param fname Must not be null
+ * @param retval Must not be null
+ * @param argc The number of items in @param argv
+ * @param argv This is a plain C array e.g. `zval args[argc]`
+ * @return
+ */
+static ZEND_RESULT_CODE ddtrace_call_method(zend_object *obj, zend_class_entry *ce, zend_function **fn_proxy,
```
SammyK, line 98: I know you're still wrapping things up, but are you planning on extracting these userland-call API's so they can be used in other places?

src/ext/php7/curl_handlers.c
```diff
@@ -0,0 +1,612 @@
...
+        zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "r", &ch) == SUCCESS) {
+        zval retval;
+        zend_call_method_with_1_params(NULL, _dd_ArrayKVStore_ce, &_dd_ArrayKVStore_deleteResource_fe, "deleteresource",
+                                       &retval, ch);
+    }
+
+    _dd_curl_close_handler(INTERNAL_FUNCTION_PARAM_PASSTHRU);
+}
+
+static void _dd_instrument_curl_close(void) {
```
SammyK, line 287: I'm a bit confused by this function name as it mentions instrumentation but I don't see any instrumentation logic. I'm guessing this is just a hook for the custom handler?

Same goes for `_dd_instrument_curl_copy_handle()`, `_dd_instrument_curl_exec()`, `_dd_instrument_curl_setopt()`, and `_dd_instrument_curl_setopt_array()` below. :)

src/ext/php7/curl_handlers.c
```diff
@@ -0,0 +1,612 @@
...
+    if (!_dd_ext_curl_loaded) {
+        return;
+    }
+
+    ddtrace_error_handling eh;
+    ddtrace_backup_error_handling(&eh, EH_THROW);
+
+    zval *curl_var = zend_call_method_with_0_params(NULL, NULL, NULL, "curl_init", &retval);
+    if (curl_var && Z_TYPE_P(curl_var) == IS_RESOURCE) {
+        le_curl = Z_RES_P(curl_var)->type;
```
SammyK, line 601: Clever! 🔝 #SneakySneaks

package.xml
```diff
@@ -42,6 +42,7 @@
                     <file name="configuration_php_iface.c" role="src" />
                     <file name="coms.c" role="src" />
                     <file name="coms.h" role="src" />
+                    <file name="curl_handlers.h" role="src" />
```
SammyK, line 45: I don't want to 🚲shed too much on naming. I do have one thought though. :)

Can we name this in such a way that all extensions that we provide custom handlers for are grouped together alphabetically? For example, if we named this `handlers_curl`, when we add PDO handlers, it would be `handlers_pdo` and be positioned directly under the curl handlers. What do you think? :)

src/ext/ddtrace.c
```diff
@@ -46,6 +48,41 @@ STD_PHP_INI_BOOLEAN("ddtrace.strict_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, st
...
+
+static void ddtrace_shutdown(struct _zend_extension *extension) {
+    PHP5_UNUSED(extension);
+    PHP7_UNUSED(extension);
+}
+
+static void ddtrace_activate(void) {}
+static void ddtrace_deactivate(void) {}
+
+static zend_extension _dd_zend_extension_entry = {"ddtrace",
```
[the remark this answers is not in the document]
morrisonlevi: I'm not sure. The only bits of the zend_extension we are using in this PR is the startup -- that should be very safe.

src/ext/ddtrace.c
morrisonlevi: No. The handlers must be overwritten in startup/minit -- rinit is too late.

tests/Integrations/Curl/CurlIntegrationTest.php
morrisonlevi: The `CurlIntegrationTest.php` sets `DD_TRACE_HTTP_CLIENT_SPLIT_BY_DOMAIN` in a test and didn't clean up. If we ran `CurlIntegrationTest.php` by itself it was fine, and `CurlSandboxTest.php` by itself it was fine, but if we ran them one after another then we had a problem. So we clean it up in teardown in case the test fails that this is still unset and won't affect other tests (cascading false failures are annoying).

src/ext/php7/curl_handlers.c
morrisonlevi: This series of functions overrides the internal function handlers -- I could change "instrument" to "install"?

src/ext/php7/curl_handlers.c
morrisonlevi: I'll think about it. In any case, we only use the autoload=0 in this PR, so it's untested API surface to actually autoload.

package.xml
morrisonlevi: Sure, reasonable enough.

src/ext/ddtrace.c
SammyK: I see why doing it in MINIT is too early, but what makes RINIT too late? Doing it for every RINIT doesn't make sense, but doing it in a one-time RINIT would allow us to avoid running as a zend extension hybrid, no? Is there something that makes it unsafe to do in RINIT at all?

I guess ultimately my concern is really about converting to a zend extension hybrid which in my opinion has more unknowns in terms of side effects versus a one-time RINIT feature. But if that's our only option, let's definitely do some nice-neighbor framework testing. :)

morrisonlevi: Replacing an internal function handler in RINIT would only be safe for NTS builds, and I seem to remember a conversation in Stack Overflow PHP chat about how it may not be valid at all with JIT -- I'll try to dig that up.

Doing it in zend_extension startup should always be safe.

src/ext/php7/curl_handlers.c
morrisonlevi: Done! Found in `php7/engine_api.{c,h}`.

src/ext/php7/curl_handlers.c
morrisonlevi: I removed the autoload bit for now.

morrisonlevi:
Everything still look good after testing on the nice-neighbor framework and also running a `build-` branch -- going to merge it in.

PR #815, opened 2020-04-10
Consolidate internal monotonic and realtime clocks

morrisonlevi:
### Description

This adds three clock functions at ~both the C and PHP levels~ the C level:
  - ddtrace_monotonic_nsec
  - ddtrace_monotonic_usec
  - ddtrace_realtime_nsec

The first two are monontonic clock ticks in nanoseconds and
microseconds respectively. The latter is a system clock time
in nanoseconds.

It consolidates the existing clock functions found in span.c and
engine_hooks.c.

It also adds three typedefs, one for each clock type+unit:
  - ddtrace_monotonic_nsec_t
  - ddtrace_monotonic_usec_t
  - ddtrace_realtime_nsec_t

This helps convey the semantics of the result, such as in
ddtrace_span_t where they were previously just uint64_t.

This does not change the `DDTrace\Time::now()` call. It
appears that it is doing microseconds, but then multiplies
it by the wrong multipliers, and the encoder appends extra
zeros. In other words, I want to leave it for another PR.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/clocks.c
+++ b/src/ext/clocks.c
@@ -1,0 +1,44 @@
+#include "clocks.h"
+
+#include <php_config.h>   // for HAVE_CLOCK_GETTIME
+#include <php_version.h>  // for PHP_VERSION_ID
+
+// UNEXPECTED is in different places in PHP 5 and 7
+#if PHP_VERSION_ID < 70000
+#include <Zend/zend.h>
+#else
+#include <Zend/zend_portability.h>
+#endif
+
+#if HAVE_CLOCK_GETTIME
+
+#include <time.h>  // for clock_gettime, CLOCK_MONOTONIC
+
+ddtrace_monotonic_nsec_t ddtrace_monotonic_nsec(void) {
+    struct timespec timespec;
+    if (UNEXPECTED(clock_gettime(CLOCK_MONOTONIC, &timespec))) {
+        return 0;
+    }
+    return ((uint64_t)timespec.tv_sec) * UINT64_C(1000000000) + ((uint64_t)timespec.tv_nsec);
+}
+
+ddtrace_monotonic_usec_t ddtrace_monotonic_usec(void) {
+    struct timespec timespec;
+    if (UNEXPECTED(clock_gettime(CLOCK_MONOTONIC, &timespec))) {
+        return 0;
+    }
+    return ((uint64_t)timespec.tv_sec) * UINT64_C(1000000) + ((uint64_t)timespec.tv_nsec / UINT64_C(1000));
+}
+
+ddtrace_realtime_nsec_t ddtrace_realtime_nsec(void) {
+    struct timespec timespec;
+    if (UNEXPECTED(clock_gettime(CLOCK_REALTIME, &timespec))) {
+        return 0;
+    }
+    return ((uint64_t)timespec.tv_sec) * UINT64_C(1000000000) + ((uint64_t)timespec.tv_nsec);
+}
+#else
+ddtrace_monotonic_nsec_t ddtrace_monotonic_nsec(void) { return 0; }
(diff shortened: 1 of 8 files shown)


src/ext/clocks.c
```diff
@@ -0,0 +1,44 @@
...
+#if HAVE_CLOCK_GETTIME
+
+#include <time.h>  // for clock_gettime, CLOCK_MONOTONIC
+
+ddtrace_monotonic_nsec_t ddtrace_monotonic_nsec(void) {
+    struct timespec timespec;
+    if (UNEXPECTED(clock_gettime(CLOCK_MONOTONIC, &timespec))) {
+        return 0;
+    }
+    return ((uint64_t)timespec.tv_sec) * UINT64_C(1000000000) + ((uint64_t)timespec.tv_nsec);
```
SammyK, line 22:
```
Seconds * 1,000,000,000 "one billion" + nano "billionth of a second"
```
✅

src/ext/clocks.c
```diff
@@ -0,0 +1,44 @@
...
+}
+
+ddtrace_monotonic_usec_t ddtrace_monotonic_usec(void) {
+    struct timespec timespec;
+    if (UNEXPECTED(clock_gettime(CLOCK_MONOTONIC, &timespec))) {
+        return 0;
+    }
+    return ((uint64_t)timespec.tv_sec) * UINT64_C(100000) + ((uint64_t)timespec.tv_nsec / UINT64_C(1000));
```
SammyK, line 30:
```
Seconds * 100,000 "one hundred thousand" + (nano "billionth of a second" / 1,000 "one thousand")
```

I'm assuming this is for microseconds (µs), which is one millionth of a second, correct? The nano -> micro conversion looks correct, but the seconds -> micro seems like it's missing a zero to make it a million `UINT64_C(1000000)`.

It's times like these that I wish C supported int literal separators. :)

src/ext/clocks.c
```diff
@@ -0,0 +1,44 @@
...
+}
+
+ddtrace_realtime_nsec_t ddtrace_realtime_nsec(void) {
+    struct timespec timespec;
+    if (UNEXPECTED(clock_gettime(CLOCK_REALTIME, &timespec))) {
+        return 0;
+    }
+    return ((uint64_t)timespec.tv_sec) * UINT64_C(1000000000) + ((uint64_t)timespec.tv_nsec);
```
SammyK, line 38:
```
Seconds * 1,000,000,000 "one billion" + nano "billionth of a second"
```
✅

src/ext/clocks.c
```diff
@@ -0,0 +1,44 @@
...
+}
+#else
+ddtrace_monotonic_nsec_t ddtrace_monotonic_nsec(void) { return 0; }
```
SammyK, line 41: Love the graceful fallbacks.

tests/ext/ddtrace_clocks.phpt
```diff
@@ -0,0 +1,12 @@
+--TEST--
+Test that clocks return > 0
+--FILE--
+<?php
+var_dump(ddtrace_monotonic_nsec() > 0);
```
SammyK, line 5: I'm not sure how I feel about exposing these to userland unless we are using them in `DDTrace\Time`. Would you be OK with holding off exposing these userland functions until we need to use them in userland?

morrisonlevi: Very reasonable; will get these removed tomorrow morning for another CR. Thanks, Sammy!

src/ext/clocks.c
morrisonlevi: Nice catch! Literal separators would definitely help here.

PR #824, opened 2020-04-15
Perf-support linuxkit

pawelchcki:

### DIFF
--- a/tests/overhead/docker-compose.yml
+++ b/tests/overhead/docker-compose.yml
@@ -1,7 +1,6 @@
-version: '3.6'
+version: "3.6"
 
 services:
-
   # nginx:
   #   image: overhead-nginx
   #   ports:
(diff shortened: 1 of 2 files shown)


tests/overhead/docker-compose.yml
```diff
@@ -55,7 +54,7 @@ services:
       - DD_TRACE_NO_AUTOLOADER=true
     volumes:
       - ./callgrind-files:/var/www/callgrind-files
-      # - ../../:/dd-trace-php
+      - ../../:/dd-trace-php
```
labbati, line 57: Can we avoid mounting the extension dir by default? Then we can mount while working on something specific before the final measure. wdyt?

krakjoe:
I'm going to close this as the feedback loop is broken and the request looks stale.

If I'm wrong to do that, please open a new PR, and accept my apologies in advance :)

PR #825, opened 2020-04-15
Convert pid from long to string for internal spans

SammyK:
### Description

The Agent only supports string values in the `meta` array. This PR ensures that when a pid is added to a root internal span, it is added as a string instead of a long.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/serializer.c
+++ b/src/ext/serializer.c
@@ -365,7 +365,9 @@ static void _serialize_meta(zval *el, ddtrace_span_t *span TSRMLS_DC) {
         add_assoc_long(el, "error", 1);
     }
     if (span->parent_id == 0) {
-        add_assoc_long(meta, "system.pid", (uint)span->pid);
+        char pid[MAX_LENGTH_OF_LONG + 1];
+        snprintf(pid, sizeof(pid), "%ld", (long)span->pid);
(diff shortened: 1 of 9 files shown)


src/ext/serializer.c
```diff
@@ -16,6 +16,8 @@
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
+#define DD_TRACE_MAX_PID_LEN 7  // Theoretical maximum pid on 64-bit: 2^22 (4194304)
```
morrisonlevi, line 19: Where does the limit of `22` come from?

SammyK: This is from the [man pages](http://man7.org/linux/man-pages/man5/proc.5.html).

> On 64-bit systems, pid_max can be set to any value up to 2^22 (PID_MAX_LIMIT, approximately 4 million).

morrisonlevi: On Linux you can change the max PIDs at runtime -- I would be safe and use the max for the PHP long. There is a define for it in PHP already: `MAX_LENGTH_OF_LONG`, which is `20` for 64 bit.

SammyK: Thanks for the suggestion! Updated.

PR #829, opened 2020-04-16
Move some Configuration methods to functions written at the C level

morrisonlevi:
### Description
This moves the implementations of these `DDTrace\Configuration` methods to the C level:

- `appName` -> `ddtrace_config_app_name`
- `isIntegrationEnabled` -> `ddtrace_config_integration_enabled`
- `isEnabled` -> `ddtrace_config_trace_enabled`

It moves the callers of these methods to directly call the new functions. This is because `Configuration` shows up as hotspots in our recent performance testing; avoiding the `Configuration::get()` is important, and speeding up the call by moving it to C are both important. We can continue moving functions from Configuration, but this is a good stopping place as the latter two were among the top hotspots for integration loading.

It also consolidates the `shouldLoad($integration)` checks into `Integration::shouldLoad($integration)`, as this code was duplicated across the codebase.

**Importantly**, it also removes previously deprecated environment variable support for `ddtrace_app_name` and `DD_TRACE_APP_NAME`. We need to be sure to highlight this in release notes.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -135,6 +135,10 @@ ZEND_BEGIN_ARG_INFO_EX(arginfo_ddtrace_config_app_name, 0, 0, 0)
 ZEND_ARG_INFO(0, default_name)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(arginfo_ddtrace_config_integration_enabled, 0, 0, 1)
+ZEND_ARG_INFO(0, integration_name)
+ZEND_END_ARG_INFO()
+
 ZEND_BEGIN_ARG_INFO_EX(arginfo_ddtrace_config_trace_enabled, 0, 0, 0)
 ZEND_END_ARG_INFO()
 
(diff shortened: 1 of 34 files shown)


src/ext/ddtrace.c
```diff
@@ -795,13 +807,169 @@ static PHP_FUNCTION(dd_tracer_circuit_breaker_info) {
...
+            if (valid_begin && (found_end == end || *found_end == ',')) {
+                return false;
+            } else {
+                // todo: remove goto?
+                begin = found + 1;
+                goto loop;
+            }
+        }
+    }
+    return true;
```
SammyK, line 958: Do you think we could make [`find_exact_match()` public](https://github.com/DataDog/dd-trace-php/blob/master/src/ext/request_hooks.c#L16-L34) and use it here instead? Or combine them somehow since they are essentially doing the same thing? Or at least very similar? :)

src/ext/ddtrace.c
```diff
@@ -795,13 +807,169 @@ static PHP_FUNCTION(dd_tracer_circuit_breaker_info) {
...
+        default_str.len = ZSTR_LEN(default_zstr);
+    }
+#endif
+
+    ddtrace_string service_name = {
+        .ptr = "DD_SERVICE_NAME",
+        .len = sizeof("DD_SERVICE_NAME") - 1,
+    };
+    ddtrace_string app_name;
+    char *value = getenv(service_name.ptr);
```
SammyK, line 841: For all the `getenv()` calls, is there a reason you did not use the existing configuration API (`get_dd_trace_*()`)? Since it memoizes the `getenv()` calls, I'm wondering if it should be considered.

src/ext/ddtrace.c
morrisonlevi: We should definitely combine them. Thanks for bringing this to my attention!

src/ext/ddtrace.c
morrisonlevi: Actually, I avoided it _because_ it caches the calls. It breaks a lot of tests which just set the env and then expect it to work.

The main reason it does "caching" is to make it thread-safe so our background sender can read this info. Because of this, there are locks involved for reading the values so it's not a particularly "cheap" cache.

At least for now I thought it would be better to avoid code churn. If we get even more performance advantage then we can measure and do that later.

laurenty:
Hi @morrisonlevi, I didn't see DD_TRACE_APP_NAME being mentioned in the release notes for 0.43.0

morrisonlevi:
@laurenty You are correct. I am sorry that it probably means you hit an issue when upgrading. Originally this was slated for 0.44 and when we decided to pull it into 0.43 we overlooked this change. I have added it to the release notes on github, but until 0.43.1 is released the `package.xml` will not have it in the release notes.

Again, I am very sorry.

PR #831, opened 2020-04-17
Refactor curl handlers

morrisonlevi:
### Description

This provides two main changes:
- Removes the curl rinit hook that found the `le_curl` resource id. Instead we install a handler for `curl_init` that fetches this same value. This removes upfront cost but slightly increases it at usage site.
- Cleans up how the handlers are installed. Previously this was duplicated for every handler; now it is consolidated into a single array and function.

Targeting ddtrace v0.43.0, which will need to cherry-pick this change if merged.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/php7/handlers_curl.c
+++ b/src/ext/php7/handlers_curl.c
@@ -326,7 +326,7 @@ static ZEND_RESULT_CODE _dd_tracer_inject_helper(zval *headers, zval *format, dd
 ZEND_FUNCTION(ddtrace_curl_exec) {
     zval *ch;
 
-    if (_dd_load_curl_integration() && Z_TYPE(_dd_curl_httpheaders) == IS_LONG &&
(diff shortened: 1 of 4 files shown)


src/ext/php7/handlers_curl.c
```diff
@@ -436,16 +398,22 @@ ZEND_FUNCTION(ddtrace_curl_setopt_array) {
...
-        curl_setopt_array->internal_function.handler = ZEND_FN(ddtrace_curl_setopt_array);
+struct _dd_curl_handler {
+    const char *name;
+    size_t name_len;
+    void (**old_handler)(INTERNAL_FUNCTION_PARAMETERS);
+    void (*new_handler)(INTERNAL_FUNCTION_PARAMETERS);
+};
+typedef struct _dd_curl_handler _dd_curl_handler;
+
+static void _dd_install_handler(_dd_curl_handler handler) {
```
SammyK, line 409: Neat! 🔝

PR #833, opened 2020-04-20
Optimize Configuration

labbati:
### Description
Optimizes the configuration by removing some of the object oriented approach. Generally, changes things like:

    DDTrace\Configuration::get()->isFrobberEnabled()

into

    ddtrace_config_frobber_enabled()

The original configuration methods are there for backwards compatibility, but delegate to the function.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
The Symfony 3.4 tests are failing because those tests use a PSR-4 autoloader. I do not know if that version is intentionally doing this or not; the other Symfony versions do not do this.

I have removed the autoloader for now. We need to have a discussion as a team eventually about the composer autoload interaction.

Basically, we try to prefer what is in composer instead of what is installed via the extension. However, the C extension and PHP code are tightly coupled and their versions must exactly match. I think we should:

- Prefer the extensions's files always.
- Eventually remove the composer package altogether. We should ship everything we need in every kind of installation we support.

src/DDTrace/Configuration.php
```diff
@@ -67,7 +67,7 @@ public function isAnalyticsEnabled()
     public function isPrioritySamplingEnabled()
     {
         return $this->isDistributedTracingEnabled()
```
SammyK, line 69: The distributed tracing check exists in `ddtrace_config_priority_sampling_enabled()` so we should be able to remove the `$this->isDistributedTracingEnabled()` check.

tests/Unit/Sampling/ConfigurableSamplerTest.php
```diff
@@ -85,98 +94,43 @@ public function samplingRulesScenarios()
...
             ],
             'fallback to priority sampling when rule does not match' => [
-                [
-                    [
-                        'service' => 'something_else',
-                        'name' => '.*',
-                        'sample_rate' => 0.7,
-                    ],
-                ],
+                '[{"service":"something_else","sample_rate":0.7}]',
```
SammyK, line 101: 🔝 for testing from a higher level.

bridge/configuration.php
```diff
@@ -0,0 +1,318 @@
...
+{
+    if (false === $value || null === $value) {
+        return $default;
+    }
+
+    $value = strtolower($value);
+    if (is_numeric($value)) {
+        $floatValue = (float)$value;
+    } else {
+        $floatValue = (float)$default;
```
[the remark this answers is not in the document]
morrisonlevi: This code is was literally lifted from `EnvVariableRegistry` or maybe a different Configuration class. The real point is that these are simply copy-pasted into simple functions. Luca did this part, but I suspect it's done this way so that it minimize risk of an unintentional change.

PR #835, opened 2020-04-22
Whitelist the cgi-fcgi SAPI

SammyK:
### Description

This PR whitelists the `cgi-fcgi` SAPI.

For build branches, the web integration tests will be run via nginx and FastCGI for PHP 7.2, 7.3 and 7.4. These tests only run on build branches since they take longer to run than the standard web integration tests (which use the built-in web server from the CLI SAPI). On non-build branches, the web integration tests will continue to run as before for all supported PHP versions.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -401,6 +401,9 @@ jobs:
       resource_class:
         type: string
         default: medium
+      sapi:
+        type: string
+        default: cli-server
     resource_class: << parameters.resource_class >>
     executor:
       name: with_integrations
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -431,7 +434,14 @@ jobs:
           command: composer test-integration -- --log-junit test-results/php-integration/results.xml
       - run:
           name: Run << parameters.integration_testsuite >> integration test
-          command: DD_TRACE_AGENT_TIMEOUT=2000 composer << parameters.integration_testsuite >>
+          command: DD_TRACE_AGENT_TIMEOUT=2000 DD_TRACE_TEST_SAPI=<< parameters.sapi >> composer << parameters.integration_testsuite >>
```
labbati, line 437: Do we have memory of why this 2000 timeout was necessary? If not I would take the opportunity to remove it as I believe it is not necessary.

SammyK: IIRC, this is set to a high limit because the `request-replayer` uses the built-in web server from the CLI SAPI which is way slower than the real Agent and so tests were regularly failing intermittently. I'll bump this down to 1000, but I think it still needs to remain pretty high until we make `request-replayer` faster.

I'll also break out those other tests into their own job. :)

PR #840, opened 2020-04-23
Compile to one single file

labbati:
### Description

Compiling our files into one single file proved to be pretty efficient and depending on the app/script under test can provide over 15% in execution time for our pre-request bootstrap phase.

How this impact our workflow?
- a `composer compile` step is necessary to generate the final compiled script
- this is not necessary for tests, where instead of `composer test`, the newly created command `composer test-no-compile` can be used.
- on CI we still use the old `composer test` command so we test the real final file, as used by our users.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bridge/_files.php
```diff
@@ -0,0 +1,115 @@
+<?php
```
labbati, line 1: These files represents the same from the old required dependencies list, including the order.

tests/bootstrap.php
```diff
@@ -2,7 +2,6 @@
 
 error_reporting(E_ALL);
 
-require __DIR__ . '/../vendor/autoload.php';
```
labbati, line 5: we should not 'default' in tests to use psr4, we should use classes from the `_generated.php` file. On the other hand, we have a tests in the `auto-instrumentation` that tests usage of `_generated.php` and composer's `psr4`.

composer.json
```diff
@@ -47,6 +48,8 @@
     },
     "prefer-stable": true,
     "scripts": {
+        "compile": "classpreloader.php compile --config=./bridge/_files.php --output=./bridge/_generated.php --strip_comments=1",
```
morrisonlevi, line 51: I think one of these [`composer` events](https://getcomposer.org/doc/articles/scripts.md#event-names) can be hooked to call the compile script. We need to remove `autoload: {"psr-4": ...}`, as we aren't only a PSR-4 autoloader (because we have functions to load).

bridge/autoload.php
```diff
@@ -0,0 +1,12 @@
...
+
+if (getenv('DD_AUTOLOAD_NO_COMPILE') === 'true') {
+    // Development
+    $files = include __DIR__ . '/_files.php';
+    foreach ($files as $file) {
+        require_once $file;
+    }
+} else {
+    // Production
+    require_once __DIR__ . '/_generated.php';
```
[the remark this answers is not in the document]
labbati: I am retracting my 👍. I agree that this would be a bug, but why we should have a bug to cause a crash when we can have that same bug not to cause a crash? Instead what I am proposing is to change the [`require_once`](https://github.com/DataDog/dd-trace-php/pull/840/files#diff-5e02ea6e1ce4e3bd4dfb13879ad38660R7) on line 7 to `require` that would still cause a crash if called twice but in development. WDYT?

composer.json
labbati: Retracting my 👍 here as well. Or at least I want to expose my reasoning first.

> I think one of these composer events can be hooked to call the compile script.

I assume that you would like to have our users to use the generate script instead of the standard autoloading mechanism used by composer. In this case I have a soft disagreement.

1. If we have our users to generate the script and something goes wrong with the script generation be blocks development. We are starting only now using this library, theoretically we should have reproducible builds but I'd feel more comfortable in having users that requires manual instrumentation to still use the composer autoloader, while users not requiring manual instrumentation (hence using our code) to use our provided generated file that is TESTED in CI. This at the cost of some performance ONLY for the classes that they need and we do not use.
2. If in a few months we observe zero issues in CI with this new approach, I will be more than open to make generate the compiled file as part of a post-X hook.

> We need to remove autoload: {"psr-4": ...}, as we aren't only a PSR-4 autoloader (because we have functions to load).

We have not functions to load as part of this PR. This PR, from the user standpoint, it is still a fully compatibile psr4 autoload. Unless I am missing something.

bridge/autoload.php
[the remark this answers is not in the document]
labbati: @morrisonlevi pinging you on this comment as GH removed from the files view as the file has changed because of the push to change line 7.

composer.json
morrisonlevi: 1. The only way I can think of right now to avoid the situation where the user's script doesn't compile the generated file is to ship it in git, as composer uses the release tag and not the uploaded builds.
2. Fair.
3. This is true, but I thought we were going to use the compile-to-single-file PR to fix the remaining testing issues that occur in optimize/config. In that PR we are no longer a PSR-4 compatible autoloader and using a single file as our autoloader should fix the issues.

bridge/autoload.php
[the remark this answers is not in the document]
morrisonlevi: If it changes to require for "development", is there a chance that users hit it? Users will still use the PSR-4 autoloader, so I think they won't. If so then it's good enough for me.

PR #852, opened 2020-04-28
Refactor module blacklist

SammyK:
### Description

This moves the module blacklist check from RINIT to STARTUP. It also removes the ability to edit the blacklist from the INI settings as this functionality does not need to be exposed. Improvements were also made in gracefully handling specific Xdebug versions on PHP 7.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/ext/php7/blacklist.c
```diff
@@ -0,0 +1,65 @@
...
+        */
+        int compare = php_version_compare(module->version, "2.9.5");
+        if (compare == -1) {
+            /*
+            PHP 7.0 was only supported from Xdebug 2.4 through 2.7
+            @see: https://xdebug.org/docs/compat
+            */
+#if PHP_VERSION_ID < 70100
+            ddtrace_log_errf("Found incompatible Xdebug version %s; disabling conflicting functionality",
+                             module->version);
```
morrisonlevi, line 40: Based on my reading of this, tracer for PHP 7.0 will not disable itself if xdebug is version 2.9.5, but PHP 7.0 can't use xdebug 2.9.5? Will xdebug fail to load earlier than this and so we're streamlining the code here?

SammyK: Very good point! I'll move the PHP 7.0 check earlier.

PR #854, opened 2020-04-28
Cache that a function is not traced

morrisonlevi:
### Description

PHP has a spot in the `zend_op_array` structure that is open for `zend_extension`s to use: `reserved`. This PR will cache that a function is not traced in the reserved slot, which saves time on future calls.

Currently the patch only works on PHP 7 but can work on PHP 5 too. Note that on PHP 7.4 we are not using this same slot because it's possibly read-only because of the preloading feature (and maybe some other scenarios) but instead use a runtime cache slot -- the engine makes this available in PHP 7.4.

Most tests are passing, but there are a few I marked as skipped because I didn't understand what they were trying to test. The biggest change is that if a function is called and it is not traced, then  for the same request it won't be traced even if there is a `dd_trace_method` that comes later. We have two integrations that violate this requirement: WordPress and Laravel. I added a known integrations list for these integrations that add NULL callables on request init so that these will continue to work as written.

I still need to add more tests for opcache specifically, although ad-hoc tests locally show that it works.

---

Note that we could also cache this same information at compile-time, in which case it will persist in opcache, which has a big potential to reduce our overhead. However, it has some challenges:
  - If opcache is not enabled then this will be more expensive than what this PR already does. This is because it will trigger lookups for all call definitions, rather than the first usages of the call, and some code that is defined will not be used. However, if opcache is enabled then this is acceptable because subsequent requests that use the cached opcodes will skip the lookup entirely. We need to be able to detect this accurately in startup.
  - All tracing must be defined before the tracing target is defined. We have a lot of tests that do not conform to that requirement, and probably many integrations too.
  - At compile-time we do not have the full inheritance hierarchy of a class -- we only have the class name and possibly its parent, but usually no more than that.
  - We currently allow tracing something like `dd_trace_method("childclass", "method", ...)` where `method` only exists on the parent class. This means when a method is defined we cannot truly know if it won't be trace. An exception is if the class is marked as `final`, but this is not particularly common in PHP.

I'm not sure how many free-standing functions get defined in userland, but they are mostly free of challenges and we could mark that we do not trace them more reliably at compile-time than methods.

---

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
I used `dtrace` to measure the impact of this change. The execution times of `ddtrace_find_dispatch` and `_dd_should_trace_call` are tracked for 3 calls to a Laravel 5.8 app from our test suite using PHP 7.4.

Here are the numbers for the commit on master this was branched from.
```
Inclusive function on-CPU times (us),
   NAME                                                TOTAL
   ddtrace_find_dispatch                               36668
   _dd_should_trace_call                              104594
```

And here are the numebrs for this PR:
```
Inclusive function on-CPU times (us),
   NAME                                                TOTAL
   ddtrace_find_dispatch                               21836
   _dd_should_trace_call                               78829
```

For these given tests, this is a reduction of 8588 us (8.588 milliseconds) per request. This is very promising and I look forward to seeing benchmark numbers for end-to-end requests.

morrisonlevi:
I've reached out to Bob Weinand to get his opinion on if the whitespace diffs in the two phpdbg tests is a real issue or not. Aside from those two tests it looks complete and ready to go.

tests/ext/closure_accessing_outside_variables.phpt
```diff
@@ -18,7 +18,8 @@ function setup($variable){
     });
 }
 
-(new Test())->m();
+// Cannot call a function while it is not traced and later expect it to trace
```
pawelchcki, line 21: I think we could keep the method if we had some way to prime the system. Marking this method as: 
"of interest" which means it will not be neg-cached. Or rather the cache will store some other marker.

Marking could be done by `-d` directive or env variable, or function we could call at startup.

This way we'll be able to not modify or delete existing tests - too much.
WDYT ?

morrisonlevi: For this case, something like like `-d ddtrace.known_functions=Test::m`? We need something like this if we want to remove runtime tracing of internal functions anyway.

tests/opcache/basic.phpt
```diff
@@ -0,0 +1,34 @@
+--TEST--
+test caching that calls are not traced works with opcache's preload script
+--INI--
+opcache.enable_cli=1
+--FILE--
+<?php
+
+if (!extension_loaded('Zend OPcache')) die("opcache is required for this test\n");
```
SammyK, line 8: I'm guessing this isn't in a `SKIPIF` to make the test fail when OPcache isn't loaded? And that's why these tests are outside of the **tests/ext** directory? And we don't ship them with PECL for all of these reasons as well? :)

src/ext/php7/engine_hooks.c
```diff
@@ -77,17 +82,79 @@ static bool _dd_should_trace_call(zend_execute_data *call, zend_function *fbc, d
...
+        /* On PHP 7.4 the op_array reserved flag can only be set on compilation.
+         * After that, you must use ZEND_OP_ARRAY_EXTENSION.
+         * We don't use it at compile-time yet, so only check this on < 7.4.
+         */
+#if PHP_VERSION_ID < 70400
+        if (fbc->op_array.reserved[ddtrace_resource] == DDTRACE_NOT_TRACED) {
+            return false;
+        }
+#else
+        ddtrace_dispatch_t *cached_dispatch = DDTRACE_OP_ARRAY_EXTENSION(&fbc->op_array);
```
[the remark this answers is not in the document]
morrisonlevi: I think so -- I think they use the same op_array extension on PHP 7.4.

tests/opcache/basic.phpt
morrisonlevi: Correct.

tests/ext/closure_accessing_outside_variables.phpt
morrisonlevi: We can cross this in another PR, I think.

Issue #865, opened 2020-05-01
[BUG] Error: Call to undefined function () in /opt/datadog-php/dd-trace-sources/bridge/functions.php:138


ThomasVdBerge:
**Describe the bug**
When running this plugin on PHP 7.4.5 we often get the following error. Out setup runs on AWS Fargate:

```
PHP message: PHP Fatal error:  Uncaught Error: Call to undefined function () in /opt/datadog-php/dd-trace-sources/bridge/functions.php:138
thrown in /opt/datadog-php/dd-trace-sources/bridge/functions.php on line 138
#0 /app/vendor/composer/autoload_real.php(25): ComposerAutoloaderInit26ebf97ffdb7575dcfd32f97a9330f1b::dd_trace_callback(Array, true, true)
#1 /app/vendor/autoload.php(7): ComposerAutoloaderInit26ebf97ffdb7575dcfd32f97a9330f1b::getLoader()
#2 /app/public_html/index.php(2): require('/app/vendor/aut...')
```

This causes the docker container to crash and restart. 

We are not seeing this behavior on 7.4.4.

**PHP Info**
- Output of `php -v`

> PHP 7.4.5 (cli) (built: Apr 24 2020 17:44:59) ( NTS )
> Copyright (c) The PHP Group
> Zend Engine v3.4.0, Copyright (c) Zend Technologies
>     with Zend OPcache v7.4.5, Copyright (c), by Zend Technologies

- Output of `php -m`

> [PHP Modules]
> apcu
> bcmath
> bz2
> calendar
> Core
> ctype
> curl
> date
> ddtrace
> dom
> exif
> fileinfo
> filter
> ftp
> gd
> geoip
> gmp
> hash
> iconv
> intl
> json
> libxml
> mbstring
> memcached
> mysqlnd
> openssl
> pcntl
> pcre
> PDO
> pdo_mysql
> pdo_sqlite
> Phar
> posix
> readline
> Reflection
> session
> SimpleXML
> soap
> sodium
> SPL
> sqlite3
> standard
> sysvmsg
> sysvsem
> sysvshm
> tokenizer
> xml
> xmlreader
> xmlwriter
> Zend OPcache
> zip
> zlib
> 
> [Zend Modules]
> Zend OPcache

**OS Info**
- Output of `cat /etc/os-release | grep -E "(NAME)|(VERSION)"`

> NAME="Alpine Linux"
> VERSION_ID=3.11.6
[...]

PR #866, opened 2020-05-01
Sandbox non-tracing usages of dd_trace on internal functions within startup

morrisonlevi:
### Description
Aside from the direct benefits of sandboxing (which would have probably prevented issue #865 from being a stability issue), these 3 functions were the only ones I could find that target internal functions and are used even if sandboxing is enabled; this is a key prerequisite to removing runtime lookups of ddtrace_dispatch_t's for internal functions. This way we can replace the internal function handler at startup and only internal functions that might get traced pay the cost of checking whether they should be traced.

But that's the future. Today is removing uses of the legacy API and sandboxing more code.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ Nearly all tests will exercise this code path.

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
Closing. This approach will generate a new span and drop it. However, while that span is open it runs the dd_init.php file which will make a new root span, so the root span ends up having a trace_id that matches the dropped span... so it doesn't work right.

PR #870, opened 2020-05-03
Add ddtrace_config_distributed_tracing_enabled

morrisonlevi:
### Description

Create `ddtrace_config_distributed_tracing_enabled()` for both PHP and the C extension.
- Migrate uses of `DDTrace\Configuration->isDistributedTracingEnabled()` to `ddtrace_config_distributed_tracing_enabled()` for both PHP and C extension.
- Migrate `DDTrace\Configuration->isDistributedTracingEnabled()` itself to use `ddtrace_config_distributed_tracing_enabled()` for backwards compatibility.

Also cleanup some recently added configuration related code in ddtrace.c by moving it into configuration.{c,h}.

This work has been split off from #833 and improved.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ I judged existing tests to be adequate.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/configuration.c
+++ b/src/ext/configuration.c
@@ -2,7 +2,10 @@
 
 #include <stdlib.h>
 
+#include "ddtrace_string.h"
 #include "env_config.h"
+
+extern inline ddtrace_string ddtrace_string_getenv(char* str, size_t len TSRMLS_DC);
 
 struct ddtrace_memoized_configuration_t ddtrace_memoized_configuration = {
 #define CHAR(...) NULL, FALSE,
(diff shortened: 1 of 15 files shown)


src/ext/php7/handlers_curl.c
```diff
@@ -362,11 +343,10 @@ ZEND_FUNCTION(ddtrace_curl_setopt) {
     ddtrace_error_handling eh;
     ddtrace_backup_error_handling(&eh, EH_THROW);
 
-    if (Z_TYPE_P(return_value) == IS_TRUE && Z_TYPE(_dd_curl_httpheaders) == IS_LONG &&
-        Z_LVAL(_dd_curl_httpheaders) == option) {
-        if (_dd_Configuration_isDistributedTracingEnabled()) {
-            _dd_ArrayKVStore_putForResource(zid, &_dd_format_curl_http_headers, zvalue);
-        }
+    bool call_succeeded = Z_TYPE_P(return_value) == IS_TRUE;
+    bool option_matches = Z_TYPE(_dd_curl_httpheaders) == IS_LONG & Z_LVAL(_dd_curl_httpheaders) == option;
```
morrisonlevi, line 347:
```suggestion
    bool option_matches = Z_TYPE(_dd_curl_httpheaders) == IS_LONG && Z_LVAL(_dd_curl_httpheaders) == option;
```

src/ext/configuration.c
```diff
@@ -131,3 +134,80 @@ DD_CONFIGURATION
...
+bool ddtrace_config_distributed_tracing_enabled(TSRMLS_D) {
+    ddtrace_string distributed_tracing_enabled =
+        ddtrace_string_cstring_ctor(ddtrace_getenv(ZEND_STRL("DD_DISTRIBUTED_TRACING") TSRMLS_CC));
+    if (distributed_tracing_enabled.len) {
+        /* We need to lowercase the str for ddtrace_config_bool.
+         * We know it's already been duplicated by ddtrace_getenv, so we can
+         * lower it in-place.
+         */
+        zend_str_tolower(distributed_tracing_enabled.ptr, distributed_tracing_enabled.len);
+        bool result = ddtrace_config_bool(distributed_tracing_enabled, true);
```
pawelchcki, line 173: Wonder if we could teach `ddtrace_config_bool` to do the lowercasing.
We could have two variants, one that will lowercase - inplace. Other taking e.g. const ptr.

src/ext/configuration.c
```diff
@@ -131,3 +134,80 @@ DD_CONFIGURATION
...
+    };
+    if (ddtrace_string_equals(subject, str_0) || ddtrace_string_equals(subject, str_false)) {
+        return false;
+    }
+    return default_value;
+}
+
+bool ddtrace_config_distributed_tracing_enabled(TSRMLS_D) {
+    ddtrace_string distributed_tracing_enabled =
+        ddtrace_string_cstring_ctor(ddtrace_getenv(ZEND_STRL("DD_DISTRIBUTED_TRACING") TSRMLS_CC));
```
pawelchcki, line 166: Can `ddtrace_string_getenv` be used here ?

src/ext/configuration.c
```diff
@@ -131,3 +134,80 @@ DD_CONFIGURATION
...
+        return result;
+    }
+    if (distributed_tracing_enabled.ptr) {
+        efree(distributed_tracing_enabled.ptr);
+    }
+    return true;
+}
+
+// note: only call this if ddtrace_config_trace_enabled() returns true
+bool ddtrace_config_integration_enabled(ddtrace_string integration TSRMLS_DC) {
```
pawelchcki, line 184: overall, these functions are a bit verbose. Some code reuse could help here.

Also each invocation is creating a temporary string and either do `getenv` or `sapi_getenv`. <- but this might be something to optimize in the future - by caching them e.g. in request context.

morrisonlevi: I'll look at them a bit more and see if I can pare them down. I think it can be done, just need to see it with fresh eyes.

src/ext/configuration.c
morrisonlevi: Probably -- I added `ddtrace_string_getenv` after I noticed the repetition and probably just missed this one.

morrisonlevi:
Thanks for your review, Pawel. I've pushed up a new commit that addresses the duplication issues and also reduced code size a bit.

As for moving more and more configuration functions into the C extension, I'm not sure what to do about that concern. I think eventually composer will go away because all methods of installing the tracer will ship the PHP components with it. The only piece that's missing today is pecl. At least for the foreseeable future the PHP code and C code have to remain in sync anyway.

PR #869, opened 2020-05-02
Use normalized URL as the resource name for curl

akdh:
### Description

Use the normalized URL rather than the sanitized URL as the resource name for curl requests. This allows for URLs like users/123 and users/456 to be grouped togerher providing more useful summary statistics in datadog.

See: https://github.com/DataDog/dd-trace-php/issues/864

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.
- [ ] Changelog has been added to the appropriate release draft. For community contributors the reviewer is in charge of this task.

PR #876, opened 2020-05-11
Bump version to 0.45.0

morrisonlevi:
### Description

If you are using dd_trace, dd_trace_function, or dd_trace_method then you need to call these functions before the first invocation of the target e.g. dd_trace('foo', ...) should be done before foo is called for the first time.

### Changed

- Compile to one single file #840, #874
- Simplify configuration #833, #870
- Refactor module blacklist #852
- Cache that a function is not traced #854, #863
- Simplify spl_autoload_register instrumentation #867
- Use normalized URL as the resource name for curl #869 (thanks, @akdh!)
- Sandbox header and http_response_code #875

### Fixed

- Fix variadic args with empty function signature #872
- Bump phpstan to 0.12 and fix issues #841

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -22,7 +22,7 @@
     <license uri="https://github.com/DataDog/dd-trace-php/blob/master/LICENSE">BSD 3-Clause</license>
     <notes>
 ## Important behavior changes in this release
-As of this release you can no longer trace a function or method if it has already been called. All instrumentation should happen before the target calls are run (and ideally before it is even defined, which can enable more aggressive optimizations in the future).
+If you are using `dd_trace`, `dd_trace_function`, or `dd_trace_method` then you need to call these functions before the first invocation of the target e.g. `dd_trace('foo', ...)` should be done before `foo` is called for the first time.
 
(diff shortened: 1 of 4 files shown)


package.xml
```diff
@@ -10,17 +10,35 @@
...
     </version>
     <stability>
         <release>stable</release>
         <api>stable</api>
     </stability>
     <license uri="https://github.com/DataDog/dd-trace-php/blob/master/LICENSE">BSD 3-Clause</license>
-    <notes>${notes}</notes>
+    <notes>
+## Important behavior changes in this release
+As of this release you can no longer trace a function or method if it has already been called. All instrumentation should happen before the target calls are run (and ideally before it is even defined, which can enable more aggressive optimizations in the future).
```
morrisonlevi, line 25:
```suggestion
If you are using `dd_trace`, `dd_trace_function`, or `dd_trace_method` then you need to call these functions before the first invocation of the target e.g. `dd_trace('foo', ...)` should be done before `foo` is called for the first time.
```

PR #637, opened 2019-10-31
Fix notice and refactor root span init

SammyK:
### Description

This PR fixes a notice that can be raised when running the request init hook outside of the sandbox on non-web SAPI's that don't set a `$_SERVER['REQUEST_METHOD']` or `$_SERVER['REQUEST_URI']`. I also tried to refactor the root-span initialization a bit, but it certainly could do with some more TLC.

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.
- ~~[ ] Tests added for this feature/bug.~~

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.
- [ ] Changelog has been added to the appropriate release draft. For community contributors the reviewer is in charge of this task.

PR #877, opened 2020-05-13
Avoid usage of spl_autoload_register while still allowing noop manual instrumentation without ext installed

labbati:
### Description

This PR implements a huge change to our loading and bootstrapping mechanism and to the way users can use our public userland API even when the extension is not installed in their environment (e.g. Windows or MAC).

Details:
- Public API has been moved to `src/api` and this is the package registered by the psr4 autoloader.
- If the extension is not installed this results mostly in a noop API.
- When the extension is installed, we register all the classes that the public api requires to not be noop
- Most of the changes are caused by avoiding leaking the `Integration` class over to the public api folder.
- To fix an issue that came up in the language tests, this converts PHP 5.6 to use `zend_execute_internal`.
    - To make this easier, some of 5.4's sources were split off from `php5` into `php5_4`.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bridge/dd_wrap_autoloader.php
```diff
@@ -1,31 +0,0 @@
-<?php
-
```
morrisonlevi, line 2: I imagine that at some point we've told people to set this ini manually, such as people installing from `pecl`. To help them upgrade I think we should emit a diagnostic or something and then include `dd_init.php`. What do you think?

phpstan.ddtrace.neon
```diff
@@ -0,0 +1,14 @@
+parameters:
+	ignoreErrors:
+		-
+			message: '#Call to an undefined method object::isHostRoot\(\)#'
+			path: src/DDTrace/Encoders/SpanEncoder.php
```
morrisonlevi, line 5: Can you defend this one a bit? Looking briefly at the file it seems this is legitimate so I'm curious what changed such that it's reported now and needs suppressed.

src/DDTrace/Integrations/Integration.php
```diff
@@ -33,6 +33,14 @@ public function __construct()
...
     }
 
+    public function addTraceAnalyticsIfEnabledLegacy(Span $span)
+    {
+        if (!$this->configuration->isTraceAnalyticsEnabled()) {
+            return;
+        }
+        $span->setMetric(Tag::ANALYTICS_KEY, $this->configuration->getTraceAnalyticsSampleRate());
+    }
+
```
morrisonlevi, line 43: Is this related to these changes or just something you grouped in here while passing by?

src/DDTrace/Span.php
```diff
@@ -392,26 +391,6 @@ public function getAllBaggageItems()
...
-        return $this;
-    }
-
-    /**
-     * @return null|Integration
-     */
-    public function getIntegration()
-    {
-        return $this->integration;
-    }
```
morrisonlevi, line 413: Are these changes related to removing spl_autoload_register and changing composer or just something you did while passing by?

labbati: they are related to it.

1. In order to remove `spl_autoload_register` we need to deliver the 'public'/noop api via composer.
2. As part of it interfaces in `DDTrace\Contracts` have to be moved to the public api
3. Span and SpanContext interface where depending on the concrete `Integration` class, which has a lot of things and we are about to remove with the work on simplifying the integrations.
4. I did not want to 'leak' that concrete class into the public API.

src/DDTrace/Integrations/Integration.php
labbati: This [answer](https://github.com/DataDog/dd-trace-php/pull/877/files#r424546314) applies to this comment too.

phpstan.ddtrace.neon
labbati: This is a very good point. Looking into it, as I had followed the wrong path before while deciding to accept the message

bridge/dd_wrap_autoloader.php
SammyK: +1 to @morrisonlevi's comment. There are also many people manually including this file from their own bootstrap files. I would change this file to `include __DIR__ . '/dd_init.php';` and leave a comment mentioning that this file was deprecated in 0.46.0.

labbati: love it

phpstan.ddtrace.neon
labbati: This was a leftover, this error is not there anymore

src/DDTrace/Tracer.php
```diff
@@ -438,12 +416,12 @@ private function addUrlAsResourceNameToRootSpan()
             return;
         }
         // Normalized URL as the resource name
-        $normalizer = new Urls(explode(',', getenv('DD_TRACE_RESOURCE_URI_MAPPING')));
-        $span->setTag(
-            Tag::RESOURCE_NAME,
-            $_SERVER['REQUEST_METHOD'] . ' ' . $normalizer->normalize($_SERVER['REQUEST_URI']),
-            true
-        );
+        $resourceName = $_SERVER['REQUEST_METHOD'];
```
SammyK, line 419: Could we add an `isset()` check here to avoid something like #637?

src/api/Configuration.php
```diff
@@ -26,7 +26,7 @@ class Configuration extends AbstractConfiguration
      */
     public function isEnabled()
     {
-        return \ddtrace_config_trace_enabled();
+        return $this->boolValue('trace.enabled', true);
```
SammyK, line 29: Did we mean to revert these changes?

labbati: yep, we don't want to use new functions at all in `/api`

src/DDTrace/Tracer.php
labbati: good idea

PR #881, opened 2020-05-15
Run sandboxed and legacy integration tests in isolated processes

labbati:
### Description

Before (e.g. for `curl` integration):
- we were running tests as `composer test -- tests/Integrations/Curl`
- this means that both sandboxed and legacy integrations where loaded (see snipped below)
- because legacy integration was executed after, even if we removed the legacy integration of `curl_exec` all the tests would still pass, because the sandboxed integration did take care already of instrumenting `curl_exec`.

After:
- legacy and sandboxed library integrations tests are not run into their own process
```
composer test-no-compile -- tests/Integrations/Curl
> DD_AUTOLOAD_NO_COMPILE=true phpunit --colors=always --configuration=phpunit.xml 'tests/Integrations/Curl'
executed dd_init.php
Loading curl SANDBOXED integration
PHPUnit 4.8.36 by Sebastian Bergmann and contributors.

Runtime:	PHP 7.3.17
Configuration:	/home/circleci/app/phpunit.xml

Loading curl LEGACY integration
.................Loading curl SANDBOXED integration
.................

Time: 1.69 seconds, Memory: 8.00MB

OK (34 tests, 452 assertions)
```

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
When running the non-sandbox tests we should add `DD_TRACE_SANDBOX_ENABLED=0` to the env before calling, I think. Otherwise it will start-up with sandboxing with the noop-api changes, right?

labbati:
> When running the non-sandbox tests we should add `DD_TRACE_SANDBOX_ENABLED=0` to the env before calling, I think. Otherwise it will start-up with sandboxing with the noop-api changes, right?

with the noop-api yes. Still, let me do it here as it is more robust in any case.

PR #895, opened 2020-05-21
Install internal function handlers at startup (PHP 7)

morrisonlevi:
### Description

This moves our internal function checks to startup instead of at runtime on PHP 7. When complete, this will have much less overhead than what we did before. This does mean that we have to know all internal functions that will ever be traced when the process starts up. To allow users (and ourselves in our tests) to instrument internal functions other than the ones in our integrations, there is a new environment variable `DD_TRACE_TRACED_INTERNAL_FUNCTIONS` which takes a csv of functions or methods that will be traced e.g. `DD_TRACE_TRACED_INTERNAL_FUNCTIONS=array_sum,mt_rand,DateTime::add`.

I use the reserved slot of the internal function to store the previous handler so I can use the same PHP_FUNCTION for all of the handlers, and also handle user-supplied functions in `DD_TRACE_TRACED_INTERNAL_FUNCTIONS`.

This PR tries to not install certain handlers when possible; if sandboxing is enabled then it won't install icall and zend_execute_internal hooks. This branch has been reworked a few times before the current state, so need to do a second-pass to make sure my sandboxing_enabled checks make sense.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -52,8 +52,6 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
                   zend_ddtrace_globals, ddtrace_globals)
 STD_PHP_INI_BOOLEAN("ddtrace.strict_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, strict_mode, zend_ddtrace_globals,
                     ddtrace_globals)
-STD_PHP_INI_ENTRY("ddtrace.traced_internal_functions", "", PHP_INI_SYSTEM, OnUpdateString, traced_internal_functions,
-                  zend_ddtrace_globals, ddtrace_globals)
 PHP_INI_END()
 
 static int ddtrace_startup(struct _zend_extension *extension) {
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -31,7 +31,7 @@
 #include "dispatch.h"
 #include "dogstatsd_client.h"
 #include "engine_hooks.h"
-#include "handlers_curl.h"
+#include "handlers_internal.h"
 #include "logging.h"
 #include "memory_limit.h"
 #include "random.h"
@@ -52,6 +52,8 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
                   zend_ddtrace_globals, ddtrace_globals)
 STD_PHP_INI_BOOLEAN("ddtrace.strict_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, strict_mode, zend_ddtrace_globals,
                     ddtrace_globals)
+STD_PHP_INI_ENTRY("ddtrace.traced_internal_functions", "", PHP_INI_SYSTEM, OnUpdateString, traced_internal_functions,
+                  zend_ddtrace_globals, ddtrace_globals)
 PHP_INI_END()
 
 static int ddtrace_startup(struct _zend_extension *extension) {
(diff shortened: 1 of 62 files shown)


src/ext/ddtrace.c
```diff
@@ -52,6 +52,8 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
                   zend_ddtrace_globals, ddtrace_globals)
 STD_PHP_INI_BOOLEAN("ddtrace.strict_mode", "0", PHP_INI_SYSTEM, OnUpdateBool, strict_mode, zend_ddtrace_globals,
                     ddtrace_globals)
+STD_PHP_INI_ENTRY("ddtrace.traced_internal_functions", "", PHP_INI_SYSTEM, OnUpdateString, traced_internal_functions,
```
SammyK, line 55: Any chance we can move this to an env var for now until we're able to get everything migrated over to INI settings at once? :) For the record, INI is totally where this should go, but until we get the post-install hook refactored, we're pretty much confined to env vars at the moment.

src/ext/ddtrace.c
```diff
@@ -554,13 +558,6 @@ static PHP_FUNCTION(dd_trace) {
         DD_PRINTF("Class name: %s", Z_STRVAL_P(class_name));
     }
     DD_PRINTF("Function name: %s", Z_STRVAL_P(function));
-    if (ddtrace_blacklisted_disable_legacy && !get_dd_trace_ignore_legacy_blacklist()) {
-        ddtrace_log_debugf(
-            "Cannot instrument '%s()' with dd_trace(). This functionality is disabled due to a potentially conflicting "
-            "module. To re-enable dd_trace(), please set the environment variable: DD_TRACE_IGNORE_LEGACY_BLACKLIST=1",
-            Z_STRVAL_P(function));
-        RETURN_BOOL(0);
-    }
```
SammyK, line 563: Did this get deleted by accident?

src/ext/php7/handlers_internal.c
```diff
@@ -0,0 +1,132 @@
...
+                    .ptr = segment.ptr,
+                    .len = colon,
+                };
+                ddtrace_string method = {
+                    .ptr = segment.ptr + colon + 2,
+                    .len = delimiter - colon - 2,
+                };
+                ddtrace_replace_internal_methods(Class, 1, &method);
+            } else {
+                // todo: should we warn?
```
SammyK, line 69: I'm +1 for a debug message here. But not necessary for merge.

src/ext/php7/handlers_mysqli.c
```diff
@@ -0,0 +1,46 @@
...
+};
+
+static ddtrace_string ddtrace_mysqli_stmt_methods[] = {
+    DDTRACE_STRING_LITERAL("execute"),
+    DDTRACE_STRING_LITERAL("get_result"),
+};
+// clang-format on
+
+void ddtrace_mysqli_handlers_startup(void) {
+    // todo: how to handle ddtrace_resource = -1?
```
SammyK, line 29: Is this something that needs to be addressed before merge you think?

src/ext/ddtrace.c
morrisonlevi: I guess so -- must have happened in the rebase for some reason.

morrisonlevi: Restored.

src/ext/php7/handlers_mysqli.c
morrisonlevi: I think so; thanks for bringing it up. Prior to this change we lose an optimization. With this change we lose core functionality if we can't get a resource. If this happens we need to:

- Do it safely and make sure we don't access `reserved[-1]` which would sigsegv.
- Emit a message saying that we aren't tracing internal functions and to open a report. Not sure what info we can log; I don't think you can reverse-map the numbers to extensions that asked for them.

src/ext/ddtrace.c
morrisonlevi: Done.

PR #896, opened 2020-05-21
fix opcache check in dd-doctor

labbati:
### Description

*Problems this PR fix*
- We said `Opcache enabled` while the fact that the function exists only tells if opcache is installed, while the following props will tell if opcache is enabled.
- We did not `return;` if the function did not exist, causing an error
- If opcache was enabled, we did say `Opcache enabled [NO]` + `Opcache enabled [YES]`

### Readiness checklist
~- [ ] Tests added for this feature/bug.~

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/dd-doctor.php
+++ b/src/dd-doctor.php
@@ -120,9 +120,10 @@ function env($key)
 function check_opcache()
 {
     if (!function_exists('opcache_get_configuration')) {
-        render('Opcache enabled:', 'NO');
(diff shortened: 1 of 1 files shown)


labbati:
The "PHP 56 language tests" that fail is a glitch in UI, as they are green 
image https://app.circleci.com/pipelines/github/DataDog/dd-trace-php?branch=labbati%2Fopcache-dd-doctor

Issue #902, opened 2020-05-24
[BUG] Segfaults with opcache.file_cache reused between FPM and CLI


Majkl578:
Hello,

**Describe the bug**
We're currently seeing segmentation faults with OPcache file cache being reused between FPM and CLI. This only happens with `ddtrace` extension loaded and under specific conditions. So far I've been able to trigger the bug with preoading enabled, then hitting a PHP script through HTTP, then running `php -m` immediately segfaults.

I was able to obtain a backtrace from a dumped core, although it doesn't seem to be particularly verbose:
```
(gdb) bt full
#0  0x0000000000000000 in ?? ()
No symbol table info available.
#1  0x00005592b8130aa9 in ZEND_USER_OPCODE_SPEC_HANDLER () at ./Zend/zend_vm_execute.h:2637
        ret = <optimized out>
#2  0x00005592b813262d in execute_ex (ex=0x7f1541e12020) at ./Zend/zend_vm_execute.h:53902
        orig_opline = <optimized out>
        orig_execute_data = <optimized out>
#3  0x00005592b8139981 in zend_execute (op_array=0x7f1541e66000, return_value=<optimized out>) at ./Zend/zend_vm_execute.h:57922
        execute_data = 0x7f1541e12020
        object_or_called_scope = <optimized out>
        call_info = <optimized out>
#4  0x00007f153e7b3a85 in dd_execute_php_file () from /opt/datadog-php/extensions/ddtrace-20190902.so
No symbol table info available.
#5  0x00007f153e7a4540 in zm_activate_ddtrace () from /opt/datadog-php/extensions/ddtrace-20190902.so
No symbol table info available.
#6  0x00005592b80b8350 in zend_activate_modules () at ./Zend/zend_API.c:2605
        module = 0x5592ba2bf890
        p = 0x5592ba512b00
#7  0x00005592b80513d6 in php_request_startup () at ./main/main.c:1834
        __orig_bailout = 0x7ffe88a47a50
        __bailout = {{__jmpbuf = {94088643238464, 8335647263630806343, 140731190900884, 2, 94088676743600, 140731190905828, 8335647263828987207, 2843541885642460487}, __mask_was_saved = 0, __saved_mask = {__val = {1, 0, 1, 140731190901520, 94088641163538, 
                140731190900736, 94088643238464, 140731190901512, 140731190900736, 0, 94085553586178, 2, 140731190900888, 0, 94088643238464, 140731190900888}}}}
        retval = 0
[...]

PR #907, opened 2020-05-28
Move the request init hook to auto_prepend_file

SammyK:
### Description

This PR addresses #902.

The request init hook runs in RINIT which is before neighboring extensions expect PHP to be executed. This can cause issues when a neighboring extension makes changes to the loaded userland definitions/`zend_op_array`s in RINIT after the request init hook has loaded. Preloading in particular has a [very late hook](https://heap.space/xref/PHP-7.4/ext/opcache/ZendAccelerator.c?r=b6492b44#4314) that conflicts with the request init hook being loaded so early.

This PR moves the request init hook execution from RINIT to later in the request (where `auto_prepend_file` executes).

Any existing `auto_prepend_file` will still execute, but if a request init hook does not call `ddtrace_init()`, the original `auto_prepend_file` will not run. This would happen when an older version of `dd_wrap_autoloader.php` is used for the request init hook. This scenario is unlikely to occur, but will occur under the following conditions:

0. The `auto_prepend_file` INI setting is used
1. The `DataDog/dd-trace-php` Composer package is installed
2. The version installed with Composer is < `0.46.0`
3. The version of the extension installed is >= `0.46.0`

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


dockerfiles/ci/xfail_tests/5.4.list
```diff
@@ -257,3 +257,5 @@ ext/curl/tests/curl_setopt_error.phpt
 ext/pdo_sqlite/tests/bug50728.phpt
 ext/pdo/tests/pdo_023.phpt
 ext/pdo_sqlite/tests/pdo_013.phpt
+tests/lang/bug23584.phpt
+Zend/tests/bug48930.phpt
```
SammyK, line 261: On PHP 5.4 the shebang throws off the line numbers when `auto_prepend_file` is used. The [bug was fixed in PHP 5.5](https://heap.space/xref/PHP-5.5/main/main.c?r=c117548e#2511-2515).

PR #911, opened 2020-06-02
Sandbox cURL (PHP 5)

SammyK:
### Description

This is a follow up PR to #814. This sandboxes the cURL integration on PHP 5.6 (PHP 5.4 is still limited to the legacy API so it does not apply.)

As in #814, this PR also provides custom internal function handlers for `curl_exec`, `curl_copy_handle`, `curl_close`, `curl_setopt`, and `curl_setopt_array` in order to inject distributed tracing headers. This PR differs slightly from #814 by injecting the distributed tracing headers with a single userland call to `DDTrace\Bridge\curl_inject_distributed_headers()` right before the `curl_exec()` call. This approach can be used for PHP 7 to reduce the dependency on the userland API.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -94,10 +94,8 @@ public function __construct(array $integrations)
                 '\DDTrace\Integrations\CakePHP\CakePHPSandboxedIntegration';
             $this->integrations[CodeIgniterSandboxedIntegration::NAME] =
                 '\DDTrace\Integrations\CodeIgniter\V2\CodeIgniterSandboxedIntegration';
-            if (\PHP_MAJOR_VERSION > 5) {
-                $this->integrations[CurlSandboxedIntegration::NAME] =
-                    '\DDTrace\Integrations\Curl\CurlSandboxedIntegration';
-            }
```
morrisonlevi, line 100: Should be > 5.4 now, right?

SammyK: `ddtrace_config_sandbox_enabled()` always returns `false` for PHP 5.4.

PR #928, opened 2020-06-19
Apply new uri to resource normalization rules

labbati:
### Description

This PR provide a new method to customize uri to resource mapping.

_Problem_
Http requests that corresponds to URI `GET http://example.com/user/123/permissions` has a resource name in the format of `GET /user/?/permissions`. Identifying the variable part `123` and converting it into `?` is what we call resource normalization. With resource normalization in place we make sure that requests to `GET http://example.com/user/123/permissions` and `GET http://example.com/user/456/permissions` are grouped under the very same resource name `GET http://example.com/user/?/permissions` in our UI.

We adopt 2 strategies to normalize `GET /user/123/permissions` into `GET /user/?/permissions`:
- known patterns for path fragments (e.g. digits, uuids)
- routing info we extract from we frameworks we have deep instrumentation for.

There are 2 use cases that the above strategies do not fully cover:
- custom web frameworks with non common resource identifiers (anything other then digits and uudis)
- web frameworks for which we haven't implemented routing parsing, yet.

Assume you have the following set of uris

| You have | You want |
|---|---|
| `/always/using/<7 CHARS HEX>/as/id/everywhere` | `/always/using/?/as/id/everywhere` |
| `/cities/london/rivers` | `/cities/?/rivers` |
| `/articles/slug-of-title` | `/articles/?` |

This PR offers the possibility to cover the remaining cases above via new settings.

`DD_TRACE_RESOURCE_URI_FRAGMENT_REGEX`: lets you provide a list of regex that are applied to every FULL path fragment. By FULL path fragment we mean what is inside to `/` chars in a uri path. E.g. in `/cities/london/rivers` we have three path fragments: `cities`, `london`, `rivers`. A list of regexes can be provided separated by comma `,`. With env you would address the first use case. `DD_TRACE_RESOURCE_URI_FRAGMENT_REGEX=^[a-f0-9]{7}$` will find and replace ALL path fragments that are a 7 chars HEX word with `?`.
This setting applies to both incoming and outgoing requests.

`DD_TRACE_RESOURCE_URI_MAPPING_INCOMING`: lets you cover the cases when you cannot provide a regex (e.g. to normalize `london` and `rome`) but you know that a given identifier appears always after a known word, e.g. in this case `cities`. In this case you would normalize the latter two examples in the table above via `DD_TRACE_RESOURCE_URI_MAPPING_INCOMING=cities/*,articles/*`. Note that in this case the provided values are NOT regexes, but they recognize the `*` replacement char.

`DD_TRACE_RESOURCE_URI_MAPPING_OUTGOING`: Same as `DD_TRACE_RESOURCE_URI_MAPPING_INCOMING` but for outgoing requests.

**NOTE: THIS PR DEPRECATES SETTING `DD_TRACE_RESOURCE_URI_MAPPING`*

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/bridge/configuration.php
+++ b/bridge/configuration.php
@@ -114,7 +114,7 @@ function _ddtrace_config_indexed_array($value, $default)
 
     return array_map(
         function ($entry) {
-            return strtolower(trim($entry));
+            return trim($entry);
         },
         explode(',', $value)
     );
--- a/bridge/configuration.php
+++ b/bridge/configuration.php
@@ -208,6 +208,30 @@ function ddtrace_config_url_resource_name_enabled()
 }
 
 /**
+ * @return string[]
+ */
+function ddtrace_config_path_fragment_regex()
+{
+    return \_ddtrace_config_indexed_array(\getenv('DD_TRACE_RESOURCE_URI_FRAGMENT_REGEX'), []);
+}
+
+/**
+ * @return string[]
+ */
+function ddtrace_config_path_mapping_incoming()
+{
+    return \_ddtrace_config_indexed_array(\getenv('DD_TRACE_RESOURCE_URI_MAPPING_INCOMING'), []);
+}
+
+/**
+ * @return string[]
+ */
+function ddtrace_config_path_mapping_outgoing()
+{
+    return \_ddtrace_config_indexed_array(\getenv('DD_TRACE_RESOURCE_URI_MAPPING_OUTGOING'), []);
+}
+
+/**
  * Set URL hostname as service name
  *
  * @return bool
--- a/src/private/functions.php
+++ b/src/private/functions.php
@@ -1,17 +1,23 @@
 <?php
 
-namespace DDTrace;
+namespace DDTrace\Private_;
 
 use DDTrace\Http\Urls;
 
-function _util_normalize_outgoing_path($uriPath)
+const DEFAULT_URI_PART_NORMALIZE_REGEXES = [
+    '/^\d+$/',
+    '/^[0-9a-f]{8}-?[0-9a-f]{4}-?[1-5][0-9a-f]{3}-?[89ab][0-9a-f]{3}-?[0-9a-f]{12}$/',
+    '/^[0-9a-f]{8,128}$/',
+];
+
+function util_uri_normalize_outgoing_path($uriPath)
 {
-    return _util_uri_apply_rules($uriPath, /* incoming */ false);
+    return util_uri_apply_rules($uriPath, /* incoming */ false);
 }
 
-function _util_normalize_incoming_path($uriPath)
+function util_uri_normalize_incoming_path($uriPath)
 {
-    return _util_uri_apply_rules($uriPath, /* incoming */ true);
+    return util_uri_apply_rules($uriPath, /* incoming */ true);
 }
 
 /**
@@ -19,7 +25,7 @@ function _util_normalize_incoming_path($uriPath)
  * @param boolean $incoming
  * @return string
  */
-function _util_uri_apply_rules($uriPath, $incoming)
+function util_uri_apply_rules($uriPath, $incoming)
 {
     if ($uriPath === '/' || $uriPath === '' || null === $uriPath) {
         return '/';
@@ -86,9 +92,4 @@ function _util_uri_apply_rules($uriPath, $incoming)
     }
 
     return implode('/', $fragments);
-}
-
-function _util_path_matching_pattern_to_regex($mappingFragments)
-{
-    $regex = '/';
 }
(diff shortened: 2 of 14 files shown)


bridge/configuration.php
```diff
@@ -114,7 +114,7 @@ function _ddtrace_config_indexed_array($value, $default)
 
     return array_map(
         function ($entry) {
-            return strtolower(trim($entry));
+            return trim($entry);
```
labbati, line 117: This is generic getter and should not make assumption about case sensitive vs case insensitive. Also, this function was not used by other settings so it is good to be changed as it is only used by the newly added functions.

tests/Unit/private/UriTest.php
```diff
@@ -0,0 +1,464 @@
...
+            '/int/?/path/?/?/then/something/?',
+            \DDtrace\Private_\util_uri_normalize_incoming_path('/int/123/path/one/two/then/something/else')
+        );
+        $this->assertSame(
+            '/int/?/path/?/?/then/something/?',
+            \DDtrace\Private_\util_uri_normalize_outgoing_path('/int/123/path/one/two/then/something/else')
+        );
+    }
+
+    public function testPatternCanNormalizeSingelFragment()
```
SammyK, line 377: Tiny nit: `Single`

src/private/functions.php
```diff
@@ -0,0 +1,121 @@
...
+
+    $fragmentRegexes = \ddtrace_config_path_fragment_regex();
+    $incomingMappings = \ddtrace_config_path_mapping_incoming();
+    $outgoingMappings = \ddtrace_config_path_mapping_outgoing();
+
+    // We can now be in one of 3 cases:
+    //   1) At least one among DD_TRACE_RESOURCE_URI_FRAGMENT_REGEX and DD_TRACE_RESOURCE_URI_MAPPING_INCOMING|OUTGOING
+    //      is defined. Then ignore legacy DD_TRACE_RESOURCE_URI_MAPPING and apply *new normalization*.
+    //   2) Only DD_TRACE_RESOURCE_URI_MAPPING is defined, then apply *legacy normalization* for backward compatibility.
+    //   3) Nothing is defined, then apply *new normalization*.
```
SammyK, line 69: Just to confirm the behavior: as of right now if someone is currently using `DD_TRACE_RESOURCE_URI_MAPPING` to configure all their incoming URL's and then set `DD_TRACE_RESOURCE_URI_MAPPING_OUTGOING` after this release, the incoming URL's will no longer get normalized. If that is the case we should definitely make that clear in the documentation. :)

src/private/functions.php
```diff
@@ -0,0 +1,121 @@
...
+ * @return string
+ */
+function _util_uri_apply_rules($uriPath, $incoming)
+{
+    if ('/' === $uriPath || '' === $uriPath || null === $uriPath) {
+        return '/';
+    }
+
+    // We always expect leading slash if it is a pure path, otherwise if it is a full url we preserve schema and port.
+    if ($uriPath[0] !== '/' && substr($uriPath, 0, 4) !== "http") {
```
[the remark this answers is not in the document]
labbati: This is a good idea. What a good list for you would be? http, https, ftp, news? Or maybe the `:`?

src/private/functions.php
labbati: This is the behavior that I proposed.

> If that is the case we should definitely make that clear in the documentation

Absolutely, I am currently working on the docs PR. Do you agree with the behavior?

SammyK: I'm good with the behavior given good docs. 😄

src/private/functions.php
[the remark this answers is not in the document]
SammyK: Matching on `:` could miss paths like `wiki/Template:Welcome`. Perhaps a regex like `/^[a-z][a-z0-9+\-.]*:\/\//` would be a bit safer?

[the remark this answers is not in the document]
labbati: Like it. Doing right now. Thanks.

tests/Unit/private/UriTest.php
```diff
@@ -427,6 +427,31 @@ public function testItWorksWithHttpsFulllUrls()
...
+    public function testItWorksWithComplexSchemePatternAsDefinedByRFC3986()
+    {
+        $this->putEnvAndReloadConfig([
+            'DD_TRACE_RESOURCE_URI_FRAGMENT_REGEX=^abc$',
+            'DD_TRACE_RESOURCE_URI_MAPPING_INCOMING=nested/*',
+            'DD_TRACE_RESOURCE_URI_MAPPING_OUTGOING=nested/*',
+        ]);
+
+        // https://tools.ietf.org/html/rfc3986#page-17
+        $rfc3986CompliantScheme = 'letter+1-2-3.CAPITAL.123';
```
SammyK, line 439: Nice! :)

PR #930, opened 2020-06-19
Add DDTrace namespace for internal functions

SammyK:
### Description

This PR adds the `DDTrace` namespace to internal functions so that going forward, all new API's can be added under this namespace. This change only applies to two existing functions:

* `dd_trace_function` -> `DDTrace\trace_function`
* `dd_trace_method` -> `DDTrace\trace_method`

The old functions `dd_trace_function` & `dd_trace_method` remain as aliases to their namespaced counterparts.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Slim/SlimSandboxedIntegration.php
+++ b/src/DDTrace/Integrations/Slim/SlimSandboxedIntegration.php
@@ -43,7 +43,7 @@ class SlimSandboxedIntegration extends SandboxedIntegration
             $rootSpan->setTag(Tag::SERVICE_NAME, $appName);
 
             // Hook into the router to extract the proper route name
-            \DDTrace\trace_method('Slim\Router', 'lookupRoute', function (SpanData $span, $args, $return) use ($rootSpan) {
+            \DDTrace\trace_method('Slim\Router', 'lookupRoute', function (SpanData $s, $a, $return) use ($rootSpan) {
                 /** @var \Slim\Interfaces\RouteInterface $route */
                 $route = $return;
                 $rootSpan->setTag(
--- a/src/DDTrace/Integrations/Slim/SlimSandboxedIntegration.php
+++ b/src/DDTrace/Integrations/Slim/SlimSandboxedIntegration.php
@@ -29,7 +29,7 @@ class SlimSandboxedIntegration extends SandboxedIntegration
         $integration = $this;
         $appName = \ddtrace_config_app_name(self::NAME);
 
-        \dd_trace_method('Slim\App', '__construct', function () use ($integration, $appName) {
+        \DDTrace\trace_method('Slim\App', '__construct', function () use ($integration, $appName) {
             // At the moment we only report internals of Slim 3.*
             $majorVersion = substr(self::VERSION, 0, 1);
             if ('3' !== $majorVersion) {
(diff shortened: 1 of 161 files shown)


src/DDTrace/Integrations/Slim/SlimSandboxedIntegration.php
```diff
@@ -43,7 +43,7 @@ public function init()
             $rootSpan->setTag(Tag::SERVICE_NAME, $appName);
 
             // Hook into the router to extract the proper route name
-            \dd_trace_method('Slim\Router', 'lookupRoute', function (SpanData $span, $args, $return) use ($rootSpan) {
+            \DDTrace\trace_method('Slim\Router', 'lookupRoute', function (SpanData $s, $a, $return) use ($rootSpan) {
```
labbati, line 46: Just asking to adopt the same approach in similar scenarios. Do we want to use `$s` and `$a` meaning 'not used'? Something like the `_` of other languages?

tests/ext/sandbox-prehook/access_modifier_method_access_hook.phpt
```diff
@@ -45,13 +45,13 @@ class Bar
     }
 }
 
-dd_trace_method('Foo', 'aPublic', ['prehook' => function() {
+DDTrace\trace_method('Foo', 'aPublic', ['prehook' => function() {
```
labbati, line 48: Missing leading `\`.

tests/ext/sandbox-prehook/access_modifier_method_access_hook.phpt
```diff
@@ -45,13 +45,13 @@ class Bar
...
     $this->getBar()->dPublic();
     var_dump($this->bProtected());
     var_dump($this->cPrivate());
 }]);
 
-dd_trace_method('Bar', 'dPublic', ['prehook' => function() {
+DDTrace\trace_method('Bar', 'dPublic', ['prehook' => function() {
```
labbati, line 54: Missing leading `\`.

tests/ext/sandbox-prehook/access_modifier_property_access_hook.phpt
```diff
@@ -15,7 +15,7 @@ class Test
     }
 }
 
-dd_trace_method("Test", "m", ['prehook' => function() {
+DDTrace\trace_method("Test", "m", ['prehook' => function() {
```
labbati, line 18: Missing leading `\`. And in some files below

tests/ext/sandbox-prehook/access_modifier_method_access_hook.phpt
SammyK: Yes, this is intentional for the phpt tests since they are all in the root namespace anyway (with the exception of the namespace test). So I just kept the calling reference the same as the original. :)

labbati: 👍  sounds good them, resolving this and the next ones

src/DDTrace/Integrations/Slim/SlimSandboxedIntegration.php
SammyK: I like this idea! I changed these just to get the linter to pass, but having a consistent "not used" convention would be good to implement across the board. Are we good with the first initial of each param: `$s`, `$a`, `$r`? Or something else?

labbati:
Thanks for the `DDTrace` in `phpt` files explanation @SammyK , resolved all the comments now, good to merge for me.

PR #932, opened 2020-06-23
Refactor dispatch.{c,h}

morrisonlevi:
### Description

Ahead of some work on the non-instrumentation API, I discovered some cruft in the dispatch process. This PR:

- Moves `dispatch_setup.c` into `dispatch.c`
- Refactors a few functions in those files and deletes some unnecessary helpers
- Extracts common code into `ddtrace_prehook` and `ddtrace_posthook`
- Removes some `DD_PRINTF` usage

Overall it deletes more code than it adds, which I am happy about. There shouldn't be any behavior changes, and don't expect it to substantially affect overhead. It will lower the amount of memory used, however, I expect the non-instrumentation API to increase it more than this reduces it.

There is more work to do, but it was getting large enough that I figured it was time for code review.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


config.m4
```diff
@@ -36,7 +36,6 @@ if test "$PHP_DDTRACE" != "no"; then
     src/ext/configuration_php_iface.c \
     src/ext/ddtrace_string.c \
     src/ext/dispatch.c \
-    src/ext/dispatch_setup.c \
```
SammyK, line 39: 🔝

src/ext/php7/engine_hooks.c
```diff
@@ -752,7 +769,7 @@ static void _dd_return_helper(zend_execute_data *execute_data) {
             ZVAL_NULL(&rv);
             retval = &rv;
         }
-        _dd_end_span(span, retval);
+        ddtrace_posthook(NULL, span, retval);
```
SammyK, line 772: We should have access to the fbc from the `execute_data` here (and in the other return handlers). Or is there a reason we need to keep it `NULL`?

morrisonlevi: It's because we don't need to check for a span stack mismatch because of the way the handlers works -- if there was a mismatch we wouldn't even be running the posthook in these handlers.

SammyK: Good point. Just FYI, that assumption will likely change in #931.

PR #945, opened 2020-07-03
Sandbox Lumen on PHP 5

morrisonlevi:
### Description

This adds some un-necessary spans that can be removed once the non-tracing API is completed, but it is worth it to get off of `dd_trace`.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Lumen/LumenSandboxedIntegration.php
+++ b/src/DDTrace/Integrations/Lumen/LumenSandboxedIntegration.php
@@ -63,14 +63,18 @@ class LumenSandboxedIntegration extends SandboxedIntegration
             }
         );
 
+        // convert to non-tracing API
+        $hook = 'posthook';
         // Extracting resource name as in legacy integration
         \DDTrace\trace_method(
             'Laravel\Lumen\Application',
             'handleFoundRoute',
             [
-                'prehook' => function (SpanData $span, $args) use ($rootSpan) {
+                $hook => function (SpanData $span, $args) use ($rootSpan, $appName) {
+                    $span->service = $appName;
+                    $span->type = 'web';
                     if (count($args) < 1 || !\is_array($args[0])) {
-                        return false;
+                        return;
                     }
                     $routeInfo = $args[0];
                     $resourceName = null;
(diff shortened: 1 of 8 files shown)


src/DDTrace/Integrations/Lumen/LumenSandboxedIntegration.php
```diff
@@ -68,9 +68,11 @@ function (SpanData $span, $args) use ($rootSpan, $integration, $appName) {
             'Laravel\Lumen\Application',
             'handleFoundRoute',
             [
-                'prehook' => function (SpanData $span, $args) use ($rootSpan) {
+                // convert to non-tracing API
+                'posthook' => function (SpanData $span, $args) use ($rootSpan) {
```
labbati, line 72: You can `use($appName)` here.

labbati, line 72: Not sure about the scope of this change. Isn't prehook returning `false` supported on 5.6 with sandboxing api, at this stage?

labbati: If not, should we change also `handleUncaughtException` and `sendExceptionToHandler`?

morrisonlevi: > Isn't prehook returning false supported on 5.6 with sandboxing api, at this stage?

No.

> If not, should we change also handleUncaughtException and sendExceptionToHandler?

Probably yes; looking into it.

src/DDTrace/Integrations/Lumen/LumenSandboxedIntegration.php
```diff
@@ -63,14 +63,17 @@ function (SpanData $span, $args) use ($rootSpan, $integration, $appName) {
...
+        // convert to non-tracing API
+        $hook = 'posthook';
         // Extracting resource name as in legacy integration
         \DDTrace\trace_method(
             'Laravel\Lumen\Application',
             'handleFoundRoute',
             [
-                'prehook' => function (SpanData $span, $args) use ($rootSpan) {
+                $hook => function (SpanData $span, $args) use ($rootSpan, $appName) {
+                    $span->service = $appName;
```
SammyK, line 74: I just realized we need to also set `$span->type = 'web';` to be consistent with the other instrumented bits from the Laravel integration.

morrisonlevi: Thanks; set this in most recent version.

PR #948, opened 2020-07-07
Add support for phpredis 3.* extension on php 7

labbati:
This PR adds support for PHPRedis extension (version 3) on PHP 7.

Highlights:
- See branch `labbati/phpredis-4-5` for work on clients 4 and 5. TO be merged immediately after.
- All functions that can be interested to observe have been traced.
- Only `Redis` class methods are traced.
- Original library's project test suites for `Redis` and `RedisArray` are tested to make sure we do not break any existing functionality.
- arguments normalization has been made compatible with the datadog agent obfuscation expectations.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
Luca, is this tracing all `Redis` methods? If so then maybe at the C level we should just iterate over all the methods and override their handler.

labbati:
> Luca, is this tracing all `Redis` methods? If so then maybe at the C level we should just iterate over all the methods and override their handler.

Not all of them but a good 50-60%

nesl247:
Any idea when this will be completed? We've been waiting on this for a long time, and are excited to get it.

labbati:
> Any idea when this will be completed? We've been waiting on this for a long time, and are excited to get it.

👋  which version of the client you are on? We are focusing on 3.x but we can easily add other versions.

nesl247:
> > Any idea when this will be completed? We've been waiting on this for a long time, and are excited to get it.
> 
> 👋 which version of the client you are on? We are focusing on 3.x but we can easily add other versions.

We're on 0.46.0 I believe at the moment. We are planning to update to 0.47.0 soon, so having a 0.48.0 with this would be great.

labbati:
> We're on 0.46.0 I believe at the moment. We are planning to update to 0.47.0 soon, so having a 0.48.0 with this would be great.

We are trying to put this in 0.48.0, I was wondering which version of phpredis extension you are on.

nesl247:
Just realized you meant phpredis. I was wondering why you guys were bumping the version of this so high 😄 . Anyways, we're on 4.3.0 for the extension.

src/DDTrace/Integrations/PHPRedis/PHPRedisSandboxedIntegration.php
```diff
@@ -0,0 +1,200 @@
...
+    {
+        \DDTrace\trace_method('Redis', $method, function (SpanData $span, $args) use ($method) {
+            PHPRedisSandboxedIntegration::enrichSpan($span, $method);
+            // Obfuscable methods: see https://github.com/DataDog/datadog-agent/blob/master/pkg/trace/obfuscate/redis.go
+            $span->meta[Tag::REDIS_RAW_COMMAND] = $method . ' ' . PHPRedisSandboxedIntegration::normalizeArgs($args);
+        });
+    }
+
+    /**
+     * Based on logic from pyhton tracer:
```
nesl247, line 143:
```suggestion
     * Based on logic from python tracer:
```

labbati:
Also @nesl247 are you using the `Redis` class or the `RedisCluster` class?

nesl247:
@labbati `Redis`.

tests/Integrations/PHPRedis/PHPRedisProjectTestsuite/3.x/RedisArrayTest.php
```diff
@@ -0,0 +1,554 @@
+<?php defined('PHPREDIS_TESTRUN') or die("Use TestRedis.php to run tests!\n");
```
SammyK, line 1: Is there a license file we can include for these vendored files?

src/DDTrace/Integrations/PHPRedis/PHPRedisSandboxedIntegration.php
```diff
@@ -0,0 +1,316 @@
...
+                $partValue = (string)$arg;
+            } elseif (\is_null($arg)) {
+                $partValue = 'null';
+            } elseif (\is_bool($arg)) {
+                $partValue = $args ? 'true' : false;
+            } elseif (\is_array($arg)) {
+                // This is best effort as specific index might be missing or be shifted, e.g. [0 => 'a', 2 => 'b'].
+                // In this case the worst that can happen is that we generate '0 a 2 b' instead of 'a b'. We accept this
+                // in order to keep things as simple as possible.
+                $isAssociative = array_values($arg) !== $arg;
```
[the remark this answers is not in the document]
labbati: Very good call. Worst that can happen is that we miss to add some keys, so it is worth the change. Thanks Sammy.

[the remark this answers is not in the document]
labbati: I am having double thoughts about this, as many times the key is numeric and would be easily confused.

I proposed a slightly more complex solution that might serve us decently in this case.

```function isAssociative($array)
{
    $expected = 0;
    foreach ($array as $key => $value) {
        if ($key !== $expected++) {
            return true;
        }
    }
    return false;
}

var_dump(isAssociative([ 'a', 'b' ])); // false
var_dump(isAssociative([ 'a' => 1, 'b' => 2 ])); // true
var_dump(isAssociative([ '0' => 1, '2' => 2 ])); // true: false positive
var_dump(isAssociative([ '0' => 1, '1' => 2 ])); // false
```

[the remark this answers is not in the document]
SammyK: Brilliant! Are we expecting string keys that are sequential ints to be considered numeric arrays? `[ '0' => 1, '1' => 2 ]` or only true int keys? `[ 0 => 1, 1 => 2 ]` Just asking because in the snippet the `foreach` has a strict comparison on `$key !== $expected++`.

Nice solution! 👍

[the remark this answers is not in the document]
labbati: > Are we expecting string keys that are sequential ints to be considered numeric arrays?

In my mind they should be associative...because if they are string then I expect them to be put there on purpose. But I don't have strong opinions about it.

tests/Integrations/PHPRedis/PHPRedisProjectTestsuite/3.x/RedisArrayTest.php
labbati: just as a followup, this has been removed.

src/DDTrace/Integrations/IntegrationsLoader.php
```diff
@@ -117,6 +118,8 @@ public function __construct(array $integrations)
                 '\DDTrace\Integrations\Mysqli\MysqliSandboxedIntegration';
             $this->integrations[PDOSandboxedIntegration::NAME] =
                 '\DDTrace\Integrations\PDO\PDOSandboxedIntegration';
+            $this->integrations[PHPRedisSandboxedIntegration::NAME] =
```
SammyK, line 121: Should this only be loaded on PHP 7?

tests/assert_extension_enabled.php
```diff
@@ -0,0 +1,8 @@
+<?php
```
SammyK, line 1: Is this file called from anywhere? I don't see where it is referenced.

src/ext/php7/handlers_phpredis.c
```diff
@@ -0,0 +1,185 @@
+#include "handlers_internal.h"
```
SammyK, line 1: I totally missed this the first time around: We'll need to add `phpredis` to the [`ddtrace_integration_name` enum](https://github.com/DataDog/dd-trace-php/blob/c8474ce/src/ext/integrations/integrations.h#L10-L30) as well as the [`ddtrace_integrations` array](https://github.com/DataDog/dd-trace-php/blob/c8474ceee6e3a50945a94ae80fc6e418acde383b/src/ext/integrations/integrations.c#L7-L27) otherwise the integration-specific config will always use the defaults for `phpredis`.

I should add a test to check this every time we add a new integration.

tests/assert_extension_enabled.php
labbati: this was actually here from when I was running the original test suite as part of our tests. now we can remove it. Thanks

src/ext/php7/handlers_phpredis.c
SammyK: Test for integrations at the extension level are added in #985.

PR #952, opened 2020-07-09
improve CGI usage in test suite

remicollet:
honors standard env. variable.

especially TEST_PHP_ARGS which may provides directive to load needed extension (json)

For ex, for RPM, we use `TEST_PHP_ARGS="-n -d extension=curl.so -d extension=json.so -d extension=posix.so -d extension=$PWD/modules/ddtrace.so"

And TEST_PHP_EXECUTABLE can be /usr/bin/zts-php (if this case, as /usr/bin/zts-php-cgi doesn't exists, tests are skipped).

### DIFF
--- a/tests/ext/startup_logging.inc
+++ b/tests/ext/startup_logging.inc
@@ -2,8 +2,12 @@
 
 function dd_get_php_cgi()
 {
-    $executable = dirname(getenv('TEST_PHP_EXECUTABLE')) . '/php-cgi';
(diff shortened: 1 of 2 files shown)


remicollet:
Second commit is related to old system where output is ""Could not resolve host: invalid_host;unknown error"

remicollet:
Sadly not in 0.47.1...

SammyK:
Yes, this should have been included and it was completely my fault for missing it. I'm very sorry @remicollet.

PR #957, opened 2020-07-13
Fix compilation on older GCC compilers

morrisonlevi:
### Description

GCC on CentOS 7 thinks (ddtrace_string) { ... } is not constant, so
we work around this by using local variables. We should start using
the newer developer toolset on CentOS so hacks like this are not
necessary.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
Closed in favor of #958.

PR #963, opened 2020-07-15
Consistent env vars: Fix propagation of DD_ENV and DD_VERSION to children spans

SammyK:
### Description

The values from `DD_ENV` and `DD_VERSION` take precedence over the same vales from `DD_TAGS` (i.e. `DD_TAGS=version:foo,env:foo`). The precedence was honored for userland children spans, but not for internal spans. This PR fixes this issue.

The precedence order is:

1. Spans with explicit conflicting tags (set via `Span::setTag()` or `SpanData::$meta[]`) will not be overwritten by `DD_ENV` and `DD_VERSION`, or `DD_TAGS`.
2. If no explicit tag exists, the value from `DD_ENV` and `DD_VERSION` will be used.
3. If `DD_ENV` and `DD_VERSION` are not set, the values from `DD_TAGS` will be used.

This was the existing behavior, but this PR addresses an issue setting `DD_TAGS` along with `DD_ENV`/`DD_VERSION`. With this fix, `env` and `version` values from `DD_TAGS` will always be overwritten by `DD_ENV` and `DD_VERSION` respectively.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Tracer.php
+++ b/src/DDTrace/Tracer.php
@@ -181,7 +181,7 @@ final class Tracer implements TracerInterface
             $options->getStartTime()
         );
 
-        $tags = $options->getTags() + $this->config['global_tags'];
+        $tags = $options->getTags() + $this->getGlobalTags();
         if ($context->getParentId() === null) {
             $tags[Tag::PID] = getmypid();
         }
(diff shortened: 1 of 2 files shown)


morrisonlevi:
> The precedence was honored for userland children spans, but not for internal spans. This PR fixes this issue.

My reading of the PR made it seem like it's the opposite, as this deals with the `Tracer:: startSpan` method?

SammyK:
@morrisonlevi I wasn't very clear so I added more context to the description. The precedence this is fixing has to do with the context where `DD_TAGS` and `DD_ENV`/`DD_VERSION` are set at the same time. It doesn't affect the existing precedence for tags.

brettlangdon:
the new precedence order here is great, makes sense and it what we expect 👍

PR #969, opened 2020-07-20
Consistent env vars: Support for DD_TRACE_<INTEGRATION>_<SUFFIX> env vars

SammyK:
### Description

This PR normalizes usage of integration-specific environment variable names to be consistent with other tracers.

| Old Name  | New Name |
| ------------- | ------------- |
| `DD_<INTEGRATION>_ANALYTICS_ENABLED` | `DD_TRACE_<INTEGRATION>_ANALYTICS_ENABLED`  |
| `DD_<INTEGRATION>_ANALYTICS_SAMPLE_RATE `  | `DD_TRACE_<INTEGRATION>_ANALYTICS_SAMPLE_RATE `  |
| `DD_INTEGRATIONS_DISABLED` | `DD_TRACE_<INTEGRATION>_ENABLED` |

For example, to disable the `pdo` and `laravel` integrations before this PR:

```
# Deprecated
DD_INTEGRATIONS_DISABLED=pdo,laravel
```

Starting from 0.47.1:

```
DD_TRACE_PDO_ENABLED=false
DD_TRACE_LARAVEL_ENABLED=false
```

The old names will continue to work but are deprecated and may be removed in a future release.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/ext/configuration.h
+++ b/src/ext/configuration.h
@@ -6,6 +6,7 @@
 #include "compatibility.h"
 #include "ddtrace_string.h"
 #include "env_config.h"
+#include "integrations/integrations.h"
 
 /**
  * Returns true if `subject` matches "true" or "1".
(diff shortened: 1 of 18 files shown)


src/ext/configuration.h
```diff
@@ -30,8 +31,22 @@ bool ddtrace_config_env_bool(ddtrace_string env_name, bool default_value TSRMLS_
 bool ddtrace_config_distributed_tracing_enabled(TSRMLS_D);
 bool ddtrace_config_trace_enabled(TSRMLS_D);
 
+#define DDTRACE_LONGEST_INTEGRATION_ENV_PREFIX_LEN 10  // "DD_TRACE_" FTW!
+#define DDTRACE_LONGEST_INTEGRATION_ENV_SUFFIX_LEN 23  // "_ANALYTICS_SAMPLE_RATE" FTW!
```
morrisonlevi, line 35: These are one larger than the string lengths; any reason for that? Usually +1 is for NULL but since we're concatenating strings there won't be nulls between them.

src/ext/configuration.h
```diff
@@ -30,8 +31,22 @@ bool ddtrace_config_env_bool(ddtrace_string env_name, bool default_value TSRMLS_
...
+    (DDTRACE_LONGEST_INTEGRATION_ENV_PREFIX_LEN + DDTRACE_LONGEST_INTEGRATION_NAME_LEN + \
+     DDTRACE_LONGEST_INTEGRATION_ENV_SUFFIX_LEN)
+
 // note: only call this if ddtrace_config_trace_enabled() returns true
 bool ddtrace_config_integration_enabled(ddtrace_string integration TSRMLS_DC);
+bool ddtrace_config_integration_enabled_ex(ddtrace_integration_name integration_name TSRMLS_DC);
+bool ddtrace_config_integration_analytics_enabled(ddtrace_string integration TSRMLS_DC);
+bool ddtrace_config_integration_analytics_enabled_ex(ddtrace_integration_name integration_name TSRMLS_DC);
+double ddtrace_config_integration_analytics_sample_rate(ddtrace_string integration TSRMLS_DC);
+double ddtrace_config_integration_analytics_sample_rate_ex(ddtrace_integration_name integration_name TSRMLS_DC);
```
morrisonlevi, line 46: Does this one need exposed? I only saw it used to implement `ddtrace_config_integration_analytics_sample_rate`.

src/ext/integrations/integrations.c
```diff
@@ -1,8 +1,50 @@
...
+};
+size_t ddtrace_integrations_len = 0;
+
+// Map of lowercase strings to the ddtrace_integration equivalent
+static HashTable _dd_string_to_integration_name_map;
+
+static void _dd_add_integration_to_map(char* name, size_t name_len, ddtrace_integration* integration);
+
+void ddtrace_integrations_minit(void) {
+    ddtrace_integrations_len = sizeof ddtrace_integrations / sizeof ddtrace_integrations[0];
```
morrisonlevi, line 36: This can be done up above on line 28.

src/ext/configuration.h
SammyK: Nope, that's just an off by one error from entering in what my IDE was telling me. 

Screen Shot 2020-07-29 at 6 56 08 AM

Nice catch!

src/ext/configuration.h
SammyK: Nice catch! I've also removed `ddtrace_config_integration_analytics_enabled_ex` since it doesn't need to be public either.

src/ext/integrations/integrations.c
SammyK: Thanks for pointing this out. My original thought was `ddtrace_integrations_len == 0` means that the integrations have not been set up yet so that you could check if the `_dd_string_to_integration_name_map` had been initialized. But since that is private, we probably don't need this. Plus I can see how giving `ddtrace_integrations_len == 0` semantics is confusing. So I moved it per your suggestion. :)

PR #978, opened 2020-07-30
Add failing test cases for Guzzle integration

RaoulMeyer:
### Description

After upgrading from Guzzle 5 to Guzzle 6, our distributed tracing doesn't work anymore. The traces still show up separately in APM, but they are not combined as they should be.

Reason for this seems to be an issue in the Guzzle integration implementation specifically for Guzzle 6. The issue is in this piece of code:

```
    /**
     * @param mixed $request
     * @param array $headers
     */
    private function addRequestHeaders($request, $headers)
    {
        if (is_a($request, '\GuzzleHttp\Message\MessageInterface')) {
            $request->setHeaders($headers);
        } elseif (is_a($request, '\Psr\Http\Message\MessageInterface')) {
            foreach ($headers as $name => $value) {
                $request->withAddedHeader($name, $value);
            }
        }
    }
```

From Guzzle 6 on, the library uses the immutable PSR Request class. The function `$request->withAddedHeader($name, $value);` returns a new request object with the header added, but doesn't change the existing request object. To fix this issue, the changed request object would have to be passed back and used in the method call.

I've reproduced this issue in the integration tests, this PR adds an extra case and makes them fail. The failing cases are skipped for now. I was looking into implementing a fix as well, but this ended up being more work than I thought, because the `$this->codeTracer->tracePublicMethod(` helper then can't be used anymore.

The reason the tests didn't catch the issue before is that the CurlIntegration was still adding the headers. It could be that I'm misunderstanding how these integrations should interact, please let me know if that's the case.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
Since you don't have distributed tracing after upgrading, does this mean you aren't using the curl back-end to Guzzle? What back-end are you using?

RaoulMeyer:
Good question, I dove into it a bit more and now have some new insights.

We don't specify any backend, so Guzzle defaults to using `CurlMultiHandler`. This explains why we are not seeing any curl spans, because the Curl integration is targeted at the regular `curl_exec` and related functions.

We are also not seeing any Guzzle spans, and this is due to a related problem. We use an adapter (https://github.com/php-http/guzzle6-adapter) that calls the `->sendAsync` method regardless of if your call is async or not, and waits if it was meant to be sync. The Guzzle integration only adds spans to the `->send` method, which calls `->sendAsync`.

Switching to the non-multi curl handler would already solve most of our issues for now.

These two issues are not directly related to this PR. I think it still might be good to make sure the Guzzle integration does add the distributed traces header in case anyone doesn't use the curl handlers for whatever reason. Of course I'm curious to hear your ideas on all this.

tests/Integrations/Guzzle/V6/GuzzleIntegrationTest.php
```diff
@@ -25,7 +25,7 @@ class GuzzleIntegrationTest extends IntegrationTestCase
     public static function setUpBeforeClass()
     {
         parent::setUpBeforeClass();
-        IntegrationsLoader::load();
+        GuzzleIntegration::load();
```
morrisonlevi, line 28: This switch basically means `curl` isn't loaded, I think? Is that the purpose here?

RaoulMeyer: Yes that is the purpose. The curl integration is what is making the tests pass currently, by only loading the guzzle integration the test starts failing.

morrisonlevi:
Okay, so I messed up a git command and accidentally closed this issue. Very sorry about that! I was just trying to push some updates after rebasing to the latest master.

PR #991, opened 2020-08-19
Enable the BGS by default on PHP 5.4

SammyK:
### Description

This PR enables the background sender by default on PHP 5.4.

This also includes a fix to CI where the phpt tests would not properly report the exit code of `run-tests.php` when running it via `make` on PHP 5.4. Running `run-tests.php` directly fixes this bug as well as a number of other issues we've run into in the past with running the test runner through `make`.

### Readiness checklist
- ~[ ] Tests added for this feature/bug.~ (Added "PHP 54 web tests with nginx + FastCGI" tests)

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ aliases:
   - &STEP_EXT_INSTALL
     run:
       name: Build and install extension
-      command: make sudo install install_ini BUILD_DIR=tmp/build_extension
+      command: make sudo install install_ini BUILD_DIR=$(pwd)/tmp/build_extension
 
   - &STEP_COMPOSER_SELF_UPDATE
     run:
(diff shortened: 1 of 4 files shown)


morrisonlevi:
It looks like in its basic form everything is okay. I wonder if we should do some manual tests to be sure?

morrisonlevi:
Moving to 0.49.0.

.circleci/config.yml
```diff
@@ -392,7 +392,7 @@ jobs:
           valgrind_config: << parameters.valgrind_config >>
       - run:
           name: Run extension tests with leak detection
-          command: make << parameters.unit_test_target >> BUILD_DIR=tmp/build_extension JUNIT_RESULTS_DIR=$(pwd)/test-results
+          command: make << parameters.unit_test_target >> BUILD_DIR=$(pwd)/tmp/build_extension JUNIT_RESULTS_DIR=$(pwd)/test-results
```
morrisonlevi, line 395: Any insight here as to why `pwd` was added? I occasionally build it somewhere else, like in `/tmp/build` (note leading slash).

SammyK: Yes, `TEST_PHP_SRCDIR` needs to be an absolute path. You can still build it somewhere else as long as `BUILD_DIR` is an absolute path.

PR #999, opened 2020-08-27
Remove any reference to legacy integrations from userland php code

labbati:
### Description

After we removed the legacy api in release 0.48.0, we want to get rid of:

- All naming explicitly referencing `Sandboxed` in the name, as now `Integration`s ARE sandboxed, no alternatives.
- Remove all differentiators between sandboxed and non sandboxed testing

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Integrations/Web/WebIntegration.php
```diff
@@ -32,6 +32,13 @@ public function getName()
         return self::NAME;
     }
 
+    public function init()
+    {
+        // For now we do nothing, as this is done in the bootstrap logic at the moment. We may consider doing this
+        // here instead, but leaving this for a future refactoring.
+        return Integration::LOADED;
+    }
+
```
morrisonlevi, line 41: This change is fine, but shouldn't we also remove `load()`?

src/api/Configuration.php
```diff
@@ -214,7 +214,7 @@ public function isHttpClientSplitByDomain()
      */
```
morrisonlevi, line 214: I think we should slap a `@deprecated` on here.

morrisonlevi:
This has been squashed, rebased, and cleaned up in PR 1003.

Issue #812, opened 2020-04-04
[BUG]PHP5.5 Not Supported


imsatender:
**PHP5.5 Not Supported with DDTRACE_0.36.1**

I am getting the error mentioned below:

`PHP Warning:  PHP Startup: Unable to load dynamic library '/opt/datadog-php/extensions/ddtrace-20121113.so' - /opt/datadog-php/extensions/ddtrace-20121113.so: cannot open shared object file: No such file or directory in Unknown on line 0
PHP 5.5.38 (cli) (built: Oct 22 2019 14:12:40) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans`

morrisonlevi:
We do not currently support PHP 5.5. It is planned, but I can't give an ETA; see #625 for more users who also want PHP 5.5 support.

morrisonlevi:
PHP 5.5 support has been merged in. It still needs more tests before it's officially supported, but wanted you to know it's coming soon, probably in version 0.49.0.

SammyK:
Thank you for opening this issue @imsatender. PHP 5.5 is now supported in [0.49.0](https://github.com/DataDog/dd-trace-php/releases/tag/0.49.0).

Issue #625, opened 2019-10-24
Add support for PHP 5.5 on Ubuntu 14.04


TomKeur:
Since we're running some legacy systems on Ubuntu 14.04 which ships with PHP 5.5, we want to get support for PHP 5.5.

I know PHP 5.5 is officially End Of Life, but since we're running Ubuntu 14.04 with [Ubuntu Extended Security Maintenance](https://ubuntu.com/esm), we're still getting security updates for PHP 5.5.

I know when we will upgrade to PHP 5.6 the extension will work, but upgrading to PHP 5.6 has no effort for us. Because when we do, we're losing our security patches since PHP 5.6 is End Of Life too.

My issue is related to #616.

PS: We're upgrading to PHP 7.x, but this is a steep road up.

labbati:
Thanks @TomKeur for opening this issue. You made a very good point and we definitely want to support your use case.
We plan to start working on this during this quarter. While I cannot commit to deliver by EOQ, we will at least be able to provide a better estimate once with start dig into this.

As a side note a good news: we (@morrisonlevi) was able to build for 5.5 with no problems and a decent amount of tests already pass. This might let us being positive.

TomKeur:
Thanks for your reply @labbati.
This is great news :)

morrisonlevi:
What is your webserver setup? Is this Apache, nginx..? When this work gets scheduled it would be nice to have this information.

TomKeur:
We're using Apache as mod_php, classic (since it's PHP 5.5 :)).

TomKeur:
@morrisonlevi Hi, any news on this? Looking really forward to start using this in production.
Thanks!

TomKeur:
@labbati @morrisonlevi Can I please have an update? Since we're really in need of this.

labbati:
Hi @TomKeur, I am bringing in this with our PM and get back to you, but let me give you some context about the work that is required to be done.

The PHP tracer hooks into the opcode handlers for function/method calls. Then from within our hooks we forward the call to the original method. This has, for us, the huge disadvantages that 1) there might potentially be stability related issues when we work along side other extensions and 2) that we have to do a lot of heavy lifting on behalf of the VM, and this can be crazy different for different versions of PHP.

[...]

PR #1008, opened 2020-09-04
Support PHP 5.5

morrisonlevi:
### Description

This PR is for adding official support for PHP 5.5. 

Addresses issues #625 and #812.

This is missing at least 3 test suites, which have been approved to be added in a follow-up PR:

- Hyperfine benchmarks.
- Language tests.
- Running the phpt tests with valgrind.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


package.xml
```diff
@@ -446,7 +446,7 @@
     <dependencies>
         <required>
             <php>
-                <min>5.6</min>
+                <min>5.5</min>
```
SammyK, line 449: We can update this to PHP 5.4 now. :)

.circleci/php-5.5_web.txt
```diff
@@ -0,0 +1,14 @@
+test-metrics
+cakephp-28-suite
+codeigniter-22-test
+laravel-42-suite
+lumen-52-suite
```
SammyK, line 5: It looks like [Slim 3 supports PHP 5.5](http://www.slimframework.com/docs/v3/start/installation.html) so it might be good to add that to the list.

composer.json
```diff
@@ -263,6 +286,34 @@
...
+            "@symfony-34-update",
+            "@yii-2-update"
+        ],
+        "test-web-55": [
+            "@test-metrics",
+            "@update-web-55",
+            "@cakephp-28-test",
+            "@codeigniter-22-test",
+            "@laravel-42-test",
+            "@lumen-52-test",
```
SammyK, line 306: Same here for Slim 3.

SammyK (commented):
Nice work @morrisonlevi! I think once we add Slim 3 to the tests, this should be good to go.

.circleci/php-5.5_web.txt
morrisonlevi: Are you okay if I handle this in the follow-up PR?

SammyK: Yep! :)

SammyK (approved):

morrisonlevi:
Thanks for the review, Sammy! I'll make sure Slim is addressed in another PR.

Issue #1012, opened 2020-09-10
[BUG] Segmentation Fault


alecavallo:
**Describe the bug**
I have configured a site in CakePHP 2.10.19 in a docker fleet where one container runs an Nginx to serve static content and the other container runs a php-fpm using the image based on [php:7.3.12-fpm-alpine3.10](https://hub.docker.com/_/php?tab=tags&page=1&name=7.3.12-fpm-alpine3.10) 
[Dockerfile.txt](https://github.com/DataDog/dd-trace-php/files/5201538/Dockerfile.txt)

The PHP application connects to MySQL, MongoDB, and Redis to get data and cache it.
After I installed ddtrace for PHP it started to show 502 errors randomly. When I check the logs in PHP FPM container it shows a message like this:
`[10-Sep-2020 11:43:17] WARNING: [pool www] child 7 exited on signal 11 (SIGSEGV - core dumped) after 391.376478 seconds from start`
 Which disappears when I uninstall ddtrace.
I tried to locate where the Segmentation Fault occurs but it happens in different parts of the code each time the error is reproduced.

Also, I enabled the trace support in Xdebug and could obtain the following:
[trace.txt](https://github.com/DataDog/dd-trace-php/files/5201585/trace.txt)

Note the trace is not finished because it was interrupted by the Segmentation Fault.

Let me know if you need more data

**PHP Info**
- Php version:
`PHP 7.3.12 (cli) (built: Nov 22 2019 11:50:51) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.12, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.9.6, Copyright (c) 2002-2020, by Derick Rethans
    with ddtrace v0.48.0, Copyright Datadog, by Datadog
`
- Php modules: 
`[PHP Modules]
apcu
Core
ctype
curl
date
ddtrace
dom
fileinfo
filter
ftp
gd
hash
iconv
json
libxml
mbstring
mongodb
mysqli
mysqlnd
openssl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
readline
redis
Reflection
session
SimpleXML
soap
sockets
sodium
SPL
sqlite3
standard
tokenizer
xdebug
xml
xmlreader
xmlrpc
xmlwriter
zip
zlib

[Zend Modules]
Xdebug
ddtrace
`

**OS Info**
- OS version (docker image):
`NAME="Alpine Linux"
[...]

PR #1013, opened 2020-09-10
Call ddtrace_engine_hooks_{rinit,rshutdown}

morrisonlevi:
### Description

These were added, but then never called :facepalm:

Honestly, we were unlucky that tests did not catch this. This means
that most of the time the function used the same address in subsequent
requests.

I also don't know how to add a test for this, as I don't know how to
force the address to change.

I believe issue #1012 is caused by this, but have not confirmed it yet.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -354,6 +354,7 @@ static PHP_RINIT_FUNCTION(ddtrace) {
         dd_request_init_hook_rinit(TSRMLS_C);
     }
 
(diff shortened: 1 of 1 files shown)


SammyK (approved):
Nice catch!

labbati:
@morrisonlevi @SammyK could you please spend some time together to brainstorm how to test this? Also, does it opens up a similar class of problems we are not testing?

PR #1011, opened 2020-09-08
Update excluded modules

SammyK:
### Description

This PR contains a number of updates to the excluded modules:

* Rename "blacklist" references to "excluded_modules" and remove legacy API references
* Add "Zend Guard Loader" to excluded modules list on PHP 5
* List conflicting modules in startup log diagnostics for easier troubleshooting

It also contains a few additional tiny changes:

* Fix build error on PHP 5.4 ZTS
* Remove legacy `debug.h`
* Remove `DD_TRACE_IGNORE_LEGACY_BLACKLIST` now that the legacy API has been removed

### Readiness checklist
- [x] (only for Members) Changelog has been added to the release document.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.
- [x] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1015, opened 2020-09-10
Release 0.48.2

SammyK:
### Description

Release `0.48.2` with the following:

### Changed
- Update excluded modules #1011
- Deprecate dd-doctor.php #1014

### Fixed
- PHP 5.4 build in ZTS mode #1011
- Call ddtrace_engine_hooks_{rinit,rshutdown} #1013


morrisonlevi (approved):
Looks good to me, Sammy. While looking over the patch once again I think we may want to mention the ZTS fix for PHP 5.4 in the notes. What do you think?

SammyK:
@morrisonlevi Nice catch! Updated.

morrisonlevi (approved):

PR #1017, opened 2020-09-15
Clear separation in testing of src/api vs src/DDTrace

labbati:
### Description

This PR achieves the following goals:

- clear separation of `src/api` tests (now in `tests_api`) from tests of `src/DDTrace` and `src/ext` (still in `tests`).
- the root composer file (the one that defines the php package) only defines `src/api` related tasks, while the root `phpunit.xml` file only define configuration for `tests_api`.
- new `tests/composer.json` and `tests/phpunit.xml` have been added. This files HAVE NO REFERENCE to ANYTHING in `src/*`, in order to simulate what happens in production, where the only link to `src` is through `_generated.php` .
- move tasks definitions from `composer.json` to the `Makefile` in order to
    - have more flexibility in how to define in different scenarios (dev/ci) the request init hook param
    - have more flexibility in how tasks can be defined (including comments etc...); and
    - avoid slowness of execution throw composer; and
    - avoid issues with memory limits
    - introduce some optimizations (e.g. composer update only if composer.lock is older than composer.json or has been cleaned).

This PR is a pre-requisite to move to built package testing.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


tests/Unit/ConfigurationTest.php
```diff
@@ -50,68 +48,56 @@ private function cleanUpEnvs()
 
     public function testTracerEnabledByDefault()
     {
-        $this->assertTrue(Configuration::get()->isEnabled());
```
labbati, line 53: Moved to equivalent class in `tests_api`

Makefile
```diff
@@ -183,5 +182,379 @@ verify_version:
...
+
+### Api tests ###
+API_TESTS_ROOT := ./tests_api
+
+test_api_unit: composer.lock
+	$(Q) $(ENV_OVERRIDE) php $(REQUEST_INIT_HOOK) vendor/bin/phpunit --config=phpunit.xml $(API_TESTS_ROOT)/Unit $(TESTS)
+
+composer.lock: composer.json
+	$(Q) composer update
+
```
SammyK, line 558: Nice idea using make! This might already be on your TODO list: could we also move `composer compile` here too? Something like `make generated` or something? That way we won't need composer to make the generated file.

tests_api/Unit/ConfigurationTest.php
```diff
@@ -0,0 +1,493 @@
+<?php
```
SammyK, line 1: One tiny nit: can we make this `tests/api` instead of `tests_api`? We have other test branches in the `tests` folder like `tests/opcache` and `tests/xdebug` so `tests/api` would follow the existing flow. :)

SammyK (commented):
Looks good so far @labbati! Just left two comments. :)

Makefile
morrisonlevi: We need composer to make the generated file because of the dev dependencies -- so it's not particularly helpful to avoid composer for that specific usage.

SammyK: Ah yes, good point.

morrisonlevi:
One thing to keep in mind is that for some of our very long test suites we want to parallelize them. What I propose that there is a command that is something like `make web-tests-5.5-list` which simply shows the commands to be run, then we can split it in CircleCI using `circleci tests split`. The output should be a list of targets like:

> test-metrics
cakephp-28-suite
codeigniter-22-test
laravel-42-suite
lumen-52-suite
symfony-23-suite
symfony-28-suite
symfony-30-suite
symfony-33-suite
symfony-34-suite
yii-2-suite
wordpress-48-test
zend-framework-1-test
custom-framework-autoloaded-suite

Note that the `suites` that update+test are essential here, because each line might be split into a different job.

When running locally we can do `make web-tests-5.5` instead which will run tests instead of printing them.

Makefile
labbati: So if you agree I leave the compilation in composer, but we will hopefully get rid of it if decide to include the required PHP files in the binary.

tests_api/Unit/ConfigurationTest.php
labbati: sure thing

labbati:
> Note that the `suites` that update+test are essential here, because each line might be split into a different node and we don't want the `update` to run in a different node that runs `test`.

Really like this idea. Created #APMPHP-412 to keep track of it.

.circleci/config.yml
```diff
@@ -511,24 +520,25 @@ jobs:
...
+      - <<: *STEP_COMPOSER_TESTS_UPDATE
       - <<: *STEP_PREPARE_TEST_RESULTS_DIR
       - <<: *STEP_EXPORT_CI_ENV
       - <<: *STEP_WAIT_AGENT
       - <<: *STEP_WAIT_MYSQL
       - <<: *STEP_WAIT_REQUEST_REPLAYER
       - run:
           name: Run auto-instrumentation tests
-          command: composer test-auto-instrumentation -- --log-junit test-results/php-auto-instrumentation/results.xml
+          command: make test_auto_instrumentation PHPUNIT_OPTS="--log-junit test-results/php-auto-instrumentation/results.xml"
```
morrisonlevi, line 532: Can we put `PHPUNIT_OPTS` in the `environment:` section so we don't repeat it in each command?

Makefile
```diff
@@ -1,10 +1,13 @@
+Q := @
```
morrisonlevi, line 1: What is `Q`?

.circleci/config.yml
```diff
@@ -309,7 +313,8 @@ jobs:
       - run:
           name: Set php.ini defaults
           command: printf "date.timezone=UTC\nmemory_limit=-1\n" | sudo tee -a /usr/local/etc/php/php.ini
-      - run: composer update
+      - <<: *STEP_COMPOSER_UPDATE
+      - <<: *STEP_COMPOSER_TESTS_UPDATE
```
morrisonlevi, line 317: At first glance I though this was wrong and shouldn't be here, but looking at it again I think this is just the equivalent of `composer update` from what we were doing before, right? It's not updating each integration suite?

morrisonlevi (commented):
I wasn't really in the groove when reviewing this, but have a few comments/questions.

.circleci/config.yml
labbati: Correct. This updates the newly created `tests/composer.json` project. Individual web apps relying on composer are updated in respective `test_web_x_y` target.

.circleci/config.yml
labbati: I am afraid not, unless we come up with any good ideas. If you take this specific example that you commented, you have for the same job 4 commands (lines 532, 535, 538, and 541) that requires different php unit options (specifically for the test results file path).

It might be possible if we split more tests into separate jobs, though.

Makefile
labbati: It is a switch to decide how 'verbose' you want to be. Basically it set the `@` in front of a command. I am doing this from forever, not sure where I stole the idea.

- you prepend all commands with `$(Q) <actual command>
- by default (`Q := @`) you silence all commands output
- if you want you can do `Q :=` and all actual command invocations get printed.

Makefile
```diff
@@ -183,5 +183,423 @@ verify_version:
...
+	$(call run_tests,tests/Integrations/WordPress/V4_8)
+test_web_yii_2:
+	$(COMPOSER) --working-dir=tests/Frameworks/Yii/Version_2_0_26 update
+	$(call run_tests,tests/Integrations/Yii/V2_0_26)
+test_web_zend_1:
+	$(call run_tests,tests/Integrations/ZendFramework/V1)
+test_web_custom:
+	$(COMPOSER) --working-dir=tests/Frameworks/Custom/Version_Autoloaded update
+	$(call run_tests,--testsuite=custom-framework-autoloaded-test)
+
```
SammyK, line 590: It looks like `DDTrace\Tests\Integrations\Custom\NotAutoloaded\FrameworkNoAutoloadedTest` is not being run anywhere from CI. This isn't something that was introduced with this PR, but it just looks like it isn't being tested currently. Perhaps we could add the non-autoloaded custom framework to the web tests for all PHP versions?

Makefile
```diff
@@ -183,5 +183,423 @@ verify_version:
...
+TEST_INTEGRATIONS_54 := \
+	test_integrations_curl \
+	test_integrations_memcached \
+	test_integrations_mysqli \
+	test_integrations_pdo \
+	test_integrations_elasticsearch1 \
+	test_integrations_guzzle5 \
+	test_integrations_predis1
+
+TEST_WEB_54 := \
```
SammyK, line 208: It looks like `test_metrics` is missing from `TEST_WEB_54`, `TEST_WEB_55`, `TEST_WEB_56`, and `TEST_WEB_70`.

tests/api/Unit/ConfigurationTest.php
```diff
@@ -0,0 +1,493 @@
...
+    {
+        $this->putEnvAndReloadConfig(['DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED=false']);
+        $this->assertFalse(Configuration::get()->isURLAsResourceNameEnabled());
+    }
+
+    public function testGlobalTags()
+    {
+        $this->putEnvAndReloadConfig(['DD_TAGS=key1:value1,key2:value2']);
+        $this->assertEquals(['key1' => 'value1', 'key2' => 'value2'], Configuration::get()->getGlobalTags());
+        $this->assertEquals(['key1' => 'value1', 'key2' => 'value2'], \ddtrace_config_global_tags());
```
SammyK, line 468: This is being tested in `tests/Unit/ConfigurationTest.php`

tests/api/Unit/ConfigurationTest.php
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    public function testGlobalTagsLegacyEnv()
+    {
+        $this->putEnvAndReloadConfig(['DD_TRACE_GLOBAL_TAGS=key1:value1,key2:value2']);
+        $this->assertEquals(['key1' => 'value1', 'key2' => 'value2'], Configuration::get()->getGlobalTags());
+        $this->assertEquals(['key1' => 'value1', 'key2' => 'value2'], \ddtrace_config_global_tags());
```
SammyK, line 475: This is being tested in `tests/Unit/ConfigurationTest.php` as well.

tests/api/Unit/ConfigurationTest.php
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    public function testGlobalTagsNewEnvWinsOverLegacyEnv()
+    {
+        $this->putEnvAndReloadConfig([
+            'DD_TRACE_GLOBAL_TAGS=key10:value10,key20:value20',
+            'DD_TAGS=key1:value1,key2:value2',
+        ]);
+        $this->assertEquals(['key1' => 'value1', 'key2' => 'value2'], Configuration::get()->getGlobalTags());
+        $this->assertEquals(['key1' => 'value1', 'key2' => 'value2'], \ddtrace_config_global_tags());
```
SammyK, line 485: One more that is being tested in `tests/Unit/ConfigurationTest.php`.

SammyK (commented):
Nice work @labbati! This is a much cleaner approach. I left a few questions/feedback. :)

Makefile
morrisonlevi: There was a reason it wasn't being run in CI, but I forget what it was. Only way to find out is to try enabling it, right?

Makefile
labbati: @morrisonlevi are you 👍  with my answer to this comment?

.circleci/config.yml
labbati: @morrisonlevi are you 👍  with my answer to this comment?

Makefile
labbati: good catch

Makefile
```diff
@@ -183,5 +183,423 @@ verify_version:
...
+	test_web_cakephp_28 \
+	test_web_laravel_42 \
+	test_web_yii_2 \
+	test_web_zend_1 \
+	test_web_custom
+
+TEST_INTEGRATIONS_55 := \
+	test_integrations_curl \
+	test_integrations_memcached \
+	test_integrations_mysqli \
```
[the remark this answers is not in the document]
labbati: another very good catch 😄 , just a bad conflict resolution.

Makefile
morrisonlevi: 👍

.circleci/config.yml
morrisonlevi: 👍

SammyK (approved):
Nice work @labbati! 💯

PR #1019, opened 2020-09-15
Add PHP 5.4 to PECL build

morrisonlevi:
### Description

PECL can only do min-max for the required PHP version\*. Since we didn't support PHP 5.5 we couldn't support PHP 5.4 either. Now that we do support PHP 5.5 we can support PHP 5.4 too.

Also adjusts some tests to use relative directories for `request_init_hook`.
Also removes some SKIPIF sections for PHP 5.4.

\* It can do `exclude`s, but it seems we didn't know about this or the excludes were somehow undesirable (such as we may have had to list all PHP 5.5.Z versions in the exclude).

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -685,7 +685,14 @@ jobs:
       - run:
           name: Run phpt tests with PECL
           command: |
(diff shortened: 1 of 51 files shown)


SammyK (approved):
Nice find on the `SKIPIF`s and `TERM` fix @morrisonlevi 👍

PR #1020, opened 2020-09-16
Add hyperfine, language, and unit testsuites to PHP 5.5

morrisonlevi:
### Description

This is a follow-up PR to Support PHP 5.5 #1008 that adds the remaining test suites to make PHP 5.5 support official.

This also changes our `bridge/dd_register_optional_deps_autoloader.php` file to use the unprefixed string callable form to avoid a bug in the language tests for PHP 5. I think it's highly unlikely that this bug would have affected users, but I made the change instead of removing the test just in case.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


SammyK (approved):
Nice work @morrisonlevi!

Issue #1021, opened 2020-09-17
[BUG] Segmentation Fault occurs when extension installed and DD_TRACE_ENABLED=false


walf443:
### Bug description
I encountered SEGV problem.
It caused by some specific endpoint of my app, and other endpoint works well.
I can't understand condition that caused.

coredump's backtrace is followings:

```
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/sbin/apache2 -k start'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fd194783fa9 in zval_ptr_dtor () from /usr/lib/apache2/modules/libphp7.so
[Current thread is 1 (Thread 0x7fd19c7dd040 (LWP 10779))]
(gdb) bt
#0  0x00007fd194783fa9 in zval_ptr_dtor () from /usr/lib/apache2/modules/libphp7.so
#1  0x00007fd18b11c6e3 in dd_should_trace_helper.isra.6.part () from /opt/datadog-php/extensions/ddtrace-20180731.so
#2  0x00007fd18b11c930 in dd_should_trace_call () from /opt/datadog-php/extensions/ddtrace-20180731.so
#3  0x00007fd18b11dec9 in zif_ddtrace_internal_function_handler () from /opt/datadog-php/extensions/ddtrace-20180731.so
#4  0x00007fd18b5caacd in xdebug_execute_internal (current_execute_data=0x7fd18b81d4e0, return_value=0x7ffe4a4328a8) at /tmp/pear/temp/xdebug/src/base/base.c:464
#5  0x00007fd19442a1bc in ?? () from /usr/lib/apache2/modules/libphp7.so
#6  0x00007fd19480c66b in execute_ex () from /usr/lib/apache2/modules/libphp7.so
#7  0x00007fd18b5ca37d in xdebug_execute_ex (execute_data=0x7fd18b81d460) at /tmp/pear/temp/xdebug/src/base/base.c:380
#8  0x00007fd19442a149 in ?? () from /usr/lib/apache2/modules/libphp7.so
#9  0x00007fd19480c66b in execute_ex () from /usr/lib/apache2/modules/libphp7.so
#10 0x00007fd18b5ca37d in xdebug_execute_ex (execute_data=0x7fd18b81d3c0) at /tmp/pear/temp/xdebug/src/base/base.c:380
[...]

PR #1022, opened 2020-09-18
Fix sigsegv in deferred loading

morrisonlevi:
### Description

A user encountered this in #1021. In the case the loading function couldn't be found the `zval retval` would be un-initialized and then sigsegv would occur when trying to `zval_ptr_dtor` it. I initialized the local variable, but also changed `ddtrace_call_function` to initialize the retval like `zend_call_function` would, which contributed to the bug occurring.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/ext/php7/engine_hooks.c
+++ b/src/ext/php7/engine_hooks.c
@@ -170,7 +170,8 @@ static bool dd_should_trace_helper(zend_execute_data *call, zend_function *fbc, 
                                    scope ? "::" : "", Z_STRVAL(fname));
             }
 
-            zval retval, *integration = &dispatch->deferred_load_integration_name;
+            zval retval = {.u1.type_info = IS_UNDEF};
+            zval *integration = &dispatch->deferred_load_integration_name;
             zend_function **fn_proxy = &dd_integrations_load_deferred_integration;
             ddtrace_string loader = DDTRACE_STRING_LITERAL("ddtrace\\integrations\\load_deferred_integration");
             ZEND_RESULT_CODE status = ddtrace_call_function(fn_proxy, loader.ptr, loader.len, &retval, 1, integration);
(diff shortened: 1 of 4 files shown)


src/ext/php7/engine_hooks.c
```diff
@@ -170,7 +170,8 @@ static bool dd_should_trace_helper(zend_execute_data *call, zend_function *fbc,
                                    scope ? "::" : "", Z_STRVAL(fname));
             }
 
-            zval retval, *integration = &dispatch->deferred_load_integration_name;
+            zval retval = {.u1.type_info = IS_UNDEF};
```
SammyK, line 173: Should we make a similar change in [handlers_curl.c](https://github.com/DataDog/dd-trace-php/blob/9a38904447dcd4ccb913194c30776726a84180de/src/ext/php7/handlers_curl.c#L269-L272)?

tests/ext/sandbox/deferred_load_missing_load_fn.phpt
```diff
@@ -0,0 +1,45 @@
+--TEST--
+deferred loading doesn't trigger nor crash if DDTrace\Integrations\load_deferred_integration is missing
```
SammyK, line 2: Would running a variant of the `distributed_tracing_curl.phpt` test without including the [fake injector](https://github.com/DataDog/dd-trace-php/blob/master/tests/ext/sandbox/distributed_tracing_curl.phpt#L8) test for a similar class of issues?

SammyK (commented):
Nice patch! I have two questions about cURL.

src/ext/php7/engine_hooks.c
morrisonlevi: That case won't segsigv without this PR because it doesn't use the fn_proxy. With this PR if it did use the fn_proxy it would still not sigsegv because I changed ddtrace_call_function to initilaize the retval like zend_call_function would.

So... yes, we probably should, but it's not an issue presently.

tests/ext/sandbox/deferred_load_missing_load_fn.phpt
morrisonlevi: Theoretically, yes it would!

SammyK (approved):

PR #1026, opened 2020-09-22
Compile src/DDTrace/version.php into _generated.php to avoid loading it at runtime

labbati:
### Description

Previously we relied on including `src/DDTrace/version.php` in order to extract the version number.
This clash with our goal to not depend on `src` folder in production.
We now compile the `version.php` file as part of the compilation step and we introduce a  new `ddtrace_src_version()` function to return the version as seen by the code.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ Manual tests were performed
    - `./tooling/bin/pecl-build` produces `x.y.z` for regular versions and `0.0.0` for nightly.
    - `dd-doctor.php` correctly works and version is read from the constant defined in generated file. `ddtrace version (userland)                         [1.0.0-nightly]`
    -  Verified in `Makefile` value of `VERSION` is correct. Both `1.0.0-nightly` and `1.2.3`.

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
I like the idea of not including the file at runtime. My two main thoughts about it are:

  1. It should be a constant, not a function. `DDTrace\VERSION`, perhaps?
  2. Does this mess up composer installs?

src/DDTrace/version.php
```diff
@@ -1,5 +1,13 @@
...
+ * Returns the tracer sources versions.
+ *
+ * @return string
+ */
+function ddtrace_src_version()
+{
+    // Must begin with a number for Debian packaging requirements
+    // Must use single-quotes for packaging script to work
+    return '1.0.0-nightly'; // Update Tracer::VERSION too
+}
```
SammyK, line 13: I like the idea of adding this to the generated file!

I'm with Levi on making the version a constant in the `DDTrace` namespace. And we can still `return` the version number from the file so that we don't have to change the build tools/dd-doctor. :)

labbati: > 'm with Levi on making the version a constant in the DDTrace namespace

Like it!

> And we can still return the version number from the file so that we don't have to change the build tools/dd-doctor. :)

I think this is not possible if we add it to generated file, as a `return` statement would cause the script to, well, return :). We could put it as the last imported file, but to be honest I'd prefer to remove one potential source of troubles given that the price to pay is limited. What do you think?

labbati:
> 2\. Does this mess up composer installs?

Any idea of the specific use case? `Tracer` is not part of `src/api`, so it is not present if the extension is not installed.

src/DDTrace/version.php
SammyK: Ah, good point. :) Sounds good to me! 👍

SammyK (approved):
Pending @morrisonlevi's concern about Composer, this LGTM! 👍

labbati:
In order to very compatibility with composer, I took a better look at what's going on, and my verdict is....I removed `src/DDTrace/version.php` altogether.

That file is never used by composer as (at least with non obsolete versions) it was ALWAYS loaded from `/opt/...`.

Moreover, we do no ship `src/DDTrace/...` as part of composer autoloader, so I do not see a way that the removal of that file can cause issues in composer. I added a line to the composer test, in any case, to make sure that the version can be accessed.

At this point, out of scope for this, we should have only one source of truth, I believe, which is the version in the ext.

Makefile
```diff
@@ -6,8 +6,7 @@ WALL_FLAGS := -Wall -Wextra
 CFLAGS := -O2 $(WALL_FLAGS)
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
-VERSION:=$(shell cat src/DDTrace/version.php | grep return | awk '{print $$2}' | cut -d\' -f2)
-VERSION_WITHOUT_SUFFIX:=$(shell cat src/DDTrace/version.php | grep return | awk '{print $$2}' | cut -d\' -f2 | cut -d- -f1)
+VERSION:=$(shell cat src/DDTrace/Tracer.php | grep 'const VERSION' | awk '{print $$4}' | cut -d\' -f2)
```
morrisonlevi, line 9: Sorry, personal pet-peeve triggered. Why use 4 tools when just `awk` will do it?

```suggestion
VERSION:=$(shell awk -F\' '/const VERSION/ {print $$2}' < src/DDTrace/Tracer.php)
```

playground/index.php
```diff
@@ -1,4 +1,8 @@
 <?php
 
+use DDTrace\Tracer;
+
 error_log("Printed to the error log");
 echo "This is the playground!\n";
+
+echo "Tracer version: " . Tracer::version() . "\n";
```
morrisonlevi, line 8: Should we wrap this in a `class_exists` check in case we want to test without a request_init_hook or disabled ddtrace or something?

```suggestion
if (class_exists('DDTrace\Tracer', $autoload = false)) {
    echo "Tracer version: " . Tracer::version() . "\n";
}
```

src/DDTrace/version.php
```diff
@@ -1,5 +0,0 @@
-<?php
-
-// Must begin with a number for Debian packaging requirements
-// Must use single-quotes for packaging script to work
```
morrisonlevi, line 4: We should preserve these comments on `DDTrace\Tracer::version()`, I think.

tests/Composer/app/index.php
```diff
@@ -39,4 +40,11 @@
 Logger::set(new ErrorLogLogger('debug'));
 Logger::get()->debug('some-debug-message');
 
+/*
+ * Accessing tracer version, which was loaded from a physical file until 0.48.3.
+ */
```
morrisonlevi, line 45: Style nit suggestion:

```suggestion
// Accessing tracer version, which was loaded from a physical file until 0.48.3.
```

tooling/bin/pecl-build
```diff
@@ -6,7 +6,8 @@ rm -f ./bridge/_generated.php
 composer compile
 
 # PECL doesn't like the "-nightly" part of the nightly version number so we have to change it
-dd_version=$(php -r "echo str_replace('1.0.0-nightly', '0.0.0', require 'src/DDTrace/version.php');")
+dd_version=$(cat src/DDTrace/Tracer.php | grep 'const VERSION' | awk '{print $$4}' | cut -d\' -f2 | sed 's/-nightly//g')
```
morrisonlevi, line 9: Awk can do all of this.

```suggestion
dd_version=$(awk -F\' '/const VERSION/ {sub(/-nightly/, "", $2); print $2}' < src/DDTrace/Tracer.php)
```

morrisonlevi (changes requested):
Luca, I left some suggestions. Aside from the ones mentioned, this branch doesn't have `build` and so we don't run the `build_packages` CI workflow, which could be affected by these changes.

Makefile
labbati: Have not idea how `awk` works but I tested it locally and it works.....so 👍  😄

playground/index.php
labbati: TIL about `$autoload = false`

src/DDTrace/version.php
labbati: Good idea, and I think that repetition helps here. Let me add all over the places where we are defining versions.

tooling/bin/pecl-build
labbati: Same as above...have no idea what's in it, but it works 😄

SammyK (approved):
Thanks for digging into the Composer concern. Nice solution @labbati! 💯

morrisonlevi (approved):

PR #1027, opened 2020-09-22
Build on PHP 8.0.0 beta4

SammyK:
### Description

This PR gets ddtrace building on PHP 8.0.0 beta4 but does not include refactoring onto the new [observer API](https://github.com/php/php-src/blob/master/Zend/zend_observer.h) yet (with the exception of `zend_observer_error_register`).

- The PHPUnit tests won't run because of the old version of PHPUnit we require for PHP 5.4 support. This will be addressed in a separate PR.
- There will be a follow up PR with better PHP 8 separation that includes the latest `ZEND_API` changes in PHP 8 RC 1 and will include a shippable build for PHP 8 RC 1. That PR will also not include observer API support yet.
- Full observer API support will come over the course of several PR's that target each release candidate after RC 1.

This PR also adds three new Buster containers:

0. `datadog/dd-trace-ci:buster`: A base Buster container with all the required dependencies to install PHP, but without any PHP version installed. This can be handy when needing to install PHP from a managed repo for debugging. It is also used as the base image for the other PHP versions to make for smaller Dockerfiles.
1. `datadog/dd-trace-ci:php-8.0_buster`: A build of PHP 8.0.0 beta4 in three flavors: `8.0`, `8.0-debug`, and `8.0-debug-zts-asan`. This will be updated to RC 1 in a follow up PR.
2. `datadog/dd-trace-ci:php-master_buster`: A build of php-src master. To pull the latest changes and make a fresh build, run `install-php-master`. In a follow up PR, this will be added to CI to ensure ddtrace is always able to build on php-src master.

### Reviewer checklist
- [x] Milestone is set.
- ~[ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.~ Not going to add this one to the change log since it does not include a shippable PHP 8 build.


src/ext/engine_hooks.h
```diff
@@ -77,6 +82,9 @@ inline void ddtrace_restore_error_handling(ddtrace_error_handling *eh TSRMLS_DC)
     PG(last_error_lineno) = eh->lineno;
     EG(error_reporting) = eh->error_reporting;
 }
+#else
+void ddtrace_restore_error_handling(ddtrace_error_handling *eh);
```
morrisonlevi, line 86: Did this need to not be inline for some reason?

src/ext/php7/engine_api.c
```diff
@@ -54,12 +54,14 @@ ZEND_RESULT_CODE ddtrace_call_method(zend_object *obj, zend_class_entry *ce, zen
...
-         * BUT if you use by-ref args, you need to free the ref if it gets separated!
-         */
-        .no_separation = 0,
         .param_count = argc,
     };
+#if PHP_VERSION_ID < 80000
+    /* Must be 0 to allow for by-ref args
+     * BUT if you use by-ref args, you need to free the ref if it gets separated!
+     */
+    fci.no_separation = 0;
```
morrisonlevi, line 63: Removal of `no_separation` means what for pass-by-ref args?

src/ext/engine_hooks.h
SammyK: Unfortunately this now has a call to `zend_string_release` which [is static](https://heap.space/xref/php-src/Zend/zend_string.h?r=573ad182#296) and causes a compiler warning about calling a static function from an inline function. This was my workaround to fix the warning, but if you have a better suggestion, I'm quite happy to accommodate! :)

src/ext/php7/engine_api.c
SammyK: Good question. I'm not sure the impact this will have at the moment. But according to the [upgrading doc](https://github.com/php/php-src/blob/3e33e1e/UPGRADING.INTERNALS#L144-L148):

> The zend_fcall_info no_separation flag has been removed, and separation is never allowed. If you wish to pass (or allow passing) arguments by reference, explicitly create those arguments as references using ZEND_MAKE_REF.

So this will require more investigation.

src/ext/engine_hooks.h
```diff
@@ -165,4 +173,36 @@ void ddtrace_close_all_open_spans(TSRMLS_D);
...
+    zend_read_property_ex(ddtrace_get_exception_base(object), (object), ZSTR_KNOWN(id), 1, &rv)
+#endif
+#endif
+
+#if PHP_VERSION_ID < 70000
+inline zval *ddtrace_exception_get_entry(zval *object, char *name, int name_len TSRMLS_DC) {
+    zend_class_entry *exception_ce = zend_exception_get_default(TSRMLS_C);
+    return zend_read_property(exception_ce, object, name, name_len, 1 TSRMLS_CC);
+}
+#endif
```
SammyK, line 206: This is... less than ideal. 😖 🐢 But the idea is that we will split `engine_hooks.h` for each PHP version in the next PR.

src/ext/ddtrace.h
```diff
@@ -107,21 +107,21 @@ ZEND_END_MODULE_GLOBALS(ddtrace)
  */
 #if PHP_VERSION_ID < 70000
 #define DDTRACE_ARG_INFO_SIZE(arg_info) ((zend_uint)(sizeof(arg_info) / sizeof(struct _zend_arg_info) - 1))
-#elif PHP_VERSION_ID < 80000
+#elif PHP_VERSION_ID < 90000
```
morrisonlevi, line 110: Maybe `80100` instead?
```suggestion
#elif PHP_VERSION_ID < 80100
```

src/ext/ddtrace.h
```diff
@@ -107,21 +107,21 @@ ZEND_END_MODULE_GLOBALS(ddtrace)
...
 #define DDTRACE_ARG_INFO_SIZE(arg_info) ((uint32_t)(sizeof(arg_info) / sizeof(struct _zend_internal_arg_info) - 1))
 #else
-#error Check if ZEND_FENTRY has changed in PHP 8 and if we need to update the macros
+#error Check if ZEND_FENTRY has changed in PHP 9 and if we need to update the macros
```
morrisonlevi, line 113: And again here:
```suggestion
#error Check if ZEND_FENTRY has changed in PHP 8.1 and if we need to update the macros
```

tests/ext/sandbox/exception_from_user_error_handler_internal.phpt
```diff
@@ -4,7 +4,7 @@ Exceptions from user error handler are tracked for instrumented internal functio
 <?php if (PHP_VERSION_ID < 50500) die('skip: PHP 5.4 not supported'); ?>
 <?php if (PHP_VERSION_ID < 70000) die('skip: Unaltered VM dispatch required for handling return value on PHP 5'); ?>
```
morrisonlevi, line 5: Do you think PHP 5 can handle this now?

SammyK: Yes. Updated. :)

Makefile
```diff
@@ -8,7 +8,7 @@ ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 
 VERSION:=$(shell awk -F\' '/const VERSION/ {print $$2}' < src/DDTrace/Tracer.php)
 
-INI_FILE := /usr/local/etc/php/conf.d/ddtrace.ini
+INI_FILE := $(shell php -i | grep 'Scan' | sed -e 's/Scan this dir for additional .ini files =>//g' | head -n 1 | awk '{print $$1}')/ddtrace.ini
```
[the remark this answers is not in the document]
SammyK: That is cleaner - thanks! And good point about `(none)/ddtrace.ini`. Since the Makefile is used for our containers only, do you think we need to come up with a solution to the `(none)` issue? Or do you think it's OK to leave as is?

src/ext/php7/engine_api.c
SammyK: So it looks like we might need to add `ZEND_MAKE_REF` to a few places to ensure this works as expected, but since none of the pass-by-reference tests failed, I'll need to dig a bit deeper. Would you be OK if we addressed this in the followup PR?

Makefile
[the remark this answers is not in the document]
morrisonlevi: I imagine users that build from source may use the `Makefile` too. So maybe we need to address it?

src/ext/php7/engine_api.c
morrisonlevi: IIRC the curl helper functions take an array by reference that might get separated. So perhaps dig into that specific case a little bit? As long as we understand what's going on we can delay to another PR, but I don't want to land something in master that is pretty suspect like this.

Makefile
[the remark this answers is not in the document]
SammyK: Building from source with PECL or manually will replace the Makefile when `phpize` is run so they won't encounter this issue.

src/ext/php7/engine_api.c
SammyK: I did a bit more digging and found that we have more blockers to do a proper investigation on the impact of this. The only place `ddtrace_call_{method|function}` is called with args is isolated to the curl handler which doesn't run on PHP 8 yet because it [needs to be upgraded](https://github.com/DataDog/dd-trace-php/blob/ba542dd/src/ext/php7/handlers_curl.c#L297) to use `CurlHandler` instead of a resource. Then we'll need to validate that the args for `GlobalTracer::get()->inject()` are handled properly.

Since this is blocked, are you OK with this landing in master now that we have a more specific plan of action to investigate this further?

morrisonlevi: Sounds good to me.

morrisonlevi (approved):

PR #767, opened 2020-02-24
Track fatal errors

SammyK:
### Description

WIP

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.
- [ ] Tests added for this feature/bug.

### Reviewer checklist
- [x] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the appropriate release draft. For community contributors the reviewer is in charge of this task.

PR #1006, opened 2020-09-03
Defer loading of PDO, Memcached, Slim, WordPress, and Yii integrations until first usage

labbati:
### Description

This PR defers the loading of PDO, Memcached, Slim, WordPress, and Yii integrations until they are used.

Cleans up Yii2 and WordPress a tiny bit:
- Remove unnecessary singletons
- Use non-tracing API for some hooks

### Readiness checklist
- [x] (only for Members) Changelog has been added to the release document.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1028, opened 2020-09-24
Close open spans on fatal errors

morrisonlevi:
### Description

This replaces PR #767: Track fatal errors.

When a fatal error is encountered the internal span stack will now be closed before flushing the trace. This targets PHP 5.5+. Additionally, if there are any open internal spans then one of them will have the fatal error information attached to it. In the future we plan to make it attach to the root span as well.

It also changes the format of reported stack traces for errors/exceptions slightly. Previously a stack trace might look like this, where the function `inner` had some arguments :

```
#0 %s: inner(...)
#1 %s: outer()
#2 {main}
```

The `...` has been removed in this PR:

```
#0 %s: inner()
#1 %s: outer()
#2 {main}
```

Note to reviewers: I verified that PHP 5.5 and PHP 5.6 leak certain bits of VM memory on fatal errors, which is why there are so  new suppressions.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/ext/php5/serializer.c
```diff
@@ -257,18 +257,79 @@ static void _serialize_exception(zval *el, zval *meta, ddtrace_exception_t *exce
...
-     * already made a copy, or it will when it duplicates with dup param, so
-     * if it did it should still be safe. */
-    add_assoc_stringl(meta, "error.type", (char *)class_name, class_name_len, needs_copied);
     add_assoc_zval(meta, "error.msg", msg);
 
+    bool use_class_name_for_error_type = true;
+    if (instanceof_function(Z_OBJCE_P(exception), ddtrace_ce_fatal_error TSRMLS_CC)) {
+        zval *code = NULL;
+        zend_call_method_with_0_params(&exception, Z_OBJCE_P(exception), NULL, "getcode", &code);
+        if (Z_TYPE_P(code) == IS_LONG) {
```
SammyK, line 269: Can `code` still be `NULL` after `zend_call_method_with_0_params`?

src/ext/php5_4/engine_hooks.c
```diff
@@ -670,3 +670,17 @@ void ddtrace_execute_internal_mshutdown(void) {
...
+    PHP5_UNUSED(DDTRACE_ERROR_CB_PARAM_PASSTHRU TSRMLS_CC);
+
+    return NULL;
+}
+
+void ddtrace_close_all_open_spans(TSRMLS_D) {
+#if ZTS
+    PHP5_UNUSED(TSRMLS_C);
+#endif
+    ddtrace_log_debug("Request to close all open spans ignored; not supported on PHP 5.4 (yet, anyway)");
```
SammyK, line 685: Should we still free the span stacks at this point?

src/ext/php7/serializer.c
```diff
@@ -222,8 +221,8 @@ static void _serialize_stack_trace(zval *meta, zval *trace) {
     add_assoc_zval(meta, "error.stack", &output);
 }
 
-static void _serialize_exception(zval *el, zval *meta, zend_object *exception_obj) {
-    zval exception, name, msg, stack;
+static void dd_serialize_exception(zval *el, zval *meta, zend_object *exception_obj) {
+    zval exception, name, msg, code, stack;
```
SammyK, line 225: Should we perhaps initialize `code` to undef? Or perhaps all of the zval's that are passed into `zend_call_method_with_0_params` just in case? :)

src/ext/php5/serializer.c
```diff
@@ -257,18 +257,79 @@ static void _serialize_exception(zval *el, zval *meta, ddtrace_exception_t *exce
...
+                case E_ERROR:
+                    error_type = DDTRACE_STRING_LITERAL("E_ERROR");
+                    break;
+                case E_WARNING:
+                    error_type = DDTRACE_STRING_LITERAL("E_WARNING");
+                    break;
+                case E_PARSE:
+                    error_type = DDTRACE_STRING_LITERAL("E_PARSE");
+                    break;
+                case E_NOTICE:
```
SammyK, line 281: If `DDTrace\FatalError` is only created for `E_ERROR`, `E_CORE_ERROR`, and `E_USER_ERROR`, do we need all the other error codes here?

SammyK (commented):
Nice work @morrisonlevi! I left some questions. :)

src/ext/ddtrace.c
```diff
@@ -1523,6 +1615,7 @@ static const zend_function_entry ddtrace_functions[] = {
     DDTRACE_SUB_NS_FE("Config\\", integration_analytics_enabled, arginfo_ddtrace_config_integration_analytics_enabled),
     DDTRACE_SUB_NS_FE("Config\\", integration_analytics_sample_rate,
                       arginfo_ddtrace_config_integration_analytics_sample_rate),
+    DDTRACE_SUB_NS_FE("Testing\\", trigger_error, arginfo_ddtrace_testing_trigger_error),
```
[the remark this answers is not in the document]
morrisonlevi: I think a runtime variable makes more sense if we do anything here. We want to run tests using the official packages, which wouldn't be built with such compile-time flags.

Currently `_DD_LOAD_TEST_INTEGRATIONS` is in use by `_dd_load_test_integrations` (which took me a bit to find because I reworked all of that in the [pending deferred loading PR](https://github.com/DataDog/dd-trace-php/pull/1006)).

Any other thoughts based on this info?

src/ext/php5_4/engine_hooks.c
morrisonlevi: I don't think so; the C stack will be holding onto variables that would be invalidated if we did that.

src/ext/engine_hooks.h
```diff
@@ -151,4 +151,18 @@ inline void ddtrace_sandbox_end(ddtrace_sandbox_backup *backup TSRMLS_DC) {
...
+
+#define DDTRACE_ERROR_CB_PARAM_PASSTHRU type, error_filename, error_lineno, format, args
+
+void ddtrace_error_cb(DDTRACE_ERROR_CB_PARAMETERS);
+ddtrace_exception_t *ddtrace_make_exception_from_error(DDTRACE_ERROR_CB_PARAMETERS TSRMLS_DC);
+
+#if PHP_VERSION_ID < 70000
+void ddtrace_close_all_open_spans(TSRMLS_D);
+#else
+void ddtrace_close_all_open_spans(void);
```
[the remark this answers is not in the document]
morrisonlevi: I don't like supporting `TSRMLS_*` on PHP 8, which also takes the `else` branch. I understand doing it in the short-term, but at least at this stage I'd prefer not to redefine things from older versions into newer versions. What do you think?

src/ext/php5/serializer.c
morrisonlevi: I'll have to check into this, especially since this is PHP 5.

src/ext/php5/serializer.c
morrisonlevi: I suppose not; was just trying to be robust and helpful. For instance, we use `"{unknown error}"` by default if it doesn't match anything, but it's a bit disingenuous if it's `E_WARNING`, `E_NOTICE`, etc.

src/ext/php7/serializer.c
morrisonlevi: On one hand this gives us better safety if something does go wrong, but if it's unnecessary then tools like ASAN are less helpful if you initialize it to something like IS_UNDEF.

I'll look into `zend_call_method_with_0_params` and company to see if we're outright having bugs, of course.

src/ext/ddtrace.c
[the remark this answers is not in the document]
SammyK: Good point about official builds. I can't think of another alternative that makes sense so we can just keep it as is. :)

src/ext/php5/serializer.c
morrisonlevi: It's not obvious; we had better be safe. Good catch, Sammy.

src/ext/engine_hooks.h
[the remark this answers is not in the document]
SammyK: I'm with you on not supporting TSRMLS macros on PHP 8 (or 7 for that matter). :) Ultimately we'll need to separate out `engine_hooks.h` by version. It's starting to get really hairy in #1027. So for now we can just leave this as-is so we can figure out a longer term solution with a follow up PHP 8 PR.

src/ext/php5/serializer.c
SammyK: I like the idea of being helpful, but to me it's a bit confusing to support non-fatal errors when serializing `DDTrace\FatalError`. Perhaps a `ZEND_ASSERT(error_type == ...)` would help communicate the specific fatal errors that are supported here?

morrisonlevi: Let me think about it a bit more. Assertion failures are hard failures but they are only present in debug builds.

SammyK: Yes, a hard failure would be ideal in this case IMO. If an `E_NOTICE` made its way into a `DDTrace\FatalError`, something is really broken. ;)

morrisonlevi: I think the best way to do this is to drop the ones that aren't fatal, then doing something in the default branch. What do you think?

SammyK: Perfect 👍

SammyK (approved):
Nice work @morrisonlevi! 💯

src/ext/php7/serializer.c
morrisonlevi: For posterity: on PHP 7 `zend_call_function` sets the retval to IS_UNDEF; `zend_call_method` doesn't do anything with it prior. This means we probably _shouldn't_ initialize it, but it's difficult to know when we should or shouldn't initialize a variable. Maybe we can look at linters that support custom rules to help us get it right.

PR #1029, opened 2020-09-28
Properly group traces sent in batch during long running scripts

labbati:
### Description

By default the background sender enqueue traces and sends them every 5 seconds (or when necessary if they are piling up). Spans are grouped together (and traces counted) based on a group id that was reset in the RINIT. Problem is: for long running scripts RINIT is run only once and we were sending all traces correctly, but header `X-Datadog-Trace-Count: 1` caused only the first trace (every 5 seconds) to be decoded.

This PR add state reset after traces are sent to the background sender and the script is ready to handle a new trace.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -1262,7 +1262,7 @@ static PHP_FUNCTION(dd_trace_send_traces_via_thread) {
     }
 
     bool result = ddtrace_send_traces_via_thread(num_traces, curl_headers, payload, payload_len TSRMLS_CC);
-    _ddtrace_prepare_for_new_trace();
+    _ddtrace_prepare_for_new_trace(TSRMLS_C);
     RETURN_BOOL(result);
 }
 
--- a/src/ext/ddtrace.c
+++ b/src/ext/ddtrace.c
@@ -89,6 +89,9 @@ static void ddtrace_shutdown(struct _zend_extension *extension) {
 static void ddtrace_activate(void) {}
 static void ddtrace_deactivate(void) {}
 
+// prepare the tracer state to start handling a new trace
+void _ddtrace_prepare_for_new_trace();
+
 static zend_extension _dd_zend_extension_entry = {"ddtrace",
                                                   PHP_DDTRACE_VERSION,
                                                   "Datadog",
@@ -377,7 +380,7 @@ static PHP_RINIT_FUNCTION(ddtrace) {
     // Reset compile time after request init hook has compiled
     ddtrace_compile_time_reset(TSRMLS_C);
 
-    DDTRACE_G(traces_group_id) = ddtrace_coms_next_group_id();
+    _ddtrace_prepare_for_new_trace();
 
     return SUCCESS;
 }
@@ -1259,7 +1262,7 @@ static PHP_FUNCTION(dd_trace_send_traces_via_thread) {
     }
 
     bool result = ddtrace_send_traces_via_thread(num_traces, curl_headers, payload, payload_len TSRMLS_CC);
-    DDTRACE_G(traces_group_id) = ddtrace_coms_next_group_id();
+    _ddtrace_prepare_for_new_trace();
     RETURN_BOOL(result);
 }
 
@@ -1548,3 +1551,9 @@ ZEND_GET_MODULE(ddtrace)
 ZEND_TSRMLS_CACHE_DEFINE();
 #endif
 #endif
+
+// the following operations are performed in order to put the tracer in a state when a new trace can be started:
+//   - set a new trace (group) id
+void _ddtrace_prepare_for_new_trace() {
+    DDTRACE_G(traces_group_id) = ddtrace_coms_next_group_id();
+}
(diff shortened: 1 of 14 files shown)


tests/Integration/LongRunning/long_running_script_manual.php
```diff
@@ -0,0 +1,29 @@
...
+    $rootSpan = $tracer->startActiveSpan("root-operation")->getSpan();
+    $rootSpan->setTag(Tag::SERVICE_NAME, "long-running-service");
+    $rootSpan->setTag(Tag::RESOURCE_NAME, "root-resource");
+
+    $subSpan = $tracer->startActiveSpan('sub-operation')->getSpan();
+    $rootSpan->setTag(Tag::RESOURCE_NAME, "sub-resource");
+    $subSpan->finish();
+    $rootSpan->finish();
+
+    if (PHP_MAJOR_VERSION === 5) {
```
labbati, line 18: @SammyK is this OK or we should not recommend (and mention in docs) tracing of long running scripts on PHP 5?

tests/Common/CLITestCase.php
```diff
@@ -1,6 +1,6 @@
 <?php
```
labbati, line 1: This was moved from `tests/Integrations` to `test/Common` in order to be properly used all over the test suites.

src/ext/ddtrace.c
```diff
@@ -89,6 +89,9 @@ static void ddtrace_shutdown(struct _zend_extension *extension) {
 static void ddtrace_activate(void) {}
 static void ddtrace_deactivate(void) {}
 
+// prepare the tracer state to start handling a new trace
+void _ddtrace_prepare_for_new_trace(TSRMLS_D);
```
morrisonlevi, line 93: I suggest making it static since it's only used in this file, and we've been using the prefix `dd_` for such functions lately.

```suggestion
static void dd_prepare_for_new_trace(TSRMLS_D);
```

src/ext/ddtrace.c
```diff
@@ -377,7 +380,7 @@ static PHP_RINIT_FUNCTION(ddtrace) {
     // Reset compile time after request init hook has compiled
     ddtrace_compile_time_reset(TSRMLS_C);
 
-    DDTRACE_G(traces_group_id) = ddtrace_coms_next_group_id();
+    _ddtrace_prepare_for_new_trace(TSRMLS_C);
```
morrisonlevi, line 383: To match my other suggestions:
```suggestion
    dd_prepare_for_new_trace(TSRMLS_C);
```

src/ext/ddtrace.c
```diff
@@ -1254,11 +1257,13 @@ static PHP_FUNCTION(dd_trace_send_traces_via_thread) {
...
     if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "las", &num_traces, &curl_headers,
                                  &payload, &payload_len) == FAILURE) {
-        ddtrace_log_debug("dd_trace_send_traces_via_thread() expects url, http headers, and http body");
+        ddtrace_log_debug("dd_trace_send_traces_via_thread() expects trace count, http headers, and http body");
         RETURN_FALSE
     }
 
-    RETURN_BOOL(ddtrace_send_traces_via_thread(num_traces, curl_headers, payload, payload_len TSRMLS_CC));
+    bool result = ddtrace_send_traces_via_thread(num_traces, curl_headers, payload, payload_len TSRMLS_CC);
+    _ddtrace_prepare_for_new_trace(TSRMLS_C);
```
morrisonlevi, line 1265: To match my other suggestions:
```suggestion
    dd_prepare_for_new_trace(TSRMLS_C);
```

src/ext/ddtrace.c
```diff
@@ -1546,3 +1551,7 @@ ZEND_GET_MODULE(ddtrace)
 ZEND_TSRMLS_CACHE_DEFINE();
 #endif
 #endif
+
+// the following operations are performed in order to put the tracer in a state when a new trace can be started:
+//   - set a new trace (group) id
+void _ddtrace_prepare_for_new_trace(TSRMLS_D) { DDTRACE_G(traces_group_id) = ddtrace_coms_next_group_id(); }
```
morrisonlevi, line 1557: To match my other suggestion; hopefully this +1 char doesn't push it over the line limit :)

```suggestion
static void dd_prepare_for_new_trace(TSRMLS_D) { DDTRACE_G(traces_group_id) = ddtrace_coms_next_group_id(); }
```

tests/Integration/LongRunning/long_running_script_manual.php
SammyK: As long as the userland API is being used (as in this example), this should be fine. We just need to add `$tracer->reset();` after the flush below.

labbati: Updated our card. I am skipping the test on PHP 5 and we are not going to suggest long running scripts on 5 until we properly and extensively test it either manually or on our reliability environment.

morrisonlevi (approved):

PR #1034, opened 2020-10-02
Attach fatal errors to all open internal spans

morrisonlevi:
### Description

Previously fatal errors were attached only to the top-most open internal span. In this case, "internal" means traced by `DDTrace\trace_method` or `DDTrace\trace_function`; it doesn't apply to spans made through Open Tracing or the `DDTrace\Tracer` API which notably includes the root span when `DD_TRACE_GENERATE_ROOT_SPAN` is on (it defaults to on).

Also avoid installing `ddtrace_error_cb` on PHP 5.4 as this feature it isn't supported yet.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


SammyK (approved):
LGTM! 👍

PR #1039, opened 2020-10-08
Enable unit testing for php 8

labbati:
### Description

TBD

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


.circleci/config.yml
```diff
@@ -87,15 +87,10 @@ aliases:
       name: Installing dependencies with composer
       command: composer update --no-interaction
 
-  - &STEP_COMPOSER_COMPILE
+  - &STEP_COMPOSER_GENERATE
```
labbati, line 90: Since now 'compilation' is in the make file, I preferred to turn to 'generation' for php, as compilation in a context closer to C might be confusing, considering what we do. Not a strong opinion though.

.circleci/config.yml
```diff
@@ -804,8 +799,7 @@ jobs:
             - ".git"
       - <<: *STEP_ATTACH_WORKSPACE
       - <<: *STEP_COMPOSER_UPDATE
-      - <<: *STEP_COMPOSER_COMPILE
-      - <<: *STEP_COMPOSER_COMPILE_VERIFY
```
labbati, line 808: Verification is now part of `Makefile`'s `make generate`.

tests/DDTestListener.php
```diff
@@ -1,19 +0,0 @@
-<?php
-
```
labbati, line 2: This functionality was moved to `ddTearDown`.

bridge/autoload.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-if (getenv('DD_AUTOLOAD_NO_COMPILE') === 'true' || !file_exists(__DIR__ . '/_generated.php')) {
+if (getenv('DD_AUTOLOAD_NO_COMPILE') === 'true') {
```
labbati, line 3: I think that rather than attempt sources, we should just return (if we want to do this check). In any case we are testing real package now (in progress), so I believe this should be removed altogether.

tests/Common/BaseTestCase.php
```diff
@@ -0,0 +1,99 @@
...
+     * @return DebugLogger
+     */
+    protected function withDebugLogger()
+    {
+        $logger = new DebugLogger();
+        Logger::set($logger);
+        return $logger;
+    }
+
+    public function composerUpdateScenario($workingDir)
```
SammyK, line 47: It looks like this was moved from the old `BaseTestCase`, but this seems to be called exclusively from `ComposerInteroperabilityTest`, so maybe we could move it there? It also might be good to send `$workingDir` through `escapeshellarg()` just in case. :)

tests/api/bootstrap.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+error_reporting(E_ALL);
+
+if (getenv('DD_AUTOLOAD_NO_COMPILE') == 'true' && (false !== getenv('CI') || false !== getenv('CIRCLECI'))) {
+    throw new Exception('Running uncopiled files in CI');
```
SammyK, line 6: Nit: s/uncopiled/uncompiled/ or perhaps: `Tests must run using the _generated.php script in CI`

tests/bootstrap.php
```diff
@@ -2,6 +2,17 @@
 
 error_reporting(E_ALL);
 
-require __DIR__ . '/bootstrap_utils.php';
+if (getenv('DD_AUTOLOAD_NO_COMPILE') == 'true' && (false !== getenv('CI') || false !== getenv('CIRCLECI'))) {
+    throw new Exception('Running uncopiled files in CI');
```
SammyK, line 6: Same nitpick here. :)

tooling/bin/pecl-build
```diff
@@ -3,7 +3,8 @@
 set -e
 
 rm -f ./bridge/_generated.php
-composer compile
+composer -dtooling/generation generate
+composer -dtooling/generation verify
```
SammyK, line 7: To ensure things don't get out of sync in the future, we could add `composer -dtooling/generation update` above these lines and remove the [`STEP_COMPOSER_UPDATE` step](https://github.com/DataDog/dd-trace-php/blob/1d940f3/.circleci/config.yml#L688) from the CircleCI config since that doesn't update from the correct Composer file for this anymore.

SammyK (commented):
Nice work tackling this issue @labbati! I just left a few bits of feedback.

tooling/bin/pecl-build
labbati: I really like the idea. I tried to think about something that I might not consider making this change, but so far I cannot identify any CON. Change pushed. Let's see if CI stays green :).
Thanks for the idea.

SammyK (approved):
Nice work @labbati! 💯

PR #1040, opened 2020-10-08
Add fatal error info to DDTrace\Tracer made root spans on PHP 7

morrisonlevi:
### Description

When a fatal PHP error triggers this will attach this information to the root span. For now this is PHP 7 only.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/ext/ddtrace.c
```diff
@@ -403,6 +403,9 @@ static PHP_RINIT_FUNCTION(ddtrace) {
         return SUCCESS;
     }
 
+    // currently only use this for error.msg, error.stack, and error.type, so use 3
+    array_init_size(&DDTRACE_G(additional_trace_meta), 3);
```
SammyK, line 407: Nit: Since this magic number is used [here](https://github.com/DataDog/dd-trace-php/blob/5b01859e2bcebc13f8b66f5a444e10660d8c9d86/src/ext/php7/serializer.c#L630), [here](https://github.com/DataDog/dd-trace-php/blob/5b01859e2bcebc13f8b66f5a444e10660d8c9d86/src/ext/php7/serializer.c#L701), and I'm guessing should be used [here too](https://github.com/DataDog/dd-trace-php/blob/5b01859e2bcebc13f8b66f5a444e10660d8c9d86/src/ext/span.c#L169), perhaps we can make it a macro?

tests/Common/SpanAssertion.php
```diff
@@ -366,4 +367,25 @@ public function isToBeSkipped()
...
+    {
+        foreach ($versions as $version) {
+            if (Versions::phpVersionMatches($version)) {
+                return $this;
+            }
+        }
+
+        $callback($this);
+        return $this;
+    }
```
morrisonlevi, line 390: The behavior here for more than one item is a bit fuzzy to me. Is there a reason for supporting anything other than a single element? So far you only have used it with exactly one item.

Otherwise, this PR looks good to me 👍

labbati: good point. My original view was to also add `ifPhpVersionMatches` and use it in conditional assertion, where now we have to define the whole thing some place else and then do `if (php7 | 8) ...`. In this case I had anticipated we would have a few php7 & php 8 vs php5.

But we can always change later or find a different approach. Let me revert it back.

tests/Unit/SpanTest.php
```diff
@@ -160,7 +160,14 @@ public function testSpanLogMessageAddsExpectedTag()
...
 
-        $this->assertFalse($span->hasError());
+        /* I'm not exactly sure what the intended behavior here is, as this
+         * previously did assertFalse($span->hasError).
+         * However, I recently made it so that when ERROR_MSG is set that the
+         * error field will automatically be set to true. This is relied on for
+         * setting the proper fields on the root span, for instance.
+         *  - Levi Morrison, October 2020
+         */
+        $this->assertTrue($span->hasError());
```
labbati, line 170: `log()` is a method inherited from our OpenTracing-like api at the very beginning of this project. Logs are one of the [types of metadata](https://opentracing.io/docs/overview/tags-logs-baggage/#logs).

I think the [behavior in the code](https://github.com/DataDog/dd-trace-php/blob/master/src/DDTrace/Span.php#L369) is wrong as I don't see why a `message` log should be translated into a `error.msg`.

Moreover, logs can be multiple (for different timestamps), so the way it is handled makes no sense to me.

I would suggest to leave this test as `$this->assertFalse($span->hasError());` and mark the test as expected failure.

In the future, we can either remove this functionality and move it to the OpenTracing adapter or to properly handle it.

labbati (changes requested):
I suggest to make this tiny change to the test as per my comment, then I am good to go with this PR.

labbati (approved):
LGTM. Pair reviewed this with @morrisonlevi

PR #1043, opened 2020-10-08
Fix Symfony 4.0 composer update issue in CI

labbati:
### Description

Some release of symfony/flex's recipes or dependencies (I believe around monolog but I was not able to track it down to the rela cause) is causing the `cache:clear` post install to fail with an error `"excluded_http_codes" cannot be used as your version of Monolog bridge does not support it.`. Attempts based on issues in various symfony recipes did not work.
Only trick that worked is running composer update once with `--no-script` (so cache clear is not run) and once without it, so all the post-update scripts required by symfony is executed.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -569,6 +569,7 @@ test_web_symfony_34:
 	php tests/Frameworks/Symfony/Version_3_4/bin/console cache:clear --no-warmup --env=prod
 	$(call run_tests,tests/Integrations/Symfony/V3_4)
 test_web_symfony_40:
+	# Trick to have symfony 4.0 update process not to fail because of an error related to monolog dependencies.
 	$(COMPOSER) --working-dir=tests/Frameworks/Symfony/Version_4_0 update --no-scripts
 	$(COMPOSER) --working-dir=tests/Frameworks/Symfony/Version_4_0 update
 	php tests/Frameworks/Symfony/Version_4_0/bin/console cache:clear --no-warmup --env=prod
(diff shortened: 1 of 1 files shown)


SammyK (approved):
Thanks for figuring out a workaround @labbati. I believe @morrisonlevi is looking into the root cause.

Makefile
```diff
@@ -569,6 +569,8 @@ test_web_symfony_34:
 	php tests/Frameworks/Symfony/Version_3_4/bin/console cache:clear --no-warmup --env=prod
 	$(call run_tests,tests/Integrations/Symfony/V3_4)
 test_web_symfony_40:
+	# Trick to have symfony 4.0 update process not to fail because of an error related to monolog dependencies.
+	$(COMPOSER) --working-dir=tests/Frameworks/Symfony/Version_4_0 update --no-scripts
 	$(COMPOSER) --working-dir=tests/Frameworks/Symfony/Version_4_0 update
```
morrisonlevi, line 574: Locally this still fails for me. The `--no-scripts` line succeeds, then this one fails.

morrisonlevi: [This comment](https://github.com/symfony/recipes/issues/639#issuecomment-647379828) makes it seem like maybe we are doing something wrong with recipes.

labbati:
Merging it for now, until we figure out a better approach

PR #1047, opened 2020-10-09
Update PHP language xfail lists

morrisonlevi:
### Description

Locally I updated PHP 7.2, 7.3, and 7.4 buster containers to the latest versions. All of these tests fail because a specific object ID is tested.

The PHP 5.5 test is one that happens in a 1 second request time limit. If it triggers in a memory routine then it is corrupted and sigsegvs in the next memory operation. This is why newer PHP versions (I think PHP 7.1+?) use a different mechanism for timeout handling.

### Readiness checklist
- [x] ~Changelog has been added to the release document.~

### Reviewer checklist
- [x] Milestone is set.
- [x] ~Changelog has been added to the release document.~


labbati (approved):
LGTM! 👍

PR #1048, opened 2020-10-11
Trace Symfony controller and more

morrisonlevi:
### Description

This adds the following enhancements:

- Trace the Symfony controller as `symfony.controller`, attaching the method name as the resource. Note that this doesn't work with controllers that are closures, only other forms of callables. Also, some versions of Symfony call this with `call_user_func` or `call_user_func_array`, and sometimes it's fully qualified and sometimes it's not. I point this out because on PHP 7 it matters; if it's unqualified then it won't be traced.
- Trace the main HttpKernel\Kernel's methods `boot` and `handle`.
- Move `Symfony\Component\HttpKernel\HttpKernel::__construct` to the non-tracing prehook API.
- Trace public methods of Doctrine Repository classes. This will probably be split off of this PR in its final form, but it was really satisfying to see it in the demo app. In the screenshot below, the spans that start with `App` underneath `symfony.controller App\Controller\BlogController::index` are the ones belonging to Doctrine. Edit: this piece has been removed from this PR.

Screenshot of Symfony demo app with changes from the initial PR

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -338,6 +338,7 @@ TEST_WEB_71 := \
 	test_web_symfony_30 \
 	test_web_symfony_33 \
 	test_web_symfony_34 \
+	test_web_symfony_40 \
 	test_web_symfony_42 \
 	test_web_yii_2 \
 	test_web_wordpress_48 \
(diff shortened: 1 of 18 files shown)


labbati (approved):
LGTM! Thanks Levi

PR #1058, opened 2020-10-23
Add tests and support for WP5

labbati:
### Description

Add tests and support for WP5:
 - we add a test app for WP5
 - we add tests for commons scenarios for WP5 (including not found)
 - we remove the check for wp4 only and now we use the very same integration for both WP4 and WP5

**Note to the reviewer**: I hid all wordpress files via `.gitattributes`. For some reason (possibly because it is the last one of thousands) a file requiring CR is hidden as well `tests/Integrations/WordPress/V5_5/CommonScenariosTest.php`.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (dismissed):
This looks good to me. Maybe Sammy can review it too since he has more experience with WordPress integrations than me?

Makefile
```diff
@@ -440,6 +445,7 @@ TEST_WEB_74 := \
 	test_web_symfony_42 \
 	test_web_symfony_44 \
 	test_web_wordpress_48 \
+	test_web_wordpress_55 \
```
SammyK, line 448: Just to confirm: does WordPress 5.5 not work on PHP 8?

tests/Integrations/WordPress/V5_5/CommonScenariosTest.php
```diff
@@ -0,0 +1,395 @@
...
+                    SpanAssertion::build(
+                        'wordpress.request',
+                        'wordpress_55_test_app',
+                        'web',
+                        'GET /error'
+                    )->withExactTags([
+                        'http.method' => 'GET',
+                        'http.url' => 'http://localhost:9999/error',
+                        // WordPress doesn't appear to automatically set the proper error code
+                        'http.status_code' => '200',
```
SammyK, line 244: This one is not a blocker, but something to think about for a separate PR. We should probably report 500 for uncaught exceptions even if WP doesn't. Otherwise they will look like normal traces in the UI (especially since the exception isn't attached to the root span yet). 

IIRC we discussed this idea about another framework, but I can't remember which one. We might want to come up with a more generic solution that will work across all frameworks, so I'll start an internal discussion so we can iron out some details. :)

SammyK (dismissed):
LGTM! 👍 If WP 5.5 doesn't support PHP 8, this one should be good to go. :)

tests/Integrations/WordPress/V5_5/CommonScenariosTest.php
labbati: Good idea to start this conversation, I am unsure I fully agree.

> We should probably report 500 for uncaught exceptions even if WP doesn't.

I think we should report what the user sees. If the app returns 200, then we should show 200. Otherwise we give a misleading info. E.g. a developer running `curl <url> -v` would see a `500` trace in the UI and a `200` response dumped by curl. Still, they should look different...read on :)

> Otherwise they will look like normal traces in the UI (especially since the exception isn't attached to the root span yet).

This is a very good point, though. Even if we return `200`, we can still (and should) mark the trace as an error. This, though, should not be the framework's integration responsibility, and it will be provided by the work that we are doing for generic error reporting improvements.

WDYT?

Makefile
labbati: It does not work as it is (unexpected status codes in case of missing and missing spans). It seems that the [version expected to be working with PHP 8 is 5.6](https://make.wordpress.org/core/2020/10/06/call-for-testing-php-8-0/) .

tests/Integrations/WordPress/V5_5/CommonScenariosTest.php
SammyK: Good point - sounds good to me. 👍

Makefile
SammyK: Perfect. Thanks for looking into this.

SammyK (approved):
Nice work @labbati 👍

PR #1068, opened 2020-10-31
Add DatadogArena and tests

morrisonlevi:
### Description

This adds an arena allocator `datadog_arena` similar to PHP's `zend_arena`. Critically it is not tied to _any_ PHPisms at all so it can be used across all PHP versions. This allocator doesn't support deallocating individual objects; it frees all its memory at once. There is also a checkpoint/restore API so you can save the current allocator position and then rewind to the same spot (caller is responsible for making sure conflicting checkpoint/restores nor frees happen between checkpoint/restore).

In particular arena allocators are good for serializing variable-length data which is then transferred and freed. This pattern comes up in both the tracer and the profiler.

Tests are written using the Catch2 library and uses the CMake build system. The arena simple enough that it can be integrated directly using whatever build system we want; if we don't count tests it's just two files, `arena.c` and `datadog/arena.h`.

------

One thing I encountered while writing this is that in buster containers we don't have write access to `/opt` as the `circleci` user, which means we must use `sudo` there. However, `sudo` doesn't exist in the CentOS containers. This is annoying and I plan to push changes to the buster containers soon that add ownership of `/opt` to `circleci`, but we should also probably get `sudo` + `circleci` into the CentOS containers too. Without these changes it makes it difficult to write scripts or commands that work across the different containers.

Another thing is that `eclint` should probably be an inclusion list instead of an exclusion list -- at this point it seems we exclude far more directories than we lint...

### Readiness checklist
- [x] ~Changelog has been added to the release document.~ This doesn't actually get used anywhere yet -- just the library and tests.

### Reviewer checklist
- [ ] Milestone is set.
- [x] ~Changelog has been added to the release document.~

### DIFF
--- a/ext/DatadogArena/arena.c
+++ b/ext/DatadogArena/arena.c
@@ -1,0 +1,59 @@
+#include "datadog/arena.h"
+
+#include <stdlib.h>
+
+_Static_assert(sizeof(size_t) == 8, "Unexpected size_t size");
+
+extern inline char *datadog_arena_alloc(datadog_arena **arena_ptr, size_t size);
+extern inline bool datadog_arena_try_alloc(datadog_arena *arena, size_t size, char **result);
+
+/* prefer powers of 2 */
+datadog_arena *datadog_arena_create(size_t size) {
+    datadog_arena *arena = (datadog_arena *)calloc(1, size);
+
+    size_t aligned = DATADOG_ARENA_ALIGNED_SIZE(sizeof(datadog_arena));
+    arena->ptr = (char *)arena + aligned;
+    arena->end = (char *)arena + size;
+    arena->prev = NULL;
+
+    return arena;
+}
+
+void datadog_arena_grow(datadog_arena **arena_ptr, size_t min_size) {
+    datadog_arena *arena = *arena_ptr;
+    size_t prev_size = arena->end - (char *)arena;
+    min_size += DATADOG_ARENA_ALIGNED_SIZE(sizeof(datadog_arena));
+    size_t size = prev_size > min_size ? prev_size : min_size;
+    // todo: align min_size to a power of 2 if it doesn't fit?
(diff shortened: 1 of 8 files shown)


ext/DatadogArena/arena.c
```diff
@@ -0,0 +1,59 @@
...
+    do {
+        datadog_arena *prev = arena->prev;
+        free(arena);
+        arena = prev;
+    } while (arena);
+}
+
+char *datadog_arena_checkpoint(datadog_arena *arena) { return arena->ptr; }
+
+// todo: what to do when checkpoint isn't found in arena?
```
SammyK, line 44: Maybe in a follow up PR we can figure out syntax to designate "debug mode" across the shared/vendored components and then do an assertion for the checkpoint in debug mode. WDYT?

SammyK (commented):
This will be very useful - thanks @morrisonlevi! Once the CI stuff is ironed out, this LGTM. 👍

ext/DatadogArena/arena.c
morrisonlevi: On one hand having a debug build that switches behavior at compile-time is pretty nice as it removes checks from runtime, but it does mean you can't enable/disable it easily. I'm not sure.

SammyK (approved):

ext/DatadogArena/arena.c
SammyK: Reading the code again, it looks like a checkpoint that doesn't exist in the arena will invalidate the arena pointer. I think ideally the pointer would not change if the checkpoint did not exist and the caller can check the return value for an error state.

morrisonlevi: Walking the entire arena to find it does make it more expensive. It's probably not a common operation so that may be okay.

PR #1085, opened 2020-11-05
Separate extension code by PHP major version

SammyK:
### Description

This is a follow up PR to #1082 that separates extension code by PHP major version. In order to make the PR more reviewable, PHP 8 compiles out of the `ext/php7` directory for now. There will be a follow up PR to copy PHP 8 code into `ext/php8`.

- The files from `src/ext/phpN` were moved to `ext/phpN`.
- `src/ext/php5_4` was moved to `ext/php5/php5_4`.
- The files from `src/ext` were moved into `ext/php5` and then duplicated into `ext/php7`. The only modifications to any of the files were include paths.
- There were two separate `dispatch.c` files: one in `src/ext/phpN` and one in `src/ext`. The one that was in `src/ext/phpN` was moved to `ext/phpN/phpN/dispatch.c`.
- There were two separate `engine_hooks.c` files: one in `src/ext/phpN` and one in `src/ext`. The one that was in `src/ext/phpN` was moved to `ext/phpN/phpN/engine_hooks.c`.
- `version.h` sits in the root of `ext/` to be shared across all major PHP versions.

### Reviewer checklist
- [x] Milestone is set.
- [ ] ~~Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.~~


morrisonlevi (commented):
Prelim review looks good; will look at it more tomorrow (or maybe later today, we'll see).

morrisonlevi (approved):
Would `#include "ext/version.h"` work? Seems like it should based on `config.m4`. I think I prefer a relative-to-root path over a relative-to-current-file path. What do you think?

I've hit approve; we can do that change in another PR if we agree to it.

SammyK:
Very good idea. I'll make sure we address that in a follow up PR.

PR #1089, opened 2020-11-06
PHP 8: Observer API

SammyK:
### Description

This PR refactors the engine hooks to use the new [observer API](https://github.com/php/php-src/blob/PHP-8.0/Zend/zend_observer.h) in PHP 8. \o/

Notable differences from PHP 7:

- The dispatch is now immutable and cannot be overwritten. Deferred loading still works as expected.
- `dd_trace_reset()` is now noop. This is due to the "ask once and cache" nature of the dispatch lookup that is dictated by the observer API design.
- Generators are fully supported on PHP 8 now. However, `yield from` is not observed properly. This is a limitation of the observer API that will be fixed in an upcoming patch release of PHP 8.0.
- Instrumenting a sub class of a parent class that is already instrumented is not possible. We might want to revisit this topic later, but at the moment this would add extra complexity and overhead.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1187,7 +1187,7 @@ workflows:
                 - php:7.2-fpm-alpine
                 - php:7.3-fpm-alpine
                 - php:7.4-fpm-alpine
-                - php:8.0.0RC3-fpm-alpine
+                - php:8.0.0RC4-fpm-alpine # Add RC5 when available
       - pecl_tests:
           requires: [ "Build PECL" ]
           name: "PHP 54 PECL tests"
--- a/dockerfiles/ci/alpine/docker-compose.yml
+++ b/dockerfiles/ci/alpine/docker-compose.yml
@@ -13,5 +13,5 @@ services:
       context: ./php-8.0
       args:
         phpVersion: 8.0
-        phpTarGzUrl: https://github.com/SammyK/php-src/archive/sammyk/php8.0.0-pre-rc5.tar.gz
-        phpSha256Hash: 9597c3497f6afa05e13db1e69e86e53bcfc19ad9581c30384be0f7f8f55c3eae
+        phpTarGzUrl: https://github.com/php/php-src/archive/php-8.0.0RC5.tar.gz
+        phpSha256Hash: ec5836f816e17fd8a281f671bb247b4c420f8de23fb425d599877cd911a54230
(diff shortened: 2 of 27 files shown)


morrisonlevi:
Overall this looks good to me. I just want to take a little bit more time to think about this bit:

> Instrumenting a sub class of a parent class that is already instrumented is not possible. We might want to revisit this topic later, but at the moment this would add extra complexity and overhead.

That sounds good, but I'm not sure if it's the complete picture. Unless I missed it in review, we still do recursive lookups, so consider a case like this:

```php
class MyPdo extends PDO {
    // note that it does not overwrite `prepare()`
}
// ...
$result = $mypdo->prepare();
```

As I understand it from a read through (but did not test) is that there is a race here: if regular PDO::prepare is called first, then the cached dispatch will be PDO's, but if MyPdo::prepare is called first then from then on out whenever we call Pdo::prepare we'll actually use MyPdo's.

SammyK:
Very good point. Yes, you are spot on. However, since the dispatch is only cached on `op_array`'s for userland calls, this limitation would not apply in the particular example of using `PDO` as the parent class. But in the case of a userland parent & child, as you pointed out, whichever instance makes the first call will cache the dispatch for all subsequent calls.

Here's some code to illustrate that point:

```php
<?php

DDTrace\trace_method('ParentPdo', 'prepare', function ($s) {
	$s->service = 'parent.class';
});

DDTrace\trace_method('ChildPdo', 'prepare', function ($s) {
	$s->service = 'child.class';
});

class ParentPdo
{
	public function __construct(
		private string $dsn,
		private string $username,
		private string $password
		)
	{
	}

	public function prepare(
		string $query,
		array $opts = []
		): array
	{
		return $opts;
	}
}

class ChildPdo extends ParentPdo
{
	// Note that it does not overwrite `prepare()`
}

$childPdo = new ChildPdo('mysql:dbname=test;host=127.0.0.1', 'child', 'pass123');
$childPdo->prepare('SELECT * FROM users WHERE name = :name');

$parentPdo = new ParentPdo('mysql:dbname=test;host=127.0.0.1', 'parent', 'pass123');
$parentPdo->prepare('SELECT * FROM users WHERE name = :name');

array_map(function ($span) {
	echo $span['name'] . ': ' . $span['service'] . PHP_EOL;
}, array_reverse(dd_trace_serialize_closed_spans()));
```

Which outputs:

```
ChildPdo.prepare: child.class
ParentPdo.prepare: child.class
```

But if we call `ParentPdo::prepare()` first and then `ChildPdo::prepare()` second, the output becomes:

```
ParentPdo.prepare: parent.class
ChildPdo.prepare: parent.class
```

morrisonlevi (approved):
As I suspected. I'm going to approve this but let's not lose track of it.

PR #1091, opened 2020-11-10
Make execution of ext tests locally faster and easier

labbati:
### Description

As part of every execution, before we copied ALL files (sources, tests, .m4) and run a complete build even if only one single test file changed, not a source file. With this PR we separate the concept of source files and test files.

When a test file changes, the file is only automatically copied to `tmp/build_extension/tests/ext/**` and a build is NOT done.
When a source file changes, the file is copied to `tmp/build_extension/**` and a build IS done automatically.

Now developers can run `make test_c TESTS=tests/ext/some_thing.phpt` and the quickest possible process will be executed, only building if necessary.

Note that the `EXTRA_CFLAG` variable is provided to provide extra CFLAGS, e.g. `EXTRA_CFLAGS=-g`.

Limitations:
  - When a `.h` file changes, it is automatically copied to the build directory but a build it NOT done.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,15 +18,30 @@ C_FILES := $(shell find ext src/dogstatsd -name '*.c' -o -name '*.h' | awk '{ pr
 TEST_FILES := $(shell find tests/ext -name '*.php*' -o -name '*.inc' | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
 M4_FILES := $(shell find m4 -name '*.m4*' | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
 
+# The following differentiation exists so we can build only (but always) the relevant files while executing tests
+#  - when a `.phpt` changes, we only copy the file to the build dir and we DO NOT rebuild
+#  - when a `.c` changes, we copy the file to the build dir and we DO the specific .lo build and linking
+#  - when a `.h` (or anything else) changes, we remove all .lo files from the build directory so a full build is done
+# The latter, avoids that during development we change something in a header included in multiple .c files, then we
+# change only one of those .c files, we only rebuild that one, we SEGFAULT.
+#
+# Note: while this adds some complexity, as a matter of facts it does not impact production builds in CI which are done
+# from scratch. But the benefits are that we can have the quickest possible `modify --> test` cycle possible.
 $(BUILD_DIR)/tests/%: tests/%
 	$(Q) echo Copying tests/$* to $@
 	$(Q) mkdir -p $(dir $@)
 	$(Q) cp -a tests/$* $@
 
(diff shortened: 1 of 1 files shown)


morrisonlevi:
I think the `clean` was thrown in there because of limitations like these ones:

> When a .h file changes, it is automatically copied to the build directory but a build it NOT done.

It is annoying to always start from scratch, but I'd rather start from scratch than sigsegv and not realize it's because of a header incongruence.

Makefile
```diff
@@ -36,7 +40,7 @@ $(BUILD_DIR)/configure: $(BUILD_DIR)/config.m4
 $(BUILD_DIR)/Makefile: $(BUILD_DIR)/configure
 	$(Q) (cd $(BUILD_DIR); ./configure)
 
-$(SO_FILE): $(ALL_FILES) $(BUILD_DIR)/Makefile
+$(SO_FILE): $(C_FILES) $(BUILD_DIR)/Makefile
```
labbati, line 43: Note that `$(BUILD_DIR)/config.m4 $(M4_FILES)` is not included as a dependency as they are already transitive dependencies of `$(BUILD_DIR)/Makefile`

labbati:
> I think the `clean` was thrown in there because of limitations like these ones:
> 
> > When a .h file changes, it is automatically copied to the build directory but a build it NOT done.
> 
> It is annoying to always start from scratch, but I'd rather start from scratch than sigsegv and not realize it's because of a header incongruence.

Good point. Do you think that removing the `.so` file when even only one file changes make sense? Something like

```
$(BUILD_DIR)/%: %
	$(Q) echo Copying $* to $@
	$(Q) mkdir -p $(dir $@)
	$(Q) cp -a $* $@
	$(Q) rm -f $(SO_FILE)                <<<<< ADDED
```

labbati:
Nevermind, let me look better into it

labbati:
@morrisonlevi I think what follows is a good compromise:

- when a `.phpt` changes, we only copy the file to the build dir and we DO NOT rebuild
- when a `.c` changes, we copy the file to the build dir and we DO the specific .lo build and linking
- when a `.h` (or anything else) changes, we remove all .lo files from the build directory so a full build is done

Note 1: as a matter of facts this does not impact CI (hence the released package) as basically there the build happens from scratch. So the added complexity here (that I  acknowledge) only impacts local development.

Note 2: the purpose of this is to avoid scenarios when we change something in a header included in multiple .c files, then we only rebuilds 1 of them and we have segfaults.

SammyK (approved):
Thanks for improving this experience @labbati! 💯

morrisonlevi (approved):

PR #1095, opened 2020-11-13
Wall-time profiler (WIP)

morrisonlevi:
### Description

Opening this PR to get some code review, but there is _no_ intent to merge, and will be closed when review is done.


ext/DDProf/src/MurmurHash3.cpp
```diff
@@ -0,0 +1,348 @@
+//-----------------------------------------------------------------------------
```
SammyK, line 1: What's the difference between this MurmurHash3 and the C version in `ext/MurmurHash3`? Are we not able to use just one implementation?

SammyK (commented):
I haven't given this a comprehensive review yet, but I left a few questions on the parts I looked at.

ext/DDProf/include/ddprof/profiler.hh
```diff
@@ -0,0 +1,87 @@
...
+    class scheduler scheduler;
+    std::mutex m;
+    enum status status;
+};
+
+inline profiler::profiler()
+    : collectors{},
+      exporters{},
+      recorder{},
+      scheduler(recorder, exporters, std::chrono::nanoseconds{10000000}),
```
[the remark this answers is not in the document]
morrisonlevi: It's the time it sleeps between its actions, which isn't _quite_ the same as interval, but yeah, that's its point. This will be configuration eventually, like `ddtrace.profiling.scheduler.interval` or such.

ext/DDProf/src/MurmurHash3.cpp
morrisonlevi: I don't think I meant to commit both of them; will look into that. Thanks!

ext/MurmurHash3/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+cmake_minimum_required(VERSION 3.10)
```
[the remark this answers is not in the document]
morrisonlevi: I thought about this, but at least with my current know-how I don't know to also make the symbols private if we do it that way. I think it's decently likely that some other extension would use murmurhash, so need to make sure the symbols are private.

ext/DDProf/src/MurmurHash3.cpp
morrisonlevi: I extracted the hashing to `ext/DatadogStringHash`, which exposes C99 and C++14 APIs:

- `datadog_string_hash` for C99+
- `datadog::string_hash` for C++14+

For now only the C++ API is used in this PR.

The API still uses MurmurHash3, but I made it so none of the symbols are exposed. This way we can shared the string hashing bits across profiling and tracing if we need to, but we won't have linker issues with other libraries.

ext/DDProf/src/pprof_exporter.cc
```diff
@@ -0,0 +1,230 @@
...
+            PProf__Line *line = add_line(&arena, function_id, frame.lineno);
+
+            PProf__Location location = PPROF__LOCATION__INIT;
+            location.line = &line;
+            location.n_line = 1;
+            uint64_t location_offset = locations.insert(location);
+
+            sample.location_id[local_locations++] = locations[location_offset].id;
+            sample.value = (int64_t *) datadog_arena_alloc(&arena, sizeof(int64_t));
+            *sample.value = 0; // todo: this is nanoseconds since... what?
```
dbenamydd, line 199: If this is the wall time "data set", what we do in the python profiler is allocate the whole slice of time since the last sample to this one- https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/profiling/collector/stack.pyx#L484, https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/profiling/collector/stack.pyx#L488, https://github.com/DataDog/dd-trace-py/blob/master/ddtrace/profiling/collector/stack.pyx#L339.

PR #1099, opened 2020-11-19
Fix distributed tracing on PHP 8

SammyK:
### Description

The curl extension API was upgraded from a `resource` to a `CurlHandle` object in php/php-src#5402. This PR updates the curl function handler overrides with the updated ZPP params and injects the distributed tracing headers using a similar implementation to PHP 5 in order to reduce the number of calls into userland.

### Reviewer checklist
- [ ] Milestone is set.


ext/php8/php8/engine_hooks.c
```diff
@@ -119,7 +119,21 @@ static bool dd_should_trace_helper(zend_execute_data *call, zend_function *fbc,
     HashTable *function_table = NULL;
     bool found = ddtrace_try_find_dispatch(scope, &fname, &dispatch, &function_table);
     if (found && dispatch->options & DDTRACE_DISPATCH_DEFERRED_LOADER) {
-        if (Z_TYPE(dispatch->deferred_load_integration_name) != IS_NULL) {
+        do {
```
labbati, line 122: Why do we need a `do { ... } while (0);` here?

SammyK: It's there for better readability. It allowed me to flatten the `if` blocks because I needed to add one more `if` branch that would further indent code that was already three blocks deep. The `while` block allowed me to skip over this block using `break` so that I didn't have to indent even further for the extra check. I could have used a `goto` instead but I find those harder to read.

labbati: smart! 😄

labbati (approved):
Thank you @SammyK for the pairing session while reviewing it.
As a follow up we discussed to enable ASAN tests for PHP 8 as well so we are 100% confident on these changes.

Great work on this 🥇

SammyK:
Thanks @labbati! I'll follow up with a PR to add ASAN tests on PHP 8. 👍

PR #1100, opened 2020-11-19
Add clean install of Laravel 8 for testing

SammyK:
### Description

This is a clean install of Laravel 8 without any modification to be used for testing in a follow up PR that will add the Laravel 8 integration.

```bash
$ composer create-project --prefer-dist laravel/laravel Version_8
```

### Reviewer checklist
- [ ] Milestone is set.


labbati (changes requested):
Thank you @SammyK! Can I suggest to:
- add the installation line to https://github.com/DataDog/dd-trace-php/tree/master/tests/Frameworks as we do for the other versions and frameworks (if we find that README useless we can always remove it);
- rename the whole directory from `Version_8` to `Version_8_0` for consistency with other versions and to be future proof when new versions will be released

SammyK:
Thanks @labbati! Nice catch on the framework docs. Updated.

> rename the whole directory from Version_8 to Version_8_0 for consistency with other versions and to be future proof when new versions will be released

This naming was intentional based on [Laravel's switch to semver](https://laravel.com/docs/6.x/releases#versioning-scheme) starting in Laravel 6. In the `composer.json`, the installed framework [version is `^8.12`](https://github.com/DataDog/dd-trace-php/blob/3f4ae862a5f357e486cdd285ae8de27b59bfa9f6/tests/Frameworks/Laravel/Version_8/composer.json#L15). Looking at the [latest Laraval releases](https://github.com/laravel/framework/releases), the minor versions can get a bump within a few days or so.

So we could rename the directory to `Version_8_12` (or actually `Version_8_4` since that's the [version of `laravel/laravel` that was installed](https://github.com/laravel/laravel/releases/tag/v8.4.2)), but since this doesn't match Laravel's release cycle under semver it might get confusing later on.

Perhaps we could name it `Version_8_x` to make the choice look more explicit? WDYT?

labbati (approved):
> based on Laravel's switch to semver starting in Laravel 6 ...

Good point. I did not know about that. 👍 LGTM!

PR #1107, opened 2020-11-24
Add functions for hashing

morrisonlevi:
### Description

This PR adds two functions for hashing:

1. `datadog::memhash`, which hashes linear bytes. Currently it uses the lower 64 bits of MurmurHash3's 128-bit hash. MurmurHash3 is in the public domain and the author has disclaimed any copyrights to it.
2. `datadog::cantor_hash`, which hashes two unsigned, 64-bit numbers using the [cantor pairing function](https://en.wikipedia.org/wiki/Pairing_function#Cantor_pairing_function). The hash of `(1, 2)` and `(2, 1)` are different. For small integers it forms a perfect hash; if the product of the two integers overflows then there can be collisions.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -737,7 +737,6 @@ jobs:
           name: Build and test Datadog::Arena
           command: |
             export PATH="/opt/cmake/<< parameters.cmake_version >>/bin:$PATH"
-            cmake --version
             mkdir -p /tmp/build/DatadogArena
             cd /tmp/build/DatadogArena
             if [ -f "/etc/debian_version" ]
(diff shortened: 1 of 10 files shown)


SammyK (approved):
LTGM 👍 If we wanted to use `datadog_memhash` in the tracer, how would that impact the way we currently build/deploy the tracer?

morrisonlevi:
We can:

1.  Build it in as a static library. This is what the tests do atm. Since no other extension will be loading this library, we should be symbol-conflict free.
2. Since it's simple enough we can build it directly as part of ddtrace. In fact if we only use the C++ bindings we don't need anything else; it's only if we use the C bindings that we need anything more than that.

I lean toward option 2.

SammyK:
Sounds good! So with option 2, the cmake stuff won't clash with PECL builds?

morrisonlevi:
I think PECL can accommodate either, but Option 2 is a little easier I think because it's more source files instead of configure steps.

PR #1109, opened 2020-11-25
Attach fatal errors to userland root span on PHP 5

SammyK:
### Description

This is a followup PR to #1040 that attaches fatal errors to root spans created in userland for PHP 5.6 and 5.5.

As mentioned in #1034, PHP 5.4 is still not supported.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/php5/serializer.c
+++ b/ext/php5/serializer.c
@@ -1,4 +1,5 @@
 #include <Zend/zend.h>
+#include <Zend/zend_builtin_functions.h>
 #include <Zend/zend_exceptions.h>
 #include <Zend/zend_interfaces.h>
 #include <php.h>
@@ -162,7 +163,7 @@ static void _add_assoc_zval_copy(zval *el, const char *name, zval *prop) {
     }
 /* }}} */
 
-/* This is modelled after _build_trace_string in PHP 5.4
+/* This is modeled after _build_trace_string in PHP 5.4
  * @see https://github.com/php/php-src/blob/PHP-5.4/Zend/zend_exceptions.c#L543-L605
  */
 static int _trace_string(zval **frame TSRMLS_DC, int num_args, va_list args, zend_hash_key *hash_key) {
@@ -228,12 +229,16 @@ static int _trace_string(zval **frame TSRMLS_DC, int num_args, va_list args, zen
     return ZEND_HASH_APPLY_KEEP;
 }
 
-/* This is modelled after getTraceAsString in PHP 5.4
+/* This is modeled after Exception::getTraceAsString() in PHP 5.4
  * @see https://github.com/php/php-src/blob/PHP-5.4/Zend/zend_exceptions.c#L607-L635
  */
-static void _serialize_stack_trace(zval *meta, zval *trace TSRMLS_DC) {
+static char *dd_serialize_stack_trace(zval *trace TSRMLS_DC) {
     char *res, **str, *s_tmp;
     int res_len = 0, *len = &res_len, num = 0;
+
+    if (Z_TYPE_P(trace) != IS_ARRAY) {
+        return NULL;
+    }
 
     res = estrdup("");
     str = &res;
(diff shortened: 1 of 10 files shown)


ext/php5/serializer.c
```diff
@@ -433,6 +441,113 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array TSR
...
+    }
+    if (error.msg) {
+        Z_ADDREF_P(error.msg);
+        add_assoc_zval(meta, "error.msg", error.msg);
+    }
+    if (error.stack) {
+        Z_ADDREF_P(error.stack);
+        add_assoc_zval(meta, "error.stack", error.stack);
+    }
+    return error.type && error.msg ? SUCCESS : FAILURE;
```
morrisonlevi, line 540: We don't use this return value. In this case I think maybe it could be removed, rather than used?

ext/php5/serializer.c
```diff
@@ -433,6 +441,113 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array TSR
...
+}
+
+static dd_error_info dd_fatal_error(int type, const char *format, va_list args) {
+    dd_error_info error = {0};
+    dd_fatal_error_type(type, &error.type);
+    dd_fatal_error_msg(format, args, &error.msg);
+    dd_fatal_error_stack(&error.stack);
+    return error;
+}
```
morrisonlevi, line 549: Is there a particular reason for passing the zval pointers as out parameters instead of using a return value? All 3 functions are void, so that would seem more natural, I think.

ext/php5/serializer.c
```diff
@@ -445,14 +560,51 @@ void ddtrace_error_cb(DDTRACE_ERROR_CB_PARAMETERS) {
...
+                    array_init_size(zmeta, ddtrace_num_error_tags);
+                    zend_update_property(ddtrace_ce_span_data, span->span.span_data, ZEND_STRL("meta"),
+                                         zmeta TSRMLS_CC);
+                    zval_ptr_dtor(&zmeta);
+
+                    meta = _read_span_property(span->span.span_data, ZEND_STRL("meta") TSRMLS_CC);
+                }
+                if (Z_TYPE_P(meta) == IS_ARRAY) {
+                    dd_fatal_error_to_meta(meta, error);
+                }
```
morrisonlevi, line 594: Let me make sure I understand this a bit:

1. We update the property with an empty array with a capacity of `ddtrace_num_error_tags` or 3.
2. We then read that property back to get the actual zval of the empty array.
3. Now we append the fatal error info to the zval of the empty array.

Why don't we write the fatal error info before we update the property?

ext/php5/serializer.c
SammyK: Good point. I was keeping it consistent with the PHP 7 version of the same function, but I agree that it makes more sense just to make it `void` return.

ext/php5/serializer.c
SammyK: Nope, no particular reason; I just got into the PHP 5 API mindset to out-param all the things. 😄 I agree that it is more natural to return. I'll update this.

ext/php5/serializer.c
SammyK: > Why don't we write the fatal error info before we update the property?

If I understand your question correctly, we cannot do this since `SpanData::$meta` might be an existing array that was set from a prehook tracing closure. Is that what you are getting at?

morrisonlevi: Yes, it might be an existing array, but the `Z_TYPE_P(meta) == IS_NULL` already rules that out for this branch. I guess I'd have to see the full code of splitting those cases out a bit more to see if I like that or what is written better, but minimizing calls into Zend APIs seems preferable.

SammyK: Ah! I see what you're saying. Yes, we can remove having to read the property again by adding the error info to the initialized array before updating the property. Good catch. I just updated this.

morrisonlevi (approved):
Looks great. Good work!

Issue #1090, opened 2020-11-09
[BUG] When installed with the PECL command, the command cannot be executed by default.


kjdev:
### Bug description

``` sh
# installed with the pecl
$ pecl install datadog_trace
downloading datadog_trace-0.50.0.tgz ...
Starting to download datadog_trace-0.50.0.tgz (304,877 bytes)
..
You should add "extension=ddtrace.so" to php.ini

# setup ini
$ echo 'extension=ddtrace.so' > /etc/php7/conf.d/ddtrace.ini
$ echo 'ddtrace.request_init_hook=/usr/share/php7/datadog_trace/bridge/dd_wrap_autoloader.php' >> /etc/php7/conf.d/ddtrace.ini

# setup enable Datadog APM
$ export DD_TRACE_ENABLED=true
$ export DD_TRACE_CLI_ENABLED=true

# create test file
$ echo '<?php echo "TEST";' > test.php

$ php test.php
# I'm not seeing anything
$ echo $?
255 # error termination
```

It seems that the definition of `NullLogger` is nowhere to be found.

Add the entity of the class loaded with `dd_register_optional_deps_autoloader.php` to `package.xml`, or
> ```
> $ mkdir -p /usr/share/php7/datadog_trace/src/api/Log
> $ cp /path/to/NullLogger.php /usr/share/php7/datadog_trace/src/api/Log/
> $ php test.php
> TEST
> $ echo $?
> 0
> ```

Add the necessary files to `_generated.php` for execution, or

> The `bridge/_files.php` has a `__DIR__ . '/... /. /src/api/Log/NullLogger.php'` to the `bridge/_files.php` and run `composer -dtooling/generation generate`
> ``` sh
> $ cp /path/to/_generated.php /usr/share/php7/datadog_trace/bridge
> $ php test.php
> TEST
> $ echo $?
> 0
> ```

We think it would be better to do so.

### PHP version
``` sh
$ php -v
PHP 7.3.24 (cli) (built: Oct 30 2020 13:32:21) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.24, Copyright (c) 1998-2018 Zend Technologies
    with ddtrace v0.50.0, Copyright Datadog, by Datadog
```

### Installed extensions
``` sh
$ php  -m
[PHP Modules]
Core
date
ddtrace
filter
hash
json
libxml
openssl
pcre
readline
Reflection
SPL
standard
xml
zlib

[...]

PR #1113, opened 2020-12-04
NullLogger is essential, not optional

morrisonlevi:
### Description

Notably, PECL builds were failing when DD_TRACE_DEBUG=false (default)
was set because they would end up using NullLogger, but it couldn't
be found because it isn't shipped in the package.xml.

Maybe we should ship the optional files in src/api too, but at the
very least NullLogger isn't optional and should be included in
_files.php so it gets shipped in _generated.php.

Relates to #1090.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/bridge/_files.php
+++ b/bridge/_files.php
@@ -87,6 +87,7 @@ return [
     __DIR__ . '/../src/api/Log/LogLevel.php',
     __DIR__ . '/../src/api/Log/AbstractLogger.php',
     __DIR__ . '/../src/api/Log/ErrorLogLogger.php',
+    __DIR__ . '/../src/api/Log/NullLogger.php',
 
     __DIR__ . '/../src/DDTrace/Obfuscation.php',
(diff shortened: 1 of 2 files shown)


morrisonlevi:
I don't know why there are so many failing tests; something major happened in composer or CI updates or something.

I still need to add a test to cover the bug report. Do we have any phpt tests that use the production `ddtrace.request_init_hook` yet?

morrisonlevi:
This is blocked until we can figure out how to either fix or migrate the CentOS 6 job failures.

morrisonlevi:
The upstream fixes have been applied. This just needs tests now.

morrisonlevi:
I have manually verified that this will take care of #1090. We're good to merge and we'll add regression tests next sprint.

labbati (approved):
LGTM!

labbati:
Thank you @morrisonlevi

PR #906, opened 2020-05-27
Fix post-install on debug build of PHP

morrisonlevi:
### Description

The post-install script currently isn't checking for debug builds, so installing a packaged extension doesn't work if the system's PHP is a debug build. You get an error like:

```
PHP Warning:  PHP Startup: ddtrace: Unable to initialize module
Module compiled with build ID=API20131226,NTS
PHP    compiled with build ID=API20131226,NTS,debug
These options need to match
 in Unknown on line 0
```

Unfortunately, this PR is incomplete because the `-debug.so` file that is shipped is not built with a debug version of PHP.

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the appropriate release draft. Create one if necessary.
- [ ] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the appropriate release draft. For community contributors the reviewer is in charge of this task.

PR #1116, opened 2020-12-09
Rebuild CentOS 6 containers

SammyK:
### Description

This PR consolidates all of the `Compile PHP *` jobs onto rebuilt CentOS 6 containers.

The base `datadog/dd-trace-ci:centos-6` container is a "frozen" image of CentOS 6 with all of the `yum` installs required to build PHP. As @morrisonlevi investigated, the `yum` commands no longer work on CentOS 6 due to it going EOL.

This PR also updates the `package verification` job to include PHP 7.3 & 7.4 and also install the tracer on ntz, zts, and debug builds across all supported PHP versions on CentOS 6.

This PR also fixes a bug that would install the wrong `.so` on debug builds of PHP.

### Reviewer checklist
- [x] Milestone is set.


dockerfiles/ci/centos/6/php-5.5/configure.sh
```diff
@@ -11,13 +16,16 @@
...
     --with-mysqli=mysqlnd \
     --with-openssl \
     --with-pdo-mysql=mysqlnd \
-    --with-pdo-pgsql \
-    --with-pdo-sqlite \
     --with-pear \
     --with-readline \
     --with-zlib \
+    --without-pdo-sqlite \
+    --without-sqlite3 \
```
morrisonlevi, line 30: Why the change here?

morrisonlevi (dismissed):

dockerfiles/ci/centos/6/php-5.6/Dockerfile
```diff
@@ -0,0 +1,72 @@
+FROM datadog/dd-trace-ci:centos-6 as base
```
labbati, line 1: This file for [5.6](https://raw.githubusercontent.com/DataDog/dd-trace-php/f0589aad7996a10397593a3592cd714a99982975/dockerfiles/ci/centos/6/php-5.6/Dockerfile) is identical to the ones for [5.5](https://raw.githubusercontent.com/DataDog/dd-trace-php/f0589aad7996a10397593a3592cd714a99982975/dockerfiles/ci/centos/6/php-5.5/Dockerfile) and [5.4](https://raw.githubusercontent.com/DataDog/dd-trace-php/f0589aad7996a10397593a3592cd714a99982975/dockerfiles/ci/centos/6/php-5.4/Dockerfile)

dockerfiles/ci/centos/6/php-7.0/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    mkdir -p ${PHP_INSTALL_DIR_NTS}/conf.d;
+
+FROM base as final
+COPY --from=php-zts $PHP_INSTALL_DIR_ZTS $PHP_INSTALL_DIR_ZTS
+COPY --from=php-debug $PHP_INSTALL_DIR_DEBUG_NTS $PHP_INSTALL_DIR_DEBUG_NTS
+COPY --from=php-nts $PHP_INSTALL_DIR_NTS $PHP_INSTALL_DIR_NTS
+
+RUN set -eux; \
+# Set the default PHP version
+    switch-php ${PHP_VERSION}-debug;
```
labbati, line 65: Same as above

> Since these are mainly the compilation images, what do you think about setting the non debug version as the default one? We still have package verification which is on non-debug builds, but I think we should play defensively here (also considering that our package verification process for non-alpine builds still requires some love).

dockerfiles/ci/centos/6/php-7.1/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    mkdir -p ${PHP_INSTALL_DIR_NTS}/conf.d;
+
+FROM base as final
+COPY --from=php-zts $PHP_INSTALL_DIR_ZTS $PHP_INSTALL_DIR_ZTS
+COPY --from=php-debug $PHP_INSTALL_DIR_DEBUG_NTS $PHP_INSTALL_DIR_DEBUG_NTS
+COPY --from=php-nts $PHP_INSTALL_DIR_NTS $PHP_INSTALL_DIR_NTS
+
+RUN set -eux; \
+# Set the default PHP version
+    switch-php ${PHP_VERSION}-debug;
```
labbati, line 65: Same as above

> Since these are mainly the compilation images, what do you think about setting the non debug version as the default one? We still have package verification which is on non-debug builds, but I think we should play defensively here (also considering that our package verification process for non-alpine builds still requires some love).

dockerfiles/ci/centos/6/php-7.2/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    mkdir -p ${PHP_INSTALL_DIR_NTS}/conf.d;
+
+FROM base as final
+COPY --from=php-zts $PHP_INSTALL_DIR_ZTS $PHP_INSTALL_DIR_ZTS
+COPY --from=php-debug $PHP_INSTALL_DIR_DEBUG_NTS $PHP_INSTALL_DIR_DEBUG_NTS
+COPY --from=php-nts $PHP_INSTALL_DIR_NTS $PHP_INSTALL_DIR_NTS
+
+RUN set -eux; \
+# Set the default PHP version
+    switch-php ${PHP_VERSION}-debug;
```
labbati, line 65: Same as above

> Since these are mainly the compilation images, what do you think about setting the non debug version as the default one? We still have package verification which is on non-debug builds, but I think we should play defensively here (also considering that our package verification process for non-alpine builds still requires some love).

dockerfiles/ci/centos/6/php-7.3/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    mkdir -p ${PHP_INSTALL_DIR_NTS}/conf.d;
+
+FROM base as final
+COPY --from=php-zts $PHP_INSTALL_DIR_ZTS $PHP_INSTALL_DIR_ZTS
+COPY --from=php-debug $PHP_INSTALL_DIR_DEBUG_NTS $PHP_INSTALL_DIR_DEBUG_NTS
+COPY --from=php-nts $PHP_INSTALL_DIR_NTS $PHP_INSTALL_DIR_NTS
+
+RUN set -eux; \
+# Set the default PHP version
+    switch-php ${PHP_VERSION}-debug;
```
labbati, line 65: Same as above

> Since these are mainly the compilation images, what do you think about setting the non debug version as the default one? We still have package verification which is on non-debug builds, but I think we should play defensively here (also considering that our package verification process for non-alpine builds still requires some love).

dockerfiles/ci/centos/6/php-7.4/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+    mkdir -p ${PHP_INSTALL_DIR_NTS}/conf.d;
+
+FROM base as final
+COPY --from=php-zts $PHP_INSTALL_DIR_ZTS $PHP_INSTALL_DIR_ZTS
+COPY --from=php-debug $PHP_INSTALL_DIR_DEBUG_NTS $PHP_INSTALL_DIR_DEBUG_NTS
+COPY --from=php-nts $PHP_INSTALL_DIR_NTS $PHP_INSTALL_DIR_NTS
+
+RUN set -eux; \
+# Set the default PHP version
+    switch-php ${PHP_VERSION}-debug;
```
labbati, line 65: Same as above

> Since these are mainly the compilation images, what do you think about setting the non debug version as the default one? We still have package verification which is on non-debug builds, but I think we should play defensively here (also considering that our package verification process for non-alpine builds still requires some love).

dockerfiles/verify_packages/verify_rpm_centos6.sh
```diff
@@ -0,0 +1,19 @@
...
+
+for phpVer in $(ls ${PHP_INSTALL_DIR}); do
+    echo "Installing ddtrace on PHP ${phpVer}..."
+    switch-php $phpVer
+    rpm -Uvh /build_src/build/packages/*.rpm
+    php --ri=ddtrace
+
+    # Uninstall the tracer
+    rpm -e datadog-php-tracer
+    rm -f /opt/datadog-php/etc/ddtrace.ini
```
labbati, line 18: Question: should we remove the `9(8|9)-ddtrace.ini`s symlinks as well?

package/post-install.sh
```diff
@@ -168,10 +168,13 @@ PHP_MAJOR_MINOR=$(invoke_php -r 'echo PHP_MAJOR_VERSION;').$(invoke_php -r 'echo
 PHP_CFG_DIR=$(invoke_php -i | grep 'Scan this dir for additional .ini files =>' | sed -e 's/Scan this dir for additional .ini files =>//g' | head -n 1 | awk '{print $1}')
 
 PHP_THREAD_SAFETY=$(invoke_php -i | grep 'Thread Safety' | awk '{print $NF}' | grep -i enabled)
+PHP_DEBUG_BUILD=$(invoke_php -i | grep 'Debug Build => ' | awk '{print $NF}' | grep -i yes)
 
 VERSION_SUFFIX=""
 if [[ -n $PHP_THREAD_SAFETY ]]; then
     VERSION_SUFFIX="-zts"
+elif [[ -n $PHP_DEBUG_BUILD ]]; then
```
labbati, line 176: I think this is correct but I want to double check with you to make sure I am understanding it correctly.

`$PHP_DEBUG_BUILD` means php's debug build (like in `--enable-debug`), while for our extension `-debug` means pretty much `CFLAGS = -g`.

So here we are saying that when we have debug builds, no matter what, we want also our symbols (which is logical). While this would not impact the actual "production" and the respective verification.

Am I correct?

labbati (dismissed):
@SammyK thank you for tackling this. Impressive work.

Overall I have very few comments that I believe require immediate action.

Then, there are a few things that can be improved as a refactoring when we will have some spare cycles. I added comments for those points. If you agree with them I'd ask you to drop them in a card so we do not loose track of them.

dockerfiles/ci/centos/6/php-5.5/configure.sh
SammyK: Very good question. IIRC this was disabled in the PHP 8 container because [PHP 7.4+ requires sqlite3 v3.7.4](https://www.php.net/manual/en/sqlite3.requirements.php) which would have required building it from source. But I don't think there's a reason to disable it on PHP 7.3 and below so I'll add this back to all of the PHP versions that support the old version of sqlite3.

dockerfiles/ci/centos/6/php-5.6/Dockerfile
SammyK: Yep! Very good point, I'll make separate Dockerfiles for the PHP major versions.

dockerfiles/verify_packages/verify_rpm_centos6.sh
SammyK: Since we're just validating that it installs properly on each of the PHP flavors, it's not necessary to remove the symlinks. We're only removing the `/opt/datadog-php/etc/ddtrace.ini` file here because the post install hook does not overwrite an existing one (which would be pointing to the wrong `.so` after the second install).

dockerfiles/ci/centos/6/php-5.4/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# Set default INI settings
+RUN set -eux; \
+    for phpVer in $(ls ${PHP_INSTALL_DIR}); \
+    do \
+        sed 's/;date.timezone =/date.timezone = UTC/' <${PHP_SRC_DIR}/php.ini-development >${PHP_INSTALL_DIR}/${phpVer}/etc/php.ini; \
+    done;
+
+RUN set -eux; \
+# Set the default PHP version
+    switch-php ${PHP_VERSION}-debug;
```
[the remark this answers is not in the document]
SammyK: This is a very good point. I originally had these set up to use a non-debug build but changed it back to be consistent with the other containers. All of the CI checks (including the package verification) are refactored to explicitly set the PHP flavor on CentOS 6 containers, but your point about being defensive is a good one and I'll make sure to update these. 👍

dockerfiles/ci/centos/6/php-5.4/Dockerfile
```diff
@@ -0,0 +1,72 @@
+FROM datadog/dd-trace-ci:centos-6 as base
+
+ARG phpVersion
+ENV PHP_INSTALL_DIR_ZTS=${PHP_INSTALL_DIR}/${phpVersion}-zts
+ENV PHP_INSTALL_DIR_DEBUG_NTS=${PHP_INSTALL_DIR}/${phpVersion}-debug
+ENV PHP_INSTALL_DIR_NTS=${PHP_INSTALL_DIR}/${phpVersion}
+ENV PHP_VERSION=${phpVersion}
+
+# Download and extract PHP source
```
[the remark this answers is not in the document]
SammyK: I do love this idea, but unfortunately if I change `download-src.sh`, it will bust the layer caches for the base image (which is frozen in an EOL state) and I wouldn't be able to add any more layers to it. But I agree that we should change this for other containers that are not EOL.

dockerfiles/ci/centos/6/php-5.4/Dockerfile
```diff
@@ -0,0 +1,72 @@
+FROM datadog/dd-trace-ci:centos-6 as base
```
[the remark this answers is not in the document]
SammyK: These Dockerfiles use multi-stage to build each PHP version from the `/tmp` directory and then copy over the installed PHP flavors into the `final` stage.

dockerfiles/ci/centos/6/Dockerfile
```diff
@@ -0,0 +1,81 @@
+FROM centos:6 as base
```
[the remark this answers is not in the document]
SammyK: Thanks for finding the copy pasta! I'll remove this.

dockerfiles/ci/centos/6/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+        vim \
+        xz \
+        zlib-devel; \
+    yum install -y devtoolset-7; \
+    yum clean all;
+
+ENV SRC_DIR=/usr/local/src
+
+COPY download-src.sh /root/
+RUN set -eux; \
```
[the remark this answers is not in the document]
SammyK: This is a good point. As a general rule I personally don't like pulling complete `RUN` commands into a bash script because 1) it's easy to accidentally bust a lot of layer caches at once if the file hash changes for any reason; including changing comments, 2) it's harder to read the Dockerfile when major parts of the build are in a separate file, and 3) it makes things more complicated when things like build args are used.

I see the value of using a separate script when sharing the script across multiple Dockerfiles, but in this case the `base` image acts as the single shared point so we wouldn't be getting that advantage.

That's just my 2 cents. 🙂 Overall I don't have a super strong opinion if we think using a script is better for future builds.

dockerfiles/ci/centos/6/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+# bison 3.0.0 or later is required to generate PHP parsers
+    /root/download-src.sh bison https://ftp.gnu.org/gnu/bison/bison-3.7.3.tar.gz; \
+    cd "${SRC_DIR}/bison"; \
+    ./configure && make && make install; \
+# re2c 0.13.4 is required to generate PHP lexers
+    /root/download-src.sh re2c https://github.com/skvadrik/re2c/releases/download/2.0.3/re2c-2.0.3.tar.xz; \
+    cd "${SRC_DIR}/re2c"; \
+    ./configure && make && make install;
+
+# CMake 3.0.2 or higher is required.  You are running version 2.8.12.2
```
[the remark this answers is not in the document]
SammyK: I agree it is worded strangely. I just copy/pasted the build errors as a came across them. 😄 I'll reword these to be less confusing. 👍

package/post-install.sh
SammyK: Thanks for double checking this!

> $PHP_DEBUG_BUILD means php's debug build (like in --enable-debug)

Correct.

> while for our extension -debug means pretty much CFLAGS = -g.

All three of the `.so` build flavors we ship ([nts](https://github.com/DataDog/dd-trace-php/blob/f0589aad7996a10397593a3592cd714a99982975/.circleci/config.yml#L866), [zts](https://github.com/DataDog/dd-trace-php/blob/f0589aad7996a10397593a3592cd714a99982975/.circleci/config.yml#L878), [debug](https://github.com/DataDog/dd-trace-php/blob/f0589aad7996a10397593a3592cd714a99982975/.circleci/config.yml#L872)) include debug symbols. 

When a PHP flavor is built with `--enable-debug` and/or `--enable-zts`, it will change the build ID. Our extension must be built against the proper PHP flavor, otherwise we'll run into errors like the one identified in #906.  The `switch-php` script also updates the symlink to `php-config` which will ensure that the `.so` will build with the proper build ID.

A while ago we had a bug where our debug builds only included debug symbols but didn't have the correct build ID for PHP flavors built with `--enable-debug` but now they build with the proper build ID. 

The change in this PR fixes another bug where the post-install script did not install the the proper `.so` for the debug flavor of PHP. Now it properly matches the build ID for debug builds of PHP.

morrisonlevi (approved):

PR #1118, opened 2020-12-14
Add StringTable component

morrisonlevi:
### Description

The StringTable is responsible for de-duplicating strings, and also storing them in a way so that they are accessible through a contiguous buffer. This should be useful for both the tracer and the profiler.

It's written C++. I had to adjust the `.clang-format` rules for this so that it plays nice with EditorConfig on certain C++ features. It uses the previously added [MemHash](#1107) and [Arena](#1068) components. I had to rename some cmake variables so that tests aren't built by default for Arena and MemHash when building tests for StringTable. Arena and MemHash are installed into `/opt/libdatadog` so StringTable can find them (in CI, not in the package).

One implementation detail I'm not sure of is actually storing the hash. If an interned string is used as a string key into a hash table, having that hash cached is great. However, it adds 8 bytes per string, which can be high overhead for short strings in particular. For now I decided to keep it, but we can revisit the balance with concrete data.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -750,6 +750,7 @@ jobs:
             tar -xf ../v<< parameters.catch2_version >>.tar.gz --strip 1
             cmake -Bbuild -H. -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX=/opt/catch2
             cmake --build build/ --target install
+
       - run:
           name: Build and test Datadog::Arena
           command: |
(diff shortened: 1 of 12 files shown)


morrisonlevi:
Closing; this work is being done [elsewhere](https://github.com/DataDog/dd-trace-php/tree/ddprof).

PR #1120, opened 2020-12-21
Make Uri to resource name normalization case insensitive

labbati:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/private/functions.php
```diff
@@ -11,8 +11,8 @@ public static function getDefaultUriPathNormalizeRegexes()
     {
         return [
             '/^\d+$/',
-            '/^[0-9a-f]{8}-?[0-9a-f]{4}-?[1-5][0-9a-f]{3}-?[89ab][0-9a-f]{3}-?[0-9a-f]{12}$/',
-            '/^[0-9a-f]{8,128}$/',
+            '/^[0-9a-fA-F]{8}-?[0-9a-fA-F]{4}-?[1-5][0-9a-fA-F]{3}-?[89abAB][0-9a-fA-F]{3}-?[0-9a-fA-F]{12}$/',
+            '/^[0-9a-fA-F]{8,128}$/',
```
SammyK, line 15: Out of curiosity, why not add the `i` modifier after the pattern delimiters to make them case insensitive?

SammyK (approved):
Just left one question that is not blocking. LGTM! 👍

src/private/functions.php
labbati: This is a good point.

The [current implementation](https://github.com/DataDog/dd-trace-php/blob/014f58d18b3eab0143886139c6b159a48d551a24/src/private/functions.php#L115) was lenient about `/` around the value provided in the regex. Meaning we trigger away `/` at the begin of end of the values in the env and then apply them again.

If I add `'/i'` instead of `'/'` at the [end of the line](https://github.com/DataDog/dd-trace-php/blob/014f58d18b3eab0143886139c6b159a48d551a24/src/private/functions.php#L115) then I will make all regex case insensitive, and I don't want to.

It is still possible to refactor the line I linked detecting optional `/.../...` and re-applying them, but it would require a refactoring.

Would it be OK if I open a RFC for this so we can reason a bit about the less confusing way to propose configuration and then we do the refactoring?

SammyK: Ah, that makes total sense. Yes, it would be good to evaluate how this is currently configured, and this is certainly not a blocker for merge. Thanks @labbati!

labbati: Thank you @SammyK

labbati:
Thanks @SammyK

PR #1121, opened 2020-12-21
Fix multiple deferred integration loading on PHP from 7.0 to 7.2

labbati:
### Description

We call the function `DDTrace\Integrations\load_deferred_integration` every time we want to defer load an integration via ext's utility function `ddtrace_call_function`
- https://github.com/DataDog/dd-trace-php/blob/014f58d18b3eab0143886139c6b159a48d551a24/ext/php7/engine_api.c#L81

Between 7.2 and 7.3 the field `zend_fcall_info_cache->initialized` was removed:
- 7.2: https://heap.space/xref/PHP-7.2/Zend/zend_API.h?r=cce2e33c#54
- 7.3: https://heap.space/xref/PHP-7.3/Zend/zend_API.h?r=cce2e33c#52

In `ext/php7/engine_api.c` we create a new `zend_fcall_info_cache` struct at each invocation and to pass it as a `function_handler` the proxied `fn_proxy` from previous integration loading.
- https://github.com/DataDog/dd-trace-php/blob/014f58d18b3eab0143886139c6b159a48d551a24/ext/php7/engine_api.c#L86

The first time `fn_proxy == null`, so the field `fcc.initialized` is set to `1` by an invocation of `zend_is_callable_ex`.
The second time `fn_proxy != null`, hence `zend_is_callable_ex` is not invoked and the field `fcc.initialized` is not set to `1`, potentially causing a segfault.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/DDTrace/Transport/Http.php
```diff
@@ -215,7 +215,7 @@ private function sendRequest($url, array $headers, $body, $tracesCount)
                 'extra' => $extra,
             ]);
 
-            if ($isDebugEnabled && isset($verbose)) {
+            if ($isDebugEnabled && $verbose !== false) {
```
labbati, line 218: Motivation:
```
------ -----------------------------------------------------------------
  Line   Transport/Http.php
 ------ -----------------------------------------------------------------
  218    Variable $verbose in isset() always exists and is not nullable.
 ------ -----------------------------------------------------------------
```
`fopen` [returns](https://www.php.net/manual/en/function.fopen.php) either a resource or `false`.

morrisonlevi (changes requested):
This looks good to me, assuming tests pass now. My only question is why Yii was chosen: was there any specific reason? If so we should document that there is or isn't, I think.

morrisonlevi (approved):

labbati:
Thank you @morrisonlevi

PR #1119, opened 2020-12-16
Support distributed tracing for calls to curl_multi_exec()

SammyK:
### Description

WIP

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1123, opened 2020-12-22
Refactor PHP 7 curl distributed tracing

SammyK:
### Description

This PR refactors the curl distributed tracing in PHP 7. The primary reasons for doing so are:

0. Reduce the number of calls into userland by making a single call to `DDTrace\Bridge\curl_inject_distributed_headers`. This was recently updated for PHP 8 (#1099).
1. Make the distributed tracing support consistent across all the PHP versions.
2. Run the existing distributed-tracing phpt tests on PHP 7.
3. Reduce the complexity of `curl_multi_exec()` support on PHP 7. (WIP PR #1119)

### Reviewer checklist
- [x] Milestone is set.


ext/php7/engine_api.c
```diff
@@ -121,10 +127,11 @@ ZEND_RESULT_CODE ddtrace_call_function(zend_function **fn_proxy, const char *nam
     // ZVAL_COPY_VALUE(&fci.function_name, fcc.function_handler->common.function_name);
 
     fci.retval = retval;
-    fci.params = argv;
     fci.no_separation = 0;  // allow for by-ref args
-    fci.param_count = argc;
     ZEND_RESULT_CODE result = zend_call_function(&fci, &fcc);
+
+    zend_fcall_info_args_clear(&fci, 1);
```
morrisonlevi, line 133: Can you explain why this moved?

morrisonlevi (approved):
Looks good to me! Just had a question about one call being moved around.

ext/php7/engine_api.c
SammyK: This was added because `zend_fcall_info_argv()` on line 92 will alloc `fci.params` and copy all the zvals instead of relying on the caller to do that.

PR #1125, opened 2020-12-30
Add function to correctly extract trace id in all context

labbati:
### Description

Add a function to correctly extract trace id in all contexts: web, short and long running cli scripts.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/tests/Common/TracerTestTrait.php
+++ b/tests/Common/TracerTestTrait.php
@@ -150,6 +150,37 @@ trait TracerTestTrait
     }
 
     /**
+     * This method executes a single script with the provided configuration.
+     */
+    public function inCli($scriptPath, $customEnvs = [], $customInis = [], $arguments = '')
+    {
+        $this->resetRequestDumper();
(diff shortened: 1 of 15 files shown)


tests/Common/TracerTestTrait.php
```diff
@@ -149,6 +149,37 @@ public function inWebServer($fn, $rootPath, $envs = [], $inis = [])
         return $this->parseTracesFromDumpedData();
     }
 
+    /**
```
labbati, line 152: I added this to the trait so we can mixin different test scenarios (web, cli) in the same test. The alternative would be to have 2 tests (one extending from `WebFrameworkTest` and the other from `CLITest`) but this would be less readable in my opinion.

Also, I think we should consider removing `CLITestCase` in favor of `IntegrationTestCase` + this function, but this is a longer conversation beyond the scope of this PR.

SammyK (approved):
LGTM! 👍

PR #1136, opened 2021-01-25
Slim 4 integration

SammyK:
### Description

This PR adds support for Slim 4. It differs from the Slim 3 integration in the following ways:

0. The span names default to `web.request` instead of `slim.request` to be consistent with other integrations.
1. The route name (if not empty) is attached to the `slim.route` span as metadata instead of changing the root span resource name.
2. Slim doesn't have controllers so there are no references to a controller in the Slim 4 integration.

Screen Shot 2021-01-25 at 2 27 30 PM

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/Frameworks/Slim/Version_4/app/routes.php
+++ b/tests/Frameworks/Slim/Version_4/app/routes.php
@@ -1,0 +1,26 @@
+<?php
+declare(strict_types=1);
+
+use App\Application\Actions\User\ListUsersAction;
+use App\Application\Actions\User\ViewUserAction;
+use Psr\Http\Message\ResponseInterface as Response;
+use Psr\Http\Message\ServerRequestInterface as Request;
+use Slim\App;
+use Slim\Interfaces\RouteCollectorProxyInterface as Group;
+
+return function (App $app) {
+    $app->options('/{routes:.*}', function (Request $request, Response $response) {
+        // CORS Pre-Flight OPTIONS Request Handler
+        return $response;
+    });
+
+    $app->get('/', function (Request $request, Response $response) {
+        $response->getBody()->write('Hello world!');
+        return $response;
+    });
+
+    $app->group('/users', function (Group $group) {
+        $group->get('', ListUsersAction::class);
+        $group->get('/{id}', ViewUserAction::class);
+    });
+};
--- a/tests/Frameworks/Slim/Version_4/app/routes.php
+++ b/tests/Frameworks/Slim/Version_4/app/routes.php
@@ -32,7 +32,7 @@ return function (App $app) {
     })->setName('simple-route');
 
     // Create Twig
-    $twig = Twig::create(__DIR__ . '/../templates'/*, ['cache' => 'path/to/cache']*/);
+    $twig = Twig::create(__DIR__ . '/../templates');
 
     // Add Twig-View Middleware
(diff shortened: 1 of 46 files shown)


morrisonlevi (dismissed):
Testing approval on a single commit.

tests/Frameworks/Slim/Version_4/app/routes.php
```diff
@@ -23,4 +25,24 @@
...
         $group->get('/{id}', ViewUserAction::class);
     });
+
+    $app->get('/simple', function (Request $request, Response $response, $args) {
+        $response->getBody()->write("Simple :)\n");
+        return $response;
+    })->setName('simple-route');
+
+    // Create Twig
+    $twig = Twig::create(__DIR__ . '/../templates'/*, ['cache' => 'path/to/cache']*/);
```
morrisonlevi, line 35: Two questions here:

1. Are we supposed to escape or sanitize this directory for information disclosure?
2. Should we keep the commented out cache bit?

tests/Integrations/Slim/V4/CommonScenariosTest.php
```diff
@@ -0,0 +1,119 @@
...
+                    ]),
+                ],
+                'A GET request with an exception' => [
+                    SpanAssertion::build(
+                        'web.request',
+                        'slim_test_app',
+                        'web',
+                        'GET /error'
+                    )->withExactTags([
+                        'slim.route' => 'Closure::__invoke',
```
morrisonlevi, line 100: We are still attaching `slim.route` to the root span; did we mean to do that? I'm not sure why we'd want it on the root span as well as having it as a span resource. I'm thinking that wherever we stick the handler name we should put the route name too, since they kind of go together. So either add `slim.route.name` to the root as well or remove `slim.route` from the root, I think?

tests/Frameworks/Slim/Version_4/app/routes.php
SammyK: 1. This functions the same as Slim 3: the `slim.view` span has the template file name in the metadata but this directory isn't part of the metadata. Is that what you're referring to?
2. Good catch; removed.

tests/Integrations/Slim/V4/CommonScenariosTest.php
SammyK: Good call. I added it to the root span as `slim.route.name`.

tests/Frameworks/Slim/Version_4/app/routes.php
morrisonlevi: 1. This is why I took a nap.
2. Thanks!

tests/Integrations/Slim/V4/CommonScenariosTest.php
morrisonlevi: Thanks. Do you think we should rename the meta on root span `span.route` -> `span.route.handler`? We have `slim.route` and `slim.route.name` now. In the Slim 3 integration we used `slim.route.controller`.

SammyK: Makes sense to me! Added `slim.route.handler` as meta to the root span.

morrisonlevi (approved):

PR #1138, opened 2021-01-27
[Profiler] Fix equality for Location and Sample; add Sample aggregation

morrisonlevi:
### Description

- Fix equality for Location, Sample
- Add naive Sample aggregation
- Fix CMake build

### DIFF
--- a/src/native/pprof.c
+++ b/src/native/pprof.c
@@ -1,3 +1,4 @@
+#include <stdbool.h>
 #include <stdio.h>
 #include <sys/stat.h>
 #include <sys/time.h>
@@ -112,7 +113,6 @@ size_t pprof_strIntern(DProf* dp, char* str) {
 // Forward decl of purely internal functions
 static uint64_t _pprof_mapNew(DProf*, uint64_t, uint64_t, uint64_t, uint64_t, int64_t);
 static uint64_t _pprof_funNew(DProf*, int64_t, int64_t, int64_t, int64_t);
-static uint64_t _pprof_locNew(DProf*, uint64_t, uint64_t, uint64_t*, int64_t*, size_t);
 static uint64_t _pprof_lineNew(DProf*, PPLocation*, uint64_t, int64_t);
 static char _pprof_mapFree(PPMapping**, size_t);
 
@@ -238,11 +238,7 @@ uint64_t pprof_funAdd(DProf* dp, const char* name, const char* system_name, cons
   return _pprof_funNew(dp, id_name, id_system_name, id_filename, start_line)+1;
 }
 
-char isEqualLocation(uint64_t id_mapping, uint64_t addr, PPLocation* B) {
-  return id_mapping == B->mapping_id && addr == B->address;
-}
-
-static uint64_t _pprof_locNew(DProf* dp, uint64_t id_mapping, uint64_t addr, uint64_t* functions, int64_t* lines, size_t n_functions) {
+static uint64_t _pprof_locNew(DProf* dp, uint64_t id_mapping, uint64_t addr, const uint64_t *functions, const int64_t *lines, size_t n_functions) {
   PPProfile* pprof = &dp->pprof;
   uint64_t id = pprof->n_location;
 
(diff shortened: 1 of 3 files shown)


src/native/pprof.c
```diff
@@ -273,21 +269,67 @@ static uint64_t _pprof_locNew(DProf* dp, uint64_t id_mapping, uint64_t addr, uin
...
+static bool isEqualLine(const PPLine *line, uint64_t function_id, int64_t lineno) {
+  uint64_t fid_a = line->function_id, fid_b = function_id;
+  int64_t line_a = line->line, line_b = lineno;
+  return !((fid_a ^ fid_b) | (line_a ^ line_b));
+}
+
+static bool isEqualLocation(const PPLocation *location, uint64_t mapping_id,
+                            uint64_t addr, const uint64_t *functions,
+                            const int64_t *lines, size_t n_functions) {
+  if (location->mapping_id != mapping_id) return false;
```
sanchda, line 281: Yeah, I agree with this early termination order.  If we were just talking about native, then exiting based on address mismatch would be more likely to fail than on mapping, but in general runtimes that can't emit a valid address will have a constant value.

src/native/pprof.c
```diff
@@ -273,21 +269,67 @@ static uint64_t _pprof_locNew(DProf* dp, uint64_t id_mapping, uint64_t addr, uin
 
 }
 
-uint64_t pprof_locAdd(DProf* dp, uint64_t id_mapping, uint64_t addr, uint64_t* functions, int64_t* lines, size_t n_functions) {
+static bool isEqualLine(const PPLine *line, uint64_t function_id, int64_t lineno) {
+  uint64_t fid_a = line->function_id, fid_b = function_id;
+  int64_t line_a = line->line, line_b = lineno;
+  return !((fid_a ^ fid_b) | (line_a ^ line_b));
```
sanchda, line 275: oldschool, but I like it :)

src/native/pprof.c
```diff
@@ -273,21 +269,67 @@ static uint64_t _pprof_locNew(DProf* dp, uint64_t id_mapping, uint64_t addr, uin
...
+                                const uint64_t *functions, const int64_t *lines,
+                                size_t n_functions) {
   PPProfile* pprof = &dp->pprof;
-  for(size_t i=0; i < pprof->n_location; i++) {
-    if(isEqualLocation(id_mapping, addr, pprof->location[i]))
-      return i+1;
+  for (size_t i = 0; i < pprof->n_location; ++i) {
+    PPLocation *location = pprof->location[i];
+    if (isEqualLocation(location, id_mapping, addr, functions, lines, n_functions)) {
+      return location;
```
sanchda, line 301: I've historically stayed away from stdbool, so potentially naive question--since C itself has no types smaller than a byte, if I cast the return (after the call returns, _not_ in a debugger on line 301 ors omething) to `(char)` will modern compilers return `1` if the location is nonzero, or will it be some other value (probably a uint64 mod 256 or whatever)?

If the latter, maybe `return !!location`, but no strong feeling either way.

src/native/pprof.c
```diff
@@ -273,21 +269,67 @@ static uint64_t _pprof_locNew(DProf* dp, uint64_t id_mapping, uint64_t addr, uin
...
 
-char isEqualSample(uint64_t* loc, size_t nloc, PPSample* B) {
-  if (nloc != B->n_location_id) return 0;
-  for (size_t i = 0; i <= nloc; i++) {
-    if (loc[i] != B->location_id[i]) return 0;
+uint64_t pprof_locAdd(DProf* dp, uint64_t id_mapping, uint64_t addr,
+                      const uint64_t *functions, const int64_t *lines,
+                      size_t n_functions) {
+  PPLocation *location = findLocation(dp, id_mapping, addr, functions, lines, n_functions);
+  return (location)
```
sanchda, line 311: Is splitting the conditional operator along the front of the line LLVM style?  Am I just a goofy old bat for immediately parsing 311 as `return location;`?

If there aren't any strong stylistic reservations, then splitting operators at the end of lines ensures the reader knows that they need to keep parsing to get the full statement (OK, yeah, formally we all need to scan for semicolons, but presence-of-token is easier for humans than lack-of-token in general)

```
  return (location) ?
             location -> id :
             ...
```


Again, no particularly strong feelings from me, just noting an ever-so-slight consideration for an alternative.

sanchda (approved):
Looks good to me.  I do prefer passing booleans and I don't even remember why I've always avoided stdbool.  It's a new world, I'll refactor the rest of the `char` returning functions to return `bool`, but no reason to do that in this PR (I mean, unless you really, really want to)

There's a little bit of disharmony in the style because you've jumped ahead to using LLVM, but this will all get resolved post-linting at a later time, so no concerns from me.

src/native/pprof.c
morrisonlevi: I couldn't find a quick reference for conversion _from_ bool to other types in C. It's guaranteed to convert _to_ `0` or `1`, and assume that implication is enough when converting from bool to other types, but I'm not sure.

src/native/pprof.c
morrisonlevi: I wrote it that style, actually. I haven't seen it with the operators at the end-of-line before, I don't think. Here's how `clang-format` does it:

```
  return (location) ? location->id
                    : _pprof_locNew(dp, id_mapping, addr, functions, lines,
                                    n_functions) +
                          1;
```

Which is... bad. Its gratuitous indenting means it line wrapped.  I probably should tweak the expression a bit so it gives a better formatting, or see what knobs there are to control ternaries.

morrisonlevi:
Thanks for the review! I'll worry about the code style in the next PR where I `clang-format` the whole library.

PR #1139, opened 2021-01-27
[Profiler] Run .clang-format on everything; adjust style

morrisonlevi:
### Description

This adjusts the `.clang-format` file and runs it across the project. A few notes on options:

- `AlignOperands: DontAlign` is there mostly because it formats ternary statements quite poorly without it.
- `AllowShortBlocksOnASingleLine: true` is supposed to permit things like `if (cond) { return; }` on a single line, but it doesn't seem to work most of the time.

Let me know if you see anything related to code style you don't like and I can try to adjust the options more.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,9 +1,5 @@
 ---
-BasedOnStyle: Google
----
-Language: Cpp
-ColumnLimit: 80
-IndentWidth: 4
-AccessModifierOffset: 0
-BreakConstructorInitializers: AfterColon
+BasedOnStyle: LLVM
+AlignOperands: DontAlign
+AllowShortBlocksOnASingleLine: true
 ---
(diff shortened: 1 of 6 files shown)


sanchda:
Overall, looks really good.  I did some sleuthing and found that most developers (and style guides) prefer placing dyadic operators at the beginning of a linebreak when a long line is split.  Since cond is written as a train of two dyadic infix operators, I think what clang-format is doing on to `?:` is fine.  I'll make myself feel better about the subject by declaring my attitude on operator-breaks as too _punk rock_ for polite company and conceding in the name of civility.

I'm surprised it formatted the wonky macros so well, that's quite pleasant.

sanchda (approved):
Looking good!

PR #1142, opened 2021-02-04
[Profiler] Add testing setup for pprof/native

morrisonlevi:
### Description

This adds the basic testing setup in CMake for running tests for the `native` folder. It also converts a parameter to take `const char*` instead of `char *` to silence a few warnings.

### DIFF
--- a/src/native/pprof.c
+++ b/src/native/pprof.c
@@ -108,8 +108,9 @@ size_t pprof_stringtable_size(void *state) {
   return ((StringTable *)state)->arena->entry_idx;
 }
 
-size_t pprof_strIntern(DProf *dp, char *str) {
-  return dp->intern_string(dp->string_table_data, str);
+size_t pprof_strIntern(DProf *dp, const char *str) {
+  // todo: change intern_string to accept `const char *` instead of `char *)
+  return dp->intern_string(dp->string_table_data, (char *)str);
 }
 
 /******************************************************************************\
--- a/test/pprof.cc
+++ b/test/pprof.cc
@@ -1,0 +1,93 @@
+#include <catch2/catch.hpp>
+
+extern "C" {
+#include "native/pprof.h"
+}
+
+const char *empty_string = "";
+
+// cannot use null; use empty_string instead
+struct frame {
+  const char *function;
+  const char *filename;
+  int64_t lineno;
+  uint64_t function_id;
+  uint64_t location_id;
+};
+
+TEST_CASE("Basic PHP frame example", "[php]") {
+  // This case is specifically crafted to find a bug found when prototyping
+  // No bug found in shared library; bug was higher up
+  DProf dprof;
+
+  // set up the dprof obj
+  dprof.table_type = 1;
+  const char *sample_names[2] = {"samples", "wall-time"};
+  const char *sample_units[2] = {"count", "nanoseconds"};
+
+  static_assert(sizeof sample_names == sizeof sample_units,
+                "number of sample names and units must match");
+
+  std::size_t nsampletypes = sizeof sample_units / sizeof *sample_units;
+
+  REQUIRE(pprof_Init(&dprof, sample_names, sample_units, nsampletypes) == 0);
+
+  const char *executable_name = "php";
+  uint64_t mapping =
+      pprof_mapAdd(&dprof, 0, 0, 0, executable_name, empty_string);
+  REQUIRE(mapping == 1);
+
+  {
+    frame call_stack[4] = {
+        {"usleep", "", 0, 1, 1},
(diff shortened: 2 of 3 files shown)


src/native/pprof.c
```diff
@@ -108,8 +108,9 @@ size_t pprof_stringtable_size(void *state) {
   return ((StringTable *)state)->arena->entry_idx;
 }
 
-size_t pprof_strIntern(DProf *dp, char *str) {
-  return dp->intern_string(dp->string_table_data, str);
+size_t pprof_strIntern(DProf *dp, const char *str) {
```
sanchda, line 111: I think this is reasonable, since these functions just copy the strings around anyway.

test/pprof.cc
```diff
@@ -0,0 +1,93 @@
+#include <catch2/catch.hpp>
+
+extern "C" {
+#include "native/pprof.h"
+}
+
+const char *empty_string = "";
+
+// cannot use null; use empty_string instead
```
sanchda, line 9: Right, this is kind of annoying.  I think the correct semantics for the string table are to intern NULL as the empty string.  I'll add it to my to-do (I went back and forth on this issue, so it's possible there's even a commented-out line implementing it).

test/pprof.cc
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    int64_t values[2] = {1, 1};
+    REQUIRE(pprof_sampleAdd(&dprof, values, sizeof values / sizeof *values,
+                            locations, 4) == 0);
+  }
+
+  // todo: what other properties should I assert?
+
+  REQUIRE(pprof_Free(&dprof) == 0);
```
sanchda, line 92: is it weird to you that the pprof_Free() function returns a value?
1.  Releasing a resource shouldn't error out.  I mean, obviously it can, and those situations are definitely bugs in the library--but the interface should ensure that the internal state is at least consistent enough to release.
2. If we can't make the assurance in 1 and it did throw an error, what could the developer possibly do about it?  I guess they could generate a new object and update all references, discarding the previous item as invalid.  Or--and this is basically the same thing--they could memset() over the struct and reuse it with new internal references (or the library could do this).  These DProf objects bind in only one direction, so failure would come from the release of a member rather than some kind of an invalid parent/global state (i.e., if an element were referenced through a global caching layer).

sanchda (approved):
Looks good to me--thanks!

test/pprof.cc
morrisonlevi: Yeah, avoiding it at the lower level forces all callers to deal with it, and if they _don't_ then there is now a failure condition that has to be resolved. If the lower level just handles it then there isn't an error condition anymore, and makes it more convenient for the caller. I think it's the right move, anyway.

test/pprof.cc
morrisonlevi: Yeah, it's weird to me. I don't know what I would do if "free" failed. What do you do if a destructor in C++ fails? You abort, basically.

At the very least it needs signaling for what failure _means_. If it means an underlying call failed and I'm going to leak some memory, okay, thanks for letting me know. Maybe I can create a metric for it or disable the tracer on the next request or something. But without knowing what a failure means I couldn't really do anything.

sanchda: I think so too.  The implementation was a little broken, since it was safeguarding against a nonsensical error state (it was checking to see whether &dp->pprof is null, but that will never happen because struct DProf embeds a PPProfile).  This was a result of careless refactoring.

Fixed that issue and refactored to NULL for a later PR I'll send.

test/pprof.cc
sanchda: One potential issue is if the user sends NULL, but the size is nonzero.  There are two obvious answers.
1.  It's an error.  Return.
2.  It's harmless.  Internally set the string to `""` and the size to 0, no matter the presented size.

It is slightly annoying for a library to be overly forgiving, which tends to delocalize (in code) the presentation of an error from its origin, but in the average case I expect a NULL string to be semantically equivalent to the empty string for the purpose of this library consumers.  I'll go with 2.

morrisonlevi: If the pointer is null and the size is non-zero, then there is definitely a bug in the calling code. To some degree, garbage in, garbage out is reasonable. In certain C libraries I'd advocate for an `assert()` only. The size is authoritative, and if they give us a nullptr with non-zero size then that's a bug in the caller; the assert is just there to help them find their own bugs.

What do you think about having an `assert` for debug builds which would catch the error, but then for production builds it coerces to the empty string in such cases? Trying to get the best of both worlds.

sanchda: That's a great idea.  It's a zero cost safety net and will probably help us catch our own bugs.

PR #1143, opened 2021-02-04
Move container ID detection to the extension level

SammyK:
### Description

This PR moves the container ID parsing and Agent HTTP header serialization from PHP to the extension level. This offers an onboarding improvement for those who have an `open_basedir` INI restriction; those users no longer have to add `/proc/self/cgroup` to the `open_basedir` INI setting.

- This PR was refactored into the component architecture that was added in #1172.
- The `DDTrace\Util\ContainerInfo` class was removed.
- In an effort to begin decoupling PHP and C transports, the Agent HTTP headers are no longer injected into the background sender with the second argument to `dd_trace_send_traces_via_thread()`. That argument is now ignored and the Agent HTTP headers are now generated at the extension level.
- The container ID is exposed to userland via `DDTrace\System\container_id()` for backward compatibility with the userland transport.
- The [original PhpUnit tests](https://github.com/DataDog/dd-trace-php/blob/bab7c70/tests/Unit/Util/ContainerInfoTest.php) for the container ID were converted into C-level unit tests.
- This PR includes running the Catch2 tests with UBSan on Buster.

### Reviewer checklist
- [x] Milestone is set.


ext/php7/ddshared.c
```diff
@@ -0,0 +1,27 @@
...
+#include "ddtrace.h"
+#include "logging.h"
+
+ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
+
+datadog_string *dd_container_id = NULL;
+
+void ddshared_minit(void) {
+    char *cgroup_file = NULL;
+    if (DDTRACE_G(test_cgroup_file) && DDTRACE_G(test_cgroup_file)[0]) {
```
morrisonlevi, line 14: This is in `minit` -- what sets the `DDTRACE_G(test_cgroup_file)` prior to this call?

ext/DatadogShared/container_id.c
```diff
@@ -0,0 +1,98 @@
...
+    for (size_t i = 0; i < dd_targets_len; ++i) {
+        char *pos = strstr(line, dd_targets[i].name);
+        if (pos) {
+            return pos + dd_targets[i].len;
+        }
+    }
+    return NULL;
+}
+
+datadog_string *datadog_container_id(const char *file) {
```
morrisonlevi, line 69: Are these container ids fixed-size? Variable? Known to have a maximum length?

morrisonlevi (commented):
I have some preliminary questions.

ext/php5/comms_php.c
```diff
@@ -1,57 +1,74 @@
...
-    zval converted;
-    TSRMLS_FETCH();
+static void dd_append_header(struct curl_slist **list, const char *key, const char *val) {
+    /* The longest Agent header should be:
+     * Datadog-Container-Id: <64-char-hash>
+     * So 256 should give us plenty of wiggle room.
+     */
+    char header[256];
+    size_t len = snprintf(header, sizeof(header), "%s: %s", key, val);
+    if (len > 0 && len < sizeof(header)) {
```
morrisonlevi, line 24: Nits: when doing `sizeof` on an object (not a type), I prefer dropping the parenthesis e.g. `sizeof header`. This because of the [grammar](https://en.cppreference.com/w/c/language/sizeof):

    sizeof( type )
    sizeof expression

So having the parenthesis on an expression is redundant and potentially confusing e.g. "Is `header` a type?"

ext/DatadogShared/container_id.c
morrisonlevi: According to the HTTP header part of this PR, it looks like this:

    Datadog-Container-Id: <64-char-hash>

Is that a max of 64 chars, or always 64 chars? In either case, I would prefer that we avoid dynamic memory here and simply store a 64 character string (plus length info). Avoiding dynamic memory is part of reliable C standards like MISRA C, and in this case it seems trivial.

ext/php8/comms_php.c
```diff
@@ -1,57 +1,74 @@
...
+     */
+    dd_append_header(list, "Expect", "");
 }
 
-static struct curl_slist *_dd_convert_hashtable_to_curl_slist(HashTable *input) {
-    if (zend_hash_num_elements(input) > 0) {
-        struct curl_slist *list = NULL;
-        ddtrace_array_walk(input, _comms_convert_append, &list);
-        return list;
+static bool dd_memoize_http_headers(void) {
```
morrisonlevi, line 47: Now that this is done at the extension level and all this data is static, we don't need to memoize it nor do atomics; we can generate it when the process starts but before the background sender is started. Clean up the headers after the background thread has been cleaned up in process shutdown.

ext/DatadogShared/container_id.c
```diff
@@ -0,0 +1,98 @@
...
+    const char *cgroup_file = file != NULL ? file : CGROUP_FILE;
+    if (access(cgroup_file, F_OK | R_OK) != 0) {
+        return NULL;
+    }
```
morrisonlevi, line 73: You don't need this; `fopen` will fail if you don't have permissions. Theoretically this would be a race condition opportunity as well.

ext/php8/ddtrace.c
```diff
@@ -67,6 +68,10 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "@php_dir@/datadog_trace/bridge/d
 STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateString, request_init_hook,
                   zend_ddtrace_globals, ddtrace_globals)
 #endif
+
+// Exposed for testing only
+STD_PHP_INI_ENTRY("ddtrace.test_cgroup_file", "", PHP_INI_SYSTEM, OnUpdateString, test_cgroup_file,
+                  zend_ddtrace_globals, ddtrace_globals)
```
morrisonlevi, line 74:
```suggestion
// Exposed for testing only
STD_PHP_INI_ENTRY("ddtrace.cgroup_file", "/proc/self/cgroup", PHP_INI_SYSTEM, OnUpdateString, cgroup_file,
                  zend_ddtrace_globals, ddtrace_globals)
```

I don't particularly love it, but I do think it is an improvement because the default value is now passed in, instead of hard-coded into the implementation.

ext/php5/ddtrace.h
```diff
@@ -61,6 +61,8 @@ ZEND_BEGIN_MODULE_GLOBALS(ddtrace)
     uint32_t open_spans_count;
     uint32_t closed_spans_count;
     int64_t compile_time_microseconds;
+
+    char *test_cgroup_file;
```
morrisonlevi, line 65:
```suggestion
    char *cgroup_file;
```

ext/php7/ddshared.c
```diff
@@ -0,0 +1,27 @@
...
+        cgroup_file = DDTRACE_G(test_cgroup_file);
+        ddtrace_log_debugf("Using test cgroup file: %s", cgroup_file);
+    }
+    dd_container_id = datadog_container_id(cgroup_file);
+}
```
morrisonlevi, line 19:
```suggestion
void ddshared_minit(void) {
    if (DDTRACE_G(cgroup_file) && DDTRACE_G(cgroup_file)[0]) {
        dd_container_id = datadog_container_id(DDTRACE_G(cgroup_file));
    }
}
```

This will break your tests, sorry.

tests/ext/background-sender/agent_headers_container_id.phpt
```diff
@@ -0,0 +1,43 @@
...
+foreach ($headers as $name => $value) {
+    echo $name . ': ' . $value . PHP_EOL;
+}
+echo PHP_EOL;
+
+echo 'Done.' . PHP_EOL;
+?>
+--EXPECTF--
+Using test cgroup file: %s/stubs/cgroup.docker
+Successfully memoized Agent HTTP headers
```
morrisonlevi, line 37:
```suggestion
```

tests/ext/background-sender/agent_headers_container_id_empty.phpt
```diff
@@ -0,0 +1,42 @@
...
+foreach ($headers as $name => $value) {
+    echo $name . ': ' . $value . PHP_EOL;
+}
+echo PHP_EOL;
+
+echo 'Done.' . PHP_EOL;
+?>
+--EXPECTF--
+Using test cgroup file: %s/stubs/cgroup.empty
+Successfully memoized Agent HTTP headers
```
morrisonlevi, line 37:
```suggestion
```

tests/ext/background-sender/agent_headers_ignore_userland.phpt
```diff
@@ -0,0 +1,40 @@
...
+]);
+foreach ($headers as $name => $value) {
+    echo $name . ': ' . $value . PHP_EOL;
+}
+echo PHP_EOL;
+
+echo 'Done.' . PHP_EOL;
+?>
+--EXPECTF--
+Successfully memoized Agent HTTP headers
```
morrisonlevi, line 35:
```suggestion
```

tests/ext/background-sender/agent_headers.phpt
```diff
@@ -0,0 +1,48 @@
...
+]);
+foreach ($headers as $name => $value) {
+    echo $name . ': ' . $value . PHP_EOL;
+}
+echo PHP_EOL;
+
+echo 'Done.' . PHP_EOL;
+?>
+--EXPECTF--
+Successfully memoized Agent HTTP headers
```
morrisonlevi, line 38:
```suggestion
```

ext/php5/comms_php.c
SammyK: I mess this up all the time. Thanks for catching it! :)

ext/php7/ddshared.c
SammyK: This happens in GINIT so it's safe to access in MINIT.

ext/DatadogShared/container_id.c
SammyK: Excellent point. This will also allow us to remove `datadog_string` for now.

ext/php8/comms_php.c
SammyK: Nice catch. This will greatly simplify Agent HTTP header initialization.

ext/php8/ddtrace.c
SammyK: I agree this is an improvement. Hopefully we'll have a better option for testing this from the embed SAPI once we get that infrastructure set up.

ext/php7/coms.c
```diff
@@ -1,5 +1,6 @@
 #include "coms.h"
 
+#include <SAPI.h>
```
morrisonlevi, line 3: Is this needed for something or did it sneak in here? Not seeing any new usage in the diff.

SammyK: This is required for `sapi_module.name` which is used in the `Datadog-Meta-Lang-Interpreter` header.

morrisonlevi: I wonder where fields like these are documented. I thought that field was things like ZendEngine vs HHVM.

SammyK: Here's where it's [documented in code](https://github.com/DataDog/datadog-agent/blob/05a509ebe00488db75969ca980f23ac21a685fb2/pkg/trace/api/api.go#L334-L336). I believe the SAPI was chosen as a best fit for this header since we don't support anything but the Zend Engine. But you're right, `Interpreter` isn't what the SAPI is. :)

ext/php5/ddshared.c
```diff
@@ -0,0 +1,13 @@
+#include "ddshared.h"
+
+#include "container_id/container_id.h"
+#include "ddtrace.h"
+#include "logging.h"
```
morrisonlevi, line 5: I'm tired, so maybe it's just me, but I don't see any use of logging in this file. I think it can be removed here and in the versions too.

morrisonlevi (approved):

morrisonlevi (changes requested):
Oops, didn't mean to approve on my previous comment.

ext/php5/ddshared.c
SammyK: Nice catch. This was left over from the old log message.

morrisonlevi (approved):

PR #1144, opened 2021-02-05
add missing dependency on curl extension for PHP 8

remicollet:
Without **curl** extension, extension load fails with "undefined symbol: curl_multi_ce" (PHP 8 only , build with --enable-rtld-now) so will segfault on standard build when curl extension is build shared and not loaded.

PHP 5 and 7 are not affected as this symbol is not used


SammyK (approved):
Thanks for the fix @remicollet! 💯

PR #1151, opened 2021-02-16
[shared libs] Formatting updates, correctness, API

sanchda:
* Applied clang-format as per the repo style.  Some of the line splits
  are horrible--they can be cleaned up in a later commit
* Remove a few commented blocks, like the pprof.c forward-decl useless
  comment block
* Add new pprof string getter function, pprof_getstr(), which will use
  the string table implementation specific to a pprof to pull out a
  pointer to its internal symbol table
* Add new pprof_print() function, a terse, incomplete printer for pprof.
  This is necessary for debugging and because go tool pprof doesn't
  always print e.g., function names that are compliant for our platform
  and for the protobuf format, but not for Go specifically
* Did we really not have pprof_flush() before?  We do now.  A later pr
  will demonstrate an example, but this should abstrat away the
  implementation-specific stuff, like generating a char** for the string
  table
* Add const qualifier in pprof.h whenever appropriate
* pprof_funAdd() pretty-prints various kinds of missing fields with "??"
* pprof_Init() has a concept of self-ownership--now returning a pointer
 to a DProf object rather than a return code, in order to allow the
 system to allocate and manage such objects entirely.
* string_table.h sanitizes input better--NULL pointers are now the empty
  string on internment.


src/native/pprof.c
```diff
@@ -455,31 +495,40 @@ void pprof_durationUpdate(DProf *dp) {
       dp, (tv.tv_sec * 1000 * 1000 + tv.tv_usec) * 1000 - dp->pprof.time_nanos);
 }
 
-char pprof_Init(DProf *dp, const char **sample_names, const char **sample_units,
-                size_t n_sampletypes) {
+DProf *pprof_Init(DProf *dp, const char **sample_names,
```
morrisonlevi, line 498: Why does this still take a `DProf *` as a parameter? I think I'm missing something.

sanchda: It's been updated to match the documentation (haa....), which is that passing a NULL pointer in the first position causes the library to heap-allocate the container and manage (free) it during pprof_free().  This is slightly inconvenient and I don't think the interface is well-suited for frequent init/free sequences since most of the inner members are heap-allocated growing arrays (which would justify the BYO-container interface), but I'd like to leave the door open a while longer for caller-provided containers, unless there are strong feelings otherwise.

sanchda:
Closing this PR because it's been completely absorbed into #1152

PR #1153, opened 2021-02-22
Add the possibility to freeze a regression in a specific randomized tests scenario

labbati:
### Description

This PR adds the possibility to freeze specific scenarios as regressions and to execute those regressions at every commit.

A note to the reviewer. All files in `tests/randomized/regressions` are collapsed by default in Github to facilitate code review. Now, and in the future, my advice is to just make sure that they are not empty. `make analyze` will make sure to fail if no tests have been run for a regression with successful results (https://github.com/DataDog/dd-trace-php/pull/1153/files#diff-28a4837595b4bca57b0799d3d7fd12a2cec420cc0ca47b4a424cedebc3a69da8R68).

### Readiness checklist
- ~[ ] (only for Members) Changelog has been added to the release document.~

### Reviewer checklist
- [x] Milestone is set.


tests/randomized/lib/EnvFileGenerator.php
```diff
@@ -0,0 +1,18 @@
...
+    public function generate($destination, $scenarioName)
+    {
+        file_put_contents(
+            $destination,
+            \str_replace(
+                ['{{scenario_name}}'],
+                [$scenarioName],
+                file_get_contents(__DIR__ . '/templates/.env.template')
+            )
+        );
```
SammyK, line 16: This pattern appears a number of times. Perhaps there could be a generic template generator akin to the generic substitutions supported by `DockerComposeFileGenerator::generate()`?

Not a blocker for merge, but just a thought for later if you agree. :)

SammyK (approved):
LGTM 👍 I just left one comment that's not a blocker for merge. Also, I didn't comb through all the regression files. Is there something I should be looking out for other than them not being empty?

Nice work @labbati! :)

tests/randomized/lib/EnvFileGenerator.php
labbati: This is a very good point. I merged this PR so we can provide value sooner in master, but I also open #1155 so we can iterate there over this feedback. Let me know what do you think and let's continue the conversation there.

PR #1160, opened 2021-02-28
[Tracer] Add runtime id, components, and plugins

morrisonlevi:
### Description

Adds top-level directories components and plugins. Components have
a single public header and their own tests. Plugins are a type of
component that are coupled to the PHP lifecycle.

The runtime id is used to associate traces to other data. It is
unique per process and should be regenerated after a fork. The
profiler will use the runtime id to associate profiles to traces.

This adds a dependency on libsodium for generating random bytes.
I expect many CI jobs will fail because of this.

This bumps the minimum version of CMake required to 3.13. The
specific feature needed is:
 - The install(TARGETS) command learned to install targets created
   outside the current directory.

This is used to install all components from the common
components/CMakeLists.txt.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -1,0 +1,52 @@
+# Need CMake 3.13 for installing targets from subdirectories
+cmake_minimum_required(VERSION 3.13)
+
+project(DatadogPHPComponents
+  VERSION 0.1.0
+  LANGUAGES C
+)
+
+option(BUILD_COMPONENTS_TESTING "Enable tests" OFF)
+if (${BUILD_COMPONENTS_TESTING})
+  # Tests uses the C++ testing framework Catch2
+  enable_language(CXX)
+
+  # The Catch2::Catch2 target has been available since 2.1.2
+  # dd-trace-cpp uses Catch2 2.4 at this time, so we can require at least 2.4
+  find_package(Catch2 2.4 REQUIRED)
+
+  #[[ This file takes a while to build, so we do it once here and every test
+      executable can link to it to save time.
+  ]]
+  add_library(catch2_main catch2_main.cc)
+  target_link_libraries(catch2_main PUBLIC Catch2::Catch2)
+  target_compile_features(catch2_main PUBLIC cxx_std_11)
+
+  include(Catch)
+  enable_testing()
+endif()
+
+include(GNUInstallDirs)
+
+add_subdirectory(string_view)
+add_subdirectory(uuid)
+
+add_library(DatadogPHPComponents INTERFACE)
+
+target_link_libraries(DatadogPHPComponents
+  INTERFACE DatadogPHPStringView DatadogPHPUUID
+)
+
+install(
+  TARGETS DatadogPHPComponents DatadogPHPStringView DatadogPHPUUID
+  EXPORT DatadogPHPComponentsTargets
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+)
+
+install(EXPORT DatadogPHPComponentsTargets
+  FILE DatadogPHPComponentsTargets.cmake
+  NAMESPACE Datadog::PHP::
+  DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake
+)
--- a/components/string_view/CMakeLists.txt
+++ b/components/string_view/CMakeLists.txt
@@ -1,0 +1,29 @@
+add_library(DatadogPHPStringView string_view.c)
+
+target_include_directories(DatadogPHPStringView
+  PUBLIC
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+    $<INSTALL_INTERFACE:include>
+)
+
+target_compile_features(DatadogPHPStringView
+  PUBLIC c_std_99
+)
+
+set_target_properties(DatadogPHPStringView PROPERTIES
+  OUTPUT_NAME datadog_php_string_view
+  VERSION ${PROJECT_VERSION}
+)
+
+add_library(Datadog::PHP::StringView
+  ALIAS DatadogPHPStringView
+)
+
+if (${BUILD_COMPONENTS_TESTING})
+  add_subdirectory(test)
+endif()
+
+# This copies the include files when `install` is ran
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/string_view.h
+	DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/string_view/
+)
--- a/components/string_view/string_view.h
+++ b/components/string_view/string_view.h
@@ -1,0 +1,40 @@
+#ifndef DATADOG_PHP_STRING_VIEW_H
+#define DATADOG_PHP_STRING_VIEW_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+/**
+ * A string view is a non-owning view into another string. The original string
+ * should not be changed from the view.
+ */
+typedef struct datadog_php_string_view {
+    size_t len;
+    const char *ptr;
+} datadog_php_string_view;
+
+/* Used to initialize an empty string e.g.
+ *   datadog_php_string_view str = DATADOG_PHP_STRING_VIEW_INIT;
+ */
+#define DATADOG_PHP_STRING_VIEW_INIT \
+    { 0, NULL }
+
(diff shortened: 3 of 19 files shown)


morrisonlevi:
It may be helpful to see the tree structure of components:

```
$ tree components
components
├── CMakeLists.txt
├── catch2_main.cc
├── string_view
│   ├── CMakeLists.txt
│   ├── string_view.c
│   ├── string_view.h
│   └── test
│       ├── CMakeLists.txt
│       └── string_view.cc
└── uuid
    ├── CMakeLists.txt
    ├── test
    │   ├── CMakeLists.txt
    │   └── uuid.cc
    ├── uuid.c
    └── uuid.h
```

The `components/CMakeLists.txt` builds all the components. Currently the CMake build is only used for tests; the `config.m4` was updated to build `string_view` and `uuid` for the PHP tooling. I think maybe someday in the future we can bridge these.

components/string_view/string_view.h
```diff
@@ -0,0 +1,40 @@
+#ifndef DATADOG_PHP_STRING_VIEW_H
```
labbati, line 1: should we use the same approach used by @SammyK in https://github.com/DataDog/dd-trace-php/pull/1143/files#diff-b291ab1c1e6a83590ba90d3ee0dbe0167d0fe987786f719eed763ab82434f034R1

TLDR, adding headers to `./datadog` and access them via `#include <datadog/header_file.h>`?

Which one you prefer?

components/uuid/uuid.c
```diff
@@ -0,0 +1,131 @@
+#include "uuid.h"
```
labbati, line 1: If, I assume you copied and pasted this snippet, can we had a link to the exact hash on GitHub for the project? So we have a direct reference to the license at this point in time.

labbati (commented):
I took a first look and the idea seems very promising.
I added a couple of questions but I am taking now this conversation offline to coordinate our efforts for container ids as well.

plugins/runtime_id_plugin/runtime_id_plugin.c
```diff
@@ -0,0 +1,36 @@
+#include "runtime_id_plugin.h"
+
+#include <sodium.h>
```
SammyK, line 3: From our conversation from earlier, I think we need to vendor this in for the short term.

plugins/runtime_id_plugin/runtime_id_plugin.c
```diff
@@ -0,0 +1,36 @@
...
+
+#include "uuid/uuid.h"
+
+datadog_php_uuid runtime_id = DATADOG_PHP_UUID_INIT;
+
+typedef enum datadog_php_plugin_result plugin_result_t;
+
+static plugin_result_t datadog_php_runtime_id_plugin_minit(int type, int module_number) {
+    (void)type;
+    (void)module_number;
```
SammyK, line 13: Do you think there is any value in passing these args in at all?

plugins/runtime_id_plugin/runtime_id_plugin.c
```diff
@@ -0,0 +1,36 @@
...
+
+    datadog_php_uuidv4_bytes_ctor(&runtime_id, bytes);
+
+    return DATADOG_PHP_PLUGIN_SUCCESS;
+}
+
+static void datadog_php_runtime_id_plugin_mshutdown(int type, int module_number) {
+    (void)type;
+    (void)module_number;
+    randombytes_close();
```
SammyK, line 30: Do we need to wait until MSHUTDOWN to call this or can we call it right after the `randombytes_buf()` call? From [the docs](https://libsodium.gitbook.io/doc/generating_random_data) it seems like we might not need to call it at all?

> This deallocates the global resources used by the pseudo-random number generator. More specifically, when the `/dev/urandom` device is used, it closes the descriptor. Explicitly calling this function is almost never required.

morrisonlevi: My worry with closing it earlier would possibly be closing `/dev/urandom` prematurely for PHP's sodium ext. Since init and close are global (no resource handle directly exposed to users), I could theoretically mess it up for others.

I have not tested this, though; just a theory.

plugins/runtime_id_plugin/runtime_id_plugin.c
morrisonlevi: I believe `type` is TEMPORARY or PERMANENT -- we definitely need that, don't we?

components/uuid/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+add_library(DatadogPHPUUID uuid.c)
```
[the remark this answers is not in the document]
morrisonlevi: Do you feel the same way with `::` between them? `Datdog::Php:Uuid`? These names don't have to match. We could do `datadog_php_uuid` (actually I kinda like that).

components/string_view/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+add_library(DatadogPHPStringView string_view.c)
+
+target_include_directories(DatadogPHPStringView
+  PUBLIC
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+    $<INSTALL_INTERFACE:include>
+)
+
+target_compile_features(DatadogPHPStringView
+  PUBLIC c_std_99
```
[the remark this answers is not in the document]
morrisonlevi: No, it just doesn't need C11 and so I didn't write it that way. I put the lowest version I can reasonably think of supporting out of habit.

components/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+  target_compile_features(catch2_main PUBLIC cxx_std_11)
+
+  include(Catch)
+  enable_testing()
+endif()
+
+include(GNUInstallDirs)
+
+add_subdirectory(string_view)
+add_subdirectory(uuid)
```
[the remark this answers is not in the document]
morrisonlevi: We can add interface and target stuff from each component, I think. If true then we can invert the relationship and do it programmatically. Maybe we can try it once we have 3+ components?

ext/php7/ddtrace.c
```diff
@@ -348,6 +357,31 @@ static PHP_MINIT_FUNCTION(ddtrace) {
...
+                    break;
+
+                case DATADOG_PHP_PLUGIN_FAILURE:
+                    plugin_enabled[i] = false;
+                    break;
+
+                case DATADOG_PHP_PLUGIN_DISABLE_PLUGIN:
+                    plugin_enabled[i] = false;
+                    // todo: what else to disable the whole ext?
+                    return FAILURE;
```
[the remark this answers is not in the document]
morrisonlevi: I think it was supposed to be `DATADOG_PHP_PLUGIN_DISABLE_EXT`, 🤦 .

components/uuid/CMakeLists.txt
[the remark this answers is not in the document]
SammyK: I like the snake case version as well: `datadog_php_uuid`. 👍

plugins/runtime_id_plugin/runtime_id_plugin.c
SammyK: Yeah, `MODULE_TEMPORARY` means the module has been loaded with `dl()` which we don't support anyway. So it might actually make sense to add a check for this in the main MINIT similar to the check we do to [test the observer API](https://github.com/php/php-src/blob/803779e/ext/zend_test/test.c#L388-L396) and fail early. Then we can ignore the args altogether. What do you think?

plugins/runtime_id_plugin/runtime_id_plugin.c
SammyK: Ah, good point. Do you think it is safe to call it at all then? I'm trying to think of a scenario where another extension uses sodium in their MSHUTDOWN after ours fires or something.

components/uuid/uuid.c
morrisonlevi: Good idea.

plugins/runtime_id_plugin/runtime_id_plugin.c
morrisonlevi: This is exactly why global shared state is bad. The docs do not say if it's reference counted or anything, so I don't know the answer! We'll have to look at the code.

krakjoe:
@morrisonlevi it looks like we moved on from here, is it okay to close this, or is some of it still salvageable/applicable ?

morrisonlevi:
Sorry for the slow response. Code Hotspots has shipped so there's no need to keep this around anymore, even as a reference.

PR #1166, opened 2021-03-03
[Profiler] Purge C++ code; restructure accordingly

morrisonlevi:
This removes now-unused C++ code. It restructures the repository accordingly.

It also cleans up the CMakeLists.txt (I think), and fixes an ownership bug (I think, I couldn't get my debugger to step into my code for some reason).


src/pprof.c
```diff
@@ -520,6 +520,8 @@ DProf *pprof_Init(DProf *dp, const char **sample_names,
     if (!dp)
       return NULL;
     dp->ownership = 1;
+  } else {
+    dp->ownership = 0;
```
sanchda, line 524: Right, this is a good point.  I'll have to update the use of this pattern elsewhere--thanks!

src/proto/.clang-format
```diff
@@ -0,0 +1,2 @@
+DisableFormat: true
```
sanchda, line 1: Oh wow, I didn't know you could do this!

sanchda (approved):
Looks good to me!

Issue #1158, opened 2021-02-25
[BUG] Odd behavior in forked CLI processes


mucha55:
### Bug description
Hello, we are experiencing some odd behavior in CLI processes that use `pcntl_fork()` with the `ddtrace` extension installed. This is in an environment where tracing is enabled for HTTP (PHP-FPM), but disabled for CLI ([due to an issue submitted by my colleague](https://github.com/DataDog/dd-trace-php/issues/1101)). Top-level processes behave well, but any forked processes take a consistent 5 seconds to exit after executing their code ([barebones example found here](https://gist.github.com/mucha55/3cd73d3258e5e9b175c2436fbedeaec7)). Tweaking the configuration (with full re-deployments in between) or manually flushing/finishing/closing spans and scopes has no effect on that delay; the only way to get forked processes to behave is by fully disabling (`ddtrace.disable=On`) or uninstalling the extension (`dpkg -r datadog-php-tracer`). Is there any way for us to remove this exit delay from forked CLI processes without affecting HTTP tracing?

### PHP version
```
PHP 7.4.11 (cli) (built: Oct  1 2020 19:44:09) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.11, Copyright (c), by Zend Technologies
    with ddtrace v0.51.0, Copyright Datadog, by Datadog
```

### Installed extensions
```
[PHP Modules]
bcmath
Core
ctype
curl
date
ddtrace
dom
fileinfo
filter
ftp
gd
hash
iconv
imap
json
libxml
mbstring
memcached
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_pgsql
pdo_sqlite
Phar
posix
readline
redis
Reflection
session
SimpleXML
soap
sockets
sodium
SPL
sqlite3
standard
sysvmsg
sysvsem
sysvshm
tokenizer
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib
zookeeper

[Zend Modules]
Zend OPcache
ddtrace
```

### OS info
```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
```

#### Output of phpinfo() (ddtrace >= 0.47.0)
```
ddtrace


Datadog PHP tracer extension
[...]

PR #1168, opened 2021-03-05
Disable tracer for good when DD_TRACE_ENABLED=false or cli and DD_TRACE_CLI_ENABLED=false

labbati:
### Description

Before this PR, even when tracing was disabled, a few things happened, including spawning the background sender thread and initializing the internal handlers.

This is not only problematic in cases like the ones uncovered by https://github.com/DataDog/dd-trace-php/issues/1158, but it is also inconvenient because it increases the risks that we take when our tracer is installed but not used.

This PR applies the concept: "exit as soon as you can doing the least possible work".

**DRAFT**: this is still in draft, as more tests have to be added and some existing tested have to be adapted, but it is ready enough to receive feedback about the approach.

Relates to #1158.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php7/ddtrace.c
+++ b/ext/php7/ddtrace.c
@@ -63,6 +63,10 @@ STD_PHP_INI_ENTRY("ddtrace.request_init_hook", "", PHP_INI_SYSTEM, OnUpdateStrin
 PHP_INI_END()
 
 static int ddtrace_startup(struct _zend_extension *extension) {
+    if (DDTRACE_G(disable) = 1) {
+        return SUCCESS;
+    }
+
     ddtrace_resource = zend_get_resource_handle(extension);
 #if PHP_VERSION_ID >= 70400
     ddtrace_op_array_extension = zend_get_op_array_extension_handle();
@@ -75,6 +79,10 @@ static int ddtrace_startup(struct _zend_extension *extension) {
 
 static void ddtrace_shutdown(struct _zend_extension *extension) {
     UNUSED(extension);
+
+    if (DDTRACE_G(disable) = 1) {
+        return;
+    }
 
     ddtrace_internal_handlers_shutdown();
 }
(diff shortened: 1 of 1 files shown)


morrisonlevi (commented):
A little while ago I suggested that we remove `DD_TRACE_CLI_ENABLED` and just use `DD_TRACE_ENABLED`, and just have a different default value for it depending on whether it's the CLI SAPI or not. Do you think we can work that into this PR? I'm asking mainly because I don't like having to think about even more edges like `DD_TRACE_ENABLED=false` and `DD_TRACE_CLI_ENABLED=true`.

ext/php7/ddtrace.c
```diff
@@ -251,24 +259,50 @@ static void dd_register_fatal_error_ce(TSRMLS_D) {
...
 static PHP_MINIT_FUNCTION(ddtrace) {
     UNUSED(type);
     REGISTER_STRING_CONSTANT("DD_TRACE_VERSION", PHP_DDTRACE_VERSION, CONST_CS | CONST_PERSISTENT);
     REGISTER_INI_ENTRIES();
 
     // config initialization needs to be at the top
     ddtrace_initialize_config(TSRMLS_C);
-    _dd_disable_if_incompatible_sapi_detected(TSRMLS_C);
+
+    dd_disable_by_configuration(TSRMLS_C);
```
SammyK, line 302: I don't think this will work as expected on the FPM and Apache SAPIs. Those SAPIs don't have access to the environment vars until the [first RINIT](https://github.com/DataDog/dd-trace-php/blob/f09dd36/ext/php7/ddtrace.c#L351-L355). INI settings do not have this same restriction so they can be reliably accessed in MINIT/startup for all SAPIs.

What are your thoughts on deprecating `DD_TRACE_ENABLED` and `DD_TRACE_CLI_ENABLED` in favor of exclusively using INI settings to disable the tracer (e.g. `ddtrace.disable=1`)? On the CLI SAPI `ddtrace.disable` could default to `1`, and all other SAPIs the default would be `0`.

labbati:
> A little while ago I suggested that we remove `DD_TRACE_CLI_ENABLED` and only use `DD_TRACE_ENABLED`, then have a different default value for it depending on whether it's the CLI SAPI or not. Do you think we can work that into this PR? I'm asking mainly because I don't like having to think about even more edges like `DD_TRACE_ENABLED=false` and `DD_TRACE_CLI_ENABLED=true`.

We should think this through. It is not uncommon to see in `www.conf` the enabling flag passed on from the host. E.g. `env[DD_TRACE_ENABLED] = $DD_TRACE_ENABLED` - especially if you have multiple pools - and then use the env from the host (or docker container) as a global switch.

I agree that the way we do it today can be confusing, though! I do not have a good idea to be honest.

labbati:
We decided to drop this change for now

PR #1169, opened 2021-03-05
[Profiler] Fix CMake build for dd_send

morrisonlevi:
### Description

This fixes some build issues and fixes an ownership bug in DDR_init.

### DIFF
--- a/src/dd_send.c
+++ b/src/dd_send.c
@@ -45,6 +45,8 @@ DDReq *DDR_init(DDReq *req) {
     if (!req)
       return NULL;
     req->ownership = 0x01;
+  } else {
+    req->ownership = 0;
   }
 
   req->as_header = as_init(NULL);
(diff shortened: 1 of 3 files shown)


src/dd_send.c
```diff
@@ -188,7 +190,7 @@ int DDR_finalize(DDReq *req) {
     return DDRC_EINVAL;
   //  as_sprintf(req->as_header, "POST http://%s%s HTTP/1.1\r\n", req->host,
   //             "/v1/input");
-  as_sprintf(req->as_header, "POST %s HTTP/1.1\r\n", "/v1/input");
+  as_sprintf(req->as_header, "POST %s HTTP/1.1\r\n", "/profiling/v1/input");
```
morrisonlevi, line 193: David, if this needs to be conditional let's sort it out in a separate PR. Thanks!

sanchda: Ah, yeah--this is going to break intake for demo applications depending on direct backend access, but that's not a big deal.  Looks good.

PR #1172, opened 2021-03-09
[Tracer] Add component system, string_view and sapi components

morrisonlevi:
### Description

Adds top-level directory components. Components have a single
public header and their own tests.

This bumps the minimum version of CMake required to 3.14. The
specific feature needed is:

> The install(TARGETS) command learned how to install to an
> appropriate default directory for a given target type, based on
> variables from the GNUInstallDirs module and built-in defaults,
> in lieu of a DESTINATION argument.

This is so each component doesn't have to copy/paste the same config
and keep it in sync.

This also drops the older Catch2 version; it was not being done
for any specific reason.

Put bad test on PHP 5.4 into xfail; already xfail this on PHP 5.5.
php/php-src@471540d

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/components/sapi/sapi.c
+++ b/components/sapi/sapi.c
@@ -1,0 +1,33 @@
+#include "sapi/sapi.h"
+
+#include <string.h>
+
+typedef datadog_php_sapi sapi_t;
+typedef datadog_php_string_view string_view_t;
+
+#define SV(cstr) DATADOG_PHP_STRING_VIEW_LITERAL(cstr)
+
+sapi_t datadog_php_sapi_from_name(string_view_t module) {
+    struct {
+        string_view_t str;
+        sapi_t type;
+    } sapis[] = {
+        {SV("apache2handler"), DATADOG_PHP_SAPI_APACHE2HANDLER},
+        {SV("cgi-fcgi"), DATADOG_PHP_SAPI_CGI_FCGI},
+        {SV("cli"), DATADOG_PHP_SAPI_CLI},
+        {SV("cli-server"), DATADOG_PHP_SAPI_CLI_SERVER},
+        {SV("embed"), DATADOG_PHP_SAPI_EMBED},
+        {SV("fpm-fcgi"), DATADOG_PHP_SAPI_FPM_FCGI},
+        {SV("litespeed"), DATADOG_PHP_SAPI_LITESPEED},
+        {SV("phpdbg"), DATADOG_PHP_SAPI_PHPDBG},
+    };
+
+    unsigned n_sapis = sizeof sapis / sizeof *sapis;
+    for (unsigned i = 0; i != n_sapis; ++i) {
+        if (datadog_php_string_view_equal(module, sapis[i].str)) {
+            return sapis[i].type;
+        }
+    }
+
+    return DATADOG_PHP_SAPI_UNKNOWN;
+}
--- a/components/sapi/sapi.h
+++ b/components/sapi/sapi.h
@@ -1,0 +1,20 @@
+#ifndef DATADOG_PHP_SAPI_H
+#define DATADOG_PHP_SAPI_H
+
+#include "string_view/string_view.h"
+
+typedef enum {
+    DATADOG_PHP_SAPI_UNKNOWN = 0,
+    DATADOG_PHP_SAPI_APACHE2HANDLER,
+    DATADOG_PHP_SAPI_CGI_FCGI,
+    DATADOG_PHP_SAPI_CLI,
+    DATADOG_PHP_SAPI_CLI_SERVER,
+    DATADOG_PHP_SAPI_EMBED,
+    DATADOG_PHP_SAPI_LITESPEED,
+    DATADOG_PHP_SAPI_FPM_FCGI,
+    DATADOG_PHP_SAPI_PHPDBG,
+} datadog_php_sapi;
+
+datadog_php_sapi datadog_php_sapi_from_name(datadog_php_string_view module);
+
+#endif  // DATADOG_PHP_SAPI_H
--- a/components/string_view/string_view.h
+++ b/components/string_view/string_view.h
@@ -1,0 +1,41 @@
+#ifndef DATADOG_PHP_STRING_VIEW_H
+#define DATADOG_PHP_STRING_VIEW_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+/**
+ * A string view is a non-owning view into another string. The original string
+ * should not be changed from the view.
+ */
+typedef struct datadog_php_string_view {
+    size_t len;
+    const char *ptr;
+} datadog_php_string_view;
+
+/* Used to initialize an empty string e.g.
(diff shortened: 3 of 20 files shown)


components/sapi/sapi.h
```diff
@@ -0,0 +1,20 @@
...
+    DATADOG_PHP_SAPI_UNKNOWN = 0,
+    DATADOG_PHP_SAPI_APACHE2HANDLER,
+    DATADOG_PHP_SAPI_CGI_FCGI,
+    DATADOG_PHP_SAPI_CLI,
+    DATADOG_PHP_SAPI_CLI_SERVER,
+    DATADOG_PHP_SAPI_EMBED,
+    DATADOG_PHP_SAPI_LITESPEED,
+    DATADOG_PHP_SAPI_FPM_FCGI,
+    DATADOG_PHP_SAPI_PHPDBG,
+} datadog_php_sapi;
```
SammyK, line 16: Do you foresee any issues/things breaking with adding more SAPIs in the future? For example if we wanted to add the fuzzer SAPI or any of the pre-PHP 7 SAPIs?

components/sapi/sapi.c
```diff
@@ -0,0 +1,34 @@
+#include "sapi/sapi.h"
+
+#include <string.h>
+
+typedef datadog_php_sapi sapi_t;
+typedef datadog_php_string_view string_view_t;
+
+#define STRING_VIEW_LITERAL(cstr) \
+    { sizeof(cstr) - 1, cstr }
```
SammyK, line 9: Was this left over before creating `DATADOG_PHP_STRING_VIEW_LITERAL`?

components/string_view/string_view.h
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Creates a string view from a C string, which is an array of char which is
+ * terminated by a null byte. Derives the length from strlen. `cstr` may be
+ * null, in which case the `.len` will be 0.
+ */
+datadog_php_string_view datadog_php_string_view_from_cstr(const char cstr[]);
+
+/**
+ * Compares the views `a` and `b` for equality. Note that the `.ptr` value of
+ * the views will be ignored when `.len` is 0.
```
SammyK, line 37: > Note that the `.ptr` value of the views will be ignored when `.len` is 0.

I don't understand this part. I see where the `.len`s are being compared, but I don't see a special case for `0`.

SammyK (commented):
Components look awesome! Thanks for figuring out a way that doesn't require maintaining the main CMake file with every new component. I just left a few questions.

SammyK:
Also don't forget labels, milestone, changelog. :)

components/sapi/sapi.h
morrisonlevi: We just have to add them. Approving specific SAPIs seems better than rejecting specific ones as SAPIs seem to come and go (though mostly go lately).

components/sapi/sapi.c
morrisonlevi: Yes, but it shortens it quite a bit ahaha. Think I should remove it?

package.xml
```diff
@@ -39,6 +39,21 @@
...
     <contents>
         <dir name="/">
+            <!-- components -->
+            <!-- todo: what to do with CMakeLists.txt?
+            <file name="components/CMakeLists.txt" role="src" />
+            <file name="components/sapi/CMakeLists.txt" role="src" />
+            <file name="components/sapi/tests/CMakeLists.txt" role="test" />
+            <file name="components/string_view/CMakeLists.txt" role="src" />
+            <file name="components/string_view/tests/CMakeLists.txt" role="test" />
+            -->
```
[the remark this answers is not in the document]
morrisonlevi: I'm unsure about that. The tests (which are in package.xml because they aren't the build script complains about missing files) are basically useless without them. But of course the pecl runner won't be able to do anything with them... so I don't know.

components/sapi/sapi.c
SammyK: I'd love to have a shorter macro for this since we'd use it similarly to `ZEND_STRL` in a lot of places. Could we possibly export a shorter version of the macro and use that instead? I can't think of a really solid short name... `DATADOG_PHP_STRL` maybe? 

What are your thoughts on having a "short" version of the `DATADOG_PHP_` namespace for special cases like this one? Something like `DDPHP_`. Then the macro could be `DDPHP_STRL`.

components/string_view/string_view.h
SammyK: Got it. Thanks for clarifying.

components/sapi/sapi.c
morrisonlevi: If we make a shorter symbol public, like `DDPHP_`, then we may as well just use it everywhere. All of my typedefs and other shortcuts are always in `.c` files so they aren't exposed.

C makes this a pain; every modern language has a solution for this, and C hasn't moved an inch; the name you see is the symbol name you get (except the compiler may or may not put an underscore in front of it).

Maybe we could use a macro to shorten symbols? Something like `DATADOG_PHP_SHORTEN(STRING_VIEW_LITERAL)` which defines a macro in the current file which maps `STRING_VIEW_LITERAL` to `DATADOG_PHP_STRING_VIEW_LITERAL`?

Not sure the macro is better than just writing each one ad-hoc...

package.xml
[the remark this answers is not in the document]
SammyK: Yeah, I don't see any value in shipping the CMake files or the Catch2 tests with PECL. I think it might be best to exclude the `components/*/tests/*.cc` files from the `verify_pecl_file_definitions` target in the Makefile. What do you think?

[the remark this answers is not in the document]
morrisonlevi: Ha, I just realized that since we've backed out every attempt to add C++ files, we don't actually check that `.cc` are in PECL. This should work as long as we never add C++... which sounds good to me! (Go Rust! 🚀 )

components/sapi/sapi.c
morrisonlevi: We decided that doing typedef and defining macros in each file that wants them is what we should do for now. Generally chop off `datadog_php` or `DATADOG_PHP` and append `_t` to types. Not sure how to define rules for things like `SV` -> `DATADOG_PHP_STRING_VIEW_LITERAL`.

Problem goes away if we move it to Rust, ha 🚀 .

SammyK (approved):
LGTM 👍 Thanks @morrisonlevi!

labbati:
@morrisonlevi what does change for our supported OSes regarding this comment?
> This bumps the minimum version of CMake required to 3.14. The
specific feature needed is:
I guess nothing, but any risk we can think of ahead of time? I am talking about production machines (so pecl or building from sources the final release), not contributors' dev machines.

morrisonlevi:
At the moment it does not affect users as the CMake is only used for testing. This may change in the future, of course, but doesn't harm users today. CMake provides binary releases as well, so updating is not hard if it ever came to it.

SammyK (approved):

PR #1174, opened 2021-03-10
Let's try composer 2, again!

morrisonlevi:
### Description

Composer clients earlier than 1.10.1 will go directly to the latest
v2 on a `self-update`, but 1.x clients newer than this will only
update to the latest 1.x. The older clients will abort on --2 being
passed as they don't understand it, so we have to update twice.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


krakjoe:
@morrisonlevi what's the status here, you updated the branch quite recently ?

PR #1175, opened 2021-03-11
[ddprof] Support v3 intake format; HTTP bugfixes

sanchda:
Various v3 intake format stuff.

### Description

dd_send should now support the v3 intake format.  We don't provide a v2 compatibility mode--just outright disable it.  It would be easy enough to add support again, but I see no reason to support a legacy interface.

Also the HTTP layer now:
 * uses Connection: close and do tail-closing for HTTP reqs
 * properly handles SIGPIPE (by not allowing it to be thrown!)

### DIFF
--- a/src/dd_send.c
+++ b/src/dd_send.c
@@ -46,6 +46,8 @@ DDReq *DDR_init(DDReq *req) {
     if (!req)
       return NULL;
     req->ownership = 0x01;
+  } else {
+    req->owernship = 0;
   }
 
   req->as_header = as_init(NULL);
--- a/src/dd_send.c
+++ b/src/dd_send.c
@@ -172,14 +172,15 @@ int DDR_finalize(DDReq *req) {
     int j = DDR_idx[i];
     char *val = req->values[j];
 
-    // If we got a value, great!  If not, check for a default.  If there's
-    // no value and no default, skip this entry.
-    if (!val && DDR_defaults[j])
-      val = (char *)DDR_defaults[j];
-    else if (!val && !DDR_defaults[j] && DDR_reqd[i])
-      return DDRC_EINVAL;
-    else if (!val && !DDR_defaults[j])
-      continue;
+    // If the request specified a value, then use that.
+    if (!val) {
+      if (DDR_reqd[i])
+        return (long)DDR_defaults[j];
+      else if (DDR_defaults[j])
+        val = (char *)DDR_defaults[j];
+      else
+        continue;
+    }
     switch (DDR_types[j]) {
     case 0: // Skipit
       break;
--- a/src/dd_send.c
+++ b/src/dd_send.c
@@ -172,15 +172,14 @@ int DDR_finalize(DDReq *req) {
     int j = DDR_idx[i];
     char *val = req->values[j];
 
-    // If the request specified a value, then use that.
-    if (!val) {
-      if (DDR_reqd[i])
-        return (long)DDR_defaults[j];
-      else if (DDR_defaults[j])
-        val = (char *)DDR_defaults[j];
-      else
-        continue;
-    }
+    // If we got a value, great!  If not, check for a default.  If there's
+    // no value and no default, skip this entry.
+    if (!val && DDR_defaults[j])
+      val = (char *)DDR_defaults[j];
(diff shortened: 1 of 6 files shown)


src/dd_send.c
```diff
@@ -45,8 +46,6 @@ DDReq *DDR_init(DDReq *req) {
     if (!req)
       return NULL;
     req->ownership = 0x01;
-  } else {
-    req->ownership = 0;
```
morrisonlevi, line 49: Any reason this was removed? I'm guessing it accidentally got lost in rebase/merge or something.

src/pprof.c
```diff
@@ -520,8 +520,6 @@ DProf *pprof_Init(DProf *dp, const char **sample_names,
     if (!dp)
       return NULL;
     dp->ownership = 1;
-  } else {
-    dp->ownership = 0;
```
morrisonlevi, line 524: This one also went away; intentional?

morrisonlevi (commented):
Just an 👀 review; will check out the code and try it out later.

src/dd_send.c
sanchda: Oops, wrong brain parity.  Hopefully this is the last two-way merge we have to do, since I changed the way things are done on the other side.  Good catch!  Updated.

src/pprof.c
sanchda: It was intentional, but wrong.  Should be fixed, but there may be other instances.  I'll double-check.

src/dd_send.c
```diff
@@ -46,7 +47,7 @@ DDReq *DDR_init(DDReq *req) {
       return NULL;
     req->ownership = 0x01;
   } else {
-    req->ownership = 0;
+    req->owernship = 0;
```
morrisonlevi, line 50: Typo:
```suggestion
    req->ownership = 0;
```

src/http.c
```diff
@@ -99,10 +108,23 @@ int HttpSend(HttpReq *req, const void *payload, size_t sz_payload) {
                            (const char *)req->port)))
       return ret;
 
-  while (-1 == (ret = send(req->conn->fd, payload, sz_payload, MSG_DONTWAIT)))
+  while (-1 ==
+         (ret = send(req->conn->fd, payload, sz_payload,
+                     MSG_DONTWAIT | MSG_NOSIGNAL)))
```
morrisonlevi, line 113: `MSG_NOSIGNAL` isn't supported on Mac.

morrisonlevi (changes requested):

src/http.c
sanchda: Should be fixed in 97b2d65, but haven't had a chance to check yet.

src/dd_send.c
sanchda: Wow, sloppy.  Fixed in 97b2d65

morrisonlevi (commented):
Just tested out the latest changes and got the following:

> app_1            | Got an error (no start value given)
> app_1            | Got an error (invalid wait, no response pending)

So, that works and is helpful :)

morrisonlevi:
Using the latest changes, I'm not sure how to properly set the `.start` and `.end` values. They are strings, and `DDR_setTimeNano` appears to create them in the right format, but does a push instead of setting these values on the `.start` and `.end`:

```c
  DDR_push(req, "name=\"start\"", NULL, (const unsigned char *)time_start,
           strlen(time_start));
  DDR_push(req, "name=\"end\"", NULL, (const unsigned char *)time_end,
           strlen(time_end));
```

How do you want to fix this? I see two options:

 1. Have `DDR_setTimeNano` set the `.start` and `.end` fields instead of doing a `DDR_push`.
 2. I don't see any uses of `DDR_setTimeNano` in the codebase, so we could refactor the function to simply do the formatting.
 
I also see a bad comment, I think:
 ```c
 void DDR_setTimeNano(DDReq *req, int64_t ti, int64_t tf) {
  ti /= 1000000000; // time_t is in seconds
  tf /= 1000000000;
  // ...
}
```

Isn't this in nanos?

morrisonlevi:
Also, it's using `localtime`; shouldn't it use `gmtime` for UTC?

PR #1179, opened 2021-03-13
Refactor the PHP 8 exception serialization

morrisonlevi:
### Description

This is now based on the PHP 8.0 version of
Exception::getTraceAsString.

Also patch up PDO exception tests. In 8.0.1 PHP fixed a bug:
php/php-src@ccb7f1c7d8518dc4548273a5acd578c57d47cc0c

In our tests it manifests as:
> SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens

This isn't what we are testing, so let's fix it so when we upgrade
our container it doesn't start failing.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/php8/engine_api.h
+++ b/ext/php8/engine_api.h
@@ -47,13 +47,13 @@ inline zval ddtrace_zval_undef(void) {
     return zv;
 }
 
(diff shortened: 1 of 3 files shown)


SammyK (approved):
👍 LGTM. That's much cleaner. (Friendly reminder about milestone + changelog.)

Issue #1157, opened 2021-02-25
Undefined symbol: curl_multi_ce on php8.0


metaxy:
### Bug description
The extension ddtrace won't load because of `undefined symbol: curl_multi_ce`

```
Warning: PHP Startup: Unable to load dynamic library 'ddtrace.so' (tried: /opt/dd-trace-php-master/modules/ddtrace.so (/opt/dd-trace-php-master/modules/ddtrace.so: undefined symbol: curl_multi_ce), /opt/dd-trace-php-master/modules/ddtrace.so.so (/opt/dd-trace-php-master/modules/ddtrace.so.so: cannot open shared object file: No such file or directory)) in Unknown on line 0
```

### PHP version
```
PHP 8.0.2 (cli) (built: Feb  5 2021 04:58:33) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.2, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.2, Copyright (c), by Zend Technologies
```

### Installed extensions
```
[PHP Modules]
bcmath
bz2
calendar
Core
ctype
curl
date
dba
dom
enchant
exif
fileinfo
filter
ftp
gd
gettext
gmp
hash
iconv
imap
intl
json
ldap
libxml
mbstring
memcached
mysqli
mysqlnd
odbc
openssl
pcntl
pcre
PDO
pdo_mysql
PDO_ODBC
pdo_pgsql
pdo_sqlite
pgsql
Phar
posix
pspell
readline
redis
Reflection
session
SimpleXML
soap
sockets
sodium
SPL
sqlite3
standard
sysvmsg
sysvsem
sysvshm
tidy
tokenizer
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
```

### OS info
```
NAME="Ubuntu"
VERSION="18.04.5 LTS (Bionic Beaver)"
PRETTY_NAME="Ubuntu 18.04.5 LTS"
VERSION_ID="18.04"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
```

#### Output of phpinfo() (ddtrace >= 0.47.0)

None. Because it is not loaded.

But maybe it significant how its compiled:
![Screenshot_20210225_101547](https://user-images.githubusercontent.com/82704/109130852-7a6e3a80-7752-11eb-91ab-6673c1fcc324.jpeg)

### Upgrading info

[...]

PR #1061, opened 2020-10-23
Run php-src master build in CI

SammyK:
### Description

WIP

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the release document.
- [ ] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1181, opened 2021-03-16
Fix "undefined symbol: curl_multi_ce" on PHP 8

SammyK:
### Description

When the curl extension is not loaded on PHP 8 (either not compiled with PHP or compiled as a shared lib) the tracer will fail to load with the error `undefined symbol: curl_multi_ce`.

The original fix (#1144) added ext/curl as a required module to load ddtrace. The fix in this PR removes that requirement by adding weak symbols for `curl_ce` and `curl_multi_ce` when the curl extension is missing. It also removes the dependency on the ext/curl headers in the event curl is loaded as a shared library and the headers are not installed. In this case the symbol addresses are fetched during process startup.

This PR adds a new Buster container (`datadog/dd-trace-ci:php-8.0-shared-ext`) with PHP 8.0 that is built with `--disable-all`. This allows us to run tests against core extensions that are built as shared libraries.

Since this PR also includes a patch update to the PHP 8.0 container (`8.0.0` -> `8.0.3`), I added three language tests to the skip list. I've verified that the tests only fail because of the object ID changing (`object(Test)#%d`) due to our request_init_hook.

Thanks to @metaxy for reporting this issue in #1157.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/php8/handlers_curl.c
```diff
@@ -10,6 +10,11 @@
 #include "logging.h"
 #include "span.h"
 
+#if !(defined(HAVE_CURL) && !defined(COMPILE_DL_CURL))
+__attribute__((weak)) zend_class_entry *curl_ce = NULL;
+__attribute__((weak)) zend_class_entry *curl_multi_ce = NULL;
+#endif
```
morrisonlevi, line 16: If these weak symbols don't get linked in, what's the behavior?

SammyK: These symbols should always be linked in when curl isn't compiled with `--with-curl` or when `--with-curl=shared` but curl isn't loaded via INI. (In those cases, none of the codepaths in this file are reached.) I'll have some tests in CI to validate this behavior by EOD tomorrow.

SammyK: Here's how [ext/uv fixed a similar issue](https://github.com/amphp/ext-uv/commit/f7d052a48b8a63ae090e069d42e4278b33e69484) so I think we also need to make use of `DL_FETCH_SYMBOL` in MINIT when curl is loaded as a shared lib. I'll make sure the tests hit that scenario and update the fix.

SammyK: > These symbols should always be linked
Turns out on some platforms the symbols need to be fetched with `dlsym()`. I've added this to MINIT.

.circleci/config.yml
```diff
@@ -619,6 +619,56 @@ jobs:
       - <<: *STEP_COMPOSER_CACHE_SAVE
       - <<: *STEP_PERSIST_TO_WORKSPACE
 
+  internal_integrations:
+    parameters:
+      docker_image:
+        type: string
+      valgrind_config:
+        type: string
+        default: musl
```
labbati, line 628: This param is not set in the workflow and defaults to musl, while the image is a buster image. Is it a copy paste left over?

docker-compose.yml
```diff
@@ -118,6 +118,7 @@ services:
   '7.4-debug': { <<: *base_php_service, image: 'datadog/dd-trace-ci:php-7.4-debug-alpine-3.11' }
   '7.4-debug-asan': { <<: *base_php_service, image: 'datadog/dd-trace-ci:php-7.4-debug-asan-buster' }
   '8.0': { <<: *base_php_service, image: 'datadog/dd-trace-ci:php-8.0_buster' }
+  '8.0-shared-ext': { <<: *base_php_service, image: 'datadog/dd-trace-ci:php-8.0-shared-ext_buster' }
```
labbati, line 121: Here we 
```suggestion
  '8.0-shared-ext': { <<: *base_php_service, image: 'datadog/dd-trace-ci:php-8.0-shared-ext' }
```

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
```diff
@@ -0,0 +1,101 @@
...
+RUN set -eux; \
+    curl -fsSL -o /tmp/php.tar.gz "${phpTarGzUrl}"; \
+    (echo "${phpSha256Hash} /tmp/php.tar.gz" | sha256sum -); \
+    tar xf /tmp/php.tar.gz -C "${PHP_SRC_DIR}" --strip-components=1; \
+    rm -f /tmp/php.tar.gz; \
+    cd ${PHP_SRC_DIR}; \
+    ./buildconf --force;
+COPY configure_shared_ext.sh /home/circleci/configure.sh
+
+FROM build as php-debug-zts-asan
```
labbati, line 21: Just asking: aren't we testing asan on nts? (I have a :palmface: reaction ready 😄 )

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
```diff
@@ -0,0 +1,101 @@
...
+FROM base as final
+COPY --chown=circleci:circleci --from=build $PHP_SRC_DIR $PHP_SRC_DIR
+COPY --chown=circleci:circleci --from=php-debug-zts-asan $PHP_INSTALL_DIR_DEBUG_ZTS_ASAN $PHP_INSTALL_DIR_DEBUG_ZTS_ASAN
+COPY --chown=circleci:circleci --from=php-debug $PHP_INSTALL_DIR_DEBUG_NTS $PHP_INSTALL_DIR_DEBUG_NTS
+COPY --chown=circleci:circleci --from=php-nts $PHP_INSTALL_DIR_NTS $PHP_INSTALL_DIR_NTS
+
+# Build core extensions as shared libraries.
+# We intentionally do not run 'make install' here so that we can test the
+# scenario where headers are not installed for the shared library.
+RUN set -eux; \
```
labbati, line 68: Not required now, but in the future we could move it to an executable the installation, something like `php-install-ext-no-header pdo` so we do not have to rebuild containers at every new extension we want to test. Not required for this specific PR. If you find it useful can drop a few lines in a card, though.

dockerfiles/ci/buster/php-master/Dockerfile
```diff
@@ -10,10 +10,8 @@ RUN set -eux; \
 COPY install-php-master /usr/local/bin/
 RUN install-php-master
 
-# Install Composer
-RUN set -eux; \
-    curl -q https://raw.githubusercontent.com/composer/getcomposer.org/76a7060ccb93902cd7576b67264ad91c8a2700e2/web/installer | \
-    sudo php -- --filename=composer --install-dir=/usr/local/bin;
+
+COPY --from=composer:2 /usr/bin/composer /usr/local/bin/composer
```
labbati, line 14: I do not see the advantage of doing this (TIL we do in other places, though).

When we do this we have to download the composer image in order only to extract a file (evidence of that)
``` => FROM docker.io/library/composer:2                                                                                                                                        14.5s
 => => resolve docker.io/library/composer:2                                                                                                                                   2.3s
 => => sha256:aca054f37f7b1a5f63bc8363a31c383ac9e2a4edd9983b92421c34efa1f35b2a 1.65kB / 1.65kB                                                                                0.0s
 => => sha256:743181516422399af3888ad5df44c6f711f992b2f86288db9e191ae13de99fa0 1.69MB / 1.69MB                                                                                0.4s
 => => sha256:e2d5b43308915ce3198dec1c19aa88719afacff02c85660a952982856a0bcc9c 268B / 268B                                                                                    0.5s
 => => sha256:6c307b60e377a3903f74fd9cde3ad69f754071faac36f81ca74e1aa9eabe4851 1.26kB / 1.26kB                                                                                0.5s
 => => sha256:ff854871a5957060e93934887728d773bd21351468fc6b4b6ac5465ad2c81ea2 13.32kB / 13.32kB                                                                              0.0s
 => => sha256:3ee24d70a08e170bdb914f6bc4cea57fee943ff43c6856a3f168ee993e53e786 3.46kB / 3.46kB                                                                                0.0s
 => => extracting sha256:743181516422399af3888ad5df44c6f711f992b2f86288db9e191ae13de99fa0                                                                                     0.2s
 => => sha256:0baf29266ab77eda55610c722f3b903364328c7c92117690eb657655e419bf9d 10.78MB / 10.78MB                                                                              4.4s
 => => sha256:dfa723e4109be76aefa6e8bc28bcacb55aa90d501f03ffc5649eebd131ceed09 493B / 493B                                                                                    0.7s
 => => sha256:7e9fbc8e5c1a07d8bdbc757bd388cbfe410d3eec2cb1ec49ea770320ae242dc7 14.95MB / 14.95MB                                                                              6.1s
 => => extracting sha256:6c307b60e377a3903f74fd9cde3ad69f754071faac36f81ca74e1aa9eabe4851                                                                                     0.0s
 => => sha256:510d7725cdd56d978d0dfb33354aea034a3da8931963851655e104f0e1db0463 2.27kB / 2.27kB                                                                                1.0s
 => => extracting sha256:e2d5b43308915ce3198dec1c19aa88719afacff02c85660a952982856a0bcc9c                                                                                     0.0s
 => => sha256:5fdc925b16f55706769b248f0cffef3239f6290a6ba31d73d3c1011723849d36 17.52kB / 17.52kB                                                                              1.4s
 => => sha256:c898ee3b4745b6a914bc0c00649d2e875d9a0affe2a2348cf3b40137347ea3a7 31.16MB / 31.16MB                                                                              6.0s
 => => sha256:f712c6bde093375cf23866a1ec9cc403af45a0b97dfa6194e82a788ae309c3b7 198.23kB / 198.23kB                                                                            4.8s
 => => extracting sha256:0baf29266ab77eda55610c722f3b903364328c7c92117690eb657655e419bf9d                                                                                     0.1s
 => => extracting sha256:dfa723e4109be76aefa6e8bc28bcacb55aa90d501f03ffc5649eebd131ceed09                                                                                     0.0s
 => => sha256:02b0dc7f879137009be18156044da5d8d5eb9b720fb2ac82021a3407d407c391 258B / 258B                                                                                    5.0s
 => => sha256:37eb1a7b3bf8c3a6cbe7b93e0a1bd979270105313f6dc3bfd6856b31f32f6367 554.12kB / 554.12kB                                                                            5.7s
 => => sha256:a53a2191b206d36a2c5b9109f227baae4f7f398253a18268080a66745a783c59 409B / 409B                                                                                    6.0s
 => => sha256:8f5c6ac570cfb9ffd956004be1ecd9bee6d150945b96c84252506fc2f779383f 126B / 126B                                                                                    6.2s
 => => extracting sha256:7e9fbc8e5c1a07d8bdbc757bd388cbfe410d3eec2cb1ec49ea770320ae242dc7                                                                                     1.8s
 => => extracting sha256:510d7725cdd56d978d0dfb33354aea034a3da8931963851655e104f0e1db0463                                                                                     0.0s
 => => extracting sha256:5fdc925b16f55706769b248f0cffef3239f6290a6ba31d73d3c1011723849d36                                                                                     0.0s
 => => extracting sha256:c898ee3b4745b6a914bc0c00649d2e875d9a0affe2a2348cf3b40137347ea3a7                                                                                     4.1s
 => => extracting sha256:f712c6bde093375cf23866a1ec9cc403af45a0b97dfa6194e82a788ae309c3b7                                                                                     0.2s
 => => extracting sha256:02b0dc7f879137009be18156044da5d8d5eb9b720fb2ac82021a3407d407c391                                                                                     0.0s
 => => extracting sha256:37eb1a7b3bf8c3a6cbe7b93e0a1bd979270105313f6dc3bfd6856b31f32f6367                                                                                     0.2s
 => => extracting sha256:a53a2191b206d36a2c5b9109f227baae4f7f398253a18268080a66745a783c59                                                                                     0.0s
 => => extracting sha256:8f5c6ac570cfb9ffd956004be1ecd9bee6d150945b96c84252506fc2f779383f                                                                                     0.0s
```

I do not find nice for who build the image to have to download extra GBs

labbati (changes requested):
Just a few comments. Thanks for putting this together @SammyK

.circleci/config.yml
SammyK: Nice catch. This is copy pasta.

docker-compose.yml
SammyK: Ah. I meant to push the container up as `datadog/dd-trace-ci:php-8.0-shared-ext_buster` originally but I accidentally truncated it when playing with a variation. We'll just go with the truncated version since that's the one that is pushed up. Thanks for noticing this!

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
SammyK: No, IIRC we combined ASan and ZTS to hit all the variations in the fewest amount of PHP builds since the container size grows significantly with each PHP build. Right now we have:
- ndbug-nzts
- debug-nzts
- debug-zts-asan

The alternative would have been four builds:
- ndbug-nzts
- debug-nzts
- debug-nzts-asan
- debug-zts

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
SammyK: I'm not opposed to adding something like this, but I have a few questions:

1. Would `php-install-ext-no-header` be intended for use directly in CI? So we would add a new command to the job to build the extension every time?
2. How would the script handle extra configure flags? I'm asking because although `curl` and `pdo` have a standard build process, `pdo_mysql`, `pdo_pgsql`, and `pdo_sqlite` may require extra config options and/or `CFLAGS`. (I tried to install them here but I ran into trouble getting them to build since ext/pdo is shared.)

FWIW I added these to a separate layer at the very bottom of the Dockerfile so that if all the other layers are cached, this layer will build fairly quickly after adding a new extension. (I just timed adding ext/pcntl to this list and rebuilt the container and it took 42 seconds with cached layers.) But I also understand not wanting to have to rebuild the container if you haven't built it yet. :)

dockerfiles/ci/buster/php-master/Dockerfile
SammyK: This was part of the cherry-picked commit (5ed52aa) from #1061. Using Composer 2 as part of any new container builds (e.g. `datadog/dd-trace-ci:php-8.0-shared-ext`) came as a special request from @morrisonlevi so I'll let him chime in here. :)

morrisonlevi: I'm not sure I understand your concern, Luca. Since we are including composer in the image one way or another, we have to download it. All I did was change it from messy commands to a simple docker-idiomatic one.

So... maybe I didn't catch a subtlety or something?

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
labbati: I have not  strong opinion. But if the problem is only size I would not consider it to be a problem. If, on the other hand, we feel safe running zts-asan only, I am totally fine with it. I have not enough knowledge to make a call here.

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
labbati: > I'm asking because although curl and pdo have a standard build process, pdo_mysql, pdo_pgsql, and pdo_sqlite may require extra config options and/or CFLAGS

This alone can suggest that you made the right decision. I am resolving this comment, then. Thanks for claryfing.

labbati (approved):
LGTM, thanks @SammyK. I only looked at the C part as a learner, so is better to have a more formal code review there.

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
labbati: Also, even if we decide to do nts+asan, it has not to be done as part f this PR.

SammyK: Ah, thanks for pointing this out. I'm fairly certain that running debug-nzts-asan isn't required if we're running debug-zts-asan, but I see your point that there might be an edge case hiding here.

@morrisonlevi Do you think it is necessary to test debug-nzts-asan in addition to debug-zts-asan? If so, do you think we should rebuild the `datadog/dd-trace-ci:php-8.0_buster` container without ASan and provide a separate container `datadog/dd-trace-ci:php-8.0_buster_asan` with all three builds as ASan builds and test all three? (In a separate PR.)

morrisonlevi: ZTS and non-ZTS may have different code paths sometimes -- so I suppose the correct answer is "yes, ship all the things on ASAN!"

Related, the 2x2 matrix is:

|      | debug | ndebug |
| --- | ----- | ------ |
| zts | [ x]  | [ ] |
| nts | [x] |  [x] |

Should we just bite the bullet and add the missing check here? Theoretically those can have differences too, but less likely so maybe not worth it?

dockerfiles/ci/buster/php-master/Dockerfile
morrisonlevi: This `composer` layer gets added to more than one image, and I use it for other images I make for playing around as well, so for me at least it's well cached and well worth the re-use.

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
SammyK: In that case, to be as comprehensive as possible, I'm all for building `datadog/dd-trace-ci:php-8.0_buster_asan` and completing the matrix with four ASan builds. #AllTheThings

labbati: Since we have a card for it, I am resolving this comment for now as the PR is not blocked y this work

dockerfiles/ci/buster/php-master/Dockerfile
labbati: I do not consider this a blocker. I am resolving this comment so we can unblock the PR.

ext/php8/handlers_curl.c
```diff
@@ -1,7 +1,12 @@
 #include <php.h>
 #include <stdbool.h>
 
+#if defined(HAVE_CURL) && !defined(COMPILE_DL_CURL)
 #include <ext/curl/php_curl.h>
+#else
+__attribute__((weak)) zend_class_entry *curl_ce = NULL;
+__attribute__((weak)) zend_class_entry *curl_multi_ce = NULL;
+#endif
```
morrisonlevi, line 9: Any reason we shouldn't just always do the weak symbols thing? Since we have to use it some of the time, why not all of the time?

morrisonlevi (commented):
Looks good to me. I do have one question before I hit approve, though.

morrisonlevi (approved):

ext/php8/handlers_curl.c
SammyK: To summarize our off-thread conversation: Since this is the first time we're introducing weak symbol resolution with `dlsym()` to the project, I'm not confident that we won't inadvertently introduce a regression on the happy path (when the curl header is present) if we go all in on weak symbols.

This technique should solve the issue for this specific edge case while also providing information about how weak symbol resolution with `dlsym()` fits within our existing infrastructure. We might go all in on this method in the future after our testing ability for this edge case matures and we bring more integrations to the extension level, but for now I would feel safer from a stability perspective to keep week symbols as a fallback to the edge case.

PR #1184, opened 2021-03-17
Run Guzzle 5/6/7 project test suites in CI with and without our tracer

labbati:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


dockerfiles/frameworks/contrib/Dockerfile
```diff
@@ -47,8 +46,17 @@ RUN set -xe; \
         php${V}-bcmath; \
     done; \
     ); \
+    apt-get install -y $(for V in 5.6 7.0 7.1 7.2 7.3 7.4; do \
```
labbati, line 49: `php8.0-json` is not available

dockerfiles/frameworks/contrib/expect-phpunit-failure.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
```
labbati, line 1: I prefer to add expected exception instead of marking the test as skipped so if suddenly it starts to work we know immediately.

PR #1186, opened 2021-03-22
Zend Abstract Interface: Method call seam (PHP 5)

SammyK:
### Description

The Zend Abstract Interface (ZAI) is an API for calling into PHP from C. The seams are decoupled from the main PHP tracer and tested using the ZAI SAPI.

This PR includes the first ZAI seam on PHP 5 for safely calling methods (sandboxed):

- `zai_class_lookup()`
- `zai_call_method_without_args()`
- `zai_call_static_method_without_args()`

This seam will be used to finish the implementation of autoflushing in #1189.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -85,6 +85,7 @@ if test "$PHP_DDTRACE" != "no"; then
     "
 
     ZAI_SOURCES="\
+      zend_abstract_interface/methods/php5/methods.c \
       zend_abstract_interface/sandbox/php5/sandbox.c \
       zend_abstract_interface/zai_sapi/php5/zai_sapi.c \
       zend_abstract_interface/zai_sapi/zai_sapi_functions.c \
@@ -128,6 +129,7 @@ if test "$PHP_DDTRACE" != "no"; then
     "
 
     ZAI_SOURCES="\
+      zend_abstract_interface/methods/php5/methods.c \
       zend_abstract_interface/sandbox/php5/sandbox.c \
       zend_abstract_interface/zai_sapi/php5/zai_sapi.c \
       zend_abstract_interface/zai_sapi/zai_sapi_functions.c \
@@ -254,6 +256,8 @@ if test "$PHP_DDTRACE" != "no"; then
 
   PHP_ADD_INCLUDE([$ext_srcdir/zend_abstract_interface])
   PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface])
+  PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/methods])
+  PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/methods/php5])
   PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/sandbox])
   PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/sandbox/php5])
   PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/sandbox/php7])
--- a/zend_abstract_interface/methods/methods.h
+++ b/zend_abstract_interface/methods/methods.h
@@ -29,8 +29,7 @@ zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len TSRMLS
  *   }
  *
  * Methods cannot be called outside of a request context so this MUST be called
- * from within a request context (after RINIT and before RSHUTDOWN). A crash
- * will occur if this is called outside of a request context.
+ * from within a request context (after RINIT and before RSHUTDOWN).
  */
 bool zai_call_method_without_args_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC);
 
--- a/zend_abstract_interface/methods/methods.h
+++ b/zend_abstract_interface/methods/methods.h
@@ -11,10 +11,12 @@
  * ZAI seams using native Zend Engine data structures.
  */
 
-/* Looks up a class entry directly from the EG(class_table). The lookup is
+/* Looks up a class entry directly from the CG(class_table). The lookup is
  * case-sensitive so the class name should be lowercase. The class name should
  * not contain a root-scope '\' prefix. Does not invoke the autoloader. Returns
- * NULL if the class entry could not be found.
+ * NULL if the class entry could not be found. Because the class lookup occurs
+ * from the compiler globals, this can be called safely outside of a request
+ * context.
  */
 zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len TSRMLS_DC);
 
@@ -25,6 +27,10 @@ zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len TSRMLS
  *   if (retval) {
  *     zval_ptr_dtor(&retval);
  *   }
+ * 
+ * Methods cannot be called outside of a request context so this MUST be called
+ * from within a request context (after RINIT and before RSHUTDOWN). A crash
+ * will occur if this is called outside of a request context. 
  */
 bool zai_call_method_without_args_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC);
 
--- a/zend_abstract_interface/methods/methods.h
+++ b/zend_abstract_interface/methods/methods.h
@@ -27,10 +27,10 @@ zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len TSRMLS
  *   if (retval) {
  *     zval_ptr_dtor(&retval);
  *   }
- * 
+ *
  * Methods cannot be called outside of a request context so this MUST be called
  * from within a request context (after RINIT and before RSHUTDOWN). A crash
- * will occur if this is called outside of a request context. 
+ * will occur if this is called outside of a request context.
  */
 bool zai_call_method_without_args_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC);
 
--- a/zend_abstract_interface/methods/methods.h
+++ b/zend_abstract_interface/methods/methods.h
@@ -1,0 +1,42 @@
+#ifndef ZAI_METHODS_H
+#define ZAI_METHODS_H
+
+#include <main/php.h>
+#include <stdbool.h>
+
+/* Work in progress
+ *
+ * The long-term goal is to provide ZAI data-structure shims so that the
+ * following APIs are the same across all PHP versions. For now we will provide
+ * ZAI seams using native Zend Engine data structures.
+ */
+
+/* Looks up a class entry directly from the EG(class_table). The lookup is
+ * case-sensitive so the class name should be lowercase. The class name should
+ * not contain a root-scope '\' prefix. Does not invoke the autoloader. Returns
+ * NULL if the class entry could not be found.
+ */
+zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len TSRMLS_DC);
+
+/* Calls a method on an instance of 'object' without passing any arguments.
+ * Caller must pass in a NULL pointer-pointer for the return value 'retval'.
+ * Caller must dtor a non-NULL 'retval' after the call:
+ *
+ *   if (retval) {
+ *     zval_ptr_dtor(&retval);
+ *   }
+ */
+bool zai_call_method_without_args_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC);
+
+/* Calls a static method on a class entry 'ce' without passing any arguments.
+ * Return value handling is the same as zai_call_method_without_args().
+ */
+bool zai_call_static_method_without_args_ex(zend_class_entry *ce, const char *method, size_t method_len,
+                                            zval **retval TSRMLS_DC);
+
+/* Mask away the TSRMLS_* macros with more macros */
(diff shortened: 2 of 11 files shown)


zend_abstract_interface/functions/tests/functions.cc
```diff
@@ -111,7 +20,7 @@ static bool zai_test_execute_script(const char *file) {
 }
 
 TEST_CASE("call internal functions with the embed SAPI", "[zai]") {
-    if (!zai_test_embed_init()) REQUIRE(false);
+    if (!zai_sapi_spinup()) REQUIRE(false);
```
morrisonlevi, line 23: Maybe write it like `REQUIRE(zai_sapi_spinup());` ?

zend_abstract_interface/zai_sapi/zai_sapi.c
```diff
@@ -0,0 +1,189 @@
...
+    ret = write(STDOUT_FILENO, str, str_length);
+    if (ret <= 0) return 0;
+    return ret;
+#else
+    size_t ret;
+
+    ret = fwrite(str, 1, MIN(str_length, 16384), stdout);
+    return ret;
+#endif
+}
```
morrisonlevi: I know this is copy/paste, but trying to think through it:
1. Taking a quick look, `PHP_WRITE_STDOUT` corresponds to this configure check: https://heap.space/xref/php-src/build/php.m4?r=fa9e2b31#2305. I'm guessing that it may fail on non-Unix platforms like Windows? Dunno.
2. Where is `MIN` defined for these non-Unix platforms?

morrisonlevi: It looks like maybe `write` doesn't buffer, and `fwrite` does: https://stackoverflow.com/a/1360034/538216. So for the CLI sapi at least it may make sense (let me see it asap, don't buffer it). Or something? I don't know.

config.m4
```diff
@@ -46,6 +46,11 @@ if test "$PHP_DDTRACE" != "no"; then
     components/string_view/string_view.c \
   "
 
+  ZAI_SOURCES="\
```
labbati, line 49: What is the motivation for having a zai dedicated root folder instead of using the `components` one? Is it to keep `components` PHP-free?

dockerfiles/ci/alpine/php-8.0/Dockerfile
```diff
@@ -11,7 +11,7 @@ ARG phpTarGzUrl
 ARG phpSha256Hash
 RUN set -eux; \
     curl -fsSL -o /tmp/php.tar.gz "${phpTarGzUrl}"; \
-    (echo "${phpSha256Hash} /tmp/php.tar.gz" | sha256sum -); \
+    (echo "${phpSha256Hash} /tmp/php.tar.gz" | sha256sum -c -); \
```
labbati, line 14: Good catch! Basically we weren't checking SHA on the development images?

labbati (commented):
Love to see this coming 😄

zend_abstract_interface/functions/tests/functions.cc
SammyK: 🤯

config.m4
SammyK: Yes, components are being tested on containers that do not have PHP installed and we want to keep them completely decoupled from the Zend Engine.

Also, ZAI is being written as if it was its own open source project completely decoupled from the tracer. The ZAI SAPI will have extensibility baked in so that we can use it to test the tracer separately. We'll be able to add custom INI settings without having to use INI files (something the embed SAPI cannot do) and also provide SAPI-level functions that we can expose to userland for testing purposes only. This also potentially opens up an easier entry point for fuzzing later on.

dockerfiles/ci/alpine/php-8.0/Dockerfile
SammyK: Correct. These would pass even when the hashes did not match.

zend_abstract_interface/zai_sapi/zai_sapi.c
SammyK: Thanks for calling this out. 

1. I found the [original commit](https://github.com/php/php-src/commit/bc7a582f8480ce92fecde267d4cd3c3f628a4ba9) where that was added back in 2001. I'm not sure about Windows compat.
2. It looks like `MIN` is [part of `zend_portability.h`](https://heap.space/xref/PHP-8.0/Zend/zend_portability.h#417).
3. I think this tries to use `write` because this is supposed to be the "unbuffered write" function for the SAPI. But for our purposes, I think we can just use `fwrite` without any issues.

morrisonlevi: Sammy, is there a "buffered" write operation? If not we may look into the buffering options and just use fwrite :)

config.m4
labbati: love it ❤️  thanks

dockerfiles/ci/alpine/php-8.0/Dockerfile
labbati: Again, good catch

SammyK: Thanks! :)

zend_abstract_interface/zai_sapi/zai_sapi.c
SammyK: AFAIK, the SAPIs [only have unbuffered writes](https://heap.space/xref/PHP-8.0/main/SAPI.h#223-224) and all output buffering is handled by [PHP's output buffer](https://www.php.net/manual/en/ref.outcontrol.php) feature. Once PHP is ready to flush it sends it to the SAPI unbuffered write operation.

I updated the [ZAI SAPI unbuffered write operation](https://github.com/DataDog/dd-trace-php/blob/5f22ec49e454c3dbc28596ba962ade4c43a49438/zend_abstract_interface/zai_sapi/zai_sapi.c#L28-L48) to use `fwrite` for the small performance gain. I added this comment:

```
/* This is supposed to be the "unbuffered write" function for the SAPI so using
 * a buffered function like fwrite() here I suppose is technically incorrect.
 * But fwrite()'s buffering gives a little performance boost over write() which
 * better serves the purposes of running many tests through the ZAI SAPI.
 */
```

WDYT?

SammyK: I'm resolving this since the ZAI SAPI has been moved to #1204 and it implements an unbuffered approach to write operations.

morrisonlevi (commented):
I am unable to build the project for PHP 5.6-debug, though I don't think the debug part is affecting it. CMake is unable to locate `libphp5.so`. This makes sense to me, because that's an artifact of the embed SAPI, which I didn't build:

```
/opt/php/5.6-debug/bin/php-config --php-sapis
```
> cli fpm phpdbg cgi


 However, I thought using `zai` meant we no longer needed the embed SAPI?

SammyK:
> However, I thought using zai meant we no longer needed the embed SAPI?

The embed SAPI is only required to run the tests with the ZAI SAPI. Although we don't use the embed SAPI directly, ZAI SAPI requires access to the PHP shared object which, as you mentioned, is only built when the embed SAPI is enabled. So the embed SAPI is required to run the tests.

Ideally we should be able to build PHP with a config flag like `--build-shared` that would build the shared object without needing to enable the embed SAPI. That's something I've got on my TODO list to investigate for a 8.1 change.

zend_abstract_interface/methods/php5/methods.c
```diff
@@ -0,0 +1,113 @@
...
+     */
+    if (*retval != NULL) return false;
+
+    /* If 'object' is NULL, the caller wants to call this method statically. */
+    zval **obj = object ? &object : NULL;
+
+    /* This one gets me all the time. Trying to save myself 5 minutes for the
+     * next time it inevitably happens.
+     */
+    assert(z_is_lower(method) && "Don't forget to send those method names in all lowercase. ;)");
```
morrisonlevi, line 43: We should add the same check in `zai_class_lookup_ex`, I think? It requires an lcname.

zend_abstract_interface/methods/php5/methods.c
```diff
@@ -0,0 +1,113 @@
...
+    char *p = (char *)str;
+    while (*p) {
+        if (isalpha(*p) && !islower(*p)) return false;
+        p++;
+    }
+    return true;
+}
+#endif
+
+static bool z_call_method_without_args_ex(zval *object, zend_class_entry *ce, const char *method, size_t method_len,
```
morrisonlevi, line 28: All the comments in this file are spot-on. Thanks for documenting all the sharp edges!

zend_abstract_interface/methods/php5/methods.c
```diff
@@ -0,0 +1,113 @@
...
+        // "Cannot call abstract method %s::%s()"
+        return false;
+    }
+
+    bool ret = false;
+
+    zai_sandbox sandbox;
+    zai_sandbox_open(&sandbox);
+
+    zend_try {
```
morrisonlevi, line 85: Couldn't this be a problem if we catch bailouts in tracing closure callbacks? I assume bailouts happen because they can't recover, so does this need to "re-throw" the bailout? I'm thinking about `ext/soap` or something. I don't know, we don't intend to call any such functions, do we? Does `exit` use bailout on these versions?

zend_abstract_interface/methods/php5/methods.c
```diff
@@ -0,0 +1,113 @@
...
+#include <sandbox/sandbox.h>
+
+zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len TSRMLS_DC) {
+    if (!cname || !cname_len) return NULL;
+    zend_class_entry **ce;
+    /* Since we do not want to invoke the autoloader and we assume the caller
+     * will pass in the lowercased class name, we look up the class entry from
+     * the EG(class_table) directly versus calling zend_lookup_class_ex().
+     */
+    return (zend_hash_find(EG(class_table), cname, (cname_len + 1), (void **)&ce) == SUCCESS) ? *ce : NULL;
```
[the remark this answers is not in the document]
SammyK: Very nice catch! Yes, this would crash outside of a request context because the executor globals haven't been initialized for class lookups. I've updated this to use compiler globals instead and added a test for a class lookup outside of a request context. I also added to the docs for method calls stating that they must be called within a request context since that is a requirement of the Zend Engine. (3f8a405)

zend_abstract_interface/methods/php5/methods.c
SammyK: Indeed! Good point. I included this assertion in class lookups. I also added an assertion for root-scope prefixes. (62c9b73)

zend_abstract_interface/methods/methods.h
```diff
@@ -0,0 +1,48 @@
...
+ *
+ *   if (retval) {
+ *     zval_ptr_dtor(&retval);
+ *   }
+ *
+ * Methods cannot be called outside of a request context so this MUST be called
+ * from within a request context (after RINIT and before RSHUTDOWN). A crash
+ * will occur if this is called outside of a request context.
+ */
+bool zai_call_method_without_args_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC);
```
morrisonlevi, line 35: Can we properly detect request context and return false if we're not inside one? Then we can run tests since they won't crash. If this is too expensive then we can consider moving it to debug-only or something.

zend_abstract_interface/methods/php5/methods.c
SammyK: Hum. I think you're right that we cannot safely catch arbitrary zend_bailouts and go on pretending like nothing happened. In the `ext/soap` case, zend_bailouts are [caught and converted into exceptions](https://github.com/php/php-src/blob/master/ext/zend_test/tests/observer_error_04.phpt), but as you pointed out that's not the case for `exit`. (In PHP 8 `exit` is a clean shutdown thanks to the uncatchable Chuck Norris exception, `UnwindExit`. 😄) But we can expect an arbitrary unclean shutdown here when `exit` is used or some fatal error. If we ignored it, this would lead to a lot of ZMM leaks and likely a number of real memory leaks.

So although it is technically a "hole in the sandbox" to remove the `zend_try` here, it's actually safer to not catch them. If we wanted a true sandbox, we could run these as "sub requests" like [`ext/sandbox` does](https://github.com/krakjoe/sandbox/blob/7930851/src/sandbox.c#L353-L412). But I'm not sure the overhead of something like this would be worth it - especially since we're reducing our C-to-PHP seams.

Very nice _catch_ on this (pun intended). I'll remove this and update the tests.

morrisonlevi: Do we need to remove it, or do we need to catch it, undo the sandbox and then re-throw?

zend_abstract_interface/methods/methods.h
SammyK: Good idea. I've aded this check in e346b91. I believe this only applies to static method calls since a crash will occur when trying to instantiate an object outside of a request context. So I updated the formally crashy test to a static method call test.

morrisonlevi (approved):
Thanks for addressing my concerns. I'm glad that you documented many of the sharp edges so we know _why_ the code is written that way; this was something sorely needed.

SammyK:
Thanks for such a high quality code review @morrisonlevi. 💯

Issue #1182, opened 2021-03-17
[BUG] Compatiblity with opentracing/opentracing:1.0.1


Agares:
### Bug description
I'm trying to use this extension with opentracing/opentracing:1.0.1. It works fine with 1.0.0-beta6, but it fails on 1.0.1 due to missing typehints.
Version >=1.0.0 is required for PHP8 compatibility.

```
<br />
<b>Fatal error</b>: Declaration of DDTrace\OpenTracer\Tracer::getScopeManager() must be compatible with
OpenTracing\Tracer::getScopeManager(): OpenTracing\ScopeManager in
<b>/opt/datadog-php/dd-trace-sources/src/DDTrace/OpenTracer/Tracer.php</b> on line <b>134</b><br />
```

### PHP version
```
PHP 8.0.2 (cli) (built: Feb  2 2021 19:28:42) ( NTS gcc x86_64 )
Copyright (c) The PHP Group
Zend Engine v4.0.2, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.2, Copyright (c), by Zend Technologies
    with Xdebug v3.0.3, Copyright (c) 2002-2021, by Derick Rethans
    with ddtrace v0.54.0, Copyright Datadog, by Datadog
```

### Installed extensions
```
[PHP Modules]
apcu
bz2
calendar
Core
ctype
curl
date
ddtrace
dom
exif
fileinfo
filter
ftp
gettext
hash
iconv
igbinary
intl
json
json_post
libxml
mbstring
msgpack
openssl
pcntl
pcre
PDO
pdo_sqlite
pdo_sqlsrv
Phar
readline
redis
Reflection
session
SimpleXML
sockets
SPL
sqlite3
sqlsrv
standard
tokenizer
xdebug
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Xdebug
Zend OPcache
ddtrace
```

### OS info
```
NAME="CentOS Linux"
VERSION="7 (Core)"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
CPE_NAME="cpe:/o:centos:centos:7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```
#### Output of phpinfo() (ddtrace >= 0.47.0)

[...]

PR #1188, opened 2021-03-24
Fix: [BUG] Compatiblity with opentracing/opentracing:1.0.1

bbaga:
### Description

- Adding type hints to maintain compatibility with OpenTracing v1.0, problem described in #1182 
- Return correct instance from `\DDTrace\OpenTracer::withBaggageItem()`
- Checked `Tests added for this feature/bug.` since tests already exists.
- Tested with `make test_opentracing_10` command

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bbaga:
Not sure why that one build failed, ran the same command in the PHP 8.0 container, and the results looked good.
Output of the local run: https://gist.github.com/bbaga/11643cfa3c192e43fb0b3ba5edaa8ab4

composer.json
```diff
@@ -78,7 +78,7 @@
         "scenarios": {
             "opentracing1": {
                 "require": {
-                    "opentracing/opentracing": "1.0.0-beta6"
+                    "opentracing/opentracing": "^1.0"
```
labbati, line 81: The `"extra"` section in the root `composer.json` can be removed altogether as it is no longer used, while the one in `tests/composer.json` is used.

labbati (changes requested):
Hi @bbaga and thanks for the PR you submitted. I need to involve @SammyK and @morrisonlevi as well in this conversation as we have to make a call on how to support all the versions of PHP we have to support with the same code base.

_Problem_
OpenTracing moved from `PHP>=5.6` to `PHP>=7.1` between [1.0.0-beta5](https://github.com/opentracing/opentracing-php/blob/1.0.0-beta5/composer.json#L14) and [1.0.0-beta6](https://github.com/opentracing/opentracing-php/blob/1.0.0-beta6/composer.json#L14) - which can be seen as a little unusual - and from no type hints to type hints on the various interfaces between [1.0.0-beta6](https://github.com/opentracing/opentracing-php/blob/1.0.0-beta6/src/OpenTracing/Span.php#L44) and [1.0.0](https://github.com/opentracing/opentracing-php/blob/1.0.0/src/OpenTracing/Span.php#L49).

Since we have to support PHP 5.6+, this PR cannot be accepted as it is.

The long term goal would be, probably to have a separate shim dedicated to OT, in the meantime I think we should find a workaround here.

One possibility that I see, is that we have two exact clones of the same folder: `src/DDTrace/OpenTracer` (identical to how it used to be before this PR) and `src/DDTrace/OpenTracer1` (with @bbaga changes), both registered in the optional dependencies class loader.

The first one would have no type-hints (5.6+), the second one would have type hints (PHP 7.1+ with OT 1.0.0+).

I think that this is the most balanced workaround to this problem, without adding the complexity of a new external package, considering that OT's future might be inside OpenTelemetry rather than as an independent project and we want to think twice before adding a new package because of OT.

labbati:
@bbaga a follow-up to my previous comment. We cannot merge the PR as it is. We would introduce a breaking change for users on PHP 5.6, who have not a possibility to update.

At the same we want to support your use case. Here is my proposal:
- there will be 2 folders: `src/DDTrace/OpenTracer` having namespace `DDTrace\OpenTracer` (exactly as it is on `master` today) and `src/DDTrace/OpenTracer1` having namespace `DDTrace\OpenTracer1` (which is exactly your code, with type hints etc..except on a new namespace);
- Devs on 5.6 will use the old namespace, other devs will use the `DDTrace\OpenTracer1` namespace (compatible with 1.0.0).
- We do not change `_files.php`, so classes will be loaded by the psr4 autoloader that provides the optional dependencies (hence you have to add the files in `src/DDTrace/OpenTracer1` to https://github.com/DataDog/dd-trace-php/blob/a8dc8fd0bb5646b3fb2fed6b8210bf31832aaca7/bridge/dd_register_optional_deps_autoloader.php#L18-L22

If you are open to make the changes above, then we will merge the PR. We will take care do add some testing around this ahead of the next release, so you do not waste any more time.

In the long term, we will elaborate a more elegant strategy, possible a separate composer package having two separate release branches.

How does that look to you?

bridge/_files.php
```diff
@@ -8,7 +8,12 @@
     __DIR__ . '/../src/api/Contracts/Scope.php',
     __DIR__ . '/../src/api/Contracts/ScopeManager.php',
     __DIR__ . '/../src/api/Contracts/SpanContext.php',
+    __DIR__ . '/../src/api/NoopSpan.php',
```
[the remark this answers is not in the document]
bbaga: I tried to follow the instruction in the [CONTRIBUTING.md](https://github.com/DataDog/dd-trace-php/blob/9417b9289a5067a4783835d3611a91ef982f6753/CONTRIBUTING.md) to run the tests in the containers, but none of the tests worked due to autoloading issues, adding these files and changing the bootstrap a bit solved the issues.
I'll remove them id they arent supposed to be there.

bbaga:
Thank you @labbati for the feedback. I expected that the PHP 5.6+ support could be an issue, but the `1.0.0-beta6` refenrece in the `composer.json` files threw me off since that doesn't support 5.6+ anymore.

Your proposal makes sense to me, I can make those changes tomorrow. However, I think we might be able to do this by declaring the classes inside same namespace behind conditions on the PHP version so users don't have to worry about which namespace they use. Although that might be somewhat noisy in the class files. What do you think about that?

Can you give me some pointers on how to run the test locally without having to hack the autoloader? I haven't had luck following the steps in the [CONTRIBUTING.md](https://github.com/DataDog/dd-trace-php/blob/9417b9289a5067a4783835d3611a91ef982f6753/CONTRIBUTING.md).

bbaga:
I think there is a bit of an issue with the current code on master:
https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/5447/workflows/40b98010-9a5b-4373-aa9a-8b334aee3d41/jobs/364585

If I'm reading things right, the code depends on a feature introduced in opentracing 1.0.0-beta6:
https://github.com/opentracing/opentracing-php/compare/1.0.0-beta5...1.0.0-beta6#diff-c8e0e5fb6329cccb6f4975b4e53f6667aae84c5e8de704ab44a273199f47e664R168-R175

Wouldn't this make this library incompatible with PHP 5.6 already? 😕

bbaga:
@SammyK Thanks for the env var tip! It seems to work now, should I add it to the contribution guide while I'm at it?
I think I'm getting close to wrapping things up, even though CircleCI has some issues with GitHub webhooks atm.

SammyK:
> It seems to work now, should I add it to the contribution guide while I'm at it?

Great! And yes, feel free to add that to the contribution guide. Thanks for the suggestion. 👍

bridge/_files.php
```diff
@@ -39,6 +45,7 @@
     __DIR__ . '/../src/DDTrace/Bootstrap.php',
     __DIR__ . '/../src/DDTrace/Encoders/SpanEncoder.php',
     __DIR__ . '/../src/DDTrace/Encoders/MessagePack.php',
+    __DIR__ . '/../src/DDTrace/Encoders/Json.php',
```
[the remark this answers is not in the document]
bbaga: reverted back to the original state

bbaga:
All :white_check_mark:

composer.json
labbati: Good catch, did not think about that. Thanks

src/DDTrace/OpenTracer/Tracer.php
```diff
@@ -58,7 +58,10 @@ private function deconstructStartSpanOptions(\OpenTracing\StartSpanOptions $obj)
         }
 
         $options['finish_span_on_close'] = $obj->shouldFinishSpanOnClose();
-        $options['ignore_active_span'] = $obj->shouldIgnoreActiveSpan();
+        if (\method_exists($obj, 'shouldIgnoreActiveSpan')) {
```
labbati, line 61: @SammyK while reviewing this PR please  pay attention to this. I added this so we are compatible with all versions also in tests. Note that were compatible with 5.6 before, unless the user used an instance of `\OpenTracing\StartSpanOptions` as `$options` that I'd say is the common approach if they just use open tracing to add so spans.

labbati (commented):
@bbaga I had to make some changes locally to verify what I was about to suggest, so I hope you don't mind if I pushed them here as from my point of view this PR is good to be merged now.

A few highlights:
- we now run tests agains beta5, beta6 an 1.0 everywhere 5.6+ (when appropriate).
- added an `if` to only call the offending method on beta5. In any case, on the simple and common use case for OT, we were working on 5.6 before, so support and  testing is important.
- I removed the OT dedicated testsuite and added it to integrations, since they are pretty quick and I wanted to avoid the overhead os spinning up a new environment.

@SammyK since I pushed a commit, please do a CR. I am OK with all the other changes (not from myself).

@bbaga thanks a for fixing long running issue. That was AWESOME.

src/DDTrace/OpenTracer/Tracer.php
SammyK: Thanks for the investigation, fix, and comment @labbati. 👍

SammyK (approved):
👍 LGTM. Thanks again @bbaga for this fantastic contribution. 💯

pardeep26k:
@SammyK any idea when we release a new version?

labbati:
We target early next week. Our goal is to deploy to our pre-release environment today or Monday, and after ~48 hours under heavy load with no issues we will release 0.57.0.

pardeep26k:
Thanks for update.

labbati:
Heads up, we are observing an increase on memory consumption under heavy load relative to the previous that we are currently investigating. We do not expect to be able to release today. I am sorry for the delay.

PR #1189, opened 2021-03-26
Add auto-flushing to PHP 5

SammyK:
### Description

WIP

This is a followup PR to #819 to add auto-flushing to PHP 5.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php5/auto_flush.c
+++ b/ext/php5/auto_flush.c
@@ -1,3 +1,39 @@
 #include "auto_flush.h"
 
-ZEND_RESULT_CODE ddtrace_flush_tracer(void) { return SUCCESS; }
+#include "engine_hooks.h"  // for ddtrace_backup_error_handling
+
+ZEND_RESULT_CODE ddtrace_flush_tracer(TSRMLS_D) {
+    bool success = true;
+
+    zval *exception = NULL, *prev_exception = NULL;
+    if (EG(exception)) {
+        exception = EG(exception);
+        EG(exception) = NULL;
+        prev_exception = EG(prev_exception);
+        EG(prev_exception) = NULL;
+        zend_clear_exception();
+    }
+
+    ddtrace_error_handling eh;
+    ddtrace_backup_error_handling(&eh, EH_SUPPRESS TSRMLS_CC);
+
+    zend_bool orig_disable = DDTRACE_G(disable_in_current_request);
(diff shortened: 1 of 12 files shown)


morrisonlevi:
I think you possibly rebased this on master, instead of the newly rebased `sammyk/zend-abstract-interface` branch, which is why it's showing 5 commits and conflicts, I think?

SammyK:
I haven't rebased this one yet. I will rebase once #1186 is merged (due to squashed commits).

ext/php5/auto_flush.c
```diff
@@ -1,3 +1,44 @@
...
+
+        // $tracer->flush();
+        // $tracer->reset();
+        result = dd_call_method_ignore_retval(tracer, ZEND_STRL("flush") TSRMLS_CC) &&
+                 dd_call_method_ignore_retval(tracer, ZEND_STRL("reset") TSRMLS_CC);
+
+        DDTRACE_G(disable_in_current_request) = orig_disable;
+    }
+
+    if (tracer) {
```
morrisonlevi, line 38: If `!result` then we shouldn't even look at `tracer`, I think?

morrisonlevi (commented):
Just one question on the finer semantics of `zai_call_static_method_without_args`.

morrisonlevi (approved):
PR looks good to me. I just have one directional question:

I think we may want/need to abstract zvals at some point so when we are writing integrations we can program against the ZAI and it works on all PHP versions. We don't want to write each integration for each PHP version; we want to write it once, like we would in .php code.

Do you think it's important to handle this sooner or later?

SammyK:
> I think we may want/need to abstract zvals at some point so when we are writing integrations we can program against the ZAI and it works on all PHP versions.

Indeed. This is certainly on [the wishlist](https://github.com/DataDog/dd-trace-php/blob/f4b5065/zend_abstract_interface/methods/methods.h#L7-L12).

> Do you think it's important to handle this sooner or later?

I think that depends on the integration API design. I think there might be a way we can write the integration API such that it can still interact with native Zend Engine data structures in an abstract, version-agnostic way. If that's the case, we can hold off on creating the ZAI data structure shims for a bit longer.

At any rate, we'll certainly have to cross that bridge when we design the ZAI integration API.

Thanks for the code review and the thoughts @morrisonlevi!

PR #1190, opened 2021-03-29
Container tagging for AWS Fargate 1.4+

SammyK:
### Description

This PR adds Fargate 1.4+ support to container tagging. The ID parsing patterns have been updated for consistency with other tracers (namely DataDog/dd-trace-java#2432 & DataDog/dd-trace-py#2132) and the ID parsing is now done with the POSIX regex API.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


components/container_id/container_id.c
```diff
@@ -1,101 +1,113 @@
...
+    regex_t task_regex = {0};
+    regex_t container_regex = {0};
+    int l_res = regcomp(&line_regex, LINE_REGEX, REG_EXTENDED | REG_NOSUB);
+    int t_res = regcomp(&task_regex, TASK_REGEX, REG_EXTENDED);
+    int c_res = regcomp(&container_regex, CONTAINER_REGEX, REG_EXTENDED);
+    if (l_res != 0 || t_res != 0 || c_res != 0) {
+        regfree(&container_regex);
+        regfree(&task_regex);
+        regfree(&line_regex);
+        fclose(fp);
```
morrisonlevi, line 80: I'm not going to block this PR, but whenever there is more than 1 thing to cleanup on an early return I try to re-examine the structure to try to make it cleaner (if possible without goto). Can you put some time on my calendar to go over this?

components/container_id/container_id.c
```diff
@@ -1,101 +1,113 @@
...
+         * we must scan the file until the very end even after finding a valid
+         * container ID.
+         */
+        size_t task_len = dd_find_id_regex(&task_regex, task_buf, line);
+        if (task_buf[0] != '\0') {
+            memcpy(buf, task_buf, task_len + 1);
+            break;
+        }
+        if (buf[0] == '\0') {
+            dd_find_id_regex(&container_regex, buf, line);
```
morrisonlevi, line 104: I would expect a `memcpy` here, like the case above.

components/container_id/container_id.c
```diff
@@ -1,101 +1,113 @@
...
+    len = (size_t)pmatch[0].rm_eo - pmatch[0].rm_so;
+    if (len < MIN_ID_LEN || len > MAX_ID_LEN) return 0;
+
+    memcpy(buf, line + pmatch[0].rm_so, len);
+    buf[len] = '\0';
+
+    return len;
 }
 
 void datadog_php_container_id(char *buf, const char *file) {
```
morrisonlevi, line 48: If this "fails", how does the caller know?

morrisonlevi (dismissed):
I have a few questions. One of them might be a blocker depending on its answer but I've approved this in case it's not so we can work async; please merge responsibly :)

components/container_id/container_id.c
SammyK: The funny thing is, this is the cleaned up version; you should have seen it before. 😆 But absolutely, if you have some ideas on how to make this even cleaner, I'd love to hear your thoughts.

components/container_id/container_id.c
SammyK: There isn't an error here, but there might be a cleaner way to write this. The reason there is no `memcpy` here is because the happy path of finding a regular container ID copies directly into the main buffer of the caller (`buf`) but the edge case of a task ID uses an intermediary buffer (`task_buf`) since we'll likely have a valid container ID before finding a task ID.

But I think we can avoid the intermediary buffer by returning a "found" status from `dd_find_id_regex`. I'll rework this a bit and see if you think it's a bit cleaner.

components/container_id/container_id.c
SammyK: Given a valid `buf` is provided by the caller, it will always be filled with a container ID or an empty string. But the caller won't know if the empty string means there was an error leading up to parsing the cgroup file or if there was no valid container ID found in the file.

Do you think returning an error code would be better so that the caller could choose to emit a debug message if there was an error parsing the cgroup file?

components/container_id/container_id.c
```diff
@@ -1,101 +1,266 @@
...
+    /* Traverse the string to find a task ID with the following pattern:
+     *
+     * [0-9a-f]{32}-[0-9]{1,20}
+     *
+     */
+    while ((size_t)(start - line + TASK_ID_MIN_LEN) <= len) {
+        end = start;
+
+        /* We start off looking for 32 hex chars in a row: [0-9a-f]{32} */
+        while (isxdigit(end[0]) && LEN_SO_FAR < 32) {
```
morrisonlevi, line 88: It might be sixes, but this sort of style ends up being cleaner in some cases:

```c
char *current = start;
char *end = start + 32;
for (; current != end; ++current) {
  if (!isxdigit(*current)) break;
}
if (current != end) {
  continue;
}
```

I don't like the `LEN_SO_FAR` macro that uses local variables "magically", which is why I suggest trying it the other style.

morrisonlevi (changes requested):
I'll go over it today, but this `LEN_SO_FAR` macro that uses local variables immediately jumped out at me. Suggestion below.

morrisonlevi (commented):
The rest of this looks good! I'm feeling good after my nap today so I'm going to take a stab at the suggestion I made earlier to get rid of `LEN_SO_FAR`.

Also, I just want to once again say how nice it is to have self-contained components. This may not be a perfect scheme, but it's so much better than what we were doing before!

labbati:
@SammyK the same tests fails on `master` as well so please do not consider this PR blocked once @morrisonlevi approves the latest changes. Thanks for working on this.

components/container_id/container_id.c
morrisonlevi: Resolved by #1194.

morrisonlevi (approved):

PR #1192, opened 2021-04-02
Exclude dockerfiles directory from the composer package

djcrock:
### Description

When installing `datadog/dd-trace` via composer, the large `dockerfiles/` directory is pulled down as well.
This directory is over 40 MB uncompressed, and makes up the majority of the package contents.
Similar to #61, this change excludes that directory from the resulting archive to slim down installs.

If some of the contents of `dockerfiles/` are needed in the release, most of the benefit of this PR could be achieved by excluding just `dockerfiles/testing-environment/`.

Comparison of the resulting archives:
image

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,5 @@
 tests               export-ignore
+dockerfiles         export-ignore
 .editorconfig       export-ignore
 .gitattributes      export-ignore
 .gitignore          export-ignore


labbati (approved):
Good idea, thank you @djcrock!

PR #1193, opened 2021-04-02
Accept multiple requests dumped from request-replayer

labbati:
### Description

While we are still able to handle only one request in the response from the request replayer, the response is not in the format `[{req_1}, {req_2}]` instead of `{req}`, so that we can assert that only one request was done.

This was required in the context of `pcntl` testing.

Note: this expects https://github.com/DataDog/dd-trace-ci/pull/33 to be merged, as well. Image is already live.

### Readiness checklist
- [x] Tests added for this feature/bug. (implicit, all integrations tests would fail)

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/tests/Common/AgentReplayerTrait.php
+++ b/tests/Common/AgentReplayerTrait.php
@@ -24,6 +24,20 @@ trait AgentReplayerTrait
      */
     public function getLastAgentRequest()
     {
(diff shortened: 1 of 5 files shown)


SammyK (approved):
LGTM 👍

PR #1194, opened 2021-04-07
Use index based approach for finding container ids

morrisonlevi:
### Description

This is shorter code. It's based on the observation that there were
many uses related to length, which is easier tracked in a variable.

It also re-orders length checks to happen first, which is generally
safer (though in this case maybe it was fine).

This targets an open PR and not master, so skip the checklist.

### DIFF
--- a/components/container_id/container_id.c
+++ b/components/container_id/container_id.c
@@ -47,8 +47,6 @@ static bool dd_parser_is_valid_line(dd_parser *parser, const char *line) {
     return regexec(&parser->line_regex, line, 0, NULL, 0) == 0;
 }
 
-#define LEN_SO_FAR (end - start)
-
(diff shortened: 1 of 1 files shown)


components/container_id/container_id.c
```diff
@@ -47,8 +47,6 @@ static bool dd_parser_is_valid_line(dd_parser *parser, const char *line) {
     return regexec(&parser->line_regex, line, 0, NULL, 0) == 0;
 }
 
-#define LEN_SO_FAR (end - start)
```
SammyK, line 50: ❤️ 🙂 👍

SammyK (approved):
LGTM! Thanks for giving this the extra TLC it needed. 🙏

PR #1203, opened 2021-04-14
Remove PHP version prefix from buster development containers

labbati:
### Description

Since we dismissed the idea of having multiple versions of PHP in our development containers due  to images size, we remove all the version prefixes from the installation dirs.

This will reduce code duplication in CI, since now you can have one `swith_php_version: <variant>` that serves well across multiple versions in a matrix configuration.

Moreover, we re-enable `memcached` extension that was installed but not enabled on 7.4-.

Note to the reviewer: tests are failing because I did not push the new images. We have probably to wait until 0.57.0 before we can push the images and consequently merge this PR, but I'd love to receive an early feedback since to rebuild images takes ages.

### Readiness checklist
- ~[] Tests added for this feature/bug.~

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1709,13 +1709,13 @@ workflows:
           name: "PHP 80 asan tests"
           docker_image: "datadog/dd-trace-ci:php-8.0_buster"
           unit_test_target: test_c_asan
-          switch_php_version: 8.0-debug-zts-asan
+          switch_php_version: debug-zts-asan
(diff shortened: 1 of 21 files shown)


SammyK (approved):
👍 LGTM. Thanks for getting `lcov` squeezed in too @labbati! 💯

labbati:
Thank you @SammyK

PR #1204, opened 2021-04-14
ZAI SAPI

SammyK:
### Description

Zend Abstract Interface (ZAI) SAPI is a new test runner for testing software components that are tightly coupled to the Zend Engine. It is a fork of the [embed SAPI](https://github.com/php/php-src/tree/master/sapi/embed).

This first edition of ZAPI SAPI comes with:

- PHP 5.4 through PHP 8.0 support.
- C-level unit tests in CI with ASan + UBSan.
- Independent control over Zend Engine startup & shutdown cycles (SINIT, MINIT, and RINIT).
- Overwritable INI settings before MINIT.

We will use ZAI SAPI to test the Zend Abstract Interface seams (WIP #1186) and parts of the tracer we have been unable to test with traditional PhpUnit and phpt tests.

Some future wishlist features include:

- [ ] Custom SAPI-level functions (e.g. `DDTrace\Testing\trigger_error()`)
- [ ] Overridable unbuffered output hook
- [ ] Fuzzer integration
- [ ] Imposter: pose as another SAPI (can be handy when testing parts of OPcache and other extensions that have code paths that are [gated by the SAPI name](https://github.com/php/php-src/blob/e0e19fd955b1fdf9f4b202343b1c9e61f1f18b35/ext/opcache/ZendAccelerator.c#L2823-L2853))

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -266,7 +266,7 @@ jobs:
           command: mkdir -p test-results/phpcs
       - run:
           name: Running eclint
-          command: node_modules/.bin/eclint check '**/*' '!dockerfiles/**/*' '!tests/ext/sandbox/**' '!tests/ext/sandbox-prehook/*' '!tests/ext/background-sender/**/*' '!config.*' '!m4/*' '!tmp/**/*' '!vendor/**/*' '!ext/vendor/**/*' '!ext/.libs/*' '!src/dogstatsd/**' '!LICENSE' '!phpstan.*.neon' '!tests/overhead/**' '!tests/Frameworks/*/Version_*/**' '!tests/dockerfiles/**' '!tests/AutoInstrumentation/**' '!.composer/**/*' '!LICENSE.*' '!tooling/*' '!tests/randomized/**' '!components/*/tests/stubs/**' || touch .failure
+          command: node_modules/.bin/eclint check '**/*' '!dockerfiles/**/*' '!tests/ext/sandbox/**' '!tests/ext/sandbox-prehook/*' '!tests/ext/background-sender/**/*' '!config.*' '!m4/*' '!tmp/**/*' '!vendor/**/*' '!ext/vendor/**/*' '!ext/.libs/*' '!src/dogstatsd/**' '!LICENSE' '!phpstan.*.neon' '!tests/overhead/**' '!tests/Frameworks/*/Version_*/**' '!tests/dockerfiles/**' '!tests/AutoInstrumentation/**' '!.composer/**/*' '!LICENSE.*' '!tooling/*' '!tests/randomized/**' '!components/*/tests/stubs/**' '!**/CMakeLists.txt' || touch .failure
       - run:
           name: Running phpcs
           command: composer lint -- --report=junit | tee test-results/phpcs/results.xml || touch .failure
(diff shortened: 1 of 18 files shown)


.circleci/config.yml
```diff
@@ -1181,6 +1267,32 @@ workflows:
               catch2_version:
                 - "2.13.4"
 
+  zend_abstract_interface:
```
labbati, line 1270: I suggest to rename it `zai` as a long name would hide all the rest in the CircleCI UI, and you have to stay over to have a full description.

zend_abstract_interface/zai_sapi/php5/zai_sapi.c
```diff
@@ -0,0 +1,220 @@
...
+#include "../zai_sapi_ini.h"
+#include "../zai_sapi_io.h"
+
+#define DEFAULT_INI          \
+    "html_errors=0\n"        \
+    "implicit_flush=1\n"     \
+    "output_buffering=0\n"   \
+    "max_execution_time=0\n" \
+    "max_input_time=-1\n"    \
+    "memory_limit=-1\n"      \
```
labbati, line 16: Are these very lenient limits inherited from the embed sapi? I do not have a strong opinions, but at the same time making test execution unbound might not be a good practice. Do you think it make sense to start from the per-version default value and then adapt/disable this param when appropriate?

zend_abstract_interface/zai_sapi/zai_sapi.h
```diff
@@ -0,0 +1,73 @@
+/*                   ____   __   __    ____   __   ____  __
+ *                  (__  ) / _\ (  )  / ___) / _\ (  _ \(  )
+ *                   / _/ /    \ )(   \___ \/    \ ) __/ )(
+ *                  (____)\_/\_/(__)  (____/\_/\_/(__)  (__)
+ *
+ * ZAI SAPI is designed to test software components that are tightly coupled to
+ * the Zend Engine.
+ *
+ * ZAI SAPI is a fork the the embed SAPI:
```
labbati, line 9:
```suggestion
 * ZAI SAPI is a fork of the embed SAPI:
```

zend_abstract_interface/zai_sapi/zai_sapi_ini.h
```diff
@@ -0,0 +1,28 @@
+#ifndef ZAI_SAPI_INI_H
+#define ZAI_SAPI_INI_H
+
+#include <stdbool.h>
+#include <stddef.h>
+
+/* Allocates memory for 'dest' and copies the entires from 'src' into it.
```
labbati, line 7:
```suggestion
/* Allocates memory for 'dest' and copies the entries from 'src' into it.
```

zend_abstract_interface/zai_sapi/zai_sapi_ini.c
```diff
@@ -0,0 +1,43 @@
+#include "zai_sapi_ini.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+size_t zai_sapi_ini_entries_alloc(const char *src, char **dest) {
```
labbati, line 7: Is it ok to return `0` to basically indicate an error, should we return `-1`?
Wouldn't `0` also be the result when invoked with `zai_sapi_ini_entries_alloc("", ...);` (I guess this is a valid invocation)?

zend_abstract_interface/zai_sapi/zai_sapi_ini.c
```diff
@@ -0,0 +1,43 @@
...
+}
+
+void zai_sapi_ini_entries_free(char **entries) {
+    if (entries != NULL && *entries != NULL) {
+        free(*entries);
+        *entries = NULL;
+    }
+}
+
+size_t zai_sapi_ini_entries_realloc_append(char **entries, size_t entries_len, const char *key, const char *value) {
```
labbati, line 30: https://github.com/DataDog/dd-trace-php/pull/1204/files#diff-1c20928ddec31938013846976fec494c8a394b606b2071ce82a80144fcdec40cR7 applies here as well

zend_abstract_interface/zai_sapi/zai_sapi_io.h
```diff
@@ -0,0 +1,46 @@
...
+ * write operation for stderr at the SAPI level. But we still perform
+ * unbuffered write operations to stderr for consistency with stdout.
+ */
+size_t zai_sapi_io_write_stderr(const char *str, size_t len);
+
+/* Flushes stdout (NOOP).
+ *
+ * ZAI SAPI only supports unbuffered write operations.
+ */
+static inline void zai_sapi_io_flush(void *server_context) { (void)server_context; }
```
labbati, line 26: Why are we adding this `static` function here in the header? Shouldn't we either remove `static` or have it declared only in the file we use it?

labbati (commented):
Thanks @SammyK, added a few questions while I leave @morrisonlevi do a proper CR.

zend_abstract_interface/zai_sapi/zai_sapi_ini.h
```diff
@@ -0,0 +1,28 @@
...
+ * 'dest' must point to NULL or no allocation will occur. Caller must free with
+ * zai_sapi_ini_entries_free().
+ */
+size_t zai_sapi_ini_entries_alloc(const char *src, char **dest);
+
+/* Frees the INI entires and sets the pointer to NULL to prevent a use after
```
labbati, line 13:
```suggestion
/* Frees the INI entries and sets the pointer to NULL to prevent a use after
```

zend_abstract_interface/zai_sapi/zai_sapi_ini.h
```diff
@@ -0,0 +1,28 @@
...
+ * free.
+ */
+void zai_sapi_ini_entries_free(char **entries);
+
+/* Reallocates the INI entries and append a new entry. Returns the new length
+ * of the entires (not counting the null terminator) after appending the new
```
labbati, line 19:
```suggestion
 * of the entries (not counting the null terminator) after appending the new
```

zend_abstract_interface/zai_sapi/zai_sapi_ini.c
SammyK: Thanks for the suggestion. I agree with you that `0` isn't always an error (as with the empty string example you provided) and it's confusing for it to represent both error and success. I'll change the return type to `ssize_t` and have `-1` represent an error. 👍

zend_abstract_interface/zai_sapi/zai_sapi_ini.c
SammyK: Agreed - we need to return `-1` for errors here for consistency with `zai_sapi_ini_entries_alloc`.  👍  Although having an empty `key` or `value` wouldn't be valid in this case so those would also return `-1`. In this case, this function should never return `0`. I have tests for `NULL` key/values but not empty strings; I need to add those tests. 🙂

zend_abstract_interface/zai_sapi/zai_sapi_ini.h
SammyK: Wow. I really don't know how to spell "entries". 😆 Thanks for detecting all these!

zend_abstract_interface/zai_sapi/php5/zai_sapi.c
SammyK: > Are these very lenient limits inherited from the embed sapi?

Yes. These are the original embed SAPI values.

> making test execution unbound might not be a good practice. Do you think it make sense to start from the per-version default value and then adapt/disable this param when appropriate?

Yes, good catch. It would be better for unexpected behavior to be detected from the tests using the default INI values. One exception is setting `output_buffering=0`. Due to the console-ish nature of the ZAI SAPI, the PHP output buffer might get in the way of some of the tests (i.e. a test examines the output before PHP flushes the buffer). That being said, we can always explicitly enable OB for a test before MINIT with:

```c
zai_sapi_append_system_ini_entry("output_buffering", "1");
```

.circleci/config.yml
SammyK: Is there a particular spot in the UI where a long workflow name truncates the job name? I can't seem to find it. I named the job `Zai` for the same reason you're mentioning and so far it's looking pretty good in the UI.

Screen Shot 2021-04-15 at 9 38 22 AM

Screen Shot 2021-04-15 at 9 39 24 AM

But are you referring to a different spot?

zend_abstract_interface/zai_sapi/zai_sapi_ini.c
SammyK: Addressed in aa2216a.

zend_abstract_interface/zai_sapi/php5/zai_sapi.c
SammyK: Addressed in 0338206.

.circleci/config.yml
```diff
@@ -266,7 +266,7 @@ jobs:
           command: mkdir -p test-results/phpcs
       - run:
           name: Running eclint
-          command: node_modules/.bin/eclint check '**/*' '!dockerfiles/**/*' '!tests/ext/sandbox/**' '!tests/ext/sandbox-prehook/*' '!tests/ext/background-sender/**/*' '!config.*' '!m4/*' '!tmp/**/*' '!vendor/**/*' '!ext/vendor/**/*' '!ext/.libs/*' '!src/dogstatsd/**' '!LICENSE' '!phpstan.*.neon' '!tests/overhead/**' '!tests/Frameworks/*/Version_*/**' '!tests/dockerfiles/**' '!tests/AutoInstrumentation/**' '!.composer/**/*' '!LICENSE.*' '!tooling/*' '!tests/randomized/**' '!components/*/tests/stubs/**' || touch .failure
+          command: node_modules/.bin/eclint check '**/*' '!dockerfiles/**/*' '!tests/ext/sandbox/**' '!tests/ext/sandbox-prehook/*' '!tests/ext/background-sender/**/*' '!config.*' '!m4/*' '!tmp/**/*' '!vendor/**/*' '!ext/vendor/**/*' '!ext/.libs/*' '!src/dogstatsd/**' '!LICENSE' '!phpstan.*.neon' '!tests/overhead/**' '!tests/Frameworks/*/Version_*/**' '!tests/dockerfiles/**' '!tests/AutoInstrumentation/**' '!.composer/**/*' '!LICENSE.*' '!tooling/*' '!tests/randomized/**' '!components/*/tests/stubs/**' '!**/CMakeLists.txt' || touch .failure
```
morrisonlevi, line 269: ... maybe we should do an include list instead of an exclude list? It's not required for this PR, of course.

zend_abstract_interface/zai_sapi/zai_sapi_io.h
morrisonlevi: If you really want to get rid of `static`, that's fine, it just means in some object (and just one) we add:

```c
extern inline void zai_sapi_io_flush(void *server_context);
```

morrisonlevi (approved):
I don't know the SAPI parts of PHP well, but to the best of my knowledge this looks good!

I did expect there to be some test that actually links against the ZAI SAPI and starts up and does like a hello-world from inside a request or something -- is that coming in another PR?

SammyK:
> I did expect there to be some test that actually links against the ZAI SAPI and starts up and does like a hello-world from inside a request or something -- is that coming in another PR?

The ZAI SAPI is linked in to run the tests in #1186, but I can also add a generic hello-world test as part of that PR.

Thanks for the reviews @morrisonlevi and @labbati!

.circleci/config.yml
SammyK: Agreed 💯%.

.circleci/config.yml
labbati: I had missed the notification about this message, so I am answering late. For some reason, I thought that this was the job definition, while it is the workflow name (which is not subject to the issue I mentioned). Thanks for looking though it, though.

PR #1213, opened 2021-04-21
ZAI Sandbox

SammyK:
### Description

ZAI Sandbox prevents PHP code execution from throwing exceptions or raising errors if executed inside the sandbox. It ensures that if there is an existing error or unhandled exception when the sandbox is opened, the error and exception state will be unchanged after closing the sandbox.

This PR includes a number of other enhancements to the ZAI SAPI test runner including:

- Fake PHP frame injection (scoped to the `Zai\noop()` internal function)
- Last error introspection
- Exception throwing
- Unhandled exception introspection

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1017,31 +1017,74 @@ jobs:
             cmake -Bbuild -H. -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX=/opt/catch2
             cmake --build build/ --target install
 
-      - run:
-          name: Build and test Zend Abstract Interface (ASan)
-          command: |
-            export PATH="/opt/cmake/<< parameters.cmake_version >>/bin:$PATH"
-            mkdir -p /tmp/build/zai_asan
-            cd /tmp/build/zai_asan
-            if [ "<< parameters.php_version >>" = "7.4" || "<< parameters.php_version >>" = "8.0" ]
-            then
-              toolchain="-DCMAKE_TOOLCHAIN_FILE=~/datadog/cmake/asan.cmake"
-            fi
-            CMAKE_PREFIX_PATH=/opt/catch2 cmake \
-              ${toolchain} \
-              -DCMAKE_BUILD_TYPE=Debug \
-              -DBUILD_ZAI_TESTING=ON \
-              -DPHP_CONFIG=$(which php-config) \
-              ~/datadog/zend_abstract_interface
-            make -j all
-            make test
+      # Builds that support ASan
+      - when:
+          condition:
+            or:
+              - equal: [ "7.4", << parameters.php_version >> ]
+              - equal: [ "8.0", << parameters.php_version >> ]
+          steps:
+            - run:
+                name: Build and test Zend Abstract Interface (nts, debug, debug-zts-asan)
+                command: |
+                  export PATH="/opt/cmake/<< parameters.cmake_version >>/bin:$PATH"
+                  for phpBuild in nts debug debug-zts-asan
+                  do
+                    switch-php ${phpBuild}
+                    mkdir -p /tmp/build/zai-${phpBuild}
+                    cd /tmp/build/zai-${phpBuild}
+                    toolchain=""
+                    if [ "${phpBuild}" = "debug-zts-asan" ]
+                    then
+                      toolchain="-DCMAKE_TOOLCHAIN_FILE=~/datadog/cmake/asan.cmake"
+                    fi
+                    CMAKE_PREFIX_PATH=/opt/catch2 cmake \
+                      ${toolchain} \
+                      -DCMAKE_BUILD_TYPE=Debug \
+                      -DBUILD_ZAI_TESTING=ON \
+                      -DPHP_CONFIG=$(which php-config) \
+                      ~/datadog/zend_abstract_interface
+                    make -j all
+                    make test
+                  done
+
+      # Builds that do not support ASan
+      - when:
+          condition:
+            or:
+              - equal: [ "5.4", << parameters.php_version >> ]
+              - equal: [ "5.5", << parameters.php_version >> ]
+              - equal: [ "5.6", << parameters.php_version >> ]
+              - equal: [ "7.0", << parameters.php_version >> ]
+              - equal: [ "7.1", << parameters.php_version >> ]
+              - equal: [ "7.2", << parameters.php_version >> ]
+              - equal: [ "7.3", << parameters.php_version >> ]
+          steps:
+            - run:
+                name: Build and test Zend Abstract Interface (nts, debug, debug-zts)
+                command: |
+                  export PATH="/opt/cmake/<< parameters.cmake_version >>/bin:$PATH"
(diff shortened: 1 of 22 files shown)


zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -0,0 +1,374 @@
...
+        if (EG(current_execute_data)) {
+            EG(current_execute_data)->opline = EG(opline_before_exception);
+        }
+    }
+
+    if (es->exception) {
+        EG(exception) = es->exception;
+        EG(prev_exception) = es->prev_exception;
+
+        EG(opline_before_exception) = es->opline_before_exception;
```
Anilm3, line 345: I'm not entirely sure if I understood this properly, on PHP 7/8 you use `zend_clear_exception` which updates `EG(current_execute_data)->opline` with the one stored in `EG(opline_before_exception)` and then you call `zend_throw_exception_internal` which restores `EG(opline_before_exception)` with the one stored in `EG(current_execute_data)->opline` (so basically unchanged), wouldn't this still be the opline before the exception which occurred within the sandboxed code? If so, shouldn't you follow the same strategy and store the `opline_before_exception` in 7 and 8?

Anilm3 (commented):
Not much to say, impressive work. I still have to go through the tests but the main code looks fine.

zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -0,0 +1,374 @@
...
+    PG(last_error_type) = es->type;
+    PG(last_error_message) = es->message;
+    PG(last_error_file) = es->file;
+    PG(last_error_lineno) = es->lineno;
+    EG(error_reporting) = es->error_reporting;
+}
+
+inline void zai_sandbox_exception_state_backup(zai_exception_state *es) {
+    es->exception = NULL;
+    es->prev_exception = NULL;
```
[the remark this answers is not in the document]
SammyK: Excellent point. Implemented your suggestion. 👍

zend_abstract_interface/sandbox/sandbox.h
SammyK: Very nice catch! You are correct, this hidden state is not being treated the same way across all the PHP versions. And also on PHP 7+, `zend_throw_exception_hook` gets invoked an extra time when the sandbox closes because of calling `zend_throw_exception_internal`. Based on your feedback I updated this in 6cd3540 and added a test to ensure `zend_throw_exception_hook` does not get called an extra time.

I still think there is something off with how we're not backing up the original opline, but I haven't thought of a test to that would prove that this isn't working as expected yet.

Anilm3:
To me the PR looks good, but I think it would be worth having a second pair of eyes.

.circleci/config.yml
```diff
@@ -1017,31 +1017,74 @@ jobs:
...
+                      -DCMAKE_BUILD_TYPE=Debug \
+                      -DBUILD_ZAI_TESTING=ON \
+                      -DPHP_CONFIG=$(which php-config) \
+                      ~/datadog/zend_abstract_interface
+                    make -j all
+                    make test
+                  done
+
+      - run:
+          name: Build and test Zend Abstract Interface (debug + UBSan)
```
morrisonlevi, line 1082: ubsan runs on all PHP versions?

zend_abstract_interface/zai_sapi/zai_sapi.h
```diff
@@ -78,13 +78,49 @@ bool zai_sapi_append_system_ini_entry(const char *key, const char *value);
...
+ * active PHP execution stack. If there is no active execution context, the
+ * fake frame will be the first frame in the stack. If there is an existing
+ * stack active in the executor globals, the fake frame will be added to the
+ * top of the stack.
+ *
+ * Motivation: sometimes it is necessary to emulate code execution during
+ * runtime (e.g. a custom opcode handler) or to create an execution context
+ * for certain functionality (e.g. throw an exception when there is no active
+ * execution context). A fake frame provides an execution context for these
+ * cases.
```
morrisonlevi, line 91: Thanks for documenting this. I saw the `.c` file first and my first immediate thought was, "Why on earth are we pushing fake frames?" So I jumped to the header and there it is.

zend_abstract_interface/sandbox/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+                                       $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+                                       $<INSTALL_INTERFACE:include>)
+
+target_compile_features(zai_sandbox PUBLIC c_std_99)
+
+target_link_libraries(zai_sandbox PUBLIC "${PHP_LIB}")
+
+set_target_properties(zai_sandbox PROPERTIES
+                                  EXPORT_NAME Sandbox
+                                  VERSION ${PROJECT_VERSION})
```
morrisonlevi, line 13: I prefer that we don't align the initial indents like this. What do you think about this style?

```suggestion
target_include_directories(zai_sandbox
  PUBLIC
    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
    $<INSTALL_INTERFACE:include>)

target_compile_features(zai_sandbox PUBLIC c_std_99)

target_link_libraries(zai_sandbox PUBLIC "${PHP_LIB}")

set_target_properties(zai_sandbox PROPERTIES
  EXPORT_NAME Sandbox
  VERSION ${PROJECT_VERSION})
```

morrisonlevi (commented):
This is not my best review; I struggled to get into the right headspace. Additionally, I can't seem to build php from source on Catalina, so I didn't have some of the tools I normally use to review things.

I'll try to get my toolchain working and do a better review.

.circleci/config.yml
SammyK: It does, but since we don't have UBSan builds of PHP, I don't think UBSan will flag anything on the PHP side of things so it should only catch issues with the ZAI components.

zend_abstract_interface/sandbox/CMakeLists.txt
SammyK: Is there a standard formatting for listfiles? All of the examples I've seen so far feel a bit foreign to me. This formatting was inspired by `cmake-format` (see [example](https://github.com/cheshirekow/cmake_format#example)). My preference is just to decided on something that we can keep consistent and add it to the linter in CI. And bonus points if we can just run, `cmake-format -i CMakeLists.txt`, so that I don't have to think about anything at all. :)

morrisonlevi: I'm not aware of a "standard", but I agree that formatting isn't a very big deal if we can run a formatter tool.

morrisonlevi (approved):
Turns out all of the nits I found technically aren't part of this PR, but are in things already committed!

Don't forget to open a JIRA for moving the INI realloc stuff to a static buffer.

SammyK:
Thanks @Anilm3 & @morrisonlevi for the code reviews!

Issue #1214, opened 2021-04-22
[BUG] [0.58.0] failed tests


remicollet:
```
=====================================================================
FAILED TEST SUMMARY
---------------------------------------------------------------------
HTTP headers are sent to the Agent from the background sender [tests/ext/background-sender/agent_headers.phpt]
The container ID is sent via HTTP headers to the Agent [tests/ext/background-sender/agent_headers_container_id.phpt]
An empty container ID is not sent via HTTP headers to the Agent [tests/ext/background-sender/agent_headers_container_id_empty.phpt]
The Fargate 1.4+ container ID is sent via HTTP headers to the Agent [tests/ext/background-sender/agent_headers_container_id_fargate.phpt]
HTTP Agent headers are ignored from userland [tests/ext/background-sender/agent_headers_ignore_userland.phpt]
=====================================================================
```

All failing with `getaddrinfo failed` (http://request-replayer)

It will be nice to have such tests, relying on dev env skipped.

SammyK:
Thanks for the report @remicollet. This should be fixed in the next release via #1215.

labbati:
[0.59.0(https://github.com/DataDog/dd-trace-php/releases/tag/0.59.0) has been released and it solves this issue. In case of any problem please comment and we will reopen.

PR #1215, opened 2021-04-22
Skip tests that have development environment dependencies

SammyK:
### Description

This addresses the issue raised in #1214 where some tests fail when run outside of the Datadog development environment.

### Reviewer checklist
- [x] Milestone is set.


tests/ext/includes/skipif_requires_dev_env.inc
```diff
@@ -0,0 +1,10 @@
+<?php
```
labbati, line 1: @SammyK what if name it `tests/ext/includes/skipif_no_dev_env.inc`

labbati (approved):
👍🏻  with only one optional comment

tests/ext/includes/skipif_requires_dev_env.inc
SammyK: Done. :)

labbati (dismissed):
Thanks @SammyK

SammyK:
Thanks @labbati. Do you mind giving this another CR? I noticed that the tests for PECL were getting skipped due to the env vars not passing correctly. To that end the `tests/ext/integrations/curl` tests were also being skipped in the PECL tests for the same reason. That should be fixed now.

.circleci/config.yml
```diff
@@ -394,6 +394,7 @@ jobs:
       DDAGENT_HOSTNAME: 127.0.0.1
       DD_AGENT_HOST: 127.0.0.1
       HTTPBIN_HOSTNAME: httpbin_integration
+      DATADOG_HAVE_DEV_ENV: 1
```
morrisonlevi, line 397: Should everything run in CircleCI set `DATADOG_HAVE_DEV_ENV`?

SammyK: Ideally, yes. Do you know how to set it across the board?

morrisonlevi: Most CIs do have this capability. In CircleCI's case, it seems like we'd [have to use the UI](https://circleci.com/docs/2.0/env-vars/)?

labbati (approved):
LGTM! Thanks @SammyK

.circleci/config.yml
SammyK: Thanks @morrisonlevi. I'll merge this as-is for now to make sure it gets into the next release, but I'll look into adding this env var across the board from the UI.

SammyK: @morrisonlevi I've added `DATADOG_HAVE_DEV_ENV=1` via the UI but I have a concern: We have cleared out the CircleCI env vars in the past for various reasons so I'm a bit concerned that at one point we may inadvertently remove `DATADOG_HAVE_DEV_ENV` from the env vars and it will silently skip over a lot of crucial tests. So I'm thinking it might be better to keep this env var set explicitly in the CI config to reduce the chances of accidentally skipping over tests.

Alternatively we could add a job that checks that `DATADOG_HAVE_DEV_ENV == 1` and will fail if it is not set. That way if the env var is removed from the UI, at least that test will fail.

I don't have any strong opinions either way. WDYT?

morrisonlevi: Avoiding the UI seems fine to me; that's hidden state that seems like it could bite us.

PR #1216, opened 2021-04-22
Tidy the components/

morrisonlevi:
### Description

Remove unused macros.
Remove unused include.
Fix mis-spellings.
Use nullptr over NULL in C++ files.

Use C++ ranged-based for loops instead of calculating the size of
the array by hand (in C++ files only, ofc). Rename cases to servers
so that the singular form is a valid variable name (cannot have a
variable named 'case', since it's part of the langauge).

With these changes, the components/ directory is free from warnings
from clang-tidy, at least with the options enabled by default in
CLion.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ This cleanup does not require new tests.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/components/container_id/container_id.c
+++ b/components/container_id/container_id.c
@@ -38,9 +38,6 @@
  */
 #define TASK_REGEX "[0-9a-f]\\{32\\}-[0-9]\\{1,20\\}"  // Original ERE: "[0-9a-f]{32}-[0-9]+"
 
-#define MIN_ID_LEN DATADOG_PHP_CONTAINER_ID_MIN_LEN
-#define MAX_ID_LEN DATADOG_PHP_CONTAINER_ID_MAX_LEN
-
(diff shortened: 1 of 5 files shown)


SammyK (approved):
That _is_ tidy! :) 👍

labbati:
I like this a lot. Does CLion let you know or access a `.clang-tidy` file with the configured checks and options that we can add to our codebase and to our CI checks?

morrisonlevi:
> I like this a lot. 

Thanks!

> Does CLion let you know or access a .clang-tidy file with the configured checks and options that we can add to our codebase and to our CI checks?

I don't know! I'll poke around a bit.

I found this stuff while looking at components to document them; I wasn't intentionally doing tool-based analysis cleanup ^_^

PR #1220, opened 2021-04-27
Nette integration

Anilm3:
### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


SammyK (commented):
This is looking great @Anilm3! 💯 I just tried to view the traces in the UI with an example Nette 3 app and I'm not seeing the spans yet.

Screen Shot 2021-05-03 at 5 40 40 PM

But it could be something on my end that I'm missing so I'll give this another look tomorrow morning.

src/DDTrace/Integrations/Nette/NetteIntegration.php
```diff
@@ -0,0 +1,173 @@
...
+    {
+        if (!self::shouldLoad(self::NAME)) {
+            return self::NOT_AVAILABLE;
+        }
+
+        $integration = $this;
+
+        \DDTrace\hook_method('Nette\Configurator', '__construct', function () use ($integration) {
+            $integration->load();
+        });
```
morrisonlevi, line 46: I think because this is defer-loaded you should call `integration->load()` instead of doing it though `hook_method`. I think.

Anilm3:
> This is looking great @Anilm3! 💯 I just tried to view the traces in the UI with an example Nette 3 app and I'm not seeing the spans yet.
> 
> Screen Shot 2021-05-03 at 5 40 40 PM
> 
> But it could be something on my end that I'm missing so I'll give this another look tomorrow morning.

This seems to be happening with the latest version of Nette, I'm investigating.

src/DDTrace/Integrations/Nette/NetteIntegration.php
SammyK: Nice catch @morrisonlevi! 

@Anilm3 As Levi pointed out, when you add this to the deferred loader, you no longer need to use `DDTrace\hook_method`. When I made this change locally, I can see the Nette spans in the UI now.

Screen Shot 2021-05-04 at 9 53 21 AM

Anilm3: For some reason I wasn't getting the spans with or without this with the Nette test application, but I think I had something not loading properly, indeed after setting everything correctly and removing removing the `hook_method` everything seems to be working as expected. Thanks!

Anilm3: Actually this had some unexpected consequences as the root span was always overridden to be a `nette.request`, trying a different approach.

morrisonlevi: My guess is that you are _also_ adding it to something in IntegrationsLoader -- IIRC we do not do that if they are deferred loaded. Is that right @SammyK ?

SammyK: Correct, deferred-loaded integrations only need to be [added to the `IntegrationsLoader` for PHP 5](https://github.com/DataDog/dd-trace-php/blob/9e0699a/src/DDTrace/Integrations/IntegrationsLoader.php#L104-L120) because PHP 5 does not support deferred loading.

Anilm3: Thanks, this was the cause of the problem I was facing, things are working as expected now and I modified the code to account for the lack of deferred loading in PHP 5.

tests/Integrations/Nette/V2_4/NetteTest.php
```diff
@@ -0,0 +1,202 @@
...
+    protected static function getAppIndexScript()
+    {
+        return __DIR__ . '/../../../Frameworks/Nette/Version_2_4/www/index.php';
+    }
+
+    protected static function getEnvs()
+    {
+        return array_merge(parent::getEnvs(), [
+            'DD_SERVICE' => 'nette_test_app',
+            'DD_TRACE_DEBUG' => true,
```
SammyK, line 22: Was this left in from debugging? (Same for the v3 test below.)

src/DDTrace/Integrations/Nette/NetteIntegration.php
```diff
@@ -0,0 +1,176 @@
...
+                }
+
+                $request = $args[0];
+                $presenter = $request->getPresenterName();
+                $action = $request->getParameter('action');
+
+                $root->setTag(Tag::HTTP_METHOD, $request->getMethod());
+                $root->setTag('nette.route.presenter', $presenter);
+                $root->setTag('nette.route.action', $action);
+                $root->setResource($presenter . ':' . $action);
```
SammyK, line 119: This is how we used to set the root span resource names, but since the route data is already being stored in the span metadata, the root span resource name will default to the HTTP method + URL endpoint. So we can remove this.

SammyK (commented):
Thanks for figuring out the workarounds for the various integration loading methods across the PHP versions. I left a few final tiny bits of feedback (and one to have Levi confirm) but this is pretty close to mergeable. 👍

tests/Integrations/Nette/V2_4/NetteTest.php
Anilm3: If I'm not mistaken the `DD_SERVICE` is required so that I can match it to the expected value on the spans, however the `DD_TRACE_DEBUG` was just set for debugging, are there any downsides to keeping it in the test? Other tests also seem to have it.

SammyK: Yes, `DD_SERVICE` is expected, but ideally we'd be testing the integration without `DD_TRACE_DEBUG=1` unless we're testing for a specific debug message (which is pretty common in the phpt tests.) But other than that, if there are other tests that have debug mode turned on, it's most likely accidentally left over from debugging. :)

src/DDTrace/Integrations/Nette/NetteIntegration.php
```diff
@@ -0,0 +1,176 @@
...
+        $scope = GlobalTracer::get()->getRootScope();
+        if (!$scope instanceof Scope) {
+            return;
+        }
+
+        $service = \ddtrace_config_app_name(NetteIntegration::NAME);
+
+        $root = $scope->getSpan();
+        $root->setTraceAnalyticsCandidate();
+        $root->overwriteOperationName('nette.request');
```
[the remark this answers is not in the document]
morrisonlevi: I think it's much better if we don't.

src/DDTrace/Integrations/Nette/NetteIntegration.php
```diff
@@ -0,0 +1,176 @@
...
+
+        \DDTrace\trace_method(
+            'Nette\Configurator',
+            'createRobotLoader',
+            function (SpanData $span) use ($service) {
+                $span->name = 'nette.configurator.createRobotLoader';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $service;
+            }
+        );
```
morrisonlevi, line 86: Is there any other meta-data that can be attached to `createRobotLoader` or `createContainer`? Based on their names, they seem like the kind of thing we should trace only if there is additional meta-data.

src/DDTrace/Integrations/Nette/NetteIntegration.php
```diff
@@ -0,0 +1,176 @@
...
+
+        \DDTrace\trace_method(
+            'Nette\Application\Routers\RouteList',
+            'match',
+            function (SpanData $span) use ($service) {
+                $span->name = 'nette.router.match';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $service;
+            }
+        );
```
morrisonlevi, line 131: Is there a concept of "route name" or something we could attach here?

src/DDTrace/Integrations/Nette/NetteIntegration.php
```diff
@@ -0,0 +1,176 @@
...
+                $span->name = 'nette.latte.render';
+                $span->type = Type::WEB_SERVLET;
+                $span->service = $service;
+
+                if (count($args) >= 1) {
+                    $span->meta['nette.latte.templateName'] = $args[0];
+                }
+            }
+        );
+    }
```
morrisonlevi, line 175: @SammyK At one point we talked about instrumenting components of frameworks like templating systems individually in case they get mixed and matched, which some small percentage of customers do, though I don't think `latte/latte` is at the top of that list. Is that something you still think we should do? If so, should we start now?

SammyK: > Is that something you still think we should do? If so, should we start now?

I think we should totally do this but I think this makes a lot more sense starting with Symfony components. I would like spend the time to put together a strategy to target Symfony components specifically and then expand from there. So, yes, but only after we have a clear plan of action. :)

src/DDTrace/Integrations/Nette/NetteIntegration.php
Anilm3: As far as I can tell there's really no useful data associated with them, they were just ported over from the original PR.

src/DDTrace/Integrations/Nette/NetteIntegration.php
Anilm3: Not really, the match function on the route list just matches the http request object itself, again this was just ported over from the original PR.

src/DDTrace/Integrations/Nette/NetteIntegration.php
Anilm3: Done

tests/Integrations/Nette/V2_4/NetteTest.php
Anilm3: I've removed it.

src/DDTrace/Integrations/Nette/NetteIntegration.php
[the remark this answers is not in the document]
Anilm3: Removed!

SammyK (dismissed):
LGTM. 👍 Nice work @Anilm3! 🎉

src/DDTrace/Integrations/Nette/NetteIntegration.php
morrisonlevi: @SammyK I think we should drop these ones; what do you think?

src/DDTrace/Integrations/Nette/NetteIntegration.php
morrisonlevi: @SammyK  What about dropping this one as well? No meta-data, including resource.

tests/Unit/Integrations/IntegrationsLoaderTest.php
```diff
@@ -167,6 +167,7 @@ public function testWeDidNotForgetToRegisterALibraryForAutoLoading()
             $excluded[] = 'slim';
             $excluded[] = 'wordpress';
             $excluded[] = 'yii';
+            $excluded[] = 'nette';
```
morrisonlevi, line 170: This list appears to be sorted; can you please stick it between memcached and pdo?

Anilm3: Done.

src/DDTrace/Integrations/Nette/NetteIntegration.php
SammyK: @morrisonlevi My apologies; I discussed this with @Anilm3 off thread and forgot to update this thread with the discussion:

> As a general rule, I think it is still useful to trace functions even if there is no metadata if that function has the possibility to call something slow. For example, some sort of plugin loader; maybe on most requests it’s a small span with no metadata, but if there’s a slow plugin doing something strange on intermittent requests, it’s nice to know it’s happening from the plugin loader. I don’t know what `createRobotLoader` does, but if it loads possibly slow application logic, it might be worth tracing still.

It looks like the [Robot Loader](https://github.com/nette/robot-loader) does some crawling+compiling of PHP files and caches the classes/paths. So perhaps it would be nice to see the non-cached timing of that? It can also [throw an exception](https://github.com/nette/bootstrap/blob/8df6751/src/Bootstrap/Configurator.php#L200-L218) so the exception would have a span to attach to. 

For `createContainer`, I'm not sure if [container initialization has potential to be slow](https://github.com/nette/bootstrap/blob/8df6751/src/Bootstrap/Configurator.php#L231-L243), but if it did, I would think it beneficial to instrument.

But I don't have enough experience with Nette to say one way or another, so I'll leave the final call and PR approval up to you @morrisonlevi.

src/DDTrace/Integrations/Nette/NetteIntegration.php
SammyK: This one calls the parent `Nette\Routing\RouteList::match()` and [I don't see how it would ever be "slow"](https://github.com/nette/routing/blob/c714f20/src/Routing/RouteList.php#L44-L72) per-se so I think this one could probably be removed.

Anilm3: I'll remove it then since there seems to be some consensus.

src/DDTrace/Integrations/Nette/NetteIntegration.php
morrisonlevi: > As a general rule, I think it is still useful to trace functions even if there is no metadata if that function has the possibility to call something slow.

I believe that in general this should be a concern for the profiler. That doesn't have to be absolute, but I think the exceptions should be for major framework concepts that are prominent in the docs. Are these things critical and major pieces to Nette applications? My gut says no, but I haven't looked at the Nette docs much.

Also, @Anilm3, message me on Slack because I think we should look at a Nette app with the profiler on -- when I did that with Laravel it was really illuminating (ha, an unintentional pun).

morrisonlevi: I took a look in the [v3 docs](https://api.nette.org/3.0/Nette.html) and the robot loader does seem worth instrumenting. Ideally we'd try to find some meta-data for it, but that kind of thing can be added later.

I'm not sure about createContainer. It seems like tracing something in Container could be useful; maybe remove this for now?

morrisonlevi (commented):
Nette 3.1 was released in January. Was it intentionally excluded for some reason? Also, I don't know how Nette does versioning, so I don't know if we should drop 3.0 for 3.1, or if we should do both (depends on if they do semantic versioning or not).

Anilm3:
> Nette 3.1 was released in January. Was it intentionally excluded for some reason? Also, I don't know how Nette does versioning, so I don't know if we should drop 3.0 for 3.1, or if we should do both (depends on if they do semantic versioning or not).

Not necessarily excluded, I did test with it and everything worked as expected. Is it common practice to have tests for each minor version or is it enough with just a major version?

src/DDTrace/Integrations/Nette/NetteIntegration.php
Anilm3: I've removed `createContainer`, thanks for double checking.

SammyK (approved):
Thanks for sticking in there with the back-and-forth @Anilm3!

PR #1222, opened 2021-04-28
Fix ingestion control reporting

labbati:
### Description

`_dd.rule_psr` has always to be set. Previously we were only setting it for sampling rules. It is necessary to have the correct values in the ingestion control pages.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/bridge/configuration.php
+++ b/bridge/configuration.php
@@ -93,6 +93,9 @@ function _ddtrace_config_json($value, $default)
         return $default;
     }
 
+    // If the char `'` used to escape the json object reaches this variable, it has to be removed.
+    $value = trim($value, "'");
+
     $parsed = \json_decode($value, true);
     if (null === $parsed) {
         return $default;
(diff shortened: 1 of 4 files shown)


morrisonlevi (dismissed):
I think the code is cleaner in this style as well, with `computePrioritySampling` only being called in one place instead of two. Good work.

labbati:
Thanks for the CR @morrisonlevi. I had to add https://github.com/DataDog/dd-trace-php/pull/1222/commits/0681320ce0a89bac6dbe92a3d0acbe92839f848d to cover the use case linked in the commit comment.

morrisonlevi (approved):
Approved. Something to check for the future is if simple `trim` is adequate or if we need a more advanced decoder.

labbati:
Good point. And I think we need more use cases. This comes from manually checking what happens with nginx.

If you set `env[SOME] = {"a": 1}` you get `{a: 1}`, and is beyond my understanding why this would ever be the case you need. For example this is not true for Apache. In any case this is a quick fix for now.

Thanks for the review @morrisonlevi

PR #1229, opened 2021-05-11
Detect ZAI ZMM leaks

SammyK:
### Description

This PR adds a way to detect Zend Memory Manager leak messages in the ZAI tests by grepping the test log in CI for:

```
=== Total n memory leaks detected ===
```

There were a few ZMM leaks that were isolated to the tests (the leaks were not in the tracer or any production code paths) and those are fixed in d8b4c3e.

Ideally the leak detection would be implemented as part of the ZAI SAPI test runner, but every way I tired to implement it, I hit a blocker.

- ZMM bypasses the SAPI error logger and [writes the leak message directly to stderr](https://github.com/php/php-src/blob/PHP-8.0.4/main/main.c#L1588) which I haven't been able to capture from ZAI SAPI during RSHUTDOWN.
- There is no way to inject a [custom hook](https://github.com/php/php-src/blob/PHP-8.0.4/main/main.c#L2034) to capture the ZMM memory leak messages.
- I cannot get Catch2 to capture stdout or stderr from ZAI SAPI using a [custom event listener](https://github.com/catchorg/Catch2/blob/devel/docs/event-listeners.md);  `ZmmLeakListener`. The event listener works, but `testCaseStats.stdOut` and `testCaseStats.stdErr` are always empty from `testCaseEnded`.
- Changing the Catch2 reporter to one that captures output, [like `junit`](https://github.com/catchorg/Catch2/blob/2c269eb/src/catch2/reporters/catch_reporter_junit.cpp#L63), the custom `ZmmLeakListener` still cannot access ZAI SAPI stdout and stderr.
- Changing ZAI SAPI write operations from `write` to `fwrite` does not seem to make any difference.
- It seems as if the reporter doesn't even have access to the output streams; for `junit`, `<system-out/>` and `<system-err/>` are always empty.
- There doesn't appear to be a way to run a post-test script using cmake.

That leaves me with "manually" grepping the test log for the ZMM leak message in CI. I'm probably missing something obvious with the Catch2 setup, so if anyone else can figure out a better solution to this, I'd love to know your thoughts.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/zend_abstract_interface/methods/tests/php5/methods.cc
+++ b/zend_abstract_interface/methods/tests/php5/methods.cc
@@ -737,6 +737,8 @@ TEST_CASE("call static method: non-static method (userland)", "[zai_methods]") {
     REQUIRE(retval != NULL);
     REQUIRE(Z_TYPE_P(retval) == IS_BOOL);
 
+    zval_ptr_dtor(&retval);
+
     ZAI_SAPI_ABORT_ON_BAILOUT_CLOSE()
     zai_sapi_spindown();
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1098,6 +1098,7 @@ jobs:
                       ~/datadog/zend_abstract_interface
                     make -j all
                     make test
+                    grep -e "=== Total .* memory leaks detected ===" Testing/Temporary/LastTest.log && exit 1 || true
                   done
 
       # Builds that do not support ASan
@@ -1128,6 +1129,7 @@ jobs:
                       ~/datadog/zend_abstract_interface
                     make -j all
                     make test
+                    grep -e "=== Total .* memory leaks detected ===" Testing/Temporary/LastTest.log && exit 1 || true
                   done
 
       - run:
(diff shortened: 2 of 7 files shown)


zend_abstract_interface/methods/tests/php5/methods.cc
```diff
@@ -737,6 +737,8 @@ TEST_CASE("call static method: non-static method (userland)", "[zai_methods]") {
     REQUIRE(retval != NULL);
     REQUIRE(Z_TYPE_P(retval) == IS_BOOL);
 
+    zval_ptr_dtor(&retval);
```
morrisonlevi, line 740: I assume this was found via the added memory leak detection?

morrisonlevi (approved):
Looks good to me!

zend_abstract_interface/methods/tests/php5/methods.cc
SammyK: It was. :)

PR #1234, opened 2021-05-13
ZAI function-call seam (PHP 8)

SammyK:
### Description

This PR adds a Zend Abstract Interface (ZAI) seam for safely calling (sandboxed) PHP functions from C in PHP 8:

- `bool zai_call_function_literal(const char *name, zval *retval, ...);`
- `bool zai_call_function(const char *name, size_t name_len, zval *retval, ...);`
- `bool zai_call_function_ex(const char *name, size_t name_len, zval *retval, int argc, ...);`

This seam will initially be used to call the [userland deferred-loading function](https://github.com/DataDog/dd-trace-php/blob/5b8ea36/ext/php8/php8/engine_hooks.c#L127-L160) in PHP 8.

This PR also includes a new macro for asserting lowercase strings (`zai_assert/zai_assert.h`):

```c
zai_assert_is_lower(str, "String must be lowercase.");
```

As with `assert()`, `zai_assert_is_lower()` will only abort on debug builds of PHP.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/zend_abstract_interface/functions/CMakeLists.txt
+++ b/zend_abstract_interface/functions/CMakeLists.txt
@@ -1,0 +1,28 @@
+add_library(zai_functions "${PHP_VERSION_DIRECTORY}/functions.c")
+
+target_include_directories(zai_functions PUBLIC
+                                         $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+                                         $<INSTALL_INTERFACE:include>)
+
+target_compile_features(zai_functions PUBLIC c_std_99)
+
+target_link_libraries(zai_functions PUBLIC "${PHP_LIB}" Zai::Sandbox)
+
+set_target_properties(zai_functions PROPERTIES
+                                    EXPORT_NAME Functions
+                                    VERSION ${PROJECT_VERSION})
+
+add_library(Zai::Functions ALIAS zai_functions)
+
+if (${BUILD_ZAI_TESTING})
+  add_subdirectory(tests)
+endif()
+
+# This copies the include files when `install` is ran
+# TODO: How to make this zai/functions.h?
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/functions.h
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/functions/)
+
+target_link_libraries(zai_zend_abstract_interface INTERFACE zai_functions)
+
+install(TARGETS zai_functions EXPORT ZendAbstractInterfaceTargets)
--- a/zend_abstract_interface/functions/functions.h
+++ b/zend_abstract_interface/functions/functions.h
@@ -1,0 +1,32 @@
+#ifndef ZAI_FUNCTIONS_H
+#define ZAI_FUNCTIONS_H
+
+#include <main/php.h>
+#include <stdbool.h>
+
+/* Work in progress
+ *
+ * The long-term goal is to provide ZAI data-structure shims so that the
+ * following APIs are the same across all PHP versions. For now we will provide
+ * ZAI seams using native Zend Engine data structures.
+ */
+
+/* Calls a function with 'argc' number of zval arguments. Caller owns the
+ * arguments and the 'retval'. Caller is responsible for freeing any refcounted
+ * arguments. Caller must provide a pointer to a stack-allocated zval for the
+ * return value, 'retval'. In error cases, 'retval' will be set to IS_UNDEF.
+ * Caller must dtor the 'retval' after the call:
+ *
+ *   zval_ptr_dtor(&retval);
+ *
+ * Functions cannot be called outside of a request context so this MUST be
+ * called from within a request context (after RINIT and before RSHUTDOWN).
+ */
+bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc, ...);
+
+/* Calls a function the same way as zai_call_function() but without passing any
+ * arguments to the function.
+ */
+bool zai_call_function_without_args(const char *name, size_t name_len, zval *retval);
+
+#endif  // ZAI_FUNCTIONS_H
--- a/zend_abstract_interface/functions/functions.h
+++ b/zend_abstract_interface/functions/functions.h
@@ -26,17 +26,16 @@ bool zai_call_function_ex(const char *name, size_t name_len, zval *retval, int a
 
 /* A wrapper for zai_call_function_ex() that automatically populates 'argc'. */
 #define zai_call_function(name, name_len, retval, ...) \
-    zai_call_function_ex(name, name_len, retval, ZAI_CALL_FUNCTION_VA_ARG_COUNT(__VA_ARGS__), ## __VA_ARGS__)
+    zai_call_function_ex(name, name_len, retval, ZAI_CALL_FUNCTION_VA_ARG_COUNT(__VA_ARGS__), ##__VA_ARGS__)
 
 /* A convenience wrapper to call zai_call_function() using a string literal.
  * This API only works when the function name is a string literal. If the
  * function name exists as a 'const char *', use zai_call_function() or
  * zai_call_function_ex() directly. */
-#define zai_call_function_literal(name, retval, ...) \
-    zai_call_function(name, sizeof(name) - 1, retval, ## __VA_ARGS__)
+#define zai_call_function_literal(name, retval, ...) zai_call_function(name, sizeof(name) - 1, retval, ##__VA_ARGS__)
 
 #define ZAI_CALL_FUNCTION_VA_ARG_COUNT(...) \
-    ZAI_CALL_FUNCTION_VA_ARG_MAX(ignore, ## __VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+    ZAI_CALL_FUNCTION_VA_ARG_MAX(ignore, ##__VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1, 0)
 #define ZAI_CALL_FUNCTION_VA_ARG_MAX(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10, ...) arg10
 
 #endif  // ZAI_FUNCTIONS_H
--- a/zend_abstract_interface/functions/functions.h
+++ b/zend_abstract_interface/functions/functions.h
@@ -22,11 +22,12 @@
  * Functions cannot be called outside of a request context so this MUST be
  * called from within a request context (after RINIT and before RSHUTDOWN).
  */
-bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc, ...);
+bool zai_call_function_ex(const char *name, size_t name_len, zval *retval, int argc, ...);
 
-/* Calls a function the same way as zai_call_function() but without passing any
- * arguments to the function.
- */
-bool zai_call_function_without_args(const char *name, size_t name_len, zval *retval);
+#define zai_call_function(name, name_len, ...) \
+    zai_call_function_ex(name, name_len, ZAI_CALL_FUNCTION_VA_ARG_COUNT(__VA_ARGS__), ## __VA_ARGS__)
+#define ZAI_CALL_FUNCTION_VA_ARG_COUNT(...) \
+    ZAI_CALL_FUNCTION_VA_ARG_MAX(ignore, ## __VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+#define ZAI_CALL_FUNCTION_VA_ARG_MAX(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10, ...) arg10
 
 #endif  // ZAI_FUNCTIONS_H
--- a/zend_abstract_interface/functions/functions.h
+++ b/zend_abstract_interface/functions/functions.h
@@ -24,8 +24,17 @@
  */
 bool zai_call_function_ex(const char *name, size_t name_len, zval *retval, int argc, ...);
 
-#define zai_call_function(name, name_len, ...) \
-    zai_call_function_ex(name, name_len, ZAI_CALL_FUNCTION_VA_ARG_COUNT(__VA_ARGS__), ## __VA_ARGS__)
+/* A wrapper for zai_call_function_ex() that automatically populates 'argc'. */
+#define zai_call_function(name, name_len, retval, ...) \
+    zai_call_function_ex(name, name_len, retval, ZAI_CALL_FUNCTION_VA_ARG_COUNT(__VA_ARGS__), ## __VA_ARGS__)
+
+/* A convenience wrapper to call zai_call_function() using a string literal.
+ * This API only works when the function name is a string literal. If the
+ * function name exists as a 'const char *', use zai_call_function() or
+ * zai_call_function_ex() directly. */
+#define zai_call_function_literal(name, retval, ...) \
+    zai_call_function(name, sizeof(name) - 1, retval, ## __VA_ARGS__)
+
 #define ZAI_CALL_FUNCTION_VA_ARG_COUNT(...) \
     ZAI_CALL_FUNCTION_VA_ARG_MAX(ignore, ## __VA_ARGS__, 8, 7, 6, 5, 4, 3, 2, 1, 0)
 #define ZAI_CALL_FUNCTION_VA_ARG_MAX(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8, arg9, arg10, ...) arg10
--- a/zend_abstract_interface/functions/php8/functions.c
+++ b/zend_abstract_interface/functions/php8/functions.c
@@ -15,7 +15,7 @@ bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc
      */
     ZVAL_UNDEF(retval);
 
-    if (!name || !name_len) return false;
+    if (!name || !name_len || argc < 0) return false;
 
     /* Functions cannot be called outside of a request context.
      * PG(modules_activated) indicates that all of the module RINITs have been
@@ -64,30 +64,20 @@ bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc
      * also allows us to avoid an unnecessary call to
      * zend_fcall_info_args_clear() since the 'param{s|_count}' members in
      * 'zend_fcall_info' will always be zeroed out in our case.
-     *
-     * Based on zend_fcall_info_argv():
-     * https://github.com/php/php-src/blob/PHP-8.0.4/Zend/zend_API.c?#L3608-L3622
      */
+    zval params[argc];
     if (argc > 0) {
         va_list argv;
         va_start(argv, argc);
-
-        zval *arg;
-        fci.param_count = (uint32_t)argc;
-        fci.params = (zval *)ecalloc(1, (fci.param_count * sizeof(zval)));
-
         for (uint32_t i = 0; i < (uint32_t)argc; ++i) {
-            arg = va_arg(argv, zval *);
-            if (!arg) {
-                /* Only dtor the zvals that were successfully copied */
-                fci.param_count = i;
-                zend_fcall_info_args_clear(&fci, /* free_mem */ true);
-                return false;
-            }
-            ZVAL_COPY(&fci.params[i], arg);
+            zval *arg = va_arg(argv, zval *);
+            if (!arg) return false;
+            params[i] = *arg;
         }
-
         va_end(argv);
+
+        fci.param_count = (uint32_t)argc;
+        fci.params = params;
     }
 
     bool ret = false;
@@ -112,13 +102,11 @@ bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc
          * best we can and then bubble up the zend_bailout.
          */
         zai_sandbox_close(&sandbox);
-        zend_fcall_info_args_clear(&fci, /* free_mem */ (argc != 0));
         zend_bailout();
     }
     zend_end_try();
 
     zai_sandbox_close(&sandbox);
-    zend_fcall_info_args_clear(&fci, /* free_mem */ (argc != 0));
 
     return ret;
 }
--- a/zend_abstract_interface/functions/php8/functions.c
+++ b/zend_abstract_interface/functions/php8/functions.c
@@ -59,6 +59,12 @@ bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc
     fci.size = sizeof(zend_fcall_info);
     fci.retval = retval;
 
+    bool should_bail = false;
+    zend_result call_fn_result = FAILURE;
+
+    zai_sandbox sandbox;
+    zai_sandbox_open(&sandbox);
+
     /* We add the zval args directly from the variable arguments, va_arg(),
      * instead of using zend_fcall_info_argv() because:
      *
@@ -68,13 +74,16 @@ bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc
      *   - We avoid an unnecessary heap allocation for the params.
      *   - We avoid unnecessary calls to zend_fcall_info_args_clear().
      */
-    zval params[argc];
     if (argc > 0) {
+        zval params[argc];
         va_list argv;
         va_start(argv, argc);
         for (uint32_t i = 0; i < (uint32_t)argc; ++i) {
             zval *arg = va_arg(argv, zval *);
-            if (!arg) return false;
+            if (!arg) {
+                zai_sandbox_close(&sandbox);
+                return false;
+            }
             /* Although we could copy the zval arg into the params array with
              * direct assignment:
              *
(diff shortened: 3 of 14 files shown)


zend_abstract_interface/functions/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+                                    VERSION ${PROJECT_VERSION})
+
+add_library(Zai::Functions ALIAS zai_functions)
+
+if (${BUILD_ZAI_TESTING})
+  add_subdirectory(tests)
+endif()
+
+# This copies the include files when `install` is ran
+# TODO: How to make this zai/functions.h?
```
morrisonlevi, line 22: To answer your question on making it `zai/functions.h`, you would set the `DESTINATION` to `"${CMAKE_INSTALL_INCLUDEDIR}/zai"` but to make that _also_ work prior to install (the common case for us) you need to adjust the directory structure so that it lives in `zai/functions.h`.

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,139 @@
...
+#ifndef NDEBUG
+static bool z_is_lower(const char *str) {
+    char *p = (char *)str;
+    while (*p) {
+        if (isalpha(*p) && !islower(*p)) return false;
+        p++;
+    }
+    return true;
+}
+#endif
```
morrisonlevi, line 15: I think this is missing an include for `<ctype.h>` when it's a debug build.

zend_abstract_interface/functions/tests/php8/functions.cc
```diff
@@ -0,0 +1,381 @@
+extern "C" {
+#include "functions/functions.h"
+#include "zai_sapi/zai_sapi.h"
```
morrisonlevi, line 3: These should use `#include <...>` right? I think that's what we decided?

zend_abstract_interface/sandbox/php8/sandbox.c
```diff
@@ -4,7 +4,30 @@ extern inline void zai_sandbox_open(zai_sandbox *sandbox);
...
 extern inline void zai_sandbox_error_state_backup(zai_error_state *es);
-extern inline void zai_sandbox_error_state_restore(zai_error_state *es);
+
+/* In PHP 8, we cannot extern inline zai_sandbox_error_state_restore() because
+ * zend_string_release() is static which generates the following compiler
+ * warning:
+ *
+ *   "warning: 'zend_string_release' is static but used in inline function
+ *    'zai_sandbox_error_state_restore' which is not static"
+ */
```
morrisonlevi, line 14: Just to document some history: I tried talking Nikita into allowing me to fix these in php-src but he didn't see the value in doing `inline` + `extern inline` vs allowing the compiler to do it via `static inline`. Plus there are `zend_always_inline` cases which have _theoretically_ been put there because of performance reasons, and we wouldn't want them to accidentally _not_ get inlined (although "always" is technically a lie anyway). I didn't feel like doing the necessary performance A/B work to demonstrate there weren't regressions and dropped it.

tl;dr It'll probably never get fixed upstream.

zend_abstract_interface/functions/tests/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+#[[ TODO When we have a consistent API across all PHP versions with ZAI
+    data-structure shims, move these tests out of PHP-versioned directories.
+ ]]
+add_executable(functions "${PHP_VERSION_DIRECTORY}/functions.cc")
+
+target_link_libraries(functions PUBLIC catch2_main Zai::Sapi Zai::Functions)
+
+file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/stubs
+     DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
morrisonlevi, line 9: I understand why the copy is necessary, but I don't like it for some reason I can't place. Did you have a similar feeling or is it just me?

morrisonlevi (commented):
I still have more review to do, but posting these comments for now.

zend_abstract_interface/functions/CMakeLists.txt
SammyK: Ah yes, I remember our conversation on this a while back. I believe we decided to not put all the headers in the same directory for some reason, but I can't remember the reason. I think we decided to keep it `thing/thing.h` until we could figure out a proper solution that we liked. Since this is also something that impacts `libddprof`, I would love to sync up with you on this again to iron out a direction we want to go.

zend_abstract_interface/functions/tests/php8/functions.cc
SammyK: I think using `<>` makes since from the perspective of including these from the tracer since they live outside of the tracer project. But since these are tests for testing _within_ the ZAI project, `""` seems to make more sense to me.

In general, this feels most natural to me:
- `<>` = external project header
- `""` = internal project header

WDYT?

zend_abstract_interface/functions/tests/CMakeLists.txt
SammyK: The only quirky thing that I've experienced with this copy is that if you make a change to a stub, you basically have to blow away the build dir and start over again (at least that's what I do with my primitive `cmake` skills. 😆) But I'm not sure how else we'd get the stubs copied over for the tests without doing a `COPY`. Of course I'm totally open to ideas if you've got a better solution!

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,139 @@
...
+    zai_sandbox sandbox;
+    zai_sandbox_open(&sandbox);
+
+    zend_try {
+        zend_result result = zend_call_function(&fci, &fcc);
+        /* An unhandled exception will not result in a zend_bailout if there is
+         * an active execution context. This is a failed call if an exception
+         * was thrown. The sandbox will clean up our mess when it closes.
+         */
+        ret = (result == SUCCESS && !EG(exception));
```
labbati, line 115: Not a change required for this PR, but rather an observation for the design of future zai modules. Do you think that a binary return state is enough to communicate the outcome? Or it would make sense to have at least 3 states:

1. `SUCCESS`;
2. error because of something you did wrong while developing (e.g. calling with a `NULL` return value);
3. runtime error (e.g. and exception because the database is not reachable).

zend_abstract_interface/functions/tests/php8/functions.cc
```diff
@@ -0,0 +1,381 @@
+extern "C" {
```
labbati, line 1: @SammyK I really love this testing infrastructure. You made an awesome job!

labbati (commented):
Great work @SammyK, I asked a few questions, but as always mostly for me to learn things :)

zend_abstract_interface/functions/php8/functions.c
SammyK: Nice catch. Since this is a pretty common use case in the Zend Engine, and it is [already being used for methods & class lookups](https://github.com/DataDog/dd-trace-php/blob/f4b5065/zend_abstract_interface/methods/php5/methods.c#L7-L16) as well, I was thinking of extracting this into `zai_assert/zai_assert.h` as something like this:

```c
zai_assert_is_lower(str, "String must be lowercase.");
```

Do you think it's worth adding that as part of this PR?

zend_abstract_interface/functions/tests/CMakeLists.txt
morrisonlevi: I think you need to add the copied file as a dependency of the test using `add_dependency`, though I'm not sure a `COPY` is recognized as a "factory" (I use quotes because I'm not sure what the proper CMake terminology is).

zend_abstract_interface/functions/functions.h
```diff
@@ -0,0 +1,32 @@
...
+ * The long-term goal is to provide ZAI data-structure shims so that the
+ * following APIs are the same across all PHP versions. For now we will provide
+ * ZAI seams using native Zend Engine data structures.
+ */
+
+/* Calls a function with 'argc' number of zval arguments. Caller owns the
+ * arguments and the 'retval'. Caller is responsible for freeing any refcounted
+ * arguments. Caller must provide a pointer to a stack-allocated zval for the
+ * return value, 'retval'. In error cases, 'retval' will be set to IS_UNDEF.
+ * Caller must dtor the 'retval' after the call:
```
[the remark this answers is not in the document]
SammyK: I'm not sure I understand what you're suggesting in terms of the refcount handling, but we certainly can (and will 🙂) add an abstraction layer/wrapper at the tracer level. (For example, auto-flushing on PHP 5 wraps the method-call seam with an [API that ignores the retval](https://github.com/DataDog/dd-trace-php/blob/5275a4f/ext/php5/auto_flush.c#L9-L17).)

On the refcounting bit - I think the comment might be a bit confusing the way it is worded.

- Basically the caller 100% owns the args (ctoring/passing/dtoring).
- The caller also owns the space for the retval (ideally stack-allocated).
- The only memory `zai_call_function` allocates is the space to hold the params to pass to the function and it is freed before returning to the caller. (As a side note we might want to consider a hard-limit on the number of args so that we can stack-allocate the params. 🤔)
- The retval may or may not need to be dtored depending on whether or not the PHP function returns a refcounted value (e.g. array, object, string, etc). But that should be known to the caller so the caller can make that choice what to do with the retval.

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,139 @@
...
+    /* For consistency with zend_call_function(), this wrapper also initializes
+     * the retval to IS_UNDEF so that the caller can expect an undefined retval
+     * in the error cases. Past wrappers of zend_call_function() did not mimic
+     * this behavior and that lead to a SIGSEGV.
+     *
+     * https://github.com/php/php-src/blob/PHP-8.0.3/Zend/zend_execute_API.c#L673
+     */
+    ZVAL_UNDEF(retval);
+
+    if (!name || !name_len) return false;
```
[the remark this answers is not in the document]
SammyK: For simple NULL checks like this, I find that going sans curly brackets is more readable, but if there's a strong technical argument for adding the curly brackets, I'm certainly open to changing it. I'm curious if @morrisonlevi has any thoughts on this as well?

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,139 @@
...
+
+    /* Functions cannot be called outside of a request context.
+     * PG(modules_activated) indicates that all of the module RINITs have been
+     * called and we are in a request context.
+     */
+    if (!PG(modules_activated)) return false;
+
+    assert(z_is_lower(name) && "Function names must be lowercase.");
```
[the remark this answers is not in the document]
SammyK: > On the other hand - unless PHP does some magic that I am not aware of - assert would cause the program execution to abort.

This is only true for debug builds of PHP because [`NDEBUG` is defined for non-debug builds of PHP](https://github.com/php/php-src/blob/c40231afbf85e0828a6c56d289cb823d65d03a54/main/php.h#L120-L126). To validate this, we [run the ZAI tests on non-debug builds](https://github.com/DataDog/dd-trace-php/blob/6f1f4b2/.circleci/config.yml#L1083) to ensure there is no abort happening for [these assertions](https://github.com/DataDog/dd-trace-php/blob/6f1f4b2/zend_abstract_interface/methods/tests/php5/methods.cc#L63-L87).

> Since the user is already "forced" to handle failure due to the return false above, do you think it would make sense here to just return false here are well?

This is the behavior on non-debug builds of PHP since the hash table lookup will fail if the function name is not lowercase (and then return `false`).

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,139 @@
...
+    ZSTR_ALLOCA_FREE(zsname, use_heap);
+
+    /* "function %s() does not exist" */
+    if (!func) return false;
+
+    zend_fcall_info_cache fcc = {0};
+    fcc.function_handler = func;
+
+    zend_fcall_info fci = {0};
+    fci.size = sizeof(zend_fcall_info);
```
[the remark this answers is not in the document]
SammyK: I believe this is a relic from an [18 year old refactoring](https://github.com/php/php-src/commit/92b4013e8d6fb3a92f958e9c01abf824bbdb4aaa). Today it is only [accessed in debug builds](https://github.com/php/php-src/blob/dd86987b2cbead2af1e95460cc5cebe03e83f736/Zend/zend_execute_API.c#L709).

I think we could probably remove it from core, but I'm curious what @bwoebi thinks.

zend_abstract_interface/functions/php8/functions.c
morrisonlevi: Sounds good to me; let's do it.

zend_abstract_interface/functions/functions.h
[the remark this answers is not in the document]
morrisonlevi: > [W]e might want to consider a hard-limit on the number of args so that we can stack-allocate the params. 🤔

Allowing the caller to manage the memory so it can stack allocate with exactly the right number without any limits seems like it might be better. I think it may lose some expressiveness, though? Is that why you went this way?

zend_abstract_interface/functions/php8/functions.c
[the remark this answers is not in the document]
morrisonlevi: I am fine with short versions of this pattern, as long as the whole thing fits on a single line:

    if (conditions) return;
    
I'm also fine with `break` and `continue`.

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,139 @@
...
+
+        for (uint32_t i = 0; i < (uint32_t)argc; ++i) {
+            arg = va_arg(argv, zval *);
+            if (!arg) {
+                /* Only dtor the zvals that were successfully copied */
+                fci.param_count = i;
+                zend_fcall_info_args_clear(&fci, /* free_mem */ true);
+                return false;
+            }
+            ZVAL_COPY(&fci.params[i], arg);
```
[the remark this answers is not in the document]
SammyK: > why should we copy the zval to invoke a function instead of refcounting +1?

[It does both](https://github.com/php/php-src/blob/dd86987b2cbead2af1e95460cc5cebe03e83f736/Zend/zend_types.h#L1259-L1262). We need to copy the `zval` because we've allocated heap memory for the `params` on `zend_fcall_info` and we need to copy the `zval`s over to it. But after the above discussion about avoiding the heap allocation altogether, I might refactor this a bit.

zend_abstract_interface/functions/php8/functions.c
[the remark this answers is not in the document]
bwoebi: No, `ZEND_FCI_INITIALIZED()` also accesses it - technically core could change that to a `zend_bool`, but seems sort of pointless to do that arbitrary ABI break. If we really strive to remove it, we'd need to devise some alternative handling for nullable callables in zpp (see `'f'` handling of `zend_parse_arg_impl`). (Which makes sense, but it's not as trivial and extension authors need to adapt to the new handling.)

zend_abstract_interface/functions/functions.h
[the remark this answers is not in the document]
bwoebi: I'm not sure why we need a hard limit for stack allocated arrays? simple variable length array: `zval params[argc];`. (At least I assume that we can use C99 here?)

zend_abstract_interface/functions/php8/functions.c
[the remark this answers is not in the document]
bwoebi: I'm fine with non-block ifs as long as they're strictly single line.

zend_abstract_interface/functions/php8/functions.c
SammyK: > Do you think that a binary return state is enough to communicate the outcome?

It is not.

> Or it would make sense to have at least 3 states:

Very good point. [ZAI error types are on the wishlist](https://github.com/DataDog/dd-trace-php/blob/f417c10/zend_abstract_interface/sandbox/sandbox.h#L7-L12). ([APMPHP-614](https://datadoghq.atlassian.net/browse/APMPHP-614)) They will likely be tightly coupled to the sandbox so that exception and error states can be moved out of the PHP global space and into a locally-managed error state.

The return states would be similar to what you proposed. Off the top of my head, this will include the following basic return states:

0. **Success**
1. **Bailout**; Allows the caller to clean up before bubbling up a `zend_bailout`.
2. **Failure**; This can have a number of generic sub types (e.g. exception thrown, error raised, NULL argument, etc) and API-specific types (e.g. method does not exist).

zend_abstract_interface/functions/tests/php8/functions.cc
SammyK: Thanks! The ZAI SAPI test runner piggybacks off of the original Catch2 work that @morrisonlevi [contributed back in October 2020](https://github.com/DataDog/dd-trace-php/commit/aafae27f64a7475ed77d502f3d06ad0510e40a48). #TeamWork 🙂

zend_abstract_interface/functions/php8/functions.c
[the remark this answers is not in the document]
bwoebi: I'm not sure how that `because` is related (the only thing is, if you do not copy you cannot use zend_fcall_info_args_clear below. But that's not an issue...)? In fact zend_call_function itself already does a copy. We do not need a copy nor RC increase here at all.

zend_abstract_interface/functions/php8/functions.c
SammyK: Added in 4e194c58.

zend_abstract_interface/functions/functions.h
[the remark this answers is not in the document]
SammyK: > I'm not sure why we need a hard limit for stack allocated arrays? simple variable length array: zval params[argc];. (At least I assume that we can use C99 here?)

Indeed we can; we use C99 and C11. Thanks for the tip. I've refactored the param handling in 8ae546bf.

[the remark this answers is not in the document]
morrisonlevi: For what it is worth, C11 moved VLA's to a conditional feature instead of mandatory. The compiler community seems to want to get rid of this feature. I think I still might prefer moving it to the caller, where the size is known.

zend_abstract_interface/functions/php8/functions.c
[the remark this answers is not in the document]
SammyK: > I'm not sure how that because is related (the only thing is, if you do not copy you cannot use zend_fcall_info_args_clear below. But that's not an issue...)?

Indeed, the heap allocation for the params was left over from expanding `zend_fcall_info_argv`, so once I removed that and the `zend_fcall_info_args_clear` calls in 8ae546bf, that simplified the code quite a bit.

**Fun fact:** I originally just expanded `zend_fcall_info_argv` in place just so that I could NULL-check the args, but after this discussion, I think we're onto something even better. :)

> In fact zend_call_function itself already does a copy. We do not need a copy nor RC increase here at all.

I didn't realize that the args were being copied and having the refcount increased from within `zend_call_function`. Thanks for the tip!

I'm not sure how we would take the args from the variable arguments `va_arg` and use them in the `zend_fcall_info` params without [some kind of copy](https://github.com/DataDog/dd-trace-php/blob/8ae546bf5e42a61a3e0e0dbe475a7b1e4644253c/zend_abstract_interface/functions/php8/functions.c#L75). But I may have misunderstood what you meant by not needing to copy?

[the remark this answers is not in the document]
bwoebi: I meant the ZVAL_COPY (vs bare assignment) - so that's fine now :-)

zend_abstract_interface/functions/php8/functions.c
```diff
@@ -0,0 +1,116 @@
...
+     * 'zend_fcall_info' will always be zeroed out in our case.
+     */
+    zval params[argc];
+    if (argc > 0) {
+        va_list argv;
+        va_start(argv, argc);
+        for (uint32_t i = 0; i < (uint32_t)argc; ++i) {
+            zval *arg = va_arg(argv, zval *);
+            if (!arg) return false;
+            params[i] = *arg;
```
bwoebi, line 75: The canonical way to assign a zval would be actually `ZVAL_COPY_VALUE()`. This copies the zend_value as two individual 32 bit assignments on 32 bit environments (I do not know why. Perhaps that's completely irrelevant?!?) and notably does _not_ copy u2.

zend_abstract_interface/functions/functions.h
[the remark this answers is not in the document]
SammyK: > I think I still might prefer moving it to the caller, where the size is known.

Do you see a significant advantage for moving this to the caller vs using a variable length array?

Although this might save us some assignments, my main concern is that the API will not be as expressive which would make it easier pass the params the wrong way. Since one of the main purposes of the ZAI API is to increase stability by reducing the chance of using the API incorrectly, my preference would be to keep it as is and reduce the cognitive load of using the API.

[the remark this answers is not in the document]
SammyK: Although, [UBSan does not like](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/5787/workflows/560dc271-42ab-4d60-8816-d46b0e29260e/jobs/435966/steps) our VLA with zero length (calling a function without args).

```
/home/circleci/datadog/zend_abstract_interface/functions/php8/functions.c:68:10: runtime error: variable length array bound evaluates to non-positive value 0
```

I'll need to fix this tomorrow. But 🎉  for having the `*San`s catch this in CI!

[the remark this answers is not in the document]
labbati: Thanks for the clarification! :)

zend_abstract_interface/functions/php8/functions.c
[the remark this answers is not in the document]
labbati: No need of course :) thanks

zend_abstract_interface/functions/php8/functions.c
labbati: Thanks @SammyK for clarifying

zend_abstract_interface/functions/php8/functions.c
SammyK: TIL! Addressed in 472d6593. Thanks for the insight.

zend_abstract_interface/functions/functions.h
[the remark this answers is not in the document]
SammyK: I've addressed the UBSan errors in e8392df. (Marking this unresolved until hearing thoughts on moving args to the caller from @morrisonlevi.)

zend_abstract_interface/functions/tests/CMakeLists.txt
SammyK: As far as I can tell from the [`add_dependencies` docs](https://cmake.org/cmake/help/latest/command/add_dependencies.html) (I couldn't find `add_dependency`), the API only supports targets to resolve build-order dependencies. I converted the stubs dir to a target with `add_custom_target`, but this did not copy the files:

```
add_custom_target(function_stubs SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/stubs)
add_dependencies(functions function_stubs)
```

I was able to copy the files using [`add_custom_command`](https://cmake.org/cmake/help/latest/command/add_custom_command.html):

```
add_custom_command(TARGET functions
                   POST_BUILD
                   COMMAND ${CMAKE_COMMAND} -E copy_directory
                           ${CMAKE_CURRENT_SOURCE_DIR}/stubs
                           ${CMAKE_CURRENT_BINARY_DIR}/stubs)
```

However, this still will not copy the stub files when they are changed. It will only recopy if you make a change to a source file from the target. Effectively it's the same behavior as `COPY`.

There might be a better solution out there, but I haven't been able to find it.

morrisonlevi: We'll just leave this alone for now, as the proper way eludes us.

zend_abstract_interface/functions/functions.h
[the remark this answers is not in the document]
morrisonlevi: If it weren't for UX and minimizing the space for users to get things wrong, I would prefer caller allocated memory over variable length arrays as compiler vendors seem to really dislike it. Notably, MSVC doesn't support VLAs (or if it does it was only recently changed). I think having an upper-bound on argument length as previously suggested would be a decent compromise, as I think we should not intentionally add any more friction to eventually supporting Windows.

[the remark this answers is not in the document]
SammyK: Thanks for sharing the info. I've updated the args to use a small fixed size array in 5d3566eb.

zend_abstract_interface/functions/functions.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ *   zval_ptr_dtor(&retval);
+ *
+ * Functions cannot be called outside of a request context so this MUST be
+ * called from within a request context (after RINIT and before RSHUTDOWN).
+ */
+bool zai_call_function(const char *name, size_t name_len, zval *retval, int argc, ...);
```
bwoebi, line 25: do we want to add like ~6 lines of macro hell to avoid manually passing the arg count? Looks to me like such an easy pitfall to mess up. The resulting macro would look similarly to https://github.com/php/php-src/blob/d60bc0e2d91d299c6cb03cb48672479bcffd03a7/Zend/zend_ast.h#L227.

SammyK: I like this idea. I'll tackle this on Monday. Thanks for the suggestion!

SammyK: Thanks for helping me figure out the compiler error (missing `retval` 🤦). This was addressed in 0648f8d and f4e4cdc.

morrisonlevi (approved):

PR #1235, opened 2021-05-17
Remove Json encoder and Stream transport

labbati:
### Description

Ahead of our migration to `uint64` IDs support, we need to get rid of all json dependent services in userland, as PHP userland does not support `uint64`.

Specifically, I have identified `DDTrace\Encoders\Json` and `DDTrace\Transport\Stream`.

While this is officially a breaking change, it will have no impact on users using automatic instrumentation.

Even if I believe it is not the case, users either requiring Json encoding to send traces to their own custom targets or relying on the `Stream` class (that was not part of the public api in `src/api`) will have to manually clone the classes from the previous release SHA and register themselves. This is discouraged because as of today, `json` causes a fallback to blocking sync transmission to the Datadog agent, while using messagepack we support the unblocking async transport to our agent.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bridge/_files.php
```diff
@@ -20,7 +20,6 @@
     __DIR__ . '/../src/DDTrace/StartSpanOptionsFactory.php',
     __DIR__ . '/../src/DDTrace/Time.php',
     __DIR__ . '/../src/DDTrace/Transport/Http.php',
-    __DIR__ . '/../src/DDTrace/Transport/Stream.php',
```
SammyK, line 23: We'll also need to remove `Json.php` from the `$autoloaderMapping` array in `bridge/dd_register_optional_deps_autoloader.php`.

SammyK (commented):
I think there's just one more line that needs to be removed and this should be good to go.

bridge/_files.php
labbati: Good catch @SammyK! For some reason it had remained un-pushed! https://github.com/DataDog/dd-trace-php/pull/1235/commits/d0e0706dda121fb84601462302f0cde8a5e5322e

SammyK (approved):
LGTM 👍 Thanks @labbati!

labbati:
Thanks @SammyK

PR #1238, opened 2021-05-19
Change config.m4 to allow in-tree build

bwoebi:
### Description

Putting dd-trace-php directly under php-src/ext should now work.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/php5/signals.c
+++ b/ext/php5/signals.c
@@ -1,6 +1,9 @@
 #include "signals.h"
 
+#ifdef HAVE_CONFIG_H
 #include "config.h"
+#endif
+
 #include "php_config.h"
 
 #if HAVE_SIGACTION
--- a/ext/php8/signals.c
+++ b/ext/php8/signals.c
@@ -1,6 +1,9 @@
 #include "signals.h"
 
+#ifdef HAVE_CONFIG_H
 #include "config.h"
+#endif
+
 #include "php_config.h"
 
 #if HAVE_SIGACTION
--- a/config.m4
+++ b/config.m4
@@ -14,8 +14,9 @@ if test "$PHP_DDTRACE" != "no"; then
     AC_MSG_RESULT([yes])
   ])
 
-  m4_include([m4/polyfill.m4])
-  m4_include([m4/ax_execinfo.m4])
+  define(DDTRACE_BASEDIR, esyscmd(printf %s "$(dirname "__file__")"))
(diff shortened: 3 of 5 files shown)


ext/php8/signals.c
```diff
@@ -1,6 +1,5 @@
 #include "signals.h"
 
-#include "config.h"
```
morrisonlevi, line 3: Doesn't `HAVE_SIGACTION` come from `config.h`?

bwoebi: ah, I need to guard it with HAVE_CONFIG_H, you are right.

ext/php5/signals.c
```diff
@@ -1,6 +1,9 @@
 #include "signals.h"
 
-#include "config.h"
+#ifdef HAVE_CONFIG_H
+# include "config.h"
```
SammyK, line 4: The linter is [failing in CI](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/5796/workflows/51e98ae7-81b8-4562-b2f6-dbeb4533fc94/jobs/438549) on the extra spaces in these files. This particular one has gotten me a few times. We use `clang-format` for our linting/formatting.

SammyK: In my experience, `clang-format` always removes all the whitespace around compiler directives.

I believe our options are:

- Add the [`IndentPPDirectives: AfterHash` rule](https://stackoverflow.com/a/48610210/443479) to `.clang-format` to allow/enforce indenting embedded compiler directives. (This will impact the entire project and require a largish diff so I would love to know @morrisonlevi's thoughts here. But I'm 👍 for this option in a separate PR.)
- Flatten it out to conform to the existing linter rules. (I usually just run `clang-format -i filename.c` before pushing up to GH so that I don't have to think about formatting.)
- Use the `// clang-format {off|on}` comments before and after the lines the linter is complaining about. ([example](https://github.com/DataDog/dd-trace-php/blob/f417c10/zend_abstract_interface/zai_sapi/zai_sapi_functions.c#L82-L88))

bwoebi: I would be in favour as well, I find properly indented preprocessor directives much more readable.

bwoebi: I reformatted it for now, until we reach a decision on the indent.

SammyK (approved):
Wow! @bwoebi you are a PR-generating machine. 🕺 Thanks for fixing the in-tree builds. 👍

PR #1239, opened 2021-05-19
Remove redundant dd_trace_tracer_is_limited() calls

bwoebi:
### Description

`dd_trace_tracer_is_limited` calls are handled internally and can be removed now.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Tracer.php
```diff
@@ -120,7 +120,7 @@ public function __construct(Transport $transport = null, array $propagators = nu
 
     public function limited()
     {
-        return dd_trace_tracer_is_limited();
+        return true;
```
labbati, line 123: Why returning `true` is preferable compared to returning `\dd_trace_tracer_is_limited();`? Since in manual tracing users actually use a tracer instance, how would their manual instrumentation be impacted?

src/api/Contracts/Tracer.php
```diff
@@ -20,6 +20,7 @@ interface Tracer
      * Tracer needs to handle any operation even if its in limited mode,
      * however users can opt not to use tracer when its in limited mode.
      *
+     * @deprecated Has no active functionality anymore and will always return true
```
labbati, line 23: What alternative method we can provide to let customers using the open tracing inspired api to know when they should create or not a span? At the moment, if I am reading the code correctly, the extension-level limited check only applies to the `DDTrace\trace_*` functions.

In case this is correct, we should remove this deprecation notice.

labbati (commented):
Added a couple of questions. Specifically, I am not sure we can remove this functionality from the tracer itself (while it is totally OK to remove it from ALL the instrumentations like you correctly did).

src/DDTrace/Tracer.php
bwoebi: Felt most sensible when looking at the NullTracer, i.e. default?

src/api/Contracts/Tracer.php
bwoebi: Don't know, at least I thought the ticket was clear in that way that we don't want to retain that functionality.

src/DDTrace/Util/CodeTracer.php
```diff
@@ -53,13 +53,6 @@ public function tracePublicMethod(
                 $isTraceAnalyticsCandidate
             ) {
                 $tracer = GlobalTracer::get();
-                if ($tracer->limited()) {
-                    if ($limitedTracerCallHook) {
-                        $limitedTracerCallHook(func_get_args());
-                    }
-
-                    return dd_trace_forward_call();
-                }
```
SammyK, line 62: `CodeTracer` is dead code and references the legacy `dd_trace()` API which is NOOP. We can remove this file and also remove references to it from `/phpstan.ddtrace.neon` and `bridge/_files.php`.

src/DDTrace/Tracer.php
SammyK: We just need to remove the `dd_trace_tracer_is_limited()` calls from the integrations, but this is part of the userland instrumentation API so we should keep the call here.

Even though limited checks are being checked internally for the sandbox API (which is what our integrations use), customers who have custom instrumentations using the userland `DDTrace\Tracer` API will still need the option of doing a limited check using `Tracer::limited()`.

src/api/Contracts/Tracer.php
SammyK: To clarify from the card: we call closures passed to the [sandbox API](https://docs.datadoghq.com/tracing/setup_overview/custom_instrumentation/php/?tab=tracingfunctioncalls) "tracing closures". The integrations instrument function calls via tracing closures in the sandbox API. These are where the redundant limited checks were occurring.

SammyK (commented):
Thanks @bwoebi! I just added a few clarifications from the card.

src/api/Contracts/Tracer.php
bwoebi: So what exactly shall I do here?

SammyK: We should remove the `@deprecated` tag since we're keeping the limited check in `Tracer::limited()`.

src/DDTrace/Util/CodeTracer.php
bwoebi: done (and also the other suggestions)

SammyK (approved):
LGTM 👍

PR #1241, opened 2021-05-20
Remove TSRMLS_* macros from PHP 7 and PHP 8

bwoebi:
### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/php7/circuit_breaker.c
+++ b/ext/php7/circuit_breaker.c
@@ -69,28 +69,28 @@ static uint64_t current_timestamp_monotonic_usec() {
     return t.tv_sec * 1000 * 1000 + t.tv_nsec / 1000;
 }
 
-static int64_t get_max_consecutive_failures(TSRMLS_D) {
+static int64_t get_max_consecutive_failures(void) {
     return ddtrace_get_int_config(DD_TRACE_CIRCUIT_BREAKER_ENV_MAX_CONSECUTIVE_FAILURES,
-                                  DD_TRACE_CIRCUIT_BREAKER_DEFAULT_MAX_CONSECUTIVE_FAILURES TSRMLS_CC);
+                                  DD_TRACE_CIRCUIT_BREAKER_DEFAULT_MAX_CONSECUTIVE_FAILURES);
 }
 
-static int64_t get_retry_time_usec(TSRMLS_D) {
+static int64_t get_retry_time_usec(void) {
     return ddtrace_get_int_config(DD_TRACE_CIRCUIT_BREAKER_ENV_RETRY_TIME_MSEC,
-                                  DD_TRACE_CIRCUIT_BREAKER_DEFAULT_RETRY_TIME_MSEC TSRMLS_CC) *
+                                  DD_TRACE_CIRCUIT_BREAKER_DEFAULT_RETRY_TIME_MSEC) *
            1000;
 }
 
-uint32_t dd_tracer_circuit_breaker_can_try(TSRMLS_D) {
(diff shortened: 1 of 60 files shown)


ext/php7/logging.h
```diff
@@ -4,10 +4,7 @@
 
 #include "configuration.h"
 
-inline void ddtrace_log_err(const char *message) {
-    TSRMLS_FETCH();
-    php_log_err((char *)message TSRMLS_CC);
-}
+inline void ddtrace_log_err(const char *message) { php_log_err((char *)message); }
```
bwoebi, line 7: wtf is clang-format converting this to a single line?!?

morrisonlevi: Yes. It bothered me at first too, but now it's my preference in all languages! Trivial things taking up trivial amounts of space on my screen is nice.

ext/php7/circuit_breaker.c
```diff
@@ -69,28 +69,28 @@ static uint64_t current_timestamp_monotonic_usec() {
     return t.tv_sec * 1000 * 1000 + t.tv_nsec / 1000;
 }
 
-static int64_t get_max_consecutive_failures(TSRMLS_D) {
+static int64_t get_max_consecutive_failures() {
```
SammyK, line 72: Should we replace `TSRMLS_D` with `void`?

ext/vendor/mpack/mpack.c
```diff
@@ -1,26 +1,26 @@
 /**
```
SammyK, line 1: We leave sources in the `ext/vendor` dir as-is so we don't need to `clang-format` any of the sources in this dir.

bwoebi: Uhm, I just did `make clang_format_fix` (i.e. `make clang_find_files_to_lint` returned these files) ... shall I then exclude them?

ext/php7/circuit_breaker.c
bwoebi: I have no opinion on this. Can do that if you like.

ext/vendor/mpack/mpack.c
SammyK: Did you run that locally on your Mac? That target is supposed to [ignore these paths](https://github.com/DataDog/dd-trace-php/blob/281bc42/Makefile#L147-L152) but I think there are some incompatibilities with `find` on Mac. (At least I vaguely remember running into a similar issue where a `find` command would return different results from the containers.)

ext/php7/circuit_breaker.c
SammyK: I think I would prefer having these be `void` since it is technically an API change to not have it.

ext/vendor/mpack/mpack.c
bwoebi: Yes, I did. Perhaps I should fix that then to not have it happen again accidentally.

ext/php7/circuit_breaker.c
bwoebi: Will do then.

ext/vendor/mpack/mpack.c
bwoebi: Fixed the find expression to return the same files (different order, but that doesn't matter :-D) and re-applied.

SammyK: Thanks for fixing that!

SammyK (approved):
Perfect. Thanks so much for tackling this @bwoebi! It feels so nice to have these finally removed from the PHP 7 and PHP 8 dirs. 💯

PR #1246, opened 2021-05-27
Create property ZAI seam for PHP 8

bwoebi:
### Description

A simple function allowing direct access into userland (and internal classes without very special property storage) properties, bypassing __get() and similar.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -56,6 +56,9 @@
             <file name="zend_abstract_interface/sandbox/php7/sandbox.c" role="src" />
             <file name="zend_abstract_interface/sandbox/php8/sandbox.c" role="src" />
             <file name="zend_abstract_interface/sandbox/sandbox.h" role="src" />
+            <file name="zend_abstract_interface/properties/php7-8/properties.c" role="src" />
+            <file name="zend_abstract_interface/properties/php7-8/properties.h" role="src" />
+            <file name="zend_abstract_interface/properties/properties.h" role="src" />
             <file name="zend_abstract_interface/zai_assert/zai_assert.h" role="src" />
             <file name="zend_abstract_interface/zai_sapi/php5/zai_sapi.c" role="src" />
             <file name="zend_abstract_interface/zai_sapi/php7/zai_sapi.c" role="src" />
--- a/zend_abstract_interface/properties/CMakeLists.txt
+++ b/zend_abstract_interface/properties/CMakeLists.txt
@@ -1,0 +1,33 @@
+if(PHP_VERSION_DIRECTORY STREQUAL "php7" OR PHP_VERSION_DIRECTORY STREQUAL "php8")
+    set(PHP_PROPERTIES_VERSION_DIRECTORY "php7-8")
+else()
+    set(PHP_PROPERTIES_VERSION_DIRECTORY PHP_VERSION_DIRECTORY)
+endif()
+
+add_library(zai_properties "${PHP_PROPERTIES_VERSION_DIRECTORY}/properties.c")
+
+target_include_directories(zai_properties PUBLIC
+                                          $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+                                          $<INSTALL_INTERFACE:include>)
+
+target_compile_features(zai_properties PUBLIC c_std_99)
+
+target_link_libraries(zai_properties PUBLIC "${PHP_LIB}")
+
+set_target_properties(zai_properties PROPERTIES
+                                     EXPORT_NAME properties
+                                     VERSION ${PROJECT_VERSION})
+
+add_library(Zai::Properties ALIAS zai_properties)
+
+if (${BUILD_ZAI_TESTING})
+  add_subdirectory(tests)
+endif()
+
+# This copies the include files when `install` is ran
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/properties.h
+        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/properties/)
+
+target_link_libraries(zai_zend_abstract_interface INTERFACE zai_properties)
+
+install(TARGETS zai_properties EXPORT ZendAbstractInterfaceTargets)
--- a/zend_abstract_interface/properties/php7-8/properties.c
+++ b/zend_abstract_interface/properties/php7-8/properties.c
@@ -1,0 +1,51 @@
+#include "properties.h"
+
+zval *zai_read_property_direct(zend_class_entry *ce, zend_object *object, zend_string *name) {
+    if (!ce) {
+        return &EG(error_zval);
+    }
+    if (!object) {
+        return &EG(error_zval);
+    }
+    if (!name) {
+        return &EG(error_zval);
+    }
+    if (!instanceof_function(object->ce, ce)) {
+        return &EG(error_zval);
+    }
+
+    zend_class_entry *prev_fake_scope = EG(fake_scope);
+    EG(fake_scope) = ce;
+    zend_property_info *prop_info = zend_get_property_info(object->ce, name, 1);
+    EG(fake_scope) = prev_fake_scope;
+
+    if (UNEXPECTED(prop_info == NULL)) {
+        if (object->properties == NULL) {
+            return &EG(uninitialized_zval);
+        }
+        zval *zv = zend_hash_find(object->properties, name);
+        if (zv == NULL) {
+            return &EG(uninitialized_zval);
+        }
+        return zv;
+    }
+
+    if (UNEXPECTED(prop_info == ZEND_WRONG_PROPERTY_INFO)) {
+        return &EG(error_zval);
+    }
+
+    return OBJ_PROP(object, prop_info->offset);
+}
+
+#undef zai_read_property_direct_literal
+zval *zai_read_property_direct_literal(zend_class_entry *ce, zend_object *obj, const char *name, int name_len) {
+    zend_string *name_str;
+    ALLOCA_FLAG(use_heap);
+    ZSTR_ALLOCA_INIT(name_str, name, name_len, use_heap);
+
+    zval *val = zai_read_property_direct(ce, obj, name_str);
+
+    ZSTR_ALLOCA_FREE(name_str, use_heap);
+
+    return val;
+}
--- a/zend_abstract_interface/properties/php7-8/properties.c
+++ b/zend_abstract_interface/properties/php7-8/properties.c
@@ -37,8 +37,7 @@ zval *zai_read_property_direct(zend_class_entry *ce, zend_object *object, zend_s
     return OBJ_PROP(object, prop_info->offset);
 }
 
-#undef zai_read_property_direct_literal
-zval *zai_read_property_direct_literal(zend_class_entry *ce, zend_object *obj, const char *name, int name_len) {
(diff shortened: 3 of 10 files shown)


zend_abstract_interface/properties/tests/php7-8/properties.cc
```diff
@@ -0,0 +1,134 @@
...
+        REQUIRE(zai_sapi_spinup()); \
+        ZAI_SAPI_TSRMLS_FETCH(); \
+        ZAI_SAPI_ABORT_ON_BAILOUT_OPEN() \
+        REQUIRE(zai_sapi_execute_script("./stubs/classes.php")); \
+        { code } \
+        ZAI_SAPI_ABORT_ON_BAILOUT_CLOSE() \
+        zai_sapi_spindown(); \
+    }
+
+static inline zval *zai_read_property_direct_literal(zend_class_entry *ce, zend_object *obj, const char *name, int name_len) {
```
morrisonlevi, line 20: Nit: in C99 and newer there is no need for `inline` on a function in a source file, so our style is to omit it.

zend_abstract_interface/properties/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+    set(PHP_PROPERTIES_VERSION_DIRECTORY PHP_VERSION_DIRECTORY)
+endif()
+
+add_library(zai_properties "${PHP_PROPERTIES_VERSION_DIRECTORY}/properties.c")
+
+target_include_directories(zai_properties PUBLIC
+                                          $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+                                          $<INSTALL_INTERFACE:include>)
+
+target_compile_features(zai_properties PUBLIC c_std_99)
```
morrisonlevi, line 13: I was expecting this to be either `INTERFACE c_std_99 PRIVATE c_std_11` or `PUBLIC c_std_11` -- any specific reason for choosing C99?

zend_abstract_interface/properties/php7-8/properties.c
```diff
@@ -0,0 +1,30 @@
+#include "properties.h"
+
+#include <assert.h>
+#include <sandbox/sandbox.h>
```
SammyK, line 4: I don't see any sandbox symbols being used. Was this left over from a refactor?

zend_abstract_interface/properties/CMakeLists.txt
bwoebi: just copy paste from the other CMakeLists. Doesn't really matter as long as it compiles?

package.xml
```diff
@@ -56,6 +56,9 @@
             <file name="zend_abstract_interface/sandbox/php7/sandbox.c" role="src" />
             <file name="zend_abstract_interface/sandbox/php8/sandbox.c" role="src" />
             <file name="zend_abstract_interface/sandbox/sandbox.h" role="src" />
+            <file name="zend_abstract_interface/properties/php7-8/properties.c" role="src" />
```
SammyK, line 59: We'll also need to add `zend_abstract_interface/properties/php7-8/properties.c` to `ZAI_SOURCES` for PHP 8 in `config.m4`. And also the new build dirs:

```
PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/properties])
PHP_ADD_BUILD_DIR([$ext_builddir/zend_abstract_interface/properties/php7-8])
```

zend_abstract_interface/properties/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+                                     VERSION ${PROJECT_VERSION})
+
+add_library(Zai::Properties ALIAS zai_properties)
+
+if (${BUILD_ZAI_TESTING})
+  add_subdirectory(tests)
+endif()
+
+# This copies the include files when `install` is ran
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/properties.h ${CMAKE_CURRENT_SOURCE_DIR}/properties_common.h
```
SammyK, line 28: I'm guessing `properties_common.h` was left over from a refactor?

zend_abstract_interface/properties/tests/php7-8/properties.cc
```diff
@@ -0,0 +1,134 @@
...
+
+#define TEST(name, code) TEST_CASE(name, "[zai properties]") { \
+        REQUIRE(zai_sapi_spinup()); \
+        ZAI_SAPI_TSRMLS_FETCH(); \
+        ZAI_SAPI_ABORT_ON_BAILOUT_OPEN() \
+        REQUIRE(zai_sapi_execute_script("./stubs/classes.php")); \
+        { code } \
+        ZAI_SAPI_ABORT_ON_BAILOUT_CLOSE() \
+        zai_sapi_spindown(); \
+    }
```
SammyK, line 18: Cool macro! 😎

zend_abstract_interface/properties/tests/php7-8/properties.cc
```diff
@@ -0,0 +1,134 @@
...
+
+static inline zval *zai_read_property_direct_literal(zend_class_entry *ce, zend_object *obj, const char *name, int name_len) {
+    zend_string *name_str = zend_string_init(name, name_len, 0);
+    zval *val = zai_read_property_direct(ce, obj, name_str);
+    zend_string_release(name_str);
+    return val;
+}
+#define zai_read_property_direct_literal(obj, ce, name) zai_read_property_direct_literal(obj, ce, ZEND_STRL(name))
```
SammyK, line 26: I think this would be useful to make public in `properties.h` so that that caller doesn't have to worry about making `zend_strings`. I believe most (or all?) of the time we will use this API, it will be with a string literal. WDYT?

zend_abstract_interface/properties/tests/php7-8/properties.cc
```diff
@@ -0,0 +1,134 @@
...
+        REQUIRE(zai_sapi_execute_script("./stubs/classes.php")); \
+        { code } \
+        ZAI_SAPI_ABORT_ON_BAILOUT_CLOSE() \
+        zai_sapi_spindown(); \
+    }
+
+static inline zval *zai_read_property_direct_literal(zend_class_entry *ce, zend_object *obj, const char *name, int name_len) {
+    zend_string *name_str = zend_string_init(name, name_len, 0);
+    zval *val = zai_read_property_direct(ce, obj, name_str);
+    zend_string_release(name_str);
```
SammyK, line 23: Do you think it's worth using the `ZSTR_ALLOCA_ALLOC` API here?

zend_abstract_interface/properties/tests/php7-8/properties.cc
```diff
@@ -0,0 +1,134 @@
...
+
+TEST("access property with invalid name is error", {
+    zval child;
+    zai_call_function_literal("zai\\properties\\test\\child", &child);
+
+    zval *val = zai_read_property_direct_literal(Z_OBJCE(child), Z_OBJ(child), "\0with zero byte");
+    REQUIRE(val == &EG(error_zval));
+
+    zval_ptr_dtor(&child);
+})
```
SammyK, line 134: In the spirit of being super defensive, could we add some tests where `ce`, `object`, and `name` args are `NULL`?

Also, do you think it's worth it to add some tests to access properties on instances of internal classes as well?

zend_abstract_interface/properties/tests/php7-8/properties.cc
bwoebi: Well, our _current_ uses do not have it as string literal, but always zend_string *. I don't mind adding it publicly though (at least for future use).

package.xml
bwoebi: I added that in a separate (yet to be committed branch), but will backport it here, makes sense…

zend_abstract_interface/properties/CMakeLists.txt
bwoebi: Yes, surprised that cmake doesn't complain about it missing.

zend_abstract_interface/properties/tests/php7-8/properties.cc
bwoebi: yes, as we move it to public API, yes.

zend_abstract_interface/properties/tests/php7-8/properties.cc
morrisonlevi: I agree with ensuring the caller does this check in principle, but until it can be enforced by static analysis or something, I would prefer to be defensive. For you and I getting a sigsegv directly at the site is nice, but for customers and their customers it's not good at all, as crashes bring down the process.

zend_abstract_interface/properties/tests/php7-8/properties.cc
bwoebi: (no, I mean, our current uses are actual zend_string * like `ZSTR_MESSAGE` - but will do anyway)

zend_abstract_interface/properties/tests/php7-8/properties.cc
bwoebi: Actually, we probably can just log that instead? I just have no idea what the proper logging within ZAI would be for that? ZAI has no access to the standard ddtrace_log functions

zend_abstract_interface/properties/tests/php7-8/properties.cc
SammyK: Ah yes, gotcha. Thanks for clarifying. Since the known-string API wasn't added until PHP 7.2, I think using a string literal in this case would be a good way to keep the API consistent across all PHP versions. It'll be great to not need so many preprocessor directives in the tracer code. :)

zend_abstract_interface/properties/tests/php7-8/properties.cc
bwoebi: Re internal class access: there is no difference in the zend_object * structure whether something is internal or not. It just might be that internal classes have some data attached and may define their own read_properties, but this function does not support that use case anyway.

zend_abstract_interface/properties/php7-8/properties.h
```diff
@@ -0,0 +1,12 @@
+#ifndef ZAI_PROPERTIES_H
+#define ZAI_PROPERTIES_H
+
+#include "Zend/zend_API.h"
+
+// Note that the function will completely bypass read_property. ce (scope) must be object ce or parent of object ce.
+zval *zai_read_property_direct(zend_class_entry *ce, zend_object *object, zend_string *name);
+
+zval *zai_read_property_direct_literal(zend_class_entry *ce, zend_object *obj, const char *name, int name_len);
+#define zai_read_property_direct_literal(obj, ce, name) zai_read_property_direct_literal(obj, ce, ZEND_STRL(name))
```
morrisonlevi, line 10: What do you think about naming the function `zai_read_property_direct_cstr` and leaving the macro as `zai_read_property_direct_literal`? This way if I have a ptr + len pair already for some reason I can call the function directly without undef'ing the macro.

bwoebi: I agree, I was already doubting that decision myself :-D

SammyK (approved):
Nice work @bwoebi! LGTM 👍

PR #1248, opened 2021-05-28
Create exceptions ZAI seam for PHP 8 

bwoebi:
### Description

Using the properties seam within an exceptions seam, and use that in the php 8 source.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/zend_abstract_interface/exceptions/exceptions_common.h
+++ b/zend_abstract_interface/exceptions/exceptions_common.h
@@ -1,0 +1,7 @@
+#include <main/php.h>
+
+#if PHP_VERSION_ID < 70100
+
+#define ZEND_STR_MESSAGE "message"
+
+#endif
--- a/zend_abstract_interface/exceptions/php8/exceptions.c
+++ b/zend_abstract_interface/exceptions/php8/exceptions.c
@@ -1,0 +1,111 @@
+#include "exceptions.h"
+
+#include <Zend/zend_smart_str.h>
+
+zend_string *zai_exception_message(zend_object *ex) {
+    if (!ex) {
+        // should never happen; TODO: fail in CI
+        return zend_string_init_interned(ZEND_STRL("(internal error retrieving exception for message)"), 1);
+    }
+
+    zval *message = ZAI_EXCEPTION_PROPERTY(ex, ZEND_STR_MESSAGE);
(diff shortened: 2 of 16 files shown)


zend_abstract_interface/exceptions/php8/exceptions.c
```diff
@@ -0,0 +1,99 @@
...
+
+        zend_array *ht = Z_ARRVAL_P(frame);
+
+        zval *file = zend_hash_find_ex(ht, ZSTR_KNOWN(ZEND_STR_FILE), 1);
+        if (file) {
+            if (Z_TYPE_P(file) != IS_STRING) {
+                /* This is not the function, but the file. However, this is what
+                 * PHP itself reports, so we'll match it for consistency.
+                 */
+                smart_str_appends(&str, "[unknown function]");
```
morrisonlevi, line 39: In 8.0.7 this is fixed, btw: https://github.com/php/php-src/commit/6fd13d0afd8f487713fa068aec9fc83358319a48.

bwoebi: Good point - I will change that :-)

SammyK (commented):
Once the conflicts are resolved, this LGTM. 👍 Nicely done @bwoebi.

zend_abstract_interface/exceptions/exceptions_common.h
```diff
@@ -0,0 +1,7 @@
+#include <main/php.h>
+
+#if PHP_VERSION_ID < 70100
+
+#define ZEND_STR_MESSAGE "message"
```
SammyK, line 5: Now that I think about this a bit more, do you think it would make sense to remove `exceptions_common.h` since each major PHP version will have its own `php<n>/exceptions.h` header?

SammyK (approved):
I left one question that's not a blocker for merge if you would rather keep it as-is. Thanks @bwoebi! 👍

zend_abstract_interface/exceptions/exceptions_common.h
bwoebi: I'm not sure yet. I think we'll see what makes most sense once we actually implement the php 5 and 7 seams. I think it should be changed (or not) at the point of implementing these.

PR #1250, opened 2021-06-02
ZAI env

SammyK:
### Description

WIP

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi:
Can we have a trivial macro with signature `(name, buf)` expanding to `name, sizeof(name), buf, sizeof(buf)`? Because that most probably will be the by far most common use case here, I suspect.

zend_abstract_interface/env/env.h
```diff
@@ -0,0 +1,35 @@
...
+ * 'name'. Must be called within a request context. If the active SAPI has a
+ * custom environment variable handler, the SAPI handler is used to access the
+ * environment variable. If there is no custom handler, the environment variable
+ * is accessed from the host using getenv().
+ *
+ * For error conditions, a return value other than ZAI_ENV_SUCCESS is returned
+ * and 'buf' is made an empty string. If the buffer size 'buf_size' is not big
+ * enough to contain the value, ZAI_ENV_BUFFER_TOO_SMALL will be returned and
+ * 'buf' will be an empty string; e.g. this API does not attempt to truncate
+ * the value to accommodate the buffer size.
```
bwoebi, line 31: Regarding the buffer size, should we maybe define a default buffer size (`#define ZAI_ENV_MAX_BUFSIZ`) to 32768? which we can then use for our "unbounded" use cases, like the array map etc. (at least on linux getenv() will never return a string longer than ~32760 currently, so that seems to me like a sane upper limit).

Then usages of zai_getenv() will do `char envbuf[ZAI_ENV_MAX_BUFSIZ]` - and when we ever notice we need bigger buffers, we can increase that centrally here.

morrisonlevi (commented):
Recently I started checking what the Rust std libraries do for things because overall they are well designed. We have an additional need (env var is not being fetched in a PHP request), but the rest should be about the same. For Rust, the equivalent function is [`std::env:var`](https://doc.rust-lang.org/std/env/fn.var.html) which returns `Result<String, VarError>`:

```rust
pub enum VarError {
    NotPresent,
    NotUnicode(OsString),
}
```

That `String` it returns is owned, so it would be copied for the user automatically.

-----

Overall, the design for ZAI env seems to be good. It takes in the additional factors of avoiding memory allocation and the env not being ready. I'm not sure that verifying it is unicode is worth it at this time, so overall good work!

I think the only thing which bothers me is the length + pointers being passed independently. These things are connected, so we should make structs for them. For the input string, we could use a `string_view` (or a zai copy of it?) because it's read-only and only temporarily borrowed. I'm not sure what to use for the output buffer; my inclination would be a byteslice (which is what I have for my WIP C->Rust exporter for the profiler):

```c
typedef struct zai_byteslice_s {
  size_t len;
  uint8_t *ptr;
} zai_byteslice;
```

Our revised signature might look like this:

```c
zai_env_result zai_getenv(zai_stringview name, zai_byteslice output);
```

What do you think?

Also, we may want to figure out the portability of the attribute `warn_unused_result`; whenever we return some sort of enum it should always be checked, I think?

bwoebi:
@Levi I also wondered about such a thing, but in the end you mostly will have two extra lines of code declaring these structs on the caller side before doing the call. While obviously a cleaner interface, the ergonomics of doing a trivial zai call suffer here.
Unless we were to actually pass arounds these stringview/byteslice structs in the ddtrace code, and not only on the ddtrace <-> zai boundary, the benefits of this aren't outweighing the costs.
With Rust however stringviews are ubiquitous and should be strongly perferred, because there they also have first-class support of declaring them inline and are also going to be passed around. On the C side that doesn't really apply.

Agree on the suggestion of warn_unused_result. Thanks for bringing that up.

zend_abstract_interface/env/env.h
SammyK: Excellent suggestion. I'll implement this.

zend_abstract_interface/env/env.c
```diff
@@ -0,0 +1,56 @@
...
+
+    buf[0] = '\0';
+
+    if (!name || !name_len) return ZAI_ENV_ERROR;
+
+    /* Some SAPIs do not initialize the SAPI-controlled environment variables
+     * until the request has started. It is for this reason we cannot reliably
+     * access environment variables outside of a request context.
+     */
+    if (!PG(modules_activated)) return ZAI_ENV_NOT_READY;
```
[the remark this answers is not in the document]
SammyK: Wow! Very nice catch. The very use-case for this API (first-time RINIT) won't even work as it is designed. 🤦 I'll make sure to change this and add a test that executes during RINIT.

morrisonlevi:
>  While obviously a cleaner interface, the ergonomics of doing a trivial zai call suffer here. Unless we were to actually pass arounds these stringview/byteslice structs in the ddtrace code, and not only on the ddtrace <-> zai boundary, the benefits of this aren't outweighing the costs.

The stringview will add trivial typing -- that part usually won't require an additional line because you do something like `zai_getenv(SV("DD_AGENT_HOST"), ...)`. You might be right about the byteslice, but the caller will use the output string in some way so I think it washes out?

> With Rust however stringviews are ubiquitous and should be strongly perferred, because there they also have first-class support of declaring them inline and are also going to be passed around. On the C side that doesn't really apply.

If I implement a C API in Rust, I still want the stringview/byteslice in the API because then I can write methods on the byteslice struct like `into_str` which does the UTF-8 validation and converts it into Rust `&str`. Of course, it can also be done as a freestanding function so it is a bit of taste.

-----

Overall, I don't feel that strongly on it for this specific function because it's still a small API, but if it took 3+ pointers and 3+ lengths I would feel stronger about it -- it gets hard to correctly use functions with many parameters, and pairing up each ptr+length helps. I'd like to hear what others think.

zend_abstract_interface/env/env.c
```diff
@@ -0,0 +1,56 @@
...
+
+    /* If 'sapi_module.getenv' is not set, sapi_getenv() will return NULL; but a
+     * NULL return value could also mean that the target environment variable
+     * does not exist. To distinguish these two paths we need to check
+     * 'sapi_module.getenv' here and fall back to getenv() only if the SAPI does
+     * not have special environment variable handling.
+     */
+    bool use_sapi_env = (bool)sapi_module.getenv;
```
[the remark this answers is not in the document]
SammyK: Yep! This is just a glimpse of how strange my mind works sometimes. 😆

zend_abstract_interface/env/env.h
morrisonlevi: > Then usages of zai_getenv() will do char envbuf[ZAI_ENV_MAX_BUFSIZ] - and when we ever notice we need bigger buffers, we can increase that centrally here.

I don't think it's smart to increase that number if we're putting it on the stack. For what it's worth, PHP's `ALLOCA_FLAG` does [32KiB stack allocations](https://heap.space/xref/php-src/Zend/zend_portability.h?r=c276c16b#349) and anything larger will go to the heap. What do you think?

bwoebi: While that's right, at that point, we can grep the usages of this `ZAI_ENV_MAX_BUFSIZ` and replace the usages by a DO_ALLOCA(). But that currently should not be needed.

SammyK:
Thank you both for the fantastic feedback. To comment on the API suggestions; macro vs struct wrappers.

I do like the idea of a macro since that will clean up the calls quite a bit:

```c
#define zai_getenv_literal(name, buf) zai_getenv(name, sizeof(name) - 1, buf, sizeof buf)
```

My main concern is when the buffer isn't owned by the caller. For ZAI config, it could be possible that `zai_getenv()` will be called via a proxy function and one might accidentally misuse the API:

```c
static void get_foo(char *buf, size_t buf_size) {
    zai_env_result res = zai_getenv_literal("FOO", buf);
    // ...
}
```

I also like the idea of struct wrappers, but it does make the API quite a bit more verbose.

```c
char buf_storage[64];
zai_byteslice buf = {
    sizeof buf_storage,
    (uint8_t *)buf_storage
};
zai_env_result res = zai_getenv(ZAI_STRL_VIEW("FOO"), buf);
```

But this method makes it safe to pass around the buffer and call via proxy.

```c
static void get_foo(zai_byteslice buf) {
    zai_env_result res = zai_getenv(ZAI_STRL_VIEW("FOO"), buf);
    // ...
}
```

So what if we combined both ideas and still used a macro API with struct wrappers under the hood?

```c
#define ZAI_BYTESLICE_INIT(name, size) \
    char name##_storage[size]; \
    zai_byteslice name = { \
        size, \
        (uint8_t *)name##_storage \
    }

ZAI_BYTESLICE_INIT(buf, 64);
zai_env_result res = zai_getenv(ZAI_STRL_VIEW("FOO"), buf);
```

We can add a macro for string literals too if y'all think it's worth it.

```c
#define zai_getenv_literal(name, buf) zai_getenv(ZAI_STRL_VIEW(name), buf)

ZAI_BYTESLICE_INIT(buf, 64);
zai_env_result res = zai_getenv_literal("FOO", buf);
```

What do y'all think?

zend_abstract_interface/env/env.c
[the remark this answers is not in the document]
SammyK: I've applied all of the suggestions except this one since it requires an update to the ZAI SAPI API that will allow us to add an "inline module" to test that we can access the environment variables in module RINIT. I started on the API, but I'm going to wrap it up tomorrow morning.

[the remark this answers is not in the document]
SammyK: This has been addressed in 20b1eb3. Thanks again for catching this one!

zend_abstract_interface/env/env.h
```diff
@@ -0,0 +1,74 @@
...
+    zai_env_buffer name = {size, name##_storage}
+
+// TODO Move this to Zai::String
+typedef struct zai_string_view_s {
+    size_t len;
+    const char *ptr;
+} zai_string_view;
+
+#define ZAI_STRL_VIEW(cstr) \
+    { sizeof(cstr) - 1, cstr }
```
bwoebi, line 55: Note that this is a C++ only feature, in C you need to put the struct name before, like `(zai_string_view){ sizeof(cstr) - 1, cstr }`

SammyK: TIL. Thanks for catching this! Fixed in a2b254a.

bwoebi (approved):
Looks ready for merge now :-)

SammyK:
Thanks @bwoebi.

PR #1251, opened 2021-06-04
Moving spans from userland to extension

bwoebi:
### Description

Doing a lot of things:
 - Nearly everything happening in Tracer->flush() is now in the extension
 - Spans are allocated and closed directly instead of their ids
 - Spans now expose duration and start time via methods
 - The root span is (if there is configured to be one) allocated and closed from within the extension - close at the very last moment: RSHUTDOWN

Also note: DRAFT, UNTESTED AND BROKEN.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php8/ddtrace.c
+++ b/ext/php8/ddtrace.c
@@ -87,9 +87,6 @@ static void ddtrace_shutdown(struct _zend_extension *extension) {
 static void ddtrace_activate(void) {}
 static void ddtrace_deactivate(void) {}
 
-// prepare the tracer state to start handling a new trace
-static void dd_prepare_for_new_trace(void);
-
 static zend_extension _dd_zend_extension_entry = {"ddtrace",
                                                   PHP_DDTRACE_VERSION,
                                                   "Datadog",
@@ -153,6 +150,15 @@ ZEND_BEGIN_ARG_INFO_EX(arginfo_dd_trace_push_span_id, 0, 0, 0)
 ZEND_ARG_INFO(0, existing_id)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(arginfo_dd_trace_close_span, 0, 0, 0)
+ZEND_ARG_INFO(0, finish_time)
+ZEND_END_ARG_INFO()
+
+ZEND_BEGIN_ARG_INFO_EX(arginfo_ddtrace_add_global_tag, 0, 0, 2)
+ZEND_ARG_INFO(0, key)
+ZEND_ARG_INFO(0, value)
+ZEND_END_ARG_INFO()
+
 ZEND_BEGIN_ARG_INFO_EX(arginfo_dd_trace_internal_fn, 0, 0, 1)
 ZEND_ARG_INFO(0, function_name)
 ZEND_ARG_VARIADIC_INFO(0, vars)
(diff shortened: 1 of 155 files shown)


ext/php8/ddtrace.c
```diff
@@ -748,6 +808,22 @@ static PHP_FUNCTION(additional_trace_meta) {
     zval_copy_ctor(return_value);
 }
 
+/* {{{ proto string DDTrace\add_global_tag(string $key, string $value) */
+static PHP_FUNCTION(add_global_tag) {
+    UNUSED(execute_data);
+
+    zend_string *key, *val;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "SS", &key, &val) == FAILURE) {
+        RETURN_THROWS();
```
labbati, line 817: We should not throw. We can log a message in debug mode and proceed the execution.
In general, we never throw exceptions, so users' applications can keep working even if with tracing not perfectly working

ext/php8/ddtrace.c
```diff
@@ -1249,6 +1325,50 @@ static PHP_FUNCTION(dd_trace_peek_span_id) {
...
+    double finish_time_seconds = 0;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "|f", &finish_time_seconds) != SUCCESS) {
+        ddtrace_log_debug("unexpected parameter. the environment variable name must be provided");
+        RETURN_FALSE;
+    }
+
+    // TODO: also prevent closing internal root spans? Would need marker though
+    if (!DDTRACE_G(open_spans_top) || !DDTRACE_G(open_spans_top)->execute_data) {
+        zend_throw_exception(NULL, "There is no user-span on the top of the stack. Cannot close.", 0);
+        RETURN_THROWS();
```
labbati, line 1357: Same as above, let's not throw. The maximum that we can accept (but we should avoid it) is that tracing is not working as expected, but users' requests never have to be interrupted because of us.

ext/php8/serializer.c
```diff
@@ -426,6 +426,36 @@ static void _serialize_meta(zval *el, ddtrace_span_fci *span_fci) {
...
+    ZEND_HASH_FOREACH_STR_KEY_VAL(global_tags, global_key, global_val) {
+        zend_hash_add(Z_ARR_P(meta), global_key, global_val);
+    }
+    ZEND_HASH_FOREACH_END();
+    zend_hash_release(global_tags);
+
+    zend_string *tag_key;
+    zval *tag_value;
+    ZEND_HASH_FOREACH_STR_KEY_VAL(DDTRACE_G(additional_global_tags), tag_key, tag_value) {
+        zend_hash_add(DDTRACE_G(additional_global_tags), tag_key, tag_value);
```
labbati, line 455: Not sure I understand what this line does. I think that here we should add from `DDTRACE_G(additional_global_tags)` into `Z_ARR_P(meta)`?

If this holds true, then we are missing a test for global + additional globals?

src/DDTrace/Bootstrap.php
```diff
@@ -51,7 +53,24 @@ public static function tracerOnce()
...
+                        }
+                        /*
+                         * Having this priority sampling here is actually a bug (should happen after service name
+                         * substitutions), but it was this way before the refactor, so let's fix this in a
+                         * subsequent release, when we will have ported everything else to the extension.
+                         */
+                        $tracer = GlobalTracer::get();
+                        if (method_exists($tracer, "enforcePrioritySamplingOnRootSpan")) {
+                            $tracer->enforcePrioritySamplingOnRootSpan();
+                        }
```
labbati, line 72: I am not understanding this part. Reaching out in private to have a quick sync.

labbati (commented):
Thank you Bob, added a few comments in the meantime. I am also going to test a specific edge case I have in mind and that I think is not covered in a test, but I will report back on that with an additional comment

src/DDTrace/Bootstrap.php
```diff
@@ -126,6 +145,14 @@ private static function initRootSpan(Tracer $tracer)
             foreach (Private_\util_extract_configured_headers_as_tags($httpHeaders, true) as $tag => $value) {
                 $span->setTag($tag, $value);
             }
+
+            if (PHP_VERSION_ID >= 80000) {
+                foreach ($httpHeaders as $header => $value) {
+                    if (stripos($header, Propagator::DEFAULT_ORIGIN_HEADER) === 0) {
```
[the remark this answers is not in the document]
bwoebi: while that's totally possible, it's in line with the current handling: https://github.com/DataDog/dd-trace-php/blob/master/src/DDTrace/Propagators/CurlHeadersMap.php#L33 If we want to change that, we probably should do that separately...?

src/DDTrace/Tracer.php
```diff
@@ -253,7 +274,7 @@ public function startActiveSpan($operationName, $options = [])
             $span->setTag(Tag::SERVICE_NAME, $parentService);
         }
 
-        return $this->scopeManager->activate($span, $options->shouldFinishSpanOnClose());
+        return $this->scopeManager->activate($span, $options->shouldFinishSpanOnClose() && (PHP_VERSION_ID >= 80000 && $span->getParentId() == 0));
```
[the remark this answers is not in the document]
bwoebi: Yeah that's not right. And there's also a failing test for that. (but as said, this current PR is fully untested, locally I've done some progress.)

ext/php8/serializer.c
bwoebi: Yes - and yes, for sure that'll get a test.

src/DDTrace/Bootstrap.php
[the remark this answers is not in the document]
labbati: good point, please create a ticket that we can address later as a low hanging fruit

ext/php8/span.h
```diff
@@ -14,25 +14,23 @@ static const int ddtrace_num_error_tags = 3;
...
     uint64_t trace_id;
     uint64_t parent_id;
     uint64_t span_id;
     uint64_t start;
-    union {
-        uint64_t duration_start;
-        uint64_t duration;
-    };
+    uint64_t duration_start;
+    uint64_t duration;
```
SammyK, line 24: What is the motivation for removing the `union`?

SammyK (commented):
Looking great so far! I'm still reviewing this but I left a few comments/questions so far.

ext/php8/span.h
bwoebi: Because the userland interface wants to distinguish between finished and unfinished spans; now we can just return duration == 0, then userland knows unfinished

ext/php8/ddtrace.c
```diff
@@ -1249,6 +1327,50 @@ static PHP_FUNCTION(dd_trace_peek_span_id) {
...
 }
 
+/* {{{ proto string dd_trace_peek_span() */
+static PHP_FUNCTION(dd_trace_peek_span) {
+    UNUSED(execute_data);
+    if (DDTRACE_G(open_spans_top)) {
+        RETURN_OBJ_COPY(&DDTRACE_G(open_spans_top)->span.std);
+    }
+    RETURN_NULL();
+}
```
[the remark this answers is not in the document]
bwoebi: The span is is currently not exposed on SpanData, but we may add getters for that...

ext/php8/span.h
SammyK: Ah, makes sense. Thanks for clarifying.

PR #1258, opened 2021-06-16
Enforce metrics to be floats in internal serializer

bwoebi:
### Description

Internal serializer is currently preserving the metrics array as-is. I'm now only preserving string keys and converting them to double as mandated by the agent. This is especially important when cutting out userland from serializing.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/php5/serializer.c
+++ b/ext/php5/serializer.c
@@ -152,14 +152,6 @@ static zval *_read_span_property(zval *span_data, const char *name, size_t name_
     return zend_read_property(ddtrace_ce_span_data, span_data, name, name_len, 1 TSRMLS_CC);
 }
 
-static void _add_assoc_zval_copy(zval *el, const char *name, zval *prop) {
-    zval *value;
-    ALLOC_ZVAL(value);
-    INIT_PZVAL_COPY(value, prop);
-    zval_copy_ctor(value);
-    add_assoc_zval(el, name, value);
-}
-
 /* gettraceasstring() macros
  * @see https://github.com/php/php-src/blob/PHP-5.4/Zend/zend_exceptions.c#L364-L395
  * {{{ */
@@ -463,8 +455,26 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array TSR
     _serialize_meta(el, span_fci TSRMLS_CC);
 
     zval *metrics = _read_span_property(span->span_data, ZEND_STRL("metrics") TSRMLS_CC);
-    if (Z_TYPE_P(metrics) == IS_ARRAY) {
-        _add_assoc_zval_copy(el, "metrics", metrics);
+    if (Z_TYPE_P(metrics) == IS_ARRAY && zend_hash_num_elements(Z_ARRVAL_P(metrics))) {
+        zval *metrics_zv;
+        ALLOC_INIT_ZVAL(metrics_zv);
+        array_init(metrics_zv);
+        HashPosition pos;
+        zval **metric_value;
+        for (zend_hash_internal_pointer_reset_ex(Z_ARRVAL_P(metrics), &pos);
+             zend_hash_get_current_data_ex(Z_ARRVAL_P(metrics), (void **)&metric_value, &pos) == SUCCESS;
+             zend_hash_move_forward_ex(Z_ARRVAL_P(metrics), &pos)) {
+            ulong num_key;
+            char *str_key;
+            if (zend_hash_get_current_key_ex(Z_ARRVAL_P(metrics), &str_key, NULL, &num_key, 0, &pos) ==
+                HASH_KEY_IS_STRING) {
+                zval value;
+                MAKE_COPY_ZVAL(metric_value, &value);
+                convert_to_double(&value);
+                add_assoc_double(metrics_zv, str_key, Z_DVAL(value));
+            }
+        }
+        add_assoc_zval(el, "metrics", metrics_zv);
     }
 
     add_next_index_zval(array, el);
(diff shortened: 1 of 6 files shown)


ext/php5/serializer.c
```diff
@@ -463,8 +455,26 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array TSR
...
+        array_init(metrics_zv);
+        HashPosition pos;
+        zval **metric_value;
+        for (zend_hash_internal_pointer_reset_ex(Z_ARRVAL_P(metrics), &pos);
+             zend_hash_get_current_data_ex(Z_ARRVAL_P(metrics), (void **)&metric_value, &pos) == SUCCESS;
+             zend_hash_move_forward_ex(Z_ARRVAL_P(metrics), &pos)) {
+            ulong num_key;
+            char *str_key;
+            if (zend_hash_get_current_key_ex(Z_ARRVAL_P(metrics), &str_key, NULL, &num_key, 0, &pos) ==
+                HASH_KEY_IS_STRING) {
```
bwoebi, line 470: clang-format, the ugly-ficator.

ext/php5/serializer.c
```diff
@@ -463,8 +455,26 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array TSR
...
+                zval value;
+                MAKE_COPY_ZVAL(metric_value, &value);
```
bwoebi, line 472: Not sure whether that's the proper way, doing a copy ctor there (so if there are strings, they'll be memory copied first to be destroyed immediately in convert_to_double - my PHP 5 knowledge is lacking) … but it works.

ext/php5/serializer.c
labbati: any other alternative you would recommend?

bwoebi: I personally hate arbitrary low line length limits (soft wraps in the editor are a thing and you always can just make your window a bit wider). At least for me they incentivize me to use less descriptive names and less obvious code structuring. This was especially worse in the past for me, without autoformatter but enforced rules.

labbati: The google Cpp coding style, on which we sit, [is even stricter](https://google.github.io/styleguide/cppguide.html#Line_Length) than [us](https://github.com/DataDog/dd-trace-php/blob/master/.clang-format#L5).

I am ok with either increasing (mostly) or removing (mmm) that limit if everyone agrees (cc @SammyK, @morrisonlevi, @Anilm3).

SammyK: I'm fine with increasing the line length limit or removing the limit. No strong opinions here.

SammyK (approved):
Thanks for identifying and normalizing this behavior @bwoebi. LGTM 👍

ext/php5/serializer.c
morrisonlevi: I actually prefer an 80 column limit. I acknowledge that I am a bit of a rarity these days, with large monitors and all. But, we look at code in a variety of situations. For example, two column flows are common, such as regular diffs, and sometimes we need to do things with 3 columns, such as solving non-trivial merge conflicts in a UI.

I will admit that C makes some of this harder for us, since it doesn't have namespaces nor methods. I also prefer 2 spaces for indents if we're using spaces, which in this case would save 6 characters; ironically enough to not make the line in the PR wrap :)

-----

Anyway, all of this is background to say I don't support increasing the column limit further; 120 columns is already very generous in my opinion. Some options to avoid the admittedly ugly line wrap in this case include:

 1. Saving the return value to a variable, and then comparing the variable in the `if`.
 2. Since `Z_ARRVAL_P(metrics)` is used a few times, it could also be saved to a variable with a name shorter than `Z_ARRVAL_P(metrics)`.

I listed these as options because I don't think they make the code worse by doing them.

bwoebi: @morrisonlevi While in this specific case you can work around it, but then there are examples like: https://github.com/DataDog/dd-trace-php/blob/f859f1227d5db2de2816d00e21fbb3099843d849/ext/php8/handlers_curl.c#L295 (or generally the zpp invocations there). `if (call(...) == RETVALUE) {` should generally be able to fit in one single line. (yes, you could make a case to add a newline before the first &user_arg, but I don't think this helps)
And in particular, we just let the format fixer run, uglifying the code. If we had to manually fix these occurrences, then you can make most of the places beautiful. But as is, nobody will care enough, we're just running the format fixer as part of a mild inconvenience.

I agree with you that short lines are usually nice, and that you should try to use shorter lines. But doing that in a nice (and helpful) way is not a result of just running the fixer, but rather actively avoiding chaining multiple 6 conditions with more complex code for example. So doing what you say will solve the issue properly in most cases. Having the code fixer automatically hard wrap lines at random places does not.

I'd say, our current usage of format fixer is just as effective if we'd have soft line wrapping and you'd shrink your window to 120 chars. I see absolutely no upside to the current way of doing it.

I would rather say "hey, let's have a soft limit of 120 chars - if there are excessively long lines e.g. thanks to a function requiring 5 arguments which have descriptive names - longer lines are allowed". I.e. when you review the code and see an obnoxiously long line for no reason, comment (or fix) it.

I definitely have written lines of 400 chars which are just a long line of nested conditions with a tenary and find them hard to read when I come back to that code. I am by no means encouraging it. 

----

Btw. I disagree with separating enum return values from their to-be-checked place (obviously when trivially avoidable). Code locality is important when reading. Rather compare it directly in an if or switch. So yes, this makes the code worse.

PR #1260, opened 2021-06-17
ZAI config

SammyK:
### Description

breaking changes:
- circuit breaker environment variable setting (`putenv("DD_TRACE_CIRCUIT_BREAKER_ENV_MAX_CONSECUTIVE_FAILURES=...")` and `putenv("DD_TRACE_CIRCUIT_BREAKER_ENV_RETRY_TIME_MSEC=...")`) will be ignored at runtime. To change these at runtime use `ini_set()` instead.
- `DD_TRACE_SANDBOX_ENABLED` has been removed
- Support for the legacy `dd_trace_env_config("trace.agent.port")` in dotted format has been removed.
- There is no distinction between not set and environment variables set to the default value anymore.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


ext/php8/ddtrace.c
```diff
@@ -433,6 +438,7 @@ static PHP_RINIT_FUNCTION(ddtrace) {
     // Things that should only run on the first RINIT
     int expected_first_rinit = 1;
     if (atomic_compare_exchange_strong(&ddtrace_first_rinit, &expected_first_rinit, 0)) {
+        zai_config_first_time_rinit();
```
bwoebi, line 441: It _runs_ only once, but let's say there are two requests coming in nearly simultaneously after a restart; will the second request then possibly run before the first_time_rinit here has completed? I suspect this needs some sort of loop (or lock or whatever) at the end of this if() branch to be race-condition free.

morrisonlevi: I recommend `pthread_once`. I'm using it in the profiler.

bwoebi: Nice, didn't know about that one, looks good to me from the docs :-)

zend_abstract_interface/config/config_ini.c
```diff
@@ -0,0 +1,140 @@
...
+        // TODO Log cannot find ID
+        return FAILURE;
+    }
+
+    zval tmp;
+    ZVAL_UNDEF(&tmp);
+    zai_string_view value_view = {.len = ZSTR_LEN(new_value), .ptr = ZSTR_VAL(new_value)};
+
+    if (zai_config_decode_value(value_view, memoized_entires[id].type, &tmp, /* persistent */ false)) {
+        zai_config_update_runtime_config(id, &tmp);
```
bwoebi, line 84: Do we assume that every config is safe to be changed at any time? … or maybe it's safe but has no effect.
I think in these cases we should forbid changing.

zend_abstract_interface/config/config_ini.c
```diff
@@ -0,0 +1,140 @@
...
+    zend_ini_entry_def entry_defs[1 + /* terminator entry */ 1] = {{0}, {0}};
+    zend_ini_entry_def *entry = &entry_defs[0];
+
+    entry->name = ini_name->ptr;
+    entry->name_length = ini_name->len;
+    entry->value = "";
+    entry->value_length = 0;
+    entry->on_modify = ZaiConfigOnUpdateIni;
+    entry->modifiable = PHP_INI_ALL;
+
```
bwoebi, line 110: Do we want to use zend_ini_boolean_displayer_cb for our purely boolean switches? (like every boolean ini in php does)

zend_abstract_interface/config/config_decode.c
```diff
@@ -0,0 +1,201 @@
...
+                            while (*++data && *data != ',') {
+                                if (*data != ' ' && *data != '\t' && *data != '\n') {
+                                    value_end = data;
+                                }
+                            }
+                        }
+
+                        zval val;
+                        zend_string *key = zend_string_init(key_start, key_end - key_start + 1, persistent);
+                        if (persistent) {
```
bwoebi, line 142: `ZVAL_NEW_STR(&val, zend_string_init(value_start, value_end - value_start + 1, persistent));` maybe instead of branching here and duplicating args

bwoebi: Similarly in zai_config_decode_string below

zend_abstract_interface/config/config_decode.c
```diff
@@ -0,0 +1,201 @@
...
+ */
+static bool zai_config_is_valid_int_format(const char *str) {
+    while (*str) {
+        if (!isdigit(*str) && !isspace(*str)) return false;
+        str++;
+    }
+    return true;
+}
+
+static bool zai_config_decode_int(zai_string_view value, zval *decoded_value) {
```
bwoebi, line 83: maybe just use `zend_atol` to mirror `ZEND_INI_MH(OnUpdateLong)` behaviour?

bwoebi: ... We seem to be validating values here, which is quite atypical for inis in PHP.  In PHP inis are generally only supposed to fail if the value cannot be worked with. (like ini value expects number > 0 and you get -1.)
I'm not sure whether we want to break with that.

zend_abstract_interface/config/config_decode.c
```diff
@@ -0,0 +1,201 @@
...
+#include <assert.h>
+#include <ctype.h>
+#include <main/php.h>
+#include <stdbool.h>
+#include <string.h>
+#include <strings.h>
+
+#include "config.h"
+
+static bool zai_config_decode_bool(zai_string_view value, zval *decoded_value) {
```
bwoebi, line 12: As per `zend_ini_parse_bool` at least on and yes are also valid values. (independently of whether we should validate or not.)

zend_abstract_interface/config/config_decode.c
```diff
@@ -0,0 +1,201 @@
...
+}
+
+static bool zai_config_decode_double(zai_string_view value, zval *decoded_value) {
+    if (!zai_config_is_valid_double_format(value.ptr)) return false;
+
+    char *endptr = (char *)value.ptr;
+
+    // The strtod function is a bit tricky, so I've quoted docs to explain code
+
+    /* Since 0 can legitimately be returned on both success and failure, the
```
bwoebi, line 52: Actually, strtod does not return 0 on failure. An invalid double is successful. The only case which fails is unrepresentability of a valid value, which may set errno to ERANGE. The code is fine, the comment not so much.

zend_abstract_interface/config/config_decode.c
```diff
@@ -0,0 +1,201 @@
...
+     * calling program should set errno to 0 before the call, and then
+     * determine if an error occurred by checking whether errno has a nonzero
+     * value after the call.
+     */
+    errno = 0;
+    double result = strtod(value.ptr, &endptr);
```
bwoebi, line 58: This is locale sensitive. Do not use strtod in PHP code unless you actually want things differently handled by locale. (e.g. the comma vs dot as decimals separator.)
`zend_strtod` handles this properly for you.

zend_abstract_interface/config/config_ini.c
```diff
@@ -0,0 +1,140 @@
...
+        return false;
+    }
+
+    /* ZEND_INI_END() adds a null terminating entry */
+    zend_ini_entry_def entry_defs[1 + /* terminator entry */ 1] = {{0}, {0}};
+    zend_ini_entry_def *entry = &entry_defs[0];
+
+    entry->name = ini_name->ptr;
+    entry->name_length = ini_name->len;
+    entry->value = "";
```
bwoebi, line 106: So, `ini_restore()` … will always set the value to an empty string?

ext/php8/configuration.h
```diff
@@ -4,10 +4,38 @@
...
+}
+
+static inline zai_string_view get_dd_service(void) {
+    zval *val = zai_config_get_value(DDTRACE_CONFIG_DD_SERVICE);
+    return (zai_string_view){.len = Z_STRLEN_P(val), .ptr = Z_STRVAL_P(val)};
+}
+
+static inline zend_array *get_dd_tags(void) {
+    zval *val = zai_config_get_value(DDTRACE_CONFIG_DD_TAGS);
+    GC_ADDREF(Z_ARRVAL_P(val));
```
bwoebi, line 25: Either addref strings as well, or don't either. I wouldn't addref either. user can incref/destroy if needed.

ext/php8/configuration.h
```diff
@@ -4,10 +4,38 @@
...
 
+// TODO Tie these into the X Macros below
+static inline bool get_dd_trace_debug(void) {
+    zval *val = zai_config_get_value(DDTRACE_CONFIG_DD_TRACE_DEBUG);
+    return Z_TYPE_P(val) == IS_TRUE;
+}
+
+static inline zai_string_view get_dd_service(void) {
+    zval *val = zai_config_get_value(DDTRACE_CONFIG_DD_SERVICE);
+    return (zai_string_view){.len = Z_STRLEN_P(val), .ptr = Z_STRVAL_P(val)};
```
bwoebi, line 20: I see no real gain here in converting to a zai_string_view, especially as some options like DD_SERVICE will end just straight up in the meta array. I.e. us ending up cycling the datastructure back and forth. Just return the string.
Also this string_view may not be stored as is, as its lifetime is dependent on the config value (i.e. when the value is changed and nothing pointing to it, this will point to unrelated memory). Keep it simple and don't add new pitfalls here.

ext/php8/configuration.h
```diff
@@ -4,10 +4,38 @@
...
+
+static inline zend_array *get_dd_tags(void) {
+    zval *val = zai_config_get_value(DDTRACE_CONFIG_DD_TAGS);
+    GC_ADDREF(Z_ARRVAL_P(val));
+    return Z_ARRVAL_P(val);
+}
+
+static inline int64_t get_dd_trace_agent_port(void) {
+    zval *val = zai_config_get_value(DDTRACE_CONFIG_DD_TRACE_AGENT_PORT);
+    return (int64_t)Z_LVAL_P(val);
```
bwoebi, line 31: We probably should just use `zend_long` here, i.e. the native data type. I wouldn't really say something is a 64 bit integer when it actually will only ever use 32 bits.

ext/php8/ddtrace_config.c
```diff
@@ -0,0 +1,56 @@
...
+#include <assert.h>
+#include <ctype.h>
+#include <stdio.h>
+#include <string.h>
+
+#define DD_CONFIG_ENTRY(name, type, default) ZAI_CONFIG_ENTRY(DDTRACE_CONFIG_##name, name, type, default)
+#define DD_CONFIG_ALIASED_ENTRY(name, type, default, aliases) \
+    ZAI_CONFIG_ALIASED_ENTRY(DDTRACE_CONFIG_##name, name, type, default, aliases)
+
+static void dd_ini_env_to_ini_name(zai_string_view env_name, zai_config_ini_name *ini_name) {
```
bwoebi, line 12: When I read this function I just think ... does it really need 30 LoC to do `strtolower(strtr($input, ["DD_TRACE_" => "datadog.trace", "DD_" => "datadog."]))` in C? :-D Maybe it does...

bwoebi:
As a general feeling, you are trying to focus too much on avoiding zend_string (may make sense for some parts of the code, but some clearly php7/8 specific things should just freely use zend_string) and memory allocations for things which are only happening at minit. I consider fixed buffers rather a micro-optimization if they aren't frequently allocated over and over.

zend_abstract_interface/config/config.h
```diff
@@ -0,0 +1,109 @@
...
+#include <main/php.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "../zai_string/string.h"
+#include "config_decode.h"
+#include "config_ini.h"
+
+#define ZAI_CONFIG_ENTRIES_COUNT_MAX 20
```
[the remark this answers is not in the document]
SammyK: > And here, within the API you artificially limit this. For ... reasons?

This number is too low, but the original thinking was to make this about 3x (or more) the number of configuration values that we currently have to make sure we don't hit that limit anytime soon (or ever). The idea was to prioritize static memory allocation and be more in line with MISRA C. @morrisonlevi might have more thoughts here.

ext/php8/configuration.h
SammyK: Yes, these are very good points. Returning a `zend_string` on PHP 7+ would make sense to me.

ext/php8/configuration.h
SammyK: 👍 Yes, let's let the caller decide what to do and not addref here.

zend_abstract_interface/config/config.h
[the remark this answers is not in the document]
bwoebi: Are you here talking about the fallibility of malloc()? Or how is that related to MISRA C? At least within the PHP ecosystem it is always assumed that malloc() won't fail.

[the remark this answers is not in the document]
morrisonlevi: We need to strongly prefer static allocations when it's feasible. It's not always feasible of course, but this seems like a case where it is -- we are in total control over the number of configuration entries we support.

ext/php8/configuration.h
SammyK: For this specific case that targets the Agent port, I agree on using 32 bit ints (or smaller). Although this getter is specific to the Agent port, I was planning on incorporating these functions into the X Macro setup we have for the existing config. In that case, I'm not sure we want to artificially limit all `int` configuration values to 32-bit.

Also, since `zend_long` is a feature of PHP 7+, what do you suggest for PHP 5? Do you see any value in providing a generic type in this case to reduce the minor API changes across the major PHP versions?

ext/php8/ddtrace_config.c
SammyK: This is the story of my life. 😆

zend_abstract_interface/config/config.h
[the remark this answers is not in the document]
bwoebi: @morrisonlevi Why do we need to? At least the big disadvantage with static allocations is that it's easy to miss a bounds check. Perhaps by chance the couple bytes after the static array are usually not _important_ and it randomly explodes.

This code is once-init init-time malloc(), so performance overhead is not a problem either.

Like, we set this to 20, and add a 21st config entry and with some bad luck we don't even get an immediate segfault.

zend_abstract_interface/config/config_decode.c
SammyK: Good point. We might want to adopt these two truthy values and their counterparts to be consistent with INI.

ext/php8/configuration.h
bwoebi: php5 uses simply a long. Okay, I agree, isn't going to do any harm to always use int64_t.

zend_abstract_interface/config/config_decode.c
SammyK: Nice catch. This comment must have eventually [gotten outdated at the source](https://github.com/DataDog/dd-trace-php/blob/fe830c017ce94436704429131bb151d20fd6391d/ext/php8/env_config.c#L124-L128).

zend_abstract_interface/config/config_decode.c
SammyK: Very good point. Although `strtod` is what we're currently using to parse doubles, I don't see any reason why we would not want to change this to the Zend flavor.

zend_abstract_interface/config/config_ini.c
bwoebi: As an example: `DD_TRACE_RETAIN_THREAD_CAPABILITIES` should really be `PHP_INI_SYSTEM` as it's a startup-only config, it just does not make sense to toggle that at runtime.

zend_abstract_interface/config/config_decode.c
SammyK: This all sounds good to me. If we end up tweaking the behavior of env var parsing to be more in line with INI, I think we should go for it. That is, unless there is an obvious BC or something. :)

zend_abstract_interface/config/config_decode.c
SammyK: Nice catch. 👍

zend_abstract_interface/config/config.h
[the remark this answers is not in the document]
bwoebi: I general, if you can make the bounds of such arrays dependent on something like `sizeof(some_array)` or such, a statically allocated array is great. An arbitrarily defined size in another component not having access to whatever is actually determining the array size is suboptimal.

zend_abstract_interface/config/config_ini.c
SammyK: > or maybe it's safe but has no effect.

Correct. As we briefly discussed a bit earlier, some of the cases where the config value was set too late could possibly have workarounds; (e.g. setting `DD_TRACE_GENERATE_ROOT_SPAN=0` at runtime could drop the existing spans.)

I guess that might mean we should consider a callback or something so that the tracer could subscribe to config changes?

> I think in these cases we should forbid changing.

How would ZAI config know it was too late for a particular value to be set? Perhaps if we add a callback to pass the decoded value to the tracer before the runtime config is updated, ZAI config can check the return value of the callback to determine if the value should be updated?

bwoebi: We have to tell ZAI config. Trivially the "too late" part is usually PHP_INI_SYSTEM vs PHP_INI_ALL. For sure, we could be more granular here, but I don't think this is worth it.

bwoebi: If we want more flexibility, especially for cases like root span, yes, then we need a callback.

zend_abstract_interface/config/config.h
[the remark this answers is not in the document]
morrisonlevi: In the private file, we can `_Static_assert` that the public constant is compatible with the array size and refuse to compile. It's not a problem.

zend_abstract_interface/config/config_ini.c
SammyK: Doesn't `ini_restore()` restore the original value, not the default value?

zend_abstract_interface/config/config_ini.c
SammyK: Makes sense to me! :)

zend_abstract_interface/config/config.h
```diff
@@ -0,0 +1,122 @@
...
+} zai_config_result;
+
+// assertions + error_zal
+// If caller wants to return to userland: Caller must refcount++ & dtor
+zval *zai_config_get_value(zai_config_id id);
+zai_config_result zai_config_set_value(zai_config_id id, zval *value);
+
+// Only really used via userland functions
+//      \DDTrace\get_config(string $name)
+//      \DDTrace\set_config(string $name, mixed $value)
```
SammyK, line 116: Some of the comments in this file are outdated or just random stubbed out notes that I was planning on removing/updating later. :)

zend_abstract_interface/config/config_ini.h
```diff
@@ -0,0 +1,33 @@
...
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include "../env/env.h"
+#include "../zai_string/string.h"
+
+#include "config.h"
+
+#endif  // ZAI_CONFIG_INI_H
```
SammyK, line 14: Is this not at the bottom for a specific reason?

zend_abstract_interface/headers/tests/php7-8/headers.cc
```diff
@@ -20,7 +20,7 @@ extern "C" {
     }
 
 static void define_server_value(zval *arr) {
-    add_assoc_string(arr, "HTTP_MY_HEADER", "Datadog");
+    add_assoc_string(arr, "HTTP_MY_HEADER", (char *) "Datadog");
```
SammyK, line 23: Should this only apply to PHP <= 7.1?

SammyK (commented):
The ZAI config parts LGTM. 👍 I left a few questions/comments.

zend_abstract_interface/config/config_ini.h
bwoebi: It is, because of the recursive include ... config.h wants to include config_ini.h. But config_ini.h references types only defined in config.h. If you put this at the bottom compilation will fail :-)

zend_abstract_interface/headers/tests/php7-8/headers.cc
bwoebi: Yes, though it fixes the compile warning on all versions. And I saw no reason to include an #ifdef if this doesn't do any harm on 7.2+.

bwoebi:
On the topic of ZAI_CONFIG_ENTRIES_COUNT_MAX we'll do a static assertion; it's trivial with the X macro to expand CFG (configuration.h) to `1, ` and use it within `sizeof((uint8_t[]){ DD_CONFIGURATION })` and compare.

src/private/functions.php
```diff
@@ -68,21 +68,8 @@ function _util_uri_apply_rules($uriPath, $incoming)
 
     // We can now be in one of 3 cases:
     //   1) At least one among DD_TRACE_RESOURCE_URI_FRAGMENT_REGEX and DD_TRACE_RESOURCE_URI_MAPPING_INCOMING|OUTGOING
-    //      is defined. Then ignore legacy DD_TRACE_RESOURCE_URI_MAPPING and apply *new normalization*.
-    //   2) Only DD_TRACE_RESOURCE_URI_MAPPING is defined, then apply *legacy normalization* for backward compatibility.
-    //   3) Nothing is defined, then apply *new normalization*.
-
-    // DEPRECATED: Applying legacy normalization for backward compatibility if preconditions are matched.
-    $legacyMappings = getenv('DD_TRACE_RESOURCE_URI_MAPPING');
```
SammyK, line 76: What was the motivation for removing this? Not necessary for this PR, but it might be a good idea to start adding checks for fully removed env vars to the startup logs/php info page.

tests/Common/BaseTestCase.php
```diff
@@ -53,7 +70,7 @@ protected function withDebugLogger()
     protected function putEnvAndReloadConfig($putenvs = [])
     {
         foreach ($putenvs as $putenv) {
-            \putenv($putenv);
+            $this->putEnv($putenv);
```
SammyK, line 73: Tiny nit
```suggestion
            self::putEnv($putenv);
```

tests/Common/TracerTestTrait.php
```diff
@@ -243,6 +243,7 @@ public function tracesFromWebRequest($fn, $tracer = null)
     private function parseRawDumpedTraces($rawTraces)
     {
         $traces = [];
+        var_dump($rawTraces);
```
SammyK, line 246:
```suggestion
```

SammyK (commented):
I reviewed the ddtrace parts but will re-review the ZAI config parts tomorrow. So far it looks like a fantastic cleanup! 💯

ext/php8/ddtrace.c
```diff
@@ -394,7 +388,6 @@ static PHP_MSHUTDOWN_FUNCTION(ddtrace) {
     UNREGISTER_INI_ENTRIES();
 
     if (DDTRACE_G(disable)) {
-        ddtrace_config_shutdown();
         return SUCCESS;
```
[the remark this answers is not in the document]
bwoebi: You're right, fixing.

ext/php8/configuration.c
```diff
@@ -1,346 +1,148 @@
 #include "configuration.h"
```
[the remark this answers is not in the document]
bwoebi: I guess the main magic is just APPLY_N(macro, ...) to apply a macro to all variadics, to apply checks to all of CALIASES

src/private/functions.php
bwoebi: I can revert that if you wish - the motivation initially was some confusion about deprecation checks.

ext/php8/configuration.h
```diff
@@ -73,71 +26,138 @@ void ddtrace_config_shutdown(void);
...
+    CALIAS(STRING, DD_TRACE_REQUEST_INIT_HOOK, DD_DEFAULT_RQUEST_INIT_HOOK_PATH,                              \
+           CALIASES("DDTRACE_REQUEST_INIT_HOOK"), .ini_change = zai_config_system_ini_change)                 \
+    CONFIG(STRING, DD_TRACE_AGENT_URL, "", .ini_change = zai_config_system_ini_change)                        \
+    CONFIG(STRING, DD_AGENT_HOST, "localhost", .ini_change = zai_config_system_ini_change)                    \
+    CONFIG(BOOL, DD_DISTRIBUTED_TRACING, "true")                                                              \
+    CONFIG(STRING, DD_DOGSTATSD_PORT, "8125")                                                                 \
+    CONFIG(STRING, DD_ENV, "")                                                                                \
+    CONFIG(BOOL, DD_AUTOFINISH_SPANS, "false")                                                                \
+    CONFIG(BOOL, DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED, "true")                                              \
+    CONFIG(SET, DD_INTEGRATIONS_DISABLED, "all")                                                              \
```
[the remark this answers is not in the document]
bwoebi: Previously the logic was:
```
if envvar("DD_TRACE_" . $integration . "_ENABLED) is defined then
   return envvar("DD_TRACE_" . $integration . "_ENABLED")
else
   return $integration not in envvar("DD_INTEGRATIONS_DISABLED")
```

Now, we cannot do an "is defined" check on config values anymore (due to the nature of always defined ini)

hence I've inverted the logic to say:

```
if config("DD_INTEGRATIONS_DISABLED") is not default then
   return config("DD_TRACE_" . $integration . "_ENABLED")
else
   return $integration not in envvar("DD_INTEGRATIONS_DISABLED")
```

`all` is probably a bad default value, but it must not be empty, as empty already has a meaning (empty set - everything enabled), and we need to detect whether the value is non-default.

I'm changing it to `default`, that's more sensible (and the default being: look at the DD_TRACE_{INTEGRATION}_ENABLED configs).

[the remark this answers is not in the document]
SammyK: Thanks for clarifying. `default` makes a lot more sense to me. 👍

src/private/functions.php
SammyK: Got it. I think we should probably keep this then unless there is a strong technical reason to remove it.

bwoebi: Will revert that then tomorrow :-)

zend_abstract_interface/config/php5/config_ini.c
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+}
+#endif
+
+void zai_config_ini_mshutdown() {
+    used_original_ini_values = 0;
+    for (int i = 0; i < used_original_ini_values; ++i) {
+        free(original_ini_values[i]);
+    }
```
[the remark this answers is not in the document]
bwoebi: you mean none? :-D Yep, fixing. (I'm sure that was correct at some point and then I somehow broke it)

zend_abstract_interface/config/config.h
```diff
@@ -0,0 +1,102 @@
...
+    uint8_t names_count;
+    zai_config_type type;
+    zval decoded_value;
+    zai_string_view default_encoded_value;
+    // The index of the name that was used to set the value
+    //     anything > 0 is deprecated
+    //     -1 == not set from env or system ini
+    int16_t name_index;
+    zai_config_apply_ini_change ini_change;
+} zai_config_memoized_entry;
```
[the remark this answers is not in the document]
bwoebi: I don't care much, so I will just apply that :-D

zend_abstract_interface/config/php5/config_ini.c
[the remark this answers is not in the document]
SammyK: Haha! Yes, I mean none. 🤦

I'm curious why none of the leak detection tests didn't catch this?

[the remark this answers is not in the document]
bwoebi: It's malloc() memory, not emalloc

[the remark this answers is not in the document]
SammyK: ~Sure, but ASan should have detected this, no?~ Ah! We don't have an ASan build on PHP 5. 😞 We definitely need to add valgrind support/some other leak detection to the ZAI tests for non-ASan builds (in a separate PR of course). And I'm assuming this didn't show up in the existing phpt leak-detection tests because this code path is never hit in any of those tests?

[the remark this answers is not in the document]
bwoebi: And the latter is correct: because there is no code using ZAI config for PHP 5 yet. Only our zai testsuite.

SammyK (commented):
Fantastic work @bwoebi! 👍 Thanks again for taking this one over and going down the config rabbit hole.

bwoebi (approved):
self-approval because Sammys PR and he cannot approve, github is funny!

PR #1262, opened 2021-06-21
Moving spans from userland to extension

bwoebi:
### Description

Doing a lot of things:
 - Nearly everything happening in Tracer->flush() is now in the extension
 - Spans are allocated and closed directly instead of their ids (via DDTrace\start_span() and DDTrace\close_span())
 - Spans now expose duration and start time via methods
 - The root span is (if there is configured to be one) allocated and closed from within the extension - close at the very last moment: RSHUTDOWN
 - Manual flushing now happens via a new DDTrace\flush()

(Originally #1251, but github doesn't let me reopen that one.)

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/bridge/_files.php
+++ b/bridge/_files.php
@@ -20,6 +20,7 @@ return [
     __DIR__ . '/../src/DDTrace/StartSpanOptionsFactory.php',
     __DIR__ . '/../src/DDTrace/Time.php',
     __DIR__ . '/../src/DDTrace/Transport/Http.php',
+    __DIR__ . '/../src/DDTrace/Transport/Internal.php',
     __DIR__ . '/../src/api/Type.php',
     __DIR__ . '/../src/DDTrace/Encoder.php',
     __DIR__ . '/../src/DDTrace/Util/Runtime.php',
--- a/ext/php8/configuration.c
+++ b/ext/php8/configuration.c
@@ -30,8 +30,24 @@ void ddtrace_reload_config(void) {
         free(ddtrace_memoized_configuration.getter_name); \
     }                                                     \
     ddtrace_memoized_configuration.__is_set_##getter_name = FALSE;
-#define HASH(getter_name, ...)                                     \
-    zend_hash_release(ddtrace_memoized_configuration.getter_name); \
+// this will leak memory if RC != 1, but prevents crashes. This could only ever happens if an user uses
+// \dd_trace_internal_fn('ddtrace_reload_config'); at the wrong point in time - which is an internal testsuite function
+// which users are not supposed to use
+#define HASH(getter_name, ...)                                                                    \
+    if (GC_REFCOUNT(ddtrace_memoized_configuration.getter_name) == 1) {                           \
+        zval* _value;                                                                             \
+        zend_string* _key;                                                                        \
+        ZEND_HASH_FOREACH_STR_KEY_VAL(ddtrace_memoized_configuration.getter_name, _key, _value) { \
+            if (GC_REFCOUNT(_key) > 1) {                                                          \
+                GC_TRY_ADDREF(_key);                                                              \
+            }                                                                                     \
+            if (Z_REFCOUNT_P(_value) > 1) {                                                       \
+                Z_TRY_ADDREF_P(_value);                                                           \
+            }                                                                                     \
+        }                                                                                         \
+        ZEND_HASH_FOREACH_END();                                                                  \
+        zend_hash_release(ddtrace_memoized_configuration.getter_name);                            \
+    }                                                                                             \
     ddtrace_memoized_configuration.__is_set_##getter_name = FALSE;
 #define BOOL(getter_name, ...) ddtrace_memoized_configuration.__is_set_##getter_name = FALSE;
 #define INT(getter_name, ...) ddtrace_memoized_configuration.__is_set_##getter_name = FALSE;
--- a/ext/php8/configuration.h
+++ b/ext/php8/configuration.h
@@ -88,6 +88,8 @@ void ddtrace_config_shutdown(void);
     BOOL(get_dd_distributed_tracing, "DD_DISTRIBUTED_TRACING", true)                                                 \
     CHAR(get_dd_dogstatsd_port, "DD_DOGSTATSD_PORT", "8125")                                                         \
     CHAR(get_dd_env, "DD_ENV", "")                                                                                   \
+    BOOL(get_dd_autofinish_spans, "DD_AUTOFINISH_SPANS", false)                                                      \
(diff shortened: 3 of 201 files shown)


Makefile
```diff
@@ -7,7 +7,7 @@ BUILD_DIR := $(PROJECT_ROOT)/tmp/build_$(BUILD_SUFFIX)
 SO_FILE := $(BUILD_DIR)/modules/ddtrace.so
 WALL_FLAGS := -Wall -Wextra
 EXTRA_CFLAGS :=
-CFLAGS := -O2 $(EXTRA_CFLAGS) $(WALL_FLAGS)
+CFLAGS := -g $(EXTRA_CFLAGS) $(WALL_FLAGS)
```
bwoebi, line 10: I'll revert that before merging...

ext/php8/configuration.h
```diff
@@ -120,10 +122,6 @@ void ddtrace_config_shutdown(void);
     BOOL(get_dd_trace_generate_root_span, "DD_TRACE_GENERATE_ROOT_SPAN", true)                                       \
     BOOL(get_dd_trace_sandbox_enabled, "DD_TRACE_SANDBOX_ENABLED", true)                                             \
     INT(get_dd_trace_spans_limit, "DD_TRACE_SPANS_LIMIT", 1000)                                                      \
-    BOOL(get_dd_trace_send_traces_via_thread, "DD_TRACE_BETA_SEND_TRACES_VIA_THREAD", true,                          \
-         "use background thread to send traces to the agent")                                                        \
-    BOOL(get_dd_trace_bgs_enabled, "DD_TRACE_BGS_ENABLED", true,                                                     \
-         "use background sender (BGS) to send traces to the agent")                                                  \
```
morrisonlevi, line 126: A question for the team, not so much Bob: should we emit a notice if these are set to something other than the default rather than just remove the settings?

ext/php8/configuration.c
```diff
@@ -30,8 +30,24 @@ void ddtrace_reload_config(void) {
         free(ddtrace_memoized_configuration.getter_name); \
     }                                                     \
     ddtrace_memoized_configuration.__is_set_##getter_name = FALSE;
-#define HASH(getter_name, ...)                                     \
-    zend_hash_release(ddtrace_memoized_configuration.getter_name); \
+// this will leak memory if RC != 1, but prevents crashes. This could only ever happens if an user uses
+// \dd_trace_internal_fn('ddtrace_reload_config'); at the wrong point in time - which is an internal testsuite function
+// which users are not supposed to use
+#define HASH(getter_name, ...)                                                                    \
```
bwoebi, line 36: ZAI config is supposed to improve upon this... Don't give this too much scrutiny, it works...

SammyK (commented):
I still have to review the test changes, but I think I finished reviewing all the non-test changes so far. I'll keep reviewing this tomorrow. :)

ext/php8/ddtrace.c
```diff
@@ -797,6 +832,23 @@ static PHP_FUNCTION(additional_trace_meta) {
...
 }
 
+/* {{{ proto string DDTrace\add_global_tag(string $key, string $value) */
+static PHP_FUNCTION(add_global_tag) {
+    UNUSED(execute_data);
+
+    zend_string *key, *val;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "SS", &key, &val) == FAILURE) {
+        ddtrace_log_debug(
+            "Unable to parse parameters for DDTrace\\add_global_tag; expected (string $key, string $value)");
```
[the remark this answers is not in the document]
bwoebi: yes it does :-)

bridge/_files.php
```diff
@@ -20,6 +20,7 @@
     __DIR__ . '/../src/DDTrace/StartSpanOptionsFactory.php',
     __DIR__ . '/../src/DDTrace/Time.php',
     __DIR__ . '/../src/DDTrace/Transport/Http.php',
+    __DIR__ . '/../src/DDTrace/Transport/Internal.php',
```
labbati, line 23: Is this a left-over or some magic happens here? Because I cannot find the file.
In case that case 1 holds true, we should probably have a check here to avoid that files that do not exist can be added to `_files.php`.

labbati: ^^ not as part of this PR of course

bwoebi: It's indeed a leftover

SammyK: The file does exist, but it is just not being used, correct?

SammyK (approved):
LGTM 👍 I did a best-effort reviewing the test changes and assumed that any tests named with the `-legacy` suffix were unchanged sans the `SKIPIF` sections.

Thanks for all the hard work on this one @bwoebi. I'm psyched to see this change finally land.

tests/Unit/TracerTest.php
```diff
@@ -269,24 +287,38 @@ public function testHonorGlobalTags()
 
         $this->assertSame('value1', $span->getAllTags()['key1']);
         $this->assertSame('value2', $span->getAllTags()['key2']);
+
+        \putenv('DD_TAGS='); // prevent memory leak
```
SammyK, line 291: After clicking "approve" yesterday, I started thinking about this leak a bit more and I wanted to confirm that we don't leak tags by default. The comment above the `HASH` macro, it states:

> This could only ever happens if an user uses \dd_trace_internal_fn('ddtrace_reload_config'); at the wrong point in time - which is an internal testsuite function which users are not supposed to use

But the tracer [invokes this function in the request init hook](https://github.com/DataDog/dd-trace-php/blob/d3bc9f01302a94bc94465507fc61a01968da050d/bridge/dd_init.php#L55). It also runs it from RINIT on [first-time RINIT](https://github.com/DataDog/dd-trace-php/blob/6e1dccff9d69b3e913be764ad661e3a965bd662e/ext/php8/ddtrace.c#L422).

I think that leaves us with a possible memleak at the end of the first request when `DD_TAGS` is non-empty on the happy path, correct?

SammyK (approved):
Thanks @bwoebi. I left one extra question, but this is 👍 if there is no memory leak. :)

tests/Unit/TracerTest.php
bwoebi: Not correct. It would only result in a memleak if someone accessed the DD_TAGS via `dd_trace_env_config` beforehand without releasing that variable.

SammyK: Got it. In that case, on the second invocation of `ddtrace_reload_config`, the RC == 1 and `zend_hash_release` will free the first allocation that `ddtrace_get_hash_config` made. 👍 Thanks for clarifying.

PR #1265, opened 2021-06-22
Allow for global debug mode, add zai testing to Makefile and more minor testing improvements

bwoebi:
### Description

See list of commits, in particular it will now be possible to add `export DD_TRACE_DOCKER_DEBUG=1` to your environment (I recommend saving it to `~/.profile`) on the host. This will always imply `dev` and `debug` during make. These can be disabled when needed by `use_generated` and `prod`, respectively.

### Readiness checklist
- ~~[ ] Tests added for this feature/bug.~~

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ all: $(BUILD_DIR)/configure $(SO_FILE)
 $(BUILD_DIR)/config.m4: $(M4_FILES)
 
 $(BUILD_DIR)/configure: $(BUILD_DIR)/config.m4
-	$(Q) (cd $(BUILD_DIR); phpize)
+	$(Q) (cd $(BUILD_DIR); phpize && sed -i 's/\/FAILED/\/\\bFAILED/' $(BUILD_DIR)/run-tests.php) # Fix PHP 5.4 exit code bug when running selected tests (FAILED vs XFAILED)
 
 $(BUILD_DIR)/Makefile: $(BUILD_DIR)/configure
 	$(Q) (cd $(BUILD_DIR); ./configure)
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ WALL_FLAGS := -Wall -Wextra
 EXTRA_CFLAGS :=
 CFLAGS := -O2 $(shell [ -n "${DD_TRACE_DOCKER_DEBUG}" ] && echo -O0 -g) $(EXTRA_CFLAGS) $(WALL_FLAGS)
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+PHP_EXTENSION_DIR=$(shell php -r 'print ini_get("extension_dir");')
 PHP_MAJOR_MINOR:=$(shell php -r 'echo PHP_MAJOR_VERSION . PHP_MINOR_VERSION;')
 
 VERSION:=$(shell awk -F\' '/const VERSION/ {print $$2}' < src/DDTrace/Tracer.php)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -4,24 +4,27 @@ REQUEST_INIT_HOOK_PATH := $(PROJECT_ROOT)/bridge/dd_wrap_autoloader.php
...
 
-RUN_TESTS_CMD := REPORT_EXIT_STATUS=1 TEST_PHP_SRCDIR=$(BUILD_DIR) USE_TRACKED_ALLOC=1 php -n -d 'memory_limit=-1' $(BUILD_DIR)/run-tests.php --show-diff -n -p $(shell which php) -q
+RUN_TESTS_CMD := REPORT_EXIT_STATUS=1 TEST_PHP_SRCDIR=$(PROJECT_ROOT) USE_TRACKED_ALLOC=1 php -n -d 'memory_limit=-1' $(BUILD_DIR)/run-tests.php --show-diff -n -p $(shell which php) -q
 
 C_FILES := $(shell find components ext src/dogstatsd zend_abstract_interface -name '*.c' -o -name '*.h' | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
 TEST_FILES := $(shell find tests/ext -name '*.php*' -o -name '*.inc' | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
 TEST_STUB_FILES := $(shell find tests/ext -type d -name 'stubs' -exec find '{}' -type f \; | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
 INIT_HOOK_TEST_FILES := $(shell find tests/C2PHP -name '*.phpt' -o -name '*.inc' | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
 M4_FILES := $(shell find m4 -name '*.m4*' | awk '{ printf "$(BUILD_DIR)/%s\n", $$1 }' )
+ZAI_FILES := $(shell find zend_abstract_interface -name '*.c' -o -name '*.h' -name '*.cc' -o -name 'CMakeLists.txt')
```
SammyK, line 27: I don't think this is used anywhere? If we're planning on using this, we'll probably want to also include the PHP files found in the test `stubs/` directories.

Makefile
```diff
@@ -109,23 +114,49 @@ test_c_asan: $(SO_FILE) $(TEST_FILES) $(TEST_STUB_FILES)
...
+
+build_zai:
+	( \
+	mkdir -p "$(ZAI_BUILD_DIR)"; \
+	cd $(ZAI_BUILD_DIR); \
+	CMAKE_PREFIX_PATH=/opt/catch2 cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_ZAI_TESTING=ON -DPHP_CONFIG=$(shell which php-config) $(PROJECT_ROOT)/zend_abstract_interface; \
+	$(MAKE) $(MAKEFLAGS); \
+	)
+
+test_zai: build_zai
```
SammyK, line 140: Is the plan to [eventually update the CircleCI config](https://github.com/DataDog/dd-trace-php/blob/d9ad858/.circleci/config.yml#L1072-L1133) to use these new targets?

SammyK (commented):
LGTM! 👍 Thanks for all the improvements. It looks like there are tests failing in CI that were fix in #1266 but also some ones that were canceled. I tried to restart the tests, but the option is greyed out for me in the UI. Maybe a rebase on master will fix the issues?

Makefile
bwoebi: No, leftover, removed.

Makefile
bwoebi: Not so trivially, we run these targets multiple times, on different php targets and collect the artifacts afterwards - test_zai* will always use the same location though.

SammyK (approved):

PR #1266, opened 2021-06-22
Fix tests: Pin Symfony 4 tests to laminas/laminas-code < 4.4

bwoebi:
See https://github.com/symfony/symfony/issues/41742 for details.


SammyK (approved):
Thanks for finding the fix for this!

PR #1267, opened 2021-06-25
Migrate _dd.origin & distributed header sending to internal

bwoebi:
### Description

- _dd.origin is appended to all spans meta upon serialization
- The userland curl injection has been removed in favour of a completely internal implementation. (Because _dd.origin is now not exposed anymore)
- The sampling priority metric of the root span is the source of truth for distributed tracing (and needs to be initialized early thus).

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/php8/ddtrace.c
+++ b/ext/php8/ddtrace.c
@@ -308,6 +308,32 @@ zval *ddtrace_spandata_property_metrics(ddtrace_span_t *span) { return OBJ_PROP_
 zval *ddtrace_spandata_property_exception(ddtrace_span_t *span) { return OBJ_PROP_NUM(&span->std, 6); }
 #pragma GCC diagnostic pop
 
+bool ddtrace_fetch_prioritySampling_from_root(int *priority) {
+    zval *priority_zv;
+    ddtrace_span_fci *root_span = DDTRACE_G(open_spans_top);
+
+    if (!root_span) {
+        return false;
+    }
+
+    while (root_span->next) {
+        root_span = root_span->next;
+    }
+
+    zval *root_metrics = ddtrace_spandata_property_metrics(&root_span->span);
+    ZVAL_DEREF(root_metrics);
+    if (Z_TYPE_P(root_metrics) != IS_ARRAY) {
+        return false;
+    }
+
+    if (!(priority_zv = zend_hash_str_find(Z_ARR_P(root_metrics), ZEND_STRL("_sampling_priority_v1")))) {
+        return false;
+    }
+
+    *priority = (int)zval_get_long(priority_zv);
+    return true;
+}
+
 /* DDTrace\FatalError */
 zend_class_entry *ddtrace_ce_fatal_error;
 
@@ -506,6 +532,10 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
     zval_dtor(&DDTRACE_G(additional_trace_meta));
     zend_array_destroy(DDTRACE_G(additional_global_tags));
     ZVAL_NULL(&DDTRACE_G(additional_trace_meta));
+
+    if (DDTRACE_G(dd_origin)) {
+        zend_string_release(DDTRACE_G(dd_origin));
+    }
 
     ddtrace_internal_handlers_rshutdown();
     ddtrace_dogstatsd_client_rshutdown();
@@ -1311,12 +1341,6 @@ static PHP_FUNCTION(dd_trace_set_trace_id) {
     RETURN_BOOL(0);
 }
 
-static inline void return_span_id(zval *return_value, uint64_t id) {
-    char buf[DD_TRACE_MAX_ID_LEN + 1];
-    snprintf(buf, sizeof(buf), "%" PRIu64, id);
-    RETURN_STRING(buf);
-}
-
 /* {{{ proto string dd_trace_push_span_id() */
 static PHP_FUNCTION(dd_trace_push_span_id) {
     UNUSED(execute_data);
@@ -1324,12 +1348,11 @@ static PHP_FUNCTION(dd_trace_push_span_id) {
     zval *existing_id = NULL;
     if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "|z!", &existing_id) == SUCCESS) {
         if (ddtrace_push_userland_span_id(existing_id) == true) {
-            return_span_id(return_value, ddtrace_peek_span_id());
-            return;
+            RETURN_STR(ddtrace_span_id_as_string(ddtrace_peek_span_id()));
         }
     }
 
-    return_span_id(return_value, ddtrace_push_span_id(0));
+    RETURN_STR(ddtrace_span_id_as_string(ddtrace_push_span_id(0)));
 }
 
 /* {{{ proto string dd_trace_pop_span_id() */
@@ -1343,13 +1366,13 @@ static PHP_FUNCTION(dd_trace_pop_span_id) {
         }
     }
 
-    return_span_id(return_value, id);
+    RETURN_STR(ddtrace_span_id_as_string(id));
 }
 
 /* {{{ proto string dd_trace_peek_span_id() */
 static PHP_FUNCTION(dd_trace_peek_span_id) {
     UNUSED(execute_data);
-    return_span_id(return_value, ddtrace_peek_span_id());
+    RETURN_STR(ddtrace_span_id_as_string(ddtrace_peek_span_id()));
 }
 
 /* {{{ proto string DDTrace\active_span() */
@@ -1448,7 +1471,7 @@ static PHP_FUNCTION(flush) {
 /* {{{ proto string \DDTrace\trace_id() */
 static PHP_FUNCTION(trace_id) {
     UNUSED(execute_data);
-    return_span_id(return_value, DDTRACE_G(trace_id));
+    RETURN_STR(ddtrace_span_id_as_string(DDTRACE_G(trace_id)));
 }
 
 /* {{{ proto array \DDTrace\current_context() */
@@ -1619,6 +1642,7 @@ void dd_read_distributed_tracing_ids(void) {
         }
     }
 
+    DDTRACE_G(distributed_parent_trace_id) = 0;
     if (success && zai_read_header_literal("X_DATADOG_PARENT_ID", &parent_id_str) == ZAI_HEADER_SUCCESS) {
         zval parent_zv;
         ZVAL_STR(&parent_zv, parent_id_str);
@@ -1629,5 +1653,10 @@ void dd_read_distributed_tracing_ids(void) {
                 DDTRACE_G(trace_id) = 0;
             }
         }
+    }
+
+    DDTRACE_G(dd_origin) = NULL;
+    if (zai_read_header_literal("X_DATADOG_ORIGIN", &DDTRACE_G(dd_origin)) == ZAI_HEADER_SUCCESS) {
+        GC_ADDREF(DDTRACE_G(dd_origin));
     }
 }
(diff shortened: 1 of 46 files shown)


ext/php8/ddtrace.c
```diff
@@ -311,6 +311,32 @@ zval *ddtrace_spandata_property_meta(ddtrace_span_t *span) { return OBJ_PROP_NUM
...
+    if (!root_span) {
+        return false;
+    }
+
+    while (root_span->next) {
+        root_span = root_span->next;
+    }
+
+    zval *root_metrics = ddtrace_spandata_property_metrics(&root_span->span);
+    ZVAL_DEREF(root_metrics);
```
SammyK, line 327: What scenario would cause a zval from `OBJ_PROP_NUM` to be `IS_REFERENCE`?

ext/php8/handlers_curl.c
```diff
@@ -79,54 +76,44 @@ static void dd_ch_duplicate_headers(zval *ch_orig, zval *ch_new) {
...
+    ZVAL_COPY(ZEND_CALL_ARG(call, 2), &http_header_const->value);
+    ZVAL_COPY_VALUE(ZEND_CALL_ARG(call, 3), &headers);
 
-        dd_free_headers_arg(&headers);
-    }
+    zval ret;
+    dd_curl_setopt_handler(call, &ret);
+
+    zend_vm_stack_free_args(call);
+    zend_vm_stack_free_call_frame(call);
```
SammyK, line 116: Would this be the preferred method to call all internal functions? If so, it might be a good candidate to add as an API to ZAI functions.

ext/php8/handlers_curl.c
```diff
@@ -79,54 +76,44 @@ static void dd_ch_duplicate_headers(zval *ch_orig, zval *ch_new) {
...
+    }
+    if (DDTRACE_G(dd_origin)) {
+        add_next_index_str(&headers, zend_strpprintf(0, "x-datadog-origin: %s", ZSTR_VAL(DDTRACE_G(dd_origin))));
     }
-    ZVAL_ARR(arg, retval);
-}
 
-static void dd_free_headers_arg(zval *arg) { zend_array_destroy(Z_ARRVAL_P(arg)); }
+    zend_function *setopt_fn = zend_hash_str_find_ptr(EG(function_table), ZEND_STRL("curl_setopt"));
+    zend_constant *http_header_const = zend_hash_str_find_ptr(EG(zend_constants), ZEND_STRL("CURLOPT_HTTPHEADER"));
```
SammyK, line 104: Any particular reason this needs to be fetched again for the call, or can we use the value of `dd_const_curlopt_httpheader` that is fetched on startup?

ext/php8/handlers_curl.c
```diff
@@ -79,54 +76,44 @@ static void dd_ch_duplicate_headers(zval *ch_orig, zval *ch_new) {
...
+                           zend_strpprintf(0, "x-datadog-parent-id: %" PRIu64, (DDTRACE_G(span_ids_top)->id)));
+    }
+    if (DDTRACE_G(dd_origin)) {
+        add_next_index_str(&headers, zend_strpprintf(0, "x-datadog-origin: %s", ZSTR_VAL(DDTRACE_G(dd_origin))));
     }
-    ZVAL_ARR(arg, retval);
-}
 
-static void dd_free_headers_arg(zval *arg) { zend_array_destroy(Z_ARRVAL_P(arg)); }
+    zend_function *setopt_fn = zend_hash_str_find_ptr(EG(function_table), ZEND_STRL("curl_setopt"));
```
SammyK, line 103: Would you see any value in obtaining this once on startup vs every function call?

SammyK (dismissed):
LGTM 👍 I left a few non-blocking questions. Do we know if the `randomized_tests` are failing due to flakiness vs something real?

ext/php8/handlers_curl.c
morrisonlevi: I have a tiny bit of experience here. It would need to be looked up on every request, not just once on startup. Remember the sigsegv when my curl cache wasn't cleared properly? It would be similar to that. _Most_ of the time it won't move address, but sometimes it will.

SammyK: > Remember the sigsegv when my curl cache wasn't cleared properly?

🤔 I'm not really sure which issue you're referring to. I remember one issue came about because the request-scoped global that stored the function call info cache wasn't being initialized on RINIT, but I'm assuming you're referring to a different issue?

> Most of the time it won't move address, but sometimes it will.

If it is read out of the compiler globals on startup, what would cause the address to change?

If we can't look it up at startup, perhaps we could look it up once per request? I don't think it's too big of an issue either way. Just thinking of how we can reduce the number of hash table lookups. :)

ext/php8/handlers_curl.c
bwoebi: It is only a preferred way under a very specific circumstance: we _wrap_ an userland function, which is called from an internal function and that function would be anyway called, and not because of our instrumentation. Thus we don't want exception handling etc.
I'm not sure whether it's a really good ZAI candidate, as this one is only sensible and even safe to do under limited cirucmstances. I doubt that it's a good idea to have a generic abstraction hiding function for something like this (e.g. let's suppose here were a ZAI call. You see this, are inspired by a seemingly neutral sounding function name and ... use it at the wrong place - where it would be technically fine, except exceptions are not handled). I fear this would be more footgun than helpful.

ext/php8/handlers_curl.c
bwoebi: @morrisonlevi The zval containing the Z_PTR might be moved, yes (due to function table resizing), but the container zend_function * pointer should never be different from MINIT, or am I missing something here?

ext/php8/handlers_curl.c
bwoebi: Wasn't aware this exists, will change :-)

ext/php8/ddtrace.c
bwoebi: `$someObj->theProp = &$myVar;` - trivial actually.

ext/php8/handlers_curl.c
bwoebi: I don't think so, in general I'd just use the normal zai framework to call functions, especially internal function which possibly throw (curl_setopt does not). This is a very specific case.

ext/php8/ddtrace.c
SammyK: Ah, of course. 😆 `E_CANNOT_SEE_FOREST_FOR_THE_TREES`

ext/php8/handlers_curl.c
morrisonlevi: I can't remember off the top of my head exactly what was being stored and caused a sigsegv. I do remember that it went through all tests and didn't fail until it reached customers.
I tried finding the PR but didn't find it right away; may have more luck looking through JIRA or release notes.

bwoebi: I cannot imagine it being an issue, but in that case let's play safe, the overhead of fetching the function from the function table is vanishingly tiny compared to the actual curl request.

ext/php8/ddtrace.c
```diff
@@ -308,6 +308,32 @@ zval *ddtrace_spandata_property_metrics(ddtrace_span_t *span) { return OBJ_PROP_
...
 
+bool ddtrace_fetch_prioritySampling_from_root(int *priority) {
+    zval *priority_zv;
+    ddtrace_span_fci *root_span = DDTRACE_G(open_spans_top);
+
+    if (!root_span) {
+        return false;
+    }
+
+    while (root_span->next) {
```
bwoebi, line 319: root_span will be cached in DDTRACE_G(root_span) in a later PR.

tests/AutoInstrumentation/scenarios/composer_with_ddtrace_dependency/composer.json
```diff
@@ -1,5 +1,8 @@
 {
     "require": {
         "datadog/dd-trace": "0.49.0"
+    },
+    "config": {
+        "disable-tls": true
```
morrisonlevi, line 6: Is there a reason these weren't committed before but are now? IIRC some composer command would run and disable it, but it wasn't committed. Or am I mis-remembering?

bwoebi: Looks like it, will remove from PR.

tests/Integrations/Guzzle/V6/GuzzleIntegrationTest.php
```diff
@@ -257,7 +257,7 @@ public function testLimitedTracerDistributedTracingIsPropagated()
             /** @var Tracer $tracer */
             $tracer = GlobalTracer::get();
             $tracer->setPrioritySampling(PrioritySampling::AUTO_KEEP);
-            $span = $tracer->startActiveSpan('custom')->getSpan();
+            $span = $tracer->startRootSpan('custom')->getSpan();
```
morrisonlevi, line 260: I suspect users may not be using `startRootSpan` as they apparently should (it existed prior to this PR). We set priority sampling before flushing if it's missing, but this is probably too late? Doesn't it need to get propagated to all the downstream distributed traces?

bwoebi: It's needed in the tests, but Bootstrap will do that for users normally. Nothing for them to change.

morrisonlevi: I was thinking more about people who are creating spans from scratch in queue workers. It won't be set for them, yes?

bwoebi: They shouldn't have autogenerated root spans, then it's not an issue either.

ext/php8/handlers_curl.c
```diff
@@ -79,54 +76,43 @@ static void dd_ch_duplicate_headers(zval *ch_orig, zval *ch_new) {
...
+    if (ddtrace_fetch_prioritySampling_from_root(&sampling_priority)) {
+        add_next_index_str(&headers, zend_strpprintf(0, "x-datadog-sampling-priority: %d", sampling_priority));
+    }
+    if (DDTRACE_G(trace_id)) {
+        add_next_index_str(&headers, zend_strpprintf(0, "x-datadog-trace-id: %" PRIu64, (DDTRACE_G(trace_id))));
+    }
+    if (DDTRACE_G(span_ids_top)) {
+        add_next_index_str(&headers,
+                           zend_strpprintf(0, "x-datadog-parent-id: %" PRIu64, (DDTRACE_G(span_ids_top)->id)));
+    }
```
morrisonlevi, line 98: Should we ensure that these are both sent or neither sent?

bwoebi: don't know. I don't know either if that makes a big difference too.

tests/Integrations/Guzzle/V6/GuzzleIntegrationTest.php
bwoebi: It's really only a problem with testing when we use autogenerated root spans, but that's not impacting real world usage.

ext/php8/handlers_curl.c
labbati: Only trace id it is worth sending it. At least the two chunks would be visible in our UI, even if disconnected.

Only span id would make no sense, but all the tracers I worked with, when they don't find a trace id they just ignore distributed tracing. So from the functional standpoint there is no harm (actually it is one info more when we try to debug, if span id is present it means that distributed tracing injection was not skipped, but trace id was empty).

In any case (not necessary for this PR unless you have to change other things), I would log an error message in debug mode since this is a misbehavior.

PR #1271, opened 2021-07-01
Move DDTrace\Format to src/api

labbati:
### Description

The fact that we did not move Format to the api packages caused a SEV-1 issue to our users. It makes sense that Format is in the public api package for now.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
I'd rather have added a comment to the constants classes than testing them.
Also, in general the values of the constants do not matter and would not need to be tested, only their mere existence.
But given this is only test code which is easily changed, I'll approve this PR.

labbati:
In general I agree. In our specific case it might make sense to test them to cover a couple of cases:
- in some cases (not so rarely) users use string values
- changing a value in the constant CAN cause a SEV-1 issue (e.g. sql queries not obfuscated and leaked to the UI). While testing them does not prevent this from happening (you can always change both) 😄  at least is one further protection

I'll merge it but we can re-evaluate this test in the future.

PR #1273, opened 2021-07-08
Fix debug/prod Makefile options

bwoebi:
Trivial fix for bogus behavior introduced in 6e1dccff9d69b3e913be764ad661e3a965bd662e:

```
make: Nothing to be done for 'prod'.
make: invalid option -- 'g'
make: invalid option -- 'g'
make: invalid option -- '0'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,10 +167,10 @@ sudo:
 	$(eval SUDO:=sudo)
 
 debug:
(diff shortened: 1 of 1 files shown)


labbati (approved):
It works locally, good catch 👍🏻

PR #1274, opened 2021-07-13
fix some API change in 8.1.0alpha3

remicollet:
Only fix the build, some tests are  failing
and as 8.1 API is not yet final...


bwoebi (approved):

bwoebi:
Thanks a lot for the fixes :-)

PR #1276, opened 2021-07-14
Set SECBIT_NO_SETUID_FIXUP to prevent crashes after setuid() calls

bwoebi:
### Description

Setuid() calls reset the effective capabilities of a thread to zero. This will prevent us from doing some operations. (including the setgroups() syscall where we encountered the crash.)

### Readiness checklist

_No tests added as I have no idea how to meaningfully test this, especially in a non-privileged docker container ..._

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -28,6 +28,8 @@ if test "$PHP_DDTRACE" != "no"; then
       [PHP_ADD_LIBRARY(execinfo, , EXTRA_LDFLAGS)])
   )
 
+  AC_CHECK_HEADERS([linux/securebits.h])
+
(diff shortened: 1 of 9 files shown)


SammyK (commented):
Thanks for the hard work uncovering the root cause of the issue and for the fix @bwoebi. Do you have any ideas how we might be able to test this from CI?

bwoebi:
@SammyK There now is a proper test, looks good?

SammyK (approved):
Using ext/ffi to test this was a brilliant idea. Thanks @bwoebi! 

The lang tests failures look like they are addressed in #1278 so this is 👍

PR #1278, opened 2021-07-15
Add ext/ffi to PHP builds

bwoebi:
This allows us to try sideeffect-ful syscalls in our testsuite.
Also fix CentOS 6 for now by hardcoding repository files from their vault (repository archive).

### DIFF
--- a/dockerfiles/ci/buster/php-8.0/configure_shared_ext.sh
+++ b/dockerfiles/ci/buster/php-8.0/configure_shared_ext.sh
@@ -7,9 +7,10 @@ ${PHP_SRC_DIR}/configure \
     --disable-all \
     --enable-cgi \
     --enable-fpm \
-    --enable-pcntl \
+    --enable-pcntl=shared \
     --enable-phpdbg \
     --enable-option-checking=fatal \
+    --with-ffi=shared \
     --with-fpm-user=www-data \
     --with-fpm-group=www-data \
     --without-pear \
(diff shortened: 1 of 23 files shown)


labbati (commented):
Changes look good to me, so from my perspective it is only a matter of fixing lang tests.

morrisonlevi:
As usual when bumping patch versions, some of the newly added tests fail. Usually this has to do with object id's being different. I actually wonder if for maintained PHP release versions if we should switch those to `#%d` unless they are specifically testing object ids.

dockerfiles/ci/buster/php-8.0/configure_shared_ext.sh
```diff
@@ -7,9 +7,10 @@ ${PHP_SRC_DIR}/configure \
     --disable-all \
     --enable-cgi \
     --enable-fpm \
-    --enable-pcntl \
+    --enable-pcntl=shared \
```
SammyK, line 10: The reason we aren't using this method to install ext/curl (for example) is to mimic the behavior of some platforms that just copy over the `.so` without doing a `make install`. That way we can test the scenario where the curl headers aren't installed but curl is still loaded as a shared lib.

I don't see any headers being installed for ext/pcntl or ext/ffi in their config.m4 files, so I think we should be OK to use this technique for these two exts (assuming that doesn't change in the future). But can you think of any other differences for using this technique vs copying over the `.so` file without running `make install`?

bwoebi: Exactly, in this scenario it does not make any difference. Otherwise I'd moved it into the post install step there.

SammyK (approved):

PR #1279, opened 2021-07-15
Apply original curl headers when curl_setopt_array() fails

SammyK:
### Description

This PR fixes a bug that would not apply the original curl headers to a request when `curl_setopt_array()` returns false. Although not strictly necessary, the `datadog/dd-trace-ci:php-8.0-shared-ext` container has been updated with two additional curl versions to test with the curl-related phpt tests in CI. The original bug was discovered due to a curl version upgrade, but the included test will trigger the same behavior on all curl versions.

This PR also fixes the `PHP 80 curl integration tests a shared lib` tests in CI; they [were failing and returning a successful exit code](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/6033/workflows/c9add20d-11af-422c-bc14-fb89c2e11415/jobs/501437).

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
+++ b/dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
@@ -72,7 +72,7 @@ RUN set -eux; \
         curl -L -o curl.tar.gz https://curl.se/download/curl-${curlVer}.tar.gz; \
         tar -xf curl.tar.gz && rm curl.tar.gz; \
         cd curl-${curlVer}; \
-        ./configure --with-openssl --prefix=/usr/local/src/curl/${curlVer}; \
+        ./configure --with-openssl --prefix=/opt/curl/${curlVer}; \
         make; make install; \
     done;
 
@@ -96,7 +96,7 @@ RUN set -eux; \
         for curlVer in ${CURL_VERSIONS}; \
         do \
             echo "Building ext/curl ${curlVer}..."; \
-            PKG_CONFIG_PATH=/usr/local/src/curl/${curlVer}/lib/pkgconfig/ \
+            PKG_CONFIG_PATH=/opt/curl/${curlVer}/lib/pkgconfig/ \
             ./configure; make; \
             mv ./modules/curl.so $(php-config --extension-dir)/curl-${curlVer}.so; \
             make clean; \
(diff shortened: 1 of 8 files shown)


dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
```diff
@@ -62,6 +62,20 @@ COPY --chown=circleci:circleci --from=php-debug-zts-asan $PHP_INSTALL_DIR_DEBUG_
...
+ENV CURL_VERSIONS="7.72.0 7.77.0"
+RUN set -eux; \
+    for curlVer in ${CURL_VERSIONS}; \
+    do \
+        echo "Build curl ${curlVer}..."; \
+        cd /tmp; \
+        curl -L -o curl.tar.gz https://curl.se/download/curl-${curlVer}.tar.gz; \
+        tar -xf curl.tar.gz && rm curl.tar.gz; \
+        cd curl-${curlVer}; \
+        ./configure --with-openssl --prefix=/usr/local/src/curl/${curlVer}; \
```
morrisonlevi, line 75: Installing it to `/usr/local/src` seems weird. Shouldn't this be `/opt/curl/${curlVer}` or something?

morrisonlevi (commented):
Looks good to me! Two things:
  1. I have a quibble about where curl is being installed to.
  2. More importantly, should `ddtrace_curl_setopt` also skip checking the retval == true?

bwoebi:
Regarding 2.: no. ddtrace_curl_setopt() only ever sets a single argument, so if this one fails, the headers are definitely not set. The issue is sepecific to the array function, as with that one we cannot distinguish what specific option failed to be set.

dockerfiles/ci/buster/php-8.0/configure_shared_ext.sh
```diff
@@ -7,7 +7,6 @@ ${PHP_SRC_DIR}/configure \
     --disable-all \
     --enable-cgi \
     --enable-fpm \
-    --enable-pcntl \
```
bwoebi, line 10: We should build pcntl as shared module here

bwoebi: Just write `--enable-pcntl=shared` instead, like here: https://github.com/DataDog/dd-trace-php/blob/424a46b7a082d67fc9350e325257b24fe7a0098e/dockerfiles/ci/buster/php-8.0/configure_shared_ext.sh#L10 (or just remove it, so that it won't merge conflict)

dockerfiles/ci/buster/php-8.0/Dockerfile_shared_ext
SammyK: Nice catch. Fixed in 8853028.

dockerfiles/ci/buster/php-8.0/configure_shared_ext.sh
SammyK: After rebasing on master, this is resolved.

bwoebi (approved):
Looks good to me now :-)

morrisonlevi (approved):

labbati:
Thank you @SammyK! Merging it, now!

PR #1280, opened 2021-07-16
Fix owner of files included in the distributed .tar.gz archive

labbati:
### Description

Our `.tar.gz` archive had files in it owned by user `3434:3434`, which is the `circleci` user from the docker images we use from our CI provider, including for `.` and `..` paths.

As a consequence, when installed, it changed ownership of the root directory to user `3434`.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/dockerfiles/verify_packages/docker-compose.yml
+++ b/dockerfiles/verify_packages/docker-compose.yml
@@ -58,5 +58,11 @@ services:
     depends_on:
       - request-replayer
 
+  tar_gz:
+    image: debian:buster
+    working_dir: /datadog
+    volumes:
+      - ../../:/datadog
+
   request-replayer:
     image: datadog/dd-trace-ci:php-request-replayer
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -800,6 +800,19 @@ jobs:
           name: Validate debian package
           command: << parameters.configuration >> bash ./dockerfiles/verify_packages/verify.sh
 
+  verify_tar_gz:
+    working_directory: ~/datadog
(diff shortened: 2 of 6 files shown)


dockerfiles/verify_packages/docker-compose.yml
```diff
@@ -58,5 +58,11 @@ services:
     depends_on:
       - request-replayer
 
+  tar_gz:
+    image: debian:buster
```
SammyK, line 62: To avoid running `install.sh` every time, would it make sense to run one of our existing Buster builds as root?

SammyK (approved):
Nice fix & test @labbati! 👍 The question is non-blocking. :)

dockerfiles/verify_packages/docker-compose.yml
labbati: Good point @SammyK! Originally the idea was that we do not want to use our development containers to do verification as we might 'overfix' things there. Instead we reproduce in all containers (you can check other scenarios in this very same folder) a path similar to what our users are doing to install php and the tracer.

I am totally open to changing this approach, but in case we need to discuss this as a separate topic from this PR and involving all the other scenarios we test, I guess?

labbati:
Failing tests are not related to this PR. Merging for now (I answered separately Sammy's comment).

dockerfiles/verify_packages/docker-compose.yml
SammyK: Good point. Thanks for clarifying!

PR #1287, opened 2021-07-28
Handle an additional case of capabilities being dropped within our background sender

bwoebi:
### Description

Turns out, while re-testing, the order of operations can be different and setuid() called before our MINIT. In this case the background sender will have to reset capabilities...
To test this, we just start the process with capabilities not present in effective set.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


SammyK (approved):
👍  LGTM. Thanks for the find, the fix, and the test!

PR #1293, opened 2021-08-13
Bringing ZAI config, internal spans and improved exception handling to PHP 7

bwoebi:
### Description

Basically all the features added over the last 1.5 months to PHP 8.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/php7/engine_hooks.c
```diff
@@ -57,14 +58,12 @@ static zend_op_array *_dd_compile_file(zend_file_handle *file_handle, int type)
 }
 
 static void _compile_minit(void) {
-    if (get_dd_trace_measure_compile_time()) {
-        _prev_compile_file = zend_compile_file;
-        zend_compile_file = _dd_compile_file;
-    }
+    _prev_compile_file = zend_compile_file;
+    zend_compile_file = _dd_compile_file;
```
morrisonlevi, line 62: Let's punt on it for this PR, but before this ships we should audit every zend fptr we override to make sure they all do an "are we enabled?" check before doing their work.

ext/php7/handlers_exception.c
```diff
@@ -0,0 +1,381 @@
...
+        }
+        zend_catch { has_bailout = true; }
+        zend_end_try();
+
+        DDTRACE_G(active_error).type = 0;
+    } else {
+        ddtrace_span_fci *root_span = DDTRACE_G(open_spans_top);
+        zend_object *exception;
+        zval *volatile span_exception;
+        volatile zval old_exception = {0};
```
morrisonlevi, line 244: Why are these marked volatile (`has_bailout` as well)?

ext/php7/integrations/integrations.c
```diff
@@ -26,29 +27,17 @@
...
-    {DDTRACE_INTEGRATION_PHPREDIS, "PHPREDIS", ZEND_STRL("phpredis")},
-    {DDTRACE_INTEGRATION_PREDIS, "PREDIS", ZEND_STRL("predis")},
-    {DDTRACE_INTEGRATION_SLIM, "SLIM", ZEND_STRL("slim")},
-    {DDTRACE_INTEGRATION_SYMFONY, "SYMFONY", ZEND_STRL("symfony")},
-    {DDTRACE_INTEGRATION_WEB, "WEB", ZEND_STRL("web")},
-    {DDTRACE_INTEGRATION_WORDPRESS, "WORDPRESS", ZEND_STRL("wordpress")},
-    {DDTRACE_INTEGRATION_YII, "YII", ZEND_STRL("yii")},
-    {DDTRACE_INTEGRATION_ZENDFRAMEWORK, "ZENDFRAMEWORK", ZEND_STRL("zendframework")},
-};
+#define INTEGRATION(id, int_name)                                      \
```
morrisonlevi, line 30: I was confused by what `id` and `int_name` were. Can we use `UCASE` and `lcase` or some other name?

ext/php7/integrations/integrations.h
```diff
@@ -1,41 +1,48 @@
...
+#undef INTEGRATION
 
 struct ddtrace_integration {
     ddtrace_integration_name name;
     char *name_ucase;
     char *name_lcase;
     size_t name_len;
+    bool (*is_enabled)();
+    bool (*is_analytics_enabled)();
+    double (*get_sample_rate)();
```
morrisonlevi, line 45: These should be `(*...)(void)`, I think? I'm unsure if it matters for pointers; it does matter for forward declares and I'd rather be consistent.

morrisonlevi (commented):
I'm only half-way down the scrollbar, but figured I'd leave what I have for now.

ext/php7/ddtrace_string.h
```diff
@@ -19,60 +19,12 @@ typedef struct ddtrace_string ddtrace_string;
 
 #define DDTRACE_STRING_ZVAL_L(zval_ptr, str) ZVAL_STRINGL(zval_ptr, str.ptr, str.len)
 
-inline ddtrace_string ddtrace_string_cstring_ctor(char *ptr) {
+static inline ddtrace_string ddtrace_string_cstring_ctor(char *ptr) {
```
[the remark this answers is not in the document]
bwoebi: There's a single occurrence in serializer.c - I intend to replace ddtrace_string.h by zai/string/string.h, but not as part of this PR.

ext/php7/signals.c
```diff
@@ -32,16 +32,17 @@ static stack_t ddtrace_altstack;
 static struct sigaction ddtrace_sigaction;
 
 #define MAX_STACK_SIZE 1024
+#define MIN_STACKSZ 16384
```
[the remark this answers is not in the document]
bwoebi: It is the next power of two larger than sizeof(void *) * MAX_STACK_SIZE

ext/php7/handlers_exception.c
bwoebi: because the compiler must not optimize these into registers due to the zend_try (a setjmp) later on

ext/php7/integrations/integrations.c
bwoebi: id is the identifier for code purposes (e.g. postfixed to the enum names) and int_name is the exposed name.

ext/php7/span.c
```diff
@@ -85,31 +60,69 @@ void ddtrace_open_span(ddtrace_span_fci *span_fci) {
...
     // Set the trace_id last so we have ID's on the stack
     span->trace_id = DDTRACE_G(trace_id);
     span->duration_start = _get_nanoseconds(USE_MONOTONIC_CLOCK);
-    span->pid = getpid();
     // Start time is nanoseconds from unix epoch
     // @see https://docs.datadoghq.com/api/?lang=python#send-traces
     span->start = _get_nanoseconds(USE_REALTIME_CLOCK);
 }
 
+ddtrace_span_fci *ddtrace_init_span() {
```
morrisonlevi, line 74:
```suggestion
ddtrace_span_fci *ddtrace_init_span(void) {
```

ext/php7/configuration.c
```diff
@@ -1,306 +1,147 @@
...
+                     module_number);
+    // We immediately initialize inis at MINIT, so that we can use a select few values already at minit.
+    // Note that we are not calling zai_config_rinit(), i.e. the get_...() functions will not work.
+    // This is intentional, so that places wishing to use values pre-RINIT do have to explicitly opt in by using the
+    // arduous way of accessing the decoded_value directly from zai_config_memoized_entries.
+    zai_config_first_time_rinit();
 }
 
-#define DD_INTEGRATION_ANALYTICS_SAMPLE_RATE_DEFAULT 1.0
+void ddtrace_config_first_rinit() {
```
morrisonlevi, line 110:
```suggestion
void ddtrace_config_first_rinit(void) {
```

ext/php7/ddtrace.c
```diff
@@ -340,17 +430,33 @@ static PHP_MSHUTDOWN_FUNCTION(ddtrace) {
...
     }
 
     ddtrace_engine_hooks_mshutdown();
 
+    zai_config_mshutdown();
+
     return SUCCESS;
 }
 
+static void dd_rinit_once() {
```
morrisonlevi, line 444:
```suggestion
static void dd_rinit_once(void) {
```

ext/php7/serializer.c
```diff
@@ -504,13 +512,66 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array) {
...
+        if (newline) {
+            size_t offset = newline - data;
+            msg = zend_string_truncate(msg, offset, 0);
+            ZSTR_VAL(msg)[offset] = 0;
+        }
+    }
+    return msg;
+}
+
+void ddtrace_save_active_error_to_metadata() {
```
morrisonlevi, line 548:
```suggestion
void ddtrace_save_active_error_to_metadata(void) {
```

morrisonlevi (commented):
I think these are optional since they are on the definition (not the declaration) but I prefer to be consistent. Apply them or not -- if not we'll talk about it once Sammy is back.

ext/php7/ddtrace.c
```diff
@@ -47,34 +50,58 @@
...
+        if (m->name == ddtrace_module_entry.name) {
+            ddtrace_module = module;
+        }
+        if (ddtrace_module && strcmp(m->name, "Zend OPcache") == 0) {
+            Bucket tmp = *ddtrace_module;
+            *ddtrace_module = *module;
+            *module = tmp;
+            break;
+        }
+    }
```
morrisonlevi, line 83: I'm struggling to understand this for some reason. I think it swaps the place of ddtrace and opcache in the Buckets, so their relative orders for any remaining things in the PHP lifecycle will be swapped. Yes?

ext/php7/ddtrace.c
```diff
@@ -47,34 +50,58 @@
...
+}
+#endif
+
 static int ddtrace_startup(struct _zend_extension *extension) {
+#if PHP_VERSION_ID >= 70300 && PHP_VERSION_ID < 70400
+    // Turns out with zai config we have dynamically allocated INI entries. This does not play well with PHP 7.3
+    // As of PHP 7.3 opcache stores INI entry values in SHM. However, only as of PHP 7.4 opcache delays detaching SHM.
+    // In PHP 7.3 SHM is freed in MSHUTDOWN, which may be executed before our extension, if we do not force an order.
+    // We have to sort this manually here, as opcache only registers itself as extension during zend_extension.startup.
```
morrisonlevi, line 92: > [...] opcache only registers itself as extension during zend_extension.startup.

You are referring here to the fact that opcache is a hybrid module + extension, or something else?

src/DDTrace/Tracer.php
```diff
@@ -172,18 +172,18 @@ public function startSpan($operationName, $options = [])
 
         $reference = $this->findParent($options->getReferences());
 
+        // avoid rounding errors, we only care about microsecond resolution here
+        $roundedStartTime = ($options->getStartTime() + 0.2) / 1000000;
```
morrisonlevi, line 176: I realize you didn't change this, but why are we adding `.2` here? Do you know?

tests/ext/start_span_with_all_properties.phpt
```diff
@@ -63,7 +65,7 @@ array(2) {
     ["start"]=>
     int(%d)
     ["duration"]=>
-    int(2000000%d)
+    int(200000%d)
```
morrisonlevi, line 68: Why are we losing a 0 here?

src/DDTrace/Tracer.php
bwoebi: because 48 bits of a double do not contain the 50 bits needed to make this precise

ext/php7/ddtrace.c
bwoebi: correct

tests/ext/start_span_with_all_properties.phpt
bwoebi: test is flaky with too much precision :-/

ext/php7/ddtrace.c
bwoebi: If swaps, _if_ opcache is loaded after ddtrace. And yes.

ext/php7/configuration.c
```diff
@@ -1,306 +1,147 @@
...
+bool runtime_config_first_init = false;
 
-// note: only call this if ddtrace_config_trace_enabled() returns true
-bool ddtrace_config_integration_enabled(ddtrace_string integration_str) {
-    ddtrace_integration* integration = ddtrace_get_integration_from_string(integration_str);
-    if (integration == NULL) {
-        return DD_INTEGRATION_ENABLED_DEFAULT;
+static void dd_copy_tolower(char *dst, const char *src) {
+    while (*src) {
+        *(dst++) = tolower(*(src++));
```
morrisonlevi, line 69:
```suggestion
static void dd_copy_tolower(char *restrict dst, const char *restrict src) {
    while (*src) {
        *(dst++) = tolower((unsigned char)*(src++));
```
Restrict is an optimization, but unsigned char cast is needed to avoid undefined behavior. See https://stackoverflow.com/a/21805970/538216 for an explanation.

-----

Note that this will not be an efficient lowercasing routine; gcc and clang seem to be unable to inline `tolower` and issue a function call. Would be better to use something like this instead of `tolower` (untested):

```c
static char dd_ascii_tolower(char in) {
    int adjust = (in >= 'A' && in <= 'Z') ? 'A' - 'a' : 0;
    int i = (unsigned char)in;
    return (char)(i - adjust);
}
```

Of course, I'm not sure it would matter. This will get called a few times, but only during MINIT or first-time RINIT, I think?

morrisonlevi (changes requested):

ext/php7/configuration.c
bwoebi: (not undefined as the inputs are all basic ASCII)

I guess using this `- ('A' - 'a')` will result in a more efficient asm at least.

bwoebi: https://godbolt.org/z/qdY8a1fos - not sure what I like more :-D (the gcc output for the lower one is inefficient though)

morrisonlevi: BTW, passing in length information allows it to vectorize with the one I provided (and I would guess yours as well).

morrisonlevi (dismissed):
Approved, though if we are not in a rush I would prefer that Sammy also reviews it given its importance and size.

morrisonlevi (approved):

ext/php7/span.c
```diff
@@ -8,6 +8,7 @@
 #include "configuration.h"
 #include "ddtrace.h"
 #include "dispatch.h"
+#include "engine_hooks.h"
```
SammyK, line 11: Is this being used anywhere?

```suggestion
```

tests/ext/sandbox/exceptions_in_original_call_rethrown_in_tracing_closure.phpt
```diff
@@ -1,7 +1,7 @@
 --TEST--
-Exceptions from original call rethrown in tracing closure (PHP 7)
+Exceptions from original call rethrown in tracing closure (PHP 7+)
 --SKIPIF--
-<?php if (PHP_VERSION_ID < 80000) die('skip: requires improved exception handling'); ?>
+<?php if (PHP_VERSION_ID < 70000) die('skip: requires improved exception handling'); ?>
```
SammyK, line 4:
```suggestion
```

SammyK (approved):
LGTM 👍 I left some non-blocking questions/suggestions.

PR #1298, opened 2021-08-23
Labbati/split generated

labbati:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
When the release is made, make sure there's a note about there now multiple generated files -- I think I've seen some customers copying the generated file. The note should also recommend against copying these files and rely on official builds and install processes, I think.

bridge/autoload.php
```diff
@@ -1,12 +1,51 @@
 <?php
 
+// The autoloading mechanism is explained in `docs/autoloading.md`. If this file is changed, make sure that
+// documentation is up to date.
```
SammyK, line 4: Totally optional: What do you think about moving the docs out of `docs/autoloading.md` and into the relevant parts of this script as big comment blocks? I think having the docs and the code in the same file would really make it easier to understand and make it easier to keep the docs/comments up to date moving forward.

SammyK (approved):
Thanks for explaining the complexities in detail. And the preloading + Composer tests in CI are fantastic. 💯 I left one optional suggestion but otherwise this LGTM! 👍

PR #1299, opened 2021-08-23
Fix static analysis execution on CI

labbati:
Recently static analysis started to fail because composer would install binaries, specifically phpstan binary, to
~/.config/composer/vendor/bin instead of ~/.composer/vendor/bin

In addition to fix this path, we also now use the most modern cimg/php images as recommended by circleci:
https://circleci.com/developer/images/image/cimg/php

### Readiness checklist
~- [ ] (only for Members) Changelog has been added to the release document.~
~- [ ] Tests added for this feature/bug.~

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -299,6 +299,9 @@ jobs:
       scenario:
         type: string
         default: ""
+      composer_root:
+        type: string
+        default: "~/.composer"
     working_directory: ~/datadog
     docker:
       - image: << parameters.docker_image >>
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -2000,7 +2003,8 @@ workflows:
       - static_analysis:
           requires: [ 'Prepare Code' ]
           name: "Static Analysis 80"
-          docker_image: circleci/php:8.0
+          docker_image: cimg/php:8.0
```
labbati, line 2006: We have to parametrize it because the change only applies to the composer version on PHP 8 images, not on 7.1.

bwoebi:
LGTM.

PR #1305, opened 2021-08-26
Fix resource name not registering ZF1 module name

wregis:
### Description

ZF1 allows do organize the application on modules and place controllers and actions under them. The current implementation ignores the modules and for so controllers and actions with the same name in different modules are treated as the same.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


labbati (commented):
Thank you for the PR @wregis. While the change is clear, I need to answer the following 2 questions before we can merge it:
1. Why our test suite did not fail?
2. Is `getModuleName()` always non `NULL`? Can I expect it to throw exceptions or to return anything that is non-concatenable to a string?

We will look into 1. asap, any link to docs/sources that can help up answering 2. would be greatly apprecited.

wregis:
@labbati I'm not sure why it didn't fail as the resource name was changed. Anyway I updated the testes to check that as well.

As far as I could find the `getModuleName` is present on zf1 since 1.0 but is is not clear if it will always return a value, for so I updated it to return "default" if empty as when fully using modules this is the values returned.

PR #1308, opened 2021-09-02
Add headers ZAI implementation for PHP 5

bwoebi:
### Description

Last missing ZAI component to enable internal spans on PHP 5.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


zend_abstract_interface/headers/php5/headers.c
```diff
@@ -0,0 +1,60 @@
...
+        // The alternative would be calling sapi_module_struct.register_server_variables manually, but this has an
+        // unacceptable overhead as that always computes the *whole* _SERVER array, even if we just want to access
+        // a single value.
+        zend_is_auto_global(ZEND_STRL("_SERVER") TSRMLS_CC);
+    }
+
+    zval *server_var = PG(http_globals)[TRACK_VARS_SERVER];
+    if (Z_TYPE_P(server_var) != IS_ARRAY) {
+        return ZAI_HEADER_NOT_READY;  // should be impossible to reach
+    }
```
morrisonlevi, line 30: Are we also sure this is always non-null?

zend_abstract_interface/headers/php5/headers.c
```diff
@@ -0,0 +1,60 @@
...
+    // array, which may have been tampered with from user side, if called after RINIT, or also by ext/filter if there
+    // is a default filter configured via ini. This should not impact us, but if it turns out to, we may have to
+    // optionally access filter globals in a best-effort attempt at getting the original raw headers.
+
+    // headers are present in HTTP_HEADERNAME from in the _SERVER array
+    ALLOCA_FLAG(use_heap)
+    size_t var_len = uppercase_header_name.len + sizeof("HTTP_");
+    char *var_name = do_alloca(var_len, use_heap);
+    memcpy(var_name, "HTTP_", 5);
+    memcpy(var_name + 5, uppercase_header_name.ptr, uppercase_header_name.len + 1);  // incl trailing NULL
```
morrisonlevi, line 43: The `+1` for the null here is not generally safe -- it would depend on whether the caller included the NULL or not. I think we should just write the `NULL` instead.

zend_abstract_interface/headers/php5/headers.c
```diff
@@ -0,0 +1,60 @@
+#include "../headers.h"
+
+#include <php.h>
+#include <zai_assert/zai_assert.h>
+
+zai_header_result zai_read_header(zai_string_view uppercase_header_name, zai_string_view *header_value TSRMLS_DC) {
```
morrisonlevi, line 6: Philosophical question: should we set `*header_value` to the empty string on entry?

zend_abstract_interface/headers/php5/headers.c
bwoebi: Actually, I'm not. I'll add a check for null here. under PHP7+ this is a bare zval, but not under PHP 5.

zend_abstract_interface/headers/php5/headers.c
```diff
@@ -0,0 +1,64 @@
...
+        // As long as this function is not called with tracing disabled, this should be fine.
+        //
+        // The alternative would be calling sapi_module_struct.register_server_variables manually, but this has an
+        // unacceptable overhead as that always computes the *whole* _SERVER array, even if we just want to access
+        // a single value.
+        zend_is_auto_global(ZEND_STRL("_SERVER") TSRMLS_CC);
+    }
+
+    zval *server_var = PG(http_globals)[TRACK_VARS_SERVER];
+    if (Z_TYPE_P(server_var) != IS_ARRAY) {
```
SammyK, line 30: It looks like we added the NULL check to PHP 7+ but not here. Did you mean to add that to PHP 5?

bwoebi: Yes...

zend_abstract_interface/zai_string/string.h
```diff
@@ -9,4 +11,9 @@ typedef struct zai_string_view_s {
 #define ZAI_STRL_VIEW(cstr) \
     (zai_string_view) { .len = sizeof(cstr) - 1, .ptr = (cstr) }
 
+#define ZAI_STRING_EMPTY \
+    (zai_string_view) { .len = 0, .ptr = "" }
+
+static inline bool zai_string_stuffed(zai_string_view s) { return s.ptr && s.len; }
```
morrisonlevi, line 17:
```suggestion
static inline bool zai_string_stuffed(zai_string_view s) { return s.len && s.ptr; }
```

I think it's slightly more cautious to not look at the ptr if `len` is 0.

bwoebi: I'm not doing *s.ptr, just checking whether it's non-NULL

morrisonlevi: True.

morrisonlevi (approved):

Issue #1311, opened 2021-09-07
[BUG] User-land span start time is wrong (differs from 0.63)


olsavmic:
### Bug description

Hi, 
after upgrading to v0.64 from 0.63, we see a very strange behaviour with user-land spans.

The start time is off in relation to other internal traces.

I've attached images showing the difference between v0.63 and 0.64.
----------------------------------
We do not hook/trace anything except for using the default integrations (Symfony, PDO).

Code:
```php
$callback = function () {
    sleep(10); //optional
    SimplePdoSelectQuery(...)
};

$scope = $this->tracer->startRootSpan(self::CONSUMER_ROOT_SPAN_NAME);
$callback();
$scope->close();
```
I made sure that before calling the `tracer->startRootSpan`, there is no other active span (`tracer->getActiveSpan === null`).

**ENV:**
DD_TRACE_GENERATE_ROOT_SPAN=false
DD_TRACE_AUTO_FLUSH_ENABLED=true
DD_TRACE_CLI_ENABLED=true
DD_APM_ENABLED=true

The log from DD_DEBUG=true is exactly the same:

```
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Attempting integrations load
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Loaded integration web
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Integration codeigniter not loaded. New attempts might be performed.
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Loaded integration curl
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Loaded integration eloquent
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Integration guzzle not loaded. New attempts might be performed.
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Integration laravel not loaded. New attempts might be performed.
[07-Sep-2021 11:40:01 America/New_York] [2021-09-07T11:40:01-04:00] [ddtrace] [debug] - Loaded integration mysqli
[...]

PR #1312, opened 2021-09-08
Fix span start times of userland spans

bwoebi:
### Description

Fix for #1311. We are now implicitly checking timing sanity for every ever tested span.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/Common/TracerTestTrait.php
+++ b/tests/Common/TracerTestTrait.php
@@ -258,6 +258,8 @@ trait TracerTestTrait
                     return;
                 }
 
+                $rawSpan["duration"] = (int)($rawSpan["duration"] / 1000);
+                $rawSpan["start"] = (int)($rawSpan["start"] / 1000);
 
                 if (PHP_VERSION_ID < 70000) {
                     $span = new Span(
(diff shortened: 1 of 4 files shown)


tests/Common/TracerTestTrait.php
```diff
@@ -258,6 +258,8 @@ private function parseRawDumpedTraces($rawTraces)
                     return;
                 }
 
+                $rawSpan["duration"] = (int)($rawSpan["duration"] / 1000);
+                $rawSpan["start"] = (int)($rawSpan["start"] / 1000);
```
labbati, line 262: While this change does not impact lines 294 and 295 (since the `/ 1000` has been removed there), it impacts lines 278 and 281 that now receive micro instead of nano. I don't see any correction to that. Is it expected?

labbati (changes requested):
I think a change is needed. But other than this, great job. Thanks @bwoebi

tests/Common/TracerTestTrait.php
bwoebi: That's right, this change had to be done _because_ they received nano. But micro is expected.

labbati: Ok then, this is why it was not failing. Thanks for the explanation.

labbati (approved):
LGTM!

PR #1318, opened 2021-09-10
Honor DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED in Lumen resource naming

labbati:
### Description

With this PR, we honor the value of `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED` when setting resource names.

This means that with the default value of `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED` (`true`), automatically users will have resources as `GET /actual/uri/path` instead of `GET action_name` or `GET App\Controller@action_method`.

This PR requires a **WARNING** to be added to the release notes, since resource names (thus possibly monitors) might have to be adjusted.

Users can go back to their default behavior setting `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED=false`.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Lumen/LumenIntegration.php
+++ b/src/DDTrace/Integrations/Lumen/LumenIntegration.php
@@ -86,8 +86,7 @@ class LumenIntegration extends Integration
                         $rootSpan->setTag('lumen.route.name', $routeInfo[1]['as']);
                         $resourceName = $routeInfo[1]['as'];
                     }
-
-                    if (null !== $resourceName) {
+                    if (null !== $resourceName && !\ddtrace_config_url_resource_name_enabled()) {
                         $rootSpan->setTag(
                             Tag::RESOURCE_NAME,
                             $rootSpan->getTag(Tag::HTTP_METHOD) . ' ' . $resourceName
--- a/src/DDTrace/Integrations/Lumen/LumenIntegration.php
+++ b/src/DDTrace/Integrations/Lumen/LumenIntegration.php
@@ -82,13 +82,12 @@ class LumenIntegration extends Integration
                         $action = $routeInfo[1]['uses'];
                         $rootSpan->setTag('lumen.route.action', $action);
                         $span->meta['lumen.route.action'] = $action;
-                        $resourceName = $routeInfo[1]['uses'];
                     }
                     if (isset($routeInfo[1]['as'])) {
                         $routeAlias = $routeInfo[1]['as'];
                         $rootSpan->setTag('lumen.route.name', $routeAlias);
-                        $span->meta['lumen.route.name'] = $routeAlias;
-                        $resourceName = $routeInfo[1]['as'];
+                        $span->resource = $routeAlias;
+                        $resourceName = $routeAlias;
                     }
                     if (null !== $resourceName && !\ddtrace_config_url_resource_name_enabled()) {
                         $rootSpan->setTag(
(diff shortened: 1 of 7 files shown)


src/DDTrace/Integrations/Lumen/LumenIntegration.php
```diff
@@ -86,8 +86,7 @@ function (SpanData $span, $args) use ($rootSpan, $integration, $appName) {
                         $rootSpan->setTag('lumen.route.name', $routeInfo[1]['as']);
                         $resourceName = $routeInfo[1]['as'];
                     }
-
-                    if (null !== $resourceName) {
+                    if (null !== $resourceName && !\ddtrace_config_url_resource_name_enabled()) {
```
morrisonlevi, line 89: Sanity check: it looks like this is done in `\DDTrace\trace_method('Laravel\Lumen\Application', 'handleFoundRoute', ...)`, so even with `ddtrace_config_url_resource_name_enabled() === false` there is a span. However, when following the logic I don't see the resource name getting set on this local span, only the  root span. Did I get that wrong?

labbati: Good that you asked, it is always better to double check. When this setting is false, then the default mechanism steps in, setting it to the FQMN. It is tested and proved here https://github.com/DataDog/dd-trace-php/pull/1318/files#diff-0459d560d408a062aaa303f1f068acfece07fec8200cb9e10c965333f42fd534R50-R53

morrisonlevi: IMO, the nested span should use `$resourceName` regardless of the `ddtrace_config_url_resource_name_enabled` setting. Currently, it does not appear to use this variable on the local span, only the root span, which seems like loss of info. Right?

labbati: I agree that the info is not on the nested span and I think that is "by design". The information is not lost, though, as it is stored in tags `lumen.route.action` and `lumen.route.name` of the root span, regardless of the setting https://github.com/DataDog/dd-trace-php/blob/labbati/lumen-resources/src/DDTrace/Integrations/Lumen/LumenIntegration.php#L79-L88

I think that it make sense that such information is stored in the root span because `handleFoundRoute` is just an implementation detail and I think it is there from the times when we were only able to have a hook (and generate a span) or not hooking at all.

labbati: @morrisonlevi as discussed in PM https://github.com/DataDog/dd-trace-php/pull/1318/commits/9311c13770792dd0b2e2e67f921d51c799f17354

tests/Integrations/Lumen/V5_2/CommonScenariosTest.php
```diff
@@ -109,7 +111,10 @@ public function provideSpecs()
                             'lumen_test_app',
                             'web',
                             'Laravel\Lumen\Application.handleFoundRoute'
-                        )->withExistingTagsNames([
+                        )->withExactTags([
+                            'lumen.route.action' => 'App\Http\Controllers\ExampleController@error',
```
labbati, line 115: `error` route has no name set, and to be honest I think we should leave as it is rather than adding a name so we add some diversity to use cases we test

morrisonlevi (dismissed):
It looks like you added `$routeAlias` as meta information on the `handleFoundRoute` span instead of as the resource -- was that intentional? If so then it looks good, but otherwise fix that ^_^

morrisonlevi (approved):

Issue #1317, opened 2021-09-10
[BUG] Accessing `$rootSpan->getService()` of internal span fails with exception


olsavmic:
### Bug description

**ddtrace-php v0.64.1**

Accessing `$rootSpan->getService()` of internal span fails with `Exception thrown in ddtrace's closure for ShipMonk\Console\Application::doRunCommand(): Undefined property: DDTrace\SpanData::$service`.

The problem disappears when setting `DD_TRACE_GENERATE_ROOT_SPAN=true`.

```php
trace_method(
    Application::class,
    'doRunCommand',
    function (SpanData $spanData, $args, $retval, ?Throwable $throwable): void {
        $tracer = GlobalTracer::get();
        $rootSpan = $tracer->getSafeRootSpan();
        if ($rootSpan === null) {
            return;
        }

        $rootSpan->getService();
});
```

**ENV**
```
DD_TRACE_GENERATE_ROOT_SPAN=false
```

### PHP version

PHP 8.0.10 (cli) (built: Aug 26 2021 16:09:29) ( NTS )

### Installed extensions

[PHP Modules]
bcmath
blackfire
calendar
Core
ctype
curl
date
ddtrace
dom
exif
FFI
fileinfo
filter
ftp
gettext
hash
iconv
igbinary
imagick
imap
intl
json
libxml
mbstring
mysqli
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
Phar
posix
readline
Reflection
session
shmop
SimpleXML
soap
sockets
sodium
SPL
standard
sysvmsg
sysvsem
sysvshm
tokenizer
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
blackfire
ddtrace

### OS info

NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster

bwoebi:
That's true, the object properties are not set. There's nothing meaningful, unless you set a service name yourself. So e.g. if the first line of your trace method would be `$spanData->service = "foo";`, the getService() would return `foo` too.

I'm not sure whether that's actually a bug.

olsavmic:
Root span by default takes the value from DD_SERVICE env so I'd expect that'd be the case.

But even if it's by design, it's unexpected behaviour as the getService function definition specifies that it returns string, not string|null or `@throws`.

bwoebi:
[...]

PR #1319, opened 2021-09-12
Default internal span service to DD_SERVICE

morrisonlevi:
### Description

As reported in #1317, internal spans are not defaulting their service property to the value of the environment variable `DD_SERVICE`.

I'm not sure if `dd_fcall_begin_tracing_hook` is the right place to stick this, but it worked locally on PHP 8 so I'm opening a draft PR to see how it goes.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php7/php7/engine_hooks.c
+++ b/ext/php7/php7/engine_hooks.c
@@ -507,6 +507,14 @@ static ddtrace_span_fci *dd_fcall_begin_tracing_hook(zend_execute_data *call, dd
     span_fci->dispatch = dispatch;
     ddtrace_open_span(span_fci);
 
+    // default service to env var DD_SERVICE
+    zend_string *service = get_DD_SERVICE();
+    if (service && ZSTR_LEN(service)) {
+        zval *prop_service = ddtrace_spandata_property_service(&span_fci->span);
+        // todo: do we need to copy this?
+        ZVAL_STR_COPY(prop_service, service);
+    }
+
     return span_fci;
 }
 
--- a/ext/php8/php8/engine_hooks.c
+++ b/ext/php8/php8/engine_hooks.c
@@ -378,6 +378,14 @@ static ddtrace_span_fci *dd_fcall_begin_tracing_hook(zend_execute_data *call, dd
     span_fci->dispatch = dispatch;
     ddtrace_open_span(span_fci);
 
+    // default service to env var DD_SERVICE
+    zend_string *service = get_DD_SERVICE();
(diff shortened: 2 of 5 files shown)


ext/php7/php7/engine_hooks.c
```diff
@@ -507,6 +507,14 @@ static ddtrace_span_fci *dd_fcall_begin_tracing_hook(zend_execute_data *call, dd
     span_fci->dispatch = dispatch;
     ddtrace_open_span(span_fci);
 
+    // default service to env var DD_SERVICE
+    zend_string *service = get_DD_SERVICE();
+    if (service && ZSTR_LEN(service)) {
+        zval *prop_service = ddtrace_spandata_property_service(&span_fci->span);
+        // todo: do we need to copy this?
+        ZVAL_STR_COPY(prop_service, service);
```
morrisonlevi, line 515: If I followed the macro expansion correctly, then the `get_DD_SERVICE()` doesn't add a ref, so we need to. CI also passed. Seem right to you too, @bwoebi?

bwoebi: yes, that's right. Most usages do not need a copy, hence we don't addref by default.

bwoebi:
While we're here https://github.com/DataDog/dd-trace-php/blob/ea6e724d6bb0a4d2c6308266afe8c526ea66379f/src/DDTrace/Span.php#L490-L504 should probably also be defaulted to "" if not set, i.e. `isset(...) ? ... : ""`.

Also, do we intentionally not want that behaviour with start_span()? If that's right, the PR is fine as is. Otherwise needs to be set during span opening in span.c.

morrisonlevi:
> While we're here
> 
> https://github.com/DataDog/dd-trace-php/blob/ea6e724d6bb0a4d2c6308266afe8c526ea66379f/src/DDTrace/Span.php#L490-L504
> 
> should probably also be defaulted to "" if not set, i.e. `isset(...) ? ... : ""`.

I'm not as sure on the userland stuff. If this shouldn't be `NULL`, then why don't we default the span resource to the empty string so it's never null?

> Also, do we intentionally not want that behaviour with start_span()? If that's right, the PR is fine as is. Otherwise needs to be set during span opening in span.c.

Again, not sure on userland expectations. If we did want it, then we should adjust `ddtrace_init_span` instead of adjusting `dd_fcall_begin_tracing_hook`, right?

morrisonlevi:
Closing, as Bob will cover this case in another PR.

PR #1322, opened 2021-09-16
Enhance exception reporting on Laravel 5+

morrisonlevi:
### Description

The Laravel integration now attaches error information to the root span
in more cases, such as an exception during routing. It also switches
from tracing hooks which drop spans to non-tracing hooks.

Thanks, Luca, for finding which methods should be hooked.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/Laravel/LaravelIntegration.php
+++ b/src/DDTrace/Integrations/Laravel/LaravelIntegration.php
@@ -76,12 +76,13 @@ class LaravelIntegration extends Integration
             }
         );
 
-        \DDTrace\trace_method(
+        \DDTrace\hook_method(
             'Illuminate\Routing\Router',
             'findRoute',
-            function (SpanData $span, $args, $route) use ($rootSpan, $integration) {
-                if (null === $route) {
-                    return false;
+            null,
+            function ($This, $scope, $args, $route) use ($rootSpan, $integration) {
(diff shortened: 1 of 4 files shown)


src/DDTrace/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -186,16 +184,34 @@ function () use ($rootSpan, $integration) {
...
+            }
+        );
+
+        \DDTrace\hook_method(
+            'Illuminate\Foundation\Http\Kernel',
+            'renderException',
+            function ($This, $scope, $args) use ($rootSpan) {
+                if (!$rootSpan->hasError()) {
+                    $rootSpan->setError($args[1]);
+                }
```
morrisonlevi, line 204: I have no particularly reason to wrap this in `hasError`, but I was thinking that if the error was already set then we'd want to avoid clobbering it. Feel free to remove, I don't feel strongly about this -- just wanted it here for discussion.

bwoebi: I would've wrapped it as well. Usually the first failure is the most important one, everything behind may just be part of a cascading failure. Should definitely be left in here.

labbati (approved):
Thanks @morrisonlevi

PR #1323, opened 2021-09-17
Pass artifacts to reliability env from dd-trace-php project

randomanderson:
### Description

This PR changes where the latest PHP build is defined. Instead of being defined in the reliability project, it is defined in this project (via a simple `curl` execution). This change gives two benefits:

1. If the artifact URL expires, it doesn't break the reliability build
2. If/when `dd-trace-php` is built inside of gitlab, no changes are needed in the reliability project.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,7 @@ variables:
 
 build:
   stage: build
-  image: ubuntu:18.04
+  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/base:bionic
   tags: [ "runner:main", "size:large" ]
   script:
     - echo $LATEST_URL | sed -E 's/.+_(.+)_.+/UPSTREAM_TRACER_VERSION=\1/' >> upstream.env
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Is it possible to trigger the GitLab job from the CircleCI job that produces the `.deb` without leaking sensitive info? Not asking you to do that right now, but I think that's the Right Way to do it.

labbati (dismissed):
It looks good to me. So now should we just update this link and the pipeline will start? How the release verification process changes for us?

randomanderson:
@morrisonlevi I don't think there's a way to do that. There has been talk of polling from gitlab for the dotnet library: both Gitlab and Azure jobs would start from the same commit, Gitlab would pull from azure once the artifact was ready. Separately, there's an initiative to move all builds into Gitlab so sensitive things like signing artifacts always happens in DD infra. Regardless, this PR decouples whatever that process eventually is from the reliability env.

@labbati There's a separate PR for the reliability env to make it all work but yes, the end goal would be for you to just update the url here, or run the dd-trace-php pipeline manually and pass the URL in as a variable.

labbati (approved):
Thanks @randomanderson

PR #1360, opened 2021-11-05
Fix http.url of internal root span

cataphract:
### Description

The internal root span defines (when `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED` is on) `http.url` as `$_SERVER['REQUEST_URI']`. This is incorrect according to the [RFC](https://github.com/DataDog/architecture/blob/master/rfcs/apm/integrations/trace-http-client-tagging/rfc.md) and other implementations. This PR build `http.url` as http(s)://<http_host or server_name><request uri>.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php5/serializer.c
+++ b/ext/php5/serializer.c
@@ -408,35 +408,72 @@ void ddtrace_set_global_span_properties(ddtrace_span_t *span TSRMLS_DC) {
     }
 }
 
+static smart_str dd_build_req_url(TSRMLS_D) {
+    smart_str url_str = {0};
+
+    const char *uri = NULL;
+    size_t uri_len;
+    zval *_server = PG(http_globals)[TRACK_VARS_SERVER];
+    if (Z_TYPE_P(_server) == IS_ARRAY || zend_is_auto_global(ZEND_STRL("_SERVER") TSRMLS_CC)) {
+        zval **req_uri;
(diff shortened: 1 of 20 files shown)


bwoebi:
Hey @cataphract, sorry for the delay, but now I've added a commit with some improvements:

* If not using an autogenerated rootspan, the start_span creating a rootspan will inherit the rootspan attributes. Hence it's enough to start and close a span instead of using a separate test method
* I prefer the built-in facilities for creating a simple string, hence used zend_strpprintf in favor of manual length and appending.
* I avoid an extra zval wrapper / zval allocation (on PHP 5).

Please verify whether everything still looks good to you, then I'll approve & merge.

cataphract:
@bwoebi seems good to me. The use of `zend_strpprintf` is a bit of a pessimization, but I understand you might prefer it for expressiveness.

bwoebi (dismissed):
Thanks for the PR then @cataphract and will merge it :-)

Btw. it is a minimal pessimization, but it's not "hot" code, the goal is not expressiveness (it's nice though), but rather avoiding manual length calculation, which in itself is a potential source of bugs and must not be off (and a risk to not be forgotten when ever the format is changed again).

bwoebi (approved):

Issue #1352, opened 2021-10-20
[BUG] Enhanced exception handling in Laravel causes false positives since 0.65.1


sserbin:
### Bug description
Enhanced exception handling for Laravel (https://github.com/DataDog/dd-trace-php/pull/1322) seem to have introduced false positive error traces.

https://github.com/DataDog/dd-trace-php/pull/1322/files#diff-65a7a50856c71af24f66a705d37f039a87410525d7e102df692ad7cc110f494fR198 hooks into `Kernel#renderException()` which however is invoked before application's `ExceptionHandler` which then determines if the exception is worth reporting - see https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Exceptions/Handler.php#L223

This means that exceptions like built-in exceptions `AuthenticationException` etc (as well as your custom exception types that you don't perceive as actual errors) are now being marked as error traces in apm, whereis previously they weren't. Doesn't seem like this is a useful/desired behavior for end-users. 

### PHP version
8.0.8

### Upgrading info
Upgraded from 0.64 to 0.66.0

morrisonlevi:
I've been working on this task. I'm a bit conflicted on the exceptions in the list `internalDontReport`. For instance, if there was an `AuthenticationException` and its a json response and I'm looking at the trace, wouldn't I want that stack trace and error? This would be true for any exception which leads to a 4xx or 5xx whether it's on the list, I think?

Can you give us an example of a custom exception and what HTTP status it leads to?

sserbin:
> For instance, if there was an AuthenticationException, it's a json response, and I'm looking at the trace, wouldn't I want that stack trace and error?

I don't personally think that's any useful in this particular case - `AuthenticationException` would usually be thrown when accessing auth-protected route and you're either not authenticated or your session has expired. In this case when I'm looking at the trace (for whatever reason) and I see it's 401 and I know what's happening. Besides the stack trace would often (if not always) be the same (bar the controller/route) 

> This would be true for any exception which leads to a 4xx or 5xx

5xx are a different story (but you normally wouldn't have an exception type on `dontReport` list *and* return 5xx), but for 4xx I think they won't be much useful stack traces in practice (to give a few examples of 4xx: `AuthenticationException`, `AuthorizationExcetpion`, `ValidationException`). 

[...]

PR #1361, opened 2021-11-05
Fix Laravel exceptions being over-reported

morrisonlevi:
### Description

As reported in #1352, the Laravel integration is now over-reporting exceptions. This PR reigns it back in.

The difference between before (16:39) and now (16:41) is that <500 are now not considered an exception, unless explicitly set.
image

Inside the internal span, though, it is still possible to see that there was an exception in the controller and what it looks like
image

Metrics now are correct, as error count is not impacted anymore by <500 for server requests.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
The latest commit has a known failure; I'm not sure what the name of the template is suppose to be yet. The integration still needs some tweaking as well.

docker-compose.yml
```diff
@@ -140,13 +138,15 @@ services:
...
+      - /proc/:/host/proc/:ro
+      - /sys/fs/cgroup/:/host/sys/fs/cgroup:ro
     environment:
-    - DD_API_KEY=${DATADOG_API_KEY}
-    - DD_APM_ENABLED=true
-    - DD_DOGSTATSD_NON_LOCAL_TRAFFIC=1
+      - DD_API_KEY=${DATADOG_API_KEY}
+      - DD_APM_ENABLED=true
+      - DD_DOGSTATSD_NON_LOCAL_TRAFFIC=1
+    ports:
```
labbati, line 148: I think it is more useful that we forward the real agent port, rather than the agent with a fake api key

docker-compose.yml
```diff
@@ -129,8 +129,6 @@ services:
       - DD_APM_ENABLED=true
       - DD_BIND_HOST=0.0.0.0
       - DD_API_KEY=invalid_key_but_its_ok
-    ports:
```
labbati, line 132: moved to the real agent

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -109,10 +109,12 @@ function (SpanData $span) use ($integration) {
         );
 
         \DDTrace\hook_method(
-            'Symfony\Component\HttpFoundation\Response',
-            'setStatusCode',
```
labbati, line 113: status code is properly handled automatically now by the extension

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -192,26 +194,6 @@ function ($This, $scope, $args) use ($rootSpan, $integration) {
             }
         );
 
-        \DDTrace\hook_method(
```
labbati, line 195: Most of this is done by the generic error reporting functionality

labbati:
Actually 2 tests are broken for good, have to fix those.

labbati:
Ready fo CR

morrisonlevi (commented):
From a technical perspective, this looks good to me. However, as I'm re-reading the related issue, it seems like they think Unauthorized shouldn't be marked as an error (nor any 4xx), because it's not something _they_ did: those are bad requests, and are under the control of the client/user, not them. But if I am reading the test right, the Unauthorized leads to an error.

morrisonlevi:
> But if I am reading the test right, the Unauthorized leads to an error.

I did not read it correctly. The errors are in children spans, but not the root span. 👍🏻

bwoebi (approved):
Seems to give the expected result :-)

PR #1364, opened 2021-11-12
Fix Laravel unnamed route with caching and domain specification

labbati:
### Description

When route caching is enabled in Laravel 7.x+ [routes are given a name](https://github.com/laravel/framework/blob/cada3d21c7f127ba01bb949b2fa045b7faaedbca/src/Illuminate/Routing/AbstractRouteCollection.php#L227) in any case that is randomly generated and follows two patterns:
- `generated:<random string>`: this was already handled by our instrumentation
- `some.domain.com.generated::<random string>`: when `Route::domain('some.domain.com')->group(...)` is used to register a route. We did not handle this case.

This PR adds support for this secondo case and refactors the route normalization code to make it more testable from the outside.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
Looks good to me :-)

PR #1365, opened 2021-11-12
Add small framework to stress test our internal API with bogus inputs

bwoebi:
Fixed these crashes and leaks:
- PHP_FUNCTION(dd_trace) assumed some zvals to be strings when printing the error message without checking
- span_id stack leaked if the tracer was disabled
- msgpack array writing function write_hash_table had problems with string+integer keys mixed
- use-after-free in PHP 5 if meta or metrics properties on SpanData were assigned non-array values
- hooking or tracing a function, with a name of an empty string on PHP 5 got an assertion failure
- Object of class %s could not be converted to string style exceptions could occur in our serialization code, removing object casting logic completely

I'm mostly unsure about the last point, I'm emulating it in our integration tracing hooks for now...

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php5/ddtrace.c
+++ b/ext/php5/ddtrace.c
@@ -328,10 +328,10 @@ static zval *OBJ_PROP_NUM_array_init(zend_object *obj, uint32_t offset) {
         return *zv;
     }
     if (*zv) {
-        zval **oldzv = zv;
+        zval *oldzv = *zv;
         MAKE_STD_ZVAL(*zv);
         array_init(*zv);
-        zval_ptr_dtor(oldzv);
+        zval_ptr_dtor(&oldzv);
     } else {
         MAKE_STD_ZVAL(*zv);
         array_init(*zv);
@@ -624,7 +624,6 @@ static void dd_clean_globals(TSRMLS_D) {
     ddtrace_internal_handlers_rshutdown(TSRMLS_C);
     ddtrace_dogstatsd_client_rshutdown(TSRMLS_C);
 
-    ddtrace_free_span_id_stack(TSRMLS_C);
     ddtrace_free_span_stacks(TSRMLS_C);
     ddtrace_coms_rshutdown();
 
@@ -638,6 +637,7 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
 
     if (!get_DD_TRACE_ENABLED()) {
         ddtrace_dispatch_destroy(TSRMLS_C);
+        ddtrace_free_span_id_stack(TSRMLS_C);
 
         return SUCCESS;
     }
@@ -655,6 +655,7 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
     dd_clean_globals(TSRMLS_C);
 
     ddtrace_dispatch_destroy(TSRMLS_C);
+    ddtrace_free_span_id_stack(TSRMLS_C);
 
     return SUCCESS;
 }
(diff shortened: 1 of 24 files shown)


ext/php5/dispatch.c
```diff
@@ -21,7 +21,7 @@ extern inline void ddtrace_dispatch_release(ddtrace_dispatch_t *dispatch);
 
 static bool dd_try_find_function_dispatch(HashTable *ht, zval *fname, ddtrace_dispatch_t **dispatch_ptr,
                                           HashTable **function_table) {
-    ddtrace_dispatch_t *dispatch = ddtrace_hash_find_ptr_lc(ht, Z_STRVAL_P(fname), Z_STRLEN_P(fname));
+    ddtrace_dispatch_t *dispatch = ddtrace_hash_find_ptr_lc(ht, Z_STRVAL_P(fname), Z_STRLEN_P(fname) + 1);
```
SammyK, line 24: This looks like it is allocating one extra byte for the [`zend_str_tolower_dup` call](https://github.com/DataDog/dd-trace-php/blob/d3bc9f01302a94bc94465507fc61a01968da050d/ext/php5/arrays.c#L32). What do you think about adding the `+ 1` to the [`zend_hash_find` call from within `ddtrace_hash_find_ptr`](https://github.com/DataDog/dd-trace-php/blob/d3bc9f01302a94bc94465507fc61a01968da050d/ext/php5/arrays.c#L19) to avoid copying PHP 5-isms into the ddtrace API?

tests/internal-api-stress-test.php
```diff
@@ -0,0 +1,123 @@
+<?php
```
SammyK, line 1: Was this intended to run in CI?

bwoebi: Yes and no - I'm not sure how I would best integrate this into CI, yet.

ext/php5/dispatch.c
bwoebi: I think that's going to be more confusing than not, especially given that we do not use only ddtrace_hash_ functions, but also zend_hash_. (like, the one call is using +1, but the other isn't, wtf?)

ext/php5/ddtrace.c
```diff
@@ -655,6 +655,7 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
     dd_clean_globals(TSRMLS_C);
 
     ddtrace_dispatch_destroy(TSRMLS_C);
+    ddtrace_free_span_id_stack(TSRMLS_C);
```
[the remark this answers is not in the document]
bwoebi: Yes, that's on my long term "cleanup" plan :-)

ext/php5/dispatch.c
bwoebi: Regarding the extra byte, should the code in `ddtrace_hash_find_ptr` do a -1 then?

SammyK: If this were a ZAI API, I would say for sure don't propagate PHP 5-isms to the API. But in this case I'm 👍 with adding a `- 1` to `zend_str_tolower_dup` if you find that preferable.

ext/php5/ddtrace.c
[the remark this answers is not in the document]
SammyK: Tracking this under APMPHP-661.

SammyK (approved):
👍 LGTM. Very nice game day results.

PR #1366, opened 2021-11-12
Bring priority sampling to internal

bwoebi:
### Description

Note that priority_sampling is not yet detached from the root_span, because it's not clear yet how metrics (like _dd.rule_psr) will be handled on the trace chunk level.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/DDTrace/Tracer.php
```diff
@@ -430,42 +409,26 @@ private function setPrioritySamplingFromSpan(Span $span)
...
-            return;
-        }
-
-        if ($prioritySampling !== Sampling\PrioritySampling::UNKNOWN) {
-            $rootSpan->setMetric('_sampling_priority_v1', $prioritySampling);
-        } elseif (isset($rootSpan->metrics)) { // official API does not allow removal, so...
-            unset($rootSpan->metrics['_sampling_priority_v1']);
-        }
+        set_priority_sampling($prioritySampling);
+        set_priority_sampling($prioritySampling, true);
```
SammyK, line 423: (This might be related to what you mentioned in the PR description, "priority_sampling is not yet detached from the root_span".) It looks like we have two sources of truth for the sampling priority value.

The request global `default_priority_sampling` is initially set from a distributed trace context (before there is a root span). We are assuming the root span to be the source of truth for priority sampling but if there isn't a root span, the request global is the priority sampling source of truth. 

The request global gets updated on this second call to `set_priority_sampling`. (I'm assuming this also means that if the tracer flushes the trace and starts a new trace during the same request, the sampling priority will be that of the previous trace, not the original value from the distributed tracing context.)

Are we planning on eventually migrating away from using the root span as the source of truth for priority sampling and rely exclusively on the request global for that value? If that is the case, can we reduce this to one call to `set_priority_sampling` and just set the request global with the call?

src/DDTrace/Propagators/CurlHeadersMap.php
```diff
@@ -1,71 +0,0 @@
...
-    /**
-     * {@inheritdoc}
-     */
-    public function inject(SpanContext $spanContext, &$carrier)
-    {
-        foreach ($carrier as $index => $value) {
-            if (
-                strpos($value, Propagator::DEFAULT_TRACE_ID_HEADER) === 0
-                || strpos($value, Propagator::DEFAULT_PARENT_ID_HEADER) === 0
-                || strpos($value, Propagator::DEFAULT_BAGGAGE_HEADER_PREFIX) === 0
```
SammyK, line 39: I don't think we have any tests for `DEFAULT_BAGGAGE_HEADER_PREFIX` outside of `CurlHeadersMapTest` which is being removed in this PR. I think OpenTracing headers was something that wasn't moved to the OpenTracing implementation a while back. But if we delete this, does this means we also drop support for propagating `ot-baggage-*` headers for curl calls?

ext/php8/span.c
```diff
@@ -189,10 +188,15 @@ void ddtrace_close_span(ddtrace_span_fci *span_fci) {
...
-    // In case we have root spans enabled, we need to always flush if we close that one (RSHUTDOWN)
-    if (DDTRACE_G(span_ids_top) == NULL && get_DD_TRACE_AUTO_FLUSH_ENABLED()) {
-        if (ddtrace_flush_tracer() == FAILURE) {
+    if (DDTRACE_G(span_ids_top) == NULL) {
+        // Enforce a sampling decision here
+        ddtrace_fetch_prioritySampling_from_root();
+
+        DDTRACE_G(root_span) = NULL;
+
+        // A userland span might still be open so we check the span ID stack instead of the internal span stack
```
SammyK, line 197: Is this still true?

ext/php5/priority_sampling/priority_sampling.h
```diff
@@ -0,0 +1,15 @@
+#ifndef DDTRACE_PRIORITY_SAMPLING_H
+#define DDTRACE_PRIORITY_SAMPLING_H
+
+#include "../ddtrace.h"
+
+#define DDTRACE_PRIORITY_SAMPLING_UNKNOWN (1 << 30)
+#define DDTRACE_PRIORITY_SAMPLING_UNSET ((1 << 30) + 1)
+
+static const int PRIORITY_SAMPLING_USER_KEEP = 2;
+static const int PRIORITY_SAMPLING_USER_REJECT = -1;
```
SammyK, line 10: Why are `AUTO_REJECT` and `AUTO_KEEP` [not migrated](https://github.com/DataDog/dd-trace-php/blob/d3bc9f01302a94bc94465507fc61a01968da050d/src/DDTrace/Sampling/PrioritySampling.php#L10-L14) to the extension level as well?

SammyK (commented):
I'm still in the process of reviewing this, but it is looking really good so far. I've left a few questions to try and get up-to-speed on the latest sampling priority handling.

ext/php5/priority_sampling/priority_sampling.h
bwoebi: You mean, I shall define them, but not use them?

ext/php8/span.c
bwoebi: this comment is obsolete, yes.

src/DDTrace/Propagators/CurlHeadersMap.php
bwoebi: We already have a couple versions ago (but I didn't notice) Here I'm just dropping the old code which hasn't been used since a month (PHP 5, for other versions - longer).
I'd like to note here that our internal span does not have any concept of baggage...

src/DDTrace/Tracer.php
bwoebi: I do not know yet. This is part of manual tracing discussion.
But I think that we'll need a default and an explicit decision (like "drop the sampling of this connection" like in an async connection handler where you don't want to sample everything).

CMakeLists.txt
```diff
@@ -81,23 +75,38 @@ else()
...
+  add_subdirectory(zend_abstract_interface/methods)
+endif()
+add_subdirectory(zend_abstract_interface/headers)
+add_subdirectory(zend_abstract_interface/properties)
+add_subdirectory(zend_abstract_interface/sandbox)
+add_subdirectory(zend_abstract_interface/uri_normalization)
+add_subdirectory(zend_abstract_interface/zai_string)
+add_subdirectory(zend_abstract_interface/zai_assert)
+
+if(NOT PHP_VERSION_DIRECTORY STREQUAL "php5")
```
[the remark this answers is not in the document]
bwoebi: Because there are none. I need to refactor a bit more on 5 in that case as we're referencing too many symbols (indirectly) from ddtrace.o here. I'm mainly proof-of-concepting the tests for 7 and 8 here now.

ext/php5/priority_sampling/priority_sampling.h
SammyK: This PR changes the default priority sampling value from `AUTO_KEEP` (1) to `USER_KEEP` (2). And `AUTO_REJECT` (0) to `USER_REJECT` (-1). This looks like a semantic change to me because this tells the Agent that all of the sampling decisions were explicitly set by the user even when they were not.

Looking at the existing implementation: The value generated from `dd_decide_on_sampling` was determined by the tracer so to mimic the exciting behavior, this would be `AUTO_KEEP`/`AUTO_REJECT`. And currently if a user explicitly set a `manual.keep`/`manual.drop` tag on a span, the sampling priority would be set to `USER_KEEP`/`USER_REJECT`.

So I'm curious if there was a reason on the Agent side of things to change this behavior in this PR. As far as I can tell, the original values are [still the ones we should be using here](https://github.com/DataDog/datadog-agent/blob/4b005758cdf6b5a63c9f3957ad8d82c17b9c558a/pkg/trace/sampler/sampler.go#L56-L71).

SammyK: Yes, this behavior is changing. Thanks for the link to the resource off thread, @bwoebi. 👍

ext/php8/span.c
bwoebi: Actually that's untrue - we still have a couple tests relying on that behavior - we'll clean it up once we get rid of the manual span id stack.

dockerfiles/verify_packages/centos/install.sh
```diff
@@ -17,7 +17,8 @@ yum --enablerepo=remi-php${PHP_MAJOR}${PHP_MINOR} install -y \
     php-fpm \
     php-opcache \
     php-pear \
-    mod_php
+    mod_php \
+    $([ ${PHP_MAJOR} == "5" ] && [ ${PHP_MINOR} -lt 6 ] && echo "php-json")
```
SammyK, line 21: The verification tests are still failing in CI after this change. Are the correct container images pushed up? If so, I'm assuming the json extension is being loaded as a shared lib in this case.

In either case, we'll need to do the same [`dlsym()` trick we used for curl](https://github.com/DataDog/dd-trace-php/blob/e8271f0/ext/php8/handlers_curl.c#L332-L359) to grab the missing symbol addresses.

ext/php5/ddtrace.c
```diff
@@ -1859,11 +1859,16 @@ static const zend_function_entry ddtrace_functions[] = {
     DDTRACE_SUB_NS_FE("Testing\\", trigger_error, arginfo_ddtrace_testing_trigger_error),
     DDTRACE_FE_END};
 
-zend_module_entry ddtrace_module_entry = {
-    STANDARD_MODULE_HEADER,  PHP_DDTRACE_EXTNAME,          ddtrace_functions,      PHP_MINIT(ddtrace),
-    PHP_MSHUTDOWN(ddtrace),  PHP_RINIT(ddtrace),           PHP_RSHUTDOWN(ddtrace), PHP_MINFO(ddtrace),
-    PHP_DDTRACE_VERSION,     PHP_MODULE_GLOBALS(ddtrace),  PHP_GINIT(ddtrace),     NULL,
-    ddtrace_post_deactivate, STANDARD_MODULE_PROPERTIES_EX};
+static const zend_module_dep ddtrace_module_reqs[] = {ZEND_MOD_REQUIRED("json") ZEND_MOD_END};
```
SammyK, line 1862: This is how we tried to resolve the same issue with the curl extension in #1144. We had to revert it (#1181 & #1209) and resolve the symbol addresses on startup because the extension might be loaded as a shared lib without the headers installed.

CMakeLists.txt
```diff
@@ -81,23 +75,38 @@ else()
...
+include_directories("${PHP_PREFIX_PATH}/include/php/"
+        "${PHP_PREFIX_PATH}/include/php/TSRM"
+        "${PHP_PREFIX_PATH}/include/php/Zend"
+        "${PHP_PREFIX_PATH}/include/php/ext"
+        "${PHP_PREFIX_PATH}/include/php/main"
+        "zend_abstract_interface"
+        "ext/vendor/mpack"
+        "ext/vendor/mt19937"
+        "src/dogstatsd"
+        "ext/${PHP_VERSION_DIRECTORY}")
```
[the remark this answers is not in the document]
bwoebi: But we need the ZAI sapi to test ... so we have a direct dependency on ZAI. Unless you explain to me how to properly break that up so that I can use ZAI SAPI in my priority sampling tests?

For me the hierarchy is component > ZAI > ddtrace. i.e. ZAI modules may use components, but not the opposite.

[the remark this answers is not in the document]
bwoebi: Also this really is just the include paths. which technically could be moved into the priority sampling cmake file. It doesn't link anything in which it shouldn't.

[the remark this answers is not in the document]
SammyK: I think this highlights that we need to move ZAI SAPI out of the `zend_abstract_interface/` directory so that it can be shared more easily. I vaguely remember some discussions @morrisonlevi and I had a while back about how we might support parts of the tracer that depend on both ZAI and ext, but we didn't come to any solid conclusions.

I think it might make sense to have a discussion about this one to find a good path forward.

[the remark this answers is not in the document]
bwoebi: TBH I prefer a more monolithic approach here, just to share the complexity and keep things simpler (in my eyes). I'm not an expert in cmakelists. I can read docs, have set up like 2-3 cmakelists projects over the last 3 years. But overall, if someone has a proposal, feel free to add to this doc, but I personally better don't touch it - at least not in this PR.

SammyK (approved):
I'm approving this, but I'm not convinced we have fully fixed the ext/json symbol addresses issue. To properly test it in CI we'd need to build a min install of PHP using all shared extensions (like `datadog/dd-trace-ci:php-8.0-shared-ext`) but for a PHP version before ext/json was bundled as core.

But I'll leave that to address in a follow up PR to unblock this one before the US holiday weekend. 🦃  Nice work @bwoebi!

ext/php5/ddtrace.c
bwoebi: But, this is for runtime? i.e. PHP will emit a proper warning for us and turn the tracer off then?

tests/Unit/TracerTest.php
```diff
@@ -148,21 +148,20 @@ public function testPrioritySamplingIsEarlyAssignedAndRefreshedOnInject()
     {
         $tracer = new Tracer(new DebugTransport());
         $span = $tracer->startRootSpan(self::OPERATION_NAME)->getSpan();
-        $this->assertSame(PrioritySampling::AUTO_KEEP, $tracer->getPrioritySampling());
+        $this->assertSame(PrioritySampling::USER_REJECT, $tracer->getPrioritySampling());
```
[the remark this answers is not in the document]
bwoebi: No. This is a side effect from another test which is not reset.

ext/php5/ddtrace.c
SammyK: Yes. But I think this check would be better suited at the call to `zai_json_setup_bindings()`. That way we also catch the unlikely case that the json extension is loaded but the symbols still could not be resolved. (This also makes it consistent with how the [tracer handles this same edge case](https://github.com/DataDog/dd-trace-php/blob/master/ext/php8/handlers_curl.c#L355-L357) with curl).

tests/ext/priority_sampling/001-default-sampling.phpt
```diff
@@ -0,0 +1,37 @@
...
+        if ($root->metrics["_dd.rule_psr"] === 1.0) {
+            echo "metrics[_dd.rule_psr] OK\n";
+
+            if (\DDTrace\get_priority_sampling() == \DD_TRACE_PRIORITY_SAMPLING_AUTO_KEEP) {
+                echo "\DDTrace\get_priority_sampling() OK\n";
+            } else {
+                echo "Default priority sampling changed\n";
+            }
+        } else {
+            echo "_dd.rule_pstr is missing from root span metrics\n";
```
morrisonlevi, line 24:
```suggestion
            echo "_dd.rule_psr is missing from root span metrics\n";
```

tests/ext/priority_sampling/002-user-reject.phpt
```diff
@@ -0,0 +1,38 @@
...
+        if ($root->metrics["_dd.rule_psr"] == 0) {
+            echo "metrics[_dd.rule_psr] OK\n";
+
+            if (\DDTrace\get_priority_sampling() == \DD_TRACE_PRIORITY_SAMPLING_USER_REJECT) {
+                echo "\DDTrace\get_priority_sampling() OK\n";
+            } else {
+                echo "Default priority sampling changed\n";
+            }
+        } else {
+            echo "_dd.rule_pstr is {$root->metrics["_dd.rule_psr"]}\n";
```
morrisonlevi, line 25:
```suggestion
            echo "_dd.rule_psr is {$root->metrics["_dd.rule_psr"]}\n";
```

tests/ext/priority_sampling/003-user-keep.phpt
```diff
@@ -0,0 +1,38 @@
...
+        if ($root->metrics["_dd.rule_psr"] == 1) {
+            echo "metrics[_dd.rule_psr] OK\n";
+
+            if (\DDTrace\get_priority_sampling() == \DD_TRACE_PRIORITY_SAMPLING_USER_KEEP) {
+                echo "\DDTrace\get_priority_sampling() OK\n";
+            } else {
+                echo "Default priority sampling changed\n";
+            }
+        } else {
+            echo "_dd.rule_pstr is {$root->metrics["_dd.rule_psr"]}\n";
```
morrisonlevi, line 25:
```suggestion
            echo "_dd.rule_psr is {$root->metrics["_dd.rule_psr"]}\n";
```

morrisonlevi (commented):
So far I've reviewed the ext/php5 changes and scanned the tests. I have a few minor questions on the code; nothing serious.

The bigger thing to me is that it seems like there is abstraction we could have pulled out here between PHP versions for parsing rules. I won't block the PR for this, though.

ext/php5/priority_sampling/priority_sampling.c
```diff
@@ -0,0 +1,121 @@
...
+
+        bool sampling = (double)genrand64_int64() < sample_rate * (double)~0ULL;
+
+        priority = sampling ? PRIORITY_SAMPLING_USER_KEEP : PRIORITY_SAMPLING_USER_REJECT;
+    }
+    add_assoc_long_ex(ddtrace_spandata_property_metrics(&span->span), "_sampling_priority_v1",
+                      sizeof("_sampling_priority_v1"), priority);
+}
+
+long ddtrace_fetch_prioritySampling_from_root(TSRMLS_D) {
```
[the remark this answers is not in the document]
bwoebi: yes, we will need to, but I guess not … yet?

ext/php7/ddtrace.c
```diff
@@ -1674,6 +1647,37 @@ static PHP_FUNCTION(dd_trace_compile_time_microseconds) {
     RETURN_LONG(ddtrace_compile_time_get());
 }
 
+static PHP_FUNCTION(set_priority_sampling) {
+    bool global = false;
+    zend_long priority;
+
+    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "l|b", &priority, &global) == FAILURE) {
+        ddtrace_log_debug("Expected an integer and an optional boolen");
```
morrisonlevi, line 1655:
```suggestion
        ddtrace_log_debug("Expected an integer and an optional boolean");
```

ext/php7/ddtrace.c
```diff
@@ -1674,6 +1647,37 @@ static PHP_FUNCTION(dd_trace_compile_time_microseconds) {
...
+    } else {
+        ddtrace_set_prioritySampling_on_root(priority);
+    }
+}
+
+static PHP_FUNCTION(get_priority_sampling) {
+    zend_bool global = false;
+
+    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "|b", &global) == FAILURE) {
+        ddtrace_log_debug("Expected an optional boolen");
```
morrisonlevi, line 1670:
```suggestion
        ddtrace_log_debug("Expected an optional boolean");
```

src/DDTrace/Tracer.php
```diff
@@ -430,42 +409,26 @@ private function setPrioritySamplingFromSpan(Span $span)
...
      */
     public function getPrioritySampling()
     {
-        // root span is source of truth for priority sampling (if it exists)
-        // @phpstan-ignore-next-line 'Undefined property $metrics' in an isset() check?! Must be a phpstan bug...
-        if (($rootSpan = $this->getSafeRootSpan()) && isset($rootSpan->metrics['_sampling_priority_v1'])) {
-            return $rootSpan->metrics['_sampling_priority_v1'];
-        }
-        return $this->prioritySampling;
+        return get_priority_sampling();
```
morrisonlevi, line 431: This PR adds a return type info of `int|null` to `getPrioritySampling` but `DDTrace\get_priority_sampling` is returning `int|false`. This discrepancy needs resolved.

I recommend changing to match `get_priority_sampling` to match `int|null`.

morrisonlevi (changes requested):

ext/php8/ddtrace.c
```diff
@@ -1615,6 +1588,37 @@ static PHP_FUNCTION(dd_trace_compile_time_microseconds) {
...
+    } else {
+        ddtrace_set_prioritySampling_on_root(priority);
+    }
+}
+
+static PHP_FUNCTION(get_priority_sampling) {
+    zend_bool global = false;
+
+    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "|b", &global) == FAILURE) {
+        ddtrace_log_debug("Expected an optional boolen");
```
morrisonlevi, line 1611:
```suggestion
        ddtrace_log_debug("Expected an optional boolean");
```

ext/php8/ddtrace.c
```diff
@@ -1615,6 +1588,37 @@ static PHP_FUNCTION(dd_trace_compile_time_microseconds) {
     RETURN_LONG(ddtrace_compile_time_get());
 }
 
+static PHP_FUNCTION(set_priority_sampling) {
+    bool global = false;
+    zend_long priority;
+
+    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS(), "l|b", &priority, &global) == FAILURE) {
+        ddtrace_log_debug("Expected an integer and an optional boolen");
```
morrisonlevi, line 1596:
```suggestion
        ddtrace_log_debug("Expected an integer and an optional boolean");
```

ext/php5/ddtrace.c
```diff
@@ -1762,6 +1759,40 @@ static PHP_FUNCTION(dd_trace_compile_time_microseconds) {
...
+    }
+}
+
+static PHP_FUNCTION(get_priority_sampling) {
+    UNUSED(return_value_used, this_ptr, return_value_ptr);
+    zend_bool global = false;
+
+    if (zend_parse_parameters_ex(ZEND_PARSE_PARAMS_QUIET, ZEND_NUM_ARGS() TSRMLS_CC, "|b", &global) == FAILURE) {
+        ddtrace_log_debug("Expected an optional boolen");
+        RETURN_FALSE;
```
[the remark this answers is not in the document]
morrisonlevi: The prototype for the callers is `int|null`, which doubly makes me think we should return null instead of false here. I have another review comment about it.

src/DDTrace/Tracer.php
krakjoe: Done in 464df10fc19fd78ab0ccb0a5cae63191a5e722f2

ext/php5/priority_sampling/priority_sampling.c
```diff
@@ -0,0 +1,127 @@
...
+    if (priority == DDTRACE_PRIORITY_SAMPLING_UNKNOWN) {
+        double sample_rate = get_DD_TRACE_SAMPLE_RATE();
+        bool explicit_rule = zai_config_memoized_entries[DDTRACE_CONFIG_DD_TRACE_SAMPLE_RATE].name_index >= 0;
+        HashPosition pos;
+        HashTable *rules = get_DD_TRACE_SAMPLING_RULES();
+        zval **rule;
+        for (zend_hash_internal_pointer_reset_ex(rules, &pos);
+             zend_hash_get_current_data_ex(rules, (void **)&rule, &pos) == SUCCESS;
+             zend_hash_move_forward_ex(rules, &pos)) {
+            if (Z_TYPE_PP(rule) != IS_ARRAY) {
```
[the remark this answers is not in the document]
krakjoe: They're json decoded from config and the shape is not verified on decode.

morrisonlevi (dismissed):
Thanks for cleaning that up, Joe!

krakjoe (approved):

PR #1367, opened 2021-11-12
implementation of parent property access

krakjoe:
Implementation of access to SpanData::$parent

### Readiness checklist

### Todo
- [ ] disallow indirect modification?
- [ ] php5

### DIFF
--- a/tests/ext/span_with_parent_property_access.phpt
+++ b/tests/ext/span_with_parent_property_access.phpt
@@ -1,0 +1,27 @@
+--TEST--
+Test DDTrace\SpanData::$parent
+--FILE--
+<?php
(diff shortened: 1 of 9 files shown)


tests/ext/span_with_parent_property_access.phpt
```diff
@@ -0,0 +1,27 @@
...
+        print $error->getMessage() . PHP_EOL;
+    } finally {
+        if ($span->parent === null) {
+            echo "FAIL\n";
+        }
+    }
+} else {
+    var_dump($root, $span);
+}
+?>
```
bwoebi, line 24: Also add `var_dump($root->parent)`, should be null.

PR #1369, opened 2021-11-15
SpanData::$parent property

krakjoe:
Implement access to SpanData::$parent.

  - This does not yet use ZEND_ACC_READONLY to keep the source simple (php9 will use ZEND_ACC_READONLY)
  - This does not prohibit indirect modification, that would be costly to all other reads and doesn't seem worth it

### DIFF
--- a/ext/php7/ddtrace.c
+++ b/ext/php7/ddtrace.c
@@ -282,6 +282,28 @@ static zend_object *ddtrace_span_data_clone_obj(zval *old_zv) {
     return new_obj;
 }
 
+#if PHP_VERSION_ID >= 70400
+static zval* ddtrace_span_data_readonly(zval *object, zval *member, zval *value, void **cache_slot) {
+#else
+static void ddtrace_span_data_readonly(zval *object, zval *member, zval *value, void **cache_slot) {
+#endif
+    if (Z_TYPE_P(member) == IS_STRING && zend_string_equals_literal(Z_STR_P(member), "parent")) {
+        zend_throw_error(zend_ce_error,
+            "Cannot modify readonly property %s::%s", ZSTR_VAL(Z_OBJCE_P(object)->name), Z_STRVAL_P(member));
+#if PHP_VERSION_ID >= 70400
+        return &EG(uninitialized_zval);
+#else
+        return;
+#endif
+    }
+
+#if PHP_VERSION_ID >= 70400
+    return zend_std_write_property(object, member, value, cache_slot);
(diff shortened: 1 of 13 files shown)


ext/php7/ddtrace.c
```diff
@@ -282,6 +282,28 @@ static zend_object *ddtrace_span_data_clone_obj(zval *old_zv) {
     return new_obj;
 }
 
+#if PHP_VERSION_ID >= 70400
+static zval* ddtrace_span_data_readonly(zval *object, zval *member, zval *value, void **cache_slot) {
+#else
+static void ddtrace_span_data_readonly(zval *object, zval *member, zval *value, void **cache_slot) {
+#endif
+    if (Z_TYPE_P(member) == IS_STRING && zend_string_equals_literal(Z_STR_P(member), "parent")) {
+        zend_throw_exception_ex(zend_ce_exception, 0,
```
bwoebi, line 291: Is there a reason why we use exception here? For PHP 5, sure, but in general Error is appropriate I'd say?

tests/ext/span_with_parent_property_access.phpt
```diff
@@ -15,10 +15,6 @@ if (!isset($root->parent) && $span->parent === $root) {
         $span->parent = null;
     } catch (\Exception $error) {
         print $error->getMessage() . PHP_EOL;
-    } finally {
-        if ($span->parent === null) {
```
bwoebi: btw. you can keep that if, given that the catch block doesn't rethrow anyway, just dropping finally {}

ext/php7/ddtrace.c
krakjoe: It started out Error, but the test is for all versions, and the only compatible test catches Exception.

tests/ext/span_with_parent_property_access.phpt
krakjoe: 5.5 is first version with finally {}

ext/php7/ddtrace.c
bwoebi: you could do `} catch (\Exception) { print $error->getMessage() . PHP_EOL; } catch (\Throwable $e) { print $error->getMessage() . PHP_EOL; }` in the test

krakjoe: I'd like to have a conversation about this.

bwoebi (approved):
Looks good to me now, thanks Joe :-)

PR #1371, opened 2021-11-16
Bump cmake and catch2 versions on Debian images

morrisonlevi:
### Description

Also build with CATCH_BUILD_STATIC_LIBRARY=on, which will speed up
the component builds a bit because catch2_main (which is slow to build)
will be replaced by this one, which is built into the images.

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the release document.
- [ ] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1372, opened 2021-11-16
Add profiler sources, components, and tests

morrisonlevi:
### Description

This PR adds profiler sources, components, and some tests. It adds build option `--enable-datadog-profiling`, but nothing in CI builds nor tests it yet; this will come in another PR (this one is already large at ~4k additions). However, the components for it are being built and tested in CI.

This PR delete's our `catch2_main` target in favor of Catch2's `Catch2::Catch2WithMain` target, and adds a polyfill. This speeds up builds when Catch2 was built with this target (which I have a PR for #1371).

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


.circleci/config.yml
```diff
@@ -1066,6 +1066,9 @@ jobs:
           name: Build and test Datadog::Php::Components (ASan)
           command: |
             export PATH="/opt/cmake/<< parameters.cmake_version >>/bin:$PATH"
+            if [ -f "/opt/libuv/lib/pkgconfig/libuv.pc" ] ; then
```
labbati, line 1069: It is not clear to me when the file is expected not to be present. In other words, it is ok to check or we should enforce existence of the file always when we build?

.circleci/config.yml
```diff
@@ -1097,7 +1100,13 @@ jobs:
...
               ~/datadog/components
             make -j all
-            make test
+            ctest --output-on-failure
+
+            # channel is non-deterministic, so run tests a few more times. At
+            # the moment, Catch2 tests are not automatically adding labels, so
+            # run all tests:
+            # https://github.com/catchorg/Catch2/issues/1590
+            ctest --output-on-failure --repeat until-fail:9
```
labbati, line 1109: 2 questions here.

1. is line 1103 a leftover?
2. `make test` used to add the `ctest`'s [flag](https://cmake.org/cmake/help/latest/manual/ctest.1.html#:~:text=--force--,new,--ctest-process) `--force-new-ctest-process`. Should we honor that?

Looking at the generated `Makefile`

```
test:
        @$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --cyan "Running tests..."
        /opt/cmake/3.20.1/bin/ctest --force-new-ctest-process $(ARGS)
```
it seems to me that you could obtain the same by `make test ARGS="--output-on-failure --repeat until-fail:9"`

labbati (commented):
I added a few minor comments only to CI integration. Thanks @morrisonlevi for the work on this.

.circleci/config.yml
morrisonlevi: At the moment it only exists on Debian, because I've only updated those images. Eventually it should always exist, or else our build images are bad.

.circleci/config.yml
morrisonlevi: Updated as per your feedback.

.circleci/config.yml
labbati: thanks 👍🏻

morrisonlevi:
I split the changes to `components/` into another PR: #1375.

PR #1373, opened 2021-11-17
PDOIntegration::parseDSN fails to parse some DSN

krakjoe:
### PDOIntegration::parseDSN

This was found to be faulty for various flavours of PDO.

Essentially some keys (discovered from manual entries) are not supported in DSN key value pairs, in addition a driver is allowed to treat keys case insensitively, so that Host=localhost, and host=localhost and HOST=localhost may be treated the same, so we should also ignore key case.

### DIFF
--- a/src/DDTrace/Integrations/PDO/PDOIntegration.php
+++ b/src/DDTrace/Integrations/PDO/PDOIntegration.php
@@ -156,13 +156,17 @@ class PDOIntegration extends Integration
                 continue;
             }
             list($key, $value) = explode('=', $valString);
-            switch ($key) {
+            switch (strtolower($key)) {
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
Thanks for that change making us compatible with more PDO drivers :-)

SammyK:
Thanks for expanding the parsing @krakjoe! Would it be possible to add a few tests for the new parsing rules? :)

krakjoe:
@SammyK we appear to test only MySQL PDO driver, I think more generally we should improve testing of PDO, so will create a task for that - it looks like, we will have to have (all/some) other drivers available in testing.

I'm going to go ahead and merge this as-is.

PR #1375, opened 2021-11-18
Merge profiling components

morrisonlevi:
### Description

This merges the profiling components and gets them running in CI. They
aren't used yet, so I haven't modified config.m4.

This will make the code review for the upcoming profiling PR easier by
reducing the amount of code looked at it in each PR.

Also rename `BUILD_COMPONENTS_TESTING` to `DATADOG_PHP_TESTING`.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/components/channel/channel.c
+++ b/components/channel/channel.c
@@ -1,0 +1,178 @@
+#include "channel.h"
+
+#include <components/queue/queue.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <uv.h>
+
+struct datadog_php_channel_impl_s {
+    datadog_php_queue queue;
+
+    // enforce single consumer (refcount should be 0 or 1)
+    uint32_t receiver_count;
+
+    /* Multiple producers are fine. When there aren't any items in the queue, the
+     * consumer will use this info to determine whether to try waiting for more
+     * data or to return empty handed. If there is a known producer, it will wait;
+     * otherwise it will return without waiting.
+     */
+    uint32_t sender_count;
+
+    uv_mutex_t mutex;
+    uv_cond_t condvar;
+    void *buffer[];
+};
+
+/**
+ * Destroys the channel. _NOT_ thread safe.
+ */
+static void channel_dtor(datadog_php_channel_impl *channel) {
+    uv_mutex_destroy(&channel->mutex);
+    uv_cond_destroy(&channel->condvar);
+    free(channel);
+}
+
+static bool receiver_recv(struct datadog_php_receiver_s *receiver, void **data, uint64_t timeout_nanos) {
+    if (!receiver || !receiver->channel) {
+        return false;
+    }
+
+    datadog_php_channel_impl *channel = receiver->channel;
+    uv_mutex_lock(&channel->mutex);
+    datadog_php_queue *queue = &channel->queue;
+    bool succeeded = queue->try_pop(queue, data);
+
+    // If there wasn't an item but there are known producers, wait and try again
+    if (!succeeded && channel->sender_count && timeout_nanos) {
+        uint64_t now = uv_hrtime();
+        uint64_t timeout_target = now + timeout_nanos;
+        do {
+            // Handle being signalled, waking up spuriously, and timing out the same.
+            (void)uv_cond_timedwait(&channel->condvar, &channel->mutex, timeout_target - now);
+            succeeded = queue->try_pop(queue, data);
+            if (succeeded) {
+                break;
+            }
+
+            now = uv_hrtime();
+        } while (channel->sender_count && now < timeout_target);
+    }
+
+    uv_mutex_unlock(&channel->mutex);
+    return succeeded;
+}
+
+static void receiver_dtor(struct datadog_php_receiver_s *receiver) {
+    if (receiver && receiver->channel) {
+        datadog_php_channel_impl *channel = receiver->channel;
+        receiver->channel = NULL;
+        uv_mutex_lock(&channel->mutex);
+        // todo: underflow checks
+        bool dtor = !(--channel->receiver_count | channel->sender_count);
+        uv_mutex_unlock(&channel->mutex);
+
+        // Cannot dtor the mutex while it is held.
+        if (dtor) {
+            channel_dtor(channel);
+        }
+    }
+}
+
+static bool sender_send(struct datadog_php_sender_s *sender, void *data) {
+    if (!sender || !sender->channel) {
+        return false;
+    }
+
+    datadog_php_channel_impl *channel = sender->channel;
+    uv_mutex_lock(&channel->mutex);
+    datadog_php_queue *queue = &channel->queue;
+    bool sent = queue->try_push(queue, data);
+    if (sent) {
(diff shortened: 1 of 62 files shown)


components/sapi/sapi.c
```diff
@@ -29,3 +33,5 @@ sapi_t datadog_php_sapi_from_name(string_view_t module) {
 
     return DATADOG_PHP_SAPI_UNKNOWN;
 }
+
+datadog_php_sapi datadog_php_sapi_detect(datadog_php_string_view module) { return datadog_php_sapi_from_name(module); }
```
SammyK, line 37: Is this an alias for the profiler to use?

SammyK (commented):
I'm still reviewing this, but so far it looks great. I'm hoping that we can start using that there log component in the tracer soon. 🪵

components/sapi/sapi.c
morrisonlevi: It's a difference I haven't unified, yeah.

components/channel/channel.c
```diff
@@ -0,0 +1,178 @@
...
+    bool succeeded = queue->try_pop(queue, data);
+
+    // If there wasn't an item but there are known producers, wait and try again
+    if (!succeeded && channel->sender_count && timeout_nanos) {
+        uint64_t now = uv_hrtime();
+        uint64_t timeout_target = now + timeout_nanos;
+        do {
+            // Handle being signalled, waking up spuriously, and timing out the same.
+            (void)uv_cond_timedwait(&channel->condvar, &channel->mutex, timeout_target - now);
+            succeeded = queue->try_pop(queue, data);
```
bwoebi, line 53: How is this ever supposed to succeed, given that we hold a lock here? Preventing anyone from modifying the queue (sender_send being properly locked).

bwoebi: In fact, as we have the assumption that there will only be one receiver at most, we can first check for availability on the queue _without_ holding a lock. Checking `queue->size` and only acquiring a lock when actually popping from the queue.

components/log/log.c
```diff
@@ -0,0 +1,159 @@
...
+    size_t written = 0;
+    unsigned tries = 0, max_tries = 3;  // arbitrarily chosen
+    while (written < message.len && tries++ < max_tries) {
+        ssize_t n = write(logger->descriptor, message.ptr + written, message.len - written);
+        if (n < 0) {
+            /* POSIX.1-2001 allows either error to be returned for this case, and does
+             * not require these constants to have the same value, so a portable
+             * application should check for both possibilities.
+             */
+            if (errno == EAGAIN || errno == EWOULDBLOCK || errno == EINTR) {
```
bwoebi, line 59: EWOULDBLOCK? Is the logging descriptor supposed to be non-blocking?

components/log/log.h
```diff
@@ -0,0 +1,113 @@
...
+ * If the `logger` is set to at least `level`, then the messages will be logged.
+ * Otherwise they will be ignored.
+ * @param logger
+ * @param level
+ * @param n_messages
+ * @param messages
+ * @return Bytes written
+ */
+int64_t datadog_php_logv(datadog_php_logger *logger, datadog_php_log_level level, size_t n_messages,
+                         datadog_php_string_view messages[C_STATIC(n_messages)]);
```
bwoebi, line 85: I'm not sure what to make of this particular design choice to merely concatenate messages?
Is this a tactic to avoid constructing a big buffer?
I'd prefer a trivial printf-style functionality (maybe additionally). This also alleviates us the need from having to pre-format any non-string values (numbers ...) in case the log level is too low for the message to be logged at all.

components/log/log.c
```diff
@@ -0,0 +1,159 @@
...
+}
+
+bool datadog_php_logger_valid(datadog_php_logger *logger) {
+    bool valid = false;
+    if (logger->mutex && pthread_mutex_lock(logger->mutex) == 0) {
+        valid = logger_valid(logger);
+        (void)pthread_mutex_unlock(logger->mutex);
+    }
+    return valid;
+}
```
bwoebi, line 33: This API does not make much sense. Either you want - as a caller - lock yourself and then do some log operation. Or you generally don't care about the thread safety of this.
But here, if you were to mutex this, as a caller, and then use this function, you'll deadlock yourself.

components/channel/channel.c
morrisonlevi: The docs for `uv_cond_timedwait` are a [bit sparse](http://docs.libuv.org/en/v1.x/threading.html#c.uv_cond_timedwait) but as you can imagine it calls `pthread_cond_timedwait` which [is documented](https://linux.die.net/man/3/pthread_cond_wait):

> The pthread_cond_timedwait() and pthread_cond_wait() functions shall block on a condition variable. They shall be called with mutex locked by the calling thread or undefined behavior results.
>
> These functions atomically release mutex and cause the calling thread to block on the condition variable cond; atomically here means "atomically with respect to access by another thread to the mutex and then the condition variable".

So the lock is released when it waits, and re-acquired when it wakes.

components/log/log.h
morrisonlevi: Almost all current messages are literals or literals plus a string generated by some other system (like `strerror`). This design avoids creating heap-allocated buffers just to call the log function (which may not log the message). So, yeah, this was basically the minimum design.

The design can improve, of course.

components/log/log.c
morrisonlevi: I didn't understand your feedback, but I do agree it's a poor function and can be removed because the profiler only uses is in these ways:

 1. Detect that initialization failed. We can make initialization return this info somehow.
 2. Avoid double-dtors. We can handle this differently.
 
 Sound good?

components/log/log.c
morrisonlevi: The comment tries to explain that `EAGAIN` and `EWOULDBLOCK` are actually the same constant on many platforms, but this isn't guaranteed, so check for both. Since this is doing `write` on a potentially arbitrary socket, we just handle things in a basic way with immediate retry mechanism. There are certainly cases it will not handle optimally, but that's fine.

This is another case of "good enough for where I am." Long-term, I think we should be sending Datadog's product logs to Datadog through a separate stream from user logs. Users have legitimately complained that sometimes dd-trace-php breaks their logs because of an unexpected format or bloat their log costs.

bwoebi: Okay, still feels bit weird, but good enough :-)

components/log/log.h
bwoebi: Yeah, no blocker here, mainly wondering and suggesting.

components/log/log.c
bwoebi: Sounds good to me.

morrisonlevi:
Note that the component `uuid` conflicts with libuuid, which has an include `<uuid/uuid.h>`. Historically I haven't hit this since we don't use libuuid at all, but on Mac OS X `pwd.h` which is included by `uv.h` will end up including libuuid!

The options I see are:

 1. Rename the component to something else.
 2. Require includes to use the `components/` prefix e.g. `#include <components/uuid/uuid.h>`.
 
 What do you think?

PR #1378, opened 2021-11-24
ZAI: JSON

krakjoe:
### ZAI: JSON

Abstract interface for JSON, with tests.

Split from in #1366

### DIFF
--- a/zend_abstract_interface/json/json.h
+++ b/zend_abstract_interface/json/json.h
@@ -1,0 +1,44 @@
+#ifndef ZAI_JSON_H
+#define ZAI_JSON_H
+
+#include "php.h"
+
+#define php_json_encode _dd_json_encode
+#define php_json_decode_ex _dd_json_decode_ex
+
+#define PHP_JSON_OBJECT_AS_ARRAY (1 << 0)
+
+#if PHP_VERSION_ID < 70000
+#include "ext/standard/php_smart_str.h"
+
+extern void (*php_json_encode)(smart_str *buf, zval *val, int options TSRMLS_DC);
+extern void (*php_json_decode_ex)(zval *return_value, char *str, int str_len, int options, long depth TSRMLS_DC);
+
+static inline void php_json_decode(zval *return_value, char *str, int str_len, zend_bool assoc, long depth TSRMLS_DC) {
+    php_json_decode_ex(return_value, str, str_len, assoc ? PHP_JSON_OBJECT_AS_ARRAY : 0, depth TSRMLS_CC);
+}
+#define smart_str_length(str) (str)->len
+#define smart_str_value(str) ((char *)(str)->c)
+#else
+#include <stdbool.h>
+
+#include "zend_smart_str.h"
+
+extern int (*php_json_encode)(smart_str *buf, zval *val, int options);
+extern int (*php_json_decode_ex)(zval *return_value, const char *str, size_t str_len, zend_long options,
+                                 zend_long depth);
+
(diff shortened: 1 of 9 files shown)


zend_abstract_interface/json/json.h
```diff
@@ -0,0 +1,44 @@
...
+
+extern int (*php_json_encode)(smart_str *buf, zval *val, int options);
+extern int (*php_json_decode_ex)(zval *return_value, const char *str, size_t str_len, zend_long options,
+                                 zend_long depth);
+
+static inline int php_json_decode(zval *return_value, const char *str, int str_len, bool assoc, zend_long depth) {
+    return php_json_decode_ex(return_value, str, str_len, assoc ? PHP_JSON_OBJECT_AS_ARRAY : 0, depth);
+}
+#define smart_str_length(str) (str)->s->len
+#define smart_str_value(str) ((char *)(str)->s->val)
```
bwoebi, line 35: Let's not define such a generic macro in unrelated header files. These are just needed for the test, right? Then these defs should also live in the test.

krakjoe: They are needed for the test, but also needed for anything that uses the zai_json interface, so defining them in the test doesn't make much sense in that light, we don't want the consumer of the api to have to define them in their code.

I'm not sure why the zend_long undef is even there, it's copy pasta ?

bwoebi: Not really. On PHP 7/8 you typically are just going to use ZSTR_LEN(str->s) - this would only matter for version independent access - and in general these macros fit better in compatibilty.h if needed. Let's put them in tests only for now.

bwoebi: zend_long undef is leftover by me, shouldn't be there anymore.

krakjoe: Okay, I think they fit in compat header, but the comment in compat header is confusing. It says they should only be used in headers or tests, but call sites (consumer of abstract api) need also to use them.

Maybe we should have an abstract interface for smart_str, but what do we gain from an abstract interface to json if around the callsite the consumer is forced to ifdef anyway ?

I think the interface needs to include everything you reasonably need to consume it.

bwoebi: Possibly we should, once the need arises. For current non-test consumers there's no need.

zend_abstract_interface/zai_compat.h
```diff
@@ -1,17 +1,20 @@
 #ifndef ZAI_COMPAT_H
 #define ZAI_COMPAT_H
 
-// macros to be used only within ZAI headers or tests to avoid duplication of code lines
```
krakjoe, line 4: Any callsite would need to use TSRMLS_ junk at the callsite, so it's not true that these must only be used in tests.

zend_abstract_interface/zai_compat.h
```diff
@@ -1,17 +1,20 @@
...
 #if PHP_VERSION_ID >= 70000
 #define ZAI_TSRMLS_C
 #define ZAI_TSRMLS_CC
 #define ZAI_TSRMLS_D
 #define ZAI_TSRMLS_DC
+#define smart_str_length(str) (str)->s->len
```
krakjoe, line 9: I don't love this, and neither does bob. But the point here is to keep seems consistent, the value of that is reduced if the consumer of any API must ifdef around the call site.

Should we decide to use smart_str more widely than we do now, maybe a zai_smart_str is a path we could follow.

This looks like the bear minimum to keep seams tidy.

bwoebi (approved):

PR #1379, opened 2021-11-24
Generate PECL file lists at packaging time

bwoebi:
Also adding Joe as maintainer.

Makes our lives a bit easier by not having to manually update package.xml for every new source file or test.


krakjoe (approved):

SammyK:
Nice! 💯 I just added Joe as [maintainer in PECL](https://pecl.php.net/package/datadog_trace) as well.

PR #1380, opened 2021-11-24
ZAI: config/json

krakjoe:
### ZAI: config/json

Used by #1366

### DIFF
--- a/zend_abstract_interface/zai_compat.h
+++ b/zend_abstract_interface/zai_compat.h
@@ -6,6 +6,8 @@
 #define ZAI_TSRMLS_CC
 #define ZAI_TSRMLS_D
 #define ZAI_TSRMLS_DC
+#define ZAI_TSRMLS_FETCH()
+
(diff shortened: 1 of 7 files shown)


zend_abstract_interface/zai_compat.h
```diff
@@ -6,13 +6,17 @@
 #define ZAI_TSRMLS_CC
 #define ZAI_TSRMLS_D
 #define ZAI_TSRMLS_DC
+#define ZAI_TSRMLS_FETCH()
```
krakjoe, line 9: We can slip this in, it's not worth a PR on it's own ...

bwoebi:
Diffing with priority_sampling branch says:

```
#undef zval_internal_ptr_dtor
#define zval_internal_ptr_dtor zval_internal_dtor
```

has been removed from the config_decode.c file

bwoebi:
The real problem here is that json.c is not included in config.m4 for PHP 5.

PR #1381, opened 2021-11-24
ZAI: json - fix php5 build

krakjoe:
We need to name these to avoid clashes/relocations (which crash alpine).

### DIFF
--- a/config.m4
+++ b/config.m4
@@ -97,6 +97,7 @@ if test "$PHP_DDTRACE" != "no"; then
       zend_abstract_interface/exceptions/php5/exceptions.c \
       zend_abstract_interface/functions/php5/functions.c \
       zend_abstract_interface/headers/php5/headers.c \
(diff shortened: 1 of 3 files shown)


krakjoe:
@bwoebi I missed the PHP5 problem, but the motivation here (beyond possible relocation errors, which I'm not sure of) is to make the seam clear; It should be obvious that we're providing php_json_* functions as part of ZAI - we should be free to provide an API on top if our usage beyond the current proposed usage converge in one direction where a different API may make sense. It should also be clear what we are actually testing under test.

bwoebi:
I found it more obvious to just have the same name here. Indicating that it's an 1:1 mirror of the PHP API. Thus I'd prefer not to do this change.

krakjoe:
@bwoebi okay, let's just fix the build here ...

bwoebi (dismissed):

bwoebi (approved):

PR #1382, opened 2021-11-25
Add parallelism to makefile for builds and test runs

krakjoe:
This doesn't have a huge impact in CI (but doesn't break anything) because CI is as slow as the slowest job anyway.

It does speed up local test runs.


Makefile
```diff
@@ -17,8 +17,15 @@ VERSION:=$(shell awk -F\' '/const VERSION/ {print $$2}' < src/DDTrace/Tracer.php
...
+RUN_TESTS_IS_PARALLEL := $(shell test $(shell php-config --vernum) -gt 70399 && echo 1 || echo 0)
+
+ifeq ($(RUN_TESTS_IS_PARALLEL), 1)
 RUN_TESTS_EXTRA_ARGS := -j$(shell nproc)
-RUN_TESTS_CMD := REPORT_EXIT_STATUS=1 TEST_PHP_SRCDIR=$(PROJECT_ROOT) USE_TRACKED_ALLOC=1 php -n -d 'memory_limit=-1' $(BUILD_DIR)/run-tests.php --show-diff -n -p $(shell which php) -q $(RUN_TESTS_EXTRA_ARGS)
+else
+RUN_TESTS_EXTRA_ARGS :=
+endif
+
+RUN_TESTS_CMD := RUN_TESTS_IS_PARALLEL=$(RUN_TESTS_IS_PARALLEL) REPORT_EXIT_STATUS=1 TEST_PHP_SRCDIR=$(PROJECT_ROOT) USE_TRACKED_ALLOC=1 php -n -d 'memory_limit=-1' $(BUILD_DIR)/run-tests.php --show-diff -n -p $(shell which php) -q $(RUN_TESTS_EXTRA_ARGS)
```
krakjoe, line 28: This (RUN_TESTS_IS_PARALLEL in env) is just for debug in CI run, I'll remove it ...

dockerfiles/compile_extension/build-dd-trace-php
```diff
@@ -35,7 +35,7 @@ apk add --no-cache \
     libexecinfo-dev \
     make \
 
-make all CFLAGS="-std=gnu11 -O2 -g -Wall -Wextra -Werror" ECHO_ARG="-e"
+make -j all CFLAGS="-std=gnu11 -O2 -g -Wall -Wextra -Werror" ECHO_ARG="-e"
```
krakjoe, line 38: I expected this to be used in the "Build extension" step, but it looks like it's not ?

bwoebi: This is just used on alpine builds.

bwoebi:
Generally looks good to me, but can we also get parallelism for CI test runs (line 447 in particular)? Quicker first results...

bwoebi (approved):
Cool, then let's merge :-)

PR #1383, opened 2021-11-26
Use internal root spans in Integrations in place of using the legacy API

labbati:
### Description

As an additional step away from our hybrid approach (legacy ve internal api), this PR:
- removes usage of `DDTrace\GlobalTracer` from `src/DDTrace/Integrations`;
- removes usage of `DDTrace\Contracts\Span` from `src/DDTrace/Integrations` with the exception of `DDTrace\Integrations\Integration` as it might cause backward compatibility issues if users have inherited from this class - even if it has always been considered internal;

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
+++ b/src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
@@ -40,22 +40,21 @@ class CakePHPIntegration extends Integration
                 return false;
             }
 
-            $scope = GlobalTracer::get()->getRootScope();
-            if (!$scope) {
+            $rootSpan = \DDTrace\root_span();
+            if (!$rootSpan) {
                 return false;
             }
+
             $integration->appName = \ddtrace_config_app_name(CakePHPIntegration::NAME);
-            $integration->rootSpan = $scope->getSpan();
-            $integration->addTraceAnalyticsIfEnabledLegacy($integration->rootSpan);
-            $integration->rootSpan->setTag(Tag::SERVICE_NAME, $integration->appName);
+            $integration->rootSpan = $rootSpan;
+            $integration->addTraceAnalyticsIfEnabled($integration->rootSpan);
+            $integration->rootSpan->service = $integration->appName;
             if ('cli' === PHP_SAPI) {
-                $integration->rootSpan->overwriteOperationName('cakephp.console');
-                $integration->rootSpan->setTag(
-                    Tag::RESOURCE_NAME,
-                    !empty($_SERVER['argv'][1]) ? 'cake_console ' . $_SERVER['argv'][1] : 'cake_console'
-                );
+                $integration->rootSpan->name = 'cakephp.console';
+                $integration->rootSpan->resource =
+                    !empty($_SERVER['argv'][1]) ? 'cake_console ' . $_SERVER['argv'][1] : 'cake_console';
             } else {
-                $integration->rootSpan->overwriteOperationName('cakephp.request');
+                $integration->rootSpan->name = 'cakephp.request';
             }
 
             \DDTrace\trace_method(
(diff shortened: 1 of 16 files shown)


src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
```diff
@@ -94,15 +90,15 @@ function (SpanData $span, array $args) use ($integration) {
...
+                    $integration->setError($integration->rootSpan, $args[0]);
                     return false;
                 },
             ]);
 
             \DDTrace\trace_method('CakeResponse', 'statusCode', [
                 'instrument_when_limited' => 1,
                 'posthook' => function (SpanData $span, $args, $return) use ($integration) {
-                    $integration->rootSpan->setTag(Tag::HTTP_STATUS_CODE, $return);
+                    $integration->rootSpan->meta[Tag::HTTP_STATUS_CODE] = $return;
```
bwoebi, line 101: There is a side-effect in current setTag(), is this intentional, or do we want to retain the `Internal Server Error` error type here?
```
            if ($key === Tag::HTTP_STATUS_CODE && $value >= 500) {
                $this->hasError = true;
                if (!isset($this->internalSpan->meta[Tag::ERROR_TYPE])) {
                    $this->internalSpan->meta[Tag::ERROR_TYPE] = 'Internal Server Error';
                }
            }
```

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
```diff
@@ -71,15 +69,13 @@ function (SpanData $span, array $args) use ($integration) {
...
-                    $integration->rootSpan->setTag(
-                        Tag::RESOURCE_NAME,
-                        $_SERVER['REQUEST_METHOD'] . ' ' . $this->name . 'Controller@' . $request->params['action']
-                    );
-                    $integration->rootSpan->setTag(Tag::HTTP_URL, Router::url($request->here, true));
-                    $integration->rootSpan->setTag('cakephp.route.controller', $request->params['controller']);
-                    $integration->rootSpan->setTag('cakephp.route.action', $request->params['action']);
+                    $integration->rootSpan->resource =
+                        $_SERVER['REQUEST_METHOD'] . ' ' . $this->name . 'Controller@' . $request->params['action'];
+                    $integration->rootSpan->meta[Tag::HTTP_URL] = Router::url($request->here, true);
```
bwoebi, line 74: `setTag()` does an `Urls::sanitize()` here (i.e. dropping a query string). If I read the CakePHP code correctly, the query string is retained in `Router::url()`.

src/DDTrace/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
```diff
@@ -79,15 +72,15 @@ public function registerIntegration(\CI_Router $router, Span $root, $service)
...
+            function (SpanData $span) use ($rootSpan, $method, $service) {
                 $class = \get_class($this);
                 $span->name = $span->resource = "{$class}.{$method}";
                 $span->service = $service;
                 $span->type = Type::WEB_SERVLET;
 
                 $this->load->helper('url');
-                $root->setTag(Tag::HTTP_URL, base_url(uri_string()));
-                $root->setTag('app.endpoint', "{$class}::{$method}");
+                $rootSpan->meta[Tag::HTTP_URL] = base_url(uri_string());
```
bwoebi, line 82: also here: sanitize

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
bwoebi: Also verify all other occurrences of Tag::HTTP_URL please.

src/DDTrace/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -88,18 +85,15 @@ function ($This, $scope, $args, $route) use ($rootSpan, $integration) {
...
-                );
+                $rootSpan->resource = $route->getActionName() . ' ' . $routeName;
 
-                $rootSpan->setTag('laravel.route.name', $routeName);
-                $rootSpan->setTag('laravel.route.action', $route->getActionName());
-                $rootSpan->setTag('http.url', $request->url());
-                $rootSpan->setTag('http.method', $request->method());
+                $rootSpan->meta['laravel.route.name'] = $routeName;
+                $rootSpan->meta['laravel.route.action'] = $route->getActionName();
+                $rootSpan->meta['http.url'] = $request->url();
```
bwoebi, line 95: Don't miss this one, for _reasons_ this does not use Tag::HTTP_URL, but the string value...

bwoebi:
Looks good to me, and is a pre-requisite to making the integrations independent from the userland APIs.

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
labbati: very good catch!

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
labbati: Need to check this better but the new exception code should take care of it. I had looked into it a long time ago so I need to go back and look. But good catch as a double-check.

labbati: About this. I think it is correct that we do not do this adjustment. We already set trace (local root span) on error when the status code is set by PHP and I don't think that further changes are required.

Users that use our legacy manual instrumentation api will have it for free.

Even the message would stay the same.

src/DDTrace/Integrations/Laravel/LaravelIntegration.php
labbati: Fixed https://github.com/DataDog/dd-trace-php/pull/1383/commits/7e375b84c45fa00e15853a4fcbb17e3ef1dc6b06

labbati:
Thank you @bwoebi

PR #1387, opened 2021-11-30
Adjust components

morrisonlevi:
### Description

Change BUILD_COMPONENTS_TESTING to DATADOG_PHP_TESTING.

Require `components/` prefix on includes e.g.

    #include <components/string_view/string_view.h>

This will prevent conflicts in the future like uuid/uuid.h and libuuid.

Set `BreakStringLiterals: false` for clang-format; wrapping string
literals makes for worse grep'ing. Re-format some tests which had
long string literals.

Remove custom catch2_main target for the standard target
Catch2::Catch2WithMain, with a polyfill for older/differently
configured versions.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/components/sapi/CMakeLists.txt
+++ b/components/sapi/CMakeLists.txt
@@ -2,7 +2,7 @@ add_library(datadog_php_sapi sapi.c)
 
 target_include_directories(datadog_php_sapi
   PUBLIC
-    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../..>
     $<INSTALL_INTERFACE:include>
 )
 
@@ -23,9 +23,9 @@ add_library(Datadog::Php::Sapi
   ALIAS datadog_php_sapi
 )
 
-if (${BUILD_COMPONENTS_TESTING})
+if (${DATADOG_PHP_TESTING})
   add_subdirectory(tests)
-endif()
+endif ()
 
 # This copies the include files when `install` is ran
 install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/sapi.h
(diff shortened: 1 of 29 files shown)


krakjoe (approved):

components/sapi/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_library(datadog_php_sapi sapi.c)
 
 target_include_directories(datadog_php_sapi
   PUBLIC
-    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../..>
```
SammyK, line 5: Was this the secret to exposing the parent directory name all along?

SammyK (approved):
👍 LTGM

components/sapi/CMakeLists.txt
morrisonlevi: Here's what the variables refer to:
```
components    <- PROJECT_SOURCE_DIR
└── sapi      <- CMAKE_CURRENT_SOURCE_DIR
```

We need `components/` to _not_ be in the include path, because then any component can be referred to by `component/component.h` instead of `components/component/component.h`.

We could probably do some build-time copies of the files so that each component only exposes its own headers and such, but that's a quality of implementation detail that could be changed later if we feel the need to safe-guard it.

SammyK: I like it. If we all agree, I think we'll eventually be able to change ZAI headers too. Although, the path would be a little different from [our former attempts/discussions](https://github.com/DataDog/dd-trace-php/blob/5f37d0e/zend_abstract_interface/functions/CMakeLists.txt#L28). I think `zend_abstract_interface/component/component.h` would have little chance of colliding with other libs. :)

PR #1388, opened 2021-12-01
Add queue and channel components

morrisonlevi:
### Description

The queue is bounded and NOT thread-safe.

The channel is a bounded, multiple-producer, single-consumer channel
suitable for inter-thread communication.

The profiler will use a channel to send stack samples to another thread
to be aggregated.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/components/channel/channel.c
+++ b/components/channel/channel.c
@@ -48,7 +48,10 @@ static bool receiver_recv(struct datadog_php_receiver_s *receiver, void **data, 
         uint64_t now = uv_hrtime();
         uint64_t timeout_target = now + timeout_nanos;
         do {
-            // Handle being signalled, waking up spuriously, and timing out the same.
+            /* Handle being signalled, waking up spuriously, and timing out the
+             * same. Note that the mutex will be released while waiting, and
+             * the mutex will be re-acquired when resuming.
+             */
             (void)uv_cond_timedwait(&channel->condvar, &channel->mutex, timeout_target - now);
             succeeded = queue->try_pop(queue, data);
             if (succeeded) {
--- a/components/channel/channel.c
+++ b/components/channel/channel.c
@@ -1,0 +1,178 @@
+#include "channel.h"
+
+#include <components/queue/queue.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <uv.h>
+
+struct datadog_php_channel_impl_s {
+    datadog_php_queue queue;
+
+    // enforce single consumer (refcount should be 0 or 1)
+    uint32_t receiver_count;
+
+    /* Multiple producers are fine. When there aren't any items in the queue, the
+     * consumer will use this info to determine whether to try waiting for more
+     * data or to return empty handed. If there is a known producer, it will wait;
+     * otherwise it will return without waiting.
+     */
+    uint32_t sender_count;
+
+    uv_mutex_t mutex;
+    uv_cond_t condvar;
+    void *buffer[];
+};
+
+/**
+ * Destroys the channel. _NOT_ thread safe.
+ */
+static void channel_dtor(datadog_php_channel_impl *channel) {
+    uv_mutex_destroy(&channel->mutex);
+    uv_cond_destroy(&channel->condvar);
+    free(channel);
+}
+
+static bool receiver_recv(struct datadog_php_receiver_s *receiver, void **data, uint64_t timeout_nanos) {
+    if (!receiver || !receiver->channel) {
+        return false;
+    }
+
+    datadog_php_channel_impl *channel = receiver->channel;
+    uv_mutex_lock(&channel->mutex);
+    datadog_php_queue *queue = &channel->queue;
+    bool succeeded = queue->try_pop(queue, data);
+
+    // If there wasn't an item but there are known producers, wait and try again
+    if (!succeeded && channel->sender_count && timeout_nanos) {
+        uint64_t now = uv_hrtime();
+        uint64_t timeout_target = now + timeout_nanos;
+        do {
+            // Handle being signalled, waking up spuriously, and timing out the same.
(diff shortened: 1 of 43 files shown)


components/channel/channel.h
```diff
@@ -0,0 +1,78 @@
...
+     * responsible for dtor'ing the data if applicable. Does not wait for the
+     * data to be received, only enqueued.
+     */
+    bool (*send)(datadog_php_sender *self, void *data);
+
+    /**
+     * Clones the sender. Can fail if there are too many senders already, in
+     * which false is returned and the caller should not dtor the clone.
+     */
+    bool (*clone)(datadog_php_sender *self, datadog_php_sender *clone);
```
krakjoe, line 41: This doesn't appear to be used or tested, and the only thing it's supposed to do (enforce sender limit) is left as a todo.

Could you talk a little bit about why this is required, where it would be used, and include a test for it ?

Alternatively, if this doesn't look necessary right now, maybe drop it ?

components/channel/channel.h
```diff
@@ -0,0 +1,78 @@
...
+
+    // private:
+    datadog_php_channel_impl *channel;
+};
+
+/**
+ * datadog_php_channel is a bounded, multiple-producer, single-consumer channel
+ * suitable for inter-thread communication.
+ *
+ * The caller is responsible for any additional cleanup of contents; it does
```
krakjoe, line 56: I'm concerned that what we have here is a buffered channel - producers don't synchronize with consumers, so the producer may free what is "buffered" and yet to be consumed. But making it the callers responsibility to ensure this doesn't happen makes the component hard to use.

Can you talk a little about your reasoning here ?

components/channel/channel.h
morrisonlevi: `clone` is supposed to be how you get an additional sender. For instance, if we had an HTTP uploader channel and the tracer and profiler both send HTTP messages then we'd have multiple senders (and still just one receiver, the thing that does the uploads).

At the moment this is unused, correct. I'll think about adding a test vs removing it for now.

components/channel/channel.h
morrisonlevi: Ah, "caller" was not a good choice there: it should say "user" instead.

An example: I malloc an object that represents a stack sample, and then send it through the channel. If it fails to enqueue, then the caller frees it. If it succeeds, then the receiver frees it. A conditional lifetime is not great but seemed better than the alternatives because we're in C. Feel free to propose something different if you think something that might be better.

SammyK (dismissed):
Not a real PR-approval (just testing the UI) 😄

components/channel/channel.h
krakjoe: I'll do thoughts and may push some more spec/tests

components/channel/channel.h
morrisonlevi: I have added a test for non-contended multiple-senders which uses `.clone`.

components/channel/channel.h
morrisonlevi: I changed `caller` to `user`, which is at least a slightly more accurate description.

SammyK (dismissed):
I remember reviewing this code earlier this year. :) I left a few non-blocking questions, but this looks great. 💯

Out of curiosity: when the profiler sends a stack sample but the channel queue buffer is full, is the sample dropped?

components/channel/channel.c
```diff
@@ -0,0 +1,178 @@
...
+        do {
+            // Handle being signalled, waking up spuriously, and timing out the same.
+            (void)uv_cond_timedwait(&channel->condvar, &channel->mutex, timeout_target - now);
+            succeeded = queue->try_pop(queue, data);
+            if (succeeded) {
+                break;
+            }
+
+            now = uv_hrtime();
+        } while (channel->sender_count && now < timeout_target);
```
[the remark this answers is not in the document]
morrisonlevi: The `uv_cond_timedwait` will release the mutex as it waits, and re-acquire the mutex when it comes back. The mutex isn't held while the channel is waiting. You are not the first person to mention something like this; I will add a comment to the `uv_cond_timedwait` call.

components/channel/channel.c
```diff
@@ -0,0 +1,178 @@
...
+static bool sender_clone(datadog_php_sender *self, datadog_php_sender *clone) {
+    if (!self || !clone || !self->channel) {
+        return false;
+    }
+
+    datadog_php_channel_impl *channel = self->channel;
+    uv_mutex_lock(&channel->mutex);
+    // todo: overflow checks
+    ++channel->sender_count;
+    uv_mutex_unlock(&channel->mutex);
```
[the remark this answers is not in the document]
morrisonlevi: It doesn't matter because `receiver_recv` will unlock the mutex while it waits, and re-acquire it when it resumes; the period where the lock is contended should be short.

morrisonlevi:
> Out of curiosity: when the profiler sends a stack sample but the channel queue buffer is full, is the sample dropped?

Yes, we discard the sample. This typically happens if there is a persistent networking error.

In the abstract, we could chose to drop the oldest sample and keep the newest, or we could drop the newest. Both have value, but dropping the newest is simpler.

-----

@krakjoe Any other feedback before merging?

krakjoe (approved):
Sorry I didn't get around to writing more tests/code for this yesterday.

I would have liked to develop this code further: I feel strongly that a component should be useful generally and in the context of the PHP extension and as it currently stands we could not use it.

All of that being said, we don't actually have any other uses for it today than the one being proposed.

Like the log component this looks fine as it is, and we can iteratively improve on this if/when the need arises.

So long as we're all going in with the limitations of this design plainly in sight, and I think we are, then all is okay.

components/channel/channel.c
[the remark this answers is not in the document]
SammyK: Thanks for adding the extra docs to clarify this one.

PR #1389, opened 2021-12-01
Add support for Code Coverage

krakjoe:
### Makefile additions for coverage collection / display

This adds the following rules:

  - test_coverage: will test components, then ZAI, then ext with appropriate flags [1]
  - test_coverage_collect: will collect coverage
  - test_coverage_output: will generate HTML report in ./coverage

### CI

CodeCov is integrated into CI, it will run on all supported PHP versions.

[1] test_coverage includes collect and output, but will fail if components/zai has fails. We ignore fails from php test suite for the purposes of coverage because we have some flaky tests.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


Makefile
```diff
@@ -85,6 +85,10 @@ test_c: export DD_TRACE_CLI_ENABLED=1
 test_c: $(SO_FILE) $(TEST_FILES) $(TEST_STUB_FILES)
 	$(RUN_TESTS_CMD) -d extension=$(SO_FILE) $(BUILD_DIR)/$(TESTS)
 
+test_c_coverage: export DD_TRACE_CLI_ENABLED=1
```
labbati, line 88: I think we do not need this line, since we are calling `make test_c` internally in any case and the sub-call to make will set it at the time tests are run. Unless I am making some wrong assumption

labbati (commented):
Only one quick comment. As a nice followup to this PR (not required to merge it) it would be cool to run them in CI (even only on master) and report the graphical output somewhere.

Thanks @krakjoe for this!

Makefile
krakjoe: Good catch

krakjoe:
@labbati that was going to be my next question, where shall we send coverage info ?

SammyK (approved):
One optional nit: would you be open to adding `coverage/` to the `.gitignore`?

Otherwise, this is working perfectly for me. Nice work @krakjoe!

krakjoe:
@SammyK will update ignore on next commit, nice catch ...

Makefile
```diff
@@ -5,10 +5,10 @@ SHELL := /bin/bash
 BUILD_SUFFIX := extension
 BUILD_DIR := $(PROJECT_ROOT)/tmp/build_$(BUILD_SUFFIX)
 ZAI_BUILD_DIR := $(PROJECT_ROOT)/tmp/build_zai
+COMPONENTS_BUILD_DIR := $(PROJECT_ROOT)/tmp/build_components
 SO_FILE := $(BUILD_DIR)/modules/ddtrace.so
 WALL_FLAGS := -Wall -Wextra
-EXTRA_CFLAGS :=
-CFLAGS := -O2 $(shell [ -n "${DD_TRACE_DOCKER_DEBUG}" ] && echo -O0 -g) $(EXTRA_CFLAGS) $(WALL_FLAGS)
+CFLAGS := $(shell [ -n "${DD_TRACE_DOCKER_DEBUG}" ] && echo -O0 -g || echo -O2 -g0) $(WALL_FLAGS)
```
labbati, line 11: @SammyK I think the `-g|-g0` was discussed previously, any memory from back then that applies to this change?

krakjoe: Oh, that's a woops ... I was sure I saw g0 somewhere, but not here :)

labbati (approved):
LGTM! But I'd like to have an eye from @SammyK especially on the build (non-)changes

PR #1390, opened 2021-12-01
Add log component

morrisonlevi:
### Description

The log component can write to a file or to stdout/stderr. It is not
threadsafe, and borrows the mutex and file descriptor used. It could
theoretically be used to write to other file descriptors without much
work, but since logging as written is a blocking operation this would
be risky.

It supports these log levels:
  - off
  - error
  - warn
  - info
  - debug

The profiler uses the component to build a thread-safe-ish abstraction
for a global profiler logger. The profiler uses the log levels to mean
roughly the following:
  - off: nothing will be logged
  - error: an error that will likely stop the profiler, such as failing
    to spawn a worker thread.
  - warn: an error that will likely not stop the profiler, such as
    an HTTP 5xx error on upload. The profile will be dropped, but the
    profiler will continue to gather samples and upload profiles.
  - info: a notable event that occurs infrequently, such as profiles
    being uploaded (happens once about every minute). If something
    happens once per request, that would be too noisy for this level.
  - debug: information intended for developers to understand internal
    order of events. May be very noisy.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/components/log/log.h
+++ b/components/log/log.h
@@ -1,0 +1,102 @@
+#ifndef DATADOG_PHP_PROFILER_LOG_H
+#define DATADOG_PHP_PROFILER_LOG_H
+
+#include <components/string_view/string_view.h>
+#include <pthread.h>
+#include <stdint.h>
+
+#if __cplusplus
+#define C_STATIC(...)
+#else
+#define C_STATIC(...) static __VA_ARGS__
+#endif
+
+typedef enum {
+    DATADOG_PHP_LOG_UNKNOWN = -1,
+    DATADOG_PHP_LOG_OFF = 0,
+    DATADOG_PHP_LOG_ERROR,
+    DATADOG_PHP_LOG_WARN,
+    DATADOG_PHP_LOG_INFO,
+    DATADOG_PHP_LOG_DEBUG,
+} datadog_php_log_level;
+
+// This is for the compiler to use; please do not use it directly!
+typedef struct datadog_php_logger_s {
+    int descriptor;  // borrowed, not owned
+    int log_level;
+    pthread_mutex_t *mutex;  // borrowed, not owned
+} datadog_php_logger;
+
+#define DATADOG_PHP_LOGGER_INIT \
+    { -1, DATADOG_PHP_LOG_UNKNOWN, NULL }
+
+/**
+ * Creates a logger instance in `logger`. The result may be valid or invalid;
+ * check the return value: `true` for valid.
+ *
+ * For now, only file descriptors are expected to work, but other types are
+ * expected to work in the future.
+ */
+__attribute__((nonnull)) bool datadog_php_logger_ctor(datadog_php_logger *logger, int descriptor,
+                                                      datadog_php_log_level log_level, pthread_mutex_t *mutex);
+
+/**
+ * Sets the .descriptor to -1, .log_level to `DATADOG_PROFILER_LOG_UNKNOWN`,
+ * and the .mutex to NULL.
+ *
+ * This is not threadsafe! It also doesn't close the descriptor nor destroy the
+ * mutex, as they are borrowed, not owned. Only do this when the program is
+ * back in single-threaded mode, or you risk a race condition.
+ *
+ * @param logger
+ */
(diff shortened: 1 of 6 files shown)


morrisonlevi:
Randomized test failure is a temporary one failing to pull a docker image.

krakjoe (dismissed):
I have some concerns here about validation, ctor/dtor stuff, raw fd support. But they are overshadowed by concerns about logging in general which needs considerable attention.

So, none of this should block us moving forward: We can iteratively improve on what you have here.

morrisonlevi:
Thanks for the review!

> I have some concerns here about validation, ctor/dtor stuff, raw fd support.

What concerns are these? Even if we don't address them now, it would be good to hear them.

components/log/log.h
```diff
@@ -0,0 +1,102 @@
...
+#endif
+
+typedef enum {
+    DATADOG_PHP_LOG_UNKNOWN = -1,
+    DATADOG_PHP_LOG_OFF = 0,
+    DATADOG_PHP_LOG_ERROR,
+    DATADOG_PHP_LOG_WARN,
+    DATADOG_PHP_LOG_INFO,
+    DATADOG_PHP_LOG_DEBUG,
+} datadog_php_log_level;
```
SammyK, line 21: Nit for later: @krakjoe brought up a good point that having `php` in the name makes this seem like a Zend Engine component. I believe the original reason for adding `php` in the namespace was to disambiguate it from other languages within APM, but that is assuming that this would become part of a shared library that other languages would use. I don't believe that will be the case so because the semantics of `php` for extensions in the wild means "Zend Engine", the name is a bit confusing.

This isn't a blocker, but after all the components are merged, we might want to discuss a naming scheme for components.

components/log/tests/log.cc
```diff
@@ -0,0 +1,71 @@
...
+    REQUIRE(mem != MAP_FAILED);
+
+    pthread_mutex_t mutex;
+    REQUIRE(pthread_mutex_init(&mutex, nullptr) == 0);
+
+    datadog_php_logger logger = DATADOG_PHP_LOGGER_INIT;
+    REQUIRE(datadog_php_logger_ctor(&logger, fd, DATADOG_PHP_LOG_DEBUG, &mutex));
+
+    datadog_php_string_view messages[3] = {
+        {datadog_php_string_view_from_cstr("[Datadog Profiling] ")},
```
SammyK, line 56: I was confused why we needed to support an array of string views, but now I see that we can more trivially prefix log messages with the product name without dealing with C-string concat. Nice.

SammyK (approved):
LGTM 👍

components/log/log.h
morrisonlevi: > I believe the original reason for adding php in the namespace was to disambiguate it from other languages within APM, but that is assuming that this would become part of a shared library that other languages would use.

Sort of: I am more worried about us having a symbol that conflicts with something in a datadog shared library, like `libddprof` or a theoretical `libdatadog` or `libddapm` or something which might also share the `datadog_php_` prefix. We could probably shorten it to `dd_php_` or something, but I want to avoid `trace` if possible because we're moving beyond just trace products.

components/log/tests/log.cc
morrisonlevi: Correct, and also can concatenate a non-literal string like `strerror`.

krakjoe:
> Thanks for the review!
> 
> > I have some concerns here about validation, ctor/dtor stuff, raw fd support.
> 
> What concerns are these? Even if we don't address them now, it would be good to hear them.

With regard to validation/ctor/dtor, specifically I'm talking about the way in which we determine a logger is still writable: It would probably look and be nicer to atomically fetch (and the dtor to atomically set) the fd and let that determine validity. I didn't suggest to change it though as I don't think supporting a shared descriptor is going to be enough moving forward, and "this works". 

Both syslog and a possible dd backend would require sockets. Both would require additional logic on top of sockets - you can only send 1024 chars at a time and have to send string specially formatted with priority/level data for syslog, similar logic would be required for a dd backend. In both cases, it won't be enough for us to pass raw sockets and write() them.

PR #1391, opened 2021-12-01
Add arena component

morrisonlevi:
### Description

This arena is used to do bump pointer style allocations inside of a
previously allocated block, such as stack memory, static storage, or
heap memory.

Individual allocations made by the arena are not tracked and are not
freed. The use-case is allocating a bunch of data which has the same
lifetime, and then everything can be freed at once by resetting the
arena. This means that care must be taken to either not store types
which have custom destructors, or that they should be tracked and dtor'd
outside of the arena before resetting it.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -53,6 +53,8 @@ add_library(datadog_php_components INTERFACE)
 
 add_subdirectory(string_view)
 
+add_subdirectory(arena)
+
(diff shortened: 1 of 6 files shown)


krakjoe (dismissed):

SammyK (approved):
LGTM 👍

SammyK (approved):

PR #1392, opened 2021-12-01
Add stack sample component

morrisonlevi:
### Description

Represents a stack trace with function/method name, file, and
line-number information, up to a bounded depth. Frames are not required
to have all members; for instance, there may be a frame that has no
function but exists in file X on line Y, and there may be a frame that
has function Z but no file nor line information.

This isn't collecting data from the Zend Engine; this just represents
the relevant information gathered.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -65,6 +65,7 @@ add_subdirectory(container_id)
 add_subdirectory(log)
 add_subdirectory(queue)
 add_subdirectory(sapi)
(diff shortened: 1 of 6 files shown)


krakjoe (dismissed):

krakjoe (approved):

SammyK (approved):
The components are starting to look more "profilery". 😎 LGTM

PR #1394, opened 2021-12-01
Add time component

morrisonlevi:
### Description

The time component accounts for differences in cpu-time calls in
different platforms. In the past it was also used for monotonic and
system time, which is why it is called "time" instead of "cpu-time".

There are no tests for this component -- I could not think of any tests
that would be meaningful.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


krakjoe (approved):

morrisonlevi:
Thanks for the review, Joe!

SammyK (approved):
👍 LGTM. Perhaps for tests we can have a simple sanity check that we're getting a non-zero return value/error from the clock.

PR #1395, opened 2021-12-02
Export only specific symbols

morrisonlevi:
### Description

The symbol `get_module` is the only symbol a PHP extension needs to
have exported; all others may be private. Right now, our most likely
symbol collisions are `mpack_*` and the Zend API polyfills which some
other extension may have also defined. This PR improves the situation.

This situation would get worse with the profiler. The profiler
statically links in libuv which is likely to conflict with other
extensions as it is a well-known event-loop library.

Edit: the appsec product uses `ddtrace_root_span_add_tag`, so expose
this as well.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/components/CMakeLists.txt
+++ b/components/CMakeLists.txt
@@ -1,8 +1,9 @@
 #[[ Need CMake 3.14 so that `install(TARGETS)` uses better defaults, which
     saves us from repeating configuration in every file (and must be the same).
-    Need CMake 3.15+ because the channel component fails on 3.14. Instead of
-    just bumping incrementally, I've bumped this to the latest version that's
-    been tested in CI for a while: 3.19.
+    Need CMake 3.15+ because the channel component fails on 3.14, as does the
+    time component's feature check. Instead of just bumping incrementally, I've
+    bumped this to the latest version that's been tested in CI for a while:
+    version 3.19.
 ]]
 cmake_minimum_required(VERSION 3.19)
 
@@ -69,6 +70,7 @@ add_subdirectory(log)
 add_subdirectory(queue)
 add_subdirectory(sapi)
 add_subdirectory(stack-sample)
+add_subdirectory(time)
 add_subdirectory(uuid)
 
 install(EXPORT DatadogPhpComponentsTargets
--- a/components/time/CMakeLists.txt
+++ b/components/time/CMakeLists.txt
@@ -1,0 +1,75 @@
+add_library(datadog-php-time OBJECT time.c time.h)
+
+target_include_directories(datadog-php-time
+  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../..>
+)
+
+include(CheckCSourceRuns)
+include(CMakePushCheckState)
+
+find_package(Threads)
+
+#[[ Prior to glibc 2.17, librt was required for clock_gettime.
+    Check for it without flags first.
+    clock_gettime gets used with pthread_getcpuclockid if the platform has it.
+]]
+cmake_push_check_state(RESET)
+set(CLOCK_GETTIME_PROGRAM "#include <time.h>
+int main(void) {
+  struct timespec ts;
+  return clock_gettime(CLOCK_REALTIME, &ts) == 0 ? 0 : 1;
+}
+")
+check_c_source_runs("${CLOCK_GETTIME_PROGRAM}" DATADOG_HAVE_CLOCK_GETTIME)
+cmake_pop_check_state()
+
+if (NOT DATADOG_HAVE_CLOCK_GETTIME)
+  # Try again with -lrt; use a different cache variable name or it won't run.
+  cmake_push_check_state(RESET)
+  set(CMAKE_REQUIRED_LIBRARIES -lrt)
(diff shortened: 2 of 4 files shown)


cataphract (commented):
As I mentioned in Slack, it's preferable to use `-fvisibility=hidden` and visibility attributes on the symbols that you wish to export.

This documents in source files which functions are public and therefore whose modification is restricted if compatibility with the dependencies is to be kept.

morrisonlevi:
> As I mentioned in Slack, it's preferable to use -fvisibility=hidden and visibility attributes on the symbols that you wish to export.

This is not generally adequate due to dependencies. For instance, if I link in libuv statically, I cannot hide its symbols, as they do not provide macros to do it. I tried to add them, but there was a lack of interest on their project's part because in their case they _do_ want the symbols to be publicly visible when linked in statically, so that all plugins can rely on it in the node.js ecosystem.

It will have to be done via linker script for now.

cataphract:
> > As I mentioned in Slack, it's preferable to use -fvisibility=hidden and visibility attributes on the symbols that you wish to export.
> 
> This is not generally adequate due to dependencies. For instance, if I link in libuv statically, I cannot hide its symbols, as they do not provide macros to do it. I tried to add them, but there was a lack of interest on their project's part because in their case they _do_ want the symbols to be publicly visible when linked in statically, so that all plugins can rely on it in the node.js ecosystem.
> 
> It will have to be done via linker script for now.

You could use `--exclude-libs`. gold seems to support it too.

morrisonlevi:
> You could use `--exclude-libs`.

`libtool -export-symbols` is portable on the relevant Mac and Linux systems. I like that it's simple, and works no matter what libraries or symbols we add because it states exactly which symbols are public, rather than excluding certain symbols.

I will ask other team members for their thoughts.

> my concern is mostly someone accidentally changing a public function. 

We could set up an ABI checker in CI; this way we will be alerted to ABI changes in PRs or on commits to release branches or something. Anyway, I don't think this PR needs blocked on that, since the situation is the same whether this PR is merged or not.

SammyK:
I don't have a very strong opinion either way, but I agree that documenting the public API directly in the header with a visibility attribute would help make more explicit which API's shouldn't be changed. This will be increasingly more important as products interdepend on one another.

morrisonlevi:
I just re-ran my tests with libuv and `-fvisibility=hidden` is definitely insufficient. I added `ZEND_API` to `ddtrace_root_span_add_tag` for PHP 7 and 8, but kept `-export-symbols` since that works with statically linked in libddprof + libuv, which the profiler will do.

Anilm3:
I generally prefer `-fvisibility=hidden` and visibility attributes, but given that there are portability concerns with `--exclude-libs` there is no other option (that I'm aware of) other than to use libtool. In any case, I concur that we should include a CI task to validate the resulting shared object.

SammyK:
Sounds good. I'm glad we can get docs in the header still. One nit: would you be open adding a `DATADOG_API` or `DATADOG_PUBLIC` macro instead of using the one from the Zend Engine?

morrisonlevi:
> One nit: would you be open adding a DATADOG_API or DATADOG_PUBLIC macro instead of using the one from the Zend Engine?

Would you be okay if I define it like this? Or do you want it re-implemented?

```c
#define DATADOG_API ZEND_API
```

SammyK:
> Would you be okay if I define it like this? Or do you want it re-implemented?

I personally would re-implement, but I have no strong feelings one way or another.

morrisonlevi:
I confirmed that using `-fvisibility=hidden` reduced the shared object size; it seems like we should do both `-fvisibility=hidden` and also `-export-symbols`.

morrisonlevi:
The PECL tests are failing; the path is not correct when built from package.xml:

> libtool: link: symbol file `./ddtrace.sym' does not exist

 I can't seem to build pecl packages locally and install them, so it's been difficult for me to figure it out. I would appreciate any help there.

config.m4
```diff
@@ -272,7 +273,11 @@ if test "$PHP_DDTRACE" != "no"; then
     [AC_MSG_ERROR([cannot find or include curl])])
 
   AC_CHECK_HEADER(time.h, [], [AC_MSG_ERROR([Cannot find or include time.h])])
-  PHP_SUBST(EXTRA_LDFLAGS)
+
+  dnl Only export symbols defined in ddtrace.sym, which should all be marked as
+  dnl DDTRACE_PUBLIC in their source files as well.
+  EXTRA_CFLAGS="$EXTRA_CFLAGS -fvisibility=hidden"
+  EXTRA_LDFLAGS="$EXTRA_LDFLAGS -export-symbols $ext_srcdir/ddtrace.sym"
```
krakjoe, line 281: these lines coming after the call to substitute looks strange, even if it works ... I'm not sure how it does work in fact ...

morrisonlevi: No problem, I can move the `PHP_SUBST` down, I'm pretty sure. The `PHP_CHECK_LIBRARY` calls just above this also append to `EXTRA_LDFLAGS`.

krakjoe (approved):

PR #1396, opened 2021-12-02
Sanitize user information from urls

labbati:
### Description

Add sanitization of the following use case from urls:

- my_user:my_password@some_url.com/path/
- my_user:@some_url.com/path/
- https://my_user:my_password@some_url.com/path/
- https://my_user:@some_url.com/path/

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/private/functions.php
```diff
@@ -127,6 +126,85 @@ function _util_uri_apply_rules($uriPath, $incoming)
     return implode('/', $fragments);
 }
 
+/**
```
labbati, line 129: Moved here because we do not want dependencies between `src/private/functions.php` and `src/DDTrace`.

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -64,7 +64,7 @@ public function init()
                 }
 
                 $info = \curl_getinfo($ch);
-                $sanitizedUrl = Urls::sanitize($info['url']);
+                $sanitizedUrl = \DDtrace\Private_\util_url_sanitize($info['url']);
```
labbati, line 67: `Urls::sanitize()` has been moved to `\DDtrace\Private_` because it is required by one of the functions in `\DDtrace\Private_` and from this namespace there should never be a dependency to `src/api`.

The original `Urls::sanitize()` is now a proxy to the new function.

tests/Unit/Http/UrlsTest.php
```diff
@@ -7,24 +7,15 @@
 
 final class UrlsTest extends BaseTestCase
 {
-    public function testSimpleUrlsAreReturned()
+    public function testSanitize()
     {
-        $this->assertSame('some_url.com/path/', Urls::sanitize('some_url.com/path/'));
```
labbati, line 12: All this use cases have been moved to https://github.com/DataDog/dd-trace-php/pull/1396/files#diff-11c158ecb066ee67275f17d6c2af030ba5e43772caf48122c9e7d1ad556eab82

src/api/Http/Urls.php
```diff
@@ -38,14 +38,78 @@ public function __construct(array $patternsWithWildcards = [])
...
+     *                           Defaults to `false`.
      */
-    public static function sanitize($url)
+    public static function sanitize($url, $dropUserInfo = false)
     {
-        return strstr($url, '?', true) ?: (string) $url;
+        /* The implementation of this method is an exact replica of \DDTrace\Private_\util_url_sanitize() - and has to
+         * be kept in sync - until \DDTrace\Private_\util_url_sanitize() will be removed as part of the PHP->C
+         * migration.
+         */
```
labbati, line 52: Please notice this comment

tests/Unit/Http/UrlsTest.php
```diff
@@ -7,24 +7,87 @@
...
+     * @param string $url
+     * @param string $expected
+     * @return void
+     */
+    public function testSanitize($url, $expected)
     {
-        $this->assertSame('some_url.com/path/', Urls::sanitize('some_url.com/path/'));
+        /* This test is an exact replica of the same method in tests/Unit/private/UriTest.php and has to be kept in sync
+         * until the other test will be removed as part of the PHP->C migration
+         */
```
labbati, line 21: Notice for reviewer

tests/Unit/Http/UrlsTest.php
```diff
@@ -7,24 +7,87 @@
...
+     * @param string $url
+     * @param string $expected
+     * @return void
+     */
+    public function testSanitizeDropUserinfo($url, $expected)
     {
-        $this->assertSame('some_url.com/path/', Urls::sanitize('some_url.com/path/?some=value'));
+        /* This test is an exact replica of the same method in tests/Unit/private/UriTest.php and has to be kept in sync
+         * until the other test will be removed as part of the PHP->C migration
+         */
```
labbati, line 62: Note for reviewer

tests/Unit/private/UriTest.php
```diff
@@ -502,4 +502,86 @@ public function testQueryStringIsRemoved()
...
+     * @dataProvider dataProviderSanitize
+     * @param string $url
+     * @param string $expected
+     * @return void
+     */
+    public function testSanitize($url, $expected)
+    {
+        /* This test is an exact replica of the same method in tests/Unit/Http/UrlsTest.php and has to be kept in sync
+         * until the current test will be removed as part of the PHP->C migration
+         */
```
labbati, line 516: Note for reviewer

tests/Unit/private/UriTest.php
```diff
@@ -502,4 +502,86 @@ public function testQueryStringIsRemoved()
...
+     * @dataProvider dataProviderSanitizeDropUserinfo
+     * @param string $url
+     * @param string $expected
+     * @return void
+     */
+    public function testSanitizeDropUserinfo($url, $expected)
+    {
+        /* This test is an exact replica of the same method in tests/Unit/Http/UrlsTest.php and has to be kept in sync
+         * until the current test will be removed as part of the PHP->C migration
+         */
```
labbati, line 557: Note for reviewer

src/private/functions.php
```diff
@@ -127,6 +126,79 @@ function _util_uri_apply_rules($uriPath, $incoming)
...
+ * @param string $url
+ * @param bool $dropUserInfo Optional. If `true`, removes the user information fragment instead of obfuscating it.
+ *                           Defaults to `false`.
+ * @return string
+ */
+function util_url_sanitize($url, $dropUserInfo = false)
+{
+    /* The implementation of this method is an exact replica of DDTrace\Http\Urls:;sanitize() - and has to
+     * be kept in sync - until this method will be removed as part of the PHP->C migration.
+     */
```
labbati, line 141: Note for reviewer

bwoebi:
I'm honestly wondering whether something like
```php
preg_replace([
    '(^(?:(?:[^/]+:)?//)?+\K[^/@:]+(?=:[^/@]+@))', // skip URL scheme, then replace up to : if followed by a later @
    '(^(?:(?:[^/]+:)?//)?+\?:\K[^/@]+)' // skip URL scheme followed by "?:", then replace up to @
], "?", strstr($url, '?', true)); // skip the query string
```
would be much more readable than url parsing, passing ?:?@ part along, path reconstruction, handling of parse_url failing to parse etc.?

(At least to me it is, but I have no problems reading and writing simple PCRE expressions at all.)

labbati:
> I'm honestly wondering whether something like
> 
> ```
> return preg_replace([
>     '(^(?:(?:[^/]+:)?//)?+\K[^/@:]+(?=:[^/@]+@))', // skip URL scheme, then replace up to : if followed by a later @
>     '(^(?:(?:[^/]+:)?//)?+\?:\K[^/@]+)' // skip URL scheme followed by "?:", then replace up to @
> ], "?", strstr($url, '?', true)); // skip the query string
> ```
> 
> would be much more readable than url parsing, passing ?:?@ part along, path reconstruction, handling of parse_url failing to parse etc.?
> 
> (At least to me it is, but I have no problems reading and writing simple PCRE expressions at all.)

Good point. I am uncertain. Reusing as much as possible of `parse_url` make sense, at the same time I agree that the regex is not too bad (but I am pretty familiar with them). I prefer to hear a third opinion, that might be enlightening.

labbati:
The regex proposal breaks idempotency of this method. I will look for a workaround to it, while still using the regex approach, as soon as I can get back to it.

krakjoe (dismissed):

tests/Unit/Http/UrlsTest.php
```diff
@@ -7,24 +7,109 @@
...
+            ['my_user:my_password@some_url.com/path/', 'some_url.com/path/'],
+            ['my_user:@some_url.com/path/', 'some_url.com/path/'],
+            ['my_user:@some_url.com/path/?key=value', 'some_url.com/path/'],
+            ['https://my_user:my_password@some_url.com/path/', 'https://some_url.com/path/'],
+            ['https://my_user:@some_url.com/path/', 'https://some_url.com/path/'],
+            ['https://my_user:@some_url.com/path/?key=value', 'https://some_url.com/path/'],
+
+            // idempotency
+            ['https://?:@some_url.com/path/?key=value', 'https://some_url.com/path/'],
+            ['?:?@some_url.com/path/?some=value#fragment', 'some_url.com/path/'],
```
bwoebi, line 104: I you truly want idempotency, also add a test without ?query_string part - because the strstr will then skip remove everything starting with the ? in the password.
That's why my regex suggestion wasn't attempting to be idempotent at all, because it hasn't been in the first place.

To be precise, we'd need to strip everything from the _first_ question mark _after_ any potential @.

tests/Unit/Http/UrlsTest.php
```diff
@@ -7,24 +7,109 @@
...
+            ['my_user:my_password@some_url.com/path/', '?:?@some_url.com/path/'],
+            ['my_user:@some_url.com/path/', '?:@some_url.com/path/'],
+            ['my_user:@some_url.com/path/?key=value', '?:@some_url.com/path/'],
+            ['https://my_user:my_password@some_url.com/path/', 'https://?:?@some_url.com/path/'],
+            ['https://my_user:@some_url.com/path/', 'https://?:@some_url.com/path/'],
+            ['https://my_user:@some_url.com/path/?key=value', 'https://?:@some_url.com/path/'],
+
+            // idempotency
+            ['https://?:@some_url.com/path/?key=value', 'https://?:@some_url.com/path/'],
+            ['?:?@some_url.com/path/?some=value#fragment', '?:?@some_url.com/path/'],
```
bwoebi, line 52: I you truly want idempotency, also add a test without ?query_string part - because the strstr will then skip remove everything starting with the ? in the password.
That's why my regex suggestion wasn't attempting to be idempotent at all, because it hasn't been in the first place.

To be precise, we'd need to strip everything from the _first_ question mark _after_ any potential @.

```suggestion
            // idempotency
            ['https://?:@some_url.com/path/?key=value', 'https://?:@some_url.com/path/'],
            ['?:?@some_url.com/path/', '?:?@some_url.com/path/'],
            ['?:@some_url.com/path/?some=?#fragment', '?:@some_url.com/path/'],
```

better test cases

bwoebi (changes requested):

tests/Unit/Http/UrlsTest.php
labbati: Thanks @bwoebi for suggesting the test case. I think that your concern should be addressed by using `<sanitized> in place of `?` if my understanding is correct. But please let me know if I am missing something or even if you have a test case in addition to the one you added above for me to add.

bwoebi (approved):
Yes, you are right :-)

PR #1400, opened 2021-12-03
Support appsec on the installer/misc installer changes

cataphract:
### Description

* Support installation of appsec.
* Several improvements to the installation script:
  * By default, install the latest version of the tracer and appsec
  * Installation of a component can be suppressed with `--no-tracer` and `--no-appsec`.
  * Fix detection of .ini dir: bail out if none exists; if more than one exists, select the first.
  * Delete temporary files on exit.
  * Some functions extracted.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


dockerfiles/verify_packages/installer/test_install_non_root_user.sh
```diff
@@ -13,7 +13,7 @@ usermod -a -G datadog datadog
 new_version="0.65.1"
 
 set +e
-output=$(su datadog -c "php /app/dd-library-php-setup.php --tracer-version ${new_version} --php-bin php")
+output=$(su datadog -c "php /app/dd-library-php-setup.php --tracer-version ${new_version} --no-trace --php-bin php")
```
krakjoe, line 16: Is this a woops ?

cataphract: ah yes, thankfully it has two mistakes that almost cancelled each other :p

dd-library-php-setup.php
```diff
@@ -47,9 +52,18 @@ function print_help_and_exit()
     --tracer-version <0.1.2>    Install a specific version. If set --tracer-url and --tracer-file are ignored.
     --tracer-url <url>          Install the tracing library from a url. If set --tracer-file is ignored.
     --tracer-file <file>        Install the tracing library from a local .tar.gz file.
+    --no tracer                 Do not install the tracing library.
```
bwoebi, line 55:
```suggestion
    --no-tracer                 Do not install the tracing library.
```

dd-library-php-setup.php
```diff
@@ -189,22 +227,147 @@ function install($options)
...
+    if (isset($options[OPT_APPSEC_FILE])) {
+        $tarball = $options[OPT_APPSEC_FILE];
+    } else {
+        $url = isset($options[OPT_APPSEC_URL])
+            ? $options[OPT_APPSEC_URL]
+            : sprintf(
+                'https://github.com/DataDog/dd-appsec-php/releases/' .
+                "download/v%s/dd-appsec-php-%s-amd64.tar.gz",
+                rawurlencode($options[OPT_APPSEC_VERSION]),
+                rawurlencode($options[OPT_APPSEC_VERSION])
```
bwoebi, line 271: I don't like redundant sprintf args :-)
```suggestion
                'https://github.com/DataDog/dd-appsec-php/releases/download/v%1$s/dd-appsec-php-%1$s-amd64.tar.gz',
                rawurlencode($options[OPT_APPSEC_VERSION])
```

dd-library-php-setup.php
```diff
@@ -189,22 +227,147 @@ function install($options)
...
+    );
+    echo "Installed dd-appsec files to '$installDir'\n";
+
+    // Actual installation
+    foreach ($selectedBinaries as $command => $fullPath) {
+        $binaryForLog = ($command === $fullPath) ? $fullPath : "$command ($fullPath)";
+        echo "Installing appsec to binary: $binaryForLog\n";
+        $phpProperties = ini_values($fullPath);
+
+        if (is_truthy($phpProperties[IS_DEBUG])) {
```
bwoebi, line 296: Maybe hard fail here in case the user explicilty gave some `--appsec-` flag as arg.

dd-library-php-setup.php
```diff
@@ -189,22 +227,147 @@ function install($options)
...
+                "the PHP version is unsupported; no such file '$extensionOrigin'\n";
+            continue;
+        }
+
+        $extensionDestination = "{$phpProperties[EXTENSION_DIR]}/$extensionFilename";
+
+        $tmpExtName = "$extensionDestination.tmp";
+        copy($extensionOrigin, $tmpExtName);
+        rename($tmpExtName, $extensionDestination);
+        echo "Copied '$extensionOrigin' '$extensionDestination'\n";
```
bwoebi, line 321:
```suggestion
        echo "Copied '$extensionOrigin' to '$extensionDestination'\n";
```

bwoebi:
As far as I remember, we intentionally did not default to use `--tracer-version latest` to make the user aware - i.e. he has to explicitly select what he installs - which versions and which components.
I still prefer this approach, it also won't give the user a feeling of being forced appsec upon by default, when he originally just wanted the tracer. I'd revert that part...

dockerfiles/verify_packages/installer/test_install_tracer_appsec_latest.sh
```diff
@@ -0,0 +1,18 @@
...
+assert_no_ddappsec
+
+# Install latest appsec and ddtrace using the php installer
+php dd-library-php-setup.php --php-bin php
+
+assert_ddtrace_installed
+assert_ddappsec_installed
+
+assert_file_exists "$(php_conf_dir)"/98-ddtrace.ini
+assert_file_exists "$(php_conf_dir)"/98-ddappsec.ini
```
labbati, line 18: This decision is still pending. @cataphract are you ok with using one file only?

dockerfiles/verify_packages/installer/test_install_tracer_appsec_latest_piecewise.sh
```diff
@@ -0,0 +1,25 @@
...
+set -e
+
+. "$(dirname ${0})/utils.sh"
+
+# Initially no ddtrace/ddappsec
+assert_no_ddtrace
+assert_no_ddappsec
+
+# Install first the latest tracer
+php dd-library-php-setup.php --php-bin php --no--appsec
```
labbati, line 12: here and on line 19 there it should be `--no-*`, why is the test not failing?

labbati (commented):
Thank you @cataphract for this PR, added a couple of comments. My main concern is about one file vs per-product config file.
I think that in the internal conversation we are leaning toward having one file only for consistency with our one library vision?

dd-library-php-setup.php
```diff
@@ -884,6 +1163,87 @@ function get_ini_template($requestInitHookPath)
...
+{
+    // phpcs:disable Generic.Files.LineLength.TooLong
+    return <<<EOD
+; Loads the dd-appsec extension
+extension = ddappsec.so
+
+; Enables or disables the loaded dd-appsec extension.
+; If disabled, the extension will do no work during the requests.
+; This value is ignored on the CLI SAPI, see ddappsec.enabled_on_cli.
+;ddappsec.enabled = Off
```
labbati, line 1176: For consistency with the tracer (`datadog.trace.*`) and profiling (`datadog.profiling.*`), can we move the naming here to `datadog.appsec.*`?

dd-library-php-setup.php
```diff
@@ -892,4 +1252,16 @@ function get_supported_php_versions()
     return ['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1'];
 }
 
+function delete_on_exit($dir)
```
labbati, line 1255: I really like this idea. Can we restrict allowed paths to the prefix `/tmp/`? Just to make sure we are not causing major issues with a bug in our installer, since it will have to be run as root.
Moreover please add a docblock specifying this restriction.

In a follow up PR, I can then make sure this function is used all over the places (even in the tracer installer).

labbati (changes requested):

dockerfiles/verify_packages/installer/test_install_tracer_appsec_latest.sh
cataphract: It wouldn't be my preference, but I can live with it.

dockerfiles/verify_packages/installer/test_install_tracer_appsec_latest_piecewise.sh
cataphract: `assert_no_ddappsec` is buggy. I've fixed this.

PR #1300, opened 2021-08-23
Add functions ZAI support for PHP 5 and 7

bwoebi:
### Description

Also adding a first use of PHP 7 functions ZAI, also existing on PHP 8.

### Readiness checklist
- [x] (only for Members) Changelog has been added to the release document.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.
- [x] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1401, opened 2021-12-03
ZAI methods on PHP 5: Args and retval

SammyK:
### Description

Before adding ZAI methods for PHP 7 and 8 we needed to update the API on PHP 5 to support calling methods with arguments. The retval handling has also been updated to work more similar to the ZAI functions API. This also will also make it easier to share the same tests on PHP 7 & 8.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zend_abstract_interface/methods/methods.h
+++ b/zend_abstract_interface/methods/methods.h
@@ -30,12 +30,13 @@ zend_class_entry *zai_class_lookup_ex(const char *cname, size_t cname_len ZAI_TS
  * Methods cannot be called outside of a request context so this MUST be called
  * from within a request context (after RINIT and before RSHUTDOWN).
  */
-bool zai_call_method_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC, int argc, ...);
+bool zai_call_method_ex(zval *object, const char *method, size_t method_len, zval **retval ZAI_TSRMLS_DC, int argc,
+                        ...);
 
 /* Calls a static method on a class entry 'ce'. Return value handling is the
  * same as zai_call_method().
  */
-bool zai_call_static_method_ex(zend_class_entry *ce, const char *method, size_t method_len, zval **retval TSRMLS_DC,
+bool zai_call_static_method_ex(zend_class_entry *ce, const char *method, size_t method_len, zval **retval ZAI_TSRMLS_DC,
                                int argc, ...);
 
 /* A convenience wrapper to call zai_class_lookup() using a string literal. This
--- a/zend_abstract_interface/methods/methods.h
+++ b/zend_abstract_interface/methods/methods.h
@@ -4,6 +4,8 @@
 #include <main/php.h>
 #include <stdbool.h>
 
+#include "../zai_compat.h"
+
 /* Work in progress
  *
  * The long-term goal is to provide ZAI data-structure shims so that the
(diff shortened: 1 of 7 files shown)


zend_abstract_interface/methods/php5/methods.c
```diff
@@ -76,27 +74,77 @@ static bool z_call_method_without_args_ex(zval *object, zend_class_entry *ce, co
...
+    fcc.function_handler = func;
+    fcc.calling_scope = ce;
+    if (!object && (EG(called_scope) && instanceof_function(EG(called_scope), ce TSRMLS_CC))) {
+        fcc.called_scope = EG(called_scope);
+    } else {
+        fcc.called_scope = ce;
     }
 
-    bool ret = false;
+    zend_fcall_info fci = {0};
```
krakjoe, line 90: fci = empty_fcall_info

zend_abstract_interface/methods/tests/methods.cc
```diff
@@ -27,17 +35,43 @@ static zval *zai_instantiate_object_from_ce(zend_class_entry *ce) {
     object_init_ex(obj, ce);
     Z_SET_REFCOUNT_P(obj, 1);
     Z_SET_ISREF_P(obj);
+
+    // Only calls constructors without passing any args
+    if (ce->constructor && (ce->constructor->common.fn_flags & ZEND_ACC_PUBLIC)) {
+        zval *retval_ptr = NULL;
+
+        zend_fcall_info fci = {0};
```
krakjoe, line 43: fci = empty_fcall_info

zend_abstract_interface/methods/tests/methods.cc
```diff
@@ -27,17 +35,43 @@ static zval *zai_instantiate_object_from_ce(zend_class_entry *ce) {
...
+        fci.size = sizeof(zend_fcall_info);
+        fci.object_ptr = obj;
+        fci.retval_ptr_ptr = &retval_ptr;
+        fci.no_separation = 1;
+
+        zend_fcall_info_cache fcc = {0};
```
krakjoe, line 49: fcc = empty_fcall_info_cache

zend_abstract_interface/methods/methods.h
```diff
@@ -18,30 +20,39 @@
...
- *   if (retval) {
- *     zval_ptr_dtor(&retval);
- *   }
+ *   zval_ptr_dtor(&retval);
  *
  * Methods cannot be called outside of a request context so this MUST be called
  * from within a request context (after RINIT and before RSHUTDOWN).
  */
-bool zai_call_method_without_args_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC);
+bool zai_call_method_ex(zval *object, const char *method, size_t method_len, zval **retval TSRMLS_DC, int argc, ...);
```
krakjoe, line 33: ZAI_TSRMLS_DC

zend_abstract_interface/methods/php5/methods.c
```diff
@@ -76,27 +74,77 @@ static bool z_call_method_without_args_ex(zval *object, zend_class_entry *ce, co
...
+        fci.params = param_ptrs;
+
+        /* The zend_call_method() API only supports up to two arguments so we
+         * use zend_call_function() instead.
          */
-        ret = !EG(exception);
+        zend_try { call_fn_result = zend_call_function(&fci, &fcc TSRMLS_CC); }
+        zend_catch { should_bail = true; }
+        zend_end_try();
+    } else {
```
bwoebi, line 141: redundant else branch, is exactly equal to the part at end end of the if, just move it out of the if.

zend_abstract_interface/methods/php5/methods.c
```diff
@@ -76,27 +74,77 @@ static bool z_call_method_without_args_ex(zval *object, zend_class_entry *ce, co
...
         // "Cannot call abstract method %s::%s()"
-        return false;
+        goto error_exit;
+    }
+
+    zend_fcall_info_cache fcc = {0};
+    fcc.initialized = 1;
+    fcc.function_handler = func;
+    fcc.calling_scope = ce;
+    if (!object && (EG(called_scope) && instanceof_function(EG(called_scope), ce TSRMLS_CC))) {
```
bwoebi, line 84: do we actually want to late static bind here? I'm not sure whether this is appropriate here, given the possibly arbitrary locations this zai method may be invoked from?

zend_abstract_interface/methods/php5/methods.c
SammyK: I know it looks weird, but in the back of my mind I thought this was intentional (looking at [ZAI functions](https://github.com/DataDog/dd-trace-php/pull/1300/files#diff-12dfce37c631fbed130a7c83f3c7318120d7aa98231eec5174f9368cadeaea95R62-R88)).

Sure enough, we ran into an [UBSan failure](https://github.com/DataDog/dd-trace-php/pull/1234#discussion_r634818384) that complained about a zero-length VLA back when we were using `params[argc]`. But now that we don't use a VLA, I'll move this out of the else block. We'll probably want to follow up with changing this on ZAI functions later as well.

zend_abstract_interface/methods/php5/methods.c
SammyK: TIL 👍

zend_abstract_interface/methods/php5/methods.c
SammyK: I'm not sure. Can you think of a specific example of how LSB might cause an issue here?

bwoebi: It doesn't cause an issue until it does :-D

Generally the scenario is you are in a parent class, an instrumentation is randomly triggered, invoking a method on the, and the child class uses static:: in its method - normally the call would go to the child method, except, if by a lot of bad luck, the call is triggered while we are in the parent class - then it goes to the parent method.

I have absolutely no idea if this will ever be an issue, but seems safer to me. Because I cannot imagine a scenario where LSB would be expected.

SammyK: You bring up a good point. I've disabled LSB to air on the safe side.

krakjoe (approved):
Approving this as it is, without 7/8 support.

There will be a follow up PR for a revised interface as discussed.

PR #1407, opened 2021-12-06
Export only specific symbols

morrisonlevi:
### Description

There are only two symbols which need to be exported:

1. `get_module`, needed by all PHP modules.
2. `ddtrace_root_span_add_tag`, needed by the appsec product.

Symbols which should be exported should be marked DDTRACE_PUBLIC, which
comes from the new `ddtrace_export.h` header. The symbols should also
be added to the `ddtrace.sym` file, one per line.

Using `-fvisibility=hidden` will hide our own symbols, reduce object
size, and to a mild degree improve performance. However, it won't hide
symbols from static libraries, which is why libtool's `-export-symbols`
is also used.

Note that for the hybrid extension hack that the symbol needs to be
public on Alpine, so it was switched to use `get_module` which is
always available on correctly built modules.

Note: this is a new PR because GH automatically closed #1395.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ JUNIT_RESULTS_DIR := $(shell pwd)
 
 all: $(BUILD_DIR)/configure $(SO_FILE)
 
-$(BUILD_DIR)/config.m4: $(M4_FILES)
+$(BUILD_DIR)/config.m4: $(M4_FILES) $(BUILD_DIR)/ddtrace.sym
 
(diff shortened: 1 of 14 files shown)


krakjoe (approved):

SammyK (approved):
This is great stuff, @morrisonlevi! 💯

PR #1409, opened 2021-12-06
Add profiler with wall-time and cpu-time

morrisonlevi:
### Description

The cpu-time is experimental, as it is biased towards functions which
do I/O. However, it is very cheap in CPU and memory to do, so it may
still be useful.

The profiler is compatible with PHP 7.1+ for NTS builds. There is some
basic ZTS disabling, but may be incomplete; need to test this.

Note that `m4/pkg.m4` is lifted directly from the PHP project.

Currently, nothing in CI builds or uses the profiler; this is just the
sources, though I have checked that it builds locally. Another PR will
do touch-up and get it going in CI.

If you want to build it locally, you'll need libuv and libddprof. I did
a pre-release of libddprof so this can be tested without using a Rust
toolchain:
https://github.com/DataDog/libddprof/releases/tag/v0.2.0-rc.1.

This PR supersedes #1372.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


config.m4
```diff
@@ -279,6 +293,26 @@ if test "$PHP_DDTRACE" != "no"; then
...
+  dnl How can we isolate these things to profiling/config.m4?
+  if test "$PHP_DATADOG_PROFILING" = "yes" ; then
+    PHP_ADD_INCLUDE([$ext_srcdir/profiling])
+
+    PHP_ADD_BUILD_DIR([$ext_builddir/components])
+    PHP_ADD_BUILD_DIR([$ext_builddir/components/arena])
+    PHP_ADD_BUILD_DIR([$ext_builddir/components/log])
+    PHP_ADD_BUILD_DIR([$ext_builddir/components/queue])
+    PHP_ADD_BUILD_DIR([$ext_builddir/components/stack-sample])
+    PHP_ADD_BUILD_DIR([$ext_builddir/components/time])
```
SammyK, line 305: Was `channel` left out intentionally?

ext/php8/ddtrace.c
```diff
@@ -80,17 +84,35 @@ static int ddtrace_startup(struct _zend_extension *extension) {
...
 }
 
 static void ddtrace_shutdown(struct _zend_extension *extension) {
+    ddtrace_internal_handlers_shutdown();
+
+#ifdef PHP_DATADOG_PROFILING_H
+    datadog_profiling_shutdown(extension);
+#else
     UNUSED(extension);
+#endif
```
SammyK, line 102: Since the profiler was started up after the internal handlers, would it make sense to shut the profiler down before the handlers? Basically it's the last to start up so I would expect it to be the first to shut down.

profiling/plugins/log_plugin/log_plugin.c
```diff
@@ -0,0 +1,142 @@
...
+    goto startup_descriptor_cleanup;
+  }
+  if (pthread_mutex_init(&logger_mutex, &mutex_attr) != 0) {
+    goto startup_descriptor_cleanup;
+  }
+
+  string_view_t env_var =
+      datadog_php_string_view_from_cstr("DD_PROFILING_LOG_LEVEL");
+
+  char *env = sapi_getenv_compat(env_var.ptr, env_var.len);
```
SammyK, line 88: Do you plan to switch this out with ZAI config in a follow up PR?

profiling/plugins/log_plugin/log_plugin.c
```diff
@@ -0,0 +1,142 @@
...
+  }
+  if (pthread_mutexattr_settype(&mutex_attr, PTHREAD_MUTEX_ERRORCHECK) != 0) {
+    goto startup_descriptor_cleanup;
+  }
+  if (pthread_mutex_init(&logger_mutex, &mutex_attr) != 0) {
+    goto startup_descriptor_cleanup;
+  }
+
+  string_view_t env_var =
+      datadog_php_string_view_from_cstr("DD_PROFILING_LOG_LEVEL");
```
SammyK, line 86: Did we ever end up figuring out if we're OK with the name being `profiling` (instead of `profiler` to be more similar to `tracer`)? Are other languages also going with `profiling` vs `profiler`? (BTW This is not something we need to figure out before merge.)

profiling/stack-collector/stack-collector.c
```diff
@@ -0,0 +1,110 @@
...
+        break;
+      }
+
+    } else if (execute_data->prev_execute_data) {
+      /* This may be a dummy frame. Dummy frames are often used by
+       * require/include but as far as I know there aren't any flags on the
+       * execute_data to let you know that's where it came from.
+       * So we skip this frame, intentionally _not_ increasing depth. I do not
+       * like this goto, but if this is the top frame then doing a --depth would
+       * cause a logic issue.
```
SammyK, line 104: Perhaps one option to avoid the goto would be to move the `++depth` to the end of the `for` loop.

config.m4
morrisonlevi: No, good catch!

profiling/plugins/log_plugin/log_plugin.c
morrisonlevi: I do not know the details of the ZAI config component yet, but I agree that at least it should be considered.

profiling/plugins/log_plugin/log_plugin.c
morrisonlevi: `DD_PROFILING_` is the standardized env var prefix for profiling. I have asked the team for more context and will relay what I hear.

Makefile
```diff
@@ -61,7 +61,7 @@ $(BUILD_DIR)/configure: $(BUILD_DIR)/config.m4
 	$(Q) (cd $(BUILD_DIR); phpize && sed -i 's/\/FAILED/\/\\bFAILED/' $(BUILD_DIR)/run-tests.php) # Fix PHP 5.4 exit code bug when running selected tests (FAILED vs XFAILED)
 
 $(BUILD_DIR)/Makefile: $(BUILD_DIR)/configure
-	$(Q) (cd $(BUILD_DIR); ./configure)
+	$(Q) (cd $(BUILD_DIR); ./configure --enable-datadog-profiling=$(PHP_DATADOG_PROFILING))
```
SammyK, line 64: Just putting a note here for others: this is strict about truthy values so this has to be `PHP_DATADOG_PROFILING=yes`.

ext/php7/ddtrace.c
```diff
@@ -111,17 +115,35 @@ static int ddtrace_startup(struct _zend_extension *extension) {
...
 }
 
 static void ddtrace_shutdown(struct _zend_extension *extension) {
+    ddtrace_internal_handlers_shutdown();
+
+#ifdef PHP_DATADOG_PROFILING_H
+    datadog_profiling_shutdown(extension);
+#else
     UNUSED(extension);
+#endif
```
SammyK, line 133: We already synced on this, but I'm just putting a reminder here on the shutdown order. :)

profiling/stack-collector/stack-collector.c
```diff
@@ -0,0 +1,101 @@
+#include "stack-collector.h"
+
+#include <Zend/zend_compile.h>
+#include <Zend/zend_portability.h>
+
+typedef datadog_php_string_view string_view_t;
+
+void datadog_php_stack_collect(zend_execute_data *execute_data,
+                               datadog_php_stack_sample *sample) {
```
SammyK, line 9: This is a perfect candidate to mark as `DATADOG_HOT`.

```c
#define DATADOG_HOT __attribute__((hot))
```

Not a blocker for this PR, but as we discussed, once we get some direct tests set up in ZAI SAPI we can comb through this more throughly for micro optimizations.

Mmmmmm, `DATADOG_HOT` makes me think of 🌭. 😋

Makefile
morrisonlevi: If anyone is aware of an existing m4 macro that handles some set of truthy values like `yes`, `true`, and `on`, please let me know.

profiling/stack-collector/stack-collector.c
morrisonlevi: Thanks for helping me with this! It ended up being pretty simple.

SammyK: For sure! I really enjoyed the pair programming session (as always). 🙂

profiling/stack-collector/stack-collector.c
morrisonlevi: I added `DDTRACE_HOT` and `DDTRACE_COLD`, and migrated existing `ZEND_COLD` usage to `DDTRACE_COLD`, and added it to a few more places while I was at it.

SammyK: Awesome! It looks like `datadog_php_stack_collect` wasn't marked with `DDTRACE_HOT`.

profiling/plugins/recorder_plugin/recorder_plugin.c
```diff
@@ -0,0 +1,741 @@
...
+    instant before = instant_now();
+    do {
+      record_msg *message;
+      if (receiver->recv(receiver, (void **)&message, sleep_for_nanos)) {
+        // an empty message can be sent, such as when we're shutting down
+        if (message) {
+          datadog_php_recorder_add(profile, message);
+          free(message);
+        }
+      }
```
SammyK, line 406: Did you figure out what you wanted to do about an upper-bounds memory limit?

profiling/plugins/recorder_plugin/recorder_plugin.c
```diff
@@ -0,0 +1,741 @@
...
+
+  ddprof_ffi_File files_[] = {{
+      .name = to_byteslice("profile.pprof"),
+      .file = &profile_buffer,
+  }};
+
+  struct ddprof_ffi_Slice_file files = {.ptr = files_,
+                                        .len = sizeof files_ / sizeof *files_};
+  ddprof_ffi_Request *request =
+      ddprof_ffi_ProfileExporterV3_build(exporter, start, end, files, 10000);
```
SammyK, line 178: Nit: When the API exists in a separate repo, it may be non-trivial to look up what this magic number is (`timeout_ms: u64`). Would you be open to switching out the literal with a const or macro? And perhaps add some context as to why `10000` was chosen as the timeout?

SammyK (approved):
This LGTM 👍 Nice job @morrisonlevi 💯 

I left some non-blocking nits that can be/not be addressed in follow up PRs.

morrisonlevi:
Thank you for the CR! I'm going to take it and... close the PR without merge. For public beta we'll ship this separately as discussed in Slack and Zoom. However, I will incorporate this feedback into the other repository. Thank you! I needed this code reviewed either way.

profiling/plugins/recorder_plugin/recorder_plugin.c
morrisonlevi: Great points. I'll extract it as a parameter to `ddprof_ffi_export` (which contains the quoted snippet) so the caller can choose the timeout, and document the choice of timeout there.

PR #1410, opened 2021-12-07
Have both legacy and new installer to fail when json PHP extension is not enabled

labbati:
### Description

The latest version (on `master`, not released yet) cannot be loaded if the json extension is not enabled and it fails with a warning:

```
[Tue Dec  7 06:05:46 2021] PHP Warning:  Cannot load module 'ddtrace' because required module 'json' is not loaded ...
```

This is the case for default installation of PHP, which does not include `json`, unless a user also add `php7-json`, for example.

This PR adds checks (and tests) to both legacy and new installers to make sure that if the extension is not installed we fail with a proper and explanatory message.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -135,3 +135,9 @@ services:
     - DD_API_KEY=${DATADOG_API_KEY}
     - DD_APM_ENABLED=true
     - DD_DOGSTATSD_NON_LOCAL_TRAFFIC=1
+
+  packager:
+    image: datadog/docker-library:ddtrace_php_fpm_packaging
+    working_dir: /app
+    volumes:
+      - .:/app
--- a/package/post-install.sh
+++ b/package/post-install.sh
@@ -6,7 +6,7 @@
 # (fpm 1.14.0, ruby 3.0, tar 1.34).
 # Packing fails if this script size in bytes is between <unknown> and 7680. This value has been found empirically and
 # measured via `wc -c`.
-# The issue seems related to this very specific file size ony, as generating a package that includes only this script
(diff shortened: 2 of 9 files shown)


package/post-install.sh
```diff
@@ -153,6 +153,23 @@ function verify_installation() {
         fail_print_and_exit
 }
 
+function verify_required_ext() {
+    # This comment has been added because without it the number of bytes of this script is 7432
+    # and the apk module of fpm generates a broken tar.
+    # For some reasons we have to reach at least a number of bytes greater or equal 7681 so I need
```
krakjoe, line 159: This is the strangest comment I have ever seen 😆 

Can we include a link to some information about this quirk ?

labbati: You are right, very unprofessional comment 😄 .

I moved it [here](https://github.com/DataDog/dd-trace-php/pull/1410/files#diff-78f4eefba9614400cb523865f64db2dc0dd3b1dc66ed89e6ef73d2ba2b937a8bR3-R12) and provided a more detailed investigation report, including the issue I opened on the fpm project https://github.com/jordansissel/fpm/issues/1866

docker-compose.yml
```diff
@@ -135,3 +135,9 @@ services:
     - DD_API_KEY=${DATADOG_API_KEY}
     - DD_APM_ENABLED=true
     - DD_DOGSTATSD_NON_LOCAL_TRAFFIC=1
+
+  packager:
```
labbati, line 139: It is useful when we need to investigate things related to the packager

krakjoe (approved):

PR #1415, opened 2021-12-15
Split INI setting in installer so they can be added separately when missing

labbati:
### Description

As we add properties and new modules (e.g. profiling) we need to add entries for the new properties to the Ini file so they are available for the user to tweak.

### Readiness checklist
- ~[ ] (only for Members) Changelog has been added to the release document.~

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/dd-library-php-setup.php
+++ b/dd-library-php-setup.php
@@ -163,7 +163,7 @@ function install($options)
                     "mkdir -p " . escapeshellarg($iniDir)
                 );
 
-                if (false === file_put_contents($iniFilePath, get_ini_template($installDirWrapperPath))) {
+                if (false === file_put_contents($iniFilePath, '')) {
                     print_error_and_exit("Cannot create INI file $iniFilePath");
                 }
                 echo "Created INI file '$iniFilePath'\n";
@@ -189,6 +189,9 @@ function install($options)
                     "sed -i 's@extension \?= \?\(.*\)@extension = ddtrace.so@g' " . escapeshellarg($iniFilePath)
                 );
             }
+
+            add_missing_ini_settings($iniFilePath, get_ini_settings($installDirWrapperPath));
+
             echo "Installation to '$binaryForLog' was successful\n";
         }
     }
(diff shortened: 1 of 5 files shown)


dd-library-php-setup.php
```diff
@@ -189,6 +191,9 @@ function install($options)
                     "sed -i 's@extension \?= \?\(.*\)@extension = ddtrace.so@g' " . escapeshellarg($iniFilePath)
                 );
             }
+
+            add_missing_ini_settings($iniFilePath, get_ini_settings($installDirWrapperPath));
```
morrisonlevi, line 195: Seems to me like there should be some error handling here, like if ini location doesn't exist or isn't reachable or something?

dockerfiles/verify_packages/installer/utils.sh
```diff
@@ -48,3 +58,7 @@ install_legacy_ddtrace() {
     tar -xf  "/tmp/legacy-${version}.tar.gz" -C /
     /opt/datadog-php/bin/post-install.sh
 }
+
+get_php_conf_dir() {
+    php -i | grep -i 'scan this dir for additional .ini files' | awk '{print $NF}'
```
morrisonlevi, line 63: I consider it my duty to educate people on AWK, so this is equivalent:

```
php -i | awk 'tolower($0) ~ /scan this dir for additional .ini files/ {print $NF}'
```

You don't have to change it; just an FYI ^_^

I don't think case insensitive matching is really necessary, so this would also suffice:

```
php -i | awk '/Scan this dir for additional .ini files/ {print $NF}'
```

dockerfiles/verify_packages/installer/utils.sh
```diff
@@ -12,6 +12,16 @@ assert_file_contains() {
     fi
 }
 
+assert_file_not_contains() {
+    output=$(cat ${1})
+    if [ -z "${output##*$2*}" ]; then
```
morrisonlevi, line 17: Sorry, what is this bash expansion doing?

dd-library-php-setup.php
```diff
@@ -775,136 +780,311 @@ function build_known_command_names_matrix()
...
+            'name' => 'datadog.trace.bgs_connect_timeout',
+            'default' => '2000',
+            'commented' => true,
+            'description' => 'Set connection timeout in milliseconds while connecting to the agent',
+        ],
+        [
+            'name' => 'datadog.trace.bgs_timeout',
+            'default' => '5000',
+            'commented' => true,
+            'description' => 'Set request timeout in millisecodns while while sending payloads to the agent',
```
morrisonlevi, line 1063:
```suggestion
            'description' => 'Set request timeout in milliseconds while while sending payloads to the agent',
```

dd-library-php-setup.php
```diff
@@ -775,136 +780,311 @@ function build_known_command_names_matrix()
...
+            'name' => 'datadog.trace.agent_port',
+            'default' => '8126',
+            'commented' => true,
+            'description' => 'Configures the agent port. If you need more flexibility use `datadog.trace.agent_url` instead',
+        ],
+        [
+            'name' => 'datadog.dogstatsd_port',
+            'default' => '8125',
+            'commented' => true,
+            'description' => 'Configures the dogstatsd agent port. If you need more flexibility use `datadog.trace.agent_url` instead',
```
morrisonlevi, line 923: Does the dogstatsd code actually use `datadog.trace.agent_url`? I don't think so, but it's been a while and could be wrong.

dd-library-php-setup.php
```diff
@@ -775,136 +780,311 @@ function build_known_command_names_matrix()
...
+        ],
+        [
+            'name' => 'datadog.trace.startup_logs',
+            'default' => 'On',
+            'commented' => true,
+            'description' => 'Enables startup logs, including diagnostic checks',
+        ],
+        [
+            'name' => 'datadog.service',
+            'default' => 'my_service',
```
morrisonlevi, line 891: I'm not sure if it's the agent or the backend, but _something_ will default to this service name if we manage to send something without a service:

```suggestion
            'default' => 'unnamed-php-service',
```

cc @brettlangdon , as this was mentioned somewhat recently in a discussion.

morrisonlevi (dismissed):

dd-library-php-setup.php
labbati: In case anything goes wrong while writing the file, `add_missing_ini_settings` already prints an error and exist. The text is  `print_error_and_exit("Cannot add additional settings to the INI file $iniFilePath");`.

If you have something else in mind feel free to let me know and I can implement.

dd-library-php-setup.php
labbati: Your are right https://github.com/DataDog/dd-trace-php/blob/271cff838c370ce06989e0fa89ab1cf67668375c/ext/php7/dogstatsd_client.c#L30-L31 very good catch. Removed it.

dockerfiles/verify_packages/installer/utils.sh
labbati: TL;DR: if the string in $2 is found, it deletes the string itself, plus everything before and after it.

---

It is [substring removal](https://tldp.org/LDP/abs/html/string-manipulation.html#:~:text=expr%20%22%24stringZ%22%20%3A%20%27.*%5C(......%5C)%27%60%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%23%20ABCabc-,Substring%20Removal,-%24%7Bstring%23substring%7D).

> `${string##substring}`
> Deletes longest match of $substring from front of $string.

Substring removal supports the wildcard `*`, meaning anything found there.

dockerfiles/verify_packages/installer/utils.sh
labbati: I appreciate this `awk` "advanced" usage (advanced if compared to my knowledge of it 😄 ). If you don't mind I'd prefer to stick with the existing implementation, that even if uses two tools instead of one I think is more immediate the the average reader (and by average I mostly mean myself 😃 )

morrisonlevi (approved):

PR #1417, opened 2021-12-15
Introduce selective output for run-tests

krakjoe:
### Reduce noise from test runs

Quick fix for makefile/CI to reduce noise


SammyK (approved):
Much better! 👍 The openssl lang test failure will be fixed in #1418. We could also probably exclude `SKIP`, but this is a big noise-reduction improvement. Thanks @krakjoe!

PR #1422, opened 2021-12-17
Add profiler installation

labbati:
### Description

Add profiling module installation to the new php installation script.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi (commented):
When installing on the 7.1-buster container, I got this:

```
Cannot load datadog-profiling - it was built with configuration API320160303,NTS, whereas running engine is API320160303,NTS,debug
Cannot load datadog-profiling - it was built with configuration API320160303,NTS, whereas running engine is API320160303,NTS,debug
Cannot load datadog-profiling - it was built with configuration API320160303,NTS, whereas running engine is API320160303,NTS,debug
```

Glancing through the changes, I don't see code for debug/zts incompatibilities, so this makes sense, I guess. I thought I saw code for this in a prior glance through at this PR; was it accidentally removed?

After switching to the nts, ndebug variant it installed cleanly and CLI diagnostics succeeded:

```
DD_AGENT_HOST=agent DD_PROFILING_ENABLED=1 php --ri=datadog-profiling

datadog-profiling

                      Profiling Recorder Diagnostics                      
Enabled => yes
Can create profiles => yes
                       Profiling Upload Diagnostics                       
info => [Datadog Profiling] Successfully uploaded profile.

Can upload profiles => yes
```

And I correctly have an empty profile in my Datadog UI from running the phpinfo diagnostic.

I think I probably need to publish rc.5 which disables the profiler by default, don't I? To make the beta -> GA transition smoother (people won't need to adjust variables?).

labbati:
Very good catch @morrisonlevi. It also uncovered a bug (zts) and an unhandled scenario (debug-zts). Added tests for all of those and fixed them.

I manually tested them replaing the github.com url with the circleci url locally.

Makefile
```diff
@@ -293,30 +293,11 @@ $(PACKAGES_BUILD_DIR):
...
-	cp -r ./bridge $(BUNDLE_TMP_GNU_TRACER_DIR); \
-	cp -r ./bridge $(BUNDLE_TMP_MUSL_TRACER_DIR); \
-	echo "$(VERSION)" > $(BUNDLE_TMP_GNU_LIBRARY_DIR)/VERSION; \
-	echo "$(VERSION)" > $(BUNDLE_TMP_MUSL_LIBRARY_DIR)/VERSION; \
-	tar -czvf $(PACKAGES_BUILD_DIR)/dd-library-php-x86_64-linux-gnu.tar.gz -C $(BUNDLE_TMP_GNU_DIR) .; \
-	tar -czvf $(PACKAGES_BUILD_DIR)/dd-library-php-x86_64-linux-musl.tar.gz -C $(BUNDLE_TMP_MUSL_DIR) .;
+	bash ./tooling/bin/generate-final-artifact.sh \
+		$(VERSION) \
+		$(PACKAGES_BUILD_DIR) \
+		https://github.com/DataDog/dd-prof-php/releases/download/v0.3.0-rc.4/datadog-profiling.tar.gz
```
morrisonlevi, line 300:
```suggestion
		https://github.com/DataDog/dd-prof-php/releases/download/v0.3.0-rc.5/datadog-profiling.tar.gz
```

morrisonlevi (commented):
Not a blocker for this PR unless you want it to be:

When a user has a build of PHP that isn't compatible with profiling (debug build, or unsupported version like PHP 5), I don't think we should silently proceed without the profiler. Since enabling profiling is opt-in, not opt-out, then if they say `--enable-profiling` and it's on a version or build of PHP that doesn't support it, then "succeeding" but actually not installing the profiler doesn't seem right to me. That's my $0.02.

Makefile
labbati: I will make this change in the release branch (and add it to the release steps)

labbati:
> then "succeeding" but actually not installing the profiler doesn't seem right to me. That's my $0.02.
I agree, taking a note and will change immediately after the release

PR #1426, opened 2021-12-22
Remove Bootstrap.php file and its usage

labbati:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi:
The primary role of Bootstrap is to initialize the userland root span object so that userland can access it. We cannot really get rid of Bootstrap altogether yet. What would be doable is moving some parts of bootstrap into a static initializer at the bottom of Tracer.php, so that the Tracer gets initialized the first time it is used in userland. What do you think about that?

I assume we will need to do that in any case so that old and new APIs can coexist for a while.

bwoebi:
By the way, we do intentionally not decide on priority sampling early. The primary reason for this is that sampling priorities can depend on some attributes on the root span, like name or service. The script itself may (and shall) set these early, yes, but only after it has been started, i.e. after automatic root span generation time.

Thus we delay deciding on it to the root span being closed (or the priority sampling being used for the first time, for distributed tracing):
https://github.com/DataDog/dd-trace-php/blob/d999fda4f8e9f33bbd5355c85430ae97de5a9c04/ext/php8/span.c#L187-L189

bwoebi:
Let's do this as a two-step process, whereby we will first decouple Integrations from the rest of the code (i.e. integration vs legacy API backend). (I'm working on a PR for this.)
Once this happened, we can do a bit of cleanup on the legacy backend side.

I"m closing this PR for now, because I will move things around a bit (conflicting here) and this PR (just removing Bootstrap altogether) is not the correct approach.

bwoebi:
Regarding your priority sampling change, I suspect this is a bug you observed, which I'm fixing in #1450.

PR #1432, opened 2021-12-29
Fix json symbol address resolving logic

SammyK:
### Description

This is a followup PR to address the [concerns raised](https://github.com/DataDog/dd-trace-php/pull/1366#discussion_r760421429) in #1366 with regard to resolving symbol addresses from ext/json. This technique implements weak symbols similar to how we resolved the same issue for ext/curl in #1181 & #1209 and gracefully handles the case where the symbol addresses could not resolve.

I also added a new shared-extension container for PHP 5.4 (`datadog/dd-trace-ci:php-5.4-shared-ext`) to test the edge cases.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1310,6 +1310,75 @@ jobs:
       - store_artifacts:
           path: /tmp/artifacts
 
+  ZaiSharedExts:
+    working_directory: ~/datadog
+    parameters:
+      php_version:
+        type: string
+    docker:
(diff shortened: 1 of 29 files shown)


krakjoe (approved):
The fix contained here is good, was tested live by two engineers.

The test for the fix in CI is not quite right, but because of the potential for badness, I'm approving this before CI is fixed so it can be tested by a customer that opened a bug this morning.

PR #1435, opened 2021-12-29
fix fake_json_ext load

krakjoe:
### Description

Fixes CI for shared exts, follow up to #1432


SammyK (approved):
Thanks @krakjoe! This was actually the way I implemented this originally but it wasn't loading the fake module for some reason. But now I realized that I was setting `zai_sapi_extension` before SINIT and `zai_sapi_reset_extension_global()` was resetting it. 🤦‍♂️ 

Thanks for digging into this and for the fix! 💯

PR #1436, opened 2021-12-29
Fix CI failures caused by unsafe signal handler

krakjoe:
### Description

This is an internal fix for our failing CI.

It's okay to mark this as XFAIL: we know why it sometimes will not produce useful output.

Since the end result is always a fault, and the only difference is the resulting output (and dump), it's not terrible to expect this to fail in CI.

It doesn't excuse use of unsafe code, and we have a plan to resolve that.

### DIFF
--- a/tests/ext/segfault_backtrace_enabled.phpt
+++ b/tests/ext/segfault_backtrace_enabled.phpt
@@ -4,8 +4,11 @@ Dump backtrace when segmentation fault signal is raised and config enables it
 <?php
 if (!extension_loaded('posix')) die('skip: posix extension required');
 if (getenv('SKIP_ASAN') || getenv('USE_ZEND_ALLOC') === '0') die("skip: intentionally causes segfaults");
+if (getenv('PHP_PEAR_RUNTESTS') === '1') die("skip: pecl run-tests does not support XFAIL");
 if (file_exists("/etc/os-release") && preg_match("/alpine/i", file_get_contents("/etc/os-release"))) die("skip Unsupported LIBC");
 ?>
+--XFAIL--
+Code called in the segv handler is not signal safe, this will sometimes result in a segfault.
 --ENV--
 DD_LOG_BACKTRACE=1
 --FILE--


tests/ext/segfault_backtrace_enabled.phpt
```diff
@@ -4,8 +4,11 @@ Dump backtrace when segmentation fault signal is raised and config enables it
 <?php
 if (!extension_loaded('posix')) die('skip: posix extension required');
 if (getenv('SKIP_ASAN') || getenv('USE_ZEND_ALLOC') === '0') die("skip: intentionally causes segfaults");
+if (getenv('DD_TRACE_CI_PECL') === '1') die("skip: pecl run-tests does not support XFAIL");
```
krakjoe, line 7: This handles the case we see in CI when we run with the very old and terrible Pear test runner.

tests/ext/segfault_backtrace_enabled.phpt
```diff
@@ -4,8 +4,11 @@ Dump backtrace when segmentation fault signal is raised and config enables it
...
 if (file_exists("/etc/os-release") && preg_match("/alpine/i", file_get_contents("/etc/os-release"))) die("skip Unsupported LIBC");
 ?>
+--XFAIL--
```
krakjoe, line 10: This handles the case where the same fault causes phpunit or run-tests.php to fail. We do not see this in CI, it's nevertheless a possibility.

tests/ext/segfault_backtrace_enabled.phpt
SammyK: We don't need to add a new env var here since we can check the existing `PHP_PEAR_RUNTESTS` ([example](https://github.com/DataDog/dd-trace-php/blob/f5666ab/tests/ext/sandbox/hook_function/01.phpt#L6)).

krakjoe: Excellent, I went looking for something like that and couldn't find it in RunTest source code ... 

RunTests is very terrible by the wa, hasn't been updated in > ten years, and almost certainly shouldn't be part of anyones CI pipeline.

labbati (approved):
LGTM!

PR #1195, opened 2021-04-07
Remove embed sapi

labbati:
### Description

As a followup to https://github.com/DataDog/dd-trace-ci/pull/34, this PR removes the `embed` sapi.

### Readiness checklist
- [x] (only for Members) Changelog has been added to the release document.
- ~[ ] Tests added for this feature/bug.~

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1447, opened 2022-01-06
Revert "remove embed sapi (#1195)"

SammyK:
### Description

This reverts commit c2bbaad6469c0db1264eecf9434082350bb5543c. I believe #1195 was merged in error.

```bash
$ ll /opt/php/8.0/lib/
total 43512
-rwxr-xr-x  1 root root 44551144 Oct  2 00:21 libphp.so
drwxr-xr-x 15 root root     4096 Oct  2 00:22 php
```

The containers still have `libphp` (which is good) so they do not appear to have been rebuilt based on this change.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/dockerfiles/ci/alpine/php-8.0/configure.sh
+++ b/dockerfiles/ci/alpine/php-8.0/configure.sh
@@ -7,6 +7,7 @@ ${PHP_SRC_DIR}/configure \
     --build="$(dpkg-architecture --query DEB_BUILD_GNU_TYPE)" \
     --enable-option-checking=fatal \
     --enable-cgi \
+    --enable-embed \
     --enable-fpm \
     --enable-ftp \
     --enable-mbstring \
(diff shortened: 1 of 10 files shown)


krakjoe:
Why do we need embed SAPI ?

SammyK:
I agree, but since we need Tea SAPI for our core tests, that includes all of our containers. We should be able to test directly on any architecture that we support.

krakjoe (approved):

krakjoe:
Okay, after further discussion we're going to just have this strange dependency for now - one SAPI relying on another is very odd. Furthermore, embed is old, unmaintained, untested, and could be removed at any moment.

This SAPI is potentially a replacement for embed SAPI and to have it as a dependency now is quite strange, our actual dependency is php-src, and ideally we want to do in-tree builds of tsapi in future.

Since this is not so straight forward to setup, we'll go ahead and revert this change for now ...

SammyK:
Agreed. Thanks for looking into this @krakjoe!

PR #1448, opened 2022-01-07
Get rid of ini-ignoring and redundant checking of ddtrace being enabled

bwoebi:
In particular datadog.trace.cli_enabled was not accounted for, leading to tracing being aborted, if the request_init_hook was active.


SammyK (approved):
Thanks for digging into this. LTGM 👍

To me this identifies a hole in the tests we should patch up as well. Perhaps something like "configuration integration tests" or something. WDYT?

bwoebi:
Yeah, I think enabling/disabling of ddtrace altogether is not well tested overall currently.

PR #1452, opened 2022-01-10
ZAI Symbols Interface

krakjoe:
### Description

Accessing symbols will be an integral part of the new integrations we are working on migrating from PHP. We currently have multiple incomplete interfaces which ZAI Symbols replaces with access to all of the following via a single interface:

   - classes (global, namespace)
   - functions (global, namespace, class, object)
   - constants (global, namespace, class, object)
   - properties (class, object)
   - locals (function static, frame)

Symbols also contains a call interface and a `new` interface.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zend_abstract_interface/symbols/lookup.c
+++ b/zend_abstract_interface/symbols/lookup.c
@@ -1,0 +1,519 @@
+#include <sandbox/sandbox.h>
+#include "symbols.h"
+#include "zend_compile.h"
+
+// clang-format off
+static inline bool zai_symbol_update(zend_class_entry *ce ZAI_TSRMLS_DC) {
+    if (ce->ce_flags & ZEND_ACC_CONSTANTS_UPDATED) {
+        return true;
+    }
+
+    volatile bool zai_symbol_updated = true;
+
+    zai_sandbox sandbox;
+    zai_sandbox_open(&sandbox);
+
+    zend_try {
+#if PHP_VERSION_ID < 70000
+        zend_update_class_constants(ce ZAI_TSRMLS_CC);
+#else
+        if (zend_update_class_constants(ce) != SUCCESS) {
+            zai_symbol_updated = false;
+        }
+#endif
+    } zend_catch {
+        zai_symbol_updated = false;
+    } zend_end_try();
+
+    if (zai_symbol_updated && !EG(exception)) {
+        zai_sandbox_close(&sandbox);
+        return true;
+    }
+
+    zai_sandbox_close(&sandbox);
+    return false;
+}
+
+static inline void *zai_symbol_lookup_table(HashTable *table, const char *name, size_t length, bool ncase, bool pointer ZAI_TSRMLS_DC) {
+#if PHP_VERSION_ID >= 70000
+    zval *result = zend_hash_str_find(table, name, length);
+#else
+    void *result = NULL;
+
+    zend_hash_find(table, name, length + 1, (void **)&result);
+#endif
+
+    if (!result && ncase) {
+        char *ptr = (char *)pemalloc(length + 1, 1);
+
+        for (uint32_t c = 0; c < length; c++) {
+            ptr[c] = tolower(name[c]);
+        }
+
+        ptr[length] = 0;
+
+#if PHP_VERSION_ID >= 70000
+        result = zend_hash_str_find(table, ptr, length);
+#else
+        zend_hash_find(table, ptr, length + 1, (void **)&result);
+#endif
+
+        pefree(ptr, 1);
+    }
+
+#if PHP_VERSION_ID >= 70000
+    if (pointer && result) {
+        return Z_PTR_P(result);
+    } else {
+        return result;
+    }
+#endif
+    return result;
+}
+
+static inline size_t zai_symbol_lookup_lengthof(zai_string_view *view) {
+    if (!view->len) {
+        return 0;
+    }
+
+    if (memcmp(view->ptr, "\\", sizeof("\\")-1) == SUCCESS) {
+        return view->len - 1;
+    }
+
+    return view->len;
+}
+
+static inline const char* zai_symbol_lookup_startof(zai_string_view *view) {
+    if (!view->len) {
+        return NULL;
+    }
+
+    if (memcmp(view->ptr, "\\", sizeof("\\")-1) == SUCCESS) {
+        return view->ptr + 1;
+    }
+
+    return view->ptr;
+}
+
+static inline zai_string_view* zai_symbol_lookup_key(zai_string_view *namespace, zai_string_view *name, bool lower ZAI_TSRMLS_DC) {
+    zai_string_view *result = pemalloc(sizeof(zai_string_view), 1);
+
+    size_t namespace_length = zai_symbol_lookup_lengthof(namespace);
+    size_t name_length      = zai_symbol_lookup_lengthof(name);
+
+    const char* namespace_start   = zai_symbol_lookup_startof(namespace);
+    const char* name_start        = zai_symbol_lookup_startof(name);
+
+#define CHAR_AT(n) (((char*) result->ptr)[n])
+    if (namespace_length) {
+        result->len = namespace_length + name_length + sizeof("\\")-1;
+        result->ptr = pemalloc(result->len + 1, 2);
+
+        memcpy(&CHAR_AT(0), namespace_start, namespace_length);
+
+        for (uint32_t c = 0; c < namespace_length; c++) {
+            CHAR_AT(c) = tolower(CHAR_AT(c));
+        }
+
+        memcpy(&CHAR_AT(namespace_length), "\\", sizeof("\\")-1);
+        memcpy(&CHAR_AT(namespace_length + (sizeof("\\")-1)), name_start, name_length);
+    } else {
+        result->len = name_length;
+        result->ptr = pemalloc(result->len + 1, 1);
+
+        memcpy(&CHAR_AT(0), name_start, name_length);
+    }
+
+    if (lower) {
+        for (uint32_t c = namespace_length ? namespace_length + (sizeof("\\")-1) : 0;
+                      c < result->len;
+                      c++) {
+            CHAR_AT(c) = tolower(CHAR_AT(c));
+        }
+    }
+
+    CHAR_AT(result->len) = 0;
+#undef CHAR_AT
+
+    return result;
+}
+
+static inline zend_class_entry *zai_symbol_lookup_class_impl(zai_symbol_scope_t scope_type, void *scope, zai_string_view *name ZAI_TSRMLS_DC) {
+    void *result = NULL;
+
+    switch (scope_type) {
+        case ZAI_SYMBOL_SCOPE_GLOBAL: {
+            result = zai_symbol_lookup_table(
+                EG(class_table),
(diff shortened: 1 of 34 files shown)


zend_abstract_interface/symbols/symbols.h
```diff
@@ -0,0 +1,127 @@
...
+    /* The next parameter is zend_execute_data* */
+    ZAI_SYMBOL_SCOPE_FRAME,
+    /* The next symbol is zend_function* with type == ZEND_USER_FUNCTION */
+    ZAI_SYMBOL_SCOPE_STATIC,
+} zai_symbol_scope_t;
+
+/* LOOKUP */
+void* zai_symbol_lookup(
+        zai_symbol_type_t symbol_type,
+        zai_symbol_scope_t scope_type, void *scope,
```
labbati, line 63: Since clang-format is off here, this happens which is very confusing to me when reading code. Some parameters are on their own line, some other are coupled with another one on the same line.

I believe that we should either follow the default approach

```
void* zai_symbol_lookup(zai_symbol_type_t symbol_type, zai_symbol_scope_t scope_type, void *scope,
                        other_param p1, int p2);
```

or (in a separate PR that I can contribute if others agree) go even further and set `BinPackParameters: false` and `BinPackArguments: false` - [docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

I am fine either way...maybe slightly leaning toward the second option.

labbati (commented):
Some non-blocking nit picks

zend_abstract_interface/symbols/symbols.h
krakjoe: I done this to try and emphasise the relationship between the _type that describes the following void* ... it's a special case because of the void*

I will look at format options, I'm not very familiar

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+    if (!view->len) {
+        return NULL;
+    }
+
+    if (memcmp(view->ptr, "\\", sizeof("\\")-1) == SUCCESS) {
+        return view->ptr + 1;
+    }
+
+    return view->ptr;
+}
```
bwoebi, line 96: Can we coalesce this to 
```suggestion
static inline zai_string_view zai_symbol_lookup_clean(zai_string_view view) {
    if (!view.len || *view.ptr == '/') {
        return view;
    }
    return (zai_string_view){ .ptr = view.ptr + 1, .len = view.len - 1 };
}
```
and then have `zai_symbol_lookup_table` take `zai_string_view` instead of this early unpacking?

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+
+static inline zai_string_view* zai_symbol_lookup_key(zai_string_view *namespace, zai_string_view *name, bool lower ZAI_TSRMLS_DC) {
+    zai_string_view *result = pemalloc(sizeof(zai_string_view), 1);
```
bwoebi, line 99:
```suggestion
static inline zai_string_view zai_symbol_lookup_key(zai_string_view *namespace, zai_string_view *name, bool lower ZAI_TSRMLS_DC) {
    zai_string_view result = pemalloc(sizeof(zai_string_view), 1);
```
don't alloc zai_string_views, return it without pointers. It's a 16 byte structure. A compiler can efficiently transfer that, an alloc is costly.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+
+    size_t namespace_length = zai_symbol_lookup_lengthof(namespace);
+    size_t name_length      = zai_symbol_lookup_lengthof(name);
+
+    const char* namespace_start   = zai_symbol_lookup_startof(namespace);
+    const char* name_start        = zai_symbol_lookup_startof(name);
+
+#define CHAR_AT(n) (((char*) result->ptr)[n])
+    if (namespace_length) {
+        result->len = namespace_length + name_length + sizeof("\\")-1;
```
bwoebi, line 109: I find these `sizeof("\\")-1` confusing. It's a single char? then just add `1`.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+        memcpy(&CHAR_AT(namespace_length + (sizeof("\\")-1)), name_start, name_length);
+    } else {
+        result->len = name_length;
+        result->ptr = pemalloc(result->len + 1, 1);
+
+        memcpy(&CHAR_AT(0), name_start, name_length);
+    }
+
+    if (lower) {
+        for (uint32_t c = namespace_length ? namespace_length + (sizeof("\\")-1) : 0;
```
bwoebi, line 128: redundant condition (the tolower of \\ is also \\)
```suggestion
        for (uint32_t c = namespace_length;
```

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+#define CHAR_AT(n) (((char*) result->ptr)[n])
+    if (namespace_length) {
+        result->len = namespace_length + name_length + sizeof("\\")-1;
+        result->ptr = pemalloc(result->len + 1, 2);
+
+        memcpy(&CHAR_AT(0), namespace_start, namespace_length);
+
+        for (uint32_t c = 0; c < namespace_length; c++) {
+            CHAR_AT(c) = tolower(CHAR_AT(c));
+        }
```
bwoebi, line 116:
```suggestion
        for (uint32_t c = 0; c < namespace_length; c++) {
            CHAR_AT(c) = tolower(namespace_start[c]);
        }
```

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+
+        memcpy(&CHAR_AT(namespace_length), "\\", sizeof("\\")-1);
```
bwoebi, line 118:
```suggestion
        CHAR_AT(namespace_length) = '\\';
```

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+static inline zai_string_view* zai_symbol_lookup_key(zai_string_view *namespace, zai_string_view *name, bool lower ZAI_TSRMLS_DC) {
+    zai_string_view *result = pemalloc(sizeof(zai_string_view), 1);
+
+    size_t namespace_length = zai_symbol_lookup_lengthof(namespace);
+    size_t name_length      = zai_symbol_lookup_lengthof(name);
+
+    const char* namespace_start   = zai_symbol_lookup_startof(namespace);
+    const char* name_start        = zai_symbol_lookup_startof(name);
+
+#define CHAR_AT(n) (((char*) result->ptr)[n])
```
bwoebi, line 107: you can get rid of this by just doing `char *name;` and assigning `name` then later to the zai_string_view instead of casting to char * on every assignment.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+    if (zai_symbol_updated && !EG(exception)) {
+        zai_sandbox_close(&sandbox);
+        return true;
+    }
+
+    zai_sandbox_close(&sandbox);
+    return false;
+}
+
+static inline void *zai_symbol_lookup_table(HashTable *table, const char *name, size_t length, bool ncase, bool pointer ZAI_TSRMLS_DC) {
```
bwoebi, line 37: Once we drop PHP 5, we can make these things nicer with just the appropriate getter function like `zend_hash_str_find_ptr_lc` (polyfilling it for PHP 7)...

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+            local = ZEND_CALL_VAR_NUM(ex, var);
+            break;
+        }
+    }
+
+    if (!local) {
+        return NULL;
+    }
+
+    ZVAL_DEREF(local);
```
bwoebi, line 427: Isn't this missing an IS_INDIRECT check? esp. when called from the main frame.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+                false, false ZAI_TSRMLS_CC);
+
+#if PHP_VERSION_ID < 70000
+            if (!property) {
+                return NULL;
+            }
+
+            return *(zval**) property;
+#else
+            return (zval*) property;
```
bwoebi, line 349: why do we deref local frame lookup, but not property?

zend_abstract_interface/symbols/symbols.h
```diff
@@ -0,0 +1,127 @@
...
+    ZAI_SYMBOL_FUNCTION_KNOWN,
+    /* The function parameter is zai_string_view* */
+    ZAI_SYMBOL_FUNCTION_NAMED,
+} zai_symbol_function_t;
+
+bool zai_symbol_call(
+    zai_symbol_scope_t scope_type, void *scope,
+    zai_symbol_function_t function_type, void *function,
+    zval **rv ZAI_TSRMLS_DC,
+    uint32_t argc, ...);
```
bwoebi, line 123: Can we just split that in two functions? (the named variant can call the known function internally)
Or maybe just provide two macros also taking care of the argc like:
https://github.com/DataDog/dd-trace-php/blob/5f37d0e0df5fb51cdf87de5cace5276423615edc/zend_abstract_interface/functions/functions.h#L42-L46

bwoebi (commented):
I don't know yet whether I'll be happy with the interface exposed, but we'll figure that out once we start using it, writing code with it...

zend_abstract_interface/symbols/lookup.c
krakjoe: Yes, nice, thankyou

zend_abstract_interface/symbols/symbols.h
krakjoe: I have solved the problem with a thin API layer on top of the single interfaces .. we have the two functions requested (and a few besides), and we can add as we need them ...

zend_abstract_interface/symbols/lookup.c
krakjoe: done

zend_abstract_interface/symbols/lookup.c
krakjoe: done

zend_abstract_interface/symbols/lookup.c
```diff
@@ -0,0 +1,519 @@
...
+
+    return view->len;
+}
+
+static inline const char* zai_symbol_lookup_startof(zai_string_view *view) {
+    if (!view->len) {
+        return NULL;
+    }
+
+    if (memcmp(view->ptr, "\\", sizeof("\\")-1) == SUCCESS) {
```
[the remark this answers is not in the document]
krakjoe: na now

zend_abstract_interface/symbols/lookup.c
krakjoe: fixed that

bwoebi (approved):
Let's move forward and put off nice macro APIs for later.

PR #1455, opened 2022-01-11
ZAI SAPI Test Harness

krakjoe:
### Description

Test harness for ZAI SAPI and consumers.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zai_sapi/testing/catch2.hpp
+++ b/zai_sapi/testing/catch2.hpp
@@ -31,7 +31,7 @@
 *    Errors in stubs will cause the test to fail
 */
 extern "C" {
-#include "zai_sapi.h"
+#include "../zai_sapi.h"
 }
 
 #include <catch2/catch.hpp>
--- a/zai_sapi/testing/catch2.hpp
+++ b/zai_sapi/testing/catch2.hpp
@@ -49,60 +49,16 @@ extern "C" {
     "[" __ZAI_SAPI_TEST_CASE_SUITE "]" __ZAI_SAPI_TEST_CASE_TAGS /* }}} */
 
 /* {{{ public void ZAI_SAPI_TEST_CASE_BARE(char *suite, char *description, block_t code) */
-#define ZAI_SAPI_TEST_CASE_BARE(suite, description, code)                    \
+#define ZAI_SAPI_TEST_CASE_BARE(suite, description, ...)                     \
     TEST_CASE(ZAI_SAPI_TEST_CASE_TAG(                                        \
         suite, description, ZAI_SAPI_TEST_TAGS_NONE)) {                      \
-        { code }                                                             \
+        { __VA_ARGS__ }                                                      \
     } /* }}} */
 
 /* {{{ public void ZAI_SAPI_TEST_CASE_WITH_TAGS_BARE(char *suite, char *description, char *tags, block_t code) */
-#define ZAI_SAPI_TEST_CASE_WITH_TAGS_BARE(suite, description, tags, code)    \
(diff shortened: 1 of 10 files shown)


bwoebi:
These test harnesses also work fine with the existing one-off test harnesses like in uri_normalization, json, headers, exceptions, env, properties, config (i.e. basically all occurrences of `#define TEST*` in zai? I.e. does it fulfil all needs there or are there use cases missing?

If this works perfectly out, then great - though obviously, this approach has the disadvantage of hiding what's going on behind a couple additional layers of macros. But I assume this is an acceptable trade-off here.

krakjoe:
The decoration API covers the vast majority of use cases, this will be mostly used by consumers of the API (ie, zend_abstract_interface).

The bare API covers those use cases within this SAPI (and outside it too) where the decorator API does not fit - ie. we don't want to test in request stage, or we are testing some sapi API.

The PR to change tests in zend_abstract_interface will follow this one.

SammyK (commented):
Overall I love the idea.

I have a concern about this change landing in ZAI SAPI which has no hard dependency on Catch2 (in terms of the public API). I would like to keep the ZAI SAPI API decoupled from any specific test runners (Catch2 is good but it's also quite limiting). ZAI SAPI should be able to work well with other testing libraries without any issues.

I think the harness would be better suited either in the `zend_abstract_interface` directory, or offer a specific Catch2 flavored name for the header (like `test_harnesses/catch2.hpp` or something) and is only installed if Catch2 is enabled.

As Bob pointed out, there might be more cases that are ZAI-specific which makes me think a harness would be appropriate there. Perhaps it could piggy-back off of a generic Catch2 harness provided by ZAI SAPI.

krakjoe:
There is no hard dependency on catch2. We want to install the header regardless of the build - to pass the dependency onto the consumer, at the moment that's zai sapi tests, or zend_abstract_interface tests, which already depend on catch2.

Moving to any layer up from here doesn't make much sense to me, one of the driving forces here is to reduce the cognitive overhead of reading and writing tests, and we take maximum advantage of that reduction if everything is sharing a harness where possible.

krakjoe:
The possibility of providing more than one harness is maybe appealing, so I've created a "testing" directory and renamed the harness.

Should we add a README.md to testing directory explaining this is where we intend to keep harnesses and they should be named accordingly ? (I'm thinking someone somewhere might PR one, and this will save them 5 minutes maybe).

SammyK (approved):
Yes, these changes are what I was referring to (but I didn't articulate very well 😄). The generic `zai_sapi_test.hpp` header seemed like it created a hard dependency on Catch2 in my eyes. But having a separate Catch2 flavored header opens the SAPI up to support more testing libs later on. This looks great! 👍

zai_sapi/testing/catch2.hpp
```diff
@@ -0,0 +1,290 @@
...
+
+/* {{{ public void ZAI_SAPI_TEST_CASE_WITH_TAGS_BARE(char *suite, char *description, char *tags, block_t code) */
+#define ZAI_SAPI_TEST_CASE_WITH_TAGS_BARE(suite, description, tags, ...)     \
+    TEST_CASE(ZAI_SAPI_TEST_CASE_TAG(suite, description, tags)) {            \
+        { __VA_ARGS__ }                                                      \
+    } /* }}} */
+
+/* {{{ private bailout handling */
+#define __ZAI_SAPI_TEST_CASE_WITHOUT_BAILOUT_BEGIN()  \
+    bool zai_sapi_test_case_without_bailout = true;   \
```
SammyK, line 66: I just realized that all the `bool`s around the try/catches need to be marked as `volatile` so that they don't get clobbered by the long jump.

PR #1458, opened 2022-01-13
ZAI Test Harness Bailout Handling

krakjoe:
### Description

As discussed, we should have a way to differentiate between test case and test code bailout handling.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zai_sapi/testing/catch2.hpp
+++ b/zai_sapi/testing/catch2.hpp
@@ -48,11 +48,11 @@ extern "C" {
 #define ZAI_SAPI_TEST_STUB_NONE     NULL
 #define ZAI_SAPI_TEST_PROLOGUE_NONE {}
 
-/* {{{ ZAI_SAPI_TEST_CASE_NO_ASAN will hide a test when running under ASAN */
+/* {{{ ZAI_SAPI_TEST_TAG_NO_ASAN will hide a test when running under ASAN */
 #ifdef __SANITIZE_ADDRESS__
-# define ZAI_SAPI_TEST_CASE_NO_ASAN "[!hide]"
+# define ZAI_SAPI_TEST_TAG_NO_ASAN "[!hide]"
 #else
-# define ZAI_SAPI_TEST_CASE_NO_ASAN ""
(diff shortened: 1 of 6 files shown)


zai_sapi/testing/catch2.hpp
```diff
@@ -112,19 +176,26 @@ extern "C" {
...
+                zai_sapi_test_case_stub_included = true;                 \
+            zend_first_try {                                             \
                 zai_sapi_test_case_stub_included =                       \
                     zai_sapi_execute_script(                             \
                         __ZAI_SAPI_TEST_CASE_STUB);                      \
             } zend_catch {                                               \
                 zai_sapi_test_case_stub_included = false;                \
             } zend_end_try();                                            \
             REQUIRE(zai_sapi_test_case_stub_included);                   \
+            if (!zai_sapi_test_case_stub_included) {                     \
```
bwoebi, line 189: REQUIRE() aborts the program altogether, i.e. the goto in the branch will never be executed.

krakjoe: Specifically, it doesn't abort the program, it returns from the test case method/function ...

zai_sapi/testing/catch2.hpp
```diff
@@ -48,39 +65,169 @@ extern "C" {
...
+    bool zai_sapi_spinup() {
+        return zai_sapi_sinit() &&
+               zai_sapi_minit() &&
+               zai_sapi_rinit();
+    }
+
+    void zai_sapi_rshutdown() {
+        ::zai_sapi_rshutdown();
+
+        stage &= ~ZAI_SAPI_TEST_CASE_STAGE_REQUEST;
```
bwoebi, line 115: I think we should swap these lines, a double shutdown is probably much worse than a partial one (if there are shutdown some requires executed during shutdown or similar)

krakjoe: ack

bwoebi (approved):
Looks good to me, more stable testing :-)

PR #1460, opened 2022-01-14
Make consistent (working) ASAN builds

krakjoe:
### Description

CI and makefile builds were configured differently causing local asan builds to fail.

This change uses the toolchain file for both makefile and CI builds for ZAI SAPI and ZAI.

It drops the option from both ZAI and ZAI SAPI in favour of the working, more precise toolchain file that was already present.

Also fix dist_clean (another cause of build failures)


labbati (approved):
LGTM

PR #1461, opened 2022-01-16
Remove --version from dd-library-php-setup.php

morrisonlevi:
### Description

This is now version locked. If `--url` or `--file` are used, it will reject
versions which do not match and emit an error like:

> ERROR: Expected version '0.68.2', found '0.68.1'.

It's not yet clear to me how we should maintain this in master. Some
options include:
 - Adjust it to be an in-file e.g. datadog-php-setup.php.in that has
   some placeholder like `@DD_TRACE_VERSION@` that gets rewritten in
   the release process to the actual version e.g. `0.68.2`.
 - Adjust it to use the 'latest' release in master. This would either
   be hard-coded to the latest (bumped as part of our release process)
   or could use the string 'latest'. The latter would require parsing
   output or returned headers from github.

Right now, I lean toward the former approach.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/dd-library-php-setup.php
+++ b/dd-library-php-setup.php
@@ -14,9 +14,11 @@ const OPT_INSTALL_DIR = 'install-dir';
 const OPT_PHP_BIN = 'php-bin';
 const OPT_FILE = 'file';
 const OPT_URL = 'url';
-const OPT_VERSION = 'version';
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Closing in favor of #1463.

PR #1462, opened 2022-01-17
ZAI SAPI -> TEA

krakjoe:
### Description

This has turned out to be a rather large pull request, so I'll summarise the changes here, and do some explaining.

#### Motivation

The original motivation was renaming ZaiSapi to TeaSapi, but because this means considerable breakage for consumers, I've taken the opportunity to refactor the code and solve some problems along the way.

#### Source code organisation

The original structure of the code was imposed by the fact that this was a component in Zend Abstract Interface, since that is no longer true, and we intend to be able to ship this by itself, possibly from a different repository in the future, the same structure no longer makes sense.

New Structure (root directory):

  - `private` headers containing private symbols
  - `include` headers containing public (installed) symbols
  - `src` all source code
  - `tests` tests
  - `cmake` cmake build files

New structure (src view):

  - `sapi.c` contains everything related directly to the SAPI, symbols prefixed with `tea_sapi_`
  - `error.c` contains everything related to errors, symbols prefixed with `tea_error_`
  - `exceptions.c` contains everything related to exceptions, symbols prefixed with `tea_exception_`
  - `frame.c` contains everything related to frames, symbols prefixed with `tea_frame_`
  - `io.c` contains everything related to io, symbols prefixed with `tea_io_`
  - `ini.c` contains everything related to ini, symbols prefixed with `tea_ini_`
  - `extension.c` contains everything related to the builtin extension, symbols prefixed with `tea_extension_`

#### Extension API

We previously allowed the manipulation of a global module structure.

Instead the extension API can be used by internal (eg frame/error) and external components (eg tests) to install functions and stage handlers in the SAPI runtime.

See `include/extension.h` for API details, `src/extension.c` for implementation, and `tests/extension.cc` for tests.

#### Private API

The IO and INI API are entirely private - the only consumer is the SAPI, no headers for these functions are installed with Tea.

Other components may (and do) declare private API - only available during build.

#### Naming

While we agreed on the name Tea SAPI, it's extremely repetitive in code when everything - even components not directly related to the SAPI - contains the word `sapi`. For example, the exceptions code does not depend on a single detail of the SAPI, nor does the IO or INI code.

This problem would extend into writing clear documentation for this code.

So we have the name Tea for the library.


bwoebi (approved):
Assuming that everything in zend_abstract_interface is just renames (looks like that to me), this looks great to me. Thanks for moving forward with this :-)

SammyK:
Thanks for tackling the rename and for the improvements. 👍 

For those watching at home, Tea SAPI stands for **Testing and Embedded Applications SAPI**.

PR #1466, opened 2022-01-20
Clear the root span when dropped

bwoebi:
### Description

Will segfault (or at least leak) when datadog.trace.generate_root_span is switched from true to false if only a root span is present.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


labbati (approved):
LGTM!

PR #1467, opened 2022-01-20
fix symbols build and call issues

krakjoe:
### Description

When I tried to move usage of ZAI Symbols into the extension, it uncovered many build errors because the ext built has stricter Wflags.

In addition, the randomized testing framework uncovered a flaw in the logic of the call interface.

Doesn't need to go into release doc, follow up to #1452, depends on #1473

### DIFF
--- a/zend_abstract_interface/symbols/call.c
+++ b/zend_abstract_interface/symbols/call.c
@@ -23,26 +23,6 @@
  *
  */
 
-#if PHP_VERSION_ID >= 50400
-#define ZAI_SYMBOL_CALL_ZE2
-#endif
-
-#if PHP_VERSION_ID >= 70000
-#define ZAI_SYMBOL_CALL_ZE3
-#endif
-
-#if PHP_VERSION_ID >= 80000
-#define ZAI_SYMBOL_CALL_ZE4
-#endif
-
-#if defined(ZEND_MAP_PTR_INIT)
-#define ZAI_SYMBOL_CALL_MAP
-#endif
-
-#if PHP_VERSION_ID < 70300
-#define ZAI_SYMBOL_CALL_FCC_INITIALIZE
-#endif
-
 /* {{{ private arena code */
 typedef struct {
     char *base;
@@ -60,13 +40,13 @@ static inline void zai_symbol_call_arena_create(zai_symbol_call_arena_t *arena, 
     if (fbc->type == ZEND_USER_FUNCTION) {
         zai_symbol_call_arena_size += ZEND_MM_ALIGNED_SIZE(sizeof(zend_op_array));
 
-#if defined(ZAI_SYMBOL_CALL_ZE3)
+#if PHP_VERSION_ID >= 70000
         zai_symbol_call_arena_size += ZEND_MM_ALIGNED_SIZE(fbc->op_array.cache_size);
-#elif defined(ZAI_SYMBOL_CALL_ZE2)
+#elif PHP_VERSION_ID >= 50400
         zai_symbol_call_arena_size += ZEND_MM_ALIGNED_SIZE(fbc->op_array.last_cache_slot * sizeof(void *));
 #endif
 
-#if defined(ZAI_SYMBOL_CALL_MAP)
+#if PHP_VERSION_ID >= 70400
         zai_symbol_call_arena_size += ZEND_MM_ALIGNED_SIZE(sizeof(void *));
 #endif
     } else {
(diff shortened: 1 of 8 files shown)


zend_abstract_interface/symbols/call.c
```diff
@@ -2,14 +2,89 @@
 
 #include <ctype.h>
 #include <sandbox/sandbox.h>
-#include <zai_assert/zai_assert.h>
 
-static inline void zai_symbol_call_argv(zend_fcall_info *fci, uint32_t argc, va_list *args ZAI_TSRMLS_DC) {
+#if PHP_VERSION_ID >= 70300
```
krakjoe, line 6: The motivation for the use of an arena comes from Zend - it uses an arena. Because of implementation details however, we are not able to use it. First, Zend uses the compilers arena and updates the function table. We don't want to do that, so we use an arena we destroy after the call. Second, we extract added benefit - making the allocation of arguments quicker, the runtime cache invisible, and leaving the call faster.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -34,7 +34,7 @@ static inline bool zai_symbol_update(zend_class_entry *ce ZAI_TSRMLS_DC) {
     return false;
 }
 
-static inline void *zai_symbol_lookup_table(HashTable *table, zai_string_view key, bool ncase, bool pointer ZAI_TSRMLS_DC) {
+static inline void *zai_symbol_lookup_table(HashTable *table, zai_string_view key, bool ncase, bool pointer) {
```
krakjoe, line 37: unused TSRMLS_JUNK

zend_abstract_interface/symbols/lookup.c
```diff
@@ -67,6 +67,8 @@ static inline void *zai_symbol_lookup_table(HashTable *table, zai_string_view ke
     } else {
         return result;
     }
+#else
```
krakjoe, line 70: unused under 7, but this is a high frequency function and I didn't want to litter the callsites with ifdefs

zend_abstract_interface/symbols/lookup.c
```diff
@@ -78,7 +80,7 @@ static inline zai_string_view zai_symbol_lookup_clean(zai_string_view view) {
     return (zai_string_view){ .ptr = view.ptr + 1, .len = view.len - 1 };
 }
 
-static inline zai_string_view zai_symbol_lookup_key(zai_string_view *namespace, zai_string_view *name, bool lower ZAI_TSRMLS_DC) {
+static inline zai_string_view zai_symbol_lookup_key(zai_string_view *namespace, zai_string_view *name, bool lower) {
```
krakjoe, line 83: unused TSRMLS_JUNK

zend_abstract_interface/symbols/lookup.c
```diff
@@ -402,7 +406,7 @@ static inline zval* zai_symbol_lookup_local_frame(zend_execute_data *ex, zai_str
     for (int var = 0; var < function->op_array.last_var; var++) {
         zend_compiled_variable *match = &function->op_array.vars[var];
 
-        if ((name->len == match->name_len) && (memcmp(name->ptr, match->name, name->len) == SUCCESS)) {
+        if ((name->len == (size_t) match->name_len) && (memcmp(name->ptr, match->name, name->len) == SUCCESS)) {
```
krakjoe, line 409: mismatch sign

zend_abstract_interface/symbols/call.c
```diff
@@ -97,24 +221,56 @@ bool zai_symbol_call_impl(
...
 
+#ifdef ZAI_SYMBOL_CALL_ARENA
+    zai_symbol_call_arena_release(&arena);
+#else
     if (argc) {
         efree(fci.params);
     }
+#endif
+
+    if (zai_symbol_call_bailed) {
```
krakjoe, line 270: we must bailout

zend_abstract_interface/symbols/call.c
krakjoe: I added a comment.

zend_abstract_interface/symbols/call.c
```diff
@@ -96,40 +341,61 @@ bool zai_symbol_call_impl(
...
+    zend_try {
+        zai_symbol_call_arena_create(&arena, &fcc, argc);
+    } zend_catch {
+        zai_symbol_call_bailed = true;
+    } zend_end_try();
+
+    if (zai_symbol_call_bailed) {
+        zai_sandbox_bailout(&sandbox);
+        zai_sandbox_close(&sandbox);
+        return false;
```
bwoebi, line 362: Nit: zai_symbol_call_bailed var is redundant, can be moved directly into the catch block

zend_abstract_interface/symbols/call.c
```diff
@@ -145,11 +411,12 @@ bool zai_symbol_new(zval *zv, zend_class_entry *ce ZAI_TSRMLS_DC, uint32_t argc,
             argc, &args);
         // clang-format on
 
-#if PHP_VERSION_ID < 70000
-        zval_ptr_dtor(&rv);
-#else
+#if defined(ZAI_SYMBOL_CALL_ZE3)
```
bwoebi, line 414: This is ZE3 and ZE4? not ZE3 exclusive? looks wrong to me

bwoebi (approved):

PR #1472, opened 2022-01-23
Support in-tree builds of Tea

krakjoe:
### Description

This allows tea to be built in-tree as a normal sapi without dependence on embed.

Supports shared/static builds, installs headers in same way as cmake based build ...

We should maybe ship a cmake, or additional autoconf tooling ...

We should probably include a simple CI job to make sure we don't break this ability in some strange way ... but later ...

to test `cp -r dd-trace-php/tea /path/to/php-src/sapi && ./buildconf --force && ./configure --enable-tea`

### Readiness checklist

### DIFF
--- a/tea/config.m4
+++ b/tea/config.m4
@@ -1,0 +1,71 @@
+PHP_ARG_ENABLE([tea],, [--enable-tea=shared|static|yes|no], [no], [no])
+
+if test "$PHP_TEA" != "no"; then
+  AC_MSG_CHECKING([Tea build configuration])
+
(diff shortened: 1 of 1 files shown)


SammyK (approved):
LGTM 👍 Ideally we wouldn't need to manage two build configurations, but I don't think there will be a way around that for a long time.

Once Tea SAPI gets to a mostly feature complete API (I think it's pretty close as is) it will be nice to make this part of the regular PHP builds in our containers; and also to remove embed.

PR #1478, opened 2022-01-25
Patch zend_alloc.c to not crash in CI for PHP 7.0 to 7.2

bwoebi:
Makes CI green again.

### DIFF
--- a/dockerfiles/ci/buster/php-7.1/Dockerfile
+++ b/dockerfiles/ci/buster/php-7.1/Dockerfile
@@ -9,13 +9,15 @@ ENV PHP_VERSION=${phpVersion}
 FROM base as build
 ARG phpTarGzUrl
 ARG phpSha256Hash
+COPY 0001-Fixed-incorrect-behavior-of-internal-memory-debugger.patch /home/circleci
 RUN set -eux; \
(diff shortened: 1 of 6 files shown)


dockerfiles/ci/buster/php-7.1/Dockerfile
```diff
@@ -72,7 +74,7 @@ RUN set -eux; \
         pecl install mcrypt-1.0.0; echo "extension=mcrypt.so" >> ${iniDir}/mcrypt.ini; \
         pecl install ast; echo "extension=ast.so" >> ${iniDir}/ast.ini; \
         yes 'no' | pecl install memcached; echo "extension=memcached.so" >> ${iniDir}/memcached.ini; \
-        pecl install mongodb; echo "extension=mongodb.so" >> ${iniDir}/mongodb.ini; \
+        pecl install mongodb-1.11.1; echo "extension=mongodb.so" >> ${iniDir}/mongodb.ini; \
```
morrisonlevi, line 77: I would expect to see a similar line in other Dockerfiles -- it seems odd to have this in only 7.1, but not 7.0 nor 7.2.

morrisonlevi: Ah, I see, it was previously pinned for 7.0.

PR #1479, opened 2022-01-25
Tea: build once during development

krakjoe:
### Description

Since we added install_tea as a dependency of zai builds, tea is built over and over again during development, slowing things down considerably.

We want tea to be rebuilt for test runs, but for everything else, we don't need to rebuild.

I've added guards to tea rules so that they run once, and clean when they have too.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


labbati (approved):
Good idea. LGTM! Thanks @krakjoe

PR #1483, opened 2022-01-27
BGS shutdown/flush

krakjoe:
### Description

The background sender is configured with a shutdown timeout, which defines the maximum amount of time the shutdown routine will wait for the writer thread before the routine ignores the request to join.

This is not safe, threads must be joined for predictable behaviour.

In the case where the timeout is reached, the shutdown routine needs to cancel the writer thread and then join with it at the last opportunity.

Failing to fulfil the contract of the function at the last opportunity leads to surprising behaviour, like the background sender trying to read from memory that was freed at mshutdown stage.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php5/coms.c
+++ b/ext/php5/coms.c
@@ -828,6 +828,17 @@ static void _dd_signal_data_processed(struct _writer_loop_data_t *writer) {
 #define TIMEOUT_SIG SIGPROF
 #endif
 
+static void _dd_writer_loop_cleanup(void *ctx) {
+    struct _writer_loop_data_t *writer = (struct _writer_loop_data_t *)ctx;
+
+    curl_slist_free_all(writer->headers);
+    writer->headers = NULL;
+
+    curl_easy_cleanup(writer->curl);
+    _dd_coms_stack_shutdown();
+    _dd_signal_writer_finished(writer);
+}
+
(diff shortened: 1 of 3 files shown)


bwoebi (dismissed):
This looks safe to me, thanks for the patch Joe :-)

bwoebi (approved):

morrisonlevi:
In the longer term, there are better ways to share data and clean up. However, this will be a more invasive change, so it's off the table for now.

I didn't know about these functions before. From reading the docs, this bit does have me concerned:

> Calling longjmp(3) (siglongjmp(3)) produces undefined results if
> any call has been made to pthread_cleanup_push() or
> pthread_cleanup_pop() without the matching call of the pair since
> the jump buffer was filled by setjmp(3) (sigsetjmp(3)).

bwoebi:
Though the BGS (nor curl https://github.com/curl/curl/search?q=longjmp is empty) is not using the zend_try nor the *jmp() API directly, so that should be fine.

morrisonlevi:
I won't block the PR -- I just don't know enough about these APIs to do that. If you two think it's better this way for the short-term and that it will handle the recent ASAN failure in CI, then go ahead.

PR #1486, opened 2022-01-28
Unmark PDO reconnect spans as an error if the (same) connection is immediately restarted afterwards

bwoebi:
### Description

I'm not sure whether that's the correct fix, or whether we should fix this at the eloquent level only. But given that Eloquent (or more precisely the laravel database backend) supports reconnecting natively, it should not register an error if it reconnects afterwards.

### Readiness checklist
- [ ] Tests added for this feature/bug. - No idea how to emulate a connection goes away...? Testing locally I killed the docker container with mysql inside...

### Reviewer checklist
- [ ] Milestone is set.


labbati (commented):
I am not convinced that what we are doing in the PR is the right solution. If the prepare fails then I think it is an error and we should report it.

Why someone would not be interested in knowing that mysql went away from the application point of view? Monitoring the server itself, by means of a core integration, it is not enough, because the issue might be in the network between the application and the server.

What sort of false positive error reporting the user complained about?

bwoebi:
Marking as do not merge for now, it needing more internal conversation first.

bwoebi:
We decided to not go with this approach. Closing it.

PR #1489, opened 2022-01-31
Improve error detection for datadog-setup.php

### LINKED ISSUE #1488: [BUG] datadog-setup.php in 0.69.0 behaves strangely for unrecognized parameters
### Bug description

For instance, if you run this with 0.69.0's datadog-setup.php, then it
goes into interactive mode:

```
php datadog-setup.php 0.69.0 --php-bin  /usr/sbin/php-fpm7.4 --enable-profiling
Searching for available php binaries, this operation might take a while.
Multiple PHP binaries detected. Please select the binaries the datadog library will be installed to:

   1. php --> /usr/bin/php7.4
   2. php7.4 --> /usr/bin/php7.4
   3. php-fpm7.4 --> /usr/sbin/php-fpm7.4

Select binaries using their number. Multiple binaries separated by space (example: 1 3): 
```

This happens for any unrecognized parameter. It should error on
unrecognized parameters, or perhaps just give a warning. It shouldn't
enter interactive mode.

### Additional Context

In this case, a customer was passing `--version 0.69.0` because they
upgraded from a previous version which required that, and when I
pointed out they should remove `--version`, they left `0.69.0` because
`datadog-setup.php --help` says it's required. I opened [a PR](https://github.com/DataDog/dd-trace-php/pull/1487) to fix the
help command.

#### Tracer version
0.69.0

morrisonlevi:
### Description

Fix #1488. It is not perfect, but it is better. It catches things like these:

    php datadog-setup.php --version 0.69.0 --enable-profiling
    php datadog-setup.php 0.69.0 --enable-profiling

But if the error happens after the first argument, then it won't catch it e.g.

    php datadog-setup.php --enable-profiling 0.69.0 --php-bin php-fpm

Will go into interactive mode because `getopt` stopped parsing at `0.69.0`.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -444,6 +444,19 @@ function parse_validate_user_options()
     ];
     $options = getopt($shortOptions, $longOptions);
 
+    global $argc;
+    if ($options === false || (empty($options) && $argc > 1)) {
+        /* Note that the above conditions are not as robust as I'd like.
+         * Consider:
+         *   php datadog-setup.php --enable-profiling 0.69.0 --php-bin php
+         * getopt will stop at 0.69.0 as it doesn't recognize it, but it will
+         * return an array that only has enable-profiling in it.
+         * I don't see an obvious way out of this, but catching some failures
+         * here is better than not catching any.
+         */
+        print_error_and_exit("Failed to parse options", true);
+    }
+
     // Help and exit
     if (key_exists('h', $options) || key_exists(OPT_HELP, $options)) {
         print_help();


datadog-setup.php
```diff
@@ -444,6 +444,19 @@ function parse_validate_user_options()
     ];
     $options = getopt($shortOptions, $longOptions);
 
+    global $argc;
```
labbati, line 447: I don't think there is much better at this time we can do. Even in this case, though, we should probably print the help message. WDYT?

labbati (changes requested):

datadog-setup.php
morrisonlevi: It should print the help message because of the `print_error_and_exit(..., true)`, I think?

labbati: I had totally forgotten of that option that I had recently added 😄

labbati (approved):
LGTM

PR #1498, opened 2022-02-04
Implement split by instance in PDO

labbati:
### Description

This feature implements split by instance for PDO spans, allowing to differentiate between different host/servers based on service name.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/PDO/PDOIntegration.php
+++ b/src/Integrations/Integrations/PDO/PDOIntegration.php
@@ -37,8 +37,11 @@ class PDOIntegration extends Integration
         // public PDO::__construct ( string $dsn [, string $username [, string $passwd [, array $options ]]] )
         \DDTrace\trace_method('PDO', '__construct', function (SpanData $span, array $args) {
             $span->name = $span->resource = 'PDO.__construct';
-            PDOIntegration::storeConnectionParams($this, $args);
-            PDOIntegration::setCommonSpanInfo($this, $span);
+            $connectionMetadata = PDOIntegration::extractConnectionMetadata($args);
+            ObjectKVStore::put($this, PDOIntegration::CONNECTION_TAGS_KEY, $connectionMetadata);
+            // We have to use $connectionMetadata as a medium, instead of $this (aka the PDO instance) because in
+            // PHP 5.* $this is NULL in this callback when there is a connection error.
+            PDOIntegration::setCommonSpanInfo($connectionMetadata, $span);
         });
 
         // public int PDO::exec(string $query)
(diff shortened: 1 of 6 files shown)


src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -72,29 +69,24 @@ public function init()
                 $span->meta = [
                     'db.rowcount' => $retval->rowCount(),
                 ];
-                PDOIntegration::storeStatementFromConnection($this, $retval);
```
labbati, line 75: @bwoebi about your comment on this. In this case works because it returns an "executed" statement and we only trace `execute` on `PDOStatement` (which is not called in this case). But thinking more about it, I'd say that it makes sense to keep it in sync in any case. Say that in 1 month we want to trace `fetchAll()` (which might not make sense since it is only an in-memory operation), then we would have a bug in case we do not assert all the fields.

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -179,31 +175,42 @@ private static function parseDsn($dsn)
...
         if (isset($constructorArgs[1])) {
             $tags['db.user'] = $constructorArgs[1];
         }
-        ObjectKVStore::put($pdo, PDOIntegration::CONNECTION_TAGS_KEY, $tags);
         return $tags;
     }
 
-    public static function storeStatementFromConnection($pdo, $stmt)
+    /**
+     * @param PDO|PDOStatement|array $pdoOrStmt
```
bwoebi, line 188: leftover varname

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -179,31 +175,42 @@ private static function parseDsn($dsn)
...
+        if (\is_array($source)) {
+            $storedConnectionInfo = $source;
+        } elseif (
+            \is_object($source)
+            && (\is_a($source, 'PDO') || \is_a($source, 'PDOStatement'))
+        ) {
+            $storedConnectionInfo = ObjectKVStore::get($source, PDOIntegration::CONNECTION_TAGS_KEY, []);
+        } else {
+            $storedConnectionInfo = [];
+        }
```
bwoebi, line 202:
```suggestion
        if (\is_array($source)) {
            $storedConnectionInfo = $source;
        } else {
            $storedConnectionInfo = ObjectKVStore::get($source, PDOIntegration::CONNECTION_TAGS_KEY, []);
        }
        if (!\is_array($storedConnectionInfo)) {
            $storedConnectionInfo = [];
        }
```
I think the code is simpler that way

bwoebi (approved):
Looks good to me, my comments aren't blocking

src/Integrations/Integrations/PDO/PDOIntegration.php
labbati: I like this, I did not remember that objectkvstore does a lot of checks on its own.

bwoebi (approved):

PR #1499, opened 2022-02-04
Trigger profiling interrupt function before internal observer_end

morrisonlevi:
### Description

If the observer_end is called before the profiler processes an
interrupt, then the profiler mis-attributes time taken to the tracing
function, instead of the real leaf function. This change pairs with one
in the profiler that exposes `datadog_profiling_interrupt_function`.

Here's the profiling function which ends up being called by
 `datadog_profiling_interrupt_function`:
[datadog_php_stack_collector_interrupt_function](https://github.com/DataDog/dd-prof-php/blob/ed9b20cd4bf15e7dad0438cd9341d6c96d576123/profiling/plugins/stack_collector_plugin/stack_collector_plugin.c#L306). It is safe to call this
function even when the profiler is disabled.

I wrote a simple "sleep" app + integration to demonstrate the bug:

```php
<?php

if (function_exists('DDTrace\trace_function')) {
  DDTrace\trace_function(
    'sleep',
    function ($span, $args) {
      $span->name = 'sleep';
      if (!empty($args[0])) {
        $span->resource = (string)$args[0];
      }
      $span->type = 'custom';
    }
  );
}

function main() {
  sleep(1);
}
main();
```

Before the change:
app datadoghq com_profiling_search_query=service%3Aunnamed-service agg_m=%40prof_php_wall_time agg_q=service agg_t=avg event=AQAAAX7GZy6wthM-QAAAAABBWDdHWnpIOUFBQi1yYkZiRlZWZThnQUE profileId=AX7GZzH9AAB-rbFbFVVe8gAA start=1644004621742  (2)

As you can see, the time of the sleep is mis-attributed to the closure.
After the change:
app datadoghq com_profiling_search_query=service%3Aunnamed-service agg_m=%40prof_php_wall_time agg_q=service agg_t=avg event=AQAAAX7GZy6wthM-QAAAAABBWDdHWnpIOUFBQi1yYkZiRlZWZThnQUE profileId=AX7GZzH9AAB-rbFbFVVe8gAA start=1644004621742  (3)

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php7/php7/engine_hooks.c
+++ b/ext/php7/php7/engine_hooks.c
@@ -3,7 +3,9 @@
 #include <Zend/zend_closures.h>
 #include <Zend/zend_compile.h>
 #include <Zend/zend_exceptions.h>
+#include <Zend/zend_extensions.h>
 #include <Zend/zend_interfaces.h>
+#include <Zend/zend_portability.h>
 #include <exceptions/exceptions.h>
 #include <stdbool.h>
 #include <symbols/symbols.h>
@@ -1060,10 +1062,29 @@ void ddtrace_opcode_mshutdown(void) {
 void ddtrace_execute_internal_minit(void) {}
 void ddtrace_execute_internal_mshutdown(void) {}
 
-/* If the profiler is loaded, then this weak symbol will be overwritten by the
- * strong symbol provided by profiler.
+static void (*profiling_interrupt_function)(zend_execute_data *) = NULL;
+
+/**
+ * The message handler is used to determine if the profiler is loaded, and if
+ * so it will locate certain symbols so cross-product features can be enabled.
  */
-__attribute__((weak)) extern void datadog_profiling_interrupt_function(zend_execute_data *execute_data);
+void ddtrace_message_handler(int message, void *arg) {
+    if (UNEXPECTED(message != ZEND_EXTMSG_NEW_EXTENSION)) {
+        // There are currently no other defined messages.
+        return;
+    }
+
+    zend_extension *extension = (zend_extension *)arg;
+    if (extension->name && strcmp(extension->name, "datadog-profiling") == 0) {
+        DL_HANDLE handle = extension->handle;
+
+        profiling_interrupt_function = DL_FETCH_SYMBOL(handle, "datadog_profiling_interrupt_function");
+        if (UNEXPECTED(!profiling_interrupt_function)) {
+            ddtrace_log_debugf("[Datadog Trace] Profiling was detected, but locating symbol %s failed: %s\n",
+                               "datadog_profiling_interrupt_function", DL_ERROR());
+        }
+    }
+}
 
 PHP_FUNCTION(ddtrace_internal_function_handler) {
     ddtrace_dispatch_t *dispatch;
@@ -1084,8 +1105,8 @@ PHP_FUNCTION(ddtrace_internal_function_handler) {
          * give the profiler an opportunity to take a sample before calling the
          * tracing funcation.
          */
-        if (datadog_profiling_interrupt_function) {
-            datadog_profiling_interrupt_function(execute_data);
+        if (profiling_interrupt_function) {
+            profiling_interrupt_function(execute_data);
         }
         dd_observer_end(EX(func), span_fci, return_value);
     }
--- a/ext/php7/php7/engine_hooks.c
+++ b/ext/php7/php7/engine_hooks.c
@@ -1063,7 +1063,7 @@ void ddtrace_execute_internal_mshutdown(void) {}
 /* If the profiler is loaded, then this weak symbol will be overwritten by the
  * strong symbol provided by profiler.
  */
-extern void datadog_profiling_interrupt_function(zend_execute_data *execute_data) __attribute__((weak));
+__attribute__((weak)) extern void datadog_profiling_interrupt_function(zend_execute_data *execute_data);
 
 PHP_FUNCTION(ddtrace_internal_function_handler) {
     ddtrace_dispatch_t *dispatch;
--- a/ext/php7/php7/engine_hooks.c
+++ b/ext/php7/php7/engine_hooks.c
@@ -1060,6 +1060,11 @@ void ddtrace_opcode_mshutdown(void) {
 void ddtrace_execute_internal_minit(void) {}
 void ddtrace_execute_internal_mshutdown(void) {}
 
+/* If the profiler is loaded, then this weak symbol will be overwritten by the
+ * strong symbol provided by profiler.
+ */
+extern void datadog_profiling_interrupt_function(zend_execute_data *execute_data) __attribute__((weak));
+
 PHP_FUNCTION(ddtrace_internal_function_handler) {
     ddtrace_dispatch_t *dispatch;
     void (*handler)(INTERNAL_FUNCTION_PARAMETERS) = EX(func)->internal_function.reserved[ddtrace_resource];
@@ -1072,6 +1077,16 @@ PHP_FUNCTION(ddtrace_internal_function_handler) {
     ddtrace_span_fci *span_fci = dd_observer_begin(execute_data, dispatch);
     handler(INTERNAL_FUNCTION_PARAM_PASSTHRU);
     if (span_fci) {
+        /* If the profiler doesn't handle a potential pending interrupt before
+         * the observer's end function, then the callback will be at the top of
+         * the stack even though it's not responsible.
+         * This is why the profiler's interrupt function is called here, to
+         * give the profiler an opportunity to take a sample before calling the
+         * tracing funcation.
+         */
+        if (datadog_profiling_interrupt_function) {
+            datadog_profiling_interrupt_function(execute_data);
+        }
         dd_observer_end(EX(func), span_fci, return_value);
     }
 }
--- a/ext/php8/php8/engine_hooks.c
+++ b/ext/php8/php8/engine_hooks.c
@@ -758,6 +758,11 @@ zend_observer_fcall_handlers ddtrace_observer_fcall_init(zend_execute_data *exec
     return (zend_observer_fcall_handlers){NULL, NULL};
 }
 
+/* If the profiler is loaded, then this weak symbol will be overwritten by the
+ * strong symbol provided by profiler.
+ */
+extern void datadog_profiling_interrupt_function(zend_execute_data *execute_data) __attribute__((weak));
+
 PHP_FUNCTION(ddtrace_internal_function_handler) {
     ddtrace_dispatch_t *dispatch;
     zend_function *fbc = EX(func);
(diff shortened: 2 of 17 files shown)


krakjoe:
Just a nit, can we put attributes before the prototype ... I missed it on the first glance ...

ext/php7/php7/engine_hooks.c
```diff
@@ -1060,6 +1060,11 @@ void ddtrace_opcode_mshutdown(void) {
 void ddtrace_execute_internal_minit(void) {}
 void ddtrace_execute_internal_mshutdown(void) {}
 
+/* If the profiler is loaded, then this weak symbol will be overwritten by the
+ * strong symbol provided by profiler.
+ */
+__attribute__((weak)) extern void datadog_profiling_interrupt_function(zend_execute_data *execute_data);
```
sanchda, line 1066: This is a convention I've used before myself, but I was curious about the degree to which undefined symbols are initialized to zero.  I put together an MVP of the following form:
```
$ gcc --shared -fPIC strong_symbol.c -o libstrong.so
$ gcc test.c -o main -L. -lstrong
$ cat strong_symbol.c
int weak_symbol(void* arg) { return 5; }
$ cat test.c
__attribute__((weak)) extern void weak_symbol(void* arg);
int foo() { return !!weak_symbol; }
int main() { return foo(); }

$ LD_PRELOAD=./libstrong.so ./main; echo $?
1
$ ./main; echo $?
0
```
I tried this test across compiler versions, with various link-time and compile-time flags.  I inspected the disassembly and the ELF header to confirm that `weak_symbol()` was residing by default in the COMMON section, and that it was being initialized to zero.

Basically, on contemporary Linux, I don't see any way that this would fail.  I do not know whether the implementation of symbol visibility is identical across platforms--I vaguely recall that Linux actually changed _something_ at some point because it conformed to what something random, like IRIX, did rather than SystemV, but I don't remember exactly.

Anyway, this is just a long-winded way of saying I checked this and it looks OK.

sanchda: Let me substantiate that comment a bit more.  I left a lot out for the sake of brevity.  If you disassemble the function in the MVP,
```
0000000000001129 <foo>:
    1129:       f3 0f 1e fa             endbr64
    112d:       55                      push   %rbp
    112e:       48 89 e5                mov    %rsp,%rbp
    1131:       48 8b 05 b0 2e 00 00    mov    0x2eb0(%rip),%rax        # 3fe8 <weak_symbol>
    1138:       48 85 c0                test   %rax,%rax
    113b:       74 07                   je     1144 <foo+0x1b>
    113d:       b8 01 00 00 00          mov    $0x1,%eax
    1142:       eb 05                   jmp    1149 <foo+0x20>
    1144:       b8 02 00 00 00          mov    $0x2,%eax
    1149:       5d                      pop    %rbp
    114a:       c3                      ret
```

You see that `weak_symbol` is loaded and checked.  It isn't obvious from that fragment, but `0x3fe8` is outside of the executable region of the code.  In fact, it's in none of the program headers.  You might have to believe me, since I stared hard at the offsets to come to that conclusion.

```
$ readelf --segments -W ./a.out

Elf file type is DYN (Shared object file)
Entry point 0x1040
There are 13 program headers, starting at offset 64

Program Headers:
  Type           Offset   VirtAddr           PhysAddr           FileSiz  MemSiz   Flg Align
  PHDR           0x000040 0x0000000000000040 0x0000000000000040 0x0002d8 0x0002d8 R   0x8
  INTERP         0x000318 0x0000000000000318 0x0000000000000318 0x00001c 0x00001c R   0x1
      [Requesting program interpreter: /lib64/ld-linux-x86-64.so.2]
  LOAD           0x000000 0x0000000000000000 0x0000000000000000 0x000610 0x000610 R   0x1000
  LOAD           0x001000 0x0000000000001000 0x0000000000001000 0x0001e5 0x0001e5 R E 0x1000
  LOAD           0x002000 0x0000000000002000 0x0000000000002000 0x000158 0x000158 R   0x1000
  LOAD           0x002de8 0x0000000000003de8 0x0000000000003de8 0x000228 0x000230 RW  0x1000
  DYNAMIC        0x002df8 0x0000000000003df8 0x0000000000003df8 0x0001c0 0x0001c0 RW  0x8
  NOTE           0x000338 0x0000000000000338 0x0000000000000338 0x000030 0x000030 R   0x8
  NOTE           0x000368 0x0000000000000368 0x0000000000000368 0x000044 0x000044 R   0x4
  GNU_PROPERTY   0x000338 0x0000000000000338 0x0000000000000338 0x000030 0x000030 R   0x8
  GNU_EH_FRAME   0x002004 0x0000000000002004 0x0000000000002004 0x000044 0x000044 R   0x4
  GNU_STACK      0x000000 0x0000000000000000 0x0000000000000000 0x000000 0x000000 RW  0x10
  GNU_RELRO      0x002de8 0x0000000000003de8 0x0000000000003de8 0x000218 0x000218 R   0x1
 ```
 
 Instead, `0x3fe8` resides within one of the data sections zero-initialized at startup.  It's probably overridden at link-time, but aside from observing that it works, I don't know which standard/phase governs that.

**Review by sanchda — DISMISSED** [review]
Looks good to me!

File: ext/php7/php7/engine_hooks.c (hunk 1 above)
Comment (morrisonlevi) [reply ext/php7/php7/engine_hooks.c:1066]: Thanks for the dive, David. I am adding a test and I'm sadly not getting the behavior I hoped for. The setup doesn't match your test.

1. I have `interrupt.cc` which defines a strong `datadog_profiling_interrupt_function` and has `main`. 
2. The `ddtrace.so` extension has a weak `datadog_profiling_interrupt_function`. 
3. At runtime, the executable built from `interrupt.cc` will `dlopen` the `ddtrace.so` extension.
4. When I call the function in `ddtrace.so` which ends up calling `datadog_profiling_interrupt_function`, it doesn't call the strong version from `interrupt.cc`. 

So I've been searching around. The `ld.so` man page has a sub-section on `LD_DYNAMIC_WEAK` under the `Environment Variables` section which includes:

> Old glibc versions (before 2.2), provided a different
behavior: if the linker found a symbol that was weak, it
would remember that symbol and keep searching in the
remaining shared libraries.  If it subsequently found a
strong definition of the same symbol, then it would
instead use that definition.  (If no further symbol was
found, then the dynamic linker would use the weak symbol
that it initially found.)

> The old glibc behavior was nonstandard.  _(Standard
practice is that the distinction between weak and strong
symbols should have effect only at static link time.)_  In
glibc 2.2, the dynamic linker was modified to provide the
current behavior (which was the behavior that was provided
by most other implementations at that time).

Emphasis mine. Apparently weak and strong symbols are only supposed to matter at static link time. This isn't what I've observed in other ad-hoc tests, and not what you observed either... and I'm using a glibc version definitely newer than v2.2. I'm not sure what's controlling this behavior we're seeing...

Comment (sanchda) [reply ext/php7/php7/engine_hooks.c:1066]: Ah, gotcha.  Let me see if I have the setup right.

```
$ cat lib.c
//__attribute__((weak)) extern int notstrong();
int notstrong() { return 2; }
extern int entrypoint() { return notstrong(); }

$ cat main.c
extern int entrypoint();
int notstrong() { return 1; } // actually not weak
int main() { return entrypoint(); }

$ gcc -fPIC -shared lib.c -o libmylib.so
$ gcc -L. main.c -o main -lmylib
LD_LIBRARY_PATH=. ./main; echo $?
1
```
AFAIK, the behavior of `dlopen()` matches the dynamic linker here, since we're talking about symbol visibility from within the .so.

I assume that making `notstrong()` purely internal isn't an option (otherwise you would have just renamed it or something), since marking `notstrong()` in lib.c as `inline static int` makes the return work correctly.

One thing to note is that the obvious indirection alone isn't enough:
```
__attribute__((weak)) extern int notstrong();
int notstrong() { return 2; }
int (*notstrong_impl)() = notstrong;
extern int entrypoint() { return notstrong_impl(); }
```
will still set `notstrong_impl()` to the `notstrong()` defined in `main.c`.

Instead, you'd need something like,
```
__attribute__((weak)) extern int notstrong();
int notstrong_impl() { return 2; }
int notstrong() { return notstrong_impl(); }
extern int entrypoint() { return notstrong_impl(); }
```
which does work.  Would that work?

File: ext/php8/php8/engine_hooks.c [hunk 2]
```diff
@@ -758,10 +760,29 @@ zend_observer_fcall_handlers ddtrace_observer_fcall_init(zend_execute_data *exec
...
+    if (UNEXPECTED(message != ZEND_EXTMSG_NEW_EXTENSION)) {
+        // There are currently no other defined messages.
+        return;
+    }
+
+    zend_extension *extension = (zend_extension *)arg;
+    if (extension->name && strcmp(extension->name, "datadog-profiling") == 0) {
+        DL_HANDLE handle = extension->handle;
+
+        profiling_interrupt_function = DL_FETCH_SYMBOL(handle, "datadog_profiling_interrupt_function");
```
sanchda, line 779: I think this is good--does the behavior of `DL_FETCH_SYMBOL()` around `DLSYM_NEEDS_UNDERSCORE` matter for you?

sanchda (dismissed):
Very nice.  LGTM!

tests/tea/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  message(FATAL_ERROR "TEA is required but not found")
+endif()
+
+add_library(ext-tea-tests INTERFACE)
+target_link_libraries(ext-tea-tests INTERFACE Catch2::Catch2WithMain Tea::Tea)
+
+# Need components, for example stack-sample.
+add_subdirectory(../../components ${CMAKE_CURRENT_BINARY_DIR}/components)
+
+add_subdirectory(profiling)
```
bwoebi, line 42: Doesn't it make more sense to completely exclude profiling with some version if check here?
https://github.com/DataDog/dd-trace-php/blob/7ace0ebe53278d783d0b1759328d9bfe742b8247/zend_abstract_interface/CMakeLists.txt#L104

Rather than doing an #if in the code?

ext/php8/php8/engine_hooks.c
bwoebi: I wondered about the underscore prefix as well, but I assume given the way the symbols are exported on the profiler (`datadog-profiling.sym`) it never will be mangled.

tests/tea/CMakeLists.txt
morrisonlevi: Adjusted this and the CI version skip stuff as well.

SammyK (approved):
Nice work adding the proper testing infrastructure for this. This was the last missing piece that was required to test `ext/` with Tea SAPI. 💯 

And let's give a big shout-out to our dear friend `dlsym` who consistently saves our bacon on symbols [time](https://github.com/DataDog/dd-trace-php/blob/7db9b0c/zend_abstract_interface/json/json.c#L42-L50) and [time](https://github.com/DataDog/dd-trace-php/blob/97b7233/ext/php8/handlers_curl.c#L349-L357) again. 😄 🎉 

LGTM 👍

PR #1503, opened 2022-02-10
Interceptor implementation

bwoebi:
### Description

There's two parts to the PR:
 - resolving, i.e. hooking into the places where new named functions or classes may be added to the symbol tables (resolve.c(c))
 - intercepting, i.e. calling hook start if hooked and hook finish if started (interceptor.c(c))

There's also a (test passing) version of PHP 8, but it still needs some love, see TODOs

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


zend_abstract_interface/interceptor/php7/interceptor.c
```diff
@@ -0,0 +1,580 @@
...
+
+    zai_interceptor_frame_memory frame_memory;
+    if (zai_hook_continue(execute_data, &frame_memory.hook_data)) {
+        frame_memory.execute_data = execute_data;
+        zai_hook_memory_table_insert(execute_data, &frame_memory);
+    }
+#endif
+
+    zai_interceptor_generator_handlers = (zend_object_handlers *)generator->std.handlers;
+    pthread_once(&zai_interceptor_replace_generator_dtor_once, zai_interceptor_replace_generator_dtor);
```
bwoebi, line 398: Compare this approach with:
https://github.com/DataDog/dd-trace-php/blob/e767a54f3113b2e886e07668f7749d079fa4549b/zend_abstract_interface/interceptor/php8/interceptor.c#L188-L205

I assume the other (in PHP 8 file) approach is probably a tiny bit safer and saving us the pthreads. We should decide for one specific approach across both versions though.
Other suggestions or comments welcome.

tea/include/extension.h
```diff
@@ -35,7 +35,11 @@ typedef zend_result zend_result_t;
 #endif
 
 typedef zend_result_t (*tea_extension_init_function)(INIT_FUNC_ARGS);
-typedef zend_result_t (*tea_extension_shutdown_function)(SHUTDOWN_FUNC_ARGS); /* }}} */
+typedef zend_result_t (*tea_extension_shutdown_function)(SHUTDOWN_FUNC_ARGS);
+typedef void (*tea_extension_op_array_function)(zend_op_array *op_array);
+typedef zend_result_t (*tea_extension_startup_function)(void); /* }}} */
+
+zend_module_entry* tea_extension_module();
```
krakjoe, line 42: This should not be leaked.

zend_abstract_interface/hook/hook.c
```diff
@@ -338,8 +325,9 @@ bool zai_hook_rinit(void) {
 }
 
 void zai_hook_rshutdown(void) {
-    zend_hash_destroy(&zai_hook_resolved);
     zend_hash_destroy(&zai_hook_request);
+    zend_hash_destroy(&zai_hook_resolved);
+    zai_hook_resolved.pDestructor = NULL;
```
krakjoe, line 330: This should not be necessary.

zend_abstract_interface/hook/hook.c
```diff
@@ -132,26 +127,37 @@ static int zai_hook_resolve_impl(zval *zv ZAI_TSRMLS_DC) {
...
+}
+
+#if PHP_VERSION_ID < 70000
+static int zai_hook_resolve_impl(zai_hook_t *hook ZAI_TSRMLS_DC) {
+#else
+static int zai_hook_resolve_impl(zval *zv ZAI_TSRMLS_DC) {
+    zai_hook_t *hook = Z_PTR_P(zv);
+#endif
+    if (zai_hook_resolve_hook(hook ZAI_TSRMLS_CC)) {
+        zai_hook_copy(hook ZAI_TSRMLS_CC);
```
krakjoe, line 157: Copy and remove looks wrong.

tea/include/extension.h
bwoebi: How would you prefer to write the test then?

https://github.com/DataDog/dd-trace-php/pull/1503/files#diff-5f9e4cde6523629a254e406c13d930a7bb1a29131ad64044e573a6212fe0495aR41

zend_abstract_interface/hook/hook.c
bwoebi: It's what has been happening before though, look at line 150 in the previous version:

https://github.com/DataDog/dd-trace-php/blob/cdc6833f3f34d52a35ed46b64282de1bd6c0c3ac/zend_abstract_interface/hook/hook.c#L150

zend_abstract_interface/hook/hook.c
```diff
@@ -460,7 +448,8 @@ bool zai_hook_install(
 
     zai_hook_t *hook;
 
-    if (!zai_hook_table_insert(table, &install, sizeof(zai_hook_t), (void **)&hook)) {
+    bool isUninitialized = zai_hook_resolved.pDestructor == NULL;
```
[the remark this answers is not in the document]
bwoebi: It needs to work when doing zai_hook_install at MINIT for example. It then must not do runtime-resolution / insert into the runtime table.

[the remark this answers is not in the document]
bwoebi: Likely `table == &zai_hook_request` could work as well...

zend_abstract_interface/hook/hook.c
krakjoe: I actually misread that, I read literally "remove" and "copy", not "resolve" and "copy" ... my mistake.

zend_abstract_interface/interceptor/php8/interceptor.c
```diff
@@ -0,0 +1,238 @@
...
+}
+
+void zai_interceptor_minit() {
+    prev_execute_internal = zend_execute_internal;
+    zend_execute_internal = prev_execute_internal ? zai_interceptor_execute_internal : zai_interceptor_execute_internal_no_prev;
+
+    zend_observer_fcall_register(zai_interceptor_observer_fcall_init);
+
+    // get hold of a generator object to access handlers
+    zend_objects_store objects_store = EG(objects_store);
```
krakjoe, line 209: super sneaky, I've never seen anything do this before and I'm not sure what side effects it might have ... it's a nice trick ...

zend_abstract_interface/interceptor/php7/interceptor.c
```diff
@@ -0,0 +1,580 @@
...
+    SET_UNUSED(zai_interceptor_generator_create_wrapper[1].result);
+    SET_UNUSED(zai_interceptor_generator_create_wrapper[1].op1);
+    SET_UNUSED(zai_interceptor_generator_create_wrapper[1].op2);
+    ZEND_VM_SET_OPCODE_HANDLER(&zai_interceptor_generator_create_wrapper[1]);
+#endif
+
+    INIT_NS_CLASS_ENTRY(zai_interceptor_bailout_ce, "Zend Abstract Interface", "BailoutHandler", NULL);
+    zai_interceptor_bailout_ce.type = ZEND_INTERNAL_CLASS;
+    zend_initialize_class_data(&zai_interceptor_bailout_ce, false);
+    zai_interceptor_bailout_ce.info.internal.module = module_entry;
```
krakjoe, line 557: This doesn't need to be an actual module, it can be a dummy, right ?

krakjoe: Another approach might be to add a dummy extension api to tea, we can export a module structure, I just don't want to export the one we use internally in tea ...

zend_abstract_interface/interceptor/php8/interceptor.c
bwoebi: well, in this specific context, where we just call the generator ctor, there should not really be side effects...
And yes, sneaky, but works. :-)

zend_abstract_interface/interceptor/php7/interceptor.c
bwoebi: yes, for the test it might be dummy memory, but in real execution at least something like xdebug will access these.

PR #1504, opened 2022-02-10
Use a new curl session for each writer iteration and clear it after

labbati:
### Description

Before this PR, we initialized a curl session at the first iteration (with loop time of 5 seconds) and then we used it until the process got killed. This means, potentially, that the same curl session could be alive for days.

This resulted in some cycles to be lost because the trace agent closed the connection (RST flag verified through tcpdump) at random intervals.

This PR changes this behavior in two ways:
1. it instantiates a new session at each writer iteration (5 seconds by default) and clears it after the transmission of the payloads that can be sent is completed.
2. increases the total usable number of buffers by 20% (from 10 to 12). This in order to account for possible delays due to slower connection operations during each cycle.

Note that this is only a first optimization, as a better approach, that can be explored as part of a refactoring that we have scheduled soon, could be to keep a session for more than one iteration, but attempting a new send in case of a network error to avoid loosing a payload.

### Readiness checklist
- ~[ ] Tests added for this feature/bug.~ I extensively tested it manually, but in order to add proper testing we will have to wait for the componentization of the BGS, that might be scheduled soon.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php5/coms.c
+++ b/ext/php5/coms.c
@@ -52,6 +52,10 @@ static bool _dd_is_memory_pressure_high(void) {
     }
 }
 
+/* Is called internally in the context of the PHP thread and actually attempts at storing the payload to send. If
+ * there is not enough memory left in the currently active stack, it does not attempt to generate a new stack, instead
+ * it returns a `ENOMEM` code that should be read by the invoker that can manually decide to ask for a larger stack.
+ */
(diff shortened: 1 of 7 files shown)


labbati:
Thanks @bwoebi

PR #1512, opened 2022-02-14
Do not mess with the span duration in curl integration

bwoebi:
### Description

Keep a distinction between the native span duration and the measured curl duration.

This also prevents type confusion in the APM itself.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


krakjoe (approved):

labbati (approved):
The change seems correct to me. Thanks @bwoebi 
Separately, we should discuss if this issue should also be addressed in the backend to make sure that user defined tags never impacts 'native' values (in this case duration).

PR #1517, opened 2022-02-16
Add Code Hotspots

morrisonlevi:
### Description

This adds support for Code Hotspots from the tracer side of things.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bwoebi:
Can we rename this to _tracing_context_? It's the context of the tracer we expose, not the profiler.

I think I'd also prefer having a single .c/h file pair instead of runtime + context as to not litter our exports all over the place.

morrisonlevi:
> Can we rename this to _tracing_context_? It's the context of the tracer we expose, not the profiler.

I want it to be clear that this is the tracing context we expose _for the profiler_. I think the name `ddtrace_tracing_context` should be reserved for the tracer's own use, probably to store all information required for a distributed trace (origin, trace id, sampling decision, and so on).

ext/php7/php7/engine_hooks.c
```diff
@@ -1080,8 +1088,16 @@ void ddtrace_message_handler(int message, void *arg) {
...
-                               "datadog_profiling_interrupt_function", DL_ERROR());
+            ddtrace_log_debugf("[Datadog Trace] Profiling v%s was detected, but locating symbol failed: \n",
+                               extension->version, DL_ERROR());
+        }
+
+        datadog_php_uuid (*runtime_id)(void) = DL_FETCH_SYMBOL(handle, "datadog_profiling_runtime_id");
+        if (EXPECTED(runtime_id)) {
+            ddtrace_profiling_runtime_id = runtime_id;
+        } else {
+            ddtrace_log_debugf("[Datadog Trace] Profiling v%s was detected, but locating symbol failed: \n",
```
morrisonlevi, line 1099: This `ddtrace_log_debugf` is happening in `MINIT` -- should I be doing this? @bwoebi @krakjoe

bwoebi: It's fine as long as it happens after the zai config minit, which is the case.

bwoebi (approved):
LGTM :-)

PR #1547, opened 2022-03-28
Remove regressions from randomized tests

labbati:
### Description

Since the very beginning, it never happened that a frozen regression test in the randomized tests caught an error
that the 'regular' randomly generated scenarios would not.

Moreover, maintaining frozen regressions is harder because the suite evolves and it forces us to consider each (or a batch of) regressions
as a special case.

For these reasons we decided to remove them.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
`rm *;`

labbati:
Thanks

PR #1556, opened 2022-04-01
Add gdbinit files for php 8.0 and 8.1

labbati:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tooling/gdbinit-8.0
+++ b/tooling/gdbinit-8.0
@@ -1,0 +1,645 @@
+define set_ts
+	set $tsrm_ls = $arg0
+end
+
+document set_ts
+	set the ts resource, it is impossible for gdb to
+	call ts_resource_ex while no process is running,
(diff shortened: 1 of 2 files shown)


bwoebi:
What do you need these for? As in, when isn't it enough to do `source /usr/local/src/php/.gdbinit` in gdb?

labbati:
For cases when you have not PHP's sources, but you mount dd-trace-php, e.g. our `dd-trace-php-solutions` repo, randomized tests and rel-env images.

If we want to remove them it is not a big deal (I can wget them, no problem) but we should either remove them for all versions in that folder or to have them for all versions.

bwoebi (approved):
Ah we already have existing gdbinit files (never seen those), then it's a no-brainer :-)

labbati:
Thanks

PR #1560, opened 2022-04-05
Update flex to v1.18.5 in composer.lock files for Symfony 4.0 and 5.0

bwoebi:
### Description

The newer flex version check against the existence of the manifest key instead of failing outright when it is missing.

This fixes our test_web_symfony runners.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/Frameworks/Symfony/Version_4_0/composer.lock
+++ b/tests/Frameworks/Symfony/Version_4_0/composer.lock
@@ -3019,16 +3019,16 @@
         },
         {
             "name": "symfony/flex",
-            "version": "v1.9.10",
+            "version": "v1.18.5",
             "source": {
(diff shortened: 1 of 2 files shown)


krakjoe (approved):

labbati (commented):
It is a mystery to me how this can fix the issue, since in theory we do `composer update` and not `composer install` (and we do not restore the `vendor` directory as far as I know). But if that work let's merge it.

labbati:
What I just wrote is not true for symfony 4.0 https://github.com/DataDog/dd-trace-php/blob/e343bdbfd528874b0c9116e97f5311f479e32b5d/Makefile#L993-L997 so go and ship it

PR #1561, opened 2022-04-05
Add a single test to xfail after upgrade of docker image to 8.0.18

bwoebi:


krakjoe (approved):

labbati (changes requested):
Please add a motivation for why we are disabling it to the README.md. If it is a common cause, there will already be a paragraph and just add the test to the list.

bwoebi:
@labbati Added

krakjoe (approved):

PR #1567, opened 2022-04-14
Update installer for AppSec GA

Anilm3:
### Description

Update installer for AppSec GA:
- Load extension whenever compatible, do not load otherwise.
- Enable or disable AppSec based on `--enable-appsec` modifier.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -268,27 +268,47 @@ function install($options)
                 // phpcs:enable Generic.Files.LineLength.TooLong
             }
 
-            // Enabling appsec
-            if (is_truthy($options[OPT_ENABLE_APPSEC])) {
-                if ($shouldInstallAppsec) {
-                    // Appsec crashes with missing symbols if tracing is not loaded
-                    execute_or_exit(
-                        'Impossible to update the INI settings file.',
-                        "sed -i 's@ \?; \?extension \?= \?ddtrace.so@extension = ddtrace.so@g' "
-                            . escapeshellarg($iniFilePath)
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
Looks correct to me

Anilm3:
Looks like a test fails because if `extension=ddtrace.so` was commented we don't uncomment it, however this causes a conflict with AppSec as we need the extension to be enabled in order to load symbols from it (this doesn't cause any crashes as far as I'm aware, just some warnings). This also prevents a user from reinstalling after uninstalling (based on the tests), which is not very intuitive.

I suggest we add either `--enable-tracing` or `--disable-tracing` to resolve this ambiguity:

- Enabling tracing will always be done if `--enable-appsec` is present, so the `--enable-tracing` flag makes it easier to avoid conflicting options and it's also more consistent. 
- If the user doesn't pass a specific enabled flag, we load the extensions but set the relevant enabled flag to `Off`. This doesn't apply to BETA extensions, which aren't loaded by default.

I suppose there could also be a small subset of customers who are using the profiler and have issues when loading the tracer, which might require a different approach.

cc: @labbati @morrisonlevi

labbati:
> I suggest we add either --enable-tracing or --disable-tracing to resolve this ambiguity
I am not sure this is feasible, especially as we move in the future more and more features to the tracer to be used by other products that at that point depend on it. What should be done to make that test green with the current api? It is not clear what the problem is here to me.

PR #1584, opened 2022-04-29
Fix installer setting check for extensions

Anilm3:
### Description

Currently the installer checks if a setting was already available before adding it to the `.ini` file, however since the extension setting is not unique, on existing installations performed with a different installation method, `extension = ddappsec.so` is not added.

This fix just makes sure we verify that the correct extensions and zend extensions are loaded by checking not only the setting but also the value.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


labbati (approved):
LGTM!

PR #1585, opened 2022-04-29
Fix inconsistent module load and unload sequence

Anilm3:
### Description

Currently the tracer calls `dlload` on itself on `MINIT` to ensure the reference count is consistent with the number of times `dlclose` should, in theory, be called against it, due to it being both a module and a zend extension. However during `MSHUTDOWN` the tracer also sets the module handle to `NULL`, thus preventing one of the two calls to `dlclose` on its handle and leaking resources.

This change consolidates the current behaviour during `MINIT` by setting the module handle to `NULL` and removing any extra calls to `dlload`.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php7/ddtrace.c
+++ b/ext/php7/ddtrace.c
@@ -450,15 +450,13 @@ static PHP_MINIT_FUNCTION(ddtrace) {
      * See http://www.phpinternalsbook.com/php7/extensions_design/zend_extensions.html#hybrid-extensions
      * {{{ */
     zend_register_extension(&_dd_zend_extension_entry, ddtrace_module_entry.handle);
-#ifdef COMPILE_DL_DDTRACE
     zend_module_entry *mod_ptr = zend_hash_str_find_ptr(&module_registry,
         PHP_DDTRACE_EXTNAME, sizeof(PHP_DDTRACE_EXTNAME) -1);
     if (mod_ptr == NULL) {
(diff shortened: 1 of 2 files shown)


ext/php7/ddtrace.c
```diff
@@ -450,12 +450,13 @@ static PHP_MINIT_FUNCTION(ddtrace) {
...
-#ifdef COMPILE_DL_DDTRACE
-    Dl_info infos;
-    // The symbol used needs to be public on Alpine.
-    dladdr(get_module, &infos);
-    dlopen(infos.dli_fname, RTLD_LAZY);
-#endif
+    zend_module_entry *mod_ptr =
+        zend_hash_str_find_ptr(&module_registry, PHP_DDTRACE_EXTNAME, sizeof(PHP_DDTRACE_EXTNAME) - 1);
+    if (mod_ptr == NULL) {
+        // This shouldn't happen, possibly a bug if it does.
```
bwoebi, line 456: This should invoke zend_error() then (core warning?), with a message similar to "Did not find ddtrace extension in registered modules. Please open a bug report."

bwoebi (dismissed):
Thanks, looks good to me now :-)

bwoebi (approved):

morrisonlevi (approved):

PR #1586, opened 2022-05-02
Eliminate the separate span id stack, fix distributed tracing trace_id

bwoebi:
### Description

The distributed_tracing/distributed_trace_overwrite*.phpt tests were having bogus EXPECT result.
The root cause was that pushing a new span id (for the parent), also caused DDTRACE_G(trace_id) to be overwritten, if there was no parent span (i.e. as a side effect of ddtrace_push_userland_span_id().

As there anyway was no reason to keep the span id stack any longer (the span stack itself has been completely internal for more than half a year), except for compatibility reasons, and it has become a pitfall, we now completely eliminate it and replace it by direct access to the span stack.

This is especially important moving forward with asynchronous tracing, where multiple traces and span stacks will be able to coexist, and managing a separate span stack is just extraneous overhead.

Thus, as of now, the sources of truth for trace and parent ids upon creating spans are:
 - if there is an active span, the current active span
 - if there is no active span, DDTRACE_G(trace_id) and DDTRACE_G(distributed_parent_trace_id)

Note: the php 7.4 failure is unrelated (related to todays tag of the latest version of symfony/flex).

### Readiness checklist
- [x] Tests added for this feature/bug. (well, fixed.)

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi (approved):
Looks good to me. Hooray for dropping more code than was written, and goodbye to one more piece of legacy.

PR #1601, opened 2022-06-01
Add support for unix domain sockets

bwoebi:
Including:
- default unix socket for dogstatsd
- default unix socket for background sender

Manually tested dogstatsd by creating a dummy socket `socat UNIX-LISTEN:/var/run/datadog/dsd.socket -` - not sure how to test that automatically, or against the real stats receiver.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php5/dogstatsd_client.c
+++ b/ext/php5/dogstatsd_client.c
@@ -63,7 +63,8 @@ void ddtrace_dogstatsd_client_rinit(TSRMLS_D) {
 
         client = dogstatsd_client_ctor(addrs, DOGSTATSD_CLIENT_RECOMMENDED_MAX_MESSAGE_SIZE, METRICS_CONST_TAGS);
         if (dogstatsd_client_is_default_client(client)) {
-            ddtrace_log_debugf("Dogstatsd client failed opening socket to %s%s%s", host, port ? ":" : "", port ? port : "");
+            ddtrace_log_debugf("Dogstatsd client failed opening socket to %s%s%s", host, port ? ":" : "",
+                               port ? port : "");
             break;
         }
 
--- a/ext/php5/dogstatsd_client.c
+++ b/ext/php5/dogstatsd_client.c
@@ -35,23 +35,58 @@ void ddtrace_dogstatsd_client_rinit(TSRMLS_D) {
 
     while (health_metrics_enabled) {
         struct addrinfo *addrs;
-        const char *host = get_DD_AGENT_HOST().ptr;
-        const char *port = get_DD_DOGSTATSD_PORT().ptr;
-        if (!*host) {
-            if (access(DEFAULT_UDS_PATH, F_OK) == SUCCESS) {
-                addrs = dd_alloc_unix_addr(DEFAULT_UDS_PATH, sizeof(DEFAULT_UDS_PATH));
-                host = "unix://" DEFAULT_UDS_PATH;
-                port = NULL;
-            } else {
-                host = "localhost";
-            }
-        }
+        const char *url = get_DD_DOGSTATSD_URL().ptr;
+        const char *host, *port;
+        if (*url) {
+            if (strlen(url) > 7 && strncmp("unix://", url, 7) == 0) {
+                addrs = dd_alloc_unix_addr(url + 7, strlen(url) - 7);
+            } else if (strlen(url) > 6 && strncmp("udp://", url, 6) == 0) {
+                char *colon = strchr(url + 6, ':');
+                if (!colon) {
+                    ddtrace_log_debugf(
+                        "Dogstatsd client encountered an invalid udp:// DD_DOGSTATSD_URL: %s, missing a colon followed by a port",
+                        url);
+                    break;
+                }
 
-        if (port) {
+                char *hostname = estrndup(url + 6, colon - url - 6);
+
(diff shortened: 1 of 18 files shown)


tests/ext/background-sender/agent_headers_unix_domain_socket.phpt
```diff
@@ -0,0 +1,58 @@
...
+DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS=1
+DD_TRACE_AGENT_FLUSH_INTERVAL=333
+DD_TRACE_GENERATE_ROOT_SPAN=0
+--FILE--
+<?php
+include __DIR__ . '/../includes/request_replayer.inc';
+
+$socketPath = getenv("DD_TRACE_AGENT_URL");
+@unlink($socketPath);
+$p = popen(PHP_BINARY . ' -r \'$s=stream_socket_server("' . $socketPath . '");print 1;$c=stream_socket_accept($s);$t=stream_socket_client("request-replayer:80");$a=$r=[$c,$t];foreach($r as $f)stream_set_blocking($f, false);while(stream_select($r,$w,$e,null)){$d=fread($f=reset($r),4096);if($d=="")exit;fwrite($f==$t?$c:$t,$d);$r=$a;}\'', 'r');
```
morrisonlevi, line 17: Must the `-r` value contain no newlines?

ext/php8/coms.c
```diff
@@ -720,15 +721,30 @@ char *ddtrace_agent_url(void) {
...
 }
 
 void ddtrace_curl_set_hostname(CURL *curl) {
     char *url = ddtrace_agent_url();
     if (url && url[0]) {
-        size_t agent_url_len = strlen(url) + sizeof(TRACE_PATH_STR);
+        char *http_url = url;
+        if (strlen(url) > 7 && strncmp(url, "unix://", 7) == 0) {
+            curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, url + 7);
+            http_url = "http://localhost";
```
morrisonlevi, line 743: For curl, when using UDS, does it matter what the host is? Could `localhost` be replaced with theoretically anything?

tests/ext/background-sender/agent_headers_unix_domain_socket.phpt
bwoebi: Can you please suggest a proper formatting? It's just a line to forward proxy data from one socket to the other.

morrisonlevi: I would probably use a heredoc and format it basically as normal PHP, I think? I get that it is "unimportant" but I'd still like to read and understand it :)

morrisonlevi: Also, it appears to be used more than once, so maybe make a function for it?

ext/php5/dogstatsd_client.c
```diff
@@ -9,38 +9,47 @@
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
 #define METRICS_CONST_TAGS "lang:php,lang_version:" PHP_VERSION ",tracer_version:" PHP_DDTRACE_VERSION
+#define DEFAULT_UDS_PATH "/var/run/datadog/dsd.socket"
```
r1viollet, line 12: the usual env var available on the machine (connecting to a random pod) is the following 
```
STATSD_URL=unix:///var/run/datadog-agent/statsd.sock
```

r1viollet:
Testing with a real Datadog agent could be a good way to test this.

ext/php5/dogstatsd_client.c
bwoebi: I haven't added configurability of the dogstatsd unix socket here. (I.e. we don't support STATSD_URL here at all.)
It just defaults to the unix socket if there's one.

Unlike with the tracer (DD_TRACE_AGENT_URL), I wasn't aware of a "full" url env var for dogstatsd. So `STATSD_URL` is the proper env var for that?

r1viollet: I'm honestly confused by the RFC. This was the only statsd socket I was aware of on a POD. I'll ask around and see if I can figure this one out.

r1viollet: The internal statsd configuration is different from the one defined in public documentations.

bwoebi: It's now configurable via DD_AGENT_HOST, so can also be used on other non-standard paths.

src/dogstatsd/client.c
```diff
@@ -60,14 +61,25 @@ dogstatsd_client dogstatsd_client_ctor(struct addrinfo *addrs, char *buffer,
     }
   }
 
+  if (addr->ai_family == PF_UNIX) {
+    if (!connect(client.socket, addr->ai_addr, addr->ai_addrlen)) {
```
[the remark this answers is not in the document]
bwoebi: That case does not matter, it then does connect(-1) which just fails again here.

morrisonlevi (approved):
The code looks good to me, and I was able to send traces using the real agent. I committed the necessary changes to `docker-compose.yml` so this will work out-of-the-box.

PR #1606, opened 2022-06-08
Add profiling sources

morrisonlevi:
### Description

This PR does not:

- Alter the tracer in any way, including its build requirements.
- Add new tests to CI, and will only fix/adjust lints jobs so that existing tracer jobs will not be affected by the `profiling` subdirectory.
- Build out of the box with `cargo build`. You have to check out Datadog/libdatadog's main branch to `profiling/libdatadog`. I didn't want to add a git submodule without discussion and the crates for libdatadog are not yet published, so this is a compromise.

This PR does:

- Add sources for a PHP profiler written almost entirely in Rust. It is essentially a port of the existing C profiler found at [Datadog/dd-prof-php](https://github.com/DataDog/dd-prof-php).
- Have some tests, but not many. As mentioned above, they are not run in CI.

Fixing these things shortcomings will likely come in further PRs, as this one is already at +2,481 additions and I expect it will take some time to be able to get it reviewed and merged, especially as it needs reviewed by both Rust and PHP experts who are busy working on their own things.

There are a variety of `todo` tasks that I do not expect to be handled in this PR, like unifying `components/sapi` with `profiling/src/sapi.rs`. However, there is one about `zend_bool` that I still need to audit. Given that I expect this PR to take a while to get fully reviewed and that it hasn't caused observable issues yet, I decided to open the PR and do this audit concurrently (and hopefully take no action except remove the todo).

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        });
+    }
+
+    ZendResult::Success
+}
+
+unsafe fn stop_profiling() {
+    if let Some(profiler) = PROFILER.replace(None) {
+        profiler.stop();
+    }
```
bruceg, line 516: Try `take`:
```suggestion
    if let Some(profiler) = PROFILER.take() {
        profiler.stop();
    }
```

profiling/src/profiling.rs
```diff
@@ -0,0 +1,806 @@
...
+            instant: Instant::now(),
+            systemtime: SystemTime::now(),
+        }
+    }
+}
+
+#[repr(C, align(16))]
+#[derive(Debug, Copy, Clone)]
+pub struct Uuid {
+    pub data: [u8; 16usize],
```
bruceg, line 54: The `usize` is superfluous here.
```suggestion
    pub data: [u8; 16],
```

profiling/src/profiling.rs
```diff
@@ -0,0 +1,806 @@
...
+        Self::from_bytes(*uuid.as_bytes())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum LabelValue {
+    Str(Cow<'static, str>),
+
+    #[allow(dead_code)]
+    Num(i64, Cow<'static, str>),
```
bruceg, line 94: Is this variant left in for compatibility reasons? I'm gathering from the `dead_code` that it is never created and so unused.

profiling/src/profiling.rs
```diff
@@ -0,0 +1,806 @@
...
+        0
+    } else {
+        module.len() + 1
+    };
+    // +2 for "::"
+    let class_len = if class_name.is_empty() {
+        0
+    } else {
+        class_name.len() + 1
+    };
```
bruceg, line 239: Should this be `+ 2`?

profiling/src/profiling.rs
```diff
@@ -0,0 +1,806 @@
...
+            let zend_function = &*func;
+            let function_result = extract_function_name(zend_function);
+
+            // ensure that str in Some(str) is non-empty, simplifies code below
+            let function = if let Ok(str) = function_result {
+                if !str.is_empty() {
+                    Some(str)
+                } else {
+                    None
+                }
```
bruceg, line 330: This can be simplified, though it is slightly awkward:
```suggestion
                (!str.is_empty()).then(|| str)
```

bruceg (commented):
I don't have enough knowledge of the broader context to comment on the overall approach, so this review just reflects on the smaller details.

profiling/src/bindings.rs
```diff
@@ -0,0 +1,291 @@
+// Of course, the C API doesn't follow Rust conventions, so suppress lints.
+
+#![allow(clippy::all)]
+#![allow(warnings)]
```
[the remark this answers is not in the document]
blt: +1, this will make the project hard to work with over time if left.

profiling/src/bindings.rs
```diff
@@ -0,0 +1,291 @@
...
+impl From<c_int> for zend_result {
+    fn from(value: c_int) -> Self {
+        match value {
+            0 => Self::Success,
+            _ => Self::Failure,
+        }
+    }
+}
+
+// In general, modify definitions which return int to return zend_result iff
```
blt, line 42:
```suggestion
// In general, modify definitions which return int to return zend_result if
```

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+use std::sync::Once;
+use std::time::Instant;
+
+static NAME: &[u8] = b"datadog-profiling\0";
+static VERSION: &[u8] = b"0.7.0\0";
+static mut SAPI: RefCell<Sapi> = RefCell::new(Sapi::Unknown);
+
+#[no_mangle]
+pub extern "C" fn get_module() -> &'static mut zend::ModuleEntry {
+    /* In PHP modules written in C, this just returns the address of a global,
```
blt, line 30: Nice comment work here.

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        let sapi_name = CStr::from_ptr(sapi_module.name);
+        *sapi = Sapi::from_name(sapi_name.to_string_lossy().as_ref())
+    }
+
+    /* Use a hybrid extension hack to load as a module but have the
+     * zend_extension hooks available:
+     * https://www.phpinternalsbook.com/php7/extensions_design/zend_extensions.html#hybrid-extensions
+     */
+    // Transfer the module handle to the zend_extension. Extensions have
+    // longer lifetimes than modules in the engine.
```
blt, line 118: It's not a problem per se, but you have mixed commenting styles in this PR, the `/* */` and `//`. Curious if there's a logic to this? As a reader it's not clear to me that there is.

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+    let handle = {
+        // The engine copies the module entry we provide it, so we have to
+        // lookup the module entry in the registry and modify it there
+        // instead of just modifying the result of get_module().
+        let str = NAME.as_ptr();
+        let len = NAME.len() - 1; // ignore trailing null byte
+        let ptr = zend::datadog_get_module_entry(str, len);
+        if ptr.is_null() {
+            // todo: error message
```
blt, line 127: Intentionally left TODO?

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+    trace!("PRSHUTDOWN");
+
+    ZendResult::Success
+}
+
+/// The version of PHP at runtime, not the version compiled against. Sent as
+/// a profile tag.
+static mut PHP_VERSION: Option<String> = None;
+
+static mut PROFILER: RefCell<Option<Profiler>> = RefCell::new(None);
```
blt, line 202: Intentionally missing comment?

profiling/src/logging.rs
```diff
@@ -0,0 +1,23 @@
...
+
+pub fn log_init(level_filter: LevelFilter) {
+    let target = unsafe {
+        let fd = libc::dup(libc::STDERR_FILENO);
+        if fd == -1 {
+            // logging isn't going to work, sorry!
+            return;
+        }
+        Box::new(File::from_raw_fd(fd))
+    };
```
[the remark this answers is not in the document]
morrisonlevi: I cleaned this up a bit, though I went with different code because `File::from_raw_fd` is also unsafe. Thanks for drawing attention here!

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+        Some(interrupt_function_wrapper)
+    } else {
+        PREV_INTERRUPT_FUNCTION.write(None);
+        Some(interrupt_function)
+    };
+
+    let prev_execute_internal = match zend::zend_execute_internal {
+        None => zend::execute_internal,
+        Some(prev) => prev,
+    };
```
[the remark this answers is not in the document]
morrisonlevi: I cleaned up this section. Thanks for the suggestions!

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+    // profiling is enabled or not.
+    static ONCE: Once = Once::new();
+    ONCE.call_once(|| {
+        #[cfg(not(debug_assertions))]
+        logging::log_init(unsafe { std::mem::transmute(globals.profiling_log_level) });
+        #[cfg(debug_assertions)]
+        log::set_max_level(globals.profiling_log_level);
+
+        unsafe {
+            // reminder: this cannot be done in minit because of Apache forking model
```
blt, line 359: I'm not sure what this is meant to be informing me of? It can't be done but is done?

profiling/src/profiling.rs
morrisonlevi: Since we're not using `try_reserve`, I simplified this whole section so the math errors just can't happen. Thanks!

profiling/src/profiling.rs
morrisonlevi: This variant matches the capabilities of the underlying `pprof` domain -- I just don't happen to use that field at the moment. It seemed better to keep this variant and have the `allow(dead_code)` than to remove it but leave a note about why there's only one variant in the enum. What do you think?

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+                    }
+                } else if globals.profiling_experimental_cpu_time_enabled {
+                    info!("CPU Time profiling enabled.");
+                }
+            }
+        }
+    });
+
+    if globals.profiling_enabled {
+        unsafe {
```
blt, line 395: This is a _very big_ unsafe block. Can it be made smaller?

profiling/src/logging.rs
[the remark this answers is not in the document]
blt: +1

profiling/src/profiling.rs
blt: +1

profiling/src/profiling.rs
```diff
@@ -0,0 +1,806 @@
...
+            }
+        };
+    }
+
+    pub fn send_sample(&self, message: SampleMessage) -> Result<(), TrySendError<SampleMessage>> {
+        self.message_sender.try_send(message)
+    }
+
+    pub fn stop(self) {
+        // todo: what should be done when a thread panics?
```
blt, line 623: Intentionally left TODO?

profiling/src/lib.rs
```diff
@@ -0,0 +1,719 @@
...
+
+    static DEPS: [zend::ModuleDep; 2] = [
+        zend::ModuleDep::optional(b"ddtrace\0"),
+        zend::ModuleDep::end(),
+    ];
+
+    static ONCE: Once = Once::new();
+
+    unsafe {
+        ONCE.call_once(|| {
```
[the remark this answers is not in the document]
morrisonlevi: It looks like I could replace all of these things with different things from the `once_cell` crate. I'll try it out soon.

profiling/src/lib.rs
morrisonlevi: In the PHP extension and module ecosystem, the "MINIT" phase is when a `zend_module_entry` is initialized. You can read more about this here if you care to: https://www.phpinternalsbook.com/php7/extensions_design/php_lifecycle.html.

The comment is there to remind me that because of the way Apache forks processes, I cannot spawn a thread (which the profiler does) during MINIT. There are more comments about this on the docs for the `minit` function.

profiling/src/lib.rs
morrisonlevi: Yes, I think it can be shrunk quite a bit. In fact, if I switch to using the `once_cell` crate I think all of the unsafe usage of this particular block goes away.

profiling/src/profiling.rs
morrisonlevi: Yes, but also I really don't know what ought to be done. In general, I don't think anything can be done except to propagate the panic.

profiling/src/lib.rs
```diff
@@ -0,0 +1,713 @@
...
+
+static NAME: &[u8] = b"datadog-profiling\0";
+static VERSION: &[u8] = b"0.7.0\0";
+static mut SAPI: RefCell<Sapi> = RefCell::new(Sapi::Unknown);
+
+#[no_mangle]
+pub extern "C" fn get_module() -> &'static mut zend::ModuleEntry {
+    /* In PHP modules written in C, this just returns the address of a global,
+     * mutable variable. In Rust, you cannot initialize such a complicated
+     * global variable because of initialization order issues that have been
```
mokomull, line 32: would the `lazy_static` crate help you avoid rolling your own `unsafe { }` later in this function?

[looks like once_cell was also mentioned below, but one or the other seems useful]

profiling/src/lib.rs
```diff
@@ -0,0 +1,713 @@
...
+static mut RUNTIME_ID: profiling::Uuid = profiling::Uuid::new();
+
+/// The version of the ZendEngine at runtime, not the version compiled against.
+/// It's currently unused, but I hope to send it as a metric or something soon
+/// so I'm keeping it here so I don't have to look up how to get it again.
+static mut ZEND_VERSION: Option<String> = None;
+
+/// # Safety
+/// Only call during an active request from the PHP thread!
+unsafe fn getenv(name: &CStr) -> Option<String> {
```
mokomull, line 217: Why does this take a `&CStr`?  You can simplify both here and at the callsites by taking an `&str`(guaranteeing UTF8-ness at compile-time) and before falling back to libc, allocating a temporary CString to get the null terminator (which is basically free since all of the environment variable names you use are a dozen bytes or so).

profiling/src/lib.rs
```diff
@@ -0,0 +1,713 @@
...
+    match string.as_str() {
+        "1" | "on" | "yes" | "true" => Some(true),
+        "0" | "no" | "off" | "false" => Some(false),
+        _ => None,
+    }
+}
+
+/// Intern the string for the duration of a request.
+/// # Safety
+/// Do not use the string for more than the request lifetime!
```
mokomull, line 252: Would it be possible to enforce this with some lifetime on a (perhaps even empty) Request struct?

(It may very well not be, since request init and teardown are two separate FFI calls and rustc can't really know much about what's happening behind the scenes)

profiling/src/lib.rs
```diff
@@ -0,0 +1,713 @@
...
+/// Used internally to gather time samples when the configured period has
+/// elapsed. Also used by the tracer to handle pending profiler interrupts
+/// before calling a tracing closure from an internal function hook; if this
+/// isn't done then the closure is erroneously at the top of the stack.
+///
+/// # Safety
+/// The zend_execute_data pointer should come from the engine to ensure it and
+/// its sub-objects are valid.
+#[export_name = "datadog_profiling_interrupt_function"]
+pub unsafe extern "C" fn interrupt_function(execute_data: *mut zend::zend_execute_data) {
```
mokomull, line 650: Why not name the Rust item `datadog_profiling_interrupt_function` and save a future reader grief when wondering where the C symbol came from :)

profiling/src/logging.rs
```diff
@@ -0,0 +1,20 @@
...
+pub fn log_init(level_filter: LevelFilter) {
+    // Safety: this is safe, it's just "unsafe" because it's a call into C.
+    let fd = unsafe { libc::dup(libc::STDERR_FILENO) };
+    if fd != -1 {
+        // Safety: the fd is a valid and open file descriptor, and the File has sole ownership.
+        let target = Box::new(unsafe { File::from_raw_fd(fd) });
+        env_logger::builder()
+            .filter_level(LevelFilter::Off)
+            .filter_module("datadog_php_profiling", level_filter)
+            .target(Target::Pipe(target))
```
mokomull, line 16: Is there a reason to do our own fd-duplicating logic, or can we use `Target::Stderr`?

profiling/src/php_ffi.c
```diff
@@ -0,0 +1,210 @@
...
+ */
+static char char_tolower_ascii(char c) { return (char)(c >= 'A' && c <= 'Z' ? (c - ('A' - 'a')) : c); }
+
+static void str_copy_tolower_ascii(size_t len, const char src[static len], char dest[static len]) {
+    for (size_t i = 0; i != len; ++i) {
+        *(dest++) = char_tolower_ascii(src[i]);
+    }
+}
+
+static zend_result datadog_php_profiling_log_level_parse(datadog_php_str str, uintptr_t *log_level) {
```
mokomull, line 60: I'm not seeing anywhere this is used other than the test for it

profiling/src/php_ffi.h
```diff
@@ -0,0 +1,133 @@
...
+ * todo: unify with ../component/log
+ */
+typedef enum datadog_php_profiling_log_level {
+    DATADOG_PHP_PROFILING_LOG_LEVEL_OFF = 0,
+    DATADOG_PHP_PROFILING_LOG_LEVEL_ERROR,
+    DATADOG_PHP_PROFILING_LOG_LEVEL_WARN,
+    DATADOG_PHP_PROFILING_LOG_LEVEL_INFO,
+    DATADOG_PHP_PROFILING_LOG_LEVEL_DEBUG,
+    DATADOG_PHP_PROFILING_LOG_LEVEL_TRACE,
+} datadog_php_profiling_log_level;
```
mokomull, line 55: likewise, I don't see this being used at all; I just see `log::LevelFilter` being transmute()d into a uintptr_t in read_env().

profiling/src/php_ffi.h
```diff
@@ -0,0 +1,133 @@
...
+
+    // The strings will be interned but potentially only for the request, so be
+    // careful to not use them outside of a request (such as from other
+    // threads).
+    datadog_php_str env;
+    datadog_php_str service;
+    datadog_php_str version;
+    datadog_php_str trace_agent_host;
+    datadog_php_str trace_agent_port;
+    datadog_php_str trace_agent_url;
```
mokomull, line 83: Can these change from request-to-request (particularly wondering about sapi_getenv() ... the libc getenv() shouldn't really change during execution).

If not, then perhaps we should hold a void* to a `Box::into_raw()` of a Rust structure so we don't have to bounce these strings through FFI so much just for them to end back up in a `Vec<Tag>`

profiling/src/profiling.rs
```diff
@@ -0,0 +1,751 @@
...
+
+impl From<Uuid> for uuid::Uuid {
+    fn from(uuid: Uuid) -> Self {
+        Self::from_bytes(*uuid.as_bytes())
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum LabelValue {
+    Str(Cow<'static, str>),
```
mokomull, line 90: This seems to only ever be used for `format!()` so perhaps just take a String for simplicity?

profiling/src/profiling.rs
mokomull: FWIW this is a Hard™ problem, and we build Mork with `panic=abort` for that reason.

profiling/src/sapi.rs
```diff
@@ -0,0 +1,92 @@
...
+                ("cli", Sapi::Cli),
+                ("cli-server", Sapi::CliServer),
+                ("embed", Sapi::Embed),
+                ("fpm-fcgi", Sapi::FpmFcgi),
+                ("litespeed", Sapi::Litespeed),
+                ("phpdbg", Sapi::PhpDbg),
+                ("tea", Sapi::Tea),
+            ]);
+            unsafe { SAPIS.write(sapis) };
+        });
```
mokomull, line 38: This seems like a job for the `maplit` crate.

mokomull (commented):
I didn't really get a chance to look too hard at `profiling.rs`, but a few comments nonetheless.  Mostly style and "a crate already exists to do that" flavoured suggestions.

profiling/src/php_ffi.h
morrisonlevi: They can change from request to request, yes. For instance, in Apache you can configure env vars to have different values for different directories, think `DD_SERVICE=myservice` but in `/admin` it's `DD_SERVICE=myservice-admin` or something.

The code is partly set up the way it is because PHP's request globals have support for being tied to `.ini` configuration. I had previously tried to add `.ini` support, but hit an issue with certain configs like `datadog.service` being shared by the tracer and profiler, and the engine does not like this (it's not a C vs Rust issue, but a PHP extension issue). So, I pulled out all of the _directly_ dead code but some of the influence is still there. I think I need to support `.ini` configuration so I'm not sure it's worth making the Rust side of things cleaner at the moment since I'm highly likely to undo that work for `.ini`. Dunno, I'll think about it.

profiling/src/profiling.rs
morrisonlevi: `panic=abort` seems nice, but to _really_ get it right you need the stdlib to be rebuilt with it. I couldn't get that to work on Alpine, at the very least, though I haven't tried the latest Alpine release.

profiling/src/logging.rs
morrisonlevi: Yes, there is! I will document this. For almost all released versions the PHP CLI will close the stderr stream between `rshutdown` and `prshutdown`, but there are still other things that write to stderr after this (not just in this extension, but in other extensions too). Certain Rust libs are definitely _not_ expecting that to happen, and I was getting panics. That's actually why I am using `env_logger`: because it supports using a target based on a file descriptor.
I fixed this in PHP for the latest patch releases of PHP 8, but I still need to support PHP 7, and it will take time for these fixes to roll out everywhere.

profiling/src/php_ffi.c
morrisonlevi: Yes, I think so. It used to be used in the `.ini` parsing bits.

profiling/src/php_ffi.h
morrisonlevi: Yes, this was also used for the `.ini` handling, but it somehow lost its `repr(C)` along the way.

profiling/src/sapi.rs
morrisonlevi: I think before Rust 1.56 that `maplit` would be nice, but since then we have `from` and `from_iter` for `HashMap`. Using `maplit` would be changing out the `(_, _),` structure for `_ => _,` via macro, and I'm not sure how that's better? If it could do it in const contexts that would be awesome, but it can't. Am I missing something?

profiling/src/lib.rs
morrisonlevi: In this particular case, even if I switch to a lazy_static or a Once cell (which I've done locally), this is _still_ handing out a mutable reference to static variable, so it's just fundamentally unsafe. However, many other usages of `unsafe` have been reduced by using these crates.

profiling/build.rs
```diff
@@ -0,0 +1,143 @@
...
+fn generate_bindings(php_config_includes: &str) {
+    println!("cargo:rerun-if-changed=src/php_ffi.h");
+    let ignored_macros = IgnoreMacros(
+        [
+            "FP_INFINITE".into(),
+            "FP_NAN".into(),
+            "FP_NORMAL".into(),
+            "FP_SUBNORMAL".into(),
+            "FP_ZERO".into(),
+            "IPPORT_RESERVED".into(),
```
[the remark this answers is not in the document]
morrisonlevi: Apparently the `blocklist_item` is sufficient.

profiling/src/lib.rs
morrisonlevi: Oops, not intentional. Fixed.

profiling/src/lib.rs
morrisonlevi: I decided to leave an error log message. Thanks!.

profiling/src/lib.rs
morrisonlevi: `/*` for multiline but apparently not consistent ^_^ it should be more consistent now.

profiling/src/bindings.rs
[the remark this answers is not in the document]
morrisonlevi: Moved into a sub-module and the lints only apply there.

profiling/src/lib.rs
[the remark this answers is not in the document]
morrisonlevi: Most, but not all of these things cleaned up with `once_cell`. Thanks!

profiling/src/bindings.rs
```diff
@@ -0,0 +1,291 @@
...
+    }
+}
+
+#[repr(C)]
+pub struct EfreePtr<T> {
+    ptr: *mut T,
+}
+
+impl EfreePtr<c_char> {
+    pub fn into_string(self) -> String {
```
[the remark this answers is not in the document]
morrisonlevi: Exclusively this is used to fetch env vars, so they are just turned into strings immediately, but I define the dtor and such so I don't forget to cleanup. I'll pass on the YAGNI principle.

morrisonlevi:
Thanks @mokomull, @blt , and @bruceg. I've addressed most of the feedback, though I'm still thinking about some of them like changing the globals struct.

morrisonlevi:
Added a CODEOWNERS file as requested, but it's currently invalid as `DataDog/profiling-php` doesn't yet have write access.

bwoebi (dismissed):
I assume that before a GA, we will still see proper ini handling.

In general the code reads pretty clean. Looks good to me for initial inclusion.

bwoebi (approved):

PR #1609, opened 2022-06-08
Transition to non-legacy circleci images

morrisonlevi:
### Description

Transition to non-legacy circleci images. I'm not sure this is all of them, but it's at least a good chunk. A job will have a notice like this if it's a deprecated image:

You’re using a deprecated Docker convenience image. Upgrade to a next-gen Docker convenience image.

This also means a newer clang-format is used, so had to fixup a few of those changes.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -270,7 +270,7 @@ jobs:
   "Lint files":
     working_directory: ~/datadog
     docker:
-      - image: circleci/php:7.2-cli-node-browsers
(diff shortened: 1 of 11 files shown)


bwoebi (approved):
Ack.

Also these failures explain what was going on on my machine, I thought I had too old of a clang format fixer locally, but apparently it's too new :-D

PR #1615, opened 2022-06-15
Collect query string by default and obfuscate

Anilm3:
### Description

Collect the query string by default as part of `http.url` and add an obfuscation step using a predefined regex. Some considerations:
- The regex is configurable with the variable `DD_OBFUSCATION_QUERY_STRING_REGEXP`.
- The obfuscation is only done when `DD_TRACE_HTTP_URL_QUERY_PARAM_ALLOWED` is set to `*` (which is now the default value as well).

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


tests/ext/root_span_url_with_query_params.phpt
```diff
@@ -0,0 +1,28 @@
+--TEST--
```
Anilm3, line 1: This test was just renamed, but another test was created with the same name this one used to have so it appears as new instead...

ext/php7/serializer.c
```diff
@@ -555,7 +556,7 @@ void ddtrace_set_root_span_properties(ddtrace_span_t *span) {
                 if (query_str) {
                     query_string =
                         zai_filter_query_string((zai_string_view){.len = strlen(query_str), .ptr = query_str},
-                                                get_DD_TRACE_RESOURCE_URI_QUERY_PARAM_ALLOWED());
```
bwoebi, line 558: This does not look right to me. With `DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED=true` the query string is sent unmodified by default to datadog, given the new `*` default. I think at this place we need the regex filter as well.

Anilm3: Well, the default for `DD_TRACE_RESOURCE_URI_QUERY_PARAM_ALLOWED` is still an empty array, but I guess it also makes sense to add the regex here for the `*` case, so I'll do that.

ext/php7/serializer.c
```diff
@@ -10,6 +10,7 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <ext/pcre/php_pcre.h>
```
bwoebi, line 13:
```suggestion
```
Seems like a leftover?

ext/php7/serializer.c
bwoebi: Right, I was confused with the different names, but yes by default the regex should apply too.

bwoebi (approved):
Looks good now :-)

PR #1619, opened 2022-06-20
Disable clang format check in CI

krakjoe:
### Description

This disables the clang format check for C source files while leaving the remaining tasks for the lint job unchanged.

Our very basic usage of clang format leads to frustration when opening PR's of any complexity. We could invest time in learning and configuring the software, however it would seem like a waste of time.

Bob and I both agree that our current usage is a hinderance rather than a help; We both agree that it's reasonable for us to just point out particularly jarring formatting during code review.


labbati (approved):
It sounds good

PR #1633, opened 2022-06-30
Instrument pcntl_fork and reset tracing when it is called

jlesueur:
This will do 4 things:
- reset the open, closed, root and dropped spans globals
- reseed the random number generator used to assign span ids
- generate a root span if configured to do so
- turn background sending off and then on again, to remove timeouts when communicating with the agent

### DIFF
--- a/ext/php7/handlers_pcntl.c
+++ b/ext/php7/handlers_pcntl.c
@@ -17,7 +17,7 @@ ZEND_FUNCTION(ddtrace_pcntl_fork) {
     if (Z_LVAL_P(return_value) == 0) {
         // CHILD PROCESS
         ddtrace_coms_kill_background_sender();
-        ddtrace_init_span_stacks();
+        ddtrace_free_span_stacks();
         ddtrace_seed_prng();
         if (get_DD_TRACE_GENERATE_ROOT_SPAN()) {
             ddtrace_push_root_span();
--- a/ext/php7/handlers_pcntl.c
+++ b/ext/php7/handlers_pcntl.c
@@ -24,7 +24,7 @@ ZEND_FUNCTION(ddtrace_pcntl_fork) {
                 DDTRACE_G(distributed_parent_trace_id) = DDTRACE_G(open_spans_top)->span.span_id;
                 DDTRACE_G(trace_id) = DDTRACE_G(open_spans_top)->span.trace_id;
             }
-            ddtrace_free_span_stacks();
+            ddtrace_free_span_stacks(true);
             if (get_DD_TRACE_GENERATE_ROOT_SPAN()) {
                 ddtrace_push_root_span();
             }
--- a/ext/php7/handlers_pcntl.c
+++ b/ext/php7/handlers_pcntl.c
@@ -17,8 +17,13 @@ ZEND_FUNCTION(ddtrace_pcntl_fork) {
     if (Z_LVAL_P(return_value) == 0) {
         // CHILD PROCESS
         ddtrace_coms_kill_background_sender();
+        int parent_span_id = 0;
+        if (DDTRACE_G(open_spans_top) != NULL) {
+            parent_span_id = DDTRACE_G(open_spans_top)->span.span_id;
+        }
         ddtrace_free_span_stacks();
         ddtrace_seed_prng();
+        DDTRACE_G(distributed_parent_trace_id) = parent_span_id;
         if (get_DD_TRACE_GENERATE_ROOT_SPAN()) {
             ddtrace_push_root_span();
         }
--- a/ext/php7/handlers_pcntl.c
+++ b/ext/php7/handlers_pcntl.c
@@ -3,6 +3,9 @@
 
 #include "coms.h"
 #include "ddtrace.h"
+#include "span.h"
+#include "configuration.h"
+#include "random.h"
 #include "handlers_internal.h"  // For 'ddtrace_replace_internal_function'
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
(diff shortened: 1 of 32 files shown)


ext/php7/handlers_pcntl.c
```diff
@@ -14,11 +17,13 @@ ZEND_FUNCTION(ddtrace_pcntl_fork) {
     dd_pcntl_fork_handler(INTERNAL_FUNCTION_PARAM_PASSTHRU);
     if (Z_LVAL_P(return_value) == 0) {
         // CHILD PROCESS
-        // Until we full support pcntl tracing:
-        //   - kill the BGS
-        //   - disable further tracing on the forked process (also ensures all spans are dropped)
         ddtrace_coms_kill_background_sender();
-        ddtrace_disable_tracing_in_current_request();
+        ddtrace_init_span_stacks();
```
bwoebi, line 21: I think this needs a `ddtrace_free_span_stacks();`  before, to avoid leaking existing spans.

bwoebi:
Hey @jlesueur, thanks for the contribution!

I think this is a pretty great start!

On top of that, I think it should first assign the original top span id `DDTRACE_G(open_spans_top)->span_id` (if it open_spans_top is not NULL) to `DDTRACE_G(distributed_parent_trace_id)` so that distributed tracing works.

bwoebi:
@jlesueur Do you mind if we take over the PR and commit to your branch to finish it up? :-)

jlesueur:
@bwoebi I do not mind at all. I did push a fix just now before I saw your comment, but am happy to hand this off from here.

bwoebi:
Great, thanks a lot for the start, I'll take over from here then :-)

tests/ext/pcntl/pcntl_fork_long_running_autoflush.phpt
```diff
@@ -51,6 +60,16 @@ function long_running_entry_point()
 
 ?>
 --EXPECTF--
+parent is enabled
+Successfully triggered flush with trace of size 3
+child is enabled
 Successfully triggered flush with trace of size 1
+Cannot run tracing closure for long_running_entry_point(); spans out of sync
```
bwoebi, line 67: I'm not sure about this: forking inherently means that spans will be dropped and thus the trace end handlers skipped.

Shall we have a special marker for spans dropped by fork to silence these then?

jlesueur:
So that you have some use cases to guide you in answering these questions, we have two main uses in our code base:
1. an event processor pulling events off a queue, then forking to handle each event. Each fork is one of a couple dozen different event handlers. We use the event handler as the resource name in the child fork. We do not care about traces from the parent that pulls events off the queue.
2. various cron jobs that fork to perform work for each tenant. We care about the parent as well as the child trace. The parent forks up to 30k times, running 20-30 processes simultaneously. This tends not to look very good in the flame chart if the parent trace contains the child traces. It would be nice if the distributed tracing connection from the parent fork to the child fork was a configuration.

In both cases, we see each child fork as a complete process, similar to an http request.

bwoebi:
So, it is looping indefinitely within that topmost zend_hash_index_del call? Otherwise, I'm not seeing any particular location which loops here in the stacktrace?

jlesueur:
yes, that's where I see it looping when I get it into gdb.

jlesueur:
GDB shows it stuck inside this while: https://heap.space/xref/PHP-7.4/Zend/zend_hash.c?r=2c508c4d#1513
(copy/paste from the source doesn't do well in github comments...)

bwoebi:
The pure action of forking should not affect this at all or require any cleanup here. In fact, that should be totally unrelated to this particular PR, but be a pre-existing bug.

I am suspecting there is a rare problem with destruction order, which causes the internal objects (we create internal objects on PHP 7 to store some curl data directly on the handles, because WeakMap is not yet a thing on PHP 7) containing the curl multi mappings to be destroyed before the curl handles themselves.
I will try to verify the bug tomorrow.

bwoebi:
It was indeed trivial to reproduce the issue (this is with a debug build and crashes as I suspected):

```
circleci(buster):~/app$ DD_TRACE_CLI_ENABLED=1 php ../r.php 
/usr/local/src/php/Zend/zend_hash.c(1497) : ht=0x7f9642e89040 is already destroyed
php: /usr/local/src/php/Zend/zend_hash.c:71: _zend_is_inconsistent: Assertion `0' failed.
Aborted (core dumped)
circleci(buster):~/app$ cat ../r.php 
<?php

$ch = curl_init();
$mh = curl_multi_init();
curl_multi_add_handle($mh, $ch);

class A {
	public static $a;
	public $ch;
	public $mh;

	function __destruct() {
		curl_multi_remove_handle($this->mh, $this->ch);
	}
}

A::$a = new A;
A::$a->ch = $ch;
A::$a->mh = $mh;
```

jlesueur:
I was able to confirm, this does fix the hang I was seeing.

bwoebi:
I've got other priorities right now, but we're definitely still planning to release this quite soon, sorry that I haven't gotten to it in time for 0.77 - but I'll try to target the next one in any case!

jlesueur:
Thank you. We tested this in local/stage, and are running this in
production for the past week, and seeing good results. We'll be happy to
get back onto a mainline version of the extension :)

VladislavRatslavCB:
Any news, our company is also very interested in this fix.

bwoebi:
I'm looking at this again now, the adding arm64 support had a higher priority. I definitely want to finish this up very soon!

krakjoe (approved):
Nicely done :)

VladislavRatslavCB:
Thank you very much, everyone!

PR #1638, opened 2022-07-01
Run Wordpress testsuite actually against PHP 5

bwoebi:
### Description

The wordpress container now runs PHP 7. Creating an extra _php5 variant for the PHP 5 testsuite.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/dockerfiles/frameworks/contrib/Dockerfile
+++ b/dockerfiles/frameworks/contrib/Dockerfile
@@ -23,7 +23,7 @@ RUN set -xe; \
         php${V}-dom \
         php${V}-gd \
         php${V}-iconv \
-        php${V}-imagick \
+        # php${V}-imagick \
         php${V}-json \
         php${V}-intl \
         php${V}-fileinfo\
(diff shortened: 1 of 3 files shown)


dockerfiles/frameworks/contrib/Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN set -xe; \
         php${V}-dom \
         php${V}-gd \
         php${V}-iconv \
-        php${V}-imagick \
+        # php${V}-imagick \
```
labbati, line 26: Should we remove this one?

bwoebi: Don't know - the reason is that recent versions of imagick do understand pdf, which makes the older wordpress testsuites fail.

labbati:
```suggestion
```
Let's remove it then! If it fails in the future we'll get back to it

labbati (approved):

PR #1640, opened 2022-07-01
Reduce memory footprint of strpprintf

bwoebi:
### Description

This generally cuts fresh span memory overhead by 40%:
strpprintf under the hood uses a smart_str. Smart strings always use at least 224 bytes of memory.
By default for each string we allocate a span_fci (256 bytes), empty metrics and meta arrays (each 64 bytes) - and two strings: the span id and the resource name. Which are ~20 + 50-100 bytes, but we alloc 448 bytes.

This improvement will be (see https://github.com/php/php-src/pull/8902) on php-src master, but we do it for versions older than PHP 8.2 as well.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php7/compatibility.h
+++ b/ext/php7/compatibility.h
@@ -1,9 +1,7 @@
 #ifndef DD_COMPATIBILITY_H
 #define DD_COMPATIBILITY_H
 
-#include <TSRM/TSRM.h>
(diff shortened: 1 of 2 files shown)


morrisonlevi (approved):
Looks good.

PR #1642, opened 2022-07-02
Fix missing query string on http.url from integrations

Anilm3:
### Description

Some integrations add `http.url` to the root span, however:

- The query string was missing even when `datadog.trace.http_url_query_param_allowed` was set to `*`.
- `cleanQueryString` was including the fragment.
- `urlSanitize` was not taking into consideration the case in which only a query string was provided ('?...').
- Query string obfuscation was not available on PHP-land.

Within the extension, `http.url` is now only added if not present before serialization.

Tests are now using the defualt value of `datadog.trace.http_url_query_param_allowed`, which means all tests had to be updated to include the query string when available.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi:
> Within the extension, http.url is now only added if not present before serialization.

Is that necessary or won't the userland code just overwrite it simply?

Anilm3:
> > Within the extension, http.url is now only added if not present before serialization.
> 
> Is that necessary or won't the userland code just overwrite it simply?

Not strictly necessary but it's better to avoid computing the URL + obfuscating if it's just going to be replaced, from a performance perspective.

bwoebi:
I suspect the relative overhead to be minimal. However from a consistency perspective, we are trying to populate as much as possible ahead of time, so that users can e.g. dump a span, and see its values when debugging. Or just manipulate the value directly in code.
I.e. this change would break anyone currently doing `$rootspan->meta["http.url"] = str_replace($rootspan->meta["http.url"], "mysecret", "");`.

Anilm3:
> I suspect the relative overhead to be minimal. However from a consistency perspective, we are trying to populate as much as possible ahead of time, so that users can e.g. dump a span, and see its values when debugging. Or just manipulate the value directly in code. I.e. this change would break anyone currently doing `$rootspan->meta["http.url"] = str_replace($rootspan->meta["http.url"], "mysecret", "");`.

That's fine, what I can do as well is change some of the integrations to check for `http.url` instead and avoid recomputing, I don't think all of them can be changed but a most probably can.

tests/composer.json
```diff
@@ -130,5 +130,10 @@
         "scenario-options": {
             "dependency-licenses": false
         }
+    },
+    "config": {
+        "allow-plugins": {
+            "g1a/composer-test-scenarios": true
+        }
```
bwoebi, line 137:
```suggestion
```
Not sure why this is here, the config for allow-plugins is in line 15

PR #1652, opened 2022-07-11
Bump PHP minimal version

remicollet:
As this branch doesn't support PHP 5 anymore

> PHP 5 is not supported on this branch. Use the PHP-5 branch to build PHP 5.

so the package.xml should be consistent.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -80,7 +80,7 @@
     <dependencies>
         <required>
             <php>
-                <min>5.4</min>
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
This looks right to me - these sources here are not able to build PHP-5 anymore, and we don't ship the PHP-5 branches code to PECL nowadays.

Thanks for the contribution @remicollet :-)

Issue #1655, opened 2022-07-13
[BUG] No APM data after upgrading from PHP 7.3 to PHP 8.0


fabioknoedt:
### Bug description
We just upgraded our Laravel 8 PHP version from version 7.3 to 8.0 and as soon as it was live we stopped collecting all APM data from the Datadog dd-trace-php (version 0.76)

![image](https://user-images.githubusercontent.com/14126548/178759827-42e85d60-4cb5-4498-81aa-a966815a4c7b.png)

### PHP version
8.0.20

### Tracer version
0.76

### Installed extensions
```
        soap \
        bcmath \
        gd \
        intl \
        pdo_mysql \
        zip \
        xsl \
```

### OS info
`php:8.0.20-fpm-alpine`

### Upgrading info

 0.75

bwoebi:
Hey @fabioknoedt,

I've heard about issues related to laravel on 0.76.0 - apparently the laravel auto-instrumentation does not properly work with 0.76.0 anymore. I have not figured out yet why exactly.

It should however still instrument everything else (like database calls for example). Or does it really not collect anything at all?

bwoebi:
We've done an a bit more in depth diagnosis and noticed that most instrumentation of user code is broken if opcache is active.

We have found a fix for this, #1654 still needs fixing, then we will do a release as soon as possible.

fabioknoedt:
Hi @bwoebi! Thanks for the quick answer! We reverted back to the 0.75 and the data started to flow again for our APM. It is definitely something with the latest version. Let me know if you need any more info which would help you debug. Cheers!

fabioknoedt:
Missed your answer just now. That's bang on. We do have OPCache enabled. Thanks!

bwoebi:
Tracer version 0.76.1 has been published and contains the fix for the issue - opcache should not impact the tracer anymore :-)

PR #1656, opened 2022-07-13
(PHP 7.3+) Fix compatibility with opcache

### LINKED ISSUE #1654: [BUG] Enabling ddtrace.so disables PHP8 JIT
### Bug description
Installing the latest ddtrace via the datadog-setup.php script disables PHP's JIT, with the warning from FPM:
```
NOTICE: PHP message: PHP Warning:  JIT is incompatible with third party extensions that setup user opcode handlers. JIT disabled. in Unknown on line 0
```

Disabling ddtrace.so with no other changes allows JIT to become enabled again.

[This blog post](https://www.datadoghq.com/blog/engineering/php-8-observability-baked-right-in/) notes that the PHP tracer has been using the observer API since 0.52.0 so I'd expect this to not interfere?

### PHP version
```
root@9057ee5627c8:/app# php -v
PHP 8.1.7 (cli) (built: Jun 23 2022 07:57:40) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.7, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.7, Copyright (c), by Zend Technologies
```

### Tracer version
```
root@848c29c9c327:/app# php -r "echo phpversion('ddtrace').PHP_EOL;"
0.76.0
```

### Installed extensions
```
root@848c29c9c327:/app# php -m
[PHP Modules]
apcu
bcmath
Core
ctype
curl
date
ddappsec
ddtrace
dom
exif
fileinfo
filter
ftp
gd
gettext
hash
iconv
imagick
json
libxml
mbstring
memcached
msgpack
mysqli
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
readline
Reflection
session
shmop
SimpleXML
sockets
sodium
SPL
sqlite3
standard
sysvmsg
sysvsem
sysvshm
tokenizer
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zlib

[Zend Modules]
Zend OPcache
ddappsec
ddtrace
```

### OS info
Contained based off `php:8.1-fpm`
```
root@848c29c9c327:/app# cat /etc/os-release | grep -E "(NAME)|(VERSION)"
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
```

#### Output of phpinfo() (ddtrace >= 0.47.0)
image

bwoebi:
### Description

In the initial commit of the interceptor, we intended to be executed after opcache, to iterate over the loaded classes and functions, but this was never properly tested (in part because the opcache tests were run with DD_TRACE_CLI_ENABLED not set).
It turns out that it did properly work on PHP 7.0 - 7.2, as the opcache logic was happening in its startup handler.
However on PHP 7.3 and 7.4, opcache was moved to post_startup, causing it to be after our code. Similarly on PHP 8, we did use post_startup, but by mistake we first called our own code, then opcaches handler, defeating the the whole point of that code.

This has been fixed, the post_startup logic of opcache is now executed before our own, allowing us to read the result of zend_compile_file.

Additionally, the JIT does not allow for opcode overloads. By ensuring that our code installing the user_opcodes is executed after that check (happening in opcaches post_startup as well), we bypass that check.
However, given that the JIT is interacting very poorly with EX(opline) replacements (It assumes there is no code installing custom oplines...) as it has installed some metadata after the end of the op_array at some offset we cannot access on PHP 8.0, we have no choice, but create custom specialized logic for the JIT.

Fixes #1654 and #1655.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
It's off topic, but here's the failing test diff:

```diff
001+ Warning: fclose(): 11 is not a valid stream resource in /usr/local/src/php/ext/standard/tests/filters/bug54350.php on line 7
001- Warning: fclose(): 5 is not a valid stream resource in %s on line %d
```

Do we shift the stream resource numbers? I know we shift object numbers and such, and we should probably hunt all those down en-masse upstream and `%d` them, but what about streams?

morrisonlevi (commented):
Looks good to me, although I can't say if the opline fixup is correct and complete. Tests pass, though, so that's a very good sign. If it's critical that I deeply review that bit, then I'll need to take a big chunk of time and step through bits in debug builds with opcache + JIT and such to try and understand it.

On that note, is anyone aware of a tool to dump PHP executor globals and such and compare them after stepping over a call, like a debugger might? Can phpdbg help with this?

PR #1663, opened 2022-07-20
Build and package datadog-profiling in CI

morrisonlevi:
### Description

This adds jobs to the `build_packages` workflow in CircleCI that build
the datadog-profiling extension on x86_64-unknown-linux-gnu and
x86_64-alpine-linux-musl. It adjusts the `datadog-setup.php` install
script to use `extension = datadog-profiling.so` instead of
`zend_extension = datadog-profiling.so` and checks for profiling 
dependencies that weren't there before, which depending on your OS
includes `libgcc_s`, `libdl`,  `libpthread`, and `librt`.

This also updated the Alpine images to v3.16, which caused a few things
to break in CI. I fixed one of them already, but there's still an issue
with the warning `return-local-addr`. This was beyond my skill to just
drop in and fix quickly, so I dropped in a
`-Wno-error=return-local-addr` that should be removed once the fix is
in place. In the testing I did, the tracer built from this workflow can
run on Alpine 3.11 and earlier for x86-64, but I suspect for arm it
will not be possible to due to the libmusl v1.2 upgrade that happens
on Alpine 3.13.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1516,8 +1516,11 @@ jobs:
               cp -v profiling/datadog-profiling-install.sh "datadog-profiling/${triplet}/bin/"
             done
             tar -czf "datadog-profiling.tar.gz" "datadog-profiling/"
-      - store_artifacts:
-          path: datadog-profiling.tar.gz
+            rm -fr "datadog-profiling"
+      - persist_to_workspace:
+          root: .
+          paths:
+            - datadog-profiling.tar.gz
 
   compile_extension_centos:
     working_directory: ~/datadog
@@ -1741,8 +1744,10 @@ workflows:
                 - "7.4"
                 - "8.0"
 
-  build_profiling:
+  build_packages:
     jobs:
+      - "Prepare Code"
+
       - "cargo fetch":
           name: "cargo fetch - x86_64-alpine-linux-musl"
           docker_image: "datadog/dd-trace-ci:php-compile-extension-alpine-8.1"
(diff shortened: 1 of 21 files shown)


.circleci/config.yml
```diff
@@ -1434,6 +1434,123 @@ jobs:
...
+        type: string
+      triplet:
+        type: string
+    docker: [ image: << parameters.docker_image >> ]
+    steps:
+      - checkout
+      - restore_cache:
+          name: Restore Cargo Package Cache
+          keys:
+            - profiling-cargo-cache-<< parameters.triplet >>-{{ checksum "profiling/Cargo.toml" }}
```
r1viollet, line 1450: This key looks nice to me

r1viollet (dismissed):
Some nice optimisations to get the Rust build times down.
I think we could have even more caching. To be discussed for next PR, this is already big !
Congrats on putting all of this together. It must have been some amount of iterations.

bwoebi (approved):
Thanks, this looks great to me. I'm not sure whether we need that extra package_profiling step, but it does no harm, so, let's move on :-)

morrisonlevi:
I took at look at `package_profiling` and I think it does about 1 minute of harm! It spends about 1 minute downloading the container, attaching the workspace, re-uploading the workspace, etc. We can just do those things in "package extension". Good eye!

bwoebi (approved):
That seems to shave off a couple seconds. Guess there's nothing left to optimize now.

r1viollet (approved):
Good thinking on these last optims !
LGTM

PR #1664, opened 2022-07-22
Fix -Werror=address-of-packed-member

morrisonlevi:
### Description

The Alpine images recently updated the compiler and now this warning triggers there:

> tmp/build_extension/zend_abstract_interface/interceptor/php7/interceptor.c:30:24: error: taking address of packed member of 'struct <anonymous>' may result in an unaligned pointer value [-Werror=address-of-packed-member]

This fixes the issue by making it aligned to the same alignment as `zend_string`. 

This only occurred on PHP 7.0 for those who are curious.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zend_abstract_interface/interceptor/php7/interceptor.c
+++ b/zend_abstract_interface/interceptor/php7/interceptor.c
@@ -14,7 +14,10 @@ static ZEND_FUNCTION(pass)
 }
 
 // HACK: define a name for XDebug compatibility..., a fully functional zend_string for "{zend_pass}"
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
Never seen such a message. Btw. when taking `.str`, the cast to `(zend_string*)` is redundant, right?

morrisonlevi:
> Btw. when taking .str, the cast to (zend_string*) is redundant, right?

No, because we're still casting away the const.

Issue #1672, opened 2022-07-27
[BUG] Cannot parse IP-address from X-Forwarded-For from Azure application gateway


joelharkes:
### Bug description

Running the datadog in php fpm (behind nginx) via Azure application gateway, creates for each request the following notice: `NOTICE: PHP message: Not recognized as IP address: "1.1.1.1:23423"`

I have deducted it triggers when port number is appended to the `x-forwarded-for` header.

### Example

It fFails parsing IP address when x-forwarded-for header contains an port number: 

<img width="746" alt="image" src="https://user-images.githubusercontent.com/2960938/181280307-f405d42c-5b6b-4ad1-9d20-0a5cdf29b8e7.png">

creates error: 
```
NOTICE: PHP message: Not recognized as IP address: "1.1.1.1:23423"
```

### PHP version

```
PHP 7.4.28 (cli) (built: Feb 28 2022 07:33:39) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Xdebug v3.0.4, Copyright (c) 2002-2021, by Derick Rethans
    with Zend OPcache v7.4.28, Copyright (c), by Zend Technologies

```
### Tracer version

`0.76.1`

### Installed extensions

```
[PHP Modules]
bcmath
bz2
calendar
Core
ctype
curl
date
dba
dom
exif
FFI
fileinfo
filter
ftp
gd
gettext
gmp
hash
iconv
intl
json
ldap
libxml
mbstring
mysqli
mysqlnd
odbc
openssl
pcntl
pcre
PDO
pdo_dblib
pdo_mysql
PDO_ODBC
pdo_pgsql
pdo_sqlite
pdo_sqlsrv
pgsql
Phar
phpdbg_webhelper
posix
pspell
readline
Reflection
session
shmop
SimpleXML
soap
sockets
sodium
SPL
sqlite3
sqlsrv
standard
sysvmsg
sysvsem
sysvshm
tidy
tokenizer
xdebug
xml
xmlreader
xmlrpc
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Xdebug
Zend OPcache
```

### OS info

NAME="Alpine Linux"
VERSION_ID=3.15.5
PRETTY_NAME="Alpine Linux v3.15"

#### Output of phpinfo() (ddtrace >= 0.47.0)

i can add if necessary but seems a general bug.

[...]

PR #1675, opened 2022-07-28
Add support for ports on x-forwarded-for header

estringana:
According to [Azzure Application Gateway documentation](https://docs.microsoft.com/en-us/azure/application-gateway/how-application-gateway-works#modifications-to-the-request) the IPs on x-forwarded-for header
can contain the port. Our ip extraction code was not expeting ports as part of this header.

This PR fixes DataDog/dd-trace-php#1672

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/php7/ip_extraction.c
+++ b/ext/php7/ip_extraction.c
@@ -242,7 +242,7 @@ static bool dd_parse_x_forwarded_for(zend_string *zvalue, ipaddr *out) {
         while (value < end && *value == ' ') ++value;
         const char *comma = memchr(value, ',', end - value);
         const char *end_cur = comma ? comma : end;
-        succ = dd_parse_ip_address(value, end_cur - value, out);
+        succ = dd_parse_ip_address_maybe_port_pair(value, end_cur - value, out);
         if (succ) {
             succ = !dd_is_private(out);
         }
@@ -446,7 +446,14 @@ static bool dd_parse_ip_address_maybe_port_pair(const char *addr, size_t addr_le
     }
     const char *colon = memchr(addr, ':', addr_len);
     if (colon) {
-        return dd_parse_ip_address(addr, colon - addr, out);
+        char *tmp_addr = safe_emalloc(addr_len, 1, 1);
+        memcpy(tmp_addr, addr, addr_len);
+        tmp_addr[addr_len] = '\0';
+        int ret = inet_pton(AF_INET6, tmp_addr, &out->v6);
+        efree(tmp_addr);
+        if (ret != 1) {  // It has colon and it is not a valid ipv6 address
+            return dd_parse_ip_address(addr, colon - addr, out);
+        }
     }
 
     return dd_parse_ip_address(addr, addr_len, out);
--- a/ext/php7/ip_extraction.c
+++ b/ext/php7/ip_extraction.c
@@ -446,13 +446,9 @@ static bool dd_parse_ip_address_maybe_port_pair(const char *addr, size_t addr_le
     }
     const char *colon = memchr(addr, ':', addr_len);
     if (colon) {
-        char *tmp_addr = safe_emalloc(addr_len, 1, 1);
-        memcpy(tmp_addr, addr, addr_len);
-        tmp_addr[addr_len] = '\0';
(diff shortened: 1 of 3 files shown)


ext/php7/ip_extraction.c
```diff
@@ -446,7 +446,14 @@ static bool dd_parse_ip_address_maybe_port_pair(const char *addr, size_t addr_le
     }
     const char *colon = memchr(addr, ':', addr_len);
     if (colon) {
-        return dd_parse_ip_address(addr, colon - addr, out);
+        char *tmp_addr = safe_emalloc(addr_len, 1, 1);
+        memcpy(tmp_addr, addr, addr_len);
+        tmp_addr[addr_len] = '\0';
+        int ret = inet_pton(AF_INET6, tmp_addr, &out->v6);
```
Anilm3, line 452: The code above already checks for ipv6, I'm not sure if I'm missing something but why check again?

estringana: The code above is looking for the following pattern `[IPv6]:Port`. So given a valid `IPv6` without port, it would not enter on that conditional. Therefore, having a colon on the `addr`(on this second section) does not mean it is a `IPv4:Port`

bwoebi: The alternative would be to check whether there are two colons - an IPv6 always has at least two colons, while IPv4:port is at most one.

estringana: Thanks for pointing that out @bwoebi. I have refactor the code using this alternative

ext/php8/ip_extraction.c
```diff
@@ -446,7 +446,10 @@ static bool dd_parse_ip_address_maybe_port_pair(const char *addr, size_t addr_le
     }
     const char *colon = memchr(addr, ':', addr_len);
     if (colon) {
-        return dd_parse_ip_address(addr, colon - addr, out);
+        size_t colon_position = colon - addr;
+        if (colon_position + 1 < addr_len && !memchr(&colon[1], ':', &addr[addr_len] - &colon[1])) {
+            return dd_parse_ip_address(addr, colon_position, out);
+        }
```
Anilm3, line 452: Alternatively you can call `memchr` and `memrchr` and if they don't coincide then you have two colons.

tests/ext/extract_ip_addr.phpt
```diff
@@ -18,9 +18,11 @@ function test($header, $value) {
 test('x_forwarded_for', '2001::1');
 test('x_forwarded_for', '::1, febf::1, fc00::1, fd00::1,2001:0000::1');
 test('x_forwarded_for', '172.16.0.1');
-test('x_forwarded_for', '172.16.0.1, 172.31.255.254, 172.32.255.1, 8.8.8.8');
+test('x_forwarded_for', '172.16.0.1, 172.31.255.254, 172.32.255.1, 8.8.8.8, 1.2.3.4:456, [2001::1]:1111');
 test('x_forwarded_for', '169.254.0.1, 127.1.1.1, 10.255.255.254,');
 test('x_forwarded_for', '127.1.1.1,, ');
+test('x_forwarded_for', '1.2.3.4:456');
+test('x_forwarded_for', '[2001::1]:1111');
```
Anilm3, line 25: Can you add to this test an `x_forwarded`, `via` and `forwarded` variant using IPv6 with no port as a regression? For example (but you can add a few more with lists and such):

```
test('x_forwarded', 'for="2001:abcf::1"');
test('forwarded', 'for=2001:abcf:1f::55');
test('via', '2001:abcf:1f::55');

ext/php8/ip_extraction.c
estringana: Oh, that's one is a nice one 👏  I'll do it now

tests/ext/extract_ip_addr.phpt
estringana: Done

ext/php8/ip_extraction.c
estringana: Done

bwoebi (approved):
I think we've now arrived at the simplest possible implementation, LGTM :-)

PR #1677, opened 2022-07-29
Add "recurse" => true option to hook/trace config array

bwoebi:
### Description

Thus, it is now possible to write code like

```php
DDTrace\hook_function('greet', [
    'prehook' => function ($args) {
        echo "greet prehook: {$args[0]}\n";
    },
    'posthook' => function ($args) {
        echo "greet posthook: {$args[0]}\n";
    },
    'recurse' => true
]);
```

to have traces also trigger upon recursion

Fixes #1105.

### Readiness checklist
- [x] (only for Members) Changelog has been added to the release document.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

Issue #1105, opened 2020-11-24
[BUG] \DDTrace\trace_method not tracing nested/recursive method invocations


numyx:
### Bug description
Hello, I noticed that `\DDTrace\trace_method` (and `trace_function`) is not properly tracing nested or recursive method invocations, only the initial invocation is traced, but all nested invocations are not.

I wrote some mini code sample that highlights the issue:

```php
function test(int $x)
{
    if ($x > 0) {
        test($x - 1);
    }
}

\DDTrace\trace_function(
    'test',
    function (\DDTrace\SpanData $span, $args) {
        echo "Traced\n";
    }
);

test(3); // Expected 3 span and 3 "Traced" outputs, got only a single one.
```

While the pure recursive usage is probably an edge case, we noticed this issue when working with Symfony EventDispatcher where we are using event listeners to dispatch other events which results in nested calls of `EventDispatcher::dispatch` which are not getting traced by the built in Symfony integration.

### PHP version
```
PHP 7.4.12 (cli) (built: Nov  5 2020 20:24:10) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.12, Copyright (c), by Zend Technologies
    with ddtrace v0.50.0, Copyright Datadog, by Datadog
```

### Installed extensions
<details>
  <summary>Click to expand!</summary>
  
```
[PHP Modules]
amqp
apcu
bcmath
Core
ctype
curl
date
ddtrace
dom
fileinfo
filter
ftp
gd
hash
iconv
igbinary
intl
json
libxml
mbstring
mysqlnd
newrelic
openssl
pcntl
pcov
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
readline
Reflection
session
SimpleXML
sockets
sodium
SPL
sqlite3
standard
tokenizer
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
ddtrace
```
</details>

### OS info
```
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
NAME="Debian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
````

#### Output of phpinfo() (ddtrace >= 0.47.0)
```
Datadog PHP tracer extension
For help, check out the documentation at https://docs.datadoghq.com/tracing/languages/php/
[...]

PR #1678, opened 2022-07-29
Allow Symfony EventDispatcher::dispatch hooks to recurse

morrisonlevi:
### Description

Related to #1105 and #1677, this allows the `EventDispatcher::dispatch`
to recurse. Historically this was not possible, and spans for various
events weren't captured because of this. Now that it is possible, we
should allow it. Tests will need to be fixed up.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
I investigated the "route" changing in some of the error test. It's because [this method](https://github.com/DataDog/dd-trace-php/blob/d8daab4622dbfd57beb904e7247f8c169c897822/src/Integrations/Integrations/Symfony/SymfonyIntegration.php#L305) is called more than once since the 'controller' event callback triggers more than once now.

morrisonlevi:
Thanks, Bob. Updating the action for the first controller that runs seems like the best thing to do for now. Thanks for doing all the busy work on the rest of the spans!

bwoebi (approved):
Sounds like it looks good to you, so let's merge it?

PR #1679, opened 2022-08-01
[Profiling] Add fake frame when truncating stacks

morrisonlevi:
### Description

The exact fake frame name may change to align with other profilers,
but currently this information is missing and we can't easily tell if
a customer has hit a stalk walking bug, genuinely has weird stacks, or
if it's been truncated.

Here's a screenshot with an artificial max depth of 4 to illustrate it:
Screenshot of Datadog profiling page with an example of a truncated stack

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


r1viollet (approved):
That should make room ! 👍 
LGTM

PR #1681, opened 2022-08-01
[Profiling] Raise max stack depth to 512

morrisonlevi:
### Description

Some customers are seeing truncated stacks. This PR allows going up
to 512 frames deep now but it no longer allocates the max stack depth
when initializing samples buffer. This avoids waste when the stack is
much smaller than the limit, which it very often will be with a higher
limit.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/profiling.rs
+++ b/profiling/src/profiling.rs
@@ -271,7 +271,15 @@ unsafe fn collect_stack_sample(
     samples.reserve(max_depth); // todo: try_reserve
     let mut execute_data = top_execute_data;
 
-    while !execute_data.is_null() && samples.len() < samples.capacity() {
+    while !execute_data.is_null() {
+        if samples.len() >= max_depth {
+            samples.push(ZendFrame {
+                function: "[truncated]".to_string(),
+                file: None,
+                line: 0,
+            });
+            break;
+        }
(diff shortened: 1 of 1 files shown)


r1viollet (dismissed):
LGTM

profiling/src/profiling.rs
```diff
@@ -266,9 +266,8 @@ unsafe fn extract_line_no(execute_data: &zend_execute_data) -> u32 {
 unsafe fn collect_stack_sample(
     top_execute_data: *mut zend_execute_data,
 ) -> Result<Vec<ZendFrame>, Utf8Error> {
-    let mut samples = Vec::new();
-    let max_depth = 128;
-    samples.reserve(max_depth); // todo: try_reserve
+    let max_depth = 512;
+    let mut samples = Vec::with_capacity(max_depth >> 3);
```
r1viollet, line 270: Minor: We could explicitly write the reserve size

morrisonlevi: We could, but then we should probably assert it is less than `max_depth`. Didn't seem worth it.

bwoebi (approved):
Re-ack

PR #1682, opened 2022-08-01
Fix profiling config in randomized tests

morrisonlevi:
### Description

This was still using `zend_extension` instead of `extension` for `datadog-profiling.so`.

### Readiness checklist
- [x] Changelog has been added to the release document.
- [x] ~Tests added for this feature/bug.~ This is tests.

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.

PR #1680, opened 2022-08-01
[Profiling] Switch `<php>` to `<?php`

morrisonlevi:
### Description

When a call frame refers to top-level PHP code, the profiler will now
display this as `<?php` mirroring an open PHP tag instead of `<php>`.
This should be more intuitive to new users.

### Readiness checklist
- [x] Changelog has been added to the release document.
- [x] ~Tests added for this feature/bug.~

### Reviewer checklist
- [x] Appropriate labels assigned.
- [x] Milestone is set.

PR #1628, opened 2022-06-28
Add single span ingestion mechanism

bwoebi:
### Description

Adding a mechanism to sample individual spans.

### Readiness checklist
- [x] (only for Members) Changelog has been added to the release document.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #1684, opened 2022-08-02
Release 0.77.0-beta.1

morrisonlevi:
### Description

⚠️ The tracer and profiling packages for PHP 7 and 8 are built on CentOS 7. These packages will not run on older GNU Linux versions like CentOS 6, Debian 7, and Ubuntu 12.04.

### Added
- Add single span ingestion mechanism #1628
- Add "recurse" => true option to hook/trace config array #1677

### Changed
- Allow Symfony EventDispatcher::dispatch hooks to recurse #1678

### Fixed
- Fix JIT compatibility under macOS #1661
- Fix -Werror=address-of-packed-member #1664
- Add support for ports on x-forwarded-for header #1675

### Internal changes
- Move to CentOS 7; begin adding profiling deps #1660
- Add profiling sources #1606
- Build and package datadog-profiling in CI #1663
- Fix profiler config in randomized tests #1682

### Added
 - Add `process_id` and `runtime_version` tags #1606.
 - Add support for changing env vars per request, such as per-directory env var settings in Apache #1606.
 - Add fake frame when truncating stacks #1679. This way users can tell when the stack is truncated.

### Changed
 - Switch <php> to <?php #1680
 - Raise max stack depth to 512 #1681
 - Enable CPU Time profile by default #1663. This can disabled by setting the environment variable `DD_PROFILING_EXPERIMENTAL_CPU_TIME_ENABLED` to `0`, `off`, or `no`.
 - Change logging format #1606. Add a new log level `trace`, which is even more verbose than `debug`.
 - Stop sending a profile on every `phpinfo()` (or the equivalent command line option `--ri datadog-profiling`) #1606

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1959,6 +1959,13 @@ workflows:
       - pecl_build:
           requires: [ 'Prepare Code' ]
           name: "Build PECL"
+          # Pecl hates various strings allowed by semver and version_compare,
+          # for example 0.77.0-beta.1.
+          # So, we don't ship pecl packages for prereleases. In practice, if it
+          # contains a dash after the X.Y.Z bit then we ignore it.
+          filters:
+            branches:
+              ignore: /^release-\d+\.\d+\.\d+-.*/
       - "package extension":
           requires:
             - compile_alpine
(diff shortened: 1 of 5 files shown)


morrisonlevi:
Closing in favor of non-beta release.

PR #1696, opened 2022-08-10
Accept pseudonym on via header as specified by rfc7230

### LINKED ISSUE #1692: [BUG] Can not parse 'Via' header when running behind a GCP load balancer
### Bug description
GCP load balancers add the HTTP `Via` header in the following format: `{http-protocol-version} google`. For example `Via: 1.1 google`. This is compliant with [RFC7230](https://httpwg.org/specs/rfc7230.html#header.via).  
The datadog php tracer only allows for ip address (and optionally ports). It treats other tokens as errors despite that they are RFC-complaint and used in the wild.  

Here is the related implementation code 
https://github.com/DataDog/dd-trace-php/blob/ea44e8b9f0068cef2fb874621ebd62f7d5795ecf/ext/php8/ip_extraction.c#L367-L384

Accordingly, we get an error log line from the tracer for every http request. For example
```
[Tue Aug 09 08:46:42.731199 2022] [php:notice] [pid 159962] [client {redacted-ip}:0] Not recognized as IP address: "google"
```

### PHP version
PHP 8.1.8 (cli) (built: Aug  2 2022 07:31:17) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.8, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.8, Copyright (c), by Zend Technologies
    with ddtrace v0.77.0, Copyright Datadog, by Datadog

### Tracer version
0.77.0

#### Output of phpinfo() (ddtrace >= 0.47.0)
Bug seems unrelated to specific setup we run. Can add it if necessary.

estringana:
This PR fixes #1692

### Description

The ip extraction module was not taking into consideration that the via header can contain pseudonym instead of IPs. When a pseudonym was parsed it was not detected as an IP and the extension was logging out an error. This PR accept pseudonyms as part of the header as specified on [rfc7230](https://httpwg.org/specs/rfc7230.html#header.via)

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/ext/extract_ip_addr.phpt
+++ b/tests/ext/extract_ip_addr.phpt
@@ -66,7 +66,8 @@ test('via', '1.0 127.0.0.1, HTTP/1.1 [2001::1]:8888');
 test('via', 'HTTP/1.1 [2001::1, HTTP/1.1 [2001::2]');
 test('via', '8.8.8.8');
 test('via', '8.8.8.8, 1.0 9.9.9.9:8888,');
-test('via', '1.0 bad_ip_address, 1.0 9.9.9.9:8888,');
+test('via', '1.0 pseudonym, 1.0 9.9.9.9:8888,');
+test('via', '1.0 172.32.255.1 comment');
 test('via', ",,8.8.8.8  127.0.0.1 6.6.6.6, 1.0\t  1.1.1.1\tcomment,");
 test('via', '2001:abcf:1f::55');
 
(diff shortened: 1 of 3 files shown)


tests/ext/extract_ip_addr.phpt
```diff
@@ -66,7 +66,8 @@ test('via', '1.0 127.0.0.1, HTTP/1.1 [2001::1]:8888');
 test('via', 'HTTP/1.1 [2001::1, HTTP/1.1 [2001::2]');
 test('via', '8.8.8.8');
 test('via', '8.8.8.8, 1.0 9.9.9.9:8888,');
-test('via', '1.0 bad_ip_address, 1.0 9.9.9.9:8888,');
+test('via', '1.0 pseudonym, 1.0 9.9.9.9:8888,');
+test('via', '1.0 172.32.255.1 comment');
```
estringana, line 70: The RFC also specifies that after the IP/pseudonym there can be a comment. This comment was already skipped by the code but I did not find a test for it so I added it.

bwoebi (approved):
I'm not particularly fond of passing the bool around everywhere, but it's probably what works... so, ack :-)

estringana:
I agree with you @bwoebi. It's not the nicest of the approaches but I can't think of any other approach which does not involves a big refactor. Did some other approach come to your mind?

bwoebi:
No, otherwise I'd proposed something, let's just merge that :-) We can refactor if we need other special cases.

PR #1697, opened 2022-08-11
Fix span memory leak and packaging 

bwoebi:
### Description

Fixes a memory leak if Z_TYPE_INFO() reaches zero and the hashtable implementation will consider the value inexistent.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1614,7 +1614,7 @@ jobs:
             set -eu
             sudo apt-get install jq
             PIPELINE_ID=$(curl 'https://circleci.com/api/v2/project/gh/DataDog/dd-trace-php/pipeline?branch=PHP-5' | jq -r '.items[0].id')
-            WORKFLOW_ID=$(curl "https://circleci.com/api/v2/pipeline/$PIPELINE_ID/workflow" | jq -r '.items[]|select(.name == "build_packages").id')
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -1614,7 +1614,7 @@ jobs:
             set -eu
             sudo apt-get install jq
             PIPELINE_ID=$(curl 'https://circleci.com/api/v2/project/gh/DataDog/dd-trace-php/pipeline?branch=PHP-5' | jq -r '.items[0].id')
-            WORKFLOW_ID=$(curl "https://circleci.com/api/v2/pipeline/$PIPELINE_ID/workflow" | jq -r '.items[]|select(.name == "build_packages").id')
+            WORKFLOW_ID=$(curl "https://circleci.com/api/v2/pipeline/$PIPELINE_ID/workflow" | jq -r '.items[]|select(.name == "build_packages").id' | head -1)
```
morrisonlevi, line 1617: Sorry, what does this fix in practice? I understand you are limiting it to one line, but what was causing multiple lines in the first place?

morrisonlevi (approved):

.circleci/config.yml
bwoebi: That happens when you re-run a workflow.

PR #1698, opened 2022-08-15
[profiling] Add Unix Domain Socket (UDS) support

morrisonlevi:
### Description

If any configuration is set, it prioritizes that over UDS. If the
socket file doesn't exist then it falls back to http://localhost:8126.

There are tests for various units, and I have tested this manually for
end-to-end for both the default and setting `DD_TRACE_AGENT_URL` to
`unix:///var/run/datadog/apm.socket`.

I had to upgrade libdatadog; this PR is blocked on a release.
Various things have been renamed:

- `ddprof::profiles` -> `datadog_profiling::profile`
- `ddprof::exporter` -> `datadog_profiling::exporter`

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/config.rs
+++ b/profiling/src/config.rs
@@ -1,0 +1,142 @@
+use crate::sapi_getenv;
+pub use datadog_profiling::exporter::Uri;
+use libc::c_char;
+use std::ffi::CStr;
+use std::fmt::{Display, Formatter};
+use std::path::Path;
+pub use std::path::PathBuf;
+
+pub struct Env {
+    pub agent_host: Option<String>,
+    pub env: Option<String>,
+    pub profiling_enabled: Option<String>,
+    pub profiling_experimental_cpu_enabled: Option<String>,
+    pub profiling_experimental_cpu_time_enabled: Option<String>,
+    pub profiling_log_level: Option<String>,
+    pub service: Option<String>,
+    pub trace_agent_port: Option<String>,
+    pub trace_agent_url: Option<String>,
+    pub version: Option<String>,
+}
+
+impl Env {
+    /// Fetches the env var represented by `name`. Treats an empty string the
+    /// same as the env var not being set.
+    unsafe fn getenv(name: &CStr) -> Option<String> {
+        // CStr doesn't have a len() so turn it into a slice.
+        let name = name.to_bytes();
+
+        // Safety: called CStr, so invariants have all been checked by this point.
+        let val = sapi_getenv(name.as_ptr() as *const c_char, name.len());
+        let val = val.into_string();
+        if val.is_some() {
+            return val;
+        }
+
+        /* If the sapi didn't have an env var, try the libc.
+         * Safety: pointer comes from valid CStr.
+         */
+        let val = libc::getenv(name.as_ptr() as *const c_char);
+        if val.is_null() {
+            return None;
+        }
+
+        /* Safety: `val` has been checked for NULL, though I haven't checked that
+         * `libc::getenv` always return a string less than `isize::MAX`.
+         */
+        let val = CStr::from_ptr(val);
+        let maybe_str = val.to_str().ok();
+        // treat empty strings the same as no string
+        maybe_str.filter(|str| !str.is_empty()).map(String::from)
+    }
+
+    /// # Safety
+    /// This can only be called during rinit.
+    pub unsafe fn get() -> Self {
+        let env = Self::getenv(CStr::from_bytes_with_nul_unchecked(b"DD_ENV\0"));
+        let profiling_enabled = Self::getenv(CStr::from_bytes_with_nul_unchecked(
+            b"DD_PROFILING_ENABLED\0",
+        ));
+        let profiling_log_level = Self::getenv(CStr::from_bytes_with_nul_unchecked(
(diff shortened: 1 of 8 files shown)


profiling/src/config.rs
```diff
@@ -0,0 +1,142 @@
...
+        /* If the sapi didn't have an env var, try the libc.
+         * Safety: pointer comes from valid CStr.
+         */
+        let val = libc::getenv(name.as_ptr() as *const c_char);
+        if val.is_null() {
+            return None;
+        }
+
+        /* Safety: `val` has been checked for NULL, though I haven't checked that
+         * `libc::getenv` always return a string less than `isize::MAX`.
```
bwoebi, line 45: We only target 64 bit pointer builds, so this is fine.

profiling/src/config.rs
```diff
@@ -0,0 +1,142 @@
...
+impl Default for AgentEndpoint {
+    /// Returns a socket configuration if the default socket exists, otherwise
+    /// it returns http://localhost:8126/. This does not consult environment
+    /// variables.
+    fn default() -> Self {
+        let path = Path::new("/var/run/datadog/apm.socket");
+        if path.exists() {
+            return AgentEndpoint::Socket(path.into());
+        }
+        AgentEndpoint::Uri(Uri::from_static("http://localhost:8126/"))
```
bwoebi, line 109:
```suggestion
        AgentEndpoint::Uri(Uri::from_static("http://localhost:8126"))
```
At least the tracer reports it without trailing slash, for consistency?

profiling/src/lib.rs
```diff
@@ -268,45 +271,31 @@ extern "C" fn prshutdown() -> ZendResult {
...
+    if n_bytes == 0 || n_bytes > 5 {
         return None;
     }
 
-    string.make_ascii_lowercase();
-    match string.as_str() {
-        "1" | "on" | "yes" | "true" => Some(true),
-        "0" | "no" | "off" | "false" => Some(false),
-        _ => None,
+    // The empty string will not match since it was handled above.
```
bwoebi, line 288: Any reason the empty string is included in the set if it never matches?

profiling/src/lib.rs
```diff
@@ -554,25 +524,44 @@ fn log_add_tag(locals: &mut RefMut<RequestLocals>, key: &str, value: &str) {
...
+        if let Some(path) = trace_agent_url.strip_prefix("unix://") {
+            let path = PathBuf::from(path);
+            if path.exists() {
+                return AgentEndpoint::Socket(path);
+            }
+        }
+        match Uri::from_str(trace_agent_url) {
+            Ok(uri) => return AgentEndpoint::Uri(uri),
+            Err(err) => warn!("DD_TRACE_AGENT_URL was not a valid URL: {}", err),
+        }
```
bwoebi, line 548:
```suggestion
            } else {
                warn!("Unix socket specified in DD_TRACE_AGENT_URL does not exist: {} ", path);
            }
        } else {
            match Uri::from_str(trace_agent_url) {
                Ok(uri) => return AgentEndpoint::Uri(uri),
                Err(err) => warn!("DD_TRACE_AGENT_URL was not a valid URL: {}", err),
            }
        }
```
The URL is valid, but it doesn't exist, we should distinguish these cases.

profiling/src/lib.rs
```diff
@@ -800,4 +789,29 @@ extern "C" fn execute_internal(
...
+
+        let falsies = [
+            "0", "no", "off", "false", "No", "Off", "False", "NO", "OFF", "FALSE",
+        ];
+        for val in falsies {
+            assert_eq!(Some(false), parse_boolean(val), "Testing {}", val);
+        }
+
+        let non_boolean = ["a", "2", "-1", "truuue", "💯", "🦀!", "🔥🔥"];
```
bwoebi, line 811:
```suggestion
        let non_boolean = ["", "a", "2", "-1", "truuue", "💯", "🦀!", "🔥🔥"];
```
Add empty value as it's special cased in the parsing function.

profiling/src/lib.rs
morrisonlevi: Yes -- the lookup table is structured based on string length. There needs to be something there to compare to, it's just there are no truthy boolean strings that are 5 characters, and no falsey boolean strings that are 4 characters.

bwoebi: Wow, is this something obvious or should it warrant a comment?

bwoebi: Ah I see, it's an index into the table. I see the optimization :-) The comment was confusing me somehow.

profiling/src/lib.rs
morrisonlevi: I'll look at this in a bit. Thanks!

morrisonlevi: Looks good, but need to call `.to_string_lossy()` on the `path` since it doesn't implement `Display` (OS strings are not guaranteed to be UTF-8), so I won't commit this directly.

bwoebi: The alternative would be renaming the path variable and use the actual string which you extracted from the tracer_agent_url string.

bwoebi (approved):
Looks good to me :-)

bwoebi (approved):

PR #1699, opened 2022-08-16
Add arm support to docker images

bwoebi:
### Description

To be built with `docker buildx bake`.

This pull request adds arm support to our docker images. In particular it:
- Sets platform configurations in our docker-compose files
- Uses cross-compilation for our buster images (which, on average get rebuild more often, also for quicker testing of new things). Other docker images use qemu, which is slower, but still reasonable. (Collapsing the buster PHP builds into a couple files to not have too much copy paste per version to maintain.)
- Streamlines the randomized tests images to depend on our centos php images, as to avoid setting up PHP compilation twice (hence the changes to centos configure, so that it's also usable in randomized tests).

This is the foundational work, creating docker images for arm, to be able to actually run them in CI. This will be a separate pull request.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


docker-compose.yml
```diff
@@ -17,7 +17,7 @@ x-aliases:
         - agent
         - ddagent_integration
         - request-replayer
-        - elasticsearch2_integration
+        #- elasticsearch2_integration
```
bwoebi, line 20: Not sure about this, but there's no elasticsearch2 container for arm... Leaving this here will just completely block the containers from running on arm, so I think we can leave them commented out?

dockerfiles/ci/buster/0001-Better-support-for-cross-compilation.patch
```diff
@@ -0,0 +1,288 @@
+From f1ad9199eff32dd805bb5b155002bdb6ca69bf01 Mon Sep 17 00:00:00 2001
```
krakjoe, line 1: Just out of interest, and so there's a record somewhere ...

Is this included in any version upstream ? If not, why not ?

krakjoe (approved):
This all looks pretty sane to me, not able to test it directly as not blessed with arm yet :(

docker-compose.yml
krakjoe: For now, this seems okay.

dockerfiles/ci/buster/0001-Better-support-for-cross-compilation.patch
bwoebi: This patch has been applied to 8.1, but not to 8.0 on upstream. We only need it for cross-compiling 8.0. So maybe something which we can backport directly to 8.0...

PR #1700, opened 2022-08-17
[profiling] Avoid PHP per-request interning

morrisonlevi:
### Description

Since opcache overrides the intern function and you still have to
provide refcounting anyway, the previous version of the code would
leak.

At this point, the Rust <-> C globals bridge isn't providing any value
at all. Although in the future we there may be value in bridging these
things, such as to support INI handling, today it's just dragging us
down. So, I removed it, resulting in about twice as many lines deleted
as I added.

Also bump profiling version to v0.9.0.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -6,23 +6,22 @@ mod sapi;
 
 use crate::profiling::Profiler;
 use bindings as zend;
-use bindings::{
-    datadog_php_str, sapi_getenv, DatadogPhpProfilingGlobals, ZendExtension, ZendResult,
-};
+use bindings::{sapi_getenv, ZendExtension, ZendResult};
 use config::AgentEndpoint;
 use datadog_profiling::exporter::{Tag, Uri};
 use lazy_static::lazy_static;
-use libc::{c_char, c_int, c_ulong, c_void};
+use libc::c_int;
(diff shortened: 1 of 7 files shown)


profiling/src/lib.rs
```diff
@@ -140,8 +124,16 @@ static mut PREV_EXECUTE_INTERNAL: MaybeUninit<
...
 extern "C" fn minit(r#type: c_int, module_number: c_int) -> ZendResult {
+    /* When developing the extension, it's useful to see log messages that
+     * occur before the user can configure the log level. However, if we
+     * initialized the logger here unconditionally then they'd have no way to
+     * hide these messages. That's why it's done only for debug builds.
+     */
     #[cfg(debug_assertions)]
-    trace!("MINIT({}, {})", r#type, module_number);
+    {
+        logging::log_init(LevelFilter::Trace);
```
bwoebi, line 134: The alternative is doing an early direct call to `libc::getenv()` for logging? And later on the proper sapi_getenv+libc::getenv() fallback is used to cover other cases too?

bwoebi (approved):
Just dropping the interning altogether definitely works.

bwoebi (approved):

bwoebi (approved):

PR #1708, opened 2022-08-23
Prevent meta information to be unset due to PDO integration

frankdejonge:
### Description

We're trying to add meta information using `trace_method`, the information is not showing up in datadog, while for other calls this works perfectly fine. The main difference between the PDO integration (most notable for PDO.exec, PDO.query, and PDOStatement.execute) is that these set a _new_ array of meta instead of appending to the current. This may cause additional trace hooks that provide meta information to be overwritten, loosing the information.

Currently looking into how to test this.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
I agree, they really oughtn't create a new meta array. Thanks, and merging :-)

bwoebi:
Oh, @frankdejonge we require all commits to be signed, is it possible for you to sign it?

frankdejonge:
@bwoebi Hi, I've signed the commit.

frankdejonge:
@bwoebi just for my understanding. Our assumption that we're loosing this data due to the integrations is correct, right?

bwoebi (approved):

bwoebi:
@frankdejonge Yes, that's correct.

bruecktech:
@bwoebi this still somehow does not work for me
I'm logging the `$span->meta` array after I add the `team` and `company_id` tags in the `trace_method` call . But the `PDOStatement.execute` span still does not contain these tags, while the same works for `Redis.get`

```
span->meta:Array
(
    [version] => fce3cb4108771f7a55442096cf3f908f012878cc
    [team] => foo
    [company_id] => 1000000
)
```

My code looks like this. Is there anything else I'm missing?

```
\DDTrace\trace_method('PDOStatement', 'execute', function (\DDTrace\SpanData $span) use ($tags) {
                foreach ($tags as $key => $value) {
                    $span->meta[$key] = $value;
                }
            });
```

bwoebi:
I cannot reproduce this with 0.78.0, with your exact code. (with 0.77.0 I can)
Do you have still an old PDOIntegration.php file somehow?

bruecktech:
thanks, we sorted it out now. We had the composer package installed and updated but the extension was still on the old version. Good to know that the composer package is actually not needed.

PR #1720, opened 2022-08-26
Add Endpoint Profiling

morrisonlevi:
### Description

Add support for the endpoint profiling feature:
https://docs.datadoghq.com/profiler/connect_traces_and_profiles/#break-down-code-performance-by-api-endpoints

This is a highly requested feature.

Note that this also hooks up the profiling Tea tests located in
tests/tea. The PHP_VERSION_ID var was not set, so it was never being
built. To get the VERNUM, I imported the FindPhpConfig.cmake that I
used in the C version of the profiler.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


ext/php8/serializer.c
```diff
@@ -893,9 +898,28 @@ void ddtrace_serialize_span_to_array(ddtrace_span_fci *span_fci, zval *array) {
...
+    ZVAL_UNDEF(&prop_type_as_string);
     if (Z_TYPE_P(prop_type) > IS_NULL) {
-        _dd_add_assoc_zval_as_string(el, "type", prop_type);
+        ddtrace_convert_to_string(&prop_type_as_string, prop_type);
+        _add_assoc_zval_copy(el, "type", &prop_resource_as_string);
+    }
+
+    // Notify profiling for Endpoint Profiling.
+    if (profiling_notify_trace_finished) {
+        bool is_root = span->parent_id == DDTRACE_G(distributed_parent_trace_id);
```
bwoebi, line 910: you can reuse `top_level_span` here instead of recalculating it.

morrisonlevi: Thanks, it allowed a good bit of simplification.

cmake/Modules/FindPhpConfig.cmake
```diff
@@ -0,0 +1,137 @@
+include(FindPackageHandleStandardArgs)
```
morrisonlevi, line 1: This `FindPhpConfig.cmake` was copied from the Datadog/dd-prof-php repository. While it provides more than what we currently need, I didn't feel the need to remove anything; let me know if you disagree.

profiling/src/capi.rs
```diff
@@ -3,6 +3,62 @@
 
 use crate::bindings::zend_execute_data;
 use crate::RUNTIME_ID;
+use std::borrow::Cow;
+use std::fmt::{Display, Formatter};
+use std::marker::PhantomData;
+
+/// A non-owning, not necessarily null terminated, not utf-8 encoded, borrowed
+/// string. Must satisfy the requirements of [core::slice::from_raw_parts],
+/// notably it must not use the nul pointer even when the length is 0.
```
bwoebi, line 12: I'd like to note that this constraint (not-NULL) doesn't hold for some usages for zai_string_view on the C side. For everything passed to the profiler, this currently holds though.

bwoebi (approved):
C side looks good... and the Rust code probably works, so... :-)

profiling/src/capi.rs
morrisonlevi: The `components/string_view/string_view.h` version notes this requirement, maybe I should use that instead? They have the same layout.

Issue #1703, opened 2022-08-22
[BUG] Symfony CLI commands are traced as symfony.request operations


numyx:
### Bug description

Symfony console command traces on version 0.70.0 are recorded with operation name "console" (name of the console.php script), when upgrading to version 0.77.0 they instead are recorded as "symfony.request" (same as other http requests). Not certain if this is a bug or an intentional change. I would've expected "symfony.request" to be used for http requests only and for symfony cli commands to either keep the script name or otherwise have something like "symfony.console".

I did some debugging and it appears be related to the changes done to the symfony integration in https://github.com/DataDog/dd-trace-php/commit/867d30df5200744a81530f49519b95dc4451b4a4 and the lazy loading of integrations. It basically overrides the root span name to "symfony.request" whenever the symfony extension is loaded, maybe it should only do so on `Kernel::handle` (so when the kernel actually handles a request).

### PHP version
```
PHP 7.4.28 (cli) (built: Mar 29 2022 03:26:50) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
    with Zend OPcache v7.4.28, Copyright (c), by Zend Technologies
    with ddtrace v0.77.0, Copyright Datadog, by Datadog
```

### Tracer version
0.77.0

### Installed extensions
```
[PHP Modules]
apcu
bcmath
Core
ctype
curl
date
ddtrace
dom
fileinfo
filter
ftp
gd
hash
iconv
igbinary
intl
json
libxml
mbstring
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
readline
Reflection
session
SimpleXML
sockets
sodium
SPL
sqlite3
standard
tokenizer
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
ddtrace
```

### OS info
```
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
```

### Upgrading info

Updated from 0.70.0 to 0.77.0

bwoebi:
Hey @numyx,

you are right, this is an unintended side effect of the change. I'll revert to the old behaviour and only set the root span name in HttpKernel::boot().

numyx:
Hey @bwoebi 

[...]

PR #1723, opened 2022-09-01
Fix symfony service and root span name

numyx:
### Description

Previous fix for #1703 didn't cover the full problem since Kernel::boot is also used in symfony cli commands, instead root span name should be set to symfony.request only when Kernel actually handles a request which is the case when Kernel::handle is called.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
Thanks for the contribution! This supersedes #1718, so I merged bits of that into this PR.

morrisonlevi (approved):

PR #1724, opened 2022-09-01
Integrate with Symfony console commands

morrisonlevi:
### Description

This adds a span 'symfony.console.command.run' whose resource is the
name of the command (e.g. `$command->getName()`) if it's not null. It
attaches the command's class name as meta-data.

Suggestions are welcome for the name of the span and the metadata.

Here's a screenshot of the span created by the test:

screenshot of symfony.console.command.run span

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/Symfony/SymfonyIntegration.php
+++ b/src/Integrations/Integrations/Symfony/SymfonyIntegration.php
@@ -70,6 +70,35 @@ class SymfonyIntegration extends Integration
             ]
         );
 
+        $symfonyCommandsIntegrated = [];
+        \DDTrace\hook_method(
+            'Symfony\Component\Console\Command\Command',
+            '__construct',
+            null,
+            function ($This, $scope) use (&$symfonyCommandsIntegrated) {
+                if (isset($symfonyCommandsIntegrated[$scope])) {
+                    return;
+                }
+
+                $symfonyCommandsIntegrated[$scope] = true;
+
+                \DDTrace\trace_method($scope, 'run', [
+                    /* Commands can evidently call other commands, so allow recursion:
+                     * > Console events are only triggered by the main command being executed.
+                     * > Commands called by the main command will not trigger any event.
+                     * - https://symfony.com/doc/current/components/console/events.html.
+                     */
+                    'recurse' => true,
+                    'prehook' => function (SpanData $span) use ($scope) {
+                        $span->name = 'symfony.console.command.run';
+                        $span->resource = $this->getName() ?: $span->name;
+                        $span->service = \ddtrace_config_app_name('symfony');
+                        $span->type = Type::CLI;
+                        $span->meta['symfony.console.command.class'] = $scope;
+                    }]);
+            }
+        );
+
         $rootSpan = \DDTrace\root_span();
         if (null == $rootSpan) {
             return Integration::NOT_LOADED;
@@ -168,13 +197,12 @@ class SymfonyIntegration extends Integration
          * Since the arguments passed to the tracing closure on PHP 7 are mutable,
          * the closure must be run _before_ the original call via 'prehook'.
         */
-        $commands = [];
         \DDTrace\trace_method(
             'Symfony\Component\EventDispatcher\EventDispatcher',
             'dispatch',
             [
                 'recurse' => true,
-                'prehook' => function (SpanData $span, $args) use ($integration, &$injectedActionInfo, &$commands) {
+                'prehook' => function (SpanData $span, $args) use ($integration, &$injectedActionInfo) {
                     if (!isset($args[0])) {
                         return false;
                     }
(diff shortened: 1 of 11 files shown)


bwoebi (approved):
Looks good according to @morrisonlevi now.

lcobucci:
Folks, I believe this is creating a root span and preventing the flush of traces until the end of the process, which makes things a bit complicated for long running processes.

Output on v0.78

```sh
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Attempting integrations load; note: some integrations are only loaded on first usage
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Loaded integration web
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Loaded integration pcntl
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Loaded integration curl
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Integration guzzle not loaded. New attempts might be performed.
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Integration laravel not loaded. New attempts might be performed.
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Integration mysqli not available. New attempts WILL NOT be performed.
[2022-10-17T22:21:51+00:00] [ddtrace] [debug] - Integration DDTrace\Integrations\Symfony\SymfonyIntegration not loaded. New attempts might be performed.
Flushing trace of size 1 to send-queue for http://datadog:8126
Flushing trace of size 1 to send-queue for http://datadog:8126
Flushing trace of size 1 to send-queue for http://datadog:8126
Flushing trace of size 1 to send-queue for http://datadog:8126
[2022-10-17T22:21:52+00:00] [ddtrace] [debug] - Loaded integration DDTrace\Integrations\PDO\PDOIntegration
Flushing trace of size 6 to send-queue for http://datadog:8126
Flushing trace of size 3 to send-queue for http://datadog:8126
Flushing trace of size 1 to send-queue for http://datadog:8126
No finished traces to be sent to the agent
```

Output on v0.80

```sh
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Attempting integrations load; note: some integrations are only loaded on first usage
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Loaded integration web
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Loaded integration pcntl
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Loaded integration curl
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Integration guzzle not loaded. New attempts might be performed.
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Integration laravel not loaded. New attempts might be performed.
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Integration mysqli not available. New attempts WILL NOT be performed.
[2022-10-17T22:00:38+00:00] [ddtrace] [debug] - Integration DDTrace\Integrations\Symfony\SymfonyIntegration not loaded. New attempts might be performed.
Flushing trace of size 1 to send-queue for http://datadog:8126
[2022-10-17T22:00:39+00:00] [ddtrace] [debug] - Loaded integration DDTrace\Integrations\PDO\PDOIntegration
Flushing trace of size 13 to send-queue for http://datadog:8126
No finished traces to be sent to the agent
```

Was that an expected result of this change?

morrisonlevi:
> Folks, I believe this is creating a root span and preventing the flush of traces until the end of the process, which makes things a bit complicated for long running processes.

I can see how that is a problem. No, it wasn't intended. Maybe it shouldn't trigger if there isn't already a root span? @bwoebi, @krakjoe , any thoughts on this?

lcobucci:
> > Folks, I believe this is creating a root span and preventing the flush of traces until the end of the process, which makes things a bit complicated for long running processes.
> 
> I can see how that is a problem. No, it wasn't intended. Maybe it shouldn't trigger if there isn't already a root span? @bwoebi, @krakjoe , any thoughts on this?

Thanks for the quick response, @morrisonlevi! 

Based on my tests there's no root span when it is executed, which makes `Command#run()` the root span - aggregating all the subsequent span until the end of the execution.

I'm not that knowledgeable on this but it seems like we should avoid initiating a span on that level. Perhaps the hook should only be used to collect tags for the scope?

bwoebi:
Checking for prior root span existence sounds sensible to me, avoiding this class of issues nicely. Good idea!

lcobucci:
> Checking for prior root span existence sounds sensible to me, avoiding this class of issues nicely. Good idea!

I think I've expressed myself in the wrong way but wouldn't that completely disable the integration? I mean, if we only trigger that when there's a root span, there will never be a span created for it because `Command#run()` is the start of everything.

morrisonlevi:
> I mean, if we only trigger that when there's a root span, there will never be a span created for it because `Command#run()` is the start of everything.

This is probably true for long-running CLI apps, but there are also not-long-running CLI apps which you may or may not want to trace. I'm not sure my proposed fix (avoiding the span if there isn't a root span yet) is a good one, but I've caused an unintentional regression that we need to fix somehow.

A possibly better way would be to make the entry point for the queue worker into a distributed trace if there is already an active trace. I'm not familiar enough with the APIs off the top of my head to know if this is easily doable, but if so it should probably be done to prevent situations like this happening (theoretically we could break this again with future integrations otherwise). It's not a backwards compatible fix, of course, as you'd have to change code.

lcobucci:
Would an attribute be useful to provide some metadata on whether or not the command should be considered the root span?

lcobucci:
Anyway, since it has been confirmed I'll open an issue to track this in an easier way. Thanks!

PR #1733, opened 2022-09-09
[profiling] Match the service names of the tracer when unset or empty

morrisonlevi:
### Description

If DD_SERVICE is unset or empty, then the tracer defaults the service name to one of these:

 - the script name, aka argv[0]
 - cli.command
 - web.request

The profiler now does the same which should result in fewer services named unnamed-php-service.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/serializer.c
+++ b/ext/serializer.c
@@ -602,11 +602,11 @@ void ddtrace_set_root_span_properties(ddtrace_span_t *span) {
     zval *prop_name = ddtrace_spandata_property_name(span);
     if (strcmp(sapi_module.name, "cli") == 0) {
         ZVAL_STR(prop_type, zend_string_init(ZEND_STRL("cli"), 0));
-        if (SG(request_info).argc > 0) {
-            ZVAL_STR(prop_name, php_basename(SG(request_info).argv[0], strlen(SG(request_info).argv[0]), NULL, 0));
-        } else {
-            ZVAL_STR(prop_name, zend_string_init(ZEND_STRL("cli.command"), 0));
-        }
+        const char *script_name;
+        ZVAL_STR(prop_name,
+            (SG(request_info).argc > 0 && (script_name = SG(request_info).argv[0]) && script_name[0] != '\0')
+                ? php_basename(script_name, strlen(script_name), NULL, 0)
+                : zend_string_init(ZEND_STRL("cli.command"), 0));
     } else {
         ZVAL_STR(prop_type, zend_string_init(ZEND_STRL("web"), 0));
         ZVAL_STR(prop_name, zend_string_init(ZEND_STRL("web.request"), 0));
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -343,7 +343,16 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
                 .or(environment.profiling_experimental_cpu_enabled.as_deref())
                 .and_then(parse_boolean)
                 .unwrap_or(true);
-            locals.service = environment.service;
+            locals.service = environment.service.or_else(|| {
+                SAPI.get().and_then(|sapi| match sapi {
+                    Sapi::Cli => {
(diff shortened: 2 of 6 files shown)


ext/serializer.c
```diff
@@ -602,11 +602,11 @@ void ddtrace_set_root_span_properties(ddtrace_span_t *span) {
...
     if (strcmp(sapi_module.name, "cli") == 0) {
         ZVAL_STR(prop_type, zend_string_init(ZEND_STRL("cli"), 0));
-        if (SG(request_info).argc > 0) {
-            ZVAL_STR(prop_name, php_basename(SG(request_info).argv[0], strlen(SG(request_info).argv[0]), NULL, 0));
-        } else {
-            ZVAL_STR(prop_name, zend_string_init(ZEND_STRL("cli.command"), 0));
-        }
+        const char *script_name;
+        ZVAL_STR(prop_name,
+            (SG(request_info).argc > 0 && (script_name = SG(request_info).argv[0]) && script_name[0] != '\0')
```
r1viollet, line 607: Just stating what we are doing
- check if we have arguments
- check if argument is not null
- check if we have at least one character
Then we use the script name, otherwise we use a literal "cli.command".
Sounds good !

Minor: Future you might like more comments

profiling/src/lib.rs
```diff
@@ -343,7 +343,16 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
                 .or(environment.profiling_experimental_cpu_enabled.as_deref())
                 .and_then(parse_boolean)
                 .unwrap_or(true);
-            locals.service = environment.service;
+            locals.service = environment.service.or_else(|| {
+                SAPI.get().and_then(|sapi| match sapi {
```
r1viollet, line 347: So SAPI can reasonably not contain a value. Usage of and_then is nice.

profiling/src/sapi.rs
```diff
@@ -40,6 +44,32 @@ impl Sapi {
...
+             * be obtained for web requests too if we care.
+             */
+            Sapi::Cli => {
+                let request_info = &sapi_globals.request_info;
+                if request_info.argc > 0 && !request_info.argv.is_null() {
+                    // Safety: It's not null; the VM should do the rest.
+                    let cstr = unsafe { CStr::from_ptr(*request_info.argv) };
+                    return if !cstr.to_bytes().is_empty() {
+                        let osstr = OsStr::from_bytes(cstr.to_bytes());
+                        Path::new(osstr)
```
r1viollet, line 60: Are there any benefits of going through the path object versus just converting to a string slice ?

r1viollet (dismissed):
LGTM
Looks like you covered a lot of the cases that can go wrong.
A Minor question around a usage of Path.

profiling/src/lib.rs
morrisonlevi: Well... sort of. There will always be a SAPI but this a consequence of Rust hating mutable globals. I could remove the optionality by using unsafe bits, but the cost in this case is reasonable enough that I went with safe code. Particularly because there's a graceful failure case here (don't have to worry about returning an error or panic'ing).

profiling/src/sapi.rs
morrisonlevi: I need only the file name, not the whole path, which we get from `Path::file_name`. Path's need something that is `AsRef<OsStr>` and `&CStr` doesn't count, so I convert to `OsStr` in the middle.

ext/serializer.c
morrisonlevi: For completeness, I have never been able to trigger `"cli.command"`. But, that's what the tracer is doing today, so that's what I'm doing in the profiler too. These additional checks I added would save us from some safety issues if it ever does happen, but also treats an empty string as the same as being not set, which is something we generally do for env vars.

r1viollet (approved):
LGTM

bwoebi (approved):
Looks good to me too.

PR #1743, opened 2022-09-13
Handle non-binaries in installer

bwoebi:
### Description

Avoid installing non-binaries even when matching the pattern by default (i.e. with `--php-bin=all`), because these may handle args differently, have unexpected side-effects or simply hang the installer.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -151,8 +151,8 @@ function install($options)
             "Cannot copy files from '$tmpArchiveAppsecEtc' to '$installDir'",
             "cp -r " . escapeshellarg("$tmpArchiveAppsecEtc") . ' ' . escapeshellarg($installDir)
         );
-        $appSecRulesPath = $installDir . '/etc/recommended.json';
     }
+    $appSecRulesPath = $installDir . '/etc/recommended.json';
 
     // Actual installation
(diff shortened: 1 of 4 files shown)


datadog-setup.php
```diff
@@ -151,8 +151,8 @@ function install($options)
             "Cannot copy files from '$tmpArchiveAppsecEtc' to '$installDir'",
             "cp -r " . escapeshellarg("$tmpArchiveAppsecEtc") . ' ' . escapeshellarg($installDir)
         );
-        $appSecRulesPath = $installDir . '/etc/recommended.json';
     }
+    $appSecRulesPath = $installDir . '/etc/recommended.json';
```
bwoebi, line 155: Gives a notice on aarch64 otherwise.

datadog-setup.php
```diff
@@ -851,28 +856,30 @@ function ini_values($binary)
 
 function is_truthy($value)
 {
+    if ($value === null) {
+        return false;
```
bwoebi, line 860: and passing null to strtolower gives a deprecation notice on PHP 8.1.

krakjoe (approved):

PR #1745, opened 2022-09-15
[profiling] Avoid crashing when pcntl_fork is called

morrisonlevi:
### Description

The profiler doesn't support forking yet. However, it shouldn't crash.
This PR reduces some crash surface, but there is more to be done.

Since the tracer already has an API for overwriting the internal
function handler of a given function, I moved those bits into their own
file and made a note that the profiler also compiles these files.

I investigated adding full forking support but decided it would be too
much work to do right now. However, I left some comments about what I
learned, and did leave in some refactoring.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/pcntl.rs
+++ b/profiling/src/pcntl.rs
@@ -1,0 +1,118 @@
+use crate::bindings::{
+    datadog_php_install_handler, datadog_php_zif_handler, zend_execute_data, zend_long, zval,
+    InternalFunctionHandler,
+};
+use crate::{RequestLocals, PROFILER, REQUEST_LOCALS};
+use log::{error, warn};
+use std::cell::RefMut;
+use std::ffi::CStr;
+use std::mem::{forget, swap};
+
+static mut PCNTL_FORK_HANDLER: InternalFunctionHandler = None;
+static mut PCNTL_RFORK_HANDLER: InternalFunctionHandler = None;
+static mut PCNTL_FORKX_HANDLER: InternalFunctionHandler = None;
+
+fn handle_child_fork(mut locals: RefMut<RequestLocals>, retval: &mut zval) {
+    let result: Result<zend_long, _> = retval.try_into();
+    match result {
+        Ok(pid) => {
+            // The child gets pid of 0. For now, stop profiling for safety.
+            if pid == 0 {
+                match PROFILER.lock() {
+                    Ok(mut profiler) => {
+                        let mut new_profiler = None;
+                        swap(&mut *profiler, &mut new_profiler);
+
+                        /* Due to the swap, this has the old profiler.
+                         * Forget about it, it has garbage in it.
+                         */
+                        forget(new_profiler);
+                        locals.profiling_enabled = false;
+
+                        /* When we fully support forking remember:
+                         *  - Reset last_cpu_time and last_wall_time
+                         *  - Reset process_id and runtime-id tags.
+                         */
+                    }
+                    Err(err) => {
+                        error!(
+                            "Forked child failed to acquire profiler lock; a crash is likely: {}",
+                            err
+                        )
+                    }
+                }
+            }
+        }
+        Err(r#type) => {
+            warn!(
+                "Return type of pcntl_*fork* function was unexpected: {}",
+                r#type
+            )
+        }
+    }
+}
+
+unsafe fn handle_pcntl_fork(return_value: *mut zval) {
+    /* This hook is installed prior to knowing if the profiler will be enabled,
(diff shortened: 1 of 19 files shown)


profiling/src/pcntl.rs
```diff
@@ -0,0 +1,118 @@
...
+static mut PCNTL_RFORK_HANDLER: InternalFunctionHandler = None;
+static mut PCNTL_FORKX_HANDLER: InternalFunctionHandler = None;
+
+fn handle_child_fork(mut locals: RefMut<RequestLocals>, retval: &mut zval) {
+    let result: Result<zend_long, _> = retval.try_into();
+    match result {
+        Ok(pid) => {
+            // The child gets pid of 0. For now, stop profiling for safety.
+            if pid == 0 {
+                match PROFILER.lock() {
```
bwoebi, line 21: May that not ... just block indefinitely when the lock is held by the other thread at time of forking? At least docs aren't mentioning mutex poisoning in context of forking.

profiling/src/pcntl.rs
```diff
@@ -0,0 +1,118 @@
...
+unsafe extern "C" fn pcntl_forkx(execute_data: *mut zend_execute_data, return_value: *mut zval) {
+    // Safety: this function is only called if PCNTL_FORKX_HANDLER was set.
+    let handler = PCNTL_FORKX_HANDLER.unwrap();
+    handler(execute_data, return_value);
+    handle_pcntl_fork(return_value);
+}
+
+/* Safety: CStr invariants are met (at least one byte, last byte is null,
+ * no null bytes before that). Note that they require their own unsafe
+ * blocks because they are const, despite being in th
```
bwoebi, line 93: in th?

profiling/src/pcntl.rs
morrisonlevi: Yes, that can happen in theory. In practice, I don't support ZTS and the profiler is only used on the PHP thread (caveat, it's also called by the tracer when a trace finishes, so please don't do that in from a non-PHP thread 😅 ).

But, that doesn't mean I'm in the clear. I'm not sure about all of the internals of the objects that the Profiler stores, for instance.

profiling/src/pcntl.rs
morrisonlevi: Removed. I used to define these inside of `startup` since they are only used there, but I had to add `unsafe` around them even though `startup` was already an `unsafe` function, so I was starting to document that.

profiling/src/pcntl.rs
bwoebi: Why does it need a mutex at all, when it's only used on a single thread? Or is this in anticipation of future ZTS support?

morrisonlevi: 1. Yes, it's in anticipation of ZTS support. Even if it's not a critical part of the business, I think it's smart to have a plan there.
2. Having a global mutable variable like this requires safe abstractions in Rust or using unsafe code. Given the above point, it made sense to me (at the time anyway) to just go ahead and use the mutex.

morrisonlevi:
I did a bit more auditing. The biggest problems with multi-threaded code and forking are locks, for example:

- Every 10ms a thread will acquire a lock which contains a list of addresses of the VM interrupt globals, and then iterate over the list to trigger them all. The lock is also acquired at the start and end of every request, to add and remove the VM interrupt address respectively. For non-ZTS we could have a simpler synchronization mechanism, but since I wanted to make sure our foundations were ZTS compatible, that's how it is today.
- Printing anything to stderr will have a lock, such as log messages.
- I haven't fully unwrapped the crossbeam channels abstractions, so there could be locking in there too. So far I'm only seeing atomics, though.

If locks are held at the time of a fork and it's possible for the thread which forked to also get that lock, then we're likely to get a deadlock eventually.

So... the current state of the PR is that several opportunities for _crashing_ have been fixed, but there are still other scenarios that need more work. I'm thinking we should merge this as it's a clear improvement, but it's insufficient, so I should continue working on addressing these issues.

bwoebi:
I agree that we shall merge the PR as is - at least to prevent the crash. Iterative progress is much better than having no progress at all here :-)

bwoebi (approved):
Looks good :-)

r1viollet:
We had some discussions about implications of "forgetting" the old instance of the old profiler. It will increase memory overhead of a single fork. For sure this is an improvement over crashes. The challenges behind calling the actual destructors and allowing a new profiler to start are not so easy, so I agree with this first step ! I'll review some more.

r1viollet (approved):
LGTM

PR #1749, opened 2022-09-23
Fix behaviour of return false in trace hooks

bwoebi:
### Description

Previously return false was ignored in any trace hook, but the first one.

This now properly allows for (as intended) to write code like:
```
DDTrace\trace_method("mysqli", "prepare", function() { return false; });
```

Dropping all spans from trace hooks on mysqli->prepare.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


Anilm3 (approved):
LGTM

PR #1751, opened 2022-09-27
Fix #1750 configurable max payload size

### LINKED ISSUE #1750: AGENT_REQUEST_BODY_LIMIT is hard coded and no longer valid
If my understanding is correct, the reference mentioned here is no longer valid (hard cap at 10MB), but it's still enforced in the PHP library. This means that when increasing the agent limit, we're still hard capped at 10MB in this library and the payload is rejected.

https://github.com/DataDog/dd-trace-php/blob/873e48296ccdd4bd1324d417b3db82794367d66f/src/DDTrace/Transport/Http.php#L21-L23

krakjoe:
### Description

Fixes #1750

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/auto_flush.c
+++ b/ext/auto_flush.c
@@ -32,12 +32,10 @@ ZEND_RESULT_CODE ddtrace_flush_tracer() {
     zend_hash_index_add(Z_ARR(traces), 0, &trace);
 
     char *payload;
-    size_t size;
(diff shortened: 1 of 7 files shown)


bwoebi (approved):
Thanks, looks good to me, not sure if wee need all the configs, but we can change our minds later.

PR #1752, opened 2022-09-27
[profiling] Prevent deadlocks when forking

morrisonlevi:
### Description

Deadlocks are prevented by the following technique:

- Prior to the fork, all profiling threads will wait on the same barrier.
- After the fork in the parent process, all threads will wait on the same barrier. They are then free to resume regular operations.
- After the fork in the child process, the barrier will not be joined, the profiler object will be forgotten, and profiling will be disabled.

It should be easy to support profiling the child process from here.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/pcntl.rs
+++ b/profiling/src/pcntl.rs
@@ -2,90 +2,112 @@ use crate::bindings::{
     datadog_php_install_handler, datadog_php_zif_handler, zend_execute_data, zend_long, zval,
     InternalFunctionHandler,
 };
-use crate::{RequestLocals, PROFILER, REQUEST_LOCALS};
+use crate::{Profiler, PROFILER, REQUEST_LOCALS};
 use log::{error, warn};
-use std::cell::RefMut;
 use std::ffi::CStr;
 use std::mem::{forget, swap};
+use std::sync::atomic::Ordering;
 
 static mut PCNTL_FORK_HANDLER: InternalFunctionHandler = None;
 static mut PCNTL_RFORK_HANDLER: InternalFunctionHandler = None;
 static mut PCNTL_FORKX_HANDLER: InternalFunctionHandler = None;
 
-fn handle_child_fork(mut locals: RefMut<RequestLocals>, retval: &mut zval) {
-    let result: Result<zend_long, _> = retval.try_into();
-    match result {
-        Ok(pid) => {
-            // The child gets pid of 0. For now, stop profiling for safety.
-            if pid == 0 {
-                match PROFILER.lock() {
-                    Ok(mut profiler) => {
-                        let mut new_profiler = None;
-                        swap(&mut *profiler, &mut new_profiler);
+fn handle_pcntl_fork(
+    name: &str,
+    handler: unsafe extern "C" fn(*mut zend_execute_data, *mut zval),
+    execute_data: *mut zend_execute_data,
+    return_value: *mut zval,
+) {
+    /* Hold mutexes across the handler. If there are any spurious wakeups by
+     * the threads while the fork is occurring, they cannot acquire locks
+     * since this thread holds them, preventing a deadlock situation.
+     */
+    let mut profiler_lock = PROFILER.lock().unwrap();
+    if let Some(profiler) = profiler_lock.as_ref() {
+        profiler.fork_prepare();
+    }
 
-                        /* Due to the swap, this has the old profiler.
-                         * Forget about it, it has garbage in it.
-                         */
-                        forget(new_profiler);
-                        locals.profiling_enabled = false;
+    // Safety: we're calling the original handler with correct args.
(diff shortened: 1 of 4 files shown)


profiling/src/pcntl.rs
```diff
@@ -2,90 +2,124 @@ use crate::bindings::{
...
+     * the threads while the fork is occurring, they cannot acquire locks
+     * since this thread holds them, preventing a deadlock situation.
+     */
+    let mut profiler_lock = PROFILER.lock().unwrap();
+    if let Some(profiler) = profiler_lock.as_ref() {
+        profiler.fork_prepare();
+    }
+
+    // Safety: we're calling the original handler with correct args.
+    unsafe { handler(execute_data, return_value) };
```
r1viollet, line 31: We are chaining handlers together, seems good. I did not see this in previous design, could this be an issue in previous version ?

profiling/src/pcntl.rs
```diff
@@ -2,90 +2,124 @@ use crate::bindings::{
...
 }
 
-unsafe fn handle_pcntl_fork(return_value: *mut zval) {
-    /* This hook is installed prior to knowing if the profiler will be enabled,
-     * so we must guard it with a runtime check.
+fn stop_and_forget_profiling(maybe_profiler: &mut Option<Profiler>) {
+    /* In a forking situation, the currently active profiler may not be valid
+     * because it has join handles and other state shared by other threads,
+     * and threads are not copied when the process is forked.
+     * Additionally, if we've hit certain other issues like not being able to
```
r1viollet, line 86: great comment

profiling/src/profiling.rs
```diff
@@ -550,85 +566,127 @@ impl Uploader {
...
+
+                },
+            }
+        }
     }
 }
 
 impl Profiler {
     pub fn new() -> Self {
+        let fork_barrier = Arc::new(Barrier::new(3));
```
r1viollet, line 618: ok, so argument is number of threads that this barrier can block

profiling/src/profiling.rs
```diff
@@ -476,41 +474,57 @@ impl TimeCollector {
...
+                recv(upload_tick) -> message => {
+                    if message.is_ok() {
+                        last_wall_export = self.handle_timeout(&mut profiles, &last_wall_export);
+                    }
                 },
+
+                recv(self.fork_receiver) -> message => {
+                    if message.is_ok() {
+                        // First, wait for every thread to finish what they are currently doing.
+                        self.fork_barrier.wait();
```
r1viollet, line 522: I'm confused as to why we would always sync on the fork barrier ? Should this not only activate when we are forking ?

profiling/src/profiling.rs
```diff
@@ -550,85 +566,127 @@ impl Uploader {
...
+        let (fork_sender0, fork_receiver0) = crossbeam_channel::bounded(1);
+        let vm_interrupts = Arc::new(Mutex::new(Vec::with_capacity(1)));
         let (message_sender, message_receiver) = crossbeam_channel::bounded(100);
-        let (time_collector_cancel_sender, time_collector_cancel_receiver) =
-            crossbeam_channel::bounded(0);
-        let vm_interrupt_lock = Arc::new(Mutex::new(HashSet::with_capacity(1)));
         let (upload_sender, upload_receiver) = crossbeam_channel::bounded(UPLOAD_CHANNEL_CAPACITY);
+        let (fork_sender1, fork_receiver1) = crossbeam_channel::bounded(1);
         let time_collector = TimeCollector {
+            fork_barrier: fork_barrier.clone(),
```
r1viollet, line 625: This was slightly confusing (for a non rust user), but they will basically share the same barrier when using clone (if I'm reading this right).

profiling/src/profiling.rs
morrisonlevi: We do this whenever we receive a message on the `fork_receiver`, which will only be sent when we are forking.

profiling/src/profiling.rs
morrisonlevi: The clone is for the `Arc<_>`, it's not cloning the barrier itself (which wouldn't make sense and Rust probably prevents). Since 3 threads share ownership of the barrier object, it needs to be refcounted (or proven to outlive all the threads, which is hard).

profiling/src/pcntl.rs
morrisonlevi: We were doing the chaining outside of this helper function before, see https://github.com/DataDog/dd-trace-php/pull/1752/files#diff-3e3ad200a04052b3bf8073c003fa246b44ec22899b71d08b721acf18c5d878a5L80. I moved it inside so we could hold things across the fork easier.

r1viollet (approved):
LGTM !
Thanks for clarifying

PR #1761, opened 2022-10-13
Add troubleshooting to `README`

realFlowControl:
### Description

This PR will add a troubleshooting section and a test section to the `README.md`

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/README.md
+++ b/profiling/README.md
@@ -28,6 +28,18 @@ To see if the profiler is recognised by your PHP version as an extension you
 may run `php -d extension=target/debug/libdatadog_php_profiling.so --ri
 datadog-profiling` and check the output.
 
+The following commands will help you run the PHPT tests from
+[tests/phpt/](tests/phpt):
+
+```sh
+cp -v "$(find $(php-config --prefix) -name run-tests.php)" .
+export TEST_PHP_EXECUTABLE=$(which php)
+php run-tests.php -d extension=target/release/libdatadog_php_profiling.dylib tests/phpt
+```
+
+Be aware that the PHPT tests will fail with the debug version of the profiler,
+if you haven't already, build the release version with `cargo build --release`.
+
 ## Troubleshooting
 
 #### ld: symbol(s) not found for architecture arm64
--- a/profiling/README.md
+++ b/profiling/README.md
@@ -24,6 +24,10 @@ cases where they are ABI compatible.
 
 The command `cargo test` will run the tests on the profiler.
 
+To see if the profiler is recognised by your PHP version as an extension you
+may run `php -d extension=target/release/libdatadog_php_profiling.so --ri
+datadog-profiling` and check the output.
+
 ## Troubleshooting
 
 #### ld: symbol(s) not found for architecture arm64
--- a/profiling/README.md
+++ b/profiling/README.md
@@ -25,16 +25,14 @@ cases where they are ABI compatible.
 The command `cargo test` will run the tests on the profiler.
 
 To see if the profiler is recognised by your PHP version as an extension you
-may run `php -d extension=target/debug/libdatadog_php_profiling.so --ri
-datadog-profiling` and check the output.
+may run `/path/to/php -d extension=target/debug/libdatadog_php_profiling.so
+--ri datadog-profiling` and check the output.
 
 The following commands will help you run the PHPT tests from
 [tests/phpt/](tests/phpt):
 
 ```sh
-cp -v "$(find $(php-config --prefix) -name run-tests.php)" .
-export TEST_PHP_EXECUTABLE=$(which php)
-php run-tests.php -d extension=target/release/libdatadog_php_profiling.dylib tests/phpt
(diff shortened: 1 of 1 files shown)


profiling/README.md
```diff
@@ -1,17 +1,38 @@
...
+definitions across versions, as well as provide more idiomatic types in some
+cases where they are ABI compatible.
+
+## Testing
+
+The command `cargo test` will run the tests on the profiler.
+
+To see if the profiler is recognised by your PHP version as an extension you
+may run `php -d extension=target/release/libdatadog_php_profiling.so --ri
+datadog-profiling` and check the output.
```
morrisonlevi, line 29: The instructions above show `cargo build`, so maybe use `debug` instead of `release`?

realFlowControl: Good find, just changed it to `debug`

profiling/README.md
```diff
@@ -1,17 +1,50 @@
...
+
+To see if the profiler is recognised by your PHP version as an extension you
+may run `php -d extension=target/debug/libdatadog_php_profiling.so --ri
+datadog-profiling` and check the output.
+
+The following commands will help you run the PHPT tests from
+[tests/phpt/](tests/phpt):
+
+```sh
+cp -v "$(find $(php-config --prefix) -name run-tests.php)" .
```
krakjoe, line 35: This looks a bit strange, it may be better to document that running phpt tests requires run-tests.php.

profiling/README.md
```diff
@@ -1,17 +1,50 @@
...
+export TEST_PHP_EXECUTABLE=$(which php)
+php run-tests.php -d extension=target/release/libdatadog_php_profiling.dylib tests/phpt
```
krakjoe, line 37: .dylib .so mismatch, stick to one - probably the thing that is helpful for most people is using .so, but you could note somewhere that SHLIB_SUFFIX is different for mac (although these users are probably aware).

profiling/README.md
```diff
@@ -1,17 +1,50 @@
...
+To see if the profiler is recognised by your PHP version as an extension you
+may run `php -d extension=target/debug/libdatadog_php_profiling.so --ri
+datadog-profiling` and check the output.
+
+The following commands will help you run the PHPT tests from
+[tests/phpt/](tests/phpt):
+
+```sh
+cp -v "$(find $(php-config --prefix) -name run-tests.php)" .
+export TEST_PHP_EXECUTABLE=$(which php)
```
krakjoe, line 36: this might not be desirable either ... I would either document in pseudo like `TEST_PHP_EXECUTABLE=/path/to/php php /path/to/run-tests.php -dextension...` or use `/path/to/php run-tests.php -P` ...

tl;dr You don't want to leave people with the expectation that this particular test suite somehow relies on TEST_PHP_EXECUTABLE being exported, because that wouldn't be normal.

profiling/README.md
realFlowControl: Nice find with the `.dylib` 👍

realFlowControl: I was not aware of the `.dylib` file ending on mac, but I also was not a mac user two weeks ago 😉 
I would also think, that most people have the `.so` extension and even if not, it is pretty easy to find the file

profiling/README.md
realFlowControl: I changed the entire section in question to:
```sh
/path/to/php /path/to/run-tests.php -d extension=target/release/libdatadog_php_profiling.so tests/phpt
```

profiling/README.md
realFlowControl: I changed the entire section to match up with your last comment

krakjoe (approved):

profiling/README.md
morrisonlevi: The trouble is that `run-tests.php` is definitely PHP specific. You can't run the one from 7.1 on 8.1 nor vice versa. That is, unless I made some mistake when trying it.

profiling/README.md
morrisonlevi: I am not aware of the nuances of `.so` and `.dylib` on Mac, but I do know:

1. `cargo` builds a `.dylib` right now.
2. Renaming this to `datadog-profiling.so` when copying it into the `php-config --extension-dir` is fine, and it will load just fine by PHP when doing `-dextension=datadog-profiling`.

So... `php run-tests.php -d extension=target/release/libdatadog_php_profiling.dylib tests/phpt` is correct... at least for now.

realFlowControl: Found in the [Rust Linkage Docs](https://doc.rust-lang.org/reference/linkage.html) that for `cdylib` crates Rust "will create *.so files on Linux, *.dylib files on macOS, and *.dll files on Windows.".

I think it is safe to assume most people will use this together with Linux, so going with the `.so` file ending should be okay from my point of view.
CircleCI running on Linux also creates `.so` files: https://github.com/DataDog/dd-trace-php/blob/master/.circleci/config.yml#L1698

We could still add a hint for Mac users, that the file ending most likely will be `.dylib` in their case.

morrisonlevi (approved):

PR #1769, opened 2022-10-20
Implement rate limiter

krakjoe:
### Description

Implement rate limiter

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/limiter/limiter.c
+++ b/ext/limiter/limiter.c
@@ -1,0 +1,152 @@
+#include "limiter.h"
+
+#include <pthread.h>
+#include <unistd.h>
+#include <sys/mman.h>
+
+// clang-format Off
+
+#define NANOSECONDS_PER_SECOND 1000000000
+
+typedef struct {
+    /* limit from configuration DD_TRACE_RATE_LIMIT */
+    uint32_t limit;
+    struct {
+        /* beginning of window */
+        uint64_t open;
+        /* last hit inside this window */
+        uint64_t current;
+        /* total count for this window */
+        uint32_t samples;
+        /* allowed count for this window */
+        uint32_t allowed;
+        /* limit (tokens) for this window */
+        uint32_t limit;
+        /* rate for the last window that passed */
+        uint32_t rate;
+    } window;
+
+    pthread_mutex_t lock;
+} ddtrace_limiter;
+
+static ddtrace_limiter* dd_limiter;
+
+static inline uint64_t ddtrace_limiter_clock() {
+    struct timespec ts;
+
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) != 0) {
+        /* According to man page, this may only happen
+            if monotonic clock is not supported on this system. */
+        return 0;
+    }
+
+    return (ts.tv_sec * NANOSECONDS_PER_SECOND) + /* seconds as nanoseconds */
+            ts.tv_nsec;                           /* plus remaining nanoseconds */
+}
+
+void ddtrace_limiter_create() {
+    uint32_t limit = (uint32_t) get_global_DD_TRACE_RATE_LIMIT();
+
+    if (!limit) {
+        return;
+    }
+
+    /*
+     We share the limiter among forks (ie, forks need to write this memory), this requires that we map the memory as anonymous and shared
+    */
+    dd_limiter = mmap(NULL, sysconf(_SC_PAGESIZE), PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, -1, 0);
+
+    if ( dd_limiter == MAP_FAILED) {
+        dd_limiter = NULL;
+        return;
+    }
+
+    dd_limiter->limit = limit;
+    dd_limiter->window.open = ddtrace_limiter_clock();
+    dd_limiter->window.limit = limit;
+
+    pthread_mutexattr_t mattr;
+
+    /*
+     Mutex must have PROCESS_SHARED set to be usable among forks
+    */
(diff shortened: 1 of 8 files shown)


ext/limiter/limiter.c
```diff
@@ -0,0 +1,158 @@
...
+        /* limit (tokens) for this window */
+        uint32_t limit;
+        /* rate for the last window that passed */
+        uint32_t rate;
+    } window;
+
+    pthread_mutex_t lock;
+} ddtrace_limiter;
+
+static ddtrace_limiter* __limiter;
```
bwoebi, line 32: Can you please rename it to `dd_limiter` to conform with our naming schema for static vars? :-)

ext/limiter/limiter.c
```diff
@@ -0,0 +1,158 @@
...
+
+static inline uint64_t ddtrace_limiter_clock() {
+    struct timespec ts;
+
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) != 0) {
+        /* TODO something, can this fail ? */
```
bwoebi, line 38: Not really, don't care I think.

ext/limiter/limiter.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+void ddtrace_limiter_create() {
+    uint32_t limit = (uint32_t) get_global_DD_TRACE_RATE_LIMIT();
+
+    if (!limit) {
+        return;
+    }
+
+    __limiter = mmap(NULL, sysconf(_SC_PAGESIZE), PROT_WRITE, MAP_SHARED|MAP_ANONYMOUS, -1, 0);
```
bwoebi, line 53: Can I please have a comment why this is not just `malloc(sizeof(*__limiter))`?

bwoebi: Or rather, the limiter not being a simple non-pointer static?

ext/limiter/limiter.c
```diff
@@ -0,0 +1,158 @@
...
+        /* no previous window, or samples */
+        effective = (double) __limiter->window.allowed / __limiter->limit;
+
+        goto __ddtrace_limiter_rate_leave;
+    }
+
+    effective = (
+                    (double) (__limiter->window.allowed / __limiter->limit) + /* current rate */
+                    __limiter->window.rate                                    /* previous window rate */
+                ) / 2.0;                                                      /* spread over last two windows */
```
bwoebi, line 138: doesn't that need to take the time elapsed into account? I.e. `1 + (window.current - window.open) / NANOSECONDS_PER_SECOND` or something like that instead of `2.0`? Otherwise there should be a steep cliff by 50% in the number (as `window.allowed` starts over at 0 again?

ext/limiter/limiter.c
```diff
@@ -0,0 +1,158 @@
...
+        __limiter->window.open = __limiter->window.current;
+        __limiter->window.samples = 0;
+        __limiter->window.allowed = 0;
+    }
+
+    if (__limiter->window.limit != __limiter->limit) {
+        __limiter->window.limit = MIN(
+            __limiter->limit,
+            __limiter->window.limit + ((elapsed / NANOSECONDS_PER_SECOND) * __limiter->limit));
+    }
```
bwoebi, line 109: When is that ever true? I don't see any assignments except at startup time?

ext/limiter/limiter.c
krakjoe: this now mirrors python

ext/limiter/limiter.c
krakjoe: > When is that ever true?

Never :)

ext/limiter/limiter.c
```diff
@@ -0,0 +1,155 @@
+#include "limiter.h"
+
+#include <pthread.h>
+#include <unistd.h>
+#include <sys/mman.h>
+
+// clang-format Off
+
+#define NANOSECONDS_PER_SECOND 1000000000
```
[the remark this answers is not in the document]
bwoebi: No, it has 9 zeroes, like it should.

ext/limiter/limiter.c
```diff
@@ -0,0 +1,155 @@
...
+
+        goto __ddtrace_limiter_rate_leave;
+    }
+
+    effective = (double) (
+                    (double) dd_limiter->window.allowed / dd_limiter->window.samples + /* current rate */
+                    dd_limiter->window.rate                                    /* previous window rate */
+                ) / 2.0;                                                      /* spread over last two windows */
+
+__ddtrace_limiter_rate_leave:
```
bwoebi, line 137: Use else instead of goto?

bwoebi:
I see no fundamental issues in this PR anymore.

I'd like to note the following behaviour (which are also present in our python implementation, so it should be fine?):
it's always the first X traces happening one second after the last X traces. Assuming the first trace was sent at instant t=0, when X is reached at t=0.01, then no other traces will be sampled until t=1. (i.e. we have no smooth sliding window, but hard cut offs.)

I'm willing to merging this PR as is. Thanks for implementing, Joe :-)

bwoebi (approved):

ext/limiter/limiter.c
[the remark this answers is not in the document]
morrisonlevi: I counted multiple times to be sure. Oh well.

ext/limiter/limiter.c
krakjoe: Yes

PR #1773, opened 2022-10-21
improve ldconfig operation in installer

krakjoe:
### Description

The installer depends on a working ldconfig (for glibc install) and assumes it is in PATH.

It is becoming common place for /sbin to be omitted from the PATH of normal users, so we must search outside of PATH in the most common paths for a working ldconfig and use its full path on execution.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -499,6 +499,57 @@ function require_binaries_or_exit($options)
     return $selectedBinaries;
 }
 
+function search_for_working_ldconfig()
+{
+    static $path;
+
+    if ($path) {
+        return $path;
+    }
+
+    $paths = [
+        "/sbin", /* this is most likely path */
+        "/usr/sbin",
+        "/usr/local/sbin",
+        "/bin",
+        "/usr/bin",
+        "/usr/local/bin",
+    ];
+
+    $search = function (&$path) {
+        exec("find $path -name ldconfig", $found, $result);
+
+        if ($result == 0) {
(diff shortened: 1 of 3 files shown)


bwoebi:
I wonder, wouldn't `shell_exec('PATH=$PATH:/sbin:/usr/sbin:... command -v ldconfig')` not be much simpler, than iterating manually with find through everything, or is that not good enough?

krakjoe:
> I wonder, wouldn't `shell_exec('PATH=$PATH:/sbin:/usr/sbin:... command -v ldconfig')` not be much simpler, than iterating manually with find through everything, or is that not good enough?
> 
> Or possibly directly `PATH=$PATH:/sbin:/usr/sbin:... ldconfig | grep ...`

Typically, find will be executed a single time (ldconfig is almost always in /sbin), and yield the full path to the binary we want to execute. We may execute that binary several more times without having to search for it in PATH.

The output of the installer, when things do go wrong, becomes more meaningful when we try to use the full path - ie, we know that if someone reports a bug with unqualified "ldconfig" error, that their system has no ldconfig in any of the paths we search, or PATH. Whereas if someone reports an error with "/sbin/ldconfig" we know the problem is not that ldconfig was not found.

tl;dr I think in this case I prefer verbosity.

bwoebi (approved):
In any case, it works. So let's include the fix then :-)

Issue #1764, opened 2022-10-17
[Feature] Allow INI Configuration for DD_PROFILING_ENABLED


JanThiel:
**Is your feature request related to a problem? Please describe.**
We want to profile code which is executed by a 3rd party CLI script. It's a scheduler system calling php "cron-like" tasks. We have no way to inject ENV variables into the call stack which calls the actual php executable we want to profile.
Tracing works fine as we can enable this within the `98-ddtrace.ini` file. But we need the Profiling...

**Describe the solution you'd like**
Allow to enable profiling using the `ini` way. That will allow us to simply enable the profiling for all executions using a particular php executable and config set.

**Describe alternatives you've considered**
Executing the final php calls manually. Which doesn't make sense at all as we are talking about several hundreds per hour where we don't really know which execution will contain the desired spans we are looking for.

**Additional context**

https://docs.datadoghq.com/tracing/trace_collection/library_config/php/#environment-variable-configuration
=> DD_PROFILING_ENABLED
![image](https://user-images.githubusercontent.com/3857275/196219634-563b2807-4b4d-4efc-91e5-c09f129bc212.png)

morrisonlevi:
Thanks for the report! This quarter we are planning to support INI configuration for all settings for the profiler, which of course would include this setting. The expectation is for it to be `datadog.profiling.enabled`.

JanThiel:
Thanks @morrisonlevi Good to hear that :-)
And by the way, thank you very much for the amazing installer. When we last tried DD PHP Tracing the installation was ... no fun ... to put it mildly ;-). 
Today it's amazingly easy. Very well done!

pierotibou:
Hello @JanThiel, this is now possible since [our last release](https://github.com/DataDog/dd-trace-php/releases/tag/0.82.0).

morrisonlevi:
Note that `datadog.profiling.enabled` defaults to being on in this release when you pass `--enable-profiling` to the installer. You should be able to upgrade, and then remove the env var if you like (but env vars are still supported). I recommend moving from env vars `DD_SERVICE`, `DD_ENV`, and `DD_VERSION` to ini settings `datadog.service`, `datadog.env`, and `datadog.version` going forward!

PR #1775, opened 2022-10-25
[profiling] Add .ini support

morrisonlevi:
### Description

Implements #1764. This uses the ZAI config component to add support for .ini configuration in profiling. Thank, @bwoebi, for laying the groundwork for this in #1765.

Note that the profiler now defaults to being enabled, but it is not installed by default. This is because it on Alpine it has a dependency on `libgcc` and the tracer does not have this dependency. If a user is only using the tracer and then upgrades, we would emit warnings about undefined symbols and such when it starts up whether they use the profiler or not.

This also moves the default of DD_AGENT_PORT from 8126 to 0. This is easier to work with when working with fallbacks to other connection settings.

There's a good bit of boilerplate to help Rust understand lifetimes for `zai_string_view`. There's also code which tries to bind lifetimes of certain strings to the owning object. In PHP's reference counted model, this is not quite correct. However, it is an improvement on what is there. Given that we are only borrowing this data for a very short amount of time, we are not executing user code during this time, and we ultimately copy it, we should be okay.

### Readiness checklist
- [x] Manually check that the current install docs still work as expected:
    - [x] Apache
    - [x] CLI
    - [x] Nginx

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -1130,6 +1130,34 @@ function get_ini_settings($requestInitHookPath, $appsecHelperPath, $appsecRulesP
             'commented' => false,
             'description' => 'Enables the appsec module',
         ],
+
+        [
+            'name' => 'datadog.profiling.enabled',
+            'default' => '1',
+            'commented' => true,
+            'description' => 'Enable the Datadog profiling module.',
+        ],
+        [
+            'name' => 'datadog.trace.endpoint_collection_enabled',
+            'default' => '1',
+            'commented' => true,
+            'description' => 'Whether to enable the endpoint data collection in profiles.',
+        ],
+        [
+            'name' => 'datadog.profiling.experimental_cpu_time_enabled',
+            'default' => '1',
+            'commented' => true,
+            'description' => 'Enable the CPU profile type.',
+        ],
+        [
+            'name' => 'datadog.profiling.log_level',
+            'default' => 'off',
+            'commented' => true,
+            'description' => 'Set the profiler’s log level.'
+                . ' Acceptable values are off, error, warn, info, and debug.'
+                . ' The profiler’s logs are written to the standard error stream of the process.',
+        ],
+
         [
             'name' => 'datadog.trace.request_init_hook',
             'default' => $requestInitHookPath,
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -1154,7 +1154,7 @@ function get_ini_settings($requestInitHookPath, $appsecHelperPath, $appsecRulesP
             'default' => 'off',
             'commented' => true,
             'description' => 'Set the profiler’s log level.'
-                . ' Acceptable values are off, error, warn, info, and debug.'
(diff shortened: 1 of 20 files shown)


datadog-setup.php
```diff
@@ -1130,6 +1130,34 @@ function get_ini_settings($requestInitHookPath, $appsecHelperPath, $appsecRulesP
...
+            'default' => '1',
+            'commented' => true,
+            'description' => 'Enable the CPU profile type.',
+        ],
+        [
+            'name' => 'datadog.profiling.log_level',
+            'default' => 'off',
+            'commented' => true,
+            'description' => 'Set the profiler’s log level.'
+                . ' Acceptable values are off, error, warn, info, and debug.'
```
realFlowControl, line 1157: Technically there is also `trace`, should we document this?

realFlowControl (approved):
LGTM 🎉

profiling/src/profiling.rs
```diff
@@ -187,29 +180,22 @@ unsafe fn extract_function_name(func: &zend_function) -> String {
...
+        buffer.push(b'|');
     }
 
-    if !func.common.scope.is_null() {
-        let class_name = zend_string_to_bytes((*func.common.scope).name);
-        if !class_name.is_empty() {
-            buffer.extend_from_slice(class_name);
-            buffer.extend_from_slice(b"::");
-        }
+    // todo: probably use EX(This) instead of func.common.scope.
```
bwoebi, line 189: I think using the func.common.scope is better here, given that stacktraces generally mirror the declaration context for methods and not the run-time late static binding or $this context.

datadog-setup.php
morrisonlevi: Yes, good point.

profiling/src/config.rs
```diff
@@ -150,3 +63,444 @@ impl Display for AgentEndpoint {
...
+            ENTRIES.as_mut_ptr(),
+            ENTRIES.len().try_into().unwrap(),
+            Some(env_to_ini_name),
+            module_number,
+        );
+        assert!(tmp); // It's literally return true in the source.
+    }
+}
+
+#[cfg(test)]
```
r1viollet, line 450: Is there value in considering more types of user inputs ? Non-ut8, long strings etc.. 
Is this already covered by tracer tests ?

r1viollet (approved):
Very thoughtful management of strings in the C/Rust bindings. We can rethink things if we need to avoid copies in the future.
I reviewed superficially and things look good. Nice to see we are converging. Minor question about tests.
LGTM

profiling/src/config.rs
bwoebi: Yeah, a lot of it is directly covered by unit tests on the zai/config component.

bwoebi (approved):
Already looked multiple times over it, didn't think you were waiting for an ack from me, but here it is :-)

morrisonlevi:
Thank you for your help with this, Bob! Thanks Erwan and Florian for the reviews.

PR #1778, opened 2022-10-27
Avoid deprecated zend_atol on PHP 8.2

morrisonlevi:
### Description

This was deprecated in PHP 8.2 with the recommendations to use:

 1. `ZEND_STRTOL()` for general purpose string to long conversion.
 2. A variant of `zend_ini_parse_quantity()` for parsing ini quantities.

For now at least, since this code is ZAI config which deals with .ini settings, I picked the latter.

We should probably re-plumb the function calls at some point to pass the ini setting name to this function so we can use `zend_ini_parse_quantity_warn()` instead.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zend_abstract_interface/config/config_decode.c
+++ b/zend_abstract_interface/config/config_decode.c
@@ -117,7 +117,20 @@ static bool zai_config_is_valid_int_format(const char *str) {
 static bool zai_config_decode_int(zai_string_view value, zval *decoded_value) {
     if (!zai_config_is_valid_int_format(value.ptr)) return false;
 
-    ZVAL_LONG(decoded_value, zend_atol(value.ptr, value.len));
+#if PHP_VERSION_ID >= 80200
+    zend_string *zstr = zend_string_init(value.ptr, value.len, 0);
+    zend_string *err = NULL;
+    zend_long l = zend_ini_parse_quantity(zstr, &err);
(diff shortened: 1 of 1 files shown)


zend_abstract_interface/config/config_decode.c
```diff
@@ -117,7 +117,20 @@ static bool zai_config_is_valid_int_format(const char *str) {
...
 
-    ZVAL_LONG(decoded_value, zend_atol(value.ptr, value.len));
+#if PHP_VERSION_ID >= 80200
+    zend_string *zstr = zend_string_init(value.ptr, value.len, 0);
+    zend_string *err = NULL;
+    zend_long l = zend_ini_parse_quantity(zstr, &err);
+    if (err) {
+        // what to do with err?
+        zend_string_free(err);
+    }
```
bwoebi, line 127: Given that we anyway check zai_config_is_valid_int_format() to be integers with a possibly leading - only (incl. spaces), this basically should never fail. So, it should not be too important, just free and `return false;` in that case I'd say.

morrisonlevi: Thanks, I've updated the PR and left a comment about this.

bwoebi (approved):
Looks good to me :-)

PR #1790, opened 2022-11-07
Bug: Invocable classes don't work

deleugpn:
### Description

I recently discovered what seems to be a bug on Datadog instrumentation where it just crashes when accessing `$this` in the context of an invocable class from a closure. I also noticed that if I don't convert the invocable class into a closure it also doesn't work at all.

Following the guide, I managed to write two tests that reproduce the issue, but this is unfortunately as far as I can get since I have no knowledge of C, PHP Extensions or the sorts. Hopefully this can be useful, even if just as a track history on why these features don't work.

I want to thank whoever wrote the Contributing.md file because that was truly amazing. We all learn to hate docker, but this project is a wonderful representation as to why we tolerate Docker.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/ext/sandbox/dd_trace_method_invocable.phpt
+++ b/tests/ext/sandbox/dd_trace_method_invocable.phpt
@@ -1,0 +1,26 @@
+--TEST--
+dd_trace_method() using Invocable Class
+--FILE--
+<?php
+use DDTrace\SpanData;
+
+class Foo
+{
+    public function bar() {}
+}
+
+class Instrumentation
+{
+    public function __invoke(SpanData $span)
+    {
+        echo 'This should be called';
+    }
+}
+
+dd_trace_method('Foo', 'bar', new Instrumentation);
+
+$foo = new Foo();
+$foo->bar();
+?>
+--EXPECTF--
+This should be called
--- a/tests/ext/sandbox/dd_trace_method_invocable_closure.phpt
+++ b/tests/ext/sandbox/dd_trace_method_invocable_closure.phpt
@@ -1,0 +1,31 @@
+--TEST--
+dd_trace_method() using Invocable Closure
+--FILE--
+<?php
+use DDTrace\SpanData;
+
+class Foo
+{
+    public function bar() {}
+}
+
+class Instrumentation
+{
+    public function __construct(private string $context) {}
+
+    public function __invoke(SpanData $span)
+    {
+        echo 'This works' . PHP_EOL;
+
+        echo 'This fails: ' . $this->context;
+    }
+}
+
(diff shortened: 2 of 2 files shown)


bwoebi:
Hey @deleugpn,

on which PHP versions do you get a crash?

By default the `trace_method()` function should just do nothing if invoked with anything but a Closure; running with `DD_TRACE_DEBUG=1`:
```
Unable to parse parameters for DDTrace\trace_method; expected (string $class_name, string $method_name, ?Closure $prehook = NULL, ?Closure $posthook = NULL)
```

deleugpn:
I ran this with php 8.0 buster as explained by the contributing guide.

> By default the trace_method() function should just do nothing if invoked with anything but a Closure

Maybe we can either support it or turn it into a fatal error? I have invested about 30 hours trying to figure out why it was just silently failing

bwoebi:
I totally can get the confusion, when you are used to PHP typically just yelling fatal error, but here it doesn't.

This has been a long-standing feature that the tracer will try as much as possible to stay silent as not to interfere with actual applications when something goes wrong inside the tracer. This may or may not be a mis-feature.

It's probably fine for production, but less so for a staging environment where you definitely _want_ to see these issues.
The question is, what the best way to distinguish these worlds from within the tracer is?

We also may want to introduce a bit more fine-grained log-levels here.

deleugpn:
> It's probably fine for production, but less so for a staging environment where you definitely want to see these issues.
The question is, what the best way to distinguish these worlds from within the tracer is?

I can understand the fear now that maybe there's code running in production that gives invocable classes to the tracer and just silently gets ignored, so changing that would break it for these people. But from a brand new implementation mindset it's quite a bad experience to send an invocable class in expecting it to work like a Closure but just getting silently ignored. This is extra hard when doing it inside AWS Lambda which is where we need observability the most and the feedback loop is slow/hard. 

I would actually recommend a `DD_STRICT=1` variable where we can opt-in to a less silent tracer. I actually don't mind production failing if I send an invocable class into the tracer because that would actually never had made into production if it had always failed in the first place.

> We also may want to introduce a bit more fine-grained log-levels here.

I'm in a bit of a tricky situation with this as well. `DEBUG` seems to be very quiet and `TRACE` seems to be extremely chatty. It also doesn't help that I'm doing this on AWS Lambda and _it feels like_ sometimes Lambda gets frozen and the agent never flushes and I never get traces on Datadog servers.

bwoebi:
Hey @deleugpn it's been a couple months, but with #1942, we've also made all these cases just emit a message in the php error log or stderr for now. So you should at least always see it.

We'll change it to always throw a TypeError in the future, but going the current way to minimize immediate impact on users.

PR #1794, opened 2022-11-10
PROF-6361 send allocation profiles

realFlowControl:
### Description

This PR adds allocation profiling to the PHP profiler:
- we aim to be a good™ neighbour, so we check on previously installed custom handlers
- and install our handlers in `startup` as we prepare for non-good™ neighbours

At this point in time there are some limitations you should be aware of:
- we are not prepared for chaotic evil neighbours 🙈 
- there is no sampling implemented at this time ⏳ (use at your own discretion)

### Readiness checklist
- [ ] guard in a 

### Reviewer checklist
- [ ] Milestone is set.


profiling/src/lib.rs
```diff
@@ -844,3 +923,57 @@ extern "C" fn execute_internal(
...
+    });
+
+    ptr
+}
+
+unsafe extern "C" fn alloc_profiling_free(ptr: *mut ::libc::c_void) {
+    if zend::executor_globals.current_execute_data.is_null() {
+        trace!("free({:p})", ptr);
+    }
+    zend::_zend_mm_free(zend::zend_mm_get_heap(), ptr);
```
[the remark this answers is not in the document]
realFlowControl: Good catch 👍

profiling/src/lib.rs
```diff
@@ -259,6 +268,56 @@ extern "C" fn minit(r#type: c_int, module_number: c_int) -> ZendResult {
         zend::zend_execute_internal = Some(execute_internal);
     };
 
+    #[cfg(feature = "allocation_profiling")]
+    {
+        let use_zend_alloc = std::env::var("USE_ZEND_ALLOC");
+        if !use_zend_alloc.is_ok() ||
+           use_zend_alloc.is_ok() && use_zend_alloc.ok().unwrap() == "1"
```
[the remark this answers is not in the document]
realFlowControl: Not anymore 😉 
Got a `zend_mm_heap corrupted`, but this was due to me not calling previous pointers in `free` and `realloc`. I removed the check for `USE_ZEND_ALLOC` and will try and run the PHP test suite against it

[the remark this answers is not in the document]
realFlowControl: Background can be found here: https://heap.space/xref/PHP-8.1/Zend/zend_alloc.c?r=c7558e2f#2811
When bypassing ZendMM using `USE_ZEND_ALLOC=0` the pointers to `__zend_malloc` are installed and we need to treat them as if a neighbouring extension would have installed custom handlers

profiling/src/lib.rs
```diff
@@ -259,6 +268,56 @@ extern "C" fn minit(r#type: c_int, module_number: c_int) -> ZendResult {
...
+            }
+            if custom_mm_free.is_none() ||
+                custom_mm_malloc.is_none() ||
+                custom_mm_realloc.is_none() ||
+                custom_mm_free != Some(alloc_profiling_free) ||
+                custom_mm_malloc != Some(alloc_profiling_malloc) ||
+                custom_mm_realloc != Some(alloc_profiling_realloc)
+            {
+                info!("Memory Allocation profiling could not be enabled. Pleas feel free to fill an issue stating the PHP version and installed modules.");
+            } else {
```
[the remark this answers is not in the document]
realFlowControl: The entire feature of a custom memory handler is bound to `ZEND_MM_CUSTOM` being defined:
https://heap.space/xref/PHP-8.1/Zend/zend_alloc.c?r=c7558e2f#2895
If this constant is initialised with `0` then a call to `zend_mm_set_custom_handlers` wont do anything. The only (IMHO) reliable way to find out if we actually succeeded is to check the pointers returned by `zend_mm_get_custom_handlers`. An alternative could be to call `is_zend_mm()` which returns `false` in case a custom memory handler is installed.

profiling/src/lib.rs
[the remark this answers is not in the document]
bwoebi: Yep, exactly - if use_zend_alloc=0 works, then we are also most likely compatible with other extensions.

profiling/src/lib.rs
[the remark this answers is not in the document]
bwoebi: Calling `is_zend_mm` sounds good to me. Also we should then actually note that in the info!() call, because this most likely means that they have a custom compiled build for extra performance, and just telling them ZEND_MM_CUSTOM being disabled is the most helpful thing to do.

profiling/src/lib.rs
```diff
@@ -844,3 +898,64 @@ extern "C" fn execute_internal(
...
+    trace!("malloc({}) -> {:p}", len, ptr);
+
+    REQUEST_LOCALS.with(|cell| {
+        let locals = cell.borrow();
+
+        if !locals.profiling_enabled {
+            return;
+        }
+
+        if let Some(profiler) = PROFILER.lock().unwrap().as_ref() {
```
morrisonlevi, line 927: I don't think this lock is strictly necessary. It was done to appease Rust's memory model in a safe way, and the cost was acceptable because we didn't lock it very often with time profiling. This PR changes that.

We don't have to fix it in this specific PR, though. I'm just pointing out that for production I suspect we'll need to revisit this.

profiling/src/profiling.rs
```diff
@@ -832,4 +832,96 @@ impl Profiler {
...
+                    },
+                    value: SampleData {
+                        frames,
+                        labels: labels.clone(),
+                        sample_values,
+                    },
+                };
+
+                // Panic: profiler was checked above for is_none().
+                match self.send_sample(message) {
```
morrisonlevi, line 910: Sorry, what's this panic referring to? I don't see an optional on the next line.

profiling/src/lib.rs
```diff
@@ -692,6 +733,41 @@ extern "C" fn mshutdown(r#type: c_int, module_number: c_int) -> ZendResult {
         profiler.stop();
     }
 
+    #[cfg(feature = "allocation_profiling")]
+    {
```
bwoebi, line 737: I think we should just skip the whole block `if zend::is_zend_mm()`. Don't even try (maybe another extension reset them, then that's also fine and we have nothing to do here), and don't emit two info!() warnings.

profiling/src/lib.rs
```diff
@@ -692,6 +733,41 @@ extern "C" fn mshutdown(r#type: c_int, module_number: c_int) -> ZendResult {
...
+                &mut custom_mm_malloc,
+                &mut custom_mm_free,
+                &mut custom_mm_realloc,
+            );
+        }
+        if custom_mm_free != Some(alloc_profiling_free) ||
+           custom_mm_malloc != Some(alloc_profiling_malloc) ||
+           custom_mm_realloc != Some(alloc_profiling_realloc)
+        {
+            info!("Memory allocation profiling could not be disabled. Please feel free to fill an issue stating the PHP version and installed modules. Most likely another extension registered ZendMM custom handlers but did not unregister properly or your PHP binary was compiled with `ZEND_MM_CUSTOM` being disabled.");
```
bwoebi, line 757: In this scenario - as noted in an older comment, we should skip unloading our extension by setting our extension handle to `NULL`.
Otherwise, it might be that some other extension forwarding calls to the allocator, like we do, has a pointer to our extension and will just hard crash then.

bwoebi (approved):
Looks good with my comments addressed. I'll leave it to Levi to decide on profiler specifics and when exactly this is ready for a merge.

profiling/Cargo.toml
```diff
@@ -27,6 +27,7 @@ uuid = { version = "1.0", features = ["v4"] }
 
 [features]
 profiling = []
```
morrisonlevi, line 29: Ha, we should remove this `profiling` feature. It's from a very old R&D idea that apparently made it in unnoticed ^_^

I mean, we don't use that feature anywhere, right?

realFlowControl: Just checked, that feature is not used anywhere

realFlowControl: ✔️

morrisonlevi (approved):

PR #1802, opened 2022-11-17
Fix $integration->setError() to also accept \Error instances

bwoebi:
### Description

We got a report that sometimes Symfony may emit \Error objects, which then emits another TypeError here.

I'm not sure exactly on how we get there, so, no test attached.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/Integration.php
+++ b/src/Integrations/Integrations/Integration.php
@@ -68,9 +68,9 @@ abstract class Integration
      * Sets common error tags for an exception.
      *
      * @param SpanData $span
-     * @param \Exception $exception
+     * @param \Throwable $exception
      */
-    public function setError(SpanData $span, \Exception $exception)
+    public function setError(SpanData $span, $exception)
     {
         $span->meta[Tag::ERROR_MSG] = $exception->getMessage();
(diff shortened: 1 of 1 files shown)


src/Integrations/Integrations/Integration.php
```diff
@@ -68,9 +68,9 @@ public function addTraceAnalyticsIfEnabledLegacy(Span $span)
      * Sets common error tags for an exception.
      *
      * @param SpanData $span
-     * @param \Exception $exception
+     * @param \Throwable $exception
      */
-    public function setError(SpanData $span, \Exception $exception)
+    public function setError(SpanData $span, $exception)
```
realFlowControl, line 73: Can we add the `\Throwable` type to the argument also?
```suggestion
    public function setError(SpanData $span, \Throwable $exception)
```

bwoebi: No we cannot, not as long as we're still releasing PHP 5 and PHP7+ code in the same release. Extension code is still independent, but PHP code is mostly still shared.

realFlowControl (approved):
Looks good to me 🎉

PR #1805, opened 2022-11-17
Add experimental WordPress 6 integration

morrisonlevi:
### Description

This is an experiment. Do not merge.

Even though it's a new integration, it's a breaking change by virtue of the WordPress 4/5 integration partially triggering on WordPress 6.

This traces actions, certain template functions, and blocks.

Screenshot of enhanced WordPress 6 integration

Here's what it looks like on master post 0.82.0 so you can compare a bit:

support-admin us1 prod dog_apm_traces_query=%40_trace_root%3A1%20env%3Adev cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http method%2Clog_http status_code historicalData=false messageDisplay=inline sort=desc spanID=1033712840

Currently, actions and templates don't have additional meta info, but blocks sometimes do (depends if it's available). Sometimes the blocks don't have a `blockName`, we probably need to investigate the WordPress source to see what's going on there. For now these unknown blocks have a resource type of `(blockName: ?)`.

The general pattern I've been using for resource names is basically an argument list using named arguments. Consider something like span name `block` with resource `$blockName`. If I adjust the resource to be `(blockName: $blockName)`, then when I look at the Span List tab in the UI, I have a lot better idea of what the resource belongs to. The same goes for span name `action` and resource `(hook_name: $hook_name)`. Note that the casing in these specific examples matches the casing WordPress uses, but we could normalize them if we felt it was better.

Screenshot showing the Span List

Of course, if this isn't a best practice we can change the details of these things. This is just me throwing things out quickly and iterating to improve my experience.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/Integrations/Integrations/WordPress/V6/WordPressComponent.php
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * @param string $service
+     */
+    public function __construct($service)
+    {
+        $this->service = $service;
+    }
+
+    public function __invoke(SpanData $span, ...$args)
```
morrisonlevi, line 26: Unused, first idea was to use the object as a callback but that doesn't work.

morrisonlevi:
Closing. The information learned here was used as part of #2126.

PR #1809, opened 2022-11-21
Fix misbehaviour with 256+ hooks on a same function

bwoebi:
### Description

Might also crash sometimes.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/span.c
+++ b/ext/span.c
@@ -184,12 +184,13 @@ void ddtrace_open_span(ddtrace_span_data *span) {
     ddtrace_set_global_span_properties(span);
 }
 
+// += 2 increment to avoid zval type ever being 0
 ddtrace_span_data *ddtrace_alloc_execute_data_span(zend_ulong index, zend_execute_data *execute_data) {
     zval *span_zv = zend_hash_index_find(&DDTRACE_G(traced_spans), index);
     ddtrace_span_data *span;
     if (span_zv) {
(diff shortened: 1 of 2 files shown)


tests/ext/sandbox-regression/handle_many_hooks.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+The tracer should not crash when many hooks are installed
```
pierotibou, line 2: Out of curiosity, did this test failed consistently before fixing?

ext/span.c
```diff
@@ -184,12 +184,13 @@ void ddtrace_open_span(ddtrace_span_data *span) {
     ddtrace_set_global_span_properties(span);
 }
 
+// += 2 increment to avoid zval type ever being 0
```
pierotibou, line 187: That looks like magic :)

tests/ext/sandbox-regression/handle_many_hooks.phpt
bwoebi: Yes.

ext/span.c
bwoebi: Yes, I don't like it very much, but it works.

morrisonlevi (approved):

PR #1814, opened 2022-11-22
Revert catch logic for PHP 7.0 only

bwoebi:
### Description

The previous logic was actually correct for 7.0, but not for 7.1 and 7.2. Not it should always be correct.

Fixes regression from d9244963484536972f14ae4540ed9ad6c221fe25.

We actually had a CI test failing for that one, but apparently, we hadn't noticed.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi (commented):
What about the 7.1 failure that occurred on this branch?

PR #1815, opened 2022-11-23
[PROF-6535] cleanup and make allocation profiling usable

realFlowControl:
### Description

This PR will:
- add sampling to allocation profiling
- add an ENV variable and INI setting to allow for enabling/disabling allocation profiling (`DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED` / `datadog.profiling.experimental_allocation_enabled`)
- move the registration/unregistration to `rinit`/`rshutdown`. This is due to ZendMM wont do garbage collection if during `php_request_shutdown` a custom handler is installed, which would lead to memory leaks
- overwrite `_zend_mm_heap.use_custom_heap` before every call to (re)alloc/free
- add allocation profiler to `minfo()`

### Readiness checklist
- [x] Make sure the `allocation_profiling` feature is not default anymore in `Cargo.toml`

### Reviewer checklist
- [x] Milestone is set.


profiling/src/lib.rs
```diff
@@ -950,6 +1072,18 @@ unsafe extern "C" fn alloc_profiling_malloc(len: u64) -> *mut ::libc::c_void {
         return ptr;
     }
 
+    let stats = ALLOCATION_PROFILING_STATS.with(|cell| {
+        let allocations = cell.borrow();
+        allocations.track_allocation(len)
+    });
+
+    if !stats.0 {
```
felixge, line 1080: What does this do? Access the first field? Why not access the field by name? This is difficult to read 🤔

profiling/src/lib.rs
```diff
@@ -336,6 +307,92 @@ pub struct RequestLocals {
...
+/// take a sample every X bytes
+const ALLOCATION_PROFILING_INTERVAL: f32 = 1024.0 * 30.0;
+
+pub struct AllocationProfilingStats {
+    /// number of bytes in "this" sampling interval
+    pub sampling_interval: AtomicU64,
+    /// number of bytes until next sample collection
+    pub remaing_bytes: AtomicI64,
+    /// number of allocations since last sample
+    pub allocation_count: AtomicU64,
```
felixge, line 319: A lot of the implementation seems to deal with atomics. FWIW Go maintains one `nextSample` (similar to what you call remaining_bytes) for each thread. This means there is no need to deal with atomics. Each threads ends up making its own independent sampling decisions. Maybe this makes sense for your PR as well?

profiling/src/lib.rs
```diff
@@ -950,6 +1072,18 @@ unsafe extern "C" fn alloc_profiling_malloc(len: u64) -> *mut ::libc::c_void {
...
+        let allocations = cell.borrow();
+        allocations.track_allocation(len)
+    });
+
+    if !stats.0 {
+        return ptr;
+    }
+
+    let samples = stats.1;
+    let total_size = stats.2;
```
felixge, line 1085: As far as I can tell, these values are still unscaled. Are you planning to add the code for scaling them up to provide an estimate of true allocations?

profiling/src/lib.rs
```diff
@@ -336,6 +307,92 @@ pub struct RequestLocals {
...
+        let mut samples: f64 = (remaing_bytes * -1 / sampling_interval as i64) as f64;
+        remaing_bytes = remaing_bytes % sampling_interval;
+
+        loop {
+            remaing_bytes += next_sampling_interval as i64;
+            samples += 1.0;
+            if remaing_bytes > 0 {
+                break;
+            }
+        }
```
felixge, line 378: The code above seems to be a little different from how the Go profiler does it. It's actually fairly important to get all the details right, otherwise it doesn't work correctly. I know this because I recently reimplemented the algorithm for a simulator. It's also difficult for me to review the correctness if you diverge too much from the Go implementation.

So I'd highly recommend you to translate this code verbatim. Otherwise you should add an extensive test suite for correctness.

https://github.com/felixge/alloc-prof-sim/blob/main/main.go#L189-L214

felixge (changes requested):
Nice work! Here is some feedback as requested.

profiling/src/lib.rs
realFlowControl: In a first version the `track_allocation()` function was inline, so the returned tuple and the usage were closer to each other, now with refactoring part of this, you are right. No one knows whats in the tuple anymore. I moved the rest of this (and also in the `realloc`-path to the `track_allocation()` function and the tuple is gone 🎉

profiling/src/lib.rs
realFlowControl: It is indeed different from the Go profiler. I had shamelessly stolen that part from the native profiler 😉 
https://github.com/DataDog/ddprof/blob/main/src/lib/allocation_tracker.cc#L164-L185

felixge: Oh. I was not aware that they came up with their own sampling variant. Looking at [their tests](https://github.com/DataDog/ddprof/blob/main/test/allocation_tracker-ut.cc#L28) I'm worried about correctness.

I'll try to add their algorithm to my simulator ([JIRA](https://datadoghq.atlassian.net/browse/PROF-6628)), but probably not until the next sprint.

For your PR I'd still recommend you to follow the Go implementation closely. It's the only one we got that has been heavily scrutinized for correctness AFAIK.

profiling/src/lib.rs
```diff
@@ -1101,7 +1121,17 @@ unsafe extern "C" fn alloc_profiling_realloc(
...
     let ptr: *mut libc::c_void;
     if PREV_CUSTOM_MM_REALLOC.is_none() {
-        ptr = zend::_zend_mm_realloc(zend::zend_mm_get_heap(), prev_ptr, len);
+        // the first element of the heap struct is the `use_custom_heap` flag. We got to reset this
+        // to 0 to allow for `zend_mm_gc()` to do garbage collection when called during allocation
+        // TODO: find out the correct size of that `use_custom_heap` element at this point I am
+        // assuming that it is a 32 bit integer ....
+        let heap = zend::zend_mm_get_heap();
+        let custom_heap: u32 = std::ptr::read(heap as *const u32);
+        std::ptr::write(heap as *mut u32, zend::ZEND_MM_CUSTOM_HEAP_NONE);
```
morrisonlevi, line 1087: This highlighted code seems to be repeated quite a few times. Can we extract a function for it?

realFlowControl: I refactored the preparation and restore into their own functions. Also added a nice comment to make clear what we are doing, why we are doing it, and why we may consider it okay (I don't want to use the word safe in this case 😉 )

profiling/src/profiling.rs
```diff
@@ -922,4 +808,90 @@ impl Profiler {
...
     }
+
+    unsafe fn prepare_sample_message(
+        &self,
+        frames: Vec<ZendFrame>,
+        sample: i64,
+        wall_time: i64,
+        cpu_time: Option<i64>,
+        alloc_samples: i64,
+        alloc_size: i64,
```
realFlowControl, line 819: @morrisonlevi I was thinking of creating a struct for this, that could also make initialisation easier (default should always be `0`). Something like this
```rust
#[derive(Default)]
struct SampleValues {
    interrupt_count: i64,
    wall_time: i64,
    cpu_time: i64,
    alloc_samples: i64,
    alloc_size: i64
}

// usage:
let samples = SampleValues {
    alloc_samples: 1,
    alloc_size: 1024,
    ..Default::default()
};
```

WDYT?

profiling/src/lib.rs
```diff
@@ -336,6 +306,68 @@ pub struct RequestLocals {
...
+            let locals = cell.try_borrow();
+            if locals.is_err() {
+                return;
+            }
+            let locals = locals.unwrap();
+
+            if !locals.profiling_experimental_allocations_enabled {
+                return;
+            }
+
```
realFlowControl, line 355: There might be no need for this check, as we register in `rinit()` (or not) and therefore there should not be a call to this function when allocation profiling is disabled

```suggestion
```

profiling/src/lib.rs
```diff
@@ -339,6 +312,67 @@ pub struct RequestLocals {
     pub vm_interrupt_addr: *const AtomicBool,
 }
 
+/// take a sample every X bytes
+#[cfg(feature = "allocation_profiling")]
+const ALLOCATION_PROFILING_INTERVAL: f32 = 1024.0 * 50.0;
```
felixge, line 317: NIT: Maybe add a comment how this value was picked? This causes 10x more frequent sampling compared to Go. Unless PHP programs have a 10x small allocation rate, this would cause 10x more overhead.

It's also fine to just add a comment that this value is temporary and will be replaced with a better one later.

Last but not least: Have you considered making this value user configurable? That's what Go does, see [runtime.MemProfileRate](https://pkg.go.dev/runtime#MemProfileRate).

profiling/src/lib.rs
```diff
@@ -339,6 +312,67 @@ pub struct RequestLocals {
...
+        if self.next_sample > 0 {
+            return;
+        }
+
+        let scale = 1.0 / (1.0 - (len as f64 * -1.0 / ALLOCATION_PROFILING_INTERVAL as f64).exp());
+
+        let count = 1.0 * scale;
+        let bytes = len as f64 * scale;
+
+        self.next_sample = AllocationProfilingStats::next_sampling_interval();
```
felixge, line 351: Sampling logic LGTM. This is exactly what the Go profiler does and should work just as well 👍

profiling/src/profiling.rs
```diff
@@ -727,102 +727,42 @@ impl Profiler {
...
                 match self.send_sample(message) {
-                    Ok(_) => trace!(
-                        "Sent stack sample of depth {} with labels {:?} to profiler.",
-                        depth,
-                        labels
-                    ),
+                    Ok(_) => trace!("Sent stack sample of depth {} with to profiler.", depth,),
                     Err(err) => warn!(
-                        "Failed to send stack sample of depth {} with labels {:?} to profiler: {}",
-                        depth, labels, err
```
felixge, line 825: NIT: These changes look unrelated to allocation profiling on first glance. Would probably be easier to split this PR into smaller ones if it contains multiple orthogonal changes.

felixge (approved):
The allocation sampling algorithm LGTM. I added a few more nit comments, but they are not critical.

⚠️ I don't see any correctness tests for the allocation sampling. I think it will work, but it'd be nice to put something in place that would catch obvious regressions in the future.

profiling/src/lib.rs
realFlowControl: Added a comment and created a new JIRA ticket for making the value configurable, thanks for the input 🙇

profiling/src/profiling.rs
```diff
@@ -848,84 +793,40 @@ impl Profiler {
...
                     ),
                     Err(err) => warn!(
-                        "Failed to send stack sample of depth {} with labels {:?} to profiler: {}",
-                        depth, labels, err
+                        "Failed to send stack sample of depth {} with size {}, labels {:?} and count {} to profiler: {}",
+                        depth,
+                        bytes,
+                        labels,
+                        count,
+                        err
```
morrisonlevi, line 829: There are a few places where we can do things like this:

```suggestion
                        "Failed to send stack sample of depth {depth} with size {bytes}, labels {labels:?} and count {count} to profiler: {err}"
```

Landed in Rust v1.58.

realFlowControl: Thats nice, especially if there are more arguments. I updated also others that I could find

profiling/src/profiling.rs
```diff
@@ -739,199 +739,179 @@ impl Profiler {
...
+
+    fn prepare_sample_message(
+        &self,
+        frames: Vec<ZendFrame>,
+        sample: i64,
+        wall_time: i64,
+        cpu_time: Option<i64>,
+        alloc_samples: i64,
+        alloc_size: i64,
+        labels: &Vec<Label>,
```
morrisonlevi, line 855: Since this is read-only, clippy suggests `&[Label]` instead:

```
warning: writing `&Vec` instead of `&[_]` involves a new object where a slice will do
   --> src/profiling.rs:855:17
    |
855 |         labels: &Vec<Label>,
    |                 ^^^^^^^^^^^
    |
    = note: `#[warn(clippy::ptr_arg)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
help: change this to
    |
855 ~         labels: &[Label],
856 |         locals: &RequestLocals,
857 |     ) -> SampleMessage {
858 |         let mut sample_types = vec![
859 |             ValueType {
860 |                 r#type: Cow::Borrowed("sample"),
```

profiling/src/profiling.rs
morrisonlevi: If we choose to not do it now, let's do this or something like it soon after merge. Clippy doesn't like how many arguments this function has :)

profiling/src/lib.rs
```diff
@@ -934,66 +1030,112 @@ extern "C" fn execute_internal(
...
+        let mut allocations = cell.borrow_mut();
+        allocations.track_allocation(len)
     });
 
     ptr
 }
 
+// This function exists just for completeness. When calling `zend_mm_set_custom_handlers()` you
+// need to pass a pointer to a `free()` function.
+#[cfg(feature = "allocation_profiling")]
```
morrisonlevi, line 1085: I didn't understand your comment here about completeness. Can you explain some more?

profiling/src/lib.rs
```diff
@@ -548,13 +592,50 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
...
             }
         });
     }
+
+    #[cfg(feature = "allocation_profiling")]
+    {
+        if profiling_experimental_allocation_enabled {
+            if !is_zend_mm() {
+                // Neighboring custom memory handlers found
+                trace!("Found another extension using the ZendMM custom handler hook");
```
morrisonlevi, line 607: I'd change this to `debug`; I'd want that info if I was debugging something.

profiling/src/lib.rs
```diff
@@ -548,13 +592,50 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
...
+                zend::ddog_php_prof_zend_mm_set_custom_handlers(
+                    zend::zend_mm_get_heap(),
+                    Some(alloc_profiling_malloc),
+                    Some(alloc_profiling_free),
+                    Some(alloc_profiling_realloc),
+                );
+            }
+
+            if is_zend_mm() {
+                info!("Memory allocation profiling could not be enabled. Please feel free to fill an issue stating the PHP version and installed modules. Most likely the reason is your PHP binary was compiled with `ZEND_MM_CUSTOM` being disabled.");
```
morrisonlevi, line 628: I'd change this to `warn`, I think? This only triggers if allocation profiling is enabled and it default to off, so when that combines, I think it's more serious than `info`.

profiling/src/lib.rs
realFlowControl: I adjusted the comment in a new commit.
The thing is: `zend_mm_set_custom_handlers()` expects three function pointers. If any of these is `null` it would unregister custom handlers. So although we are - currently - not interested in tracking `free()` calls, we need to provide a function pointer.
As there will be a custom handler installed, we need to overwrite the `use_custom_heap` flag before calling the ZendMM default free function.
Hence the only reason this function exists is because we are using the `zend_mm_set_custom_handlers()` hook.

profiling/src/lib.rs
```diff
@@ -934,66 +1024,115 @@ extern "C" fn execute_internal(
...
+// before calling the original `zend::_zend_mm_free()` function as we do in the custom alloc and
+// realloc handlers.
+#[cfg(feature = "allocation_profiling")]
 unsafe extern "C" fn alloc_profiling_free(ptr: *mut ::libc::c_void) {
     if PREV_CUSTOM_MM_FREE.is_none() {
-        zend::_zend_mm_free(zend::zend_mm_get_heap(), ptr);
+        let heap = zend::zend_mm_get_heap();
+        let custom_heap = prepare_zend_heap(heap);
+        zend::_zend_mm_free(heap, ptr);
+        restore_zend_heap(heap, custom_heap);
```
bwoebi, line 1088: For efree() that's not really necessary as the flag is never checked within.

morrisonlevi (approved):
Looks good! Let's move to the next stage!

PR #1819, opened 2022-11-28
fix compile time warning when compiling with features

realFlowControl:
### Description

This PR fixes compile time warnings when compiling without the feature `allocation_profiling`

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/profiling/src/bindings/mod.rs
+++ b/profiling/src/bindings/mod.rs
@@ -10,8 +10,13 @@ use std::sync::atomic::AtomicBool;
 
 pub type VmInterruptFn = unsafe extern "C" fn(execute_data: *mut zend_execute_data);
 
+#[cfg(feature = "allocation_profiling")]
 pub type VmMmCustomAllocFn = unsafe extern "C" fn(u64) -> *mut libc::c_void;
+
+#[cfg(feature = "allocation_profiling")]
 pub type VmMmCustomReallocFn = unsafe extern "C" fn(*mut libc::c_void, u64) -> *mut libc::c_void;
+
+#[cfg(feature = "allocation_profiling")]
 pub type VmMmCustomFreeFn = unsafe extern "C" fn(*mut libc::c_void);
 
 // todo: this a lie on some PHP versions; is it a problem even though zend_bool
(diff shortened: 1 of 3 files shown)


profiling/src/bindings/mod.rs
```diff
@@ -10,8 +10,13 @@ use std::sync::atomic::AtomicBool;
...
 pub type VmInterruptFn = unsafe extern "C" fn(execute_data: *mut zend_execute_data);
 
+#[cfg(feature = "allocation_profiling")]
 pub type VmMmCustomAllocFn = unsafe extern "C" fn(u64) -> *mut libc::c_void;
+
+#[cfg(feature = "allocation_profiling")]
 pub type VmMmCustomReallocFn = unsafe extern "C" fn(*mut libc::c_void, u64) -> *mut libc::c_void;
+
+#[cfg(feature = "allocation_profiling")]
 pub type VmMmCustomFreeFn = unsafe extern "C" fn(*mut libc::c_void);
```
morrisonlevi, line 20: IIRC, you can do this:

```suggestion
#[cfg(feature = "allocation_profiling")]
{
    pub type VmMmCustomAllocFn = unsafe extern "C" fn(u64) -> *mut libc::c_void;
    pub type VmMmCustomReallocFn = unsafe extern "C" fn(*mut libc::c_void, u64) -> *mut libc::c_void;
    pub type VmMmCustomFreeFn = unsafe extern "C" fn(*mut libc::c_void);
}
```

Do you think that's better?

realFlowControl: Tried it and I'd love to, this would nicely group together what belongs together. Same in [profiling/src/lib.rs](https://github.com/DataDog/dd-trace-php/blob/955d8e10ecb7d2af26863f93c2cf26ae5e0ba5c3/profiling/src/lib.rs#L133-L143), but Rust won't allow it though:
```
error: expected item after attributes
  --> src/bindings/mod.rs:13:1
   |
13 | #[cfg(feature = "allocation_profiling")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Same when trying to make this an inner attribute of a block, Rust expects an item before the opening curly braces (at least at the root level of a file)

morrisonlevi (approved):

PR #1821, opened 2022-11-29
chore(tracing): change error.msg to error.message

wconti27:
### Description
Changes `error.msg` to `error.message` to match new Datadog Span Attributes spec. Also changes `ERROR_MSG` constant to `ERROR_MESSAGE` for consistency. Part of a larger effort to make tag naming more consistent among Datadog tracing languages

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


src/api/Tag.php
```diff
@@ -15,7 +15,7 @@ class Tag
     const RESOURCE_NAME = 'resource.name';
     const DB_STATEMENT = 'sql.query';
     const ERROR = 'error';
-    const ERROR_MSG = 'error.msg'; // string representing the error message
+    const ERROR_MESSAGE = 'error.message'; // string representing the error message
```
morrisonlevi, line 18: I think I would leave `ERROR_MSG` that points to the updated `error.message` name, with a comment about it being deprecated and to use `ERROR_MESSAGE` instead. This would prevent breakage for custom integrations users have written.

```suggestion
    /** @deprecated Use ERROR_MESSAGE instead. */
    const ERROR_MSG = 'error.message';
    const ERROR_MESSAGE = 'error.message'; // string representing the error message
```

src/api/Tag.php
```diff
@@ -15,7 +15,8 @@ class Tag
...
+    const ERROR_MSG = 'error.msg';/** @deprecated Use ERROR_MESSAGE instead. */
```
morrisonlevi, line 19: Documentation comments need to be placed _before_ the thing they apply to. I know it's not consistent with the comment style being used right here, but that's how it goes.

As written, this `@deprecated` would apply to `ERROR_TYPE`.

wconti27: Good to know! Just fixed this

src/api/Tag.php
```diff
@@ -15,7 +15,9 @@ class Tag
     const RESOURCE_NAME = 'resource.name';
     const DB_STATEMENT = 'sql.query';
     const ERROR = 'error';
-    const ERROR_MSG = 'error.msg'; // string representing the error message
+    const ERROR_MESSAGE = 'error.message'; // string representing the error message
+    /** @deprecated Use ERROR_MESSAGE instead. */
+    const ERROR_MSG = 'error.msg';
```
morrisonlevi, line 20: We can migrate the value here, I think?

```suggestion
    const ERROR_MSG = 'error.message';
```

morrisonlevi (approved):

pierotibou (approved):

PR #1826, opened 2022-12-05
Change process_id to metrics tag

wconti27:
### Description
Change process id to be a metrics field instead of meta field.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (commented):
In some tests like `tests/ext/active_span.phpt`, the `process_id` field is an `int`, and in others like `tests/ext/distributed_tracing/distributed_trace_bogus_ids.phpt`, it is a `float`. Which should metrics be? I assume float?

ext/serializer.c
```diff
@@ -1060,9 +1060,12 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         zend_string *str_key;
         zval *val;
         ZEND_HASH_FOREACH_STR_KEY_VAL_IND(Z_ARR_P(metrics), str_key, val) {
-            if (str_key) {
+            if (str_key && strcmp(ZSTR_VAL(str_key), "process_id") != 0) {
```
morrisonlevi, line 1063: Instead of `strcmp(a, b)`, I think you need to use `zend_string_equals_literal`:

```suggestion
            if (str_key && zend_string_equals_literal(str_key, "process_id")) {
```

Same idea below as well.

Though, I wonder where the meta key is coming from, and maybe we should add its value to the `metrics` thing automatically? Meaning, if someone sets `$span->meta["process_id"] = getpid();` or whatever, that we move it at serialization time to metrics? Dunno.

tests/Common/SpanChecker.php
```diff
@@ -416,18 +418,29 @@ function ($key) use ($pattern) {
...
             }
         }
         $metrics = $exp->getExactMetrics();
         if ($metrics !== SpanAssertion::NOT_TESTED) {
             // Ignore compilation-time metric unless explicitly tested
             if (!isset($metrics['php.compilation.total_time_ms'])) {
-                unset($span['metrics']['php.compilation.total_time_ms']);
+                unset($spanMetrics['php.compilation.total_time_ms']);
+            }
+            if (isset($metrics['process_id'])) {
```
morrisonlevi, line 435: Are these supposed to be negated with `!` as well? Meaning, unless they are specifically tested, we shouldn't test them?

wconti27: The purpose of these lines is to unset the process_id field in both our expected and actual metric arrays. Doing this because we already asserted on the existence of the process_id field on lines 421-425. If we don't unset the field, we will have a failure on line 441 when comparing exact arrays equal since the process_id values will not be equal.

ext/serializer.c
wconti27: Changed the string comparison for use with a zend_string. For the meta array and possibly setting PID, do we see users ever setting that manually? Especially given that we are setting the PID within metrics no matter what. Also, the span meta is serialized during this call: `_serialize_meta(el, span);` on line 1053.

bwoebi (approved):

PR #1835, opened 2022-12-16
[Tracer][CodeIgniter] Fix CodeIgniter integration to not depend on CI_Controller class

pierotibou:
### Description

In a [recent escalation](https://datadoghq.atlassian.net/browse/APMS-8370), the app throws a 500 error when using ddtrace as we expect any controller to extend CI_Controller class, which isn't done there for performance reasons.

In the integration, I've just checked the existence of the field we're calling. The extra tests now works. It's hard to tell if the change could have side effects are the other CodeIgniter tests are flaky locally, even without my change.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
+++ b/src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
@@ -70,7 +70,11 @@ class CodeIgniterIntegration extends Integration
                 $span->service = $service;
                 $span->type = Type::WEB_SERVLET;
 
-                $this->load->helper('url');
+                // We took the assumption that all controllers will extend CI_Controller.
+                // But we've at least seen one healthcheck controller not extending it.
+                if ($this->load) {
+                    $this->load->helper('url');
+                }
 
                 if (!array_key_exists(Tag::HTTP_URL, $rootSpan->meta)) {
                     $rootSpan->meta[Tag::HTTP_URL] = \DDTrace\Util\Normalizer::urlSanitize(base_url(uri_string()))
@@ -98,7 +102,12 @@ class CodeIgniterIntegration extends Integration
                 $span->service = $service;
                 $span->type = Type::WEB_SERVLET;
 
-                $this->load->helper('url');
+                // We took the assumption that all controllers will extend CI_Controller.
+                // But we've at least seen one healthcheck case where it wasn't the case.
+                if ($this->load) {
+                    $this->load->helper('url');
+                }
+
                 $rootSpan->meta[Tag::HTTP_URL] = \DDTrace\Util\Normalizer::urlSanitize(base_url(uri_string()))
                     . Normalizer::sanitizedQueryString();
                 $rootSpan->meta['app.endpoint'] = "{$class}::_remap";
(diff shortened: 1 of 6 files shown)


src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
```diff
@@ -70,7 +70,11 @@ function (SpanData $span) use ($rootSpan, $method, $service) {
...
 
-                $this->load->helper('url');
+                // We took the assumption that all controllers will extend CI_Controller.
+                // But we've at least seen one healthcheck controller not extending it.
+                if ($this->load) {
+                    $this->load->helper('url');
+                }
 
                 if (!array_key_exists(Tag::HTTP_URL, $rootSpan->meta)) {
                     $rootSpan->meta[Tag::HTTP_URL] = \DDTrace\Util\Normalizer::urlSanitize(base_url(uri_string()))
```
realFlowControl, line 80: I would expect this to fail at the call to `base_url()` in case line 76 won't get executed as the `base_url()` function seems to be provided by the `url` helper: https://codeigniter.com/userguide3/helpers/url_helper.html#base_url

src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
```diff
@@ -70,7 +70,11 @@ function (SpanData $span) use ($rootSpan, $method, $service) {
                 $span->service = $service;
                 $span->type = Type::WEB_SERVLET;
 
-                $this->load->helper('url');
+                // We took the assumption that all controllers will extend CI_Controller.
+                // But we've at least seen one healthcheck controller not extending it.
+                if ($this->load) {
```
morrisonlevi, line 75: Is this trying to guard against the presence of the `load` method? If so, it should be:

```suggestion
                if (\method_exists($this, 'load')) {
```

src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
```diff
@@ -98,7 +102,12 @@ function (SpanData $span, $args, $retval, $ex) use ($rootSpan, $service) {
                 $span->service = $service;
                 $span->type = Type::WEB_SERVLET;
 
-                $this->load->helper('url');
+                // We took the assumption that all controllers will extend CI_Controller.
+                // But we've at least seen one healthcheck case where it wasn't the case.
+                if ($this->load) {
+                    $this->load->helper('url');
+                }
```
morrisonlevi, line 109: Same as the other comment.

src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
pierotibou: I'll retry this, it was my first try and didn't work. But just in case I'll retry

pierotibou: Actually, load don't seem to be a method right? it's initialized with `load_class`. I've added a check on the helper method while I was at it though inc75e1a80ea427923d66779fffa11e16bdf825e6b

src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
pierotibou: Thanks a lot. I've regrouped the whole lot in the if in c75e1a80ea427923d66779fffa11e16bdf825e6b.

src/Integrations/Integrations/CodeIgniter/V2/CodeIgniterIntegration.php
morrisonlevi: Doh! Your commit makes sense.

morrisonlevi (approved):

realFlowControl (approved):

PR #1839, opened 2022-12-22
Show error message when PDO returns an error

pierotibou:
### Description

Adds the error message in the error message tag when available. This gives more information on where the issue is. Error messages are contained inside the [errorInfo array](https://www.php.net/manual/en/pdo.errorinfo.php). The specs state it contains 3 elements but [SQLServer seems to set more than that.](https://learn.microsoft.com/en-us/sql/connect/php/pdo-errorinfo?view=sql-server-ver16#additional-odbc-messages)

This extra data shouldn't contain sensitive data as it should be close to what we get on PDOException when there is one. 

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/PDO/PDOIntegration.php
+++ b/src/Integrations/Integrations/PDO/PDOIntegration.php
@@ -136,13 +136,15 @@ class PDOIntegration extends Integration
 
         // Driver-specific error message will be in the rest of the array
         // The spec suggests the error message to be at pos 2 only,
-        // but e.g. SQLServer doc states that the array can contain monre than 3 elements
+        // for ODBC: the SQLSTATE in [3] (but e.g. SQLServer doc states that the array can contain monre than 3 elements)
+        // for DBLIB: in [3] an OSerror (integer), in [4] a severity and optionally a string representation of the OSerror in [5].
         if (count($errorInfo) > 2) {
-            $span->meta[Tag::ERROR_MSG] .= '. Driver-specific error message: ' ;
-
+            $additionalInfo = [];
             for ($i = 2; $i < count($errorInfo); $i++) {
-                $span->meta[Tag::ERROR_MSG] .= $errorInfo[$i];
+                $additionalInfo[] = $errorInfo[$i];
             }
+
+            $span->meta[Tag::ERROR_MSG] .= '. Driver-specific error data: ' . implode('. ', $additionalInfo);
         }
 
         $span->meta[Tag::ERROR_TYPE] = get_class($pdoOrStatement) . ' error';
--- a/src/Integrations/Integrations/PDO/PDOIntegration.php
+++ b/src/Integrations/Integrations/PDO/PDOIntegration.php
@@ -136,15 +136,10 @@ class PDOIntegration extends Integration
 
         // Driver-specific error message will be in the rest of the array
         // The spec suggests the error message to be at pos 2 only,
-        // for ODBC: the SQLSTATE in [3] (but e.g. SQLServer doc states that the array can contain monre than 3 elements)
-        // for DBLIB: in [3] an OSerror (integer), in [4] a severity and optionally a string representation of the OSerror in [5].
+        // ODBC: the SQLSTATE in [3] (but e.g. SQLServer doc states that the array can contain monre than 3 elements)
+        // DBLIB: [3] an OSerror (int), [4] a severity and optionally a string representation of the OSerror in [5]
         if (count($errorInfo) > 2) {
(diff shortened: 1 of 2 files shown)


src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -133,6 +133,18 @@ public static function detectError($pdoOrStatement, SpanData $span)
...
+
+        // Driver-specific error message will be in the rest of the array
+        // The spec suggests the error message to be at pos 2 only,
+        // but e.g. SQLServer doc states that the array can contain monre than 3 elements
+        if (count($errorInfo) > 2) {
+            $span->meta[Tag::ERROR_MSG] .= '. Driver-specific error message: ' ;
+
+            for ($i = 2; $i < count($errorInfo); $i++) {
+                $span->meta[Tag::ERROR_MSG] .= $errorInfo[$i];
+            }
```
labbati, line 145: I would break at the first non-string argument and use the `implode` function to put separators only when appropriate.

```suggestion
            $additionalInfo = [];
            for ($i = 2; $i < count($errorInfo); $i++) {
                if (! is_string($errorInfo[$i])) {
                    break;
                }
                $additionalInfo[] = $errorInfo[$i];
            }
            $span->meta[Tag::ERROR_MSG] .= implode('. ', $additionalInfo);
```

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -133,6 +133,18 @@ public static function detectError($pdoOrStatement, SpanData $span)
         }
         $errorInfo = $pdoOrStatement->errorInfo();
         $span->meta[Tag::ERROR_MSG] = 'SQL error: ' . $errorCode . '. Driver error: ' . $errorInfo[1];
+
+        // Driver-specific error message will be in the rest of the array
+        // The spec suggests the error message to be at pos 2 only,
+        // but e.g. SQLServer doc states that the array can contain monre than 3 elements
+        if (count($errorInfo) > 2) {
+            $span->meta[Tag::ERROR_MSG] .= '. Driver-specific error message: ' ;
```
labbati, line 141: This is not a blocker for me, but I am unsure `Tag::ERROR_MSG` is where we want to add this information, which to me seems to be pretty unpredictable. Should we add a tag like `error|err.details`? Again, very weakly opinionated on whether we should do it or not

labbati (changes requested):

bwoebi:
The extended error_info is:
- for ODBC: the [SQLSTATE](https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/sqlstates?view=sql-server-ver16) in `[3]`.
- for DBLIB: in `[3]` a possible OSerror (integer), in `[4]` a severity and optionally a string representation of the OSerror in `[5]`.

Merely chaining these info together is not going to look nicely.

src/Integrations/Integrations/PDO/PDOIntegration.php
pierotibou: When we redefined the span attributes, we didn't include a specific one for such a need. 
The error message in some languages will contain the exception which will contain the same kind of data, so I don't think we should add a new one

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -133,6 +133,20 @@ public static function detectError($pdoOrStatement, SpanData $span)
...
+        // The spec suggests the error message to be at pos 2 only,
+        // for ODBC: the SQLSTATE in [3] (but e.g. SQLServer doc states that the array can contain monre than 3 elements)
+        // for DBLIB: in [3] an OSerror (integer), in [4] a severity and optionally a string representation of the OSerror in [5].
+        if (count($errorInfo) > 2) {
+            $additionalInfo = [];
+            for ($i = 2; $i < count($errorInfo); $i++) {
+                $additionalInfo[] = $errorInfo[$i];
+            }
+
+            $span->meta[Tag::ERROR_MSG] .= '. Driver-specific error data: ' . implode('. ', $additionalInfo);
```
bwoebi, line 147:
```suggestion
            $span->meta[Tag::ERROR_MSG] .= '. Driver-specific error data: ' . implode('. ', array_slice($additionalInfo, 2));
```

pierotibou: Introduced in last commit (didn't use the exact suggestion as we need to slice errorInfo)

bwoebi (approved):

PR #1846, opened 2023-01-10
Support custom ini names in installer

### LINKED ISSUE #1833: Ini file created by datadog-setup.phpdoes not align with rpm package
Hey guys,

we encountered several "PHP Warning: Module 'ddtrace' already loaded in Unknown on line 0" warnings after we used your installer over the last couple of weeks. We had DD as well as the profiler installed before we used your installer for the first time.

The reason was a bit tough to find but logical at the end :-)

https://github.com/DataDog/dd-trace-php/blob/7cdb199db6513d10eaaceb1ca3c5ae33465dae8c/datadog-setup.php#L227

We install all DD packages using the remi repo rpm packages on Centos ( let's take a jump back to the past: https://github.com/DataDog/dd-trace-php/issues/214 ).
Remi adds the ini files with the prefix `40-ddtrace.ini`. Thus all updates to the rpm packages leads to a duplicate ini file being present.
Your installer creates `98-ddtrace.ini`.

We now simply moved your ini file to `40-ddtrace.ini`. But still running the installer again or setting up a fresh system will always lead to this ini clash.

Maybe you could add a check to your installer to see if a `xx-ddtrace.ini` file is already present in `/php.d/` and offer to overwrite these instead of creating a new file?

Best regards,

Jan

bwoebi:
### Description

Fixes #1833.

Also update randomized tests instructions and slightly lower the recursion depth for internal-api-stress-test.php

This PR also updates the installer tests so that they run natively on arm dev machines locally.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -221,6 +221,17 @@ function install($options)
         // Writing the ini file
         if ($phpProperties[INI_SCANDIR]) {
             $iniFileName = '98-ddtrace.ini';
+            // Search for pre-existing files with extension = ddtrace.so to avoid conflicts
+            // See issue https://github.com/DataDog/dd-trace-php/issues/1833
+            foreach (scandir($phpProperties[INI_SCANDIR]) as $ini) {
+                $path = "{$phpProperties[INI_SCANDIR]}/$ini";
+                if (is_file($path)) {
+                    if (preg_match('(^\s*extension\s*=.+ddtrace\.so)m', file_get_contents($path))) {
+                        $iniFileName = $ini;
+                    }
+                }
+            }
+
             $iniFilePaths = [$phpProperties[INI_SCANDIR] . '/' . $iniFileName];
 
             if (\strpos($phpProperties[INI_SCANDIR], '/cli/conf.d') !== false) {
@@ -270,7 +281,7 @@ function install($options)
                  */
                 execute_or_exit(
                     'Impossible to update the INI settings file.',
-                    "sed -i 's@extension \?= \?.*ddtrace.*\(.*\)@extension = ddtrace.so@g' "
+                    "sed -i 's@ \?;\? \?extension \?= \?.*ddtrace.*\(.*\)@extension = ddtrace.so@g' "
                         . escapeshellarg($iniFilePath)
                 );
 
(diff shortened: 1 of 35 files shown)


datadog-setup.php
```diff
@@ -308,12 +319,6 @@ function install($options)
 
             // phpcs:disable Generic.Files.LineLength.TooLong
             if ($shouldInstallAppsec) {
-                // Appsec crashes with missing symbols if tracing is not loaded
-                execute_or_exit(
-                    'Impossible to update the INI settings file.',
-                    "sed -i 's@ \?; \?extension \?= \?ddtrace.so@extension = ddtrace.so@g' "
-                        . escapeshellarg($iniFilePath)
-                );
```
realFlowControl, line 316: It looks like this can not be the case (missing `ddtrace.so` in `$iniFilePath`), as running `datadog-setup.php` will always install `ddtrace` extension and if it fails to do so, it will not reach this point, correct?

bwoebi: Correct. Just previously, for some reason it did not uncomment the line if present, which was not expected and making the test fail under arm where appsec is not loaded :-)

realFlowControl (approved):

datadog-setup.php
```diff
@@ -221,6 +221,17 @@ function install($options)
         // Writing the ini file
         if ($phpProperties[INI_SCANDIR]) {
             $iniFileName = '98-ddtrace.ini';
+            // Search for pre-existing files with extension = ddtrace.so to avoid conflicts
+            // See issue https://github.com/DataDog/dd-trace-php/issues/1833
+            foreach (scandir($phpProperties[INI_SCANDIR]) as $ini) {
+                $path = "{$phpProperties[INI_SCANDIR]}/$ini";
+                if (is_file($path)) {
+                    if (preg_match('(^\s*extension\s*=.+ddtrace\.so)m', file_get_contents($path))) {
```
remicollet, line 229: .so suffix is not mandatory since 7.2, so pref may fail (and is different in other OS)

remicollet: Notice `=.+` (mandatory char) is also wrong, can be "extension=ddtrace" (no space), so need "=.*"  to swallow optional path
`preg_match("(^\s*extension\s*=.*ddtrace(\b|\.so))m"` seems OK.

remicollet: Perhaps, even:

```
if (preg_match("(^\s*extension\s*=\s*(\S*ddtrace)\b)m", file_get_contents($path), $res)
    && basename($res[1])=="ddtrace")  {
```

PR #1852, opened 2023-01-12
Avoid putting .package-cache into the cache, commit Cargo.lock

morrisonlevi:
### Description

This `.package-cache` file will cause cargo to think some other process is running and holding the lock. I don't know how it got there, but this should help prevent it from getting in there.

Additionally, commit `Cargo.lock` so that build spend less time trying to resolve packages, which sometimes for unknown reasons will take longer than 10 minutes and produce no output, and fail the CI.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bwoebi (approved):
Weird that Cargo.lock needs to be there, but if it fixes it, then we'll have to.
Thanks for investigating and fixing, Levi!

PR #1865, opened 2023-01-17
fix(profiling): prepare heap for `gc_mem_caches()`

realFlowControl:
### Description

The PHP userland function `gc_mem_caches()` just forwards calls to `zend_mm_gc()` which will do exactly nothing and return `int(0)` in case a custom allocator is installed.
This PR makes sure we intercept userland calls to `gc_mem_caches()` and prepare the heap for such a call.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -1066,13 +1066,6 @@ unsafe extern "C" fn datadog_allocation_profiling_gc_mem_caches(
     execute_data: *mut zend::zend_execute_data,
     return_value: *mut zend::zval,
 ) {
-    // Better safe than sorry: `gc_mem_caches()` is a Zend core function which
-    // always exists
-    if GC_MEM_CACHES_HANDLER == None {
-        ddog_php_prof_copy_long_into_zval(return_value, 0);
-        return;
-    }
-
     let allocation_profiling: bool = REQUEST_LOCALS.with(|cell| {
         // Panic: there might already be a mutable reference to `REQUEST_LOCALS`
         let locals = cell.try_borrow();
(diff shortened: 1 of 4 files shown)


profiling/src/php_ffi.c
```diff
@@ -75,6 +75,12 @@ void datadog_php_profiling_copy_string_view_into_zval(zval *dest, zai_string_vie
     }
 }
 
+void datadog_php_profiling_copy_long_into_zval(zval *dest, long num) {
```
morrisonlevi, line 78: Can we use the newer `ddog_php_prof_` prefix instead of `datadog_php_profiling_`?

profiling/src/lib.rs
```diff
@@ -1045,6 +1061,43 @@ unsafe fn restore_zend_heap(heap: *mut zend::_zend_mm_heap, custom_heap: c_int)
...
 }
 
+#[cfg(feature = "allocation_profiling")]
+unsafe extern "C" fn datadog_allocation_profiling_gc_mem_caches(
+    execute_data: *mut zend::zend_execute_data,
+    return_value: *mut zend::zval,
+) {
+    // Better safe than sorry: `gc_mem_caches()` is a Zend core function which
+    // always exists
+    if GC_MEM_CACHES_HANDLER == None {
```
bwoebi, line 1071: wouldn't something like `let Some(func) = GC_MEM_CACHES_HANDLER else {` be much more idiomatic than having Safety annotations below?

profiling/src/php_ffi.c
```diff
@@ -75,6 +75,12 @@ void datadog_php_profiling_copy_string_view_into_zval(zval *dest, zai_string_vie
     }
 }
 
+void ddog_php_prof_copy_long_into_zval(zval *dest, long num) {
+    ZEND_ASSERT(dest);
+    ZVAL_LONG(dest, num);
+    return;
+}
```
morrisonlevi, line 82: I suppose we could add a `ZEND_ASSERT` to be sure `dest` is undefined or a value which doesn't need refcounts nor dtor'd?

profiling/tests/phpt/gc_mem_caches_01.phpt
```diff
@@ -0,0 +1,36 @@
...
+function should actually cleanup some memory and return the amount in bytes.
+--SKIPIF--
+<?php
+if (!extension_loaded('datadog-profiling'))
+    echo "skip: test requires Datadog Continuous Profiler\n";
+?>
+--ENV--
+DD_PROFILING_ENABLED=yes
+DD_PROFILING_EXPERIMENTAL_CPU_TIME_ENABLED=yes
+DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED=yes
```
morrisonlevi, line 16: Probably add `DD_PROFILING_LOG_LEVEL=off` to this?

morrisonlevi (approved):

profiling/tests/phpt/gc_mem_caches_01.phpt
realFlowControl: See #1876

PR #1867, opened 2023-01-17
Implement #[DDTrace\Trace] attribute

bwoebi:
### Description

Adds a DDTrace\Trace Attribute for PHP 8, to trace functions inline with annotations.

It provides the following named args:
```
        string $name = "",
        string $resource = "",
        string $type = "",
        string $service = "",
        array $tags = [],
        bool|array $args = false,
        bool $return = false,
        bool $recurse = true,
        bool $run_if_limited = false
```

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/hook/uhook_attributes.c
+++ b/ext/hook/uhook_attributes.c
@@ -371,7 +371,7 @@ void dd_uhook_on_function_resolve(zend_function *func) {
                 if (!def->type) {
                     def->type = zend_string_copy(Z_STR(value));
                 }
-            } else if (zend_string_equals_literal(name, "args")) {
+            } else if (zend_string_equals_literal(name, "saveArgs")) {
                 if (!def->arg_whitelist) {
                     if (Z_TYPE(value) == IS_ARRAY) {
                         def->args = true;
@@ -386,7 +386,7 @@ void dd_uhook_on_function_resolve(zend_function *func) {
                         def->args = zend_is_true(&value);
                     }
                 }
-            } else if (zend_string_equals_literal(name, "return")) {
+            } else if (zend_string_equals_literal(name, "saveReturn")) {
                 def->retval = zend_is_true(&value);
             } else if (zend_string_equals_literal(name, "run_if_limited")) {
                 def->run_if_limited = zend_is_true(&value);
@@ -412,7 +412,7 @@ void dd_uhook_on_function_resolve(zend_function *func) {
 void zai_uhook_attributes_minit(void) {
     zai_hook_on_function_resolve = dd_uhook_on_function_resolve;
 
-    ddtrace_hook_attribute_ce = register_class_DDTrace_Traced();
+    ddtrace_hook_attribute_ce = register_class_DDTrace_Trace();
 #if PHP_VERSION_ID >= 80200
     zend_mark_internal_attribute(ddtrace_hook_attribute_ce);
 #endif
@@ -423,6 +423,6 @@ void zai_uhook_attributes_mshutdown(void) {
     zend_string_release(dd_hook_attribute_lcname);
 }
 
-ZEND_METHOD(DDTrace_Traced, __construct) {
+ZEND_METHOD(DDTrace_Trace, __construct) {
     (void) execute_data, (void) return_value;
 }
--- a/ext/hook/uhook_attributes.c
+++ b/ext/hook/uhook_attributes.c
@@ -1,0 +1,428 @@
+#include <php.h>
+#include <Zend/zend_attributes.h>
+#include <Zend/zend_smart_str.h>
+#include "uhook_attributes_arginfo.h"
+#include "../ddtrace.h"
+#include "../configuration.h"
+#include "uhook.h"
+
+#include <hook/hook.h>
+
+zend_class_entry *ddtrace_hook_attribute_ce;
+static zend_string *dd_hook_attribute_lcname;
+
+typedef struct {
+    zend_string *name;
+    zend_string *resource;
+    zend_string *service;
+    zend_string *type;
+    zend_array *tags;
+    bool args;
+    zend_array *arg_whitelist;
+    bool retval;
+    bool run_if_limited;
+    bool active;
+    bool disallow_recursion;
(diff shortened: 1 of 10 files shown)


ext/hook/uhook_attributes.stub.php
```diff
@@ -0,0 +1,22 @@
...
+        string $resource = "",
+        string $type = "",
+        string $service = "",
+        array $tags = [],
+        bool|array $args = false,
+        bool $return = false,
+        bool $recurse = true,
+        bool $run_if_limited = false
+    ) {}
+}
```
morrisonlevi, line 22: I think I understand most of these options, but not `$args` nor `$return`. What are these?

Also, won't the linter complain about `$run_if_limited`? IIRC it likes `$runIfLimited`.

pierotibou: And just to be sure I'd add to the list `$recurse`, `$run_if_limited`

ext/hook/uhook_attributes.c
```diff
@@ -0,0 +1,432 @@
...
+                                            ZAI_HOOK_AUX(def, dd_uhook_dtor), sizeof(dd_uhook_dynamic));
+
+    }
+}
+
+void zai_uhook_attributes_minit(void) {
+    zai_hook_on_function_resolve = dd_uhook_on_function_resolve;
+
+    ddtrace_hook_attribute_ce = register_class_DDTrace_Traced();
+#if PHP_VERSION_ID >= 80200
```
pierotibou, line 420: For posterity, shouldn't you try to explain in comments why the special cases like this one? Or will it be obvious to any PHP dev?

ext/hook/uhook_attributes.stub.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+/** @generate-class-entries */
+
+namespace DDTrace;
+
+require "Zend/zend_attributes.stub.php";
+
+#[\Attribute(\Attribute::TARGET_FUNCTION | \Attribute::TARGET_METHOD)]
+final class Traced {
```
pierotibou, line 10: Other languages implemeting this (at least Java and .NET, haven't checked the others) use `Trace`. We could probably mimick that

pierotibou (commented):
Nice, thanks. A few comments about the feature itself. 
Also, even though I'm not asking you to do it, what would be the extra work needed to support `DD_TRACE_METHODS`, ie listing the methods you want to instrument in a env variable?

ext/hook/uhook_attributes.stub.php
bwoebi: $args and $return log the args and return value into tags. I.e. `$args = ["foo"]` will add a tag `arg.foo = <a possibly shortened representation of the passed parameter>`. (`$args = true` will log them all).

ext/hook/uhook_attributes.c
bwoebi: The method does not exist yet - that's the only reason, which is immediately obvious when commenting the #if out :-D

ext/hook/uhook_attributes.stub.php
bwoebi: Would be fine by me too.

ext/hook/uhook_attributes.stub.php
bwoebi: @morrisonlevi The linter does not complain in fact :-D It's named that way because the array option for trace_function() is also named that way. Won't object to changing, thought it'd be more consistent.

bwoebi:
Supporting a DD_TRACE_FUNCTIONS would be little work. But that is a completely orthogonal feature to this :-) At least it'd be less work than this feature.

ext/hook/uhook_attributes.stub.php
pierotibou: Thanks. And the api looks good to me now that you've renamed a few things.

morrisonlevi:
Do we have customer requests or a specific desire to have args and retval logging? As far as I can remember, we use those things sometimes, but never as recorded like this PR does it. I think I'd rather omit them for now, unless there's some specific use for it.

bwoebi:
I'm not aware of specific requests, it's also disabled by default. However, there are definitely instances, where users (and we ourselves in our integrations) are simply assigning the arg or retval to a meta tag.
That's why, given that this is sometimes very much needed, I'm just giving an easy way out here.

Is there a specific reason why you would omit it except "maybe not needed"?

morrisonlevi:
A few reasons:

- It's more likely to accidentally expose sensitive info.
- Less code to maintain is better.
- I'd rather support operations like "assign the first argument to resource" than put them in meta, and it seems weird to support both strategies, IMO.
- What about meta vs metric? How do you know which it ought to go to?

So, if it's not known to be necessary, then I'd rather avoid it.

bwoebi:
1. If you accidentally opt in into it, yes. But ultimately, it's the users responsibility.
2. That I can agree with :-D
3. I wouldn't object to additionally having resourceArg: int|string (arg num or arg name)
4. If in doubt, meta. Meta is just string and can represent everything. Metric is sort of "optimize as 8 byte float instead of string representation". TBH, I've been considering multiple times in the past whether we should just drop metric...

bwoebi:
However, I would have use cases for arg and retval in my personal project, if I were to instrument it via annotations, that's why I added it - I felt there would be a need for it.

morrisonlevi:
I'm not sure the way to structure the API in case of attributes, but I think `ArgsAsTags(prefix: 'wordpress.hook.', args: ['slug', 'theme', 'area', 'tagName'])` would be better. The ability to prefix the names with a user-chosen thing seems necessary, I think?

morrisonlevi:
I think I'd be happier shipping it without args and retval and then have a separate discussion and PR for those features.

bwoebi:
Okay, let me `#if 0` it out for now.

morrisonlevi (approved):
Approved, though I'm not very familiar with the attribute implementation in PHP. Looks good as far as I can tell.

PR #1871, opened 2023-01-18
refactor(profiling): fix clippy::too_many_arguments

realFlowControl:
### Description

This PR will please clippy!
The `Profiler::prepare_sample_message` method has too many arguments so in this PR I will group all the sample information together in one struct, this also makes the API nicer IMHO.

### Readiness checklist
- [x] ~~Tests added for this feature/bug.~~

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/profiling.rs
+++ b/profiling/src/profiling.rs
@@ -26,6 +26,15 @@ const UPLOAD_PERIOD: Duration = Duration::from_secs(67);
 // magnitude for the capacity.
 const UPLOAD_CHANNEL_CAPACITY: usize = 8;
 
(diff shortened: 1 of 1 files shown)


morrisonlevi:
It's unrelated to this PR, I think, but is this always sending the sample types for allocation profiling even if it's not enabled? If allocation profiling is:

- present but disabled
- not present

then the sample types shouldn't exist in the profile.

realFlowControl:
You are right, I'll create a task for me and fix this in another PR

morrisonlevi (approved):

PR #1872, opened 2023-01-19
fix(profiling): do not send empty profile samples

realFlowControl:
### Description

This is branched of #1871 

Goal is to only send pprof files with samples that were configured to be taken.
For example: Wall time and CPU time profiling is enabled, but allocation profiling is disabled, then we should not send empty allocations, but no allocations at all to save on traffic and processing time.

This PR also changes how we determine if we send CPU time samples:
```diff
-        if locals.last_cpu_time.is_some() {
+        if locals.profiling_experimental_cpu_time_enabled {
```

This would not be strictly necessary, but I would argue that `locals.last_cpu_time` is "just" and indicator, that there are CPU time samples, whereas `locals.profiling_experimental_cpu_time_enabled` is explicit. If we don't like this change, I have no hard feelings reverting it.

Another small changes is some internal variables so we can use the short hand syntax for creating the struct.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/profiling.rs
```diff
@@ -861,24 +861,34 @@ impl Profiler {
...
                 r#type: Cow::Borrowed("wall-time"),
                 unit: Cow::Borrowed("nanoseconds"),
-            },
-            ValueType {
+            });
+            sample_values.push(samples.interrupt_count);
+            sample_values.push(samples.wall_time);
+        }
+
+        if locals.profiling_experimental_allocation_enabled {
```
morrisonlevi, line 880: I think we need to refactor a bit: if profiling isn't enabled, why are we preparing a sample message in the first place? We shouldn't have even collected the samples in the first place, yeah?

morrisonlevi (approved):

PR #1875, opened 2023-01-20
Add 128-bit trace id generation and propagation via _dd.p.tid

bwoebi:
### Description

Adds a new `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED=false` config option.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/ddtrace.h
+++ b/ext/ddtrace.h
@@ -78,8 +78,11 @@ typedef struct {
     union {
         uint64_t high;
         struct {
-            uint32_t padding; // zeroes
-            uint32_t time;
+            ZEND_ENDIAN_LOHI(
+                uint32_t padding // zeroes
+            ,
+                uint32_t time
+            )
         };
     };
 } ddtrace_trace_id;
--- a/ext/ddtrace.h
+++ b/ext/ddtrace.h
@@ -75,7 +75,13 @@ typedef struct {
 
 typedef struct {
     uint64_t low;
-    uint64_t high;
+    union {
+        uint64_t high;
+        struct {
+            uint32_t padding; // zeroes
+            uint32_t time;
+        };
+    };
 } ddtrace_trace_id;
 
 // clang-format off
--- a/ext/span.c
+++ b/ext/span.c
@@ -142,6 +142,11 @@ void ddtrace_open_span(ddtrace_span_data *span) {
     // All open spans hold a ref to their stack
     ZVAL_OBJ_COPY(&span->property_stack, &stack->std);
 
+    span->duration_start = _get_nanoseconds(USE_MONOTONIC_CLOCK);
+    // Start time is nanoseconds from unix epoch
+    // @see https://docs.datadoghq.com/api/?lang=python#send-traces
+    span->start = _get_nanoseconds(USE_REALTIME_CLOCK);
+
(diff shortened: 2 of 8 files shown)


ext/span.c
```diff
@@ -140,6 +140,11 @@ void ddtrace_open_span(ddtrace_span_data *span) {
     // All open spans hold a ref to their stack
     ZVAL_OBJ_COPY(&span->property_stack, &stack->std);
 
+    span->duration_start = _get_nanoseconds(USE_MONOTONIC_CLOCK);
+    // Start time is nanoseconds from unix epoch
+    // @see https://docs.datadoghq.com/api/?lang=python#send-traces
+    span->start = _get_nanoseconds(USE_REALTIME_CLOCK);
```
morrisonlevi, line 146: This seems unrelated to the other changes. Are we expecting these to be set elsewhere, and this is being defensive, or was this a bug?

morrisonlevi (approved):

ext/span.c
morrisonlevi: I see it's related, but I need to read the docs.

bwoebi: Yeah, using the start time afterwards for the high bits.

ext/ddtrace.h
```diff
@@ -75,7 +75,13 @@ typedef struct {
 
 typedef struct {
     uint64_t low;
-    uint64_t high;
+    union {
+        uint64_t high;
+        struct {
+            uint32_t padding; // zeroes
+            uint32_t time;
+        };
```
morrisonlevi, line 83: Based on your commit comment, this is endian specific, yes?

ext/span.c
```diff
@@ -163,7 +163,7 @@ void ddtrace_open_span(ddtrace_span_data *span) {
 set_trace_id_from_span_id:
         span->trace_id = (ddtrace_trace_id){
             .low = span->span_id,
-            .high = get_DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED() ? span->start : 0,
+            .time = get_DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED() ? span->start / 1000000000L : 0,
```
morrisonlevi, line 166: I think you want `UINT64_C` instead of `L`:

```suggestion
            .time = get_DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED() ? span->start / UINT64_C(1000000000) : 0,
```

I'm not feeling well today, but this is going from nanoseconds down to seconds, I think? Just checking that's the intention? I'd add a comment about the unit in `ddtrace_trace_id.time`'s definition.

morrisonlevi (approved):
Approved, with some suggestions. Are we fine to merge this, though? Is the behavior solidified?

ext/ddtrace.h
bwoebi: Yes it does, I'll just do it (though we only really target little endian).

ext/span.c
bwoebi: I did just copy the L from somewhere else and it had the expected value :-D But UINT64_C is more correct, let me do that.

bwoebi:
Yes, we're fine to merge that now :-)

Issue #1880, opened 2023-01-25
[BUG] PHP-FPM + profiling + preload = segfault


B-Galati:
Hello!

### Bug description
I am currently using Datadog tracer and it's fine. Trying to enable Datadog profiler extension is causing PHP-FPM to produce segfault. PHP CLI is doing well :heavy_check_mark:

I found 2 things that fixes the issue:
- Either disable preload
- Or execute PHP-FPM with root (and keep preload of course)

In some ways it looks like to these 2 issues:
- https://github.com/DataDog/dd-trace-php/issues/1795
- https://github.com/DataDog/dd-trace-php/issues/1797

Output of `gdb`:

```
$ gdb php-fpm coredump-php-fpm.3091 
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.1) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from php-fpm...
(No debugging symbols found in php-fpm)

warning: core file may not match specified executable file.
[New LWP 3091]
[New LWP 3093]
[New LWP 3094]
[New LWP 3092]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `php-fpm -v'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f7e7082e2fe in ddtrace_get_profiling_context () at /home/circleci/datadog/tmp/build_extension/ext/profiling.c:11
11    /home/circleci/datadog/tmp/build_extension/ext/profiling.c: No such file or directory.
[Current thread is 1 (Thread 0x7f7e73fbe5c0 (LWP 3091))]
(gdb) 
```

### PHP version
```
$ php -v
PHP 8.1.14 (cli) (built: Jan 13 2023 10:43:22) (NTS)
[...]

PR #1883, opened 2023-01-25
Make active_stack infaillible: there always exists a stack

bwoebi:
### Description

This fixes a crash with toggling the tracing within fibers.

Additionally add another safety NULL-check to ensure #1880 cannot happen. (Blackfire executes code before ddtrace RINIT.)

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/ddtrace.c
+++ b/ext/ddtrace.c
@@ -838,6 +838,7 @@ static PHP_RINIT_FUNCTION(ddtrace) {
     if (!DDTRACE_G(disable)) {
         // With internal functions also being hookable, they must not be hooked before the CG(map_ptr_base) is zeroed
         zai_hook_activate();
+        DDTRACE_G(active_stack) = ddtrace_init_root_span_stack();
     }
 
     if (get_DD_TRACE_ENABLED()) {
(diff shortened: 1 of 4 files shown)


ext/ddtrace.c
```diff
@@ -2048,7 +2050,7 @@ static PHP_FUNCTION(set_priority_sampling) {
         RETURN_FALSE;
     }
 
-    if (global || !DDTRACE_G(active_stack) || !DDTRACE_G(active_stack)->root_span) {
+    if (global || DDTRACE_G(disable) || !DDTRACE_G(active_stack)->root_span) {
```
morrisonlevi, line 2053: This assumes if not disabled that there's an active stack, but in the rshutdown handler there is this:

    DDTRACE_G(active_stack) = NULL;

Which means there is a possibility that it's null, even though it's remote. Thoughts?

bwoebi: Yeah, agreed, see my new commit :-)

morrisonlevi (approved):

PR #1891, opened 2023-01-30
Proposal for standardized storage of installable artifacts

robertomonteromiguel:
### Description

For internal tooling needs (e.g. System Tests), and for having consistent and reliable experience fetching the tracer artifacts. We can leverage a simplified approach of [OCI Artifacts](https://github.com/opencontainers/artifacts) to store the final build artifacts (e.g. dd-java-agent.jar) in GitHub Container Registry.

This PR creates a docker base image with:
- Tag  latest snapshot with php tracer and php appsec tracer
- Tag latest release with php tracer and php appsec tracer

see also: https://docs.google.com/document/d/1HbMGgZFRnIYjfnXGJGyk0LCwWs2op12R0BE58wma448/edit#

The build process will be launched in each commit on main branch.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


tooling/ci/ddtrace-info.sh
```diff
@@ -0,0 +1,28 @@
...
+    "${INSTALLER_ARGS[@]}"\
+    --php-bin all
+
+export DD_APPSEC_ENABLED=1
+
+php -d error_reporting='' -d extension=ddtrace.so -d extension=ddappsec.so -r 'echo phpversion("ddtrace");' > \
+  ./LIBRARY_VERSION
+
+php -d error_reporting='' -d extension=ddtrace.so -d extension=ddappsec.so -r 'echo phpversion("ddappsec");' > \
+  ./PHP_APPSEC_VERSION
```
bwoebi, line 22: Is there a specific reason why there is `-d error_reporting='' -d extension=ddtrace.so -d extension=ddappsec.so` at all? The error presumably is about the extension already being loaded, which it is, thanks to the installer.

bwoebi (approved):
I can't judge the correctness of the gitlab-ci pipeline, but in general it looks sound to me.
If one wants to, download-binary-php.sh could be made generic with respect to the platform (i.e. arm vs amd).

Issue #966, opened 2020-07-16
[Feature] Allow a way to remove some spans from default integrations


khepin:
**Is your feature request related to a problem? Please describe.**
The default laravel tracing of every single `event` becomes problematic when loading large amounts of models. Here's an example of a trace:

![Screen Shot 2020-07-16 at 2 59 49 PM](https://user-images.githubusercontent.com/455656/87727139-4327bc00-c775-11ea-9746-c748ca01e6c3.png)

This query retrieves ~200 models and later on, we retrieve models attached to each of those models, which very rapidly brings us into the thousands of spans which are all of the type `laravel.event.handle` which in our case are not valuable to trace and bring us over the limit of 999 spans per trace.

**Describe the solution you'd like**
Since I seem to be able to replace the default handler for tracing those methods, I'd like to be able to disable it entirely as well and have no span created.

Potentially, this should be part of the `SpanData` itself so that we could keep some events while ignoring others though we don't have a use case for this at the moment.

**Describe alternatives you've considered**
Here's everything I've tried: 
```php
\dd_trace_method('Illuminate\Events\Dispatcher', 'fire', function() {});
\dd_trace_method('Illuminate\Events\Dispatcher', 'fire', null);
\dd_trace_method('Illuminate\Events\Dispatcher', 'fire', false);
\dd_trace_method('Illuminate\Events\Dispatcher', 'fire');
\dd_trace_method('Illuminate\Events\Dispatcher', 'fire', function ($span) {
    $span->setTag(\DDTrace\Tag::MANUAL_DROP, true);
});
```
The final alternative would be to not use the Laravel integration at all and rebuild it on our own. Which isn't a great solution in my opinion.

**Additional context**
Nope

khepin:
![image](https://user-images.githubusercontent.com/455656/87727752-7ae33380-c776-11ea-8dba-f5d24e0b87ff.png)
This is all I've been able to get when setting the trace handler to `false` or `null`

morrisonlevi:
You can choose to drop a span by using returning `false`, but there are some issues with it because it effectively orphans (or drops) all its children spans too. If that's a side-effect you can live with then that's your work-around.

A month or so ago I put in a feature request for being able to drop these kinds of things more reliably for exactly the kind of use case you see here! It hasn't been scheduled, but I'll be sure that it's brought up at the next planning meeting to see when we think we can schedule it.

khepin:
[...]

PR #1897, opened 2023-01-31
Add support for tracing Dispatched Events in Laravel 5.8+

ralphschindler:
### Description

Add the appropriate 5.8+ Event Dispatcher method for tracing. See: https://laravel.com/docs/5.8/upgrade#events

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ralphschindler:
These tests made me think, and maybe you're handling this with other code?....

Should this integration be tracking the dispatching of the event, even if there are no listeners to the event? Or, will the datadog ui be smart enough that if there are no listeners, an empty span for a dispatched event won't show up?

Unsure what it looked like in pre 5.8 when the method `fire` was being tracked.

Thanks for your work on this!

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -140,6 +141,20 @@ function (SpanData $span, $args) use ($integration) {
             }
         );
 
+        // used by Laravel >= 5.8
+        // More details: https://laravel.com/docs/5.8/upgrade#events
+        \DDTrace\trace_method(
+            'Illuminate\Events\Dispatcher',
+            'dispatch',
```
morrisonlevi, line 148: This needs to allow tracing recursion, I think?

bwoebi:
Pre 5.8 it looked exactly like it does in the tests here: https://github.com/DataDog/dd-trace-php/blob/79fe2443ed28a539900531a04e4eed4bfd88dd9d/tests/Integrations/Laravel/V4/CommonScenariosTest.php#L57-L86.

No, the datadog UI does not magically know about there being an event handler.

src/Integrations/Integrations/Laravel/LaravelIntegration.php
bwoebi: Added :-)

bwoebi (approved):
Thanks for the PR!

Anyway, on the topic of:
> Should this integration be tracking the dispatching of the event, even if there are no listeners to the event.

Currently, the way to go is yes.
In particular it isn't that simple to determine whether there's a receiver: it might be broadcast conditionally, there might be listeners on an interface, which are not directly tied to the event itself (Dispatcher::addInterfaceListeners e.g.).
Now the question is, how would we, the tracer, know whether the event actually was dispatched anywhere? There is in fact a way to do that: wrapping any returned Closure from `$this->getListeners()`, but that will be visible in stacktraces. Just simply tracing the returned Closures would be an option - but that functionality is going to only be implemented next week.

Anyway, we may consider improving upon that in future, but for now, that's what we're going to do.

ralphschindler:
Thanks @bwoebi for digging into this PR and taking it to the finish line- and so promptly!
We will be updating to 0.84.0 ASAP.

bwoebi:
0.84.0 just has been released, hope it works just as you expect :-)

joelharkes:
Seems to work well :)

image



It does all of a sudden adds a lot more spans to each process though. People (for example us) wit heavy applications, could have problems reaching the 1000+ spans now or see their datadog bill all of a sudden increase when upgrading the tracer package. good to keep this in mind when adding these things.

joelharkes:
We also trace these things:

```php
  \DDTrace\trace_method(Worker::class, 'runJob', function (\DDTrace\SpanData $span, $args) {
            /** @var Job $job */
            $job = $args[0];
            $span->service = 'worker';
            $span->resource = $job->resolveName();
            $span->name = 'laravel.queue';
            $span->meta['job.name'] = $job->getName();
            $span->meta['job.id'] = $job->getJobId();
            $span->meta['connection'] = $job->getConnectionName();
            $span->meta['queue'] = $job->getQueue();
            $span->meta['attempts'] = $job->attempts();
        });

        \DDTrace\trace_method(Dispatcher::class, 'dispatchNow', function (\DDTrace\SpanData $span, $args) {
            /** @var BaseJob $job */
            $job = $args[0];
            $span->service = 'job';
            $span->name = 'laravel.job';
            $span->resource = class_basename($job);
            if (method_exists($job, 'context')) {
                $context = $job->context();
                if (is_array($context)) {
                    foreach ($context as $key => $value) {
                        if (null !== $value) {
                            $span->meta[$key] = (string) $value;
                        }
                    }
                }
            }
        });

```

and another one for our cron/command dispatcher but that one is completely custom.

jaewun:
Sorry for dredging this up but after updating our staging environment, this caused a bit of trouble for us. Just in case anyone comes here trying to figure out how to reduce the insane amount of spans this produces...

```php
\DD_Trace\trace_method('Illuminate\Events\Dispatcher', 'dispatch', function ($span) {

	//$span->resource contains the event name, you can use this 
	//to decide if you want to drop the span and its children
	
    return false;
});
```

We added some logic in the closure to only drop the noise from Laravel events and keep our own events being tracked. $span->resource holds the event name. As mentioned [here](https://github.com/DataDog/dd-trace-php/issues/966#issuecomment-659720664) it's not perfect but seemed to do the trick for us.

>You can choose to drop a span by using returning false from trace_method/trace_function, but there are some issues with it because it effectively orphans (or drops) all its children spans too. If that's a side-effect you can live with (in many cases you can, because they won't make database calls or do anything else that is interesting) then that's your work-around.

kirkbushell:
> Seems to work well :)
> 
> image
> It does all of a sudden adds a lot more spans to each process though. People (for example us) wit heavy applications, could have problems reaching the 1000+ spans now or see their datadog bill all of a sudden increase when upgrading the tracer package. good to keep this in mind when adding these things.

Yeah this really needed more thought/planning from the product team. The amount of span noise as a result of this change is immense, and has resulted in us not being able to appropriately diagnose performance issues as a result (due to the truncated trace as we have literally thousands of spans).

This needs to be addressed.

PR #1901, opened 2023-02-01
Export ddtrace_flush_tracer and allow flushing during request startup

Anilm3:
### Description

Export `ddtrace_flush_tracer` and allow flushing during request startup. Currently appsec blocks the request primarily on RINIT, which results in no traces in the backend, this should allow the ASM extension to flush the traces before blocking.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
Not sure if we will need it - but we always can remove it later on.

PR #1904, opened 2023-02-03
Add hooking for Closures

bwoebi:
### Description

Extracted from my experiments with revolt; the DDTrace\install_hook() API now got proper instrumentation for Closures. Note that hooks applied to closures only apply to the lifetime of the specific hooked closure.

This makes it trivial to check with a WeakMap whether the Closure already has been hooked.

Additionally the DDTrace\HookData API got a span() method which must be called in the begin handler at least once if the hook is to be traced.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/compatibility.h
+++ b/ext/compatibility.h
@@ -60,6 +60,14 @@
 #define RETURN_COPY(zv) do { RETVAL_COPY(zv); return; } while (0)
 #define RETURN_OBJ_COPY(r) do { RETVAL_OBJ_COPY(r); return; } while (0)
 
+ZEND_API const zend_function *zend_get_closure_method_def(zval *obj);
+static inline const zend_function *dd_zend_get_closure_method_def(zend_object *obj) {
+    zval zv;
+    ZVAL_OBJ(&zv, obj);
+    return zend_get_closure_method_def(&zv);
+}
+#define zend_get_closure_method_def dd_zend_get_closure_method_def
+
 #define ZEND_ARG_OBJ_TYPE_MASK(pass_by_ref, name, class_name, type_mask, default_value) ZEND_ARG_INFO(pass_by_ref, name)
 #define zend_declare_typed_property(ce, name, default, visibility, doc_comment, type) zend_declare_property_ex(ce, name, default, visibility, doc_comment); (void)type
 #define ZEND_TYPE_INIT_MASK(type) NULL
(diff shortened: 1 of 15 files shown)


ext/hook/uhook.c
```diff
@@ -5,32 +5,29 @@
...
+extern void (*profiling_interrupt_function)(zend_execute_data *);
 
 zend_class_entry *ddtrace_hook_data_ce;
 
+static __thread HashTable dd_closure_hooks;
 static __thread HashTable dd_active_hooks;
 
+typedef struct {
+    size_t size;
+    zend_long id[1]; // struct hack
```
morrisonlevi, line 28: Since this exists in a `.c` file and not a header file, I would prefer the standards-compliant flexible array member instead.

```suggestion
    zend_long id[];
```

Obviously other changes are required as well to do the math right.

ext/compatibility.h
```diff
@@ -58,6 +58,14 @@
 #define RETURN_COPY(zv) do { RETVAL_COPY(zv); return; } while (0)
 #define RETURN_OBJ_COPY(r) do { RETVAL_OBJ_COPY(r); return; } while (0)
 
+ZEND_API const zend_function *zend_get_closure_method_def(zval *obj);
+static inline const zend_function *dd_zend_get_closure_method_def(zend_object *obj) {
+    zval zv;
+    ZVAL_OBJ(&zv, obj);
+    return zend_get_closure_method_def(&zv);
+}
+#define zend_get_closure_method_def dd_zend_get_closure_method_def
```
morrisonlevi, line 67: What's the point of the macro, here?

ext/hook/uhook.c
```diff
@@ -103,32 +122,50 @@ static void dd_uhook_call_hook(zend_execute_data *execute_data, zend_object *clo
...
+
+    ddtrace_span_data *span = dyn->hook_data->span;
+    if (span && span->duration != DDTRACE_DROPPED_SPAN && span->duration != DDTRACE_SILENTLY_DROPPED_SPAN) {
+        zval *exception_zv = ddtrace_spandata_property_exception(span);
+        if (EG(exception) && Z_TYPE_P(exception_zv) <= IS_FALSE) {
+            ZVAL_OBJ_COPY(exception_zv, EG(exception));
+        }
+
+        dd_trace_stop_span_time(span);
+    }
```
morrisonlevi, line 167: Apparently, it was only called before if we dropped `.end`, and now we want to always call it.

PR #1905, opened 2023-02-03
Add a DDTrace\generate_distributed_tracing_headers() function

bwoebi:
### Description

It's a direct opposite function to consume_distributed_tracing_headers, thereby trivially enabling users to use it for custom propoagation and later consumption.

Also needed for easy testing of parametric tests.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/ddtrace.c
```diff
@@ -2005,6 +2006,16 @@ static PHP_FUNCTION(consume_distributed_tracing_headers) {
     RETURN_NULL();
 }
 
+/* {{{ proto array generate_distributed_tracing_headers() */
+static PHP_FUNCTION(generate_distributed_tracing_headers) {
+    UNUSED(execute_data);
+
+    array_init(return_value);
+    if (get_DD_TRACE_ENABLED()) {
+        ddtrace_inject_distributed_headers(Z_ARR_P(return_value), true);
```
morrisonlevi, line 2015: If a given request had distributed headers passed to it, do we need to pass on those headers in outgoing requests _even if the tracer is disabled_? My memory isn't clear here, but IIRC there's something like that we should be doing.

morrisonlevi (approved):
Approved, but have a question.

ext/ddtrace.c
bwoebi: If the tracer is disabled, we don't even read the distributed headers. So, here, no, we should not do now.

Maybe, if it's something we should do, we need to refactor a couple things, but that should be separate work.

PR #1908, opened 2023-02-04
refactor(profiling): clean up stalk walking

morrisonlevi:
### Description

- Filename and lineno are both only used for user functions. Merging them into a single function simplifies the code.
- Using as_ref() on pointers cleans up the code in many places as it allows using methods like `filter` and `map`.
- Renames `datadog_php_profiling_zend_string_view` to `ddog_php_prof_zend_string_view` as per our modern naming.
- Adds new `collect_call_frame` function which is called by `collect_stack_sample` on each iteration of the stack-walking.
- Checking for stack truncation is moved from being done every loop, to only when a sample is pushed. For performance this should be slightly better but almost negligible. It's done because it's more maintainable code, the performance is just a nice cherry on top.

### Motivation

I was investigating the stalk walking in preparation to use the per-function runtime cache to store a `pprof.Function.id`, which should speed us up on a critical performance path. This has a few other prerequisites, but I noticed I could clean this code up. I figured I may as well get this part committed to make that PR smaller when the time comes.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ It's pure refactoring, no test changes.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/profiling.rs
+++ b/profiling/src/profiling.rs
@@ -162,7 +162,7 @@ pub struct ZendFrame {
 
 // todo: dedup
 unsafe fn zend_string_to_bytes(zstr: Option<&mut zend_string>) -> &[u8] {
-    bindings::datadog_php_profiling_zend_string_view(zstr).into_bytes()
+    bindings::ddog_php_prof_zend_string_view(zstr).into_bytes()
 }
 
 /// Extract the "function name" component for the frame. This is a string which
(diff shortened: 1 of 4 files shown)


profiling/src/profiling.rs
```diff
@@ -183,7 +183,7 @@ unsafe fn extract_function_name(func: &zend_function) -> String {
      * erring on the side of caution and returning early.
      */
     if method_name.is_empty() {
-        return String::new();
+        return None;
```
realFlowControl, line 186: praise: I like this API change very much 👍

morrisonlevi: Thank you :)

realFlowControl (approved):
LGTM 🚀

PR #1910, opened 2023-02-07
User tracking through DDTrace\set_user

Anilm3:
### Description

This PR introduces a new function `DDTrace\set_user(user_id, metadata, propagate)` which allows the user to conveniently add relevant user tracking tags to the root span. The expected parameters are the following:

- `user_id`: a non-empty string.
- (optional) `metadata`: a hash table containing relevant user information, e.g. `['email' => 'someone@somewhere.com', 'role' => 'artist']`
- (optional) `propagate`: a boolean indicating whether to propagate the user ID or not, the default value is false.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/ddtrace.c
```diff
@@ -327,6 +327,12 @@ ZEND_ARG_INFO(0, key)
 ZEND_ARG_INFO(0, value)
 ZEND_END_ARG_INFO()
 
+ZEND_BEGIN_ARG_INFO_EX(arginfo_ddtrace_set_user, 0, 0, 3)
```
bwoebi, line 330: the last param is required_num_args, should be 1.

ext/ddtrace.c
```diff
@@ -1185,6 +1191,71 @@ static PHP_FUNCTION(add_distributed_tag) {
     RETURN_NULL();
 }
 
+static PHP_FUNCTION(set_user) {
+    UNUSED(execute_data);
+
+    zend_string *user_id = NULL;
```
bwoebi, line 1197: Can never be NULL after zpp (not optional), you can remove the NULL assign and NULL check later.

ext/ddtrace.c
```diff
@@ -1185,6 +1191,71 @@ static PHP_FUNCTION(add_distributed_tag) {
...
+    HashTable *metadata = NULL;
+    zend_bool propagate = false;
```
bwoebi, line 1199: I think there should be a config for default usr.id propagation; especially when we will use DDTrace\set_user() in integrations code, that needs to be toggleable anyway. So I propose doing it here.

ext/ddtrace.c
```diff
@@ -1185,6 +1191,71 @@ static PHP_FUNCTION(add_distributed_tag) {
...
+        ZVAL_COPY(&value_zv, &user_id_zv);
+        zend_hash_str_update(target_table, ZEND_STRL("_dd.p.usr.id"), &value_zv);
+
+        zend_hash_str_add_empty_element(&DDTRACE_G(propagated_root_span_tags),
+            ZEND_STRL("_dd.p.usr.id"));
+    }
+
+    if (metadata != NULL) {
+        zend_string *key = NULL;
+        zval *value = NULL;
```
bwoebi, line 1238: I'm a proponent of trying to assign default values as little as possible so that a compiler may warn about possibly uninitialized values instead of running into a NULL at runtime.

ext/ddtrace.c
Anilm3: Okay, I'll add a `DD_TRACE_PROPAGATE_USER_ID_DEFAULT` set to false by default in configuration and I'll still keep the parameter so that the user can later still decide regardless of the configuration.

ext/ddtrace.c
Anilm3: Fair enough I'll follow that convention where feasible.

ext/ddtrace.c
bwoebi: Yes, exactly, definitely retain the arg :-)

bwoebi (approved):
LGTM :-)

PR #1912, opened 2023-02-07
tracing: add db.system tag to all relevant integrations

wconti27:
### Description
Add new `db.system` tag - An identifier for the database management system (DBMS) product being used - to all database and caching integrations. 

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/Integrations/Integrations/Memcached/MemcachedIntegration.php
```diff
@@ -185,6 +185,7 @@ public function setCommonData(SpanData $span, $command)
         $span->meta['memcached.command'] = $command;
         $span->meta[Tag::SPAN_KIND] = 'client';
         $span->meta[Tag::COMPONENT] = MemcachedIntegration::NAME;
+        $span->meta['db.system'] = MemcachedIntegration::NAME;
```
morrisonlevi, line 188: Maybe add `Tag::DB_SYSTEM` since this is a well-known tag?

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -14,6 +14,10 @@ class PDOIntegration extends Integration
 
     const CONNECTION_TAGS_KEY = 'connection_tags';
 
+    const ENGINE_TO_SYSTEM_ARRAY = [
```
morrisonlevi, line 17: As far as I understood your code, a better word than "engine" would be "drivers".

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -14,6 +14,10 @@ class PDOIntegration extends Integration
...
+        'mysql' => 'mysqldb',
+    ];
```
morrisonlevi, line 19: As far as I understood it, "driver" is a better name than "engine". You can find the driver names in the docs:

https://www.php.net/manual/en/pdo.drivers.php

Then for each one, look for their `*_DSN` docs.

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -149,6 +149,7 @@ private static function parseDsn($dsn)
     {
         $engine = substr($dsn, 0, strpos($dsn, ':'));
         $tags = ['db.engine' => $engine];
+        $tags[Tag::DB_SYSTEM] = $tags['db.engine'];
```
morrisonlevi, line 152: The driver name is not always the same as the standardized value for that system. For instance, `pgsql` is the driver name, but the standardized name for that is `postgresql`. You need to do a mapping, at least for this one case (maybe others), hence my previous comment about going through the docs.

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -14,6 +14,21 @@ class PDOIntegration extends Integration
...
+        'cubrid' => 'other_sql',
+        'dblib' => 'other_sql',
+        # may be mssql or Sybase, not supported anymore so shouldn't be a problem
+        'firebird' => 'firebird',
+        'ibm' => 'db2',
+        'informix' => 'informix',
+        'mysql' => 'mysql',
+        'sqlsrv' => 'mssql',
+        'oci' => 'oracle',
+        'odbc' => 'other',
```
morrisonlevi, line 27: `'other_sql'`, yeah?

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -14,6 +14,21 @@ class PDOIntegration extends Integration
 
     const CONNECTION_TAGS_KEY = 'connection_tags';
 
+    const DB_DRIVER_TO_SYSTEM = [
+        'cubrid' => 'other_sql',
+        'dblib' => 'other_sql',
+        # may be mssql or Sybase, not supported anymore so shouldn't be a problem
```
morrisonlevi, line 20: Comment style should be `//` or `/* */` these days, not `#`, which have been repurposed for attributes.

src/Integrations/Integrations/PDO/PDOIntegration.php
wconti27: This should be all set.

src/Integrations/Integrations/PDO/PDOIntegration.php
wconti27: Fixed this.

src/Integrations/Integrations/PDO/PDOIntegration.php
wconti27: fixed

src/Integrations/Integrations/PDO/PDOIntegration.php
wconti27: fixed

bwoebi: (Using # followed by a space is and will always be fine.)

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -149,6 +164,7 @@ private static function parseDsn($dsn)
     {
         $engine = substr($dsn, 0, strpos($dsn, ':'));
         $tags = ['db.engine' => $engine];
+        $tags[Tag::DB_SYSTEM] = self::DB_DRIVER_TO_SYSTEM[$tags['db.engine']];
```
morrisonlevi, line 167: I'd probably do something more like this:
```suggestion
        $tags[Tag::DB_SYSTEM] = isset(self::DB_DRIVER_TO_SYSTEM[$engine]) ? self::DB_DRIVER_TO_SYSTEM[$engine] : 'other_sql';
```

This is a little bit more resilient, and it also uses `$engine` directly instead of using `$tags['db.engine']` (which gets its info from `$engine`).

wconti27: Good idea! Just added this suggestion

morrisonlevi (approved):
Looks good to me!

morrisonlevi (changes requested):
Actually, wait, I noticed a CI job that failed, checking.

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -164,9 +164,8 @@ private static function parseDsn($dsn)
     {
         $engine = substr($dsn, 0, strpos($dsn, ':'));
         $tags = ['db.engine' => $engine];
-        $tags[Tag::DB_SYSTEM] = isset(self::DB_DRIVER_TO_SYSTEM[$engine])
-            ? self::DB_DRIVER_TO_SYSTEM[$engine]
-            : 'other_sql';
+        $db_system = self::DB_DRIVER_TO_SYSTEM[$engine];
+        $tags[Tag::DB_SYSTEM] = isset($db_system) ? $db_system : 'other_sql';
```
morrisonlevi, line 168: This will generate a warning, which we swallow, but does this really fix an issue with the compiler? Seems odd, the former looks correct.

wconti27: I've now replaced this with just `$tags[Tag::DB_SYSTEM] = self::DB_DRIVER_TO_SYSTEM[$engine] ?? 'other_sql` but still getting the build errors. This was the only change that was made, the currently failing tests were passing yesterday when this line was different :thinking_face:

bwoebi: Yes, we've been rebuilding the docker containers, needs #1922 to be merged.

bwoebi: If you rebase it on master, then build should be green now.

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -14,6 +14,21 @@ class PDOIntegration extends Integration
 
     const CONNECTION_TAGS_KEY = 'connection_tags';
 
+    const DB_DRIVER_TO_SYSTEM = [
```
bwoebi, line 17: The PHP files listed in `_files_integrations.PHP5.php` must all still confirm to PHP 5.4 syntax, because they're still distributed from the master branch. We need to eventually fully separate these, but for now, arrays are not allowed in constants, nor is `??`.

wconti27: was not aware of this, thanks for letting me know. This is should be all set.

bwoebi (approved):

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -14,6 +14,21 @@ class PDOIntegration extends Integration
...
+    private static $DB_DRIVER_TO_SYSTEM = [
+        'cubrid' => 'other_sql',
+        'dblib' => 'other_sql',
+        // may be mssql or Sybase, not supported anymore so shouldn't be a problem
+        'firebird' => 'firebird',
+        'ibm' => 'db2',
+        'informix' => 'informix',
+        'mysql' => 'mysql',
+        'sqlsrv' => 'mssql',
+        'oci' => 'oracle',
```
BridgeAR, line 26: I believe these should be

```suggestion
        'firebird' => 'firebirdsql',
        'ibm' => 'ibm.db2',
        'informix' => 'ibm.informix',
        'mysql' => 'mysql',
        'sqlsrv' => 'mssql',
        'oci' => 'oracle.db',
```

as defined in https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/

wconti27: agreed, we should follow OTel conventions as much as possible

PR #1913, opened 2023-02-08
perf(profiling): increased allocation sampling distance to 512KiB

realFlowControl:
### Description

Did some benchmarks in relenv and saw that we do not get a better resolution with 50 KB sampling distance compared to 512 KB, but we pay with a higher CPU time overhead.
This PR will lower the sampling distance to 512 KB.

PROF-7186

### Readiness checklist
- [ ] ~~Tests added for this feature/bug.~~

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (approved):
This is exactly what I recommended in slack when tracking the average number of stalk walks  per requests. This may also still be too high, or the app I was tracking may generate a lot of garbage, but _at least this_ is necessary. Thanks!

PR #1914, opened 2023-02-08
Add hooking for abstract methods

bwoebi:
### Description

This PR allows tracing of interface and abstract parent methods.

Abstract trait methods are not supported due to complexity with renaming etc. They may be considered in the future, but not now.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/tests/ext/sandbox/abstract_trait_hook.phpt
+++ b/tests/ext/sandbox/abstract_trait_hook.phpt
@@ -1,0 +1,49 @@
+--TEST--
+[Sandbox] Hook implementations of abstract trait methods are not supported
+--FILE--
+<?php
+
+DDTrace\hook_method("Ancestor", "Method", function() {
+    echo "EARLY Ancestor HOOK\n";
+});
+
+DDTrace\hook_method("Base", "Method", function() {
+    echo "EARLY Base HOOK\n";
+});
+
+// Ensure run-time resolving
+if (true) {
+    trait Ancestor {
+        public abstract function Method();
+    }
+}
+
+class Base {
+    use Ancestor;
+
+    public function Method() {
+        echo "METHOD Base\n";
+    }
+}
+
+DDTrace\hook_method("Ancestor", "Method", function() {
+    echo "LATE Ancestor HOOK\n";
+});
+
+DDTrace\hook_method("Base", "Method", function() {
+    echo "LATE Base HOOK\n";
+});
+
+(new Base())->Method();
+
+dd_untrace("Method", "Ancestor");
+(new Base())->Method();
+
+?>
+--EXPECT--
+EARLY Base HOOK
+LATE Base HOOK
+METHOD Base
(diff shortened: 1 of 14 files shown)


zend_abstract_interface/hook/hook.c
```diff
@@ -58,6 +59,19 @@ __thread HashTable zai_hook_request_classes;
 // if indirect, then it's pointing to some hashtable in zai_hook_request_functions/classes
 __thread HashTable zai_hook_resolved;
 
+// zai_hook_inheritors is a map of persistent class entries (interfaces and abstract classes) to a list of persistent class entries
+static HashTable zai_hook_static_inheritors;
```
realFlowControl, line 63: question: "persistent classes" as in preloaded? Like: which classes survive a request shutdown? Oh, maybe builtin classes?

zend_abstract_interface/hook/hook.c
```diff
@@ -124,6 +144,15 @@ static void zend_hash_iterators_remove(HashTable *ht) {
     }
 }
 
+
+static void zai_hook_static_inheritors_destroy(zval *zv) {
+    free(Z_PTR_P(zv));
```
realFlowControl, line 149: question: I have not found where these pointers are allocated, but AFAIK this function is to free the elements in `zai_hook_static_inheritors` (as the call to `zend_hash_init` suggests).
Long story short, is `free(Z_PTR_P(zv));` correct or should it be `pefree(Z_PTR_P(zv),1);`?
I understood, that `pefree(void *, 1)` is just a call to `free(void *)` but if my thought is right, this would be more consistent.

Is this the line corresponding to this `free()`?
https://github.com/DataDog/dd-trace-php/pull/1914/files#diff-f4b7c344c209da085fdb03ad5b08198ef92790f7fc43751e23486fc56a1bc857R446

tests/ext/sandbox/abstract_trait_hook.phpt
```diff
@@ -0,0 +1,49 @@
+--TEST--
+[Sandbox] Hook implementations of abstract trait methods are not supposed
```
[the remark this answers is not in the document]
bwoebi: Yes :-)

zend_abstract_interface/hook/hook.c
bwoebi: I think in my code I've nearly always been using free() - only pefree() when it's dynamic.

The distinction actually only matters for malloc/pemalloc(), as malloc is faillible, while pemalloc(, 1) is not.

zend_abstract_interface/hook/hook.c
bwoebi: Yes, persistent classes include preloaded (there's also a comment in line 928) - builtin classes are "internal".

realFlowControl (approved):
LGTM 😄

PR #1916, opened 2023-02-09
Fetch the latest release from github API instead of manually updating it

bwoebi:
### Description

Why automate it when you also can do it manually :-)

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -4,8 +4,8 @@ stages:
 
 variables:
   LATEST_LIBRARY_x86_64_LINUX_GNU:
-    value: "https://github.com/DataDog/dd-trace-php/releases/download/0.84.0/dd-library-php-0.84.0-x86_64-linux-gnu.tar.gz"
-    description: "Location where to download latest dd-library-php-*-x86_64-linux-gnu.tar.gz archive"
+    value: ""
+    description: "Location where to download latest dd-library-php-*-x86_64-linux-gnu.tar.gz archive. Leave empty to take it from the latest released github tag."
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -4,8 +4,8 @@ stages:
 
 variables:
   LATEST_LIBRARY_x86_64_LINUX_GNU:
-    value: "https://github.com/DataDog/dd-trace-php/releases/download/0.83.1/dd-library-php-0.83.1-x86_64-linux-gnu.tar.gz"
-    description: "Location where to download latest dd-library-php-*-x86_64-linux-gnu.tar.gz archive"
+    value: ""
+    description: "Location where to download latest dd-library-php-*-x86_64-linux-gnu.tar.gz archive. Leave empty to take it from the latest released github tag."
```
labbati, line 8: Should be default point to "master" (no how good ideas for how to do that, maybe some public scraping from circleci) or the latest release?
Even if we prefer master, no need to do it as part of this PR — latest release is still better then manually updating — but maybe an idea for the future

bwoebi: The point of last release is being able to compare a new release (which is typically the state of master) with the last release. Makes no sense to have the current master here thus.

labbati: ah ok so this is the baseline. so yeah then makes no sense, I thought it was the library to the actual library to deploy to the rel env! Thanks for clarify that

pierotibou (approved):

PR #1917, opened 2023-02-09
Fix automatic injection of w3c headers

bwoebi:
### Description

And manual injection of x-datadog-origin.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/ddtrace.c
+++ b/ext/ddtrace.c
@@ -2382,7 +2382,7 @@ void ddtrace_read_distributed_tracing_ids(bool (*read_header)(zai_string_view, c
     }
 
     // "{version:2}-{trace-id:32}-{parent-id:16}-{trace-flags:2}"
-    if (parse_tracestate && read_header(ZAI_STRL_VIEW("X_TRACEPARENT"), "traceparent", &traceparent, data)) {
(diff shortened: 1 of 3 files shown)


ext/ddtrace.c
```diff
@@ -2332,7 +2332,7 @@ void ddtrace_read_distributed_tracing_ids(bool (*read_header)(zai_string_view, c
         ZVAL_STR(&trace_zv, trace_id_str);
         DDTRACE_G(distributed_trace_id) = (ddtrace_trace_id){ .low = ddtrace_parse_userland_span_id(&trace_zv) };
         zend_string_release(trace_id_str);
-    } else if (parse_b3 && read_header(ZAI_STRL_VIEW("X_B3_TRACEID"), "b3-traceid", &trace_id_str, data)) {
+    } else if (parse_b3 && read_header(ZAI_STRL_VIEW("X_B3_TRACEID"), "x-b3-traceid", &trace_id_str, data)) {
```
pierotibou, line 2335: Would it make sense to start using constants to avoid mistakes like this?

bwoebi: It's the only usage of these strings though... we could have constants, sure, but the constants can also have the wrong values then :-D

PR #1926, opened 2023-02-15
Use checked_sub duration ops for handling cpu time

bwoebi:
### Description

This is never supposed to fail, but better check for it than avoiding it.

Fixes an issue reported in #1880.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (approved):
Thanks. Really odd as this shouldn't be an issue since we're on Rust 1.60, but nonetheless, here we are.

PR #1927, opened 2023-02-15
fix: mask Zend Signals in profiling threads

morrisonlevi:
### Description

Mask Zend Signals in profiling threads.
Also, add thread names for profiling threads.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi:
Tests fixed here: #1929.

profiling/src/profiling.rs
```diff
@@ -646,8 +687,8 @@ impl Profiler {
             fork_senders: [fork_sender0, fork_sender1],
             vm_interrupts,
             message_sender,
-            time_collector_handle: std::thread::spawn(move || time_collector.run()),
-            uploader_handle: std::thread::spawn(move || uploader.run()),
+            time_collector_handle: Self::spawn("ddprof_time", move || time_collector.run()),
+            uploader_handle: Self::spawn("ddprof_upload", move || uploader.run()),
```
realFlowControl, line 691: So we will stop seeing `thread '<unnamed>' panicked at` and instead see `thread 'ddprof_time' panicked at` in the future?
🤞 for hopefully not seeing this at all 😉

realFlowControl (approved):
LGTM 🚀

profiling/src/profiling.rs
morrisonlevi: We _might_, yeah. It depends on the implementation, but I believe the Linux one will call `pthread_setname_np`, which has a limit of 16 chars and must be null terminated, so 15 non-null chars. This is why I opted for the very short abbreviation instead of `ddog_pprof_*`.

PR #1930, opened 2023-02-15
Upgrade CentOS 7 images to Rust 1.64

morrisonlevi:
### Description

These images are already pushed, I just forgot to commit them.

We don't need protoc-c anymore since we use Rust.

Edit: We found the need to downgrade PHP 8.2 to 8.2.1 due to this bug:
https://github.com/php/php-src/commit/95016138a54b3352a0988878cd71c2ebe7cdeca5

So, we're taking the opportunity to rebuild and fix this in a few places.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


realFlowControl (approved):
This is the way! 🎉

Issue #1919, opened 2023-02-10
[BUG] FPM shows warnings about rust errors when activating profiler


Spea:
### Bug description

We activated the datadog profiler in our php applications, but FPM reports a warning about an error in rust 👇 

```
WARNING: child 77 said into stderr: "thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/thread/mod.rs:1311:40"
WARNING: child 77 said into stderr: "stack backtrace:"
WARNING: child 77 said into stderr: "   0:     0x7f4043390f74 - std::backtrace_rs::backtrace::libunwind::trace::hee598835bc88d35b"
WARNING: child 77 said into stderr: "                               at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5"
WARNING: child 77 said into stderr: "   1:     0x7f4043390f74 - std::backtrace_rs::backtrace::trace_unsynchronized::h9cdc730ba5cf5d72"
WARNING: child 77 said into stderr: "                               at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5"
WARNING: child 77 said into stderr: "   2:     0x7f4043390f74 - std::sys_common::backtrace::_print_fmt::h75aeaf7ed30e43fa"
WARNING: child 77 said into stderr: "                               at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/sys_common/backtrace.rs:66:5"
WARNING: child 77 said into stderr: "   3:     0x7f4043390f74 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h606862f787600875"
WARNING: child 77 said into stderr: "                               at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/sys_common/backtrace.rs:45:22"
WARNING: child 77 said into stderr: "   4:     0x7f40431a556c - core::fmt::write::he803f0f418caf762"
WARNING: child 77 said into stderr: "                               at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/fmt/mod.rs:1190:17"
[...]

PR #1932, opened 2023-02-16
Mitigate panic in profiler shutdown, bound timeout

morrisonlevi:
### Description

This mitigation is for issue #1919. It doesn't fix the root cause, which we haven't been able to determine yet.

This also bounds our shutdown timeouts. Before, they _ought_ to have been bounded as a consequence of how the code was written, but this is explicit. It's also a tighter bound than before, 2 seconds instead of 10 seconds.

Also, the new Rust 1.64 formatter apparently formats some things differently, so those are included as well.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -1194,47 +1194,28 @@ mod tests {
     #[test]
     fn detect_uri_from_config_works() {
         // expected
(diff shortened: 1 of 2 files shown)


realFlowControl (approved):
LGTM 👍

PR #1936, opened 2023-02-21
Add PSR18 tracing support in Guzzle 7

gmessier:
### Description

Guzzle 7 implement the PSR18 (Http client) interface. That means the client has a new method named "sendRequest" with an implementation very similar to the "send" method. 

This PR is to add tracing to this "sendRequest" method.

NOTE: I didn't add tests as I'm not really sure how to handle them (currently the tests are for Guzzle V5 and V6 only, but V7 was released in 2020. The V5 is not supported anymore and the V6 is in security fixes only.)

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/Guzzle/GuzzleIntegration.php
+++ b/src/Integrations/Integrations/Guzzle/GuzzleIntegration.php
@@ -63,6 +63,32 @@ class GuzzleIntegration extends Integration
             }
         );
 
+        /* Until we support both pre- and post- hooks on the same function, do
+         * not send distributed tracing headers; curl will almost guaranteed do
+         * it for us anyway. Just do a post-hook to get the response.
+         */
+        \DDTrace\trace_method(
+            'GuzzleHttp\Client',
+            'sendRequest',
+            function (SpanData $span, $args, $retval) use ($integration) {
+                $span->resource = 'sendRequest';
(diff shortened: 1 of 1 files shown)


bwoebi:
Hey @gmessier,

This sounds like a great idea. I wonder whether it makes sense here to instrument the interface method directly instead of the Guzzle-specific implementation?

We'll have a look at adding tests soon.

gmessier:
I though that abstract classes and interfaces were not working anymore, but if it's fixed before this is added, yes it should be a good idea to trace the PSR interface directly instead of the Guzzle implementation as it will be a lot more generic. 

But that means another integration (as it should not be in the Guzzle one) and changing a little bit the values (the service will not be "guzzle" anymore, but maybe simply "http-client" or something like that). I could help, but I will probably forgot things.

I will create another PR and we will see at the end which one we choose.

bwoebi:
It would need to be based on top of https://github.com/DataDog/dd-trace-php/pull/1914 (which supports that), but we are going to merge that soon I think.

gmessier:
The other PR for a more generic implementation (any PSR18 class): https://github.com/DataDog/dd-trace-php/pull/1938

gmessier:
I will close this PR as the generic one for any PSR18 implementation is better.

PR #1937, opened 2023-02-22
Add HookData->overrideArguments() method to replace arguments in a pre-hook

bwoebi:
### Description

This functionality is required to inject some data on method calls (like adding logging info, mostly trace/span ids).

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/hook/uhook.stub.php
+++ b/ext/hook/uhook.stub.php
@@ -12,6 +12,7 @@ class HookData {
     public ?\Throwable $exception;
     /**
      * Creates a span if none exists yet, otherwise returns the span attached to the current function call.
+     * If called outside the pre-hook and no span is attached yet, it will also return null.
      *
      * @param SpanStack|SpanData|null $parent May be specified to start a span on a specific stack.
      *                                        As an example, when instrumenting closures, it might conceptually make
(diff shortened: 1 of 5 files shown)


ext/hook/uhook.stub.php
```diff
@@ -20,7 +21,16 @@ class HookData {
...
     public function span(SpanStack|SpanData|null $parent = null): SpanData;
+
+    /**
+     * Replaces the arguments of a function call. Must be called within a pre-hook.
+     * It is not allowed to pass more arguments to a function that currently on the stack or total number or arguments,
+     * whichever is greater.
+     *
+     * @param array $arguments An array of arguments, which will replace the hooked functions arguments.
+     */
+    public function overrideArguments(array $arguments): void;
```
realFlowControl, line 32:
```suggestion
    public function overrideArguments(array $arguments): bool;
```
idea: I was thinking that it would be nice to know if this actually worked so the developer can react on this. I don't know if this would be different from how all the other stuff is currently handled, so it is just a suggestion and non blocking

realFlowControl (approved):
Awesome, just had one remark, but this is non blocking 🚀

PR #1939, opened 2023-02-25
[CI] Switch to smaller instances

pierotibou:
### Description

It seems that most of our stages can run with medium instances that are 10 times cheaper.
Also, reduced the number of scenarios in the random tests. Will speed it up, reduce usage and from Bob's experience, it's been very rare that a crash occurred only in single randomized test.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
Looks good, hasn't become too slow, so let's go ahead with this :-)

PR #1949, opened 2023-03-01
perf(profiling): speed up stalk walking by using function run_time_cache

morrisonlevi:
### Description

This is for PHP 8 only. It uses two slots in the function run_time_cache to store:

- The function name, which looks something like `"module|Namespace\\ClassName::method"`.
- The file name.

These two strings are interned and live between requests. This allows us to share them as `Cow<'static, str>` with the `ddprof_time` thread.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/profiling/src/php_ffi.c
+++ b/profiling/src/php_ffi.c
@@ -42,8 +42,8 @@ void datadog_php_profiling_startup(zend_extension *extension) {
 
 void *datadog_php_profiling_vm_interrupt_addr(void) { return &EG(vm_interrupt); }
 
-zend_module_entry *datadog_get_module_entry(const uint8_t *str, uintptr_t len) {
-    return zend_hash_str_find_ptr(&module_registry, (const char *)str, len);
+zend_module_entry *datadog_get_module_entry(const char *str, uintptr_t len) {
+    return zend_hash_str_find_ptr(&module_registry, str, len);
 }
 
 ddtrace_profiling_context (*datadog_php_profiling_get_profiling_context)(void) =
@@ -95,8 +95,7 @@ zai_string_view ddog_php_prof_zend_string_view(zend_string *zstr) {
 void ddog_php_prof_zend_mm_set_custom_handlers(zend_mm_heap *heap,
                                                void* (*_malloc)(size_t),
                                                void  (*_free)(void*),
-                                               void* (*_realloc)(void*, size_t))
-{
+                                               void* (*_realloc)(void*, size_t)) {
     zend_mm_set_custom_handlers(heap, _malloc, _free, _realloc);
 #if PHP_VERSION_ID < 70300
     if (!_malloc && !_free && !_realloc) {
@@ -105,7 +104,53 @@ void ddog_php_prof_zend_mm_set_custom_handlers(zend_mm_heap *heap,
 #endif
 }
 
-zend_execute_data* ddog_php_prof_get_current_execute_data()
-{
+zend_execute_data* ddog_php_prof_get_current_execute_data() {
     return EG(current_execute_data);
+}
+
+#if PHP_VERSION_ID >= 80000
+static int ddog_php_prof_run_time_cache_handle = -1;
+#endif
+
+void ddog_php_prof_function_run_time_cache_init(const char *module_name) {
+#if PHP_VERSION_ID >= 80000
+    // Grab 2, one for function name and one for filename.
+    ddog_php_prof_run_time_cache_handle =
+        zend_get_op_array_extension_handles(module_name, 2);
+#endif
+
+    /* It's possible to work on PHP 7.4 as well, but there are opcache bugs
+     * that weren't truly fixed until PHP 8:
+     * https://github.com/php/php-src/pull/5871
+     * I would rather avoid these bugs for now.
+     */
+}
+
+uintptr_t *ddog_php_prof_function_run_time_cache(zend_function *func) {
+#if PHP_VERSION_ID < 80000
+    /* It's possible to work on PHP 7.4 as well, but there are opcache bugs
+     * that weren't truly fixed until PHP 8:
+     * https://github.com/php/php-src/pull/5871
+     * I would rather avoid these bugs for now.
+     */
+    return NULL;
+#else
+
+    // It should be initialized by this point, or we failed.
+    if (ddog_php_prof_run_time_cache_handle < 0) return NULL;
+
+#if PHP_VERSION_ID < 80200
+    // internal functions don't have a runtime cache until PHP 8.2
+    if (func->type == ZEND_INTERNAL_FUNCTION) return NULL;
+
+    uintptr_t *cache_addr = RUN_TIME_CACHE(&func->op_array);
+#else
+    uintptr_t *cache_addr = RUN_TIME_CACHE(&func->common);
+#endif
+
+    // To my knowledge, this is always a bug, but it has happened.
+    if (!cache_addr) return 0;
+
+    return cache_addr + ddog_php_prof_run_time_cache_handle;
+#endif
 }
--- a/profiling/src/profiling/stalk_walking.rs
+++ b/profiling/src/profiling/stalk_walking.rs
@@ -101,6 +101,7 @@ unsafe fn handle_file_cache_slot_helper(
     // Safety: changing the lifetime to 'static is safe because
     // the other threads using it are joined before this thread
     // ever dies.
+    // todo: this is _not_ ZTS safe.
     Some(Cow::Borrowed(transmute(str)))
 }
 
@@ -139,6 +140,7 @@ unsafe fn handle_function_cache_slot(
     // Safety: changing the lifetime to 'static is safe because
     // the other threads using it are joined before this thread
     // ever dies.
+    // todo: this is _not_ ZTS safe.
     Some(Cow::Borrowed(transmute(str)))
 }
 
--- a/profiling/src/profiling/stalk_walking.rs
+++ b/profiling/src/profiling/stalk_walking.rs
@@ -1,15 +1,32 @@
 use crate::bindings::{
-    ddog_php_prof_zend_string_view, zend_execute_data, zend_function, zend_string,
-    ZEND_USER_FUNCTION,
+    ddog_php_prof_function_run_time_cache, ddog_php_prof_zend_string_view, zend_execute_data,
+    zend_function, zend_string, ZEND_USER_FUNCTION,
 };
+use datadog_profiling_store::{OwnedStringTable, StringTable};
+use log::debug;
+use std::borrow::Cow;
+use std::cell::{RefCell, RefMut};
+use std::mem::transmute;
 use std::str::Utf8Error;
+
+#[derive(Debug, Default)]
+pub struct FunctionRunTimeCacheStats {
+    hit: usize,
+    missed: usize,
+    not_applicable: usize,
+}
+
+thread_local! {
+    static CACHED_STRINGS: RefCell<OwnedStringTable> = RefCell::new(OwnedStringTable::new());
+    pub static FUNCTION_CACHE_STATS: RefCell<FunctionRunTimeCacheStats> = RefCell::new(Default::default())
+}
 
 #[derive(Default, Debug)]
 pub struct ZendFrame {
     // Most tools don't like frames that don't have function names, so use a
-    // fake name if you need to like "<php>".
-    pub function: String,
-    pub file: Option<String>,
+    // fake name if you need to like "<?php".
+    pub function: Cow<'static, str>,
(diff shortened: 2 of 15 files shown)


morrisonlevi:
Numbers time! Using the Datadog native profiler, I analyzed memory and CPU characteristics. Here's the short version:

- When just wall-time and cpu-time are enabled, CPU and latency are about the same, but down slightly. The number of allocations and bytes allocated are also down.
- When wall-time, cpu-time, and allocation profiling are enabled, then it's a clear win. CPU time spent stack walking is down ~55%, and allocated bytes are down ~35%.

In all cases, the total memory in use goes up a bit as the `OwnedStringTable` lives essentially forever in thread-local storage.

Some screenshots:

CPU time before/after. This is with allocation profiling enabled.

app datadoghq com_profiling_AYaeLMt7AABn13XN6ikcfQAA_query=service%3Asymfony-demo%20-runtime%3Azendengine event=AQAAAYaeLMoEJtXgqQAAAABBWWFlTE10N0FBQm4xM1hONmlrY2ZRQUE group_by=line my_code=disabled op_filter=show_from%28function%3A%22datad

app datadoghq com_profiling_AYaeLMZ8AABBkYx4H153YwAA_query=service%3Asymfony-demo%20-runtime%3Azendengine event=AQAAAYaeLMUJB0wRTgAAAABBWWFlTE1aOEFBQkJrWXg0SDE1M1l3QUE group_by=line my_code=disabled op_filter=show_from%28function%3A%22datad

Bytes allocated before/after. Again, with allocation profiling enabled.

app datadoghq com_profiling_AYaeE3pmAABn13XN6iIVegAA_query=runtime%3Anative%20service%3Asymfony-demo event=AQAAAYaeE3mCH9LgqQAAAABBWWFlRTNwbUFBQm4xM1hONmlJVmVnQUE group_by=line my_code=disabled op_filter=show_from%28function%3A%22datado (1)

app datadoghq com_profiling_AYaeE3fvAACx9iETN3_ewgAA_query=runtime%3Anative%20service%3Asymfony-demo event=AQAAAYaeE3dHRc9nFgAAAABBWWFlRTNmdkFBQ3g5aUVUTjNfZXdnQUE group_by=line my_code=disabled op_filter=show_from%28function%3A%22datado (1)

profiling/src/string_table/bump_owned.rs
```diff
@@ -0,0 +1,70 @@
+use bumpalo::{collections, Bump};
+use std::ops::Range;
+
+pub struct BorrowedStringTable<'b> {
+    arena: &'b Bump,
```
r1viollet, line 5: Interesting pattern, this comment is mostly trying to understand more.
Why is it better to do this (take a ref to an existing arena guaranteeing the lifetime (using `b`) rather than just have the arena within the object ?

profiling/src/profiling/stalk_walking.rs
```diff
@@ -2,14 +2,43 @@ use crate::bindings::{
...
+use std::cell::{RefCell, RefMut};
+use std::mem::transmute;
 use std::str::Utf8Error;
 
+/// Used to help track the function run_time_cache hit rate. It glosses over
+/// the fact that there are two cache slots used, and they don't have to be in
+/// sync. However, they usually are, so we simplify.
+#[cfg(php8)]
+#[derive(Debug, Default)]
+pub struct FunctionRunTimeCacheStats {
```
r1viollet, line 16: good idea having stats around

profiling/src/profiling/stalk_walking.rs
```diff
@@ -2,14 +2,43 @@ use crate::bindings::{
...
+}
+
 #[derive(Default, Debug)]
 pub struct ZendFrame {
     // Most tools don't like frames that don't have function names, so use a
-    // fake name if you need to like "<php>".
-    pub function: String,
-    pub file: Option<String>,
+    // fake name if you need to like "<?php".
+    pub function: Cow<'static, str>,
```
r1viollet, line 40: Mostly reading to see if I understand well: So in some cases we want to own the string and give it static lifetime (What are these cases?), otherwise we will use the borrowed lifetime from the cache.

profiling/src/php_ffi.c
```diff
@@ -105,7 +104,53 @@ void ddog_php_prof_zend_mm_set_custom_handlers(zend_mm_heap *heap,
...
-{
+zend_execute_data* ddog_php_prof_get_current_execute_data() {
     return EG(current_execute_data);
 }
+
+#if PHP_VERSION_ID >= 80000
+static int ddog_php_prof_run_time_cache_handle = -1;
+#endif
+
+void ddog_php_prof_function_run_time_cache_init(const char *module_name) {
```
r1viollet, line 115: I don't really understand if we had an alternative to having a C layer part. I'll come back to it.

profiling/src/php_ffi.c
```diff
@@ -105,7 +104,53 @@ void ddog_php_prof_zend_mm_set_custom_handlers(zend_mm_heap *heap,
...
+#if PHP_VERSION_ID >= 80000
+    // Grab 2, one for function name and one for filename.
+    ddog_php_prof_run_time_cache_handle =
+        zend_get_op_array_extension_handles(module_name, 2);
```
bwoebi, line 119: I added this API with 8.0.18 / 8.1.4. If you want to support older versions, i.e. 8.0.0-8.0.17 & 8.1.0-8.1.3, use zend_get_op_array_extension_handle() twice:
```suggestion
#if PHP_VERSION_ID < 80200
        zend_get_op_array_extension_handles(module_name);
    zend_get_op_array_extension_handles(module_name);
#else
        zend_get_op_array_extension_handles(module_name, 2);
#endif
```
which effectively has the same effect.

profiling/src/string_table/bump_owned.rs
morrisonlevi: Rust doesn't have a great self-referential story yet. That's why I built 3 structs (borrowed, bump-owned, and owned) so that this is done in safe Rust and the risky bits are in a well tested crate.

Maybe it can be done in a simpler way, but I didn't see it at the time.

profiling/src/php_ffi.c
morrisonlevi: In this case, the C macros aren't available to Rust code through bindgen, so I made a little wrapper to expose it. Additionally, Rust doesn't like global mutable state like `ddog_php_prof_run_time_cache_handle`, but in this case it's simple enough and we already have C code for the previous point.

profiling/src/php_ffi.c
morrisonlevi: Thanks for the heads ups!

profiling/src/php_ffi.c
r1viollet: Ah yes, you mentioned that! Thanks

profiling/src/string_table/owned.rs
```diff
@@ -0,0 +1,109 @@
...
+impl super::StringTable for OwnedStringTable {
+    #[inline]
+    fn len(&self) -> usize {
+        self.inner.with_dependent(|_arena, set| set.len())
+    }
+
+    #[inline]
+    fn insert_full(&mut self, str: &str) -> (usize, bool) {
+        self.inner
+            .with_dependent_mut(|_arena, set| set.insert_full(str))
```
realFlowControl, line 75: @morrisonlevi could you help me understand what `with_dependent_mut` and `with_dependent` are doing? I could not find any kind of documentation on this

profiling/src/string_table/borrowed.rs
```diff
@@ -0,0 +1,80 @@
+use ahash::RandomState;
+use std::collections::HashMap;
+use std::ops::Range;
+
+pub struct BorrowedStringTable<'a> {
+    pub vec: Vec<&'a str>,
+    pub map: HashMap<&'a str, usize, RandomState>,
```
realFlowControl, line 7: @morrisonlevi excuse my most likely super stupid question, but:

- `vec` holds references to `str` in the arena
- `map` uses the `&str` (the one also stored in `vec`) as key and the offset in `vec` as value

we do this because we do not want to / can't store pointers in the PHP_RUNTIME_CACHE, but offsets, correct?
Follow-up: why do we use a `usize`, If my assumption are correct, I would have defaulted to an `unsigned int`, just want to understand.

profiling/src/string_table/borrowed.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl<'a> Default for BorrowedStringTable<'a> {
+    fn default() -> Self {
+        /// The initial size of the Vec. At the time of writing, Vec would
+        /// choose size 4. This is expected to be much too small for the
+        /// use-case, so use a larger initial capacity to save a few
+        /// re-allocations in the beginning.
+        /// This is just an educated estimate, not a finely tuned value.
+        const INITIAL_VEC_CAPACITY: usize = 1024 / std::mem::size_of::<&str>();
```
realFlowControl, line 17: @morrisonlevi just for my understanding: assuming we compile for a 64 bit platform, `size_of::<&str>()` is 8 (64 bit), so we reserve initial memory for 128 string references, on a 32 bit platform we would initially reserve memory for 256 string references. Did you mean to pre alloc space for 1024 string references?

profiling/src/string_table/owned.rs
```diff
@@ -0,0 +1,109 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// If this fails, bumpalo may have changed its allocation patterns, and
+    /// [OwnedStringTable::new] may need adjusted.
+    #[test]
+    fn test_bump() {
+        let arena = Bump::with_capacity(4000);
+        assert_eq!(4096 - 64, arena.chunk_capacity());
```
realFlowControl, line 100: @morrisonlevi could you explain those numbers (why `- 64`) and why `OwnedStringTable::new` may need adjustment in this case?

profiling/src/string_table/bump_owned.rs
```diff
@@ -0,0 +1,70 @@
...
+impl<'b> super::StringTable for BorrowedStringTable<'b> {
+    #[inline]
+    fn len(&self) -> usize {
+        self.set.len()
+    }
+
+    fn insert_full(&mut self, str: &str) -> (usize, bool) {
+        match self.set.map.get(str) {
+            None => {
+                let owned = collections::String::from_str_in(str, self.arena);
```
realFlowControl, line 28: @morrisonlevi just for my understanding: this is where we actually store the string in the arena?

profiling/src/string_table/owned.rs
morrisonlevi: This is part of the `self_cell` crate:
https://docs.rs/self_cell/latest/self_cell/macro.self_cell.html
Basically, they are borrowing the "dependent" which is the borrowed string table for the duration of the closure. This is how they scope the lifetime of the borrow for safety.
These things are needed because self-referential structs (structs which internally have references to other parts of the same struct), are currently not supported in safe Rust.

profiling/src/string_table/borrowed.rs
morrisonlevi: Offsets in Rust are `usize`, so it's ergonomic to work with those. Serendipitously for this specific use case, the cache slot in the Zend Engine stores a pointer, and pointer-sized-integers are represented as `usize` in Rust. Make sense?

profiling/src/string_table/borrowed.rs
morrisonlevi: `sizeof::<&str>()` is actually 16 bytes; this is a "fat pointer." The capacity is therefore only 64 strings, and on 32-bit (which we don't support) it would be double that. The general goals were:

- Keep it within 1 page of memory.
- Reduce allocations during the startup, which already has a lot of memory churn.

Honestly, this `1024` number should probably be `2048` -- half a page of memory. Then the `HashMap` would use roughly the other half of the page (logically, not guaranteed by any allocator).

profiling/src/string_table/owned.rs
morrisonlevi: Said differently, if I ask Bump for `4000` bytes, I should get a number closer to `4096`.

profiling/src/string_table/bump_owned.rs
morrisonlevi: Yes, this is where the sting is actually being stored. Docs:
https://docs.rs/bumpalo/latest/bumpalo/collections/string/struct.String.html#method.from_str_in

> Construct a new String<'bump> from a string slice.

```rust
use bumpalo::{Bump, collections::String};

let b = Bump::new();

let s = String::from_str_in("hello", &b);
assert_eq!(s, "hello");
```

realFlowControl (approved):

PR #1951, opened 2023-03-01
feat: add config mode to `datadog-setup.php`

realFlowControl:
### Description

This PR will add a config mode to the `datadog-setup.php`, see PROF-6905

- adds an implementation of `getopt()` that allows for commands
- adds a `config get`, `config set` and `config list` command
- updates the help output

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (commented):
I can't finish the review right now, but it's looking good so far!

datadog-setup.php
```diff
@@ -58,6 +58,182 @@ function print_help()
...
+ * This function will print out all Datadog specific PHP INI settings. The list
+ * of Datadog specific settings is retrieved by calling `get_ini_settings()`. It
+ * will also show the default and INI file this setting was found.
+ * The output will be grouped by PHP binary, example:
+ *
+ * $ php datadog-setup.php config list --php-bin all
+ * Searching for available php binaries, this operation might take a while.
+ * Datadog configuration for binary: php (/opt/php/8.2/bin/php)
+ * datadog.profiling.enabled => On // default: 1, INI file: /opt/php/etc/conf.d/98-ddtrace.ini
+ * datadog.profiling.experimental_allocation_enabled => On // default: 1, INI file: /opt/php/etc/conf.d/98-ddtrace.ini
```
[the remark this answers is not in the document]
realFlowControl: Fair point, I had chosen this format as `php --ri=extension` does it with the `=>`, but your point is valid!

datadog-setup.php
```diff
@@ -1059,16 +1062,17 @@ function parse_cli_arguments(array $argv = null): array
...
+            if (strpos($key, '=') !== false) {
                 list($key, $value) = explode('=', $key, 2);
+            } else {
+                // look ahead to next $token
+                if (isset($argv[0]) && substr($argv[0], 0, 1) !== '-') {
+                    $value = array_shift($argv);
+                    if ($value === null) {
+                        $value = false;
+                    }
+                }
```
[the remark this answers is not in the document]
realFlowControl: This prompt currently leads to
```bash
$ php datadog-setup.php config get --php-bin -ddatadog.profiling.enabled
ERROR: Provided PHP binary '' was not found.
```
Could you elaborate on the `else`? The error is due to 'php-bin' argument being set to false, it must have a value, so I will add a check on the mandatory option to give a better error message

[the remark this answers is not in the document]
morrisonlevi: I tried a few variations and couldn't get it to fail, so I guess I must not have been thinking clearly.

datadog-setup.php
```diff
@@ -47,23 +71,337 @@ function print_help()
...
+                    echo "Warning: '$setting[0]' was found in multiple places, ",
+                        "you might want to remove duplicates.", PHP_EOL;
+                }
+                continue;
+            }
+
+            // If we are here, we could not find the INI setting in any files, so
+            // we try and look for commented versions
+
+            $iniFilePaths = find_ini_files($phpProps) + get_ini_files($phpProps);
```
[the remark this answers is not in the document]
realFlowControl: I cleaned up the multiple calls to the same function.
Regarding the naming and what those functions do:
- renamed to `find_all_ini_files()` and `find_main_ini_files()`
- updated the docs on each function to better show what they are doing
Both might look similar, but have different logic on what they should find and retrieve. Also, both functions do fit on one screen, so I think it is safe to leave them as is. WDYT?

datadog-setup.php
```diff
@@ -400,6 +701,118 @@ function ($el) {
...
+             *   - <...>/cli/conf.d       <-- we know this from php -i
+             *   - <...>/apache2/conf.d   <-- we derive this from relative path
+             *   - <...>/fpm/conf.d       <-- we derive this from relative path
+             */
+            $apacheConfd = str_replace('/cli/conf.d', '/apache2/conf.d', $phpProperties[INI_SCANDIR]);
+            $addIniFiles($apacheConfd);
+        }
+    }
+
+    if ($phpProperties[INI_MAIN] && is_file($phpProperties[INI_MAIN])) {
```
morrisonlevi, line 753: I got this error on one of my configurations:
> Warning: Undefined array key "Loaded Configuration File" in %s/datadog-setup.php on line 753

I assume this was supposed to have `isset()` around it or something?

datadog-setup.php
```diff
@@ -400,6 +701,118 @@ function ($el) {
...
+ *
+ * The `$phpProperties` can be retrieved by calling `ini_values($pathToPHPBinary)`.
+ *
+ * @see ini_values
+ * @return string[]
+ */
+function find_main_ini_files(array $phpProperties): array
+{
+    $iniFilePaths = [];
+    if ($phpProperties[INI_SCANDIR]) {
```
morrisonlevi, line 780: I assume an `isset` is needed here as well?

datadog-setup.php
```diff
@@ -400,6 +701,118 @@ function ($el) {
...
+        $iniFileName = '98-ddtrace.ini';
+        // Search for pre-existing files with extension = ddtrace.so to avoid conflicts
+        // See issue https://github.com/DataDog/dd-trace-php/issues/1833
+        foreach (scandir($phpProperties[INI_SCANDIR]) as $ini) {
```
morrisonlevi, line 784: If `scandir` fails, such as:

```
% php --ini
Configuration File (php.ini) Path: /opt/php/8.2/etc/php.ini
Loaded Configuration File:         (none)
Scan for additional .ini files in: /opt/php/8.2/etc/conf.d
Additional .ini files parsed:      (none)
```

And that directory `/opt/php/8.2/etc/conf.d` doesn't exist`, then we fail.

morrisonlevi: I added a few things to handle this better. If the directory doesn't exist, it will tell you to make it and try again.

datadog-setup.php
```diff
@@ -47,23 +71,353 @@ function print_help()
...
+    // Build an index by unique names for filtering.
+    $indexByName = array_column($iniSettings, null, 'name');
+
+    foreach (require_binaries_or_exit($options) as $command => $fullPath) {
+        $binaryForLog = ($command === $fullPath) ? $fullPath : "$command ($fullPath)";
+        echo "Datadog configuration for binary: $binaryForLog", PHP_EOL;
+
+        $iniFilePaths = find_all_ini_files(ini_values($fullPath));
+
+        foreach ($iniFilePaths as $iniFilePath) {
```
bwoebi, line 134: If multiple binaries point to the same ini, it'll dump the ini files from the same ini multiple times.
This needs to be two separate loops, with the second loop iterating over `array_unique($iniFilePaths);`

```suggestion
    $iniFilePaths = [];
    foreach (require_binaries_or_exit($options) as $command => $fullPath) {
        $binaryForLog = ($command === $fullPath) ? $fullPath : "$command ($fullPath)";
        echo "Datadog configuration for binary: $binaryForLog", PHP_EOL;

        array_push($iniFilePaths, ...find_all_ini_files(ini_values($fullPath));
    }

    foreach (array_unique($iniFilePaths) as $iniFilePath) {
```

datadog-setup.php
```diff
@@ -47,23 +71,353 @@ function print_help()
...
+function cmd_config_list(array $options)
+{
+    foreach (config_list($options) as $record) {
+        echo "$record->setting = $record->currentValue ; default: $record->defaultValue, INI file: $record->iniFile\n";
+    }
+}
+
+/**
+ * This function will print the specified PHP INI settings. It only prints
+ * information for Datadog's own settings; it can't be used to parse opache,
```
bwoebi, line 175:
```suggestion
 * information for Datadog's own settings; it can't be used to parse opcache,
```

datadog-setup.php
```diff
@@ -47,23 +71,353 @@ function print_help()
...
+ * xdebug, etc.
+ * The output will be grouped by PHP binary, example:
```
bwoebi, line 177: It's grouped by INI, really, not saying which binary (which is fine), right?

morrisonlevi: There is an echo in `config_list` which prints out the binary. The separation isn't clean as it probably ought to be. I wrote that when unifying config get and config list a bit.

bwoebi: Oh. Bit awkward with the yield & echo mix :-D

realFlowControl: I'm about to check this, as with multiple binaries it gives
```bash
php datadog-setup.php config get -d datadog.profiling.enabled --php-bin /opt/php/8.2/bin/php --php-bin /opt/php/8.1/bin/php
Datadog configuration for binary: /opt/php/8.2/bin/php
Datadog configuration for binary: /opt/php/8.1/bin/php
datadog.profiling.enabled = On ; default: 1, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
```
where it should be
```bash
php datadog-setup.php config get -d datadog.profiling.enabled --php-bin /opt/php/8.2/bin/php --php-bin /opt/php/8.1/bin/php
Datadog configuration for binary: /opt/php/8.2/bin/php
datadog.profiling.enabled = On ; default: 1, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
Datadog configuration for binary: /opt/php/8.1/bin/php
```

datadog-setup.php
realFlowControl: This would change the output for `php datadog-setup.php config list --php-bin /opt/php/8.2/bin/php --php-bin /opt/php/8.1/bin/php` from:
```bash
datadog.profiling.enabled = On; default: 1, binary: /opt/php/8.2/bin/php, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
datadog.profiling.enabled = On; default: 1, binary: /opt/php/8.1/bin/php, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
```
to
```bash
datadog.profiling.enabled = On; default: 1, binary: /opt/php/8.2/bin/php, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
```
This is also correct, but I would argue that the first output makes it more explicit what happened:
- both PHP binaries are using the same INI file
- in both PHP binaries the INI setting is set to a certain value

It doubles the amount of lines shown, but it is the more explicit approach. If we `array_unique()` the INI files, we should present this information (that multiple binaries share an INI file) somehow to the user.

An example could be something like this:
```
Hint: The INI file /opt/php/8.2/etc/conf.d/98-ddtrace.ini is used for the PHP binaries /opt/php/8.2/bin/php and /opt/php/8.1/bin/php.
datadog.profiling.enabled = On; default: 1, binary: /opt/php/8.2/bin/php, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
```

This could also make it explicit what happened. I think that if we do not somehow point out this special case, people might get confused about missing INI settings for certain PHP binaries. WDYT?

datadog-setup.php
realFlowControl: I moved the PHP binary to the `IniRecord` class, so that it can be printed out with the record itself.

realFlowControl: The output should now look like this:
```bash
$ php datadog-setup.php config get -d datadog.profiling.enabled --php-bin /opt/php/8.2/bin/php --php-bin /opt/php/8.1/bin/php
datadog.profiling.enabled = On; default: 1, binary: /opt/php/8.2/bin/php, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
datadog.profiling.enabled = On; default: 1, binary: /opt/php/8.1/bin/php, INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
```

datadog-setup.php
realFlowControl: @bwoebi If this is okay for you, do you see something else blocking this PR?

datadog-setup.php
```diff
@@ -47,23 +71,357 @@ function print_help()
...
+    }
+    return $setting;
+}
+
+
+/**
+ * This function will try and update a given `$setting` in the INI file given in
+ * `$iniFile`. First it tries to find an uncommented version of the setting in
+ * the file and replace this with the new value. In case no uncommented version
+ * was found it tries to find commented versions of this INI setting.
```
bwoebi, line 393: That's not true, right?
If I read the function code correctly, it either replaces the first match, commented or not, or if $promoteComment is false, it replaces all occurrences of that INI.

bwoebi: Also note that, if any directive happens to be duplicate, the last occurrence of that directive is relevant. It does not make sense to only replace the first, esp. in case there is a commented directive, followed by an uncommented one.

bwoebi: Ah, I see, the caller site of update_ini_setting is ensuring it's first called without $promoteComment, then with it.
I guess in that case it's fine.

I was just looking at the function itself.

realFlowControl: Nevertheless, I will test this behaviour again to make sure there is no blind spot, thanks for the tip!

bwoebi (approved):
I think the PR does a good job at handling all the possible cases :-)

PR #1952, opened 2023-03-02
perf(profiling): optimize prepare_sample_message

morrisonlevi:
### Description

This function is called once per stack walk, so it's on the hot path but not as important to optimize as walking the stack itself.

However, it was showing up more than I would have expected in my profiles when examining it with the native profiler. Unfortunately I didn't have granular feedback on what was slowing it down, so I did some general optimization techniques:

 1. Statically store the sample types as an array instead of creating them on-demand.
 2. Replace Cow<'static, str> with &'static str in ValueType. All usages were already static strings, and I don't see this changing.
 3. Simplify the branching logic into slicing operations.
 4. Move tag creation to rinit, and share them with `Arc<Vec<Tag>>` instead of copying them. This makes the work to do on each stack sample be a simple atomic refcount increment. This will also reduce the number of total copies, which is also good.
 5. Initialize the sample's labels to the right size, and stop copying them unconditionally just for `trace` level diagnostics.

Aside from performance, it also adds `ValueType::new` to shorten up consecutive constructions (which happen frequently with this type). Between this and the changes to remove `Cow`, code is simplified:

```rust
    // before
    ValueType {
        r#type: Cow::Borrowed("sample"),
        unit: Cow::Borrowed("count"),
    },
    ValueType {
        r#type: Cow::Borrowed("wall-time"),
        unit: Cow::Borrowed("nanoseconds"),
    },

    // after
    ValueType::new("sample", "count"),
    ValueType::new("wall-time", "nanoseconds"),
```

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -565,23 +565,23 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
                 // later.
                 let mut tags = (*locals.tags).clone();
 
-                local_add_optional_tag(&mut tags, "service", &locals.service);
-                local_add_optional_tag(&mut tags, "env", &locals.env);
-                local_add_optional_tag(&mut tags, "version", &locals.version);
+                add_optional_tag(&mut tags, "service", &locals.service);
+                add_optional_tag(&mut tags, "env", &locals.env);
+                add_optional_tag(&mut tags, "version", &locals.version);
 
                 let runtime_id = runtime_id();
                 if !runtime_id.is_nil() {
-                    local_add_tag(&mut tags, "runtime-id", &runtime_id.to_string());
+                    add_tag(&mut tags, "runtime-id", &runtime_id.to_string());
                 }
 
                 if let Some(version) = PHP_VERSION.get() {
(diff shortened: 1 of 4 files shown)


morrisonlevi:
Numbers from ddprof for default profile types and how much overhead prepare_sample_message has within `collect_time`:

- Old: 12/101 = 0.1188
- New: 9/93 = 0.0968

This was for a given endpoint in the symfony-demo under `ab` load for 560-ish seconds. The win is small, but real.

realFlowControl (approved):
🚀

morrisonlevi:
I added another optimization by switching tags from `Vec<Tag>` to `Arc<Vec<Tag>>` and just sharing them for every message (they don't change within a request). Stats :
- Old: 5/46 = 0.1087
- New (cumulative): 2/32 = 0.0625

It's a bigger win, but still overall a small win.

morrisonlevi:
I made a small improvement to `message_labels`. It's not worth measuring, I just noticed it while driving by (it's the code right above `prepare_sample_message`).

PR #1954, opened 2023-03-02
Slightly reword the exception message to use Caught and Uncaught only at top-level

bwoebi:
### Description

This should reduce confusion about exception thrown at function boundaries: it's only at top-level that we actually known about completely uncaught.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/serializer.c
```diff
@@ -664,6 +674,7 @@ void ddtrace_set_root_span_properties(ddtrace_span_data *span) {
 
 static void _serialize_meta(zval *el, ddtrace_span_data *span) {
     bool top_level_span = span->parent_id == DDTRACE_G(distributed_parent_trace_id);
+    bool local_root_span = span->parent_id == 0 || top_level_span;
```
pierotibou, line 677: `is_local_root_span` would make it more understandable

pierotibou (approved):
Looks good functionally speaking.

realFlowControl (approved):
There is one open comment from @pierotibou which I would also support, but besides that it's looking good

PR #1959, opened 2023-03-09
Use ZEND_TLS/TSRM_TLS macros instead of __thread to only use thread local storage in ZTS builds

### LINKED ISSUE #1957: SIGSEGV when used with ilimit
(Not using the bug or feature-request templates because this falls very much in between the two...)

Hi,

we use https://github.com/krakjoe/ilimit to implement timeouts for a C library (which we don't control) that does sync network IO.

When the timeout is hit, we get a SIGSEGV:

```
(omitted)
#4  <signal handler called>
#5  0x000055e77d5ce52b in zend_hash_index_find_bucket (h=h@entry=35423727, ht=0x7fc6dc00f4c0) at /build/downloads/php-8.1.7/Zend/zend_hash.c:713
#6  _zend_hash_index_find (ht=0x7fc6dc00f4c0, h=h@entry=35423727) at /build/downloads/php-8.1.7/Zend/zend_hash.c:2323
#7  0x00007fc6eca4d5a4 in zai_hook_table_find (found=<synthetic pointer>, index=<optimized out>, table=<optimized out>) at /home/circleci/datadog/tmp/build_extension/zend_abstract_interface/hook/table.h:14
#8  zai_hook_continue (ex=ex@entry=0x7fc6f4015900, memory=memory@entry=0x7fc6e65fe8b0) at /home/circleci/datadog/tmp/build_extension/zend_abstract_interface/hook/hook.c:398
#9  0x00007fc6eca47de3 in zai_interceptor_observer_begin_handler (execute_data=0x7fc6f4015900) at /home/circleci/datadog/tmp/build_extension/zend_abstract_interface/interceptor/php8/interceptor.c:123
#10 0x000055e77d64a0d1 in _zend_observe_fcall_begin (execute_data=0x7fc6f4015900) at /build/downloads/php-8.1.7/Zend/zend_observer.c:170
#11 0x000055e77d64a345 in zend_observer_fcall_begin (execute_data=<optimized out>) at /build/downloads/php-8.1.7/Zend/zend_observer.c:183
#12 0x000055e77d5ad785 in zend_call_function (fci=fci@entry=0x7ffc194c12f0, fci_cache=fci_cache@entry=0x7ffc194c1330) at /build/downloads/php-8.1.7/Zend/zend_execute_API.c:907
#13 0x00007fc6eccc5fdb in __php_ilimit_call_thread (arg=0x7fc6e65fe9f0) at /build/downloads/ilimit/src/ilimit.c:179
#14 0x00007fc6f6846ea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
#15 0x00007fc6f69b6def in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

The null pointer is in this hashtable:

https://github.com/DataDog/dd-trace-php/blob/0.84.0/zend_abstract_interface/hook/hook.c#L44-L46

By the time the dd-trace hook on zend_call_function runs, the hashtable looks like this:

```
#8  zai_hook_continue (ex=ex@entry=0x7fc6f4015900, memory=memory@entry=0x7fc6e65fe8b0) at /home/circleci/datadog/tmp/build_extension/zend_abstract_interface/hook/hook.c:398
398	    if (!zai_hook_table_find(&zai_hook_resolved, zai_hook_frame_address(ex), (void**)&hooks)) {
(gdb) p zai_hook_resolved
$4 = {gc = {refcount = 0, u = {type_info = 0}}, u = {v = {flags = 0 '\000', _unused = 0 '\000', nIteratorsCount = 0 '\000', _unused2 = 0 '\000'}, flags = 0}, nTableMask = 0, arData = 0x0, nNumUsed = 0, nNumOfElements = 0, nTableSize = 0, nInternalPointer = 0, nNextFreeElement = 0, pDestructor = 0x0}
```

(`arData` is what gets dereferenced later in `zend_hash_index_find_bucket`)

which tells me that this is indeed running on another thread, where the init code https://github.com/DataDog/dd-trace-php/blob/0.84.0/zend_abstract_interface/hook/hook.c#L550-L559
has not run.

If I read this right, the problem here is ilimit, which calls zend_call_function from another thread. But this is probably within its right, since by design it only works with a non-threaded php runtime and explicitly states so.

What do you think, would it be possible in any way for dd-trace to detect this and not crash? I could absolutely live with not having any tracing for requests that hit this path.

In the future we will probably have to solve this in a different way, but for now I'd love to be able to continue to run dd-trace without the segfaults.

Thanks for your feedback!

````
php --version
PHP 8.1.7 (cli) (built: Jun 10 2022 09:40:09) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.7, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.7, Copyright (c), by Zend Technologies
    with ddtrace v0.84.0, Copyright Datadog, by Datadog
    with datadog-profiling v0.13.0, Copyright Datadog, by Datadog
    with ddappsec v0.6.0, Copyright Datadog, by Datadog
```

bwoebi:
### Description

Fixes #1957.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/hook/uhook.c
+++ b/ext/hook/uhook.c
@@ -21,8 +21,8 @@ extern void (*profiling_interrupt_function)(zend_execute_data *);
 
 zend_class_entry *ddtrace_hook_data_ce;
 
-static __thread HashTable dd_closure_hooks;
-static __thread HashTable dd_active_hooks;
+ZEND_TLS HashTable dd_closure_hooks;
+ZEND_TLS HashTable dd_active_hooks;
 
 typedef struct {
     size_t size;
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -439,6 +439,7 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
         profiling_experimental_cpu_time_enabled,
         profiling_experimental_allocation_enabled,
         log_level,
+        output_pprof,
     ) = unsafe {
         (
             config::profiling_enabled(),
@@ -446,6 +447,7 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
             config::profiling_experimental_cpu_time_enabled(),
             config::profiling_experimental_allocation_enabled(),
             config::profiling_log_level(),
+            config::profiling_output_pprof(),
         )
     };
 
@@ -543,7 +545,7 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
          */
         let mut profiler = PROFILER.lock().unwrap();
         if profiler.is_none() {
-            *profiler = Some(Profiler::new())
+            *profiler = Some(Profiler::new(output_pprof))
         }
     };
 
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -391,7 +391,7 @@ pub struct UploadMessage {
 }
 
 impl Profiler {
-    pub fn new() -> Self {
+    pub fn new(output_pprof: Option<Cow<'static, str>>) -> Self {
         let fork_barrier = Arc::new(Barrier::new(3));
         let (fork_sender0, fork_receiver0) = crossbeam_channel::bounded(1);
         let interrupt_manager = Arc::new(InterruptManager::new(Mutex::new(Vec::with_capacity(1))));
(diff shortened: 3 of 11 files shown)


morrisonlevi (commented):
The profiler doesn't have such an easy time switching. What exactly is the issue? Is this going to be a worry for the profiler?

bwoebi:
The issue is that extensions, which require NTS, can use a thread to interact with php from that other thread, but will run into uninitialized values from that other thread. With NTS everything is supposed to be global. This PR ensures that.

So, no, the profiler, just supporting NTS, has nothing to worry about.

morrisonlevi (approved):

morrisonlevi:
Just in case it causes issues, I pushed a commit that removed the profiler's single use of config from another thread. It now fetches it in the PHP thread and passes it along, which is better anyway.

r1viollet (commented):
LGTM! This seems safer indeed

Issue #1958, opened 2023-03-09
[Feature] Per-request configuration without environment variables


brainlock:
Hi,

in the DataDog docs I read that the recommended way to configure ddtrace-php is via environment variables.

In our setup though, we use each pool of php-fpm processes to serve requests for multiple customers, for multiple versions of our backend, and even for different environments (prod/stage/test). Environment variables configured on the fpm pool won't cut it in this situation.

In order to tag each trace with the correct values, I've been using the php api at the beginning of the request to annotate the root span:

```
$span = \DDTrace\root_span();
if ($span) {
    $span->service = '…';
    $span->type = 'web';
    $span->meta['branch'] = …;
    $span->meta['version'] = …;
    $span->meta['commit'] = …;
    $span->meta['site_name'] = …;
    $span->meta['env'] = …;
}
```

and then later, when the routing has found the endpoint which will handle the request, to annotate the active span:

```
$span = \DDTrace\active_span();
if ($span) {
    $span->resource = $endpoint_name;
}
```

This seems to work well for the traces, I can add the custom fields as columns, filter, facets.

The profiles though aren't annotated with this information: each profile appears to be under the default service name of "web.request", and I don't see any of the other fields (env, version…). It does seem to pick up the endpoint name though, the one I'm setting on the active span->resource.

It would be great if we could also filter the profiles by the fields above.

Is this something you plan to support at some point? Or am I missing something and there is another obvious way of doing this?

(Another way I could pass some of the parameters is via fastcgi parameters, because many of the routing decisions have been made by that point, but not all. Ideally the per-request configuration via the php API would also be picked up in the profiles.)

Thanks for your feedback!

Alberto

bwoebi:
Hey @brainlock,

As described in our docs https://docs.datadoghq.com/tracing/trace_collection/library_config/php/#environment-variable-configuration, each env var has its corresponding ini.

So, in your case with php-fpm, you can add an [.user.ini](https://www.php.net/manual/en/configuration.file.per-user.php) for each target. (Assuming your routing is to different files)

Obviously, if it's fully dynamic, you might have to still do some configuration in code.

bwoebi:
Also, I am noticing that we always fetch version and env from the ini for each span. Do you think it would make sense to fetch it from the parent span if one is available? That should be probably better for your use case too.

bwoebi:
I did miss that you were talking about the profiler and read "profile" as "the profile of a trace".

[...]

PR #1960, opened 2023-03-09
Inherit version, env and _dd.origin from parent span

bwoebi:
### Description

Fetch things from parent span in order to ensure consistency of global tags, if the root spans metadata is changed after start.

Relates to #1958.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/serializer.c
+++ b/ext/serializer.c
@@ -407,24 +407,6 @@ static void dd_add_header_to_meta(zend_array *meta, const char *type, zend_strin
 
 void ddtrace_set_global_span_properties(ddtrace_span_data *span) {
     zend_array *meta = ddtrace_spandata_property_meta(span);
-    zval value;
-
-    zend_string *version = get_DD_VERSION();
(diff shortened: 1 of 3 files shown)


morrisonlevi (commented):
If service is not explicitly set, it should also be inherited, I think?

bwoebi:
@morrisonlevi that's already the case, a couple lines above the newly added code.

morrisonlevi:
I think it sets it based on `DD_SERVICE`, not on the parent's value? IMO, that should be the second fallback. I could have also misread this.

bwoebi:
@morrisonlevi 
https://github.com/DataDog/dd-trace-php/blob/741b875db09136e6cf6c1f87fdf2ad5ceecd7609/ext/span.c#L188-L190

Not sure, but it just copies from parent to new span?

morrisonlevi (approved):

PR #1964, opened 2023-03-13
Cleanup _asan usages in Makefile

bwoebi:
### Description

Instead recognize libasan in the active php binary. "ASAN=" to disable, "ASAN=1" to enable explicitly.

I noticed that even though I have debug-zts-asan in my container, I usually did not have the extension itself built with asan, which sort of defeated the point mostly, given that I've not been adding `_asan` to my commands all the time.

Also optimizing Makefile to only do a `composer update` on missing lockfile or lockfile for different PHP version being installed when running `make test_*` commands.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/zend_abstract_interface/interceptor/php8/interceptor.c
+++ b/zend_abstract_interface/interceptor/php8/interceptor.c
@@ -56,7 +56,7 @@ static void zai_hook_safe_finish(zend_execute_data *execute_data, zval *retval, 
     const size_t stack_size = 1 << 17;
     void *volatile stack = malloc(stack_size);
     if (SETJMP(target) == 0) {
-        void *stacktop = stack + stack_size, *stacktarget = stacktop - 0x400;
+        void *stacktop = stack + stack_size, *stacktarget = stacktop - 0x800;
 
 #ifdef __SANITIZE_ADDRESS__
         void *volatile fake_stack;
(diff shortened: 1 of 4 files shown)


zend_abstract_interface/interceptor/php8/interceptor.c
```diff
@@ -39,6 +39,9 @@ static inline bool zai_hook_memory_table_del(zend_execute_data *index) {
 }
 
 #if defined(__x86_64__) || defined(__aarch64__)
+# if defined(__GNUC__) && !defined(__clang__)
+__attribute__((no_sanitize_address))
+# endif
```
morrisonlevi, line 44: GCC's asan doesn't handle this part right, or something?

morrisonlevi (approved):
Looks good to me, as long as CI passes. Just have one question that doesn't block merging.

zend_abstract_interface/interceptor/php8/interceptor.c
bwoebi: Yep. At least some versions of GCC. For instance, I can't get it to repro within my buster container. It seems to put some things onto the stack and expects them to be still there after the stack base swap.

PR #1965, opened 2023-03-13
Add DD_TRACE_HOOK_LIMIT

### LINKED ISSUE #1956: [BUG] Memory leak. Memory increases every time trace_method(...) is called
### Bug description
Hi,

We have been fighting memory leaks in one of our Kafka consumer commands and yesterday we found a cause. 

Every time we consume a new message from Kafka, we start a tracing like this:

```php
        $tracer = new Tracer();
        GlobalTracer::set($tracer);

        $resourceName = $topicName . '::process';

        $options = ['start_time' => Time::now()];
        $rootSpan = $tracer->startRootSpan(
            'Jobs',
            $options
        )
            ->getSpan();

        $rootSpan->setTag(Tag::SPAN_TYPE, Type::CLI);
        $rootSpan->setTag(Tag::SERVICE_NAME, 'broker.service-x');
        $rootSpan->setTag(Tag::RESOURCE_NAME, $resourceName, true);
        $rootSpan->setTag('command_origin', $this->signature);

        // we add tracing for whatever classes we want to trace
        trace_method(Producer::class, 'produce', function (SpanData $span) {
            $span->service = 'broker.offer-service.kafka';
        });
       // ... many more
```

That means that if we consume 10 000 messages in a minute, we will execute this pseudo piece of code 10 000 times. After this block, we added a simple log that prints out current memory usage. **It was increasing every time we executed this block**.

Once we identified that the memory leak is somewhere here, we started removing some lines of code to find out which ones are causing it and we found it. 

**Every time we call *trace_method(...)* MEM usage increases.**

I would expect that the SDK would already make sure that if I call the same trace_method(*) multiple times, **it would ignore all subsequent calls since it sees it as a duplicate**.

### PHP version
8.1

### Tracer version
0.84.0

### OS info
NAME="Alpine Linux"
VERSION_ID=3.17.2
PRETTY_NAME="Alpine Linux v3.17"

bwoebi:
### Description

The primary motivation is users mistakenly installing hooks in a loop and observing memory leaking / increasing overhead.

Closes #1956.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/configuration.h
+++ b/ext/configuration.h
@@ -142,6 +142,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(BOOL, DD_TRACE_CLIENT_IP_ENABLED, "false")                                                          \
     CONFIG(STRING, DD_TRACE_CLIENT_IP_HEADER, "")                                                              \
     CONFIG(BOOL, DD_TRACE_FORKED_PROCESS, "true")                                                              \
+    CONFIG(INT, DD_TRACE_HOOK_LIMIT, "100")                                                                    \
     CONFIG(INT, DD_TRACE_AGENT_MAX_PAYLOAD_SIZE, "52428800", .ini_change = zai_config_system_ini_change)       \
     CONFIG(INT, DD_TRACE_AGENT_STACK_INITIAL_SIZE, "131072", .ini_change = zai_config_system_ini_change)       \
     CONFIG(INT, DD_TRACE_AGENT_STACK_BACKLOG, "12", .ini_change = zai_config_system_ini_change)                \
--- a/ext/hook/uhook.c
+++ b/ext/hook/uhook.c
@@ -350,11 +350,23 @@ PHP_FUNCTION(DDTrace_install_hook) {
     }
     def->id = -1;
 
+    uint32_t hook_limit = get_DD_TRACE_HOOK_LIMIT();
+
     zend_long id;
     if (resolved) {
         def->resolved = resolved;
         def->function = NULL;
         def->closure = closure;
+
+        if (hook_limit > 0 && zai_hook_count_resolved(resolved) >= hook_limit) {
+            ddtrace_log_onceerrf(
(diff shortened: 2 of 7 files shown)


ext/configuration.h
```diff
@@ -142,6 +142,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(BOOL, DD_TRACE_CLIENT_IP_ENABLED, "false")                                                          \
     CONFIG(STRING, DD_TRACE_CLIENT_IP_HEADER, "")                                                              \
     CONFIG(BOOL, DD_TRACE_FORKED_PROCESS, "true")                                                              \
+    CONFIG(INT, DD_TRACE_HOOK_LIMIT, "100")                                                                   \
```
realFlowControl, line 145: Nitpick: there is a space missing in this line, so the `\` in the end don't line up currently
```suggestion
    CONFIG(INT, DD_TRACE_HOOK_LIMIT, "100")                                                                    \
```

realFlowControl (approved):

guillaumepeano:
Hi, since we have updated our ddtrace-php version to 0.85.0 including this fix, we are experiencing memory limit exceeded issues in many jobs because of too many logs "Could not add hook ..." are occuring. To fix this, we override DD_TRACE_HOOK_LIMIT value in our script commands (DD_TRACE_HOOK_LIMIT=200 ourcommand.php) but I don't think it's a right way to do.

Could you tell me what's in our code is increasing the number of hooks ? Is overring value the only way ? Thank you.

Technical infos : 

Symfony 2.8+
PHP 7+
version dd-trace-php : https://github.com/DataDog/dd-trace-php/releases/tag/0.92.2

PROFeNoM:
Hi @guillaumepeano :wave:

An ongoing [issue](https://github.com/DataDog/dd-trace-php/issues/2427) relates to what you're saying right now. 

Is this `"Could not add hook ..."` occurring on the controller as well?

(Note that I'd preferably prefer that you answer on the issue, for better visibility/tracking)

Thanks

Issue #1740, opened 2022-09-12
[Feature] HTTP POST Data and headers


kissifrot:
**Is your feature request related to a problem? Please describe.**
Currently we don't have POST data when span concerns POST requests, which makes debugging sometimes quite troublesome, and that's something we miss in Datadog compared to tools like Sentry.

**Describe the solution you'd like**
Have a breakdown of POSTed variables as well as request headers in the Info tab.

**Describe alternatives you've considered**
Write a listener to add tags in symfony (as it's the main target), bit it's time consuming and not that optimized (and also out of scope)

**Additional context**
(n/a)

bwoebi:
Hey @kissifrot,

Are you looking at full POST data (also values, which may or may not contain sensitive information), or only the information whether some fields have been set (i.e. keys only)?

Headers can already be sent via `DD_TRACE_HEADER_TAGS` as a comma-separated whitelist of header fields you'd like to see.

kissifrot:
Well goal for eventually debugging is having the POST values along with their keys.
But you're right, it causes the risk of getting PII in them.
Maybe there could be a blacklist or a preprocess option to replace some values by `[REDACTTED]` when keys contains expression or regexes such as '*password*' or something equivalent.

PR #1971, opened 2023-03-15
[Feature] HTTP POST Data and headers

PROFeNoM:
### Description

Related to Issue #1740. Provides a breakdown of posted variables

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/serializer.c
```diff
@@ -651,6 +731,18 @@ void ddtrace_set_root_span_properties(ddtrace_span_data *span) {
         ZEND_HASH_FOREACH_END();
     }
 
+    if (getenv("DD_TRACE_HTTP_POST_DATA_PARAM_ALLOWED")
```
bwoebi, line 734:
```suggestion
    if (zend_hash_num_elements(get_DD_TRACE_HTTP_POST_DATA_PARAM_ALLOWED())
```
env is not necessarily set. Don't use getenv().

ext/serializer.c
```diff
@@ -405,6 +405,86 @@ static void dd_add_header_to_meta(zend_array *meta, const char *type, zend_strin
     }
 }
 
+static void replace_non_alpha_with_underscores(zend_string *str) {
+    for (char *ptr = ZSTR_VAL(str); *ptr; ++ptr) {
+        if ((*ptr < 'a' || *ptr > 'z') && (*ptr < 'A' || *ptr > 'Z')) {
```
bwoebi, line 410: don't replace `-` either.

src/Integrations/Util/Normalizer.php
```diff
@@ -200,6 +200,67 @@ public static function sanitizedQueryString()
...
+        return trim(preg_replace('/[^a-zA-Z0-9.\_]+/', '-', $noSpaces), '- ');
+    }
+
+    private static function generateFilteredPostFields(string $postKey, $postVal, array $whitelist): array
+    {
+        if (is_array($postVal)) {
+            $filteredPostFields = [];
+            foreach ($postVal as $key => $val) {
+                $key = self::normalizeString($key);
+                // If the current postkey is the empty string, we don't want ot add a '.' to the beginning of the key
```
bwoebi, line 215:
```suggestion
                // If the current postkey is the empty string, we don't want to add a '.' to the beginning of the key
```

ext/serializer.c
bwoebi: And numbers are fine too.

ext/serializer.c
```diff
@@ -405,6 +405,90 @@ static void dd_add_header_to_meta(zend_array *meta, const char *type, zend_strin
...
+    zend_string_release(posttag);
+}
+
+static void dd_add_post_fields_to_meta_recursive(zend_array *meta, const char *type, zend_string *postkey,
+                                                 zval *postval, zend_array* post_whitelist) {
+    if (Z_TYPE_P(postval) == IS_ARRAY) {
+        zend_string *key;
+        zval *val;
+
+        ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARRVAL_P(postval), key, val) {
```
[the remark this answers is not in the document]
bwoebi: I.e. `foo[]=a&foo[]=b` should result in keys `http.request.post.foo.0` and `http.request.post.foo.1`

tests/ext/root_span_url_with_post_array_allowed.phpt
```diff
@@ -0,0 +1,25 @@
+--TEST--
+Values of a an array are redacted when the array key is in the DD_TRACE_HTTP_POST_DATA_PARAM_ALLOWED env var
```
bwoebi, line 2: This doesn't really makes sense, you don't want to redact these, basically if the prefix matches exactly, then we just submit it.

bwoebi (approved):
Thanks, looks good now :-)

PR #1972, opened 2023-03-15
Add test agent to CI

wconti27:
### Description

This PR adds the test agent as a docker container to the tests and sets the tracer URL to that of the Test Agent. All traces are handled by the Test Agent and proxied to the original test endpoint, the request-replayer, once received.

## Motivation
We're currently working on integrating the test agent into each tracer's test suites, and for the test agent to handle all traces produced by integrations during testing. Within the Test-Agent, we are then creating checks for Service Naming, Unified Naming Conventions tagging, etc in order to be a centralized location for asserting on integration traces. Allows us to leverage traces produced from integrations during testing.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


.circleci/config.yml
```diff
@@ -56,6 +53,19 @@ aliases:
     environment:
       DD_REQUEST_DUMPER_FILE: dump.json
 
+  - &IMAGE_DOCKER_TEST_AGENT
+    image: williamconti549/dd-apm-test-agent:latest
```
pierotibou, line 57: Why not use the "official" test agent?

.circleci/config.yml
```diff
@@ -113,10 +123,10 @@ aliases:
       name: Disable Xdebug for built-in web tests
       command: test -e /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini && sudo rm -f $_ || true
 
-  - &STEP_WAIT_AGENT
-    run:
-      name: Waiting for Dockerized agent
-      command: wait-for 127.0.0.1:8126 --timeout=60
+  # - &STEP_WAIT_AGENT
+  #   run:
```
pierotibou, line 127: Can you remove instead of commenting?

.circleci/config.yml
```diff
@@ -128,6 +138,23 @@ aliases:
       name: Waiting for Dockerized request replayer
       command: wait-for request-replayer:80 --timeout=120
 
+  - &STEP_WAIT_TEST_AGENT
+    run:
+      name: Waiting for Dockerized APM Test Agent
+      command: wait-for test-agent:9126 --timeout=120
```
pierotibou, line 144: Do we really need a 2 minutes timeout ? We're starting to chase unnecessary wait time to reduce ci cost

.circleci/config.yml
```diff
@@ -128,6 +138,23 @@ aliases:
...
+
+  - &STEP_DOCKER_LOGS
+    run:
+      name: Outputting Docker Logs for APM Test Agent
+      command: docker logs -f test-agent
+      background: true
+
+  - &STEP_RUN_TEST_AGENT_SPAN_VALIDATION
+    run:
+      name: Getting APM Test Agent Span Validation Results
```
pierotibou, line 154: Do you have more details/doc on this one?

.circleci/config.yml
pierotibou: Oh I saw the description.  So can we wait for the PR to be merged on the test agent side ?

wconti27: yes exactly, had to make changes so I need to host my own image

.circleci/config.yml
wconti27: removing, future work

.circleci/config.yml
wconti27: reduced this to 30 seconds

.circleci/config.yml
wconti27: I've switched this back to use the official image, ready for review.

bwoebi:
Looking at this PR, I'm wondering what the motivation was, to have the agent forward to the request-replayer, instead of transparently handling it in the request-replayer, to forward traces to the test agent, given the request-replayer anyway gets all the traces already.

wconti27:
> Looking at this PR, I'm wondering what the motivation was, to have the agent forward to the request-replayer, instead of transparently handling it in the request-replayer, to forward traces to the test agent, given the request-replayer anyway gets all the traces already.
> 
> (Not a blocker, just curious.)

The primary reason for having the Test-Agent receive the traces first was mainly because of ease of implementation. The test agent already has proxying ability enabled (in order to send traces to an actual DD Agent), so making the change to proxy to the request replayer was simple. Also, doing it this way gives us the option to remove the request-replayer in the future (if desired) since the Test Agent **should** contain all functionality currently within the request-replayer.

PR #1976, opened 2023-03-17
Allow for hooking of extended internal classes

bwoebi:
### Description

In particular, test for \Redis being able to be extended.

Note: I discovered that we also don't hook anon classes properly (that's why they were commented out in the test). This is to be addressed in a future PR.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/ext/sandbox/hook_method/hook_extended_internal_class.phpt
+++ b/tests/ext/sandbox/hook_method/hook_extended_internal_class.phpt
@@ -1,0 +1,58 @@
+--TEST--
+Extended internal classes are hookable
+--FILE--
+<?php
+
+DDTrace\hook_method('Countable', 'count', function () {
+    echo " Countable";
+});
+
+DDTrace\hook_method('ArrayObject', 'count', function () {
+    echo " ArrayObject";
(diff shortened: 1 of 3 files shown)


tests/ext/sandbox/hook_method/hook_extended_internal_class.phpt
```diff
@@ -0,0 +1,58 @@
...
+dd_untrace('count', 'ArrayObject');
+
+echo "\nNormal: Nohook";
+(new ArrayObject([1]))->count();
+echo "\nExtended: Nohook";
+(new A)->count();
+//echo "\nAnonymous: Nohook";
+//(new class extends ArrayObject {})->count();
+
+$anon = new class extends ArrayObject {};
```
morrisonlevi, line 37: Just to check: this line doesn't matter at all, right? Was just missed when commenting out anon stuff?

bwoebi: Yes (not quite missed, but, yes, it does not matter).

PR #1982, opened 2023-03-21
Fix system ini handling of invalid values in zai config

bwoebi:
### Description

Also handle the special case of the empty string in the profiling log level parser because of an unquoted "off" in ini becoming an empty string.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/profiling/src/config.rs
+++ b/profiling/src/config.rs
@@ -316,13 +316,22 @@ unsafe extern "C" fn parse_level_filter(
 
     let decoded_value = &mut *decoded_value;
     match value.into_utf8() {
-        Ok(level) => match LevelFilter::from_str(level) {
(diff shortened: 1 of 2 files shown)


morrisonlevi:
For historical reference, this came up specifically when testing:

```ini
; 98-ddtrace.ini
datadog.profiling.log_level=true

; 99-whatever.ini
datadog.profiling.log_level=off
```

And I got a warning in the profiler:

```
[2023-03-20T22:54:55.430860Z WARN  datadog_php_profiling::config] zval of type 15 encountered when calling config::profiling_log_level(), expected type int (4)
```

Thanks, @bwoebi, for fixing this so quickly.

morrisonlevi (approved):

PR #1992, opened 2023-03-28
Sync profiling version with tracer; append git sha1 to non-release builds

morrisonlevi:
### Description

- The profiling version will now be synchronized with the tracer version.
- The tracer version on master will be the latest feature release instead of `1.0.0-nightly`.
- Non-release builds will append '+' followed by the git sha1 to the version number.
- Bump rust version in the Cargo.toml file. We had previously already bumped the version in other places and just missed this one.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2076,6 +2076,7 @@ jobs:
             - source-v1-{{ .Branch }}-{{ .Revision }}
       - <<: *STEP_CHECKOUT
       - <<: *STEP_ATTACH_WORKSPACE
+      - <<: *STEP_APPEND_BUILD_ID
       - unless:
           condition: << parameters.asan_build >>
           steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ aliases:
       name: Append build id to version number
       command: |
         githash="${CIRCLE_SHA1?}"
-        if [ "$CIRCLE_BRANCH" =~ "ddtrace-" ] ; then
+        if [[ "$CIRCLE_BRANCH" =~ "ddtrace-" ]] ; then
           echo "Release branch detected; not adding git sha1 to version number."
         else
           sed -E -i "s/const\s+VERSION\s*=\s*'(.*)'/const VERSION = '\1+$githash'/g" "src/DDTrace/Tracer.php"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1900,14 +1900,14 @@ jobs:
       - CARGO_TARGET_DIR: /mnt/ramdisk/cargo
     steps:
       - <<: *STEP_CHECKOUT
-      - setup_docker:
-          docker_image: << parameters.docker_image >>
       - restore_cache:
           name: Restore Cargo Package Cache
           keys:
             - profiling-cargo-cache-<< parameters.triplet >>-{{ checksum "profiling/Cargo.toml" }}
-      # Append the build id after restoring cache
+      # Append the build id after restoring cache, and before docker
       - <<: *STEP_APPEND_BUILD_ID
+      - setup_docker:
+          docker_image: << parameters.docker_image >>
       - run:
           name: Build Profiler
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2061,7 +2061,12 @@ jobs:
           command: ls -al ~/datadog/bridge
       - persist_to_workspace:
           root: '.'
-          paths: ['./bridge/_generated*.php']
+          paths:
+            - './bridge/_generated*.php'
+            # Persist files below because we generate a new version number.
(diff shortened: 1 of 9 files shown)


profiling/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
 name = "datadog-php-profiling"
-version = "0.15.0"
+version = "0.87.0-alpha.1"
```
bwoebi, line 3: Is it necessary to keep _master_ on a version number, which needs to regularly updated? I'd strongly prefer having something like `1.0.0-nightly` here on master, if possible.
It's not like we're distributing this via cargo anyway...

morrisonlevi: Customers sometimes run builds from master. I would prefer them to have a real but pre-release version number than being `1.0.0-nightly` all the time.

bwoebi: @morrisonlevi Would it be possible to instead insert the version in the circleci checkout step (for tracer as well then, replacing it in the 3 relevant files via `sed`), or are we talking rather about manual builds from source?

morrisonlevi: We discussed in person to change our release process so that the version on master is the latest feature release. We'll append `+$gitsha` e.g. `0.86.0+a885854666d58b3e06792856f8d32f0a2459bafb` in CI when the branch doesn't start with `ddtrace-`. For PECL we'll have to replace `+` with `-` or just strip it. These PECL builds aren't released anyway, so it doesn't exactly matter, but it will choke on `+`.

tooling/bin/generate-installers.sh
```diff
@@ -9,6 +9,6 @@ packages_build_dir=$2
 # Installers
 ########################
 sed "s|@release_version@|${release_version}|g" ./datadog-setup.php > "${packages_build_dir}/datadog-setup.php"
-if [[ ${release_version} == "1.0.0-nightly" && -n ${CIRCLE_WORKFLOW_JOB_ID:-} ]]; then
+if echo "${release_version}" | grep -qE '\+' && [ -n ${CIRCLE_WORKFLOW_JOB_ID:-} ]; then
```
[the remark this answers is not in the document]
morrisonlevi: The `[[` isn't valid `sh`, and we invoke this via `sh` in multiple places. The existing code was invalid -- there were warnings/errors in the logs. I'm not sure why they didn't outright fail, though.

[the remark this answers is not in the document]
morrisonlevi: I thought I reverted that part. Guess not. I actually tried using bash everywhere, but I kept hitting more places it was using `sh`, and since it wasn't causing fatal errors (just complaints, which is weird, as `set -e` is set), I wasn't confident in going forward.

bwoebi (approved):
Generally looks nice to me now :-)

tooling/bin/generate-installers.sh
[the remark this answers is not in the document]
morrisonlevi: Yes, it uses `Alpine:3.12` as a test image, which doesn't have bash.

morrisonlevi:
Installing from `php datadog-setup.php --enable-profiling --php-bin all` didn't work. However, the package extension jobs for `ddtrace-0.86.3.2` look good and it deployed to the relenv.

.circleci/config.yml
```diff
@@ -60,6 +60,20 @@ aliases:
...
 
+  - &STEP_APPEND_BUILD_ID
+    run:
+      name: Append build id to version number
+      command: |
+        githash="${CIRCLE_SHA1?}"
+        if [[ "$CIRCLE_BRANCH" =~ "ddtrace-" ]] ; then
+          echo "Release branch detected; not adding git sha1 to version number."
+        else
+          sed -E -i "s/const\s+VERSION\s*=\s*'(.*)'/const VERSION = '\1+$githash'/g" "src/DDTrace/Tracer.php"
```
r1viollet, line 71: Considering this modifies sources, can we easily make it part of the build process ? Like a different makefile target ?

.gitlab-ci.yml
```diff
@@ -7,7 +7,7 @@ variables:
     value: ""
     description: "Location where to download latest dd-library-php-*-x86_64-linux-gnu.tar.gz archive. Leave empty to take it from the latest released github tag."
   DOWNSTREAM_REL_BRANCH:
-    value: "master"
```
r1viollet, line 10: Thanks for thinking about relenv and testing this!

.circleci/config.yml
r1viollet: we talked about potential improvements to unify the version management across languages (rust / php / C) which would involve having a make target to propagate version numbers.
Accepting PR for now as this won't change the result of the build.

r1viollet (approved):
LTGM

PR #1997, opened 2023-03-31
feat(profiling): timeline support (not compiled by default)

realFlowControl:
### Description

Add support in the PHP profiler for the upcoming timeline feature.

This PR brings the following:
- CPU time with timestamps for timeline
- Wall time are shown as dots for timeline (nothing to do for us) 
- Added a `timeline` sample type for the virtual "GC" thread to put garbage collecting cycles on
- show the reason why we did a garbage collection run
- make the feature configurable (default disable, `datadog.profiling.experimental_timeline_enabled`)
- the entire feature is behind a cargo feature flag (`timeline`) which is default disabled (the code will not be compiled at this time, if you might need it, compile with `cargo build --features timeline`)

There is a few open things yet to decide on what to do:
- we could add more features (file compilation and the profiler itself)
- we need to check the gc-time vs cpu-time issue (time spend in garbage collection is also reported as cpu time, which is technically correct, but might also be misleading)

I would vote for not adding more features in this PR, but adding them in another PR to not grow this one too big and validate on the gc- vs. cpu-time thingy in PROF-7403.

PROF-7400
PROF-7401

The overhead in filesize for the pprof file we generate has been 55% in a test case where I hit the `/en/blog/` of a Symfony Demo app endpoint for 20 seconds. In the resulting pprof files you could see that the number of locations was the same, but due to sending timestamps in labels, we can not aggregate samples anymore.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.
- [ ] ~Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.~


profiling/src/lib.rs
```diff
@@ -1071,6 +1099,60 @@ extern "C" fn execute_internal(
...
+/// to run it.
+#[cfg(feature = "timeline")]
+unsafe fn gc_reason() -> &'static str {
+    let execute_data = zend::ddog_php_prof_get_current_execute_data();
+    let fname = || execute_data.as_ref()?.func.as_ref()?.name();
+    match fname() {
+        Some(name) if name == b"gc_collect_cycles" => "induced",
+        _ => "engine",
+    }
+}
```
morrisonlevi, line 1114: @realFlowControl I refactored this bit, partly because I wasn't sure func would always be there and so I didn't trust the pointer dereferencing.

realFlowControl:
Converted this to a draft, as @chrisnas  mentioned that @gleocadie found a rounding issue in the `libdatadog` upscaling, especially when combined with timeline

realFlowControl:
Version bump to `libdatadog` version `2.2.0` containing the fix is done

profiling/src/profiling/mod.rs
```diff
@@ -658,6 +664,54 @@ impl Profiler {
...
+        lazy_static! {
+            static ref TIMELINE_GC_LABELS: Vec<Label> = vec![
+                Label {
+                    key: "event",
+                    value: LabelValue::Str("gc".into()),
+                },
+                Label {
+                    key: "thread id",
+                    value: LabelValue::Str("GC".into()),
+                },
```
morrisonlevi, line 683: I am confused. The labels "event: gc" and "thread id: GC" mean... what?

morrisonlevi (commented):
I pushed some minor code things (like making clippy happy). I have a few questions.

profiling/src/profiling/mod.rs
realFlowControl: - the `thread id` makes it go into the extra "virtual" garbage collection thread in the visualisation
- the `gc` in `event` marks it a garbage collection event so it will be displayed as such in the frontend

Our alternatives (and I would have to play with this) would be to not set a `thread id` and have this GC-event in the "main" thread, which would be more correct in the case of PHP, but IDK if the backend supports this for visualisation, but I think I will try this out.

profiling/src/profiling/mod.rs
```diff
@@ -658,6 +664,54 @@ impl Profiler {
...
+        labels.extend_from_slice(&TIMELINE_GC_LABELS);
+        labels.push(Label {
+            key: "gc reason",
+            value: LabelValue::Str(String::from(reason).into()),
+        });
+        let n_labels = labels.len();
+
+        match self.send_sample(Profiler::prepare_sample_message(
+            vec![ZendFrame {
+                function: "internal|gc_collect_cycles".to_string(),
```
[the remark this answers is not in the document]
bwoebi: In .NET the garbage collector is a separate thread not having a proper (managed) call stack. Hence that's the right thing to do in .NET, but not so in PHP.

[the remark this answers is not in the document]
morrisonlevi: Currently, the stack trace for this profile type is unused by the UI, right? So it would avoid unnecessary overhead.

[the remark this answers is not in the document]
morrisonlevi: We have `[truncated]` as a fake frame name already, and I think native has a few other ones using square brackets as well. If we decide to use call stacks at all, I think we could push `[gc]` or `[garbage collection]` to the top of the profiler stack (not the executor vm stack). Any thoughts on that?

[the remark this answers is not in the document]
bwoebi: I'm not sure I understand the question. Do you suggest truncating the stack to the entrypoint of the gc in general?
Unsure about that, may be helpful to locate code exhibiting heavy GC overhead.

I don't know about unused or not in the UI.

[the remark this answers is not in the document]
morrisonlevi: The PR currently pushes `internal|gc_collect_cycles` to the stack, and it's the only item. It doesn't take a full stack trace. I am suggesting:

- Since the stack trace is unused, to save on time and bytes, we shouldn't send any stack at all.
- But if we do send some kind of stack now or in the future, we should put a fake frame `[gc]` or `[garbage collection]` at the top of whatever stack we send, instead of `internal|gc_collect_cycles`.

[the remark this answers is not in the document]
realFlowControl: FYI: just found out that in the current version of the backend the stack trace for gc events will be shown, but as we do not know how much GC events there are (should be really just very few). Not sending stack frames make these events missing in the resulting pprof file, IDK if the pprof spec actually allows for "no frames".
I'd suppose we do send `[gc]` and measure how many garbage collection events there are when we have this feature out. This is really low overhead (as we omit stack walking) and when we have numbers, we can decide again if want the stack trace to show up for customers.
WDYT?

profiling/src/profiling/mod.rs
realFlowControl: I removed the `thread id`, as in the current version of the backend `gc` type events will be shown in this virtual gc thread anyway. We can solve this part on the backend then.

realFlowControl:
With a PHP script the creates a lot of garbage, the visualisation of the timeline looks something like this:

image

profiling/src/lib.rs
```diff
@@ -675,6 +686,7 @@ unsafe extern "C" fn minfo(module_ptr: *mut zend::ModuleEntry) {
             },
         );
 
+        #[cfg(feature = "allocation_profiling")]
```
morrisonlevi, line 689: Let's print out a message saying the feature was not compiled in, since it's on by default and we've shipped it. Let's not do it for timeline since it's not ready.

profiling/src/profiling/mod.rs
```diff
@@ -722,6 +789,24 @@ impl Profiler {
...
+            if locals.profiling_experimental_timeline_enabled
+                && (samples.timeline > 0 || samples.cpu_time > 0)
+            {
+                if let Ok(now) = SystemTime::now().duration_since(UNIX_EPOCH) {
+                    labels.push(Label {
+                        key: "end_timestamp_ns",
+                        value: LabelValue::Num(now.as_nanos() as i64, Some("nanoseconds")),
+                    });
+                }
+            }
```
morrisonlevi, line 809: We should be able to combine these two `cfg()`s into an outer if with a nested inner if.

profiling/src/timeline.rs
```diff
@@ -0,0 +1,99 @@
...
+            gc_protected: false,
+            #[cfg(php_gc_status_extended)]
+            full: false,
+            runs: 0,
+            collected: 0,
+            threshold: 0,
+            #[cfg(php_gc_status_extended)]
+            buf_size: 0,
+            num_roots: 0,
+        };
```
morrisonlevi, line 52: Florian mentioned using `MaybeUninit` and I think that's better. Just be sure to use `.as_ptr()` instead of using a reference when passing it because taking the reference to an uninitialized struct is undefined behavior in Rust.

profiling/tests/phpt/gc_collect_cycles_01.phpt
```diff
@@ -0,0 +1,54 @@
...
+DD_PROFILING_ENABLED=yes
+DD_PROFILING_EXPERIMENTAL_CPU_TIME_ENABLED=no
+DD_PROFILING_EXPERIMENTAL_TIMELINE_ENABLED=yes
+DD_PROFILING_ALLOCATION_ENABLED=no
+DD_PROFILING_LOG_LEVEL=debug
+--FILE--
+<?php
+
+class SomeClass {
+    public SomeClass $someProperty;
```
morrisonlevi, line 26: This is too new of a feature, needs to go back to PHP 7.1. It didn't fail in CI because this feature is not enabled for the phpt tests, so it gets skipped.

morrisonlevi (approved):

Issue #1993, opened 2023-03-29
[BUG] Profiling SIGSEGV in 0.86.0


zeroxs:
Edit by Datadog: For anyone landing here, here's the summary of the root cause: https://github.com/DataDog/dd-trace-php/issues/1993#issuecomment-1491105610.

### Bug description
We had a production outage last week after a release that spanned many hours trying to determine what was going on. Code reversions did not help solve this issue and it was only at the end did we try to unload the Datadog PHP extensions and suddenly everything proceeded to work. The issues encountered were excessive PHP segfaults and when PHP did not segfault, a large amount of requests hit the PHP memory limit and would "gracefully" die on PHP's side. Very few requests would make it through successfully but there was no observable strain on the resources of the systems in question. This occurred BOTH when PHP was running as FPM and as an Apache module. Difference is Apache would segfault and die instead of the PHP process so it would segfault all the same.

After much digging, and to explain why only some of our servers are experiencing this issue, the problem version of the extension is 0.86.0. Downgrading to 84 works fine, moving to 85 works fine, 86 breaks again. This makes sense considering our other servers are running 0.85.0 and under and were not experiencing any issues.

Utilizing my history as a native (C/++) developer, I have a sneaking suspicion that there is a possibility of the thread-local storage change for stack sample collection being overconsumed and causing a crash. https://github.com/DataDog/dd-trace-php/pull/1949#issuecomment-1450856002

This would explain why the segfaults are seemingly random because requests succeed as the storage fills up until it pops.

(this is a copy of the ticket I created with Datadog support as this directly impacted our production environment, but am posting it here for awareness and for anyone else that may encounter it)

### PHP version
```
PHP 8.0.28 (cli) (built: Feb 14 2023 11:05:44) ( NTS gcc x86_64 )
Copyright (c) The PHP Group
Zend Engine v4.0.28, Copyright (c) Zend Technologies
    with Zend OPcache v8.0.28, Copyright (c), by Zend Technologies
```

### Tracer version
0.86.0

### Installed extensions
<details>
<summary>Extensions</summary>

```
[PHP Modules]
bcmath
bz2
calendar
Core
ctype
curl
datadog-profiling
date
ddappsec
ddtrace
dom
exif
fileinfo
filter
ftp
gd
gettext
hash
iconv
igbinary
intl
json
libxml
mbstring
msgpack
mysqli
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
[...]

PR #1998, opened 2023-03-31
perf(profiling): speed up stack walking by using function run_time_cache 

morrisonlevi:
### Description

This essentially re-applies #1949, with some changes:

- Exclude accessing the `run_time_cache` on calls with trampolines, which causes a crash, see  #1993.
- Massage the source of `has_invalid_run_time_cache` to get good codegen out of it.
- ~Fix a typo I kept making (stalk_walking instead of stack-walking).~ Reverted so performance can be more easily be compared when using native profiler.
- Reset the string table on each request, preventing the table from unbounded memory growth across request.
     - Exclude the CLI from using the run_time_cache optimization, since it only has 1 PHP request and is often long lived.

Despite fixing the sigsegv and taking precautions about unbounded memory growth, I think we need to be cautious about rolling this out.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/php_ffi.c
```diff
@@ -142,7 +156,83 @@ void ddog_php_prof_zend_mm_set_custom_handlers(zend_mm_heap *heap,
...
+
+#if PHP_VERSION_ID < 80200
+    // Internal functions don't have a runtime cache until PHP 8.2.
+    uintptr_t *cache_addr = RUN_TIME_CACHE(&func->op_array);
+#else
+    uintptr_t *cache_addr = RUN_TIME_CACHE(&func->common);
+#endif
+
+    // todo: how sure can I be, here? Can I be so confident as to omit it?
+    ZEND_ASSERT(cache_addr);
```
morrisonlevi, line 226: @bwoebi Can you think of any reason for the `cache_addr` to be null, given the guards in `has_invalid_run_time_cache`?

bwoebi: No. That's it.

profiling/Cargo.toml
```diff
@@ -6,16 +6,18 @@ license = "Apache-2.0"
 rust-version = "1.64"
 
 [lib]
-crate-type = ["cdylib"]
+crate-type = ["cdylib", "rlib"]
```
morrisonlevi, line 9: Is rlib okay here? Doesn't it break LTO? Or did we fix that by using newer CLI invocations to specify crate-types during the build in CI?

profiling/Cargo.toml
```diff
@@ -25,9 +27,10 @@ libc = "0.2"
 # TRACE set to max to support runtime configuration.
 log = { version = "0.4", features = ["max_level_trace", "release_max_level_trace"]}
 once_cell = { version = "1.12" }
-uuid = { version = "1.0", features = ["v4"] }
+ouroboros = { version = "0.15.6" }
```
morrisonlevi, line 30: Need version 0.17+ to avoid a soundness bug:
https://github.com/joshua-maros/ouroboros/issues/88

profiling/src/php_ffi.c
morrisonlevi: Thanks. I need to update the comment then.

profiling/Cargo.toml
realFlowControl: Nope, that was a left over from me locally running the benchmarks
```suggestion
crate-type = ["cdylib"]
```

bwoebi (approved):

realFlowControl (approved):

PR #1999, opened 2023-04-01
[Tracer][YII] Do not use module property if it doesn't exist

pierotibou:
### Description
When using modules in Yii, the controller doesn't have the same properties at hand. We used to try and access the module property which was failing in some cases and caused 500 in a user application.

We now check the existence of the property and fallback on another method that work for modules. 

In terms of implementation, I've first confirmed I could reproduce by extending the tests, and then fixed the code. 

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


PROFeNoM (approved):
LGTM!

PR #2016, opened 2023-04-14
Prevent execution of install_hook() hooks after disabling tracer

bwoebi:
### Description

This led to crashes in combination with the database integrations, if database calls happened after the tracer was disabled.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/tests/ext/sandbox/install_hook/hook_not_invoked_if_tracer_disabled.phpt
+++ b/tests/ext/sandbox/install_hook/hook_not_invoked_if_tracer_disabled.phpt
@@ -1,0 +1,19 @@
+--TEST--
+Ensure hooks are not invoked after the tracer has been disabled
+--FILE--
+<?php
+
+function foo() {}
+DDTrace\install_hook("foo", function($h) {
+    print "Invoked\n";
+    ini_set("datadog.trace.enabled", 0);
+    $h->span(); // must not crash
+}, function() {
+    print "Skipped\n";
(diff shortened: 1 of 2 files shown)


tests/ext/sandbox/install_hook/hook_not_invoked_if_tracer_disabled.phpt
```diff
@@ -0,0 +1,19 @@
+--TEST--
+Ensure hooks are not invoked after the tracer has been disabled
+--FILE--
+<?php
+
+function foo() {}
+DDTrace\install_hook("foo", function($h) {
+    print "Invoked\n";
+    ini_set("datadog.trace.enabled", 0);
+    $h->span(); // must not crash
```
Anilm3, line 10: If I understand correctly, on the first call this is a dummy span, wouldn't it also make sense to test that the span isn't pushed?

Anilm3 (approved):
lgtm

tests/ext/sandbox/install_hook/hook_not_invoked_if_tracer_disabled.phpt
bwoebi: This is not testable as, when the tracer is disabled, the accessors are shortcircuited - pushing a span would crash, given it'd push onto destroyed structures anyway.

PR #2017, opened 2023-04-19
Fix install_hook on functions returning by-reference

bwoebi:
### Description

Two issues:
- IS_VAR passed to ZEND_RETURN(_BY_REF) on PHP 7 may be IS_INDIRECT. When copying these, we must ensure to not copy the IS_INDIRECT itself. It largely was unproblematic, but manipulating $hookData->returned could have unexpected side effects.
- On PHP 8 we need to track type sources as the $returned property is typed mixed and thus, the type source needs to be carried around. Otherwise we run into an assertion failure.

Note that these changes on the PHP 7 side also enable changing the return value (on PHP 8 no extra handling is needed).

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/zend_abstract_interface/interceptor/php7/interceptor.c
+++ b/zend_abstract_interface/interceptor/php7/interceptor.c
@@ -127,7 +127,20 @@ void zai_interceptor_op_array_pass_two(zend_op_array *op_array) {
         }
 
         // For generators we need our own temporary to store a constant array which is converted to an iterator
-        if (op_array->fn_flags & ZEND_ACC_GENERATOR) {
+        bool need_temporary = op_array->fn_flags & ZEND_ACC_GENERATOR;
+        if (!need_temporary) {
+            for (uint32_t op = i; op < op_array->last; ++op) {
+                // For RETURN opcodes, we may need a temporary to replace the return values
+                if ((opcodes[op].opcode == ZEND_RETURN || opcodes[op].opcode == ZEND_RETURN_BY_REF)) {
+                    if (opcodes[op].op1_type == IS_CV || opcodes[op].op1_type == IS_CONST) {
+                        need_temporary = true;
+                        break;
+                    }
+                }
+            }
+        }
+
+        if (need_temporary) {
             // For Optimizer to allocate a temporary for us, the temporary must exist
             // To not interfere with live range calculation, the temporary must be defined as a result
             opcodes[i].result_type = IS_TMP_VAR;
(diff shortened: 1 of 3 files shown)


zend_abstract_interface/interceptor/php7/interceptor.c
```diff
@@ -181,18 +203,59 @@ static inline zval *zai_interceptor_get_zval_ptr(const zend_op *opline, int op_t
...
 
+ZEND_TLS zend_op zai_interceptor_custom_return_op;
 static inline void zai_interceptor_return_impl(zend_execute_data *execute_data) {
     zai_interceptor_frame_memory *frame_memory;
     if (zai_hook_memory_table_find(execute_data, &frame_memory)) {
         if (!frame_memory->implicit) {
             zval rv;
             zval *retval = zai_interceptor_get_zval_ptr_op1(execute_data);
-            if (!retval || Z_TYPE_INFO_P(retval) == IS_UNDEF) {
+            bool needs_copy = EX(opline)->op1_type == IS_CONST || EX(opline)->op1_type == IS_CV;
```
realFlowControl, line 213: **question (non-blocking):** @bwoebi just for my understanding, `needs_copy` is true for stuff like
```php
function foo() {
    return 1; // IS_CONST
}
```
or
```php
function foo() {
    $bar = 42;
    return $bar; // IS_CV
}
```
correct?

zend_abstract_interface/interceptor/php7/interceptor.c
```diff
@@ -181,18 +203,59 @@ static inline zval *zai_interceptor_get_zval_ptr(const zend_op *opline, int op_t
...
+                        // it's sad, but we cannot support it properly right now if we don't find a temporary? shouldn't happen though
+                        zval_ptr_dtor(&rv);
+                    }
+                } else {
+                    // Override any IS_INDIRECT, hence directly to EX_VAR() instead of retval
+                    ZVAL_COPY_VALUE(EX_VAR(EX(opline)->op1.var), &rv);
+                }
+            } else if (needs_copy) {
+                zval_ptr_dtor_nogc(retval);
+            }
```
realFlowControl, line 258: **question:** @bwoebi I am not 100% whats going on in this block 😵 
My understanding so far: the `zai_interceptor_return_impl()` function is to process the `ZEND_RETURN` opcode of a hook that has been installed in user land using `DDTrace\install_hook`.
So in `op1` we should find the value to be returned by the hook?

zend_abstract_interface/interceptor/php7/interceptor.c
bwoebi: Yes.

realFlowControl (approved):

PR #2025, opened 2023-04-27
perf(profiling): only wake every 10ms if PHP is serving a request

morrisonlevi:
### Description

We accomplish this by tracking if we're between request init and request shutdown. I split the interrupt parts into a new thread.  It has two nearly `select!`s in a loop, where one `select!` doesn't have have the code path for `recv`ing the `crossbeam_channel::tick(Duration::from_millis(10))`. A bit inelegant, but it works and is cross-platform.

~Also change `sample/count` profile type to `wall-samples/count`, since that's what it is. This name is better aligned with conventions used such as `exception-samples`, `heap-live-samples`, `lock-acquire-samples`, etc. The `sample/count` part of our profiles is currently unused in backend.~ I kept the code changes surrounding this change, but for now I have reverted the part that actually renames it in the profiles and tests and will leave that for another PR.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/profiling/mod.rs
```diff
@@ -690,7 +678,7 @@ impl Profiler {
     ) -> SampleMessage {
         // Lay this out in the same order as SampleValues
         static SAMPLE_TYPES: &[ValueType; 5] = &[
-            ValueType::new("sample", "count"),
+            ValueType::new("wall-samples", "count"),
```
r1viollet, line 681: Are these not `time-samples` ? As they apply to CPU.
Does this imply `backend` changes ?

profiling/src/interrupter.rs
```diff
@@ -0,0 +1,192 @@
...
+                                Ok(Message::Stop) => {
+                                    active = false;
+                                }
+                                _ => {
+                                    break;
+                                }
+                            },
+
+                            recv(ticker) -> message => match message {
+                                Ok(_) => signal_pointers.notify(),
```
morrisonlevi, line 127: Add an `if active` check here.

profiling/src/interrupter.rs
```diff
@@ -0,0 +1,192 @@
...
+            });
+
+            Self {
+                sender,
+                join_handle: Some(join_handle),
+            }
+        }
+
+        pub fn start(&self) -> anyhow::Result<()> {
+            self.sender.send(Message::Start).map_err(|err| {
```
morrisonlevi, line 165: Move this to send with timeout in case the other thread has crashed.

morrisonlevi:
I need to fix a few things:

- Not crash on fork.
- Handle potential deadlocks if thread is de-sync'd.

morrisonlevi:
In a recent commit, I removed some barriers in exchange for thread parking/unparking. _This solution is not as robust_. If the threads wake spuriously, we could deadlock or have other race conditions during the fork. I was a bit tired when I made those changes and I need to revisit them when I'm fresh.

morrisonlevi:
I was looking around at thread park implementations, and on Linux it handles spurious wakeup:

```rs
    pub unsafe fn park(self: Pin<&Self>) {
        // Change NOTIFIED=>EMPTY or EMPTY=>PARKED, and directly return in the
        // first case.
        if self.state.fetch_sub(1, Acquire) == NOTIFIED {
            return;
        }
        loop {
            // Wait for something to happen, assuming it's still set to PARKED.
            futex_wait(&self.state, PARKED, None);
            // Change NOTIFIED=>EMPTY and return in that case.
            if self.state.compare_exchange(NOTIFIED, EMPTY, Acquire, Acquire).is_ok() {
                return;
            } else {
                // Spurious wake up. We loop to try again.
            }
        }
    }
```

So, I think we can do this simpler thing!

morrisonlevi:
Florian and I spent some time together in Zoom adding panics to various places in the `ddprof_time` thread in `interrupter.rs`. At least with our efforts, we couldn't get it to deadlock. The general behavior is that time samples will stop being generated for the rest of the request, and then on the next request it will spawn a new interrupter and try again.

profiling/src/interrupter.rs
morrisonlevi: The precaution of timeouts seem unnecessary as Florian and I could not get this to deadlock when causing the associated thread to panic in various places.

morrisonlevi:
Forking has been broken. We don't support forking, but we still shouldn't deadlock nor crash if our users fork, and currently we can deadlock.

PR #2026, opened 2023-04-27
Add Laravel Queue Integration with Span Links

PROFeNoM:
### Description

This PR comprises three features:
- A Laravel queue integration for Laravel 5+
- Implements Span Links
- Fixes an issue where internal server errors would not be set on the root span
- Fixes a bug where the tracer would enter limited mode because of an underflow in the number of open spans with multiple parallel traces

The Laravel Queue integration traces the following methods:
- `Illuminate\Queue\Worker::process`
- `Illuminate\Queue\Jobs\Job::fire`
- `Illuminate\Queue\Jobs\Job::resolve`
- (Laravel 8+) `Illuminate\Bus\Batch::add`
- (Laravel 8+) `Illuminate\Queue\Queue::enqueueUsing`
- `Illuminate\Contracts\Queue\Queue::push`
- `Illuminate\Contracts\Queue\Queue::later`
- The job's action

Things to note:
- If applicable, the batch id will be retrieved and set in the metadata of the span. This will only be the case in Laravel 8+, as batches were implemented in this version.
- It could be argued that tracing `Illuminate\Queue\Queue::enqueueUsing` doesn't provide much information. It just provides some context onto _'why are all these events happening'_. However, I don't mind removing it.
- When setting a span's metadata for a job that is an object - i.e., not an instance of `Job`, nor a string (e.g., redis), information about the connection and the queue will be tried to be retrieved. In the event that it wasn't possible to retrieve it in from the object or the parameters, the configuration defaults (`config/queue.php`) will be used.
- (Artisan) In laravel's tests, the migrations will be run during the `post-autoload-dump` phase. Considering the sample apps are in production mode, the migrations are forced and the whole database is basically dropped and rebuilt from the migrations. This was needed to ran the jobs-related migrations.
- (Laravel 4) This Laravel Queue integration doesn't provide support with Laravel 4.2
- (Apache) While testing, I faced an issue where the apache server would not be restarted/reloaded after the configuration changed, hence that's why there is a change in `PhpApache.php`.
- (CI) The Laravel Testsuite was disabled in the CI, as some Queue tests weren't passing as they would include additional information in their payload because of the distributed tracing. This could be fixed by modifying the corresponding docker image, but it would require cross-compilation.  

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.


pierotibou (commented):
Thanks. I've just reviewed the tags you used based on the documentation for messaging system in confluence. Could you check your entries? I'd expect a bit more `Tag::MQ_*` to be used honestly, but maybe I'm wrong.
Can you also then fill the confluence page with the laravel specifics? It's not mandatory as won't be reused by others but it's a good habit to take I believe

src/Integrations/Integrations/LaravelQueue/LaravelQueueIntegration.php
```diff
@@ -0,0 +1,330 @@
...
+        if ($exception) {
+            $this->setError($span, $exception);
+        }
+    }
+
+    public function getMetadataFromJob(Job $job)
+    {
+        $metadata = [
+            'messaging.laravel.attempts' => $job->attempts(),
+            'messaging.laravel.connection' => $job->getConnectionName() ?? config('queue.default'),
```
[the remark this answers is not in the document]
PROFeNoM: The connection name will be something like 'database', 'sync', or 'redis' (non-exhaustive, these are defined in the `config/queue.php` file). It doesn't really have the format we would expect from a url in my opinion. However, if you feel like this makes sense, then that's ok for me.

src/Integrations/Integrations/LaravelQueue/LaravelQueueIntegration.php
```diff
@@ -0,0 +1,330 @@
...
+            'messaging.laravel.id' => $job->getJobId(),
+            'messaging.laravel.max_tries' => $job->maxTries(),
+            'messaging.laravel.queue' => $job->getQueue(),
```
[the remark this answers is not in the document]
PROFeNoM: Indeed, I think using `messaging.destination`makes sense

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -234,6 +234,15 @@ function ($This, $scope, $args) use ($rootSpan, $integration) {
...
         );
 
+        // Used by Laravel >= 5.0
+        \DDTrace\hook_method(
+            'Illuminate\Contracts\Debug\ExceptionHandler',
+            'report',
+            function ($This, $scope, $args) use ($rootSpan, $integration) {
+                $integration->setError($rootSpan, $args[0]);
+            }
+        );
```
bwoebi, line 244: Needs to be wrapped in `if (PHP_MAJOR_VERSION > 5)`

src/Integrations/Integrations/IntegrationsLoader.php
```diff
@@ -76,6 +77,8 @@ public function __construct(array $integrations)
             '\DDTrace\Integrations\Laravel\LaravelIntegration';
         $this->integrations[MysqliIntegration::NAME] =
             '\DDTrace\Integrations\Mysqli\MysqliIntegration';
+        $this->integrations[LaravelQueueIntegration::NAME] =
+            '\DDTrace\Integrations\LaravelQueue\LaravelQueueIntegration';
```
bwoebi, line 81: Are there no sensible entrypoints to be defined in integrations.c?

src/Integrations/Integrations/LaravelQueue/LaravelQueueIntegration.php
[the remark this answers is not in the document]
amarziali: is this like this? `'mysql:host=mysql_integration;dbname=test'` 
In those cases my concerns are more about leaking sensitive info?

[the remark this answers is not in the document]
PROFeNoM: Laravel connections are defined using an associative array ([as can be seen in this sample configuration](https://github.com/agence-webup/laravel-skeleton/blob/2f93dfd40ede154361cc2a9fd3f2000560e071fc/config/queue.php#L31)) with specific **configuration keys**, rather than a single string containing all parameters. Therefore, the value of `messaging.laravel.connection` would be one of these **configuration keys**, referencing to one of the defined queue connection types. 

Note: A 'connection' in the context of laravel queue refers to the method by which the queue system communicates with the underlying storage mechanism used to store and retrieve queued jobs.

zend_abstract_interface/json/json.c
```diff
@@ -22,16 +22,32 @@ __attribute__((weak)) int php_json_encode(smart_str *buf, zval *val, int options
...
+    }
+
+    zend_class_entry **tmp_json_serializable_ce = (zend_class_entry **)DL_FETCH_SYMBOL(json_me->handle, "php_json_serializable_ce");
+    if (tmp_json_serializable_ce == NULL) {
+        tmp_json_serializable_ce = (zend_class_entry **)DL_FETCH_SYMBOL(json_me->handle, "_php_json_serializable_ce");
+    }
+    if (tmp_json_serializable_ce != NULL) {
+        php_json_serializable_ce = *tmp_json_serializable_ce;
+    }
+}
```
bwoebi, line 41: This is ... dead code, right? I see no usages of that.

src/Integrations/Integrations/LaravelQueue/LaravelQueueIntegration.php
```diff
@@ -0,0 +1,333 @@
...
+
+    public function getMetadataFromJob(Job $job)
+    {
+        $metadata = [
+            'messaging.laravel.attempts' => $job->attempts(),
+            'messaging.laravel.connection' => $job->getConnectionName() ?? config('queue.default'),
+            'messaging.laravel.max_tries' => $job->maxTries(),
+            'messaging.laravel.timeout' => $job->timeout(),
+            'messaging.laravel.name' => $job->resolveName(),
+            Tag::MQ_SYSTEM => 'laravel',
```
morrisonlevi, line 270: Isn't this supposed to be like `kafka`, `rabbitmq`, `rocketmq`, `activemq`, `AmazonSQS`, etc?

bwoebi: Well, Laravel queues aren't dispatched via these mechanisms. It's a separate mechanism.

zend_abstract_interface/json/json.c
PROFeNoM: yep

src/Integrations/Integrations/LaravelQueue/LaravelQueueIntegration.php
PROFeNoM: On this, I was also planning to update the confluence page to include the newly introduced span attributes; `messaging.system` would be one of these.

src/Integrations/Integrations/IntegrationsLoader.php
```diff
@@ -10,6 +10,7 @@
 use DDTrace\Integrations\Guzzle\GuzzleIntegration;
 use DDTrace\Integrations\Laminas\LaminasIntegration;
 use DDTrace\Integrations\Laravel\LaravelIntegration;
+use DDTrace\Integrations\LaravelQueue\LaravelQueueIntegration;
```
bwoebi, line 13:
```suggestion
```

src/Integrations/Integrations/IntegrationsLoader.php
```diff
@@ -10,6 +10,7 @@
 use DDTrace\Integrations\Guzzle\GuzzleIntegration;
 use DDTrace\Integrations\Laminas\LaminasIntegration;
```
bwoebi, line 11:
```suggestion
```

ext/ddtrace.c
```diff
@@ -392,6 +430,7 @@ static zend_object *ddtrace_span_stack_clone_obj(zend_object *old_obj) {
 static void ddtrace_span_data_free_storage(zend_object *object) {
     zend_object_std_dtor(object);
     // Prevent use after free after zend_objects_store_free_object_storage is called (e.g. preloading) [PHP < 8.1]
+    memset(&object->properties_table[8], 0, sizeof(((ddtrace_span_data *)NULL)->properties_table_placeholder[8])); // links property
```
bwoebi, line 433:
```suggestion
```
`object->properties_table[8]` is pointing to the `parent` property, not the links property. Unless this was necessary, but then there's another bug.

ext/ddtrace.stub.php
```diff
@@ -23,6 +23,40 @@
      */
     const DBM_PROPAGATION_FULL = UNKNOWN;
 
+    class SpanLink implements \JsonSerializable {
+        /**
+         * @var string $traceId A 32-character, lower-case hexadecimal encoded string of the linked trace ID. This field
+         * shouldn't be directly assigned. Use the SpanData::getLinks() method instead.
```
bwoebi, line 29:
```suggestion
         * shouldn't be directly assigned an id from SpanData. Use the SpanData::getLinks() method instead.
```
directly assigning it when getting it from distributed tracing or such is fine, just not assigning a decimal id.

PR #2032, opened 2023-05-03
feat(profiling): emit metrics for stack-walking overhead

morrisonlevi:
### Description

Currently emits a metric `datadog.profiling.php.stack_walk_ns` as a histogram. It's gated behind the feature `profiling_metrics` which is off by default.

This is needed to more accurately track the cost of stack walking. 

Also fixes some clippy warnings.

PROF-7576

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


realFlowControl:
It's working 😄
I was just thinking we could maybe move the `dogstatsd.rs` to libdatadog. We don't necessarily need to it in this PR, we could also move later. WDYT?

realFlowControl (approved):

morrisonlevi:
Do we need to turn this on for releases (not release builds, the releases we ship to customers)? My thought was we don't, and this lowers our risk substantially.

realFlowControl:
I thought the same, that's why I approved it.
Additionally I like the idea of somewhen in the future having our profiler sending those kind of data to us, so we would have more insight in how it performs on customer machines.

bwoebi:
I also think dogstatsd would be a good fit for a small crate in libdatadog. Can we move it there? :-)

morrisonlevi:
I worry if we put the dogstatsd bits into libdatadog that it may unintentionally get used in prod by libdatadog users. There is a tiny bit of testing but I'm not sure it's production ready. That doesn't matter for this PR, since it's not enabled for our production builds, but users in libdatadog may not realize this.

I do agree it would be nice to have a solid dogstatsd client in libdatadog, though.

PR #2034, opened 2023-05-05
test(profiling) add stack walking test

realFlowControl:
### Description

This PR will add a test for stack walking to the profiler, that can be run without PHP being around. The main reason for this PR is that it provides the foundation of building micro benchmarks for the profiler that run without PHP.

- adds a unit test that validates the resulting stack frames against a fake PHP stack
- adds function to create a fake stack frame
- which is only compiled when enabling the `stack_walking_tests` feature

When reviewing please keep on eye on the following: The `ddog_php_test_*` functions in the `php_ffi.c` should not be compiled when building a release, but only when running tests. Currently this is implemented with a cargo feature `stack_walking_tests`, but maybe in the future we find a more automatic way.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/php_ffi.c
+++ b/profiling/src/php_ffi.c
@@ -4,6 +4,10 @@
 #include <stdbool.h>
 #include <string.h>
 
+#if CFG_STACK_WALKING_TESTS
+#include <dlfcn.h> // for dlsym
+#endif
+
 const char *datadog_extension_build_id(void) { return ZEND_EXTENSION_BUILD_ID; }
 const char *datadog_module_build_id(void) { return ZEND_MODULE_BUILD_ID; }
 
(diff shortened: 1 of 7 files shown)


profiling/src/php_ffi.c
```diff
@@ -146,3 +146,68 @@ zend_execute_data* ddog_php_prof_get_current_execute_data()
 {
     return EG(current_execute_data);
 }
+
+#if CFG_STACK_WALKING_TESTS
+zend_execute_data* ddog_php_test_create_fake_zend_execute_data(int depth) {
+    if (depth <= 0 || depth > 99) {
```
r1viollet, line 152: Minor: There is probably value in allowing the test to work for more extreme stack depths.

profiling/src/php_ffi.c
```diff
@@ -146,3 +146,68 @@ zend_execute_data* ddog_php_prof_get_current_execute_data()
...
+        return NULL;
+    }
+
+    zend_execute_data *execute_data = (zend_execute_data *) malloc(sizeof(zend_execute_data));
+    memset(execute_data, 0, sizeof(zend_execute_data));
+
+    execute_data->prev_execute_data = ddog_php_test_create_fake_zend_execute_data(depth - 1);
```
r1viollet, line 159: So recursive execution of all stack depths :smile: ?

profiling/src/php_ffi.h
```diff
@@ -120,3 +120,10 @@ void ddog_php_prof_zend_mm_set_custom_handlers(zend_mm_heap *heap,
                                                void* (*_realloc)(void*, size_t));
 
 zend_execute_data* ddog_php_prof_get_current_execute_data();
+
+/**
+ * The following two functions exist for the sole purpose of creating fake stack
+ * frames that can be used in testing/benchmarking scenarios
+ */
+zend_execute_data* ddog_php_test_create_fake_zend_execute_data(int depth);
```
r1viollet, line 128: Not too familiar with codebase, though I'm guessing it could be better to create a dedicated ffi header ?

profiling/src/profiling/stalk_walking.rs
```diff
@@ -123,3 +123,33 @@ pub(super) unsafe fn collect_stack_sample(
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::bindings as zend;
+
+    #[test]
+    #[cfg(feature = "stack_walking_tests")]
+    fn test_collect_stack_sample() {
+        unsafe {
+            let fake_execute_data = zend::ddog_php_test_create_fake_zend_execute_data(3);
```
r1viollet, line 136: This is super nice! 
This will also allow us to have better tests around what we can expect from zend_execute_data.

profiling/src/php_ffi.c
```diff
@@ -146,3 +146,68 @@ zend_execute_data* ddog_php_prof_get_current_execute_data()
...
+
+    // add function name to stack frame
+    zend_string *func_name = malloc(sizeof(zend_string) + 16);
+    func_name->h = 0;
+    func_name->len = 16;
+    memcpy(func_name->val, "function name 00", 16);
+    if (depth > 9) {
```
r1viollet, line 166: snprintf seems like less work ^^

r1viollet (approved):
LGTM
The only important comment is around splitting the FFI header
More tests ! :heart:

profiling/src/php_ffi.c
realFlowControl: I actually found a way using `dlsym()` to look for the original symbol, CI currently running

morrisonlevi (approved):
I'm approving, but please check check the commit I pushed up. I added more error checking, extracted a few things so I could better see what was going on, and also used `zend_string_init(..., persistent: true)` instead of re-creating that. It did require me to copy/paste some code for `__zend_malloc` and mark it as `__attribute__((weak))` though.

PR #2040, opened 2023-05-09
Reduce the allocated heap size of the elasticsearch7 container

PROFeNoM:
### Description

It sometimes happened that the elasticsearch7 container would exit with the error code 137 in the CI, but with the introduction of the SQLSRV integration and its additional container, elasticsearch would often exit, hence making the integrations tests fail. 

This PR just changes the heap size of this container (or more specifically, the amount of memory the JVM is allowed to use), which is 2gb by default, to 1gb, which should be enough for testing purposes.

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,7 @@ aliases:
     name: elasticsearch7_integration
     environment:
       - discovery.type=single-node
-      - ES_JAVA_OPTS=-Xms1g -Xmx1g
+      - ES_JAVA_OPTS=-Xms1536m -Xmx1536m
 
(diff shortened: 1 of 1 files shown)


pierotibou (approved):

morrisonlevi:
I assume this fixes it because the container doesn't have 2GB of memory, so it gets OOM killed?

PROFeNoM:
> I assume this fixes it because the container doesn't have 2GB of memory, so it gets OOM killed?

The container would indeed get OOM killed (more information [https://github.com/10up/wp-local-docker/issues/6](url))

PR #2046, opened 2023-05-11
Do not force cycle collection at runtime on flush by default

bwoebi:
### Description

We had an user where it massively impacted their performance: And _practically_ it should rarely make a difference (it only matters with autoflush active and forgetting to close all of a stacks spans and GC running next to never in a long running script. But if it _does_ delay traces from arriving, there's an escape hatch configuration option.

### Readiness checklist
- [ ] ~Tests added for this feature/bug.~ Performance issue rather.

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/span.c
+++ b/ext/span.c
@@ -513,7 +513,7 @@ static void dd_close_entry_span_of_stack(ddtrace_span_stack *stack) {
             ddtrace_switch_span_stack(stack->parent_stack);
         }
 
-        if (get_DD_TRACE_AUTO_FLUSH_ENABLED() && ddtrace_flush_tracer(false) == FAILURE) {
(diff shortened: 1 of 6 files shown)


ext/span.c
```diff
@@ -671,10 +672,7 @@ void ddtrace_drop_span(ddtrace_span_data *span) {
 }
 
 void ddtrace_serialize_closed_spans(zval *serialized) {
-    array_init(serialized);
-
-    // We need to loop here, as closing the last span root stack could add other spans here
-    while (DDTRACE_G(top_closed_stack)) {
+    if (DDTRACE_G(top_closed_stack)) {
```
realFlowControl, line 675: **question:** Is this comment not true anymore and is this related to not calling a garbage collection cycle?

bwoebi: Yes, that's related to the garbage collection. The code in this function does not have any side effects by itself.

realFlowControl (approved):

PR #2050, opened 2023-05-12
feat: add environment variable mapping to datadog-setup.php

realFlowControl:
### Description

This PR will add the missing environment variable to INI setting mapping from #1951 to `datadog-setup.php`, so we do support the documented environment variables and map them to INI settings in for example the `config set` command.

This will allow the following use case:
```
$ php datadog-setup.php config set --php-bin=all DD_PROFILING_ENABLED=1
Searching for available php binaries, this operation might take a while.
Setting configuration for binary: php (/opt/php/8.2/bin/php)
Set 'datadog.profiling.enabled' to '1' in INI file: /opt/php/8.2/etc/conf.d/98-ddtrace.ini
```

Additionally this adds the `datadog-setup.php` to the PHP5 lint job and fixes some violations.

PROF-7617

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


datadog-setup.php
```diff
@@ -1630,6 +1645,35 @@ function ($line) {
...
 
+function map_env_to_ini($env)
+{
+    $setting = explode('_', $env, 3);
+    if ($setting[0] !== 'DD') {
+        return null;
+    }
+
+    $ini = 'datadog.';
+    switch ($setting[1]) {
```
morrisonlevi, line 1656: I suppose technically index 1 ought to have an isset guard as well.

realFlowControl: Done

datadog-setup.php
```diff
@@ -1630,6 +1645,35 @@ function ($line) {
...
+        case 'PROFILING':
+        case 'TRACE':
+        case 'APPSEC':
```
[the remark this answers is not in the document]
bwoebi: Yes, appsec has the same zai mechanisms than tracer and profiler.

datadog-setup.php
```diff
@@ -1630,6 +1645,39 @@ function ($line) {
...
+            $ini .= strtolower($setting[1]) . '.' . strtolower($setting[2]);
+            break;
+        default:
+            // for cases like DD_ENV or DD_DOGSTATSD_URL
+            $ini .= strtolower($setting[1]);
+            if (isset($setting[2])) {
+                $ini .= '_' . strtolower($setting[2]);
+            }
+    }
+    foreach (get_ini_settings('', '', '') as $setting) {
```
bwoebi, line 1673: Why are we checking the names here? get_ini_settings() does _not_ give you all valid ini settings, but only the selected ini settings for default inclusion in the INI files. E.g. `datadog-setup.php config set DD_TRACE_REPORT_HOSTNAME=0` will fail.

realFlowControl: The main reason was to make sure to give guidance to users. If you do a typo it will tell you and reduce possible support cases. I get the point though. I will think about how we could tackle this as I would like to see some kind of validation in here. OTOH we could also remove that check and see if there are actually support cases.
WDYT?

bwoebi (approved):
Looks good. But we likely will require PHP 5 compat for the datadog-setup for a long time (as the sole file), so that users may just grab the datadog-setup.php and install PHP 5 from any current version (though we'll pin the downloaded version when it's EOL).

PR #2058, opened 2023-05-16
Add snapshot testing

PROFeNoM:
### Description

Use the [test agent](https://github.com/DataDog/dd-apm-test-agent/tree/master#snapshot-testing) to add the snapshot testing feature. As such, snapshots for the Laminas _web framework_ were made.

To allow snapshot testing on the CI, the web tests' jobs (only) were moved from using docker executors to a machine executor, and the individual docker containers needed are now run through the `setup_docker` command with the `extra` parameter set to `with_snapshots`. Using a machine executor was necessary to allow volume mounting on the test agent's image so it could access the `tests/snapshots` directory.

Practically speaking, a (web) test with snapshot would use the new `tracesFromWebRequestSnapshot` method instead of `tracesFromWebRequest`. The logic behind this new method is basically the following:
- A token (function namespace) will be generated and a corresponding snapshot session started
- The test will be run (as before)
- The snapshot session will be stopped and the traces compared (agent-side)

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.


PROFeNoM:
I'll add a github action so the diff are easily seen when snapshots are getting regenerated

PROFeNoM:
Merged the two other branches (github workflow + removed attributes). Examples of the github action's output can be seen on these two PRs: #2078 & #2076.

tests/snapshots/tests.integrations.laminas.v1_4.common_scenarios_test.test_scenario_get_to_missing_route.json
```diff
@@ -0,0 +1,310 @@
+[[
+  {
+    "name": "laminas.request",
+    "service": "test_laminas_20",
+    "resource": "GET /does_not_exist",
+    "trace_id": 0,
+    "span_id": 1,
+    "parent_id": 1625545689840073902,
```
bwoebi, line 8: Why is this parent_id hardcoded? I wonder, are we incidentally doing distributed tracing from the testrunner or such?

build/_build.csproj.DotSettings
```diff
@@ -0,0 +1,27 @@
+<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
```
bwoebi, line 1: Do we need that file at all? Looks like some Rider specific file to configure project local values.

build/.editorconfig
```diff
@@ -0,0 +1,11 @@
+[*.cs]
```
bwoebi, line 1: We have an .editorconfig in our project root. Mind merging it there?

bwoebi (commented):
Is it possible to move build and build.{sh,cmd,ps1} into an adequately named folder? (like, what's build? build process of the tracer? no.)

It's just for github-actions, so maybe just put it in a github-actions/ dir?

build/_build.csproj.DotSettings
PROFeNoM: According to [Nuke's documentation](https://nuke.build/docs/getting-started/setup/), it indeed is a ReSharper/Rider formatting file, so removing it shouldn't break the workflow. 

However, I don't know whether if it should be removed or not. On [dd-trace-dotnet](https://github.com/DataDog/dd-trace-dotnet/blob/master/tracer/build/_build/_build.csproj.DotSettings), they aren't removing it.

tests/snapshots/tests.integrations.laminas.v1_4.common_scenarios_test.test_scenario_get_to_missing_route.json
PROFeNoM: I think that's because in our traces, we are setting a `parent_id` on the root span (even if that's the same as the `trace_id` in the root span). Therefore, on the [test-agent side](https://github.com/DataDog/dd-apm-test-agent/blob/793ec3c107ffcb420d924ad3ff9f1397f7ed106d/ddapm_test_agent/trace.py#L146), since it sees a `parent_id`, it uses this one.

Below is an example of a root span for the same query:
```
{
            "trace_id": "2161630347673372554",
            "span_id": "7241179534834786024",
            "name": "laminas.request",
            "resource": "Application\\Controller\\CommonSpecsController@simple simple",
            "service": "test_laminas_20",
            "start": 1685602874633867000,
            "error": 0,
            "type": "web",
            "duration": 339176000,
            "parent_id": "2161630347673372554",
            "meta": {
                "_dd.p.dm": "-1",
                "http.url": "http:\/\/localhost:9999\/simple?key=value&<redacted>",
                "http.method": "GET",
                "span.kind": "server",
                "component": "laminas",
                "http.version": "1.1",
                "http.route": "simple",
                "laminas.route.action": "Application\\Controller\\CommonSpecsController@simple",
                "http.status_code": "200"
            },
            "metrics": {
                "process_id": 12829,
                "_sampling_priority_v1": 1,
                "php.compilation.total_time_ms": 79.091
            }
        }
```

bwoebi: Other question: does the parent id disappear, if we add `DD_DISTRIBUTED_TRACING=0` to the phpunit runner?

PROFeNoM: Yep, running `DD_DISTRIBUTED_TRACING=0 DD_TRACE_CLI_ENABLED=true php  -d ddtrace.request_init_hook=/home/circleci/app/bridge/dd_wrap_autoloader.php ./tests/vendor/bin/phpunit  --config=./tests/phpunit.xml tests/Integrations/Laminas/V2_0 --filter=.`, I no longer have the `parent_id` in the root span (and `metrics._dd.rule_psr` now appears, just noting it).

build/_build.csproj.DotSettings
PROFeNoM: Removed :+1:

tests/snapshots/tests.integrations.laminas.v1_4.common_scenarios_test.test_scenario_get_to_missing_route.json
bwoebi: This would be sort of the expected output, and we need to change things so that phpunit itself will be called with `-d datadog.distributed_tracing=0`. But that's not for this PR.

bwoebi (approved):
Thanks for introducing snapshot testing! Looking forward at what it'll make easier for us :-)

PR #2065, opened 2023-05-19
build(ci): execute workflows based on which files are modified

morrisonlevi:
### Description

This uses [CircleCI's dynamic configuration to execute specific workflows based on which files are modified](https://circleci.com/docs/using-dynamic-configuration/#execute-specific-workflows-or-steps-based-on-which-files-are-modified). It does this by comparing the changed files to the `master` branch, so in a stacked PR scenario, you may run more tests than you expect.

The `build` and `build_packages` workflows are on by default, and currently no rules disable them. The workflows which are enabled this way are:

- components
- zend_abstract_interface
- profiling (new, this is just the profiling tests extracted to a workflow so they can be conditionally run and tracked individually)

Note that the eclint job is broken when ran from a CircleCI continuation because of the gulp-reporter, which hard-codes the configuration as `.circleci/config.yml`, which is broken by using `.circleci/continue_config.yml`. It seems [eclint is unmaintained](https://github.com/jednano/eclint/issues/226), and has been for around 5 years now. We could copy the lint job to the main workflow if we want, or we could drop the `eclint` check instead. For now I have removed it.

These PRs demonstrate the filters work correctly for the relevant workflows (but disabled the build workflows to save credits, as they aren't needed).

- #2066
- #2067
- #2068

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ Draft PRs test the job continuations, but this just a one time test.

### Reviewer checklist
- [x] Milestone is set.


bwoebi (approved):
Trusting that it's working, and the pipeline conditions seem sensible.

Issue #1840, opened 2022-12-28
[Feature] php-amqplib Integration


orlandothoeny:
As a developer, I'd like to see information about RabbitMQ messages inside traces.

**Solution**
Support for the [php-amqplib](https://github.com/php-amqplib/php-amqplib) library, which is as far as I could research the most widely used RabbitMQ client for PHP.

**Alternatives I've considered**
Implementing custom instrumentation myself. I would prefer to have this out of the box though.

**Interesting methods**
Methods I personally am interested in:

[AMQPChannel](https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Channel/AMQPChannel.php)
* `PhpAmqpLib\Channel\AMQPChannel::basic_publish()`
* `PhpAmqpLib\Channel\AMQPChannel::basic_consume()`
* `PhpAmqpLib\Channel\AMQPChannel::basic_get()`
* `PhpAmqpLib\Channel\AMQPChannel::basic_ack()`
* `PhpAmqpLib\Channel\AMQPChannel::basic_nack()`
* `PhpAmqpLib\Channel\AMQPChannel::basic_cancel()`

Especially interesting parameters: Exchange, Queue, Routing Key

[AbstractConnection](https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Connection/AbstractConnection.php)
* `PhpAmqpLib\Connection\AbstractConnection::connect()`
* `PhpAmqpLib\Connection\AbstractConnection::reconnect()`

It might make sense to extend this list.

morrisonlevi:
Thank you for including some interesting bits for us to investigate with this feature request!

bwoebi:
Hey @orlandothoeny,

We've just released 0.87.0 which includes an php-ampqlib integration!

Feel free to try it out and tell us whether it satisfies your expectations :-)

orlandothoeny:
Hi @bwoebi,

Cool, great work!

What I observed in our setup:
We only have 2 resources: `connect` and one where we had a connection error and the whole PHP stack trace is used as resource name. But no `basic_ack` or `basic_get`, etc. which we also call.

![image](https://github.com/DataDog/dd-trace-php/assets/17168783/4ac16429-1053-4436-a009-e1335145bc13)

orlandothoeny:
Hi @PROFeNoM

Now we have proper resources, I can't reproduce it anymore.

PR #2069, opened 2023-05-22
Fix error stack in span's resource when a AMQP connection error occurs

PROFeNoM:
### Description

Bugfix #1840. The call to `setGenericTags` was made using the exception as the resource detail in some cases. It will also retrieve the protocol version in the prehook (instead of the posthook) of the `AbstractConnection::_construct` method, since it wouldn't have been otherwise retrieved when a connection error would occur.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi (approved):
Thanks, looks good :-)

PR #2072, opened 2023-05-22
Work around fiber->execute_data only being correct upon yielding

bwoebi:
### Description

Fixes language tests.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ Existing tests exposed the bug.

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (approved):
🙈

Someone more qualified like Joe should review this when they have time, but it fixes the issue and exposes no new ones so it seems to be an improvement, at least.

PR #2073, opened 2023-05-23
Update removed methods used in Guzzle V6 & V7 test suites

PROFeNoM:
### Description

Change some methods in the guzzle test suites that were removed with the v2 of guzzle promises:
- [all](https://github.com/guzzle/promises/blob/136a635e2b4a49b9d79e9c8fee267ffb257fdba0/src/functions.php#L177): `Promises\all` --> `Utils::all`
- [is_settled](https://github.com/guzzle/promises/blob/136a635e2b4a49b9d79e9c8fee267ffb257fdba0/src/functions.php#L346): `Promises\is_settled`--> `Is:settled`

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.

### DIFF
--- a/tests/composer.json
+++ b/tests/composer.json
@@ -65,7 +65,8 @@
             },
             "guzzle7": {
                 "require": {
-                    "guzzlehttp/guzzle": "~7.0"
+                    "guzzlehttp/guzzle": "~7.0",
+                    "guzzlehttp/promises": "~2.0"
                 },
                 "scenario-options": {
                     "create-lockfile": false
--- a/tests/composer.json
+++ b/tests/composer.json
@@ -65,8 +65,7 @@
             },
             "guzzle7": {
                 "require": {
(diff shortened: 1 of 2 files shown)


tests/composer.json
```diff
@@ -65,8 +65,7 @@
             },
             "guzzle7": {
                 "require": {
-                    "guzzlehttp/guzzle": "~7.0",
-                    "guzzlehttp/promises": "^1.5.3"
+                    "guzzlehttp/guzzle": "~7.0"
```
realFlowControl, line 68: **suggestion:** As we are using `guzzlehttp/promises` in our code we might keep it in the `composer.json` to make this explicit. Otherwise we are relying on Guzzle bringing this dependency, which we do not control.

```suggestion
                    "guzzlehttp/guzzle": "~7.0",
                    "guzzlehttp/promises": "~2.0"
```

realFlowControl (approved):

PR #2084, opened 2023-05-31
docs(tests): add ASAN section to randomised tests `README.md`

realFlowControl:
### Description

This will add a section about ASAN and how to recompile PHP with it in case you need to debug memory issues.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/tests/randomized/README.md
+++ b/tests/randomized/README.md
@@ -156,6 +156,46 @@ Then load it in `gdb`:
 gdb --core=/tmp/core php-fpm|httpd|php
 ```
 
+### Analyzing using ASAN
+
+In some cases it might be tricky to find the memory corruption, but you can
+always use an address sanitizer to try and find problems.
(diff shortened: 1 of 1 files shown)


tests/randomized/README.md
```diff
@@ -156,6 +156,46 @@ Then load it in `gdb`:
 gdb --core=/tmp/core php-fpm|httpd|php
 ```
 
+### Analyzing using ASAN
+
+In some cases it might be tricky to find the memory corruption, but you can
+always use an address sanitizer to try and find problems.
```
bwoebi, line 162: Could you please add one sentence that the installer needs be be fetched from the `package extension zts-debug-asan` step instead of `package extension` if the tracer itself should also be linked against asan?

bwoebi (approved):

PR #2100, opened 2023-06-14
Add automatic Laravel 5.8 user tracking events

estringana:
### Description

First ATO integration for PHP login success, login failure, and signup events for Laravel.

The event model is designed to be consistent across frameworks: same semantics, same span tags, regardless of how the framework internally handles authentication. Laravel was picked first based on customer usage data (largest share of PHP AppSec customers at the time).

The wrapped methods are:
- `Illuminate\Events\Dispatcher::fire`: This generic method was already wrapped for some other purposes. I filtered within this method the event `eloquent.created: User` to trigger appsec event `signup` on Laravel 4
- `Illuminate\Auth\SessionGuard::attempt`: This method will trigger the appsec event login attempt failure on Laravel >=5
- `Illuminate\Auth\Guard::attempt`:  This method will trigger the appsec event login attempt failure on Laravel <5
- `Illuminate\Auth\SessionGuard::setUser`: This method will trigger the appsec event login succesful on Laravel >=5
- `Illuminate\Auth\Guard::setUser`: This method will trigger the appsec event login succesful on Laravel <5
- `Illuminate\Auth\Events\Registered::__construct`: This method will trigger the appsec event sign up

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/Laravel/LaravelIntegration.php
+++ b/src/Integrations/Integrations/Laravel/LaravelIntegration.php
@@ -389,7 +389,7 @@ class LaravelIntegration extends Integration
             }
         );
 
-        // Used by Laravel >= 5.0
+        // Used by Laravel < 5.0
         \DDTrace\hook_method(
             'Illuminate\Auth\Guard',
             'attempt',
--- a/src/Integrations/Integrations/Laravel/LaravelIntegration.php
+++ b/src/Integrations/Integrations/Laravel/LaravelIntegration.php
@@ -166,24 +166,20 @@ class LaravelIntegration extends Integration
 
                     //New user created, assume sign up
                     if ($span->resource == 'eloquent.created: User') {
-                        if (!function_exists('\datadog\appsec\track_user_signup_event'))
-                        {
+                        $authClass = '\User';
+                        if (
+                            !function_exists('\datadog\appsec\track_user_signup_event') ||
+                            !isset($args[1]) ||
+                            !$args[1] ||
+                            !($args[1] instanceof $authClass)
+                        ) {
                             return;
                         }
-                        if (!isset($args[1])) {
-                            return;
-                        }
-                        $user = $args[1];
-                        $authClass = 'User';
-                        if (!$user || !($user instanceof $authClass)) {
-                            return;
-                        }
-                        $metadata = [];
                         $id = null;
-                        if (isset($user['id'])) {
-                          $id = $user['id'];
+                        if (isset($args[1]['id'])) {
+                          $id = $args[1]['id'];
                         }
-                        \datadog\appsec\track_user_signup_event($id, $metadata, true);
+                        \datadog\appsec\track_user_signup_event($id, [], true);
                     }
 
                 },
@@ -317,11 +313,7 @@ class LaravelIntegration extends Integration
             'attempt',
             null,
             function ($This, $scope, $args, $loginSuccess) use ($rootSpan, $integration) {
-                if ($loginSuccess) {
-                    return;
-                }
-
-                if (!function_exists('\datadog\appsec\track_user_login_failure_event'))
+                if ($loginSuccess || !function_exists('\datadog\appsec\track_user_login_failure_event'))
                 {
                     return;
                 }
@@ -334,28 +326,23 @@ class LaravelIntegration extends Integration
             'Illuminate\Auth\SessionGuard',
             'setUser',
             function ($This, $scope, $args) use ($rootSpan, $integration) {
-                if (!function_exists('\datadog\appsec\track_user_login_success_event'))
-                {
-                    return;
-                }
-                if (!isset($args[0])) {
-                    return;
-                }
-                $user = $args[0];
-                $authClass = 'Illuminate\Contracts\Auth\Authenticatable';
-                if (!$user || !($user instanceof $authClass)) {
-                    return;
-                }
-
-                $metadata = [];
-                if (isset($user['name'])) {
-                    $metadata['name'] = $user['name'];
-                }
-                if (isset($user['email'])) {
-                    $metadata['email'] = $user['email'];
-                }
-
-                \datadog\appsec\track_user_login_success_event($user->getAuthIdentifier(), $metadata, true);
+                $authClass = '\Illuminate\Contracts\Auth\Authenticatable';
+                if (
+                   !function_exists('\datadog\appsec\track_user_login_success_event') ||
+                   !isset($args[0]) ||
+                   !$args[0] ||
+                   !($args[0] instanceof $authClass)
+               ) {
+                   return;
+               }
+               $metadata = [];
+               if (isset($args[0]['name'])) {
+                   $metadata['name'] = $args[0]['name'];
+               }
+               if (isset($args[0]['email'])) {
+                   $metadata['email'] = $args[0]['email'];
+               }
+               \datadog\appsec\track_user_login_success_event($args[0]->getAuthIdentifier(), $metadata, true);
             }
         );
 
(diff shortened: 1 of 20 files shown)


estringana:
You can start the review but plz don't merge yet

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -163,6 +163,29 @@ function ($This, $scope, $args) use ($rootSpan, $integration) {
                     $span->service = $integration->getServiceName();
                     $span->resource = $args[0];
                     $span->meta[Tag::COMPONENT] = LaravelIntegration::NAME;
+
+                    //New user created, assume sign up
+                    if ($span->resource == 'eloquent.created: User') {
```
estringana, line 168: On Laravel 4, this is the only way I found I can detect when a user is created. It does not filter only sign up. Open to listen new approaches

PROFeNoM: From my experience, in 'Vanilla Laravel,' the problem is that the registration process can be completely customized (e.g., with a registration controller responsible for creating a `User`). However, one thing is certain: when a user signs up, their account information is typically stored in a database table or another data source.

I can't think of a way to differentiate between a `User` created as a result of registration and one created through another process...

💭 However, isn't the only one capable of creating a new `User` outside of the registration process the admin? Is that truly an issue?

estringana: I don't think that's truly an issue neither. This project is aiming to automate user events as much as possible. Developers doing custom approaches will have to still use the SDK.

I was checking with you in case there was a more elegant approach on Laravel 4 but if not, this is the way :)

Thanks for your answer

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -288,6 +311,125 @@ function ($exceptionHandler, $scope, $args) use ($rootSpan, $integration) {
             }
         );
 
+        // Used by Laravel >= 5.0
+        \DDTrace\hook_method(
+            'Illuminate\Auth\SessionGuard',
+            'attempt',
+            null,
+            function ($This, $scope, $args, $loginSuccess) use ($rootSpan, $integration) {
+                if ($loginSuccess) {
```
Anilm3, line 320: Is there enough information to call `track_user_login_success_event` here or is there an advantage to having a separate hook for it?

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -288,6 +311,125 @@ function ($exceptionHandler, $scope, $args) use ($rootSpan, $integration) {
...
+                    return;
+                }
+
+                if (!function_exists('\datadog\appsec\track_user_login_failure_event'))
+                {
+                    return;
+                }
+                \datadog\appsec\track_user_login_failure_event(null, false, [], true);
```
Anilm3, line 328: Is it not possible to provide the user id in the event of login failure? Or does the RFC suggest otherwise?

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -288,6 +311,125 @@ function ($exceptionHandler, $scope, $args) use ($rootSpan, $integration) {
...
+                }
+                if (isset($user['email'])) {
+                    $metadata['email'] = $user['email'];
+                }
+
+                \datadog\appsec\track_user_login_success_event($user->getAuthIdentifier(), $metadata, true);
+            }
+        );
+
+        // Used by Laravel >= 5.0
```
Anilm3, line 392: Should it be `< 5.0`?

bateller (approved):

src/Integrations/Integrations/Laravel/LaravelIntegration.php
estringana: Milestone 0 does not require id on this event. I have an idea of how to get it for failure logins but I will do that on another PR.

src/Integrations/Integrations/Laravel/LaravelIntegration.php
estringana: You're correct

Anilm3 (approved):
lgtm

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -288,6 +311,125 @@ function ($exceptionHandler, $scope, $args) use ($rootSpan, $integration) {
...
+            function ($This, $scope, $args) use ($rootSpan, $integration) {
+                if (!function_exists('\datadog\appsec\track_user_login_success_event'))
+                {
+                    return;
+                }
+                if (!isset($args[0])) {
+                    return;
+                }
+                $user = $args[0];
+                $authClass = 'Illuminate\Auth\UserInterface';
```
realFlowControl, line 375: I was wondering if this should not be rather this (with the `\` in the beginning):
```suggestion
                $authClass = '\Illuminate\Auth\UserInterface';
```
But there seems some magic going on ...
https://3v4l.org/G1UNq

In the function bellow (following the same schema), you where adding the `\` at the beginning, maybe we should do it here as well. Although it looks like its working, I would think it should not 😆 
The same applies to the function above this one

src/Integrations/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -288,6 +311,125 @@ function ($exceptionHandler, $scope, $args) use ($rootSpan, $integration) {
...
+                    return;
+                }
+                $user = $args[0];
+                $authClass = '\Illuminate\Contracts\Auth\Authenticatable';
+                if (!$user || !($user instanceof $authClass)) {
+                    return;
+                }
+
+                $metadata = [];
+                \datadog\appsec\track_user_signup_event($user->getAuthIdentifier(), $metadata, true);
```
realFlowControl, line 429: **nitpick (non-blocking):** this is just a coding style thing
```suggestion
                if (!function_exists('\datadog\appsec\track_user_signup_event') ||
                    !isset($args[0]) ||
                    !$args[0] ||
                    !($args[0] instanceof \Illuminate\Contracts\Auth\Authenticatable))
                {
                    return;
                }
                \datadog\appsec\track_user_signup_event($args[0]->getAuthIdentifier(), [], true);
```
I do not have any hard feelings about this and I do not block the PR because of this, If you like to change this, you could also do in a separate PR later.

realFlowControl (approved):
I'll approve, but there are some comments especially the missing `\` in the `$authClass` variables is interesting, no idea why it is working at all 😉 
Most likely some PHP special kicking in.

src/Integrations/Integrations/Laravel/LaravelIntegration.php
estringana: Thanks for this. I will change it everywhere

src/Integrations/Integrations/Laravel/LaravelIntegration.php
estringana: Since I was doing the other changes on this file I have changed this and all other functions I added :)

bwoebi (approved):

src/Integrations/Integrations/Laravel/LaravelIntegration.php
bwoebi: The better style would be rather the opposite - stringified names are always fully qualified. I fact `\Illuminate\Auth\UserInterface::class` will give you `'Illuminate\Auth\UserInterface'`.

realFlowControl: TIL

estringana: ok, I will remove the initial `\` from all these places

PR #2119, opened 2023-06-20
feat(profiling): add file compilation to timeline

realFlowControl:
### Description

This PR will add the file compilation time for files included via the `include()`/`require()` function family. We do so by hooking into the `zend_compile_file` hook. This PR will also send the filename and the include type (`include`/`require`) with the event.

This PR will _not_ add support for code that is `eval()`'ed.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/timeline.rs
```diff
@@ -1,20 +1,84 @@
...
+
+            let include_type = match r#type as u32 {
+                zend::ZEND_INCLUDE => "include",
+                zend::ZEND_REQUIRE => "require",
+                _default => "",
+            };
+
+            cfg_if::cfg_if! {
+                if #[cfg(php_zend_stream_api_uses_zend_string)] {
+                    let filename = Some(String::from_utf8_lossy(
```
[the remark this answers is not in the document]
realFlowControl: Well yes, but actually:
```
error[E0658]: attributes on expressions are experimental
  --> profiling/src/timeline.rs:53:28
   |
53 |               let filename = cfg_if::cfg_if! {
   |  ____________________________^
54 | |                 if #[cfg(php_zend_stream_api_uses_zend_string)] {
55 | |                     Some(String::from_utf8_lossy(
56 | |                         ddog_php_prof_zend_string_view((*handle).filename.as_mut()).into_bytes(),
...  |
61 | |                 }
62 | |             };
   | |_____________^
```

Regarding "out of scope": It is not only compiling, it also works 🎉 
I would think, that all the `cfg`-stuff is evaluated before compilation so when compiling it either the one or the other `let filename = ...` thats there and nothing else. I could be wrong, but it's working.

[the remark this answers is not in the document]
morrisonlevi: Right, I knew it works, I was just trying to say that my brain was like "wait, that should be out of scope" ^_^

profiling/src/timeline.rs
```diff
@@ -1,20 +1,84 @@
...
+            let locals = locals.unwrap();
+
+            if !locals.profiling_experimental_timeline_enabled {
+                return;
+            }
+
+            let include_type = match r#type as u32 {
+                zend::ZEND_INCLUDE => "include",
+                zend::ZEND_REQUIRE => "require",
+                _default => "",
```
morrisonlevi, line 50: Eval goes here too, right?

profiling/src/profiling/mod.rs
```diff
@@ -665,6 +665,54 @@ impl Profiler {
...
+        }
+
+        labels.extend_from_slice(&TIMELINE_COMPILE_FILE_LABELS);
+        labels.push(Label {
+            key: "filename",
+            value: LabelValue::Str(Cow::from(filename)),
+        });
+        let n_labels = labels.len();
+
+        match self.send_sample(Profiler::prepare_sample_message(
```
morrisonlevi, line 692: I realized just now that there's a decent number of lines of code between the measuring of the elapsed time and preparing the sample message, which is the timestamp. Do you think we should push the system time to the caller and pass it around?

profiling/src/profiling/mod.rs
```diff
@@ -665,6 +665,54 @@ impl Profiler {
...
+            }],
+            SampleValues {
+                timeline: duration,
+                ..Default::default()
+            },
+            labels,
+            locals,
+        )) {
+            Ok(_) => {
+                trace!("Sent event 'compile file' with {n_labels} and labels to profiler.")
```
morrisonlevi, line 706: There's a superfluous `and` here:

```suggestion
                trace!("Sent event 'compile file' with {n_labels} labels to profiler.")
```

profiling/src/timeline.rs
```diff
@@ -9,12 +10,66 @@ use std::time::Instant;
...
+                    filename.to_string(),
+                    include_type,
+                    &locals,
+                );
+            }
+        });
+
+        return op_array;
     }
+    panic!("should not happen");
```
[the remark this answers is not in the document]
morrisonlevi: I suppose use a `warn!` and do nothing, but yes, this is likely to be a very serious bug... somewhere.

profiling/src/timeline.rs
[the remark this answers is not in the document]
realFlowControl: I was not able to get this structure to work, so I fell back to:
```rust
#[cfg(php_zend_stream_api_uses_zend_string)]
let filename = Some(String::from_utf8_lossy(
    ddog_php_prof_zend_string_view((*handle).filename.as_mut()).into_bytes(),
)).unwrap();
#[cfg(not(php_zend_stream_api_uses_zend_string))]
let filename = CStr::from_ptr((*handle).filename).to_str().unwrap();
```

profiling/src/profiling/mod.rs
realFlowControl: Thats a good idea 👍

realFlowControl: I'll created a task for this (PROF-7844) and will start working on it after this is merged.

morrisonlevi:
Is cardinality an issue on this at all? That's the only concern I can think of for the filename, since we already send and display filenames in profiles.

profiling/src/profiling/mod.rs
morrisonlevi: Sounds good! Remember to take the sample time after the event's duration has measured (timestamps match the end of the event, not the beginning).

morrisonlevi (approved):
Approved, assuming my question about filename cardinality is not an issue.

realFlowControl:
Thanks for the hint. I did a deeper investigation and found something that needs fixing!
In the pprof file we generate most (if not all) of the filenames are in the string tables due to stack walking anyways and the labels will just reference them.
**BUT:** The `filename` in the `zend_compile_file`-hook is not a normalized version, but whatever the caller gave to the `include()` function. So in case of Symfony you would see a lot if filenames like `/www/symfony-demo/vendor/composer/../symfony/polyfill-intl-icu/IntlDateFormatter.php` but when stack walking you get the normalized version wich is `/www/symfony-demo/vendor/symfony/polyfill-intl-icu/IntlDateFormatter.php` (notice the missing `composer/../` in the second version).
So prior to merging this I will investigate and see if I can find other cases and write a normalizer.

PR #2120, opened 2023-06-21
ci: make sure core file is where it should be

realFlowControl:
### Description

In the internal API randomized tests the core file (in case a segfault happens) would not be in `/tmp/core`,  because `/proc/sys/kernel/core_pattern` is set to `|/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E` on the host which is inherited by the container. Now inside the container the `apport` binaray does not exists, so core dumps are lost.
We can not change the core pattern from inside the container, as this is not a privileged one, so we do it as an extra step from the outside.

Find the failed job here that did not have a core dump in the artifacts:
https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/11602/workflows/0ac9f7b9-8200-4c94-ae58-7edd648f90fc/jobs/2461257/steps

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -696,6 +696,10 @@ jobs:
             - source-v1-{{ .Branch }}-{{ .Revision }}
       - <<: *STEP_CHECKOUT
       - <<: *STEP_ATTACH_WORKSPACE
+      - run:
+          name: Set core pattern
+          command: |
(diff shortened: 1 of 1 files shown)


morrisonlevi (approved):
It looks like it would work, but I don't know 🤷🏻. Feel free to merge!

PR #2124, opened 2023-06-23
Add `horizon` as artisan queue command

markwalet:
### Description

The default horizon work command is `php artisan horizon`. Which doesn't get recognised right now.

I've chosen to do an exact match, because using `strpos()` will match on all Horizon subcommands.

I'm not familiar with the codebase, but I did some quick testing on the `$artisanCommand` variable locally and saw that it contains the command without any options. So this should work I think. I'm not sure why we aren't using an exact match (or `in_array()`) for all options. Is this speed related?

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/Laravel/LaravelIntegration.php
+++ b/src/Integrations/Integrations/Laravel/LaravelIntegration.php
@@ -44,7 +44,8 @@ class LaravelIntegration extends Integration
 
         return !empty($artisanCommand)
             && (strpos($artisanCommand, 'horizon:work') !== false
-                || strpos($artisanCommand, 'queue:work') !== false);
+                || strpos($artisanCommand, 'queue:work') !== false
+                || $artisanCommand === 'horizon');
     }
 
     /**


PROFeNoM:
Hi Mark :wave:

First of all, thanks for bringing it up! After some testing on my side as well, there was indeed a confusion around the relationship between `Artisan::call` and `$_SERVER['argv']`, which was the reason for using `strpos` instead of an exact match. 

Therefore, it appears that checking exact matches with `in_array` is just the easiest way to go. I would imagine something like this:
```php
public function isArtisanQueueCommand()
{
    $artisanCommand = isset($_SERVER['argv'][1]) ? $_SERVER['argv'][1] : '';
    return in_array($artisanCommand, ['horizon:work', 'queue:work', 'horizon', 'horizon:supervisor']);
}
```

Would it corresponds to what you were thinking as well? If it does, either you or I can do the commit and merge this PR 😃

PROFeNoM (approved):
Thanks, Mark!

PR #2130, opened 2023-06-28
Add SpanData::peerServiceSources

labbati:
### Description

The new `SpanData` field `peerServiceSources` will be used by integrations to notify the spans post-processor that the span requires `peer.service` to be set. If any (in the provided order) of the tags will be present in the span, then its value will be used to set `peer.service`.

Note for the reviewer:
- It is **OK** if `var_dump($span)` changes and adds a `peerServiceSources` field;
- It is **NOT OK**  if `peerServiceSources` is added to the serialized span, e.g. `var_dump(dd_trace_serialize_closed_spans());`

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


PROFeNoM (approved):
LGTM 😃

ext/ddtrace.stub.php
```diff
@@ -105,6 +105,11 @@ class SpanData {
          */
         public array $links = [];
 
+        /**
+         * @var string[] $peerServiceSources A sorted list of tag names used to set peer.service tag
```
bwoebi, line 109: I think here's place for an extended documentation comment as to how peerServiceSources works in detail :-)

labbati: @bwoebi  Good idea. Are you cool with https://github.com/DataDog/dd-trace-php/pull/2130/commits/fd8a9a3a2d8d54919ce694c5b90433e8fdd507fd

bwoebi (approved):
Looks good, but please hold off the merge until the full feature is ready :-)

PR #2132, opened 2023-06-29
Automated user events tracking symfony 5.2

estringana:
### Description

ATO support for Symfony: login success, login failure, and signup events.

Same event model as [#2100](https://github.com/DataDog/dd-trace-php/pull/2100) (Laravel). Symfony uses a different authentication architecture (Security component, event listeners) so the instrumentation approach differs, but the output is identical from AppSec's perspective.

Methods wrapped:
- `Doctrine\ORM\UnitOfWork::executeInserts`: This method will trigger the appsec event signup
- `Symfony\Component\Security\Http\Firewall\AbstractAuthenticationListener::onFailure`:  This method will trigger the appsec event login attempt failure
- `Symfony\Component\Security\Http\Firewall\AbstractAuthenticationListener::onSuccess`: This method will trigger the appsec event login succesful

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/src/Integrations/Integrations/Symfony/SymfonyIntegration.php
+++ b/src/Integrations/Integrations/Symfony/SymfonyIntegration.php
@@ -86,6 +86,10 @@ class SymfonyIntegration extends Integration
             'Doctrine\ORM\UnitOfWork',
             'executeInserts',
             function ($This, $scope, $args) {
+                  if (!function_exists('\datadog\appsec\track_user_signup_event'))
+                  {
+                    return;
+                  }
                   $metadataClass = 'Doctrine\ORM\Mapping\ClassMetadata';
                   if (
                        !$args ||
--- a/src/Integrations/Integrations/Symfony/SymfonyIntegration.php
+++ b/src/Integrations/Integrations/Symfony/SymfonyIntegration.php
@@ -83,6 +83,42 @@ class SymfonyIntegration extends Integration
         );
 
         \DDTrace\hook_method(
+            'Doctrine\ORM\UnitOfWork',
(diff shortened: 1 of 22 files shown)


src/Integrations/Integrations/Symfony/SymfonyIntegration.php
```diff
@@ -82,6 +82,77 @@ public function init()
...
+                    }
+                    $found++;
+                    $userEntity = $entity;
+                  }
+
+                  if ($found != 1) {
+                    return;
+                  }
+
+                  \datadog\appsec\track_user_signup_event($userEntity->getUsername(), [], true);
```
PROFeNoM, line 115: Just to be sure, so here we are going through **all** the entity insertions, and only tracking the signup event on **the last one**? Can't we directly grab the last element of the `entityInsertions` array? Maybe there is something I understood wrong there 🤔

estringana: What this is trying to achieve is to trigger `track_user_signup_event` event when there is a sign up. On the loop I need to go through all the entities and check if there is one and only one user entity.  If there are multiple, it probably is a mass user creation process or something else but not a sign up

bwoebi, line 115: function_exists check missing

estringana: Good one! I have just changed it and also used this opportunity to change it on the Laravel integration we already had

bwoebi (approved):

PR #2134, opened 2023-06-29
Add peer.service to http client instrumentations

labbati:
### Description

Add peer.service to http client instrumentations

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/api/Tag.php
```diff
@@ -8,6 +8,11 @@ class Tag
     const ENV = 'env';
     const SPAN_TYPE = 'span.type';
     const SPAN_KIND = 'span.kind';
+    const SPAN_KIND_VALUE_SERVER = 'server';
+    const SPAN_KIND_VALUE_CLIENT = 'client';
+    const SPAN_KIND_VALUE_PRODUCER = 'producer';
+    const SPAN_KIND_VALUE_CONSUMER = 'consumer';
+    const SPAN_KIND_VALUE_INTERNAL = 'internal';
```
labbati, line 15: This can be controversial (so far we have used `Tag.php` for keys only. I do not consider it a problem, but happy to revert back if the reviewer is not happy with this approach.

bwoebi: I think it's fine :-)

src/Integrations/Integrations/Curl/CurlIntegration.php
```diff
@@ -69,8 +70,13 @@ public function init()
                 $info = \curl_getinfo($ch);
                 $sanitizedUrl = \DDTrace\Util\Normalizer::urlSanitize($info['url']);
                 $normalizedPath = \DDTrace\Util\Normalizer::uriNormalizeOutgoingPath($info['url']);
+                $host = Urls::hostname($sanitizedUrl);
                 unset($info['url']);
 
+                if ($host) {
```
bwoebi, line 76: Checking `Urls::hostname` it will never fail, so this if is meaningless, unless you mean to avoid parse failures?
However, this needs to check for `$host != "unknown-host" && $host != "unparsable-host"`.

labbati: Good catch. I think there is still value in reporting which services (in the service map) reports to the `unknown/unparseable-host`, so I think I'll add it in any case, rather than pass silently.

bwoebi (approved):
Looks good to me now :-)

labbati:
Ignoring flaky randomized tests as we wait for containers to be rebuild. cc @bwoebi

PR #2135, opened 2023-06-29
Handle peer.service-related post-process in the extension

PROFeNoM:
### Description

This PR processes the newly introduced `peerServiceSources` during the serialization phase.

Double-check the tests so we're sure we are on the same page from a functional pov 😃 

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.


ext/serializer.c
```diff
@@ -607,6 +607,28 @@ static zend_string *dd_get_user_agent() {
...
+    zend_array *peer_service_mapping = get_DD_TRACE_PEER_SERVICE_MAPPING();
+    if (zend_hash_num_elements(peer_service_mapping) == 0) {
+        return NULL;
+    }
+
+    // The mappings of peer_service_mapping are strings in the format <peer_service>:<mapped_service>
+    zend_string *mapped_service = NULL;
+    zend_string *mapping;
+    ZEND_HASH_FOREACH_STR_KEY(peer_service_mapping, mapping) {
+        if (mapping && memcmp(ZSTR_VAL(mapping), ZSTR_VAL(peer_service), ZSTR_LEN(peer_service)) == 0) {
```
labbati, line 620: I think that this would be true, while it should be false, in the following scenario:
- `peer_service = something`
- `mapping = something_and_something_else:hello`

tests/ext/peer_service_remapping.phpt
```diff
@@ -0,0 +1,60 @@
+--TEST--
+Properly remap peer service's sources based on DD_TRACE_PEER_SERVICE_MAPPING
+--ENV--
+DD_TRACE_PEER_SERVICE_MAPPING=foo:bar,db.instance:db,xyz:abc,db1:database,net.peer.name:net
```
labbati, line 4: Looking at the code I think this case is well covered, but I would also add a test case where users set a wrong value, e.g. `DD_TRACE_PEER_SERVICE_MAPPING=foo=bar,only_tag` and verify that your code does not break.

labbati (changes requested):
I could only find one edge case that I believe is not covered and added it to a comment. Let me know if that makes sense to you. In case it holds true, I would also add a test for it.

ext/serializer.c
PROFeNoM: Indeed!

labbati (approved):
Now it looks good from my point of view, and I am marking it as approved 👍🏻, even thus I'd prefer to have a C expert to take a look at it! Thanks @PROFeNoM!

ext/configuration.h
```diff
@@ -109,6 +109,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(STRING, DD_SPAN_SAMPLING_RULES_FILE, "", .ini_change = ddtrace_alter_sampling_rules_file_config)    \
     CONFIG(SET_LOWERCASE, DD_TRACE_HEADER_TAGS, "")                                                            \
     CONFIG(INT, DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH, "512")                                                     \
+    CONFIG(SET, DD_TRACE_PEER_SERVICE_MAPPING, "")                                                             \
```
labbati, line 112: `peer.service` should be _opt-in_, based on the spec (for now, as we want to enable it by default as we pass private beta). However, this is not a blocker for me to approve this PR, as I can add it the `mysqli` work https://github.com/DataDog/dd-trace-php/pull/2142

labbati (commented):
I did not notice, with the first review, that peer.service was not opt-in. Not a blocker to me

ext/configuration.h
PROFeNoM: What should be the default value in this case?

bwoebi, line 112: We have a config type MAP (instead of SET) which will split comma separated key:value into a key/value array. This seems more suitable here than doing string parsing to the next colon in iteration later.

ext/serializer.c
```diff
@@ -828,6 +856,33 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
         add_assoc_str(meta, "_dd.span_links", buf.s);
     }
 
+    zend_array *peer_service_sources_zv = ddtrace_spandata_property_peer_service_sources(span);
```
bwoebi, line 859: can you please not use a _zv suffix for something not a zval? That's confusing :-D

ext/serializer.c
```diff
@@ -828,6 +856,33 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
...
+            if (Z_TYPE_P(tag) == IS_STRING) {
+                // Use the first tag that is found in the span, if any
+                zval * value = zend_hash_find(Z_ARRVAL_P(meta), Z_STR_P(tag));
+                if (value) {
+                    add_assoc_str(meta, "_dd.peer.service.source", zend_string_copy(Z_STR_P(tag)));
+
+                    zend_string * substitute_value = dd_get_mapped_peer_service(Z_STR_P(value));
+                    if (substitute_value) {
+                        add_assoc_str(meta, "peer.service", substitute_value);
+                        add_assoc_str(meta, "peer.service.remapped_from", zend_string_copy(Z_STR_P(value)));
```
bwoebi, line 874: This logic is missing in the explicitly set peer.service case. I assume it should be remapping then, too?

ext/configuration.h
labbati: `false`

PROFeNoM: One thing, though, is that the `peer_service_wrong_values.phpt` test would now fail because of the wrong formatting `foo=bar,only_tag,foo:bar` (the last `foo:bar` won't get interpreted). If this is what we want, then great, else I would need to change `zai_config_decode_map`.

bwoebi: `bar,only_tag,foo:bar` will be parsed as `bar,only_tag,foo` key, `bar` value, right?

But I don't care too much about that case, it's wrong, it should give some not totally wrong value and not crash.

PROFeNoM: > bar,only_tag,foo:bar will be parsed as bar,only_tag,foo key, bar value, right?

Well, yes indeed... See that way...

Alright then :+1:

PROFeNoM: Alright, added `DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED`

ext/serializer.c
PROFeNoM: Indeed, it was missing. I changed the logic a bit (refactored) to make use of the `MAP` configuration type

ext/serializer.c
```diff
@@ -821,13 +838,45 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
...
-    if (zend_hash_num_elements(span_links_zv) > 0) {
+    zend_array *span_links = ddtrace_spandata_property_links(span);
+    if (zend_hash_num_elements(span_links) > 0) {
         smart_str buf = {0};
-        _dd_serialize_json(span_links_zv, &buf, 0);
+        _dd_serialize_json(span_links, &buf, 0);
         add_assoc_str(meta, "_dd.span_links", buf.s);
     }
 
+    if (get_DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED()) { // opt-in
```
bwoebi, line 848: shall this apply to the whole block or only for the automated `peer_service_sources` mapping? (i.e. line 858 currently)

PROFeNoM: From my understanding of the RFC (Service Representation, part FAQ: Tag peer.service), I think this should apply to the whole block as I don't see anything mapping-related, as well as saying they "_would like to avoid releasing `peer.service` to the public and then have to think about a migration strategy_".

However, I would be more confident in this if @labbati also shares the same point of view 😃

labbati: The whole block is fine.

PROFeNoM:
A [fiber test](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/11792/workflows/3555176a-916f-47bc-8298-7296a141382b/jobs/2534130) in 8.1 only is leaking memory. I thought this was a flaky one but it isn't, I'll look at it asap

ext/serializer.c
```diff
@@ -821,13 +838,45 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
...
+            ZEND_HASH_FOREACH_VAL(peer_service_sources, tag)
+            {
+                if (Z_TYPE_P(tag) == IS_STRING) {
+                    // Use the first tag that is found in the span, if any
+                    zval *peer_service = zend_hash_find(Z_ARRVAL_P(meta), Z_STR_P(tag));
+                    if (peer_service) {
+                        add_assoc_str(meta, "_dd.peer.service.source", zend_string_copy(Z_STR_P(tag)));
+
+                        if (!dd_set_mapped_peer_service(meta, Z_STR_P(peer_service))) {
+                            add_assoc_str(meta, "peer.service", zend_string_copy(Z_STR_P(peer_service)));
```
bwoebi, line 869:
```suggestion
                        zend_string *peer = zval_get_string(peer_service);
                        if (!dd_set_mapped_peer_service(meta, peer)) {
                            add_assoc_str(meta, "peer.service", peer);
```
And remove the zend_string_copy in dd_set_mapped_peer_service.

ext/serializer.c
```diff
@@ -821,13 +838,45 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
...
+                dd_set_mapped_peer_service(meta, Z_STR_P(peer_service));
+            }
+        } else if (zend_hash_num_elements(peer_service_sources) > 0) {
+            zval *tag;
+            ZEND_HASH_FOREACH_VAL(peer_service_sources, tag)
+            {
+                if (Z_TYPE_P(tag) == IS_STRING) {
+                    // Use the first tag that is found in the span, if any
+                    zval *peer_service = zend_hash_find(Z_ARRVAL_P(meta), Z_STR_P(tag));
+                    if (peer_service) {
```
bwoebi, line 865:
```suggestion
                    if (peer_service && Z_TYPE_P(peer_service) == IS_STRING) {
```

PROFeNoM:
This leak, in fact, doesn't originate from this PR (it all of a sudden happens on master as well), which would have seemed weird anyway, considering the whole block of code wasn't being run 😅

PR #2136, opened 2023-06-29
CI: don't run benchmarks on non profiling PRs

realFlowControl:
### Description

This PR will actually really (this time for real) make sure that benchmarks are only run on changes to the `profiling/` directory. The former PR #2133 missed the fact, that if a `when` keyword is around, the `only` keyword will not be evaluated anymore.
Also the docs state the using the `rules` key is the current way to go an `only` and others are not further developed anymore.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -5,11 +5,9 @@ variables:
 
 benchmarks:
   stage: benchmarks
-  when: on_success
   only:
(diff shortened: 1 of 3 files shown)


bwoebi (approved):
That's such a trap, I messed that up myself already as well...

PR #2137, opened 2023-06-30
fix: generate runtime-id on first request

morrisonlevi:
### Description

Without this, in SAPIs which fork after MINIT, the runtime id will be shared by multiple processes. This issue was introduced in #2029 in v0.88.0 when the tracer took over runtime generation. Previously the profiler was in charge of generating the runtime id.

This affects cross-product features such as Code Hotspots in the profiler.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


bwoebi (approved):
Seems good - we only actually need the run-time ID on first request start.

PR #2142, opened 2023-07-04
Add peer.service sources to Mysqli instrumentation

labbati:
### Description

Define sources for mysqli spans so `peer.service` can be set by the post-processor.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


src/Integrations/Integrations/Mysqli/MysqliIntegration.php
```diff
@@ -40,20 +43,30 @@ public function init()
         $integration = $this;
 
         \DDTrace\trace_function('mysqli_connect', function (SpanData $span, $args, $result) use ($integration) {
-            list($host) = $args;
+            list($host,,, $dbName) = $args;
```
bwoebi, line 46: You cannot use list() with optional args, this will give you an undefined index warning if the arg isn't given (under DD_TRACE_DEBUG=1). Needs `!empty($args[3])` directly.

bwoebi:
This needs additional hooking to mysqli::select_db / mysqli_select_db to ensure the target db is correct.

src/Integrations/Integrations/Mysqli/MysqliIntegration.php
labbati: Good catch! thank you

labbati:
> This needs additional hooking to mysqli::select_db / mysqli_select_db to ensure the target db is correct.
TIL about those methods, thanks

labbati:
@bwoebi added handling of select db and the other issues related to `list` and `DatabaseIntegrationHelper` not being compatible with PHP 5. This should be good for your review.

src/Integrations/Integrations/DatabaseIntegrationHelper.php
```diff
@@ -3,9 +3,15 @@
 namespace DDTrace\Integrations;
 
 use DDTrace\HookData;
+use DDTrace\Tag;
 
 class DatabaseIntegrationHelper
 {
+    public static $PEER_SERVICE_SOURCES = [
```
bwoebi, line 10: This behaves like a const, so it probably should be a const:
```suggestion
    public const PEER_SERVICE_SOURCES = [
```
(and remove $ signs from usages)

tests/Common/BaseTestCase.php
```diff
@@ -36,9 +36,22 @@ protected function ddTearDown()
         if (\class_exists('DDTrace\Log\Logger')) {
             Logger::reset();
         }
+        foreach ($this->envsToCleanUpAtTearDown() as $env) {
+            self::putEnv($env . '=');
```
bwoebi, line 40:
```suggestion
            self::putEnv($env);
```
`$env=` sets it to empty, `$env` only sets it to undefined, i.e. default

bwoebi (approved):
Overall, looks good to me, just added two comments :-)

bwoebi (approved):
Yes, we can :-)

PR #2149, opened 2023-07-06
feat(profiling): Explicit shutdown allocation profiling on `fork()`

realFlowControl:
### Description

This PR will explicit shutdown allocation profiling on `fork()`ing. This does not block currently, as we do a:
https://github.com/DataDog/dd-trace-php/blob/564415412ea4e4c333898daeff3cf5c733297679/profiling/src/allocation.rs#L73

which is not `Some()` anymore after `fork()`ing, but this PR makes this explicit and removes the custom handler from the ZendMM as well.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/allocation.rs
```diff
@@ -4,6 +4,7 @@ use crate::PROFILER;
 use crate::PROFILER_NAME;
 use crate::REQUEST_LOCALS;
 use libc::{c_char, c_int, c_void, size_t};
+use log::trace;
```
morrisonlevi, line 7: Combine this with the `use` statements on the next line.

profiling/src/pcntl.rs
```diff
@@ -94,6 +95,8 @@ fn stop_and_forget_profiling(maybe_profiler: &mut Option<Profiler>) {
     swap(&mut *maybe_profiler, &mut old_profiler);
     forget(old_profiler);
 
+    allocation_profiling_rshutdown();
```
morrisonlevi, line 98: Is this really safe to call? Maybe we need to go through it line-by-line together in Zoom.

realFlowControl: I scheduled an appointment 👍

morrisonlevi (approved):

PR #2153, opened 2023-07-10
Enable instrumentation telemetry, fix telemetry leaks, and move all integrations to deferred loading

bwoebi:
### Description

And more stable tests :-)

Any instrumentation which was not loaded in a deferred way would be always loaded, and thus never be registered as loaded integration, as the deferred loading is also the entry point for telemetry.
Also, we want to have accurate metrics which integrations are needed - hence the usage of the precise deferred loading mechanism.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (approved):
Looks good as far as I could tell by reviewing the code and taking it out for a spin. I can't vouch for the instrumentation telemetry bits, but moving all remaining integrations to be deferred loaded certainly cleans up `DD_TRACE_DEBUG` output.

I would change this log line now that all official integrations are deferred:
> Attempting integrations load; note: some integrations are only loaded on first usage

That's the only thing I could see.

ext/configuration.h
```diff
@@ -86,7 +86,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(BOOL, DD_TRACE_MEASURE_COMPILE_TIME, "true")                                                        \
     CONFIG(BOOL, DD_TRACE_DEBUG, "false")                                                                      \
     CONFIG(BOOL, DD_TRACE_ENABLED, "true", .ini_change = ddtrace_alter_dd_trace_disabled_config)               \
-    CONFIG(BOOL, DD_TRACE_TELEMETRY_ENABLED, "false", .ini_change = zai_config_system_ini_change)              \
+    CONFIG(BOOL, DD_TRACE_TELEMETRY_ENABLED, "true", .ini_change = zai_config_system_ini_change)               \
```
pierotibou, line 89: I just realize that now but all languages (but JS) use `DD_TRACE_TELEMETRY_ENABLED` . We should change

pierotibou (changes requested):

ext/configuration.h
bwoebi: Done :-)

PR #2158, opened 2023-07-12
Add log levels

bwoebi:
### Description

Adding RUSTC_BOOTSTRAP=1 to use select unstable features. In this case, #[thread_local], to avoid slow RefCell handling.

This will also allow using log.h in ZAI components, where we have a lot of // TODO log

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/components-rs/log.rs
+++ b/components-rs/log.rs
@@ -16,9 +16,7 @@ bitflags! {
         const Warn = 1 << 2;
         const Info = 1 << 3;
         const Deprecated = (1 << 4) | 1 /* Once */;
-        const SpanTracing = 1 << 5;
-        const Sampling = 1 << 6;
-        const Startup = 1 << 7;
+        const Startup = 1 << 5;
     }
 }
 
@@ -103,7 +101,7 @@ pub extern "C" fn ddog_log(level: Log, msg: CharSlice) {
                 let msgstr = unsafe { msg.to_utf8_lossy() };
                 if !logged.contains(msgstr.as_ref()) {
                     let msg = msgstr.to_string();
-                    let once_msg = format!("{}; This message is only displayed once. Append 'Once' DD_TRACE_DEBUG= to show all messages (e.g. DD_TRACE_DEBUG=Error,Once).\0", msg);
+                    let once_msg = format!("{}; This message is only displayed once. Specify DD_TRACE_ONCE_LOGS=0 to show all messages.\0", msg);
                     cb(level, unsafe { CharSlice::new(once_msg.as_ptr() as *const c_char, once_msg.len() - 1) });
                     logged.insert(msg);
                 }
--- a/components-rs/log.rs
+++ b/components-rs/log.rs
@@ -1,0 +1,118 @@
+use std::cell::RefCell;
+use std::collections::BTreeSet;
+use std::ffi::c_char;
+use std::slice;
+use bitflags::bitflags;
+use ddcommon_ffi::CharSlice;
+use ddcommon_ffi::slice::AsBytes;
+
+bitflags! {
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    #[repr(C)]
+    pub struct Log: u32 {
+        const None = 0;
+        const Once = 1 << 0;
+        const Error = 1 << 1;
+        const Warn = 1 << 2;
+        const Deprecated = (1 << 3) | 1 /* Once */;
+        const SpanTracing = 1 << 4;
+        const Sampling = 1 << 5;
+        const Startup = 1 << 6;
+    }
+}
+
+#[macro_export]
+macro_rules! log {
+    ($source:ident, $msg:expr) => { if ($crate::ddog_shall_log($crate::Log::$source)) { $crate::log($crate::Log::$source, $msg) } }
+}
+
+#[allow_internal_unstable(thread_local)]
+macro_rules! thread_local {
+    ($($tokens:tt)*) => {
+        #[thread_local]
+        $($tokens)*
+    }
+}
+
+#[no_mangle]
+#[allow(non_upper_case_globals)]
+pub static mut ddog_log_callback: Option<extern "C" fn(Log, CharSlice)> = None;
+
+// Avoid RefCell for performance
+thread_local! { static mut LOG_LEVEL: Log = Log::Once.union(Log::Error); }
+std::thread_local! {
+    static LOGGED_MSGS: RefCell<BTreeSet<String>> = RefCell::default();
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_shall_log(level: Log) -> bool {
+    unsafe { LOG_LEVEL }.contains(level & !Log::Once)
+}
+
+pub fn log<S>(level: Log, msg: S) where S: AsRef<str> {
+    ddog_log(level, CharSlice::from(msg.as_ref()))
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_set_log_level(level: Log) {
+    unsafe { LOG_LEVEL = level; }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn ddog_parse_log_level(levels: *const CharSlice, num: usize, startup_logs_by_default: bool) {
+    let mut total_level = Log::None;
+    let levels = slice::from_raw_parts(levels, num);
+
+    if levels.len() == 1 {
+        let first_level = levels[0].to_utf8_lossy();
+        if first_level == "1" || first_level == "true" || first_level == "On" {
+            total_level = Log::Error | Log::Warn | Log::Deprecated;
+            if startup_logs_by_default {
+                total_level |= Log::Startup;
+            }
+        }
+    }
+
+    for level_name in levels {
+        for (name, level) in Log::all().iter_names() {
+            if name.eq_ignore_ascii_case(&level_name.to_utf8_lossy()) {
+                total_level |= level;
+            }
+        }
+    }
+
+    // Warn always implies error
+    if total_level.contains(Log::Warn) {
+        total_level |= Log::Error;
+    }
+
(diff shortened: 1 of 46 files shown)


ext/auto_flush.c
```diff
@@ -28,7 +29,7 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
 
     if (zend_hash_num_elements(Z_ARR(trace)) == 0) {
         zend_array_destroy(Z_ARR(trace));
-        ddtrace_log_debug("No finished traces to be sent to the agent");
+        LOG(Warn, "No finished traces to be sent to the agent");
```
pierotibou, line 32: I would have said this is debug

ext/auto_flush.c
```diff
@@ -40,13 +41,13 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
...
         if (size > limit) {
-            ddtrace_log_errf("Agent request payload of %zu bytes exceeds configured %zu byte limit; dropping request", size, limit);
+            LOG(Error, "Agent request payload of %zu bytes exceeds configured %zu byte limit; dropping request", size, limit);
             success = false;
         } else {
             success = ddtrace_send_traces_via_thread(1, payload, size);
             if (success) {
                 char *url = ddtrace_agent_url();
-                ddtrace_log_debugf("Flushing trace of size %d to send-queue for %s",
+                LOG(Warn, "Flushing trace of size %d to send-queue for %s",
```
pierotibou, line 50: I would have said this is debug

ext/comms_php.c
```diff
@@ -2,15 +2,16 @@
...
+#include <components/log/log.h>
 #include "mpack/mpack.h"
 
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
 bool ddtrace_send_traces_via_thread(size_t num_traces, char *payload, size_t payload_len) {
     if (!get_DD_TRACE_ENABLED()) {
         // If the tracer is set to drop all the spans, we do not signal an error.
-        ddtrace_log_debugf("Traces are dropped by PID %ld because tracing is disabled.", getpid());
+        LOG(Warn, "Traces are dropped by PID %ld because tracing is disabled.", getpid());
```
pierotibou, line 14: If this is what the customer want, I assume we should not flood with warning.
Maybe logging once would be good as it seems a possibility

ext/configuration.h
```diff
@@ -84,7 +84,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(BOOL, DD_TRACE_AUTO_FLUSH_ENABLED, "false")                                                         \
     CONFIG(BOOL, DD_TRACE_CLI_ENABLED, "false")                                                                \
     CONFIG(BOOL, DD_TRACE_MEASURE_COMPILE_TIME, "true")                                                        \
-    CONFIG(BOOL, DD_TRACE_DEBUG, "false")                                                                      \
+    CONFIG(SET, DD_TRACE_DEBUG, "Error", .ini_change = ddtrace_alter_dd_trace_debug)                           \
```
pierotibou, line 87: `DD_TRACE_DEBUG` is a bool in all languages https://docs.datadoghq.com/tracing/troubleshooting/tracer_debug_logs/?code-lang=java#enable-debug-mode. And I don't think we have a log level one, by default we log everything starting at Info, unless `DD_TRACE_DEBUG` is true (at least in .NET)

ext/ddtrace.c
```diff
@@ -593,7 +596,7 @@ static bool dd_is_compatible_sapi(datadog_php_string_view module_name) {
 static void dd_disable_if_incompatible_sapi_detected(void) {
     datadog_php_string_view module_name = datadog_php_string_view_from_cstr(sapi_module.name);
     if (UNEXPECTED(!dd_is_compatible_sapi(module_name))) {
-        ddtrace_log_debugf("Incompatible SAPI detected '%s'; disabling ddtrace", sapi_module.name);
+        LOG(Warn, "Incompatible SAPI detected '%s'; disabling ddtrace", sapi_module.name);
```
pierotibou, line 599: Can happen only once per SAPI I assume, right?

ext/excluded_modules.c
```diff
@@ -56,9 +56,9 @@ void ddtrace_excluded_modules_startup() {
         if (module && module->name && module->version && ddtrace_is_excluded_module(module, error)) {
             ddtrace_has_excluded_module = true;
             if (strcmp("xdebug", module->name) == 0) {
-                ddtrace_log_err(error);
+                LOG(Error, error);
             } else {
-                ddtrace_log_debug(error);
+                LOG(Warn, error);
```
pierotibou, line 61: Why not debug?

components-rs/log.rs
```diff
@@ -0,0 +1,118 @@
...
+use std::slice;
+use bitflags::bitflags;
+use ddcommon_ffi::CharSlice;
+use ddcommon_ffi::slice::AsBytes;
+
+bitflags! {
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    #[repr(C)]
+    pub struct Log: u32 {
+        const None = 0;
```
pierotibou, line 13: you reinvent log levels :)

pierotibou (commented):
> Also adding Log levels Sampling and SpanTracing, which are yet unused, but will be used to show sampling decisions and verbose span tracing.

I'd stick to what other tracer do personally. So log levels, Debug to error, by default we log starting at info. We log debug logs when `DD_TRACE_DEBUG` is on.

I like the idea of the log once, haven't looked how it's been implemented

ext/configuration.h
bwoebi: Note that boolean functionality is retained in log.rs; if the first arg is 1, On or true, it will have the same behaviour as before.

ext/comms_php.c
bwoebi: This specific one does not matter much, it would only happen if one would use the manual API to send bytes directly to the agent. (which, I hope, nobody uses. It was used by us in the past. A candidate to deprecate.)

ext/excluded_modules.c
bwoebi: That else branch is dead code, I don't know why exactly it was written like that.

components-rs/log.rs
bwoebi: Yeah, though I need log levels interoperable between Rust and C - assuming we have rust code which will need logging.

bwoebi:
I agree, starting at info is the right thing, let me add info quickly.

components-rs/log.rs
pierotibou: I still think we should have `None`, `Debug`, `Info`, `Warn` and `Error` I believe. I think once can be a nice addition indeed. 
`Deprecated` to me would be a `once` `Warning` and `Startup` is orthogonal I believe so should be removed

ext/auto_flush.c
```diff
@@ -28,7 +29,7 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
 
     if (zend_hash_num_elements(Z_ARR(trace)) == 0) {
         zend_array_destroy(Z_ARR(trace));
-        ddtrace_log_debug("No finished traces to be sent to the agent");
+        LOG(Info, "No finished traces to be sent to the agent");
```
pierotibou, line 32: I have a feeling that you set `Info` where I'd set `Debug`.  
So as setting `DD_TRACE_DEBUG` to true is allowing info and warning logs, it's basically the same result. But for consistency with other tracers (at least the one I know), I would have set `Debug`. Because again, it could help in the future if we have a centralized UI (or shared saved views) to use the same levels across tracers. That said I haven't checked all tracers so can't be 100% sure. 
And this can be reworked later as when `DD_TRACE_DEBUG` is set to false we don't get those logs

ext/ip_extraction.c
```diff
@@ -337,15 +338,15 @@ static bool dd_parse_ip_address(const char *_addr, size_t addr_len, bool ip_or_e
         static const uint8_t ip4_mapped_prefix[12] = {[10 ... 11] = 0xFF};
         if (memcmp(s6addr, ip4_mapped_prefix, sizeof(ip4_mapped_prefix)) == 0) {
             // IPv4 mapped
-            ddtrace_log_debugf("Parsed as IPv4 mapped address: %s", addr);
+            LOG(Warn, "Parsed as IPv4 mapped address: %s", addr);
```
pierotibou, line 341: I'd say it's debug as well. I'd be afraid to flood customers with logs

components-rs/log.rs
morrisonlevi: I agree that `deprecated` is a `once` `warning` level. At the bitfield level I'm okay with Startup being smushed in there, but if we start using this system, I'd like to see a bit more separation in these concepts at the API level.

morrisonlevi (approved):
I'm approving, but mostly because the log levels aren't yet respected, so it's functionally the same, or close enough. I'd want more scrutiny on categories/modules and some review on every individual log level if we do this for real.

components-rs/log.rs
```diff
@@ -0,0 +1,121 @@
...
+        const Once = 1 << 0;
```
morrisonlevi, line 14: I would document this is once per request, not once per minit/mshutdown.

ext/signals.c
```diff
@@ -50,34 +50,34 @@ static void ddtrace_sigsegv_handler(int sig) {
...
-                ddtrace_log_err(backtraces[i]);
+                LOG(Error, backtraces[i]);
             }
             free(backtraces);
         }
 #endif
     }
 
-    exit(128 + sig);
+    _Exit(128 + sig);
```
morrisonlevi, line 80: I would leave a little comment about the difference here.

PR #2163, opened 2023-07-13
Opt-in to flat service names

labbati:
### Description

Set all service name to `DD_SERVICE` (or whatever the local root span service is) for internal libraries if the use opts-in via `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED=true`.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/configuration.h
```diff
@@ -112,6 +112,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(INT, DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH, "512")                                                     \
     CONFIG(MAP, DD_TRACE_PEER_SERVICE_MAPPING, "")                                                             \
     CONFIG(BOOL, DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED, "false")                                              \
+    CONFIG(BOOL, DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED, "false")                                   \
```
bwoebi, line 115: May I note that _remove_ + _enabled_ is quite redundant? `DD_TRACE_INTEGRATION_SERVICE_NAMES_ENABLED=true` would make more sense to me.

src/Integrations/Integrations/SpanTaxonomy.php
```diff
@@ -0,0 +1,35 @@
...
+            \PHP_MAJOR_VERSION > 5
+                && \dd_trace_env_config('DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED');
+
+        if ($flatServiceNames) {
+            $span->service = empty(self::$currentRootService)
+                ? \ddtrace_config_app_name($fallbackName)
+                : self::$currentRootService;
+        } else {
+            $span->service = empty($fallbackName) ? \ddtrace_config_app_name() : $fallbackName;
+        }
```
bwoebi, line 28: I think this would be the better logic:
```suggestion
        if (!$flatServiceNames && !empty($fallbackName)) {
            $span->service = $fallbackName;
        }
```
The reason being: service names are automatically inherited on span creation. If nothing happens, it will be simply the service of the parent span, i.e. we remain in context of the actually executing service.

For example, you have a non-trivial application which is forked or using fibers, where you want to assign different forks/fibers different services. With the current approach, the DD_SERVICE would appear in the leaf functions, instead of being part of their actual parent service where it logically pertains to.

With the logic simplified that much, registerCurrentRootService will be unnecessary and this a simple helper function which could live in `Integration.php`.

ext/configuration.h
labbati: I agree with you, this is a battle that I fought (and lost) myself 😄. On top of that, there is a negation (remove*) and an affirmation (enable) in the same name, which is also a bad practice in my opinion.
I have to resolve this comment without changes for consistency with the whole project, but I feel what you say.

src/Integrations/Integrations/SpanTaxonomy.php
labbati: Applying some changes as discussed in person

bwoebi (approved):

PR #2165, opened 2023-07-14
Reflect changes to datadog.version and datadog.env in existing spans

bwoebi:
### Description

Generally, users have been trying to change datadog.version or datadog.env at runtime (because it fits their setup easier) and were disappointed it to not affect the root span.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi:
Just double-check that when the profiler is also loaded that this isn't broken, since it has a different ini_change hook.

PR #2168, opened 2023-07-16
perf(profiling): only wake every 10ms if PHP is serving a request

morrisonlevi:
### Description

We accomplish this by tracking if we're between request init and request shutdown. Earlier versions of this PR spawned a new thread, but the current version is able to slot it into the existing `ddprof_time` thread. This section of the code is conscience about supporting ZTS and NTS builds, even though we do not support ZTS builds yet.

When a new request comes in, we add the VM interrupt info to a set in the `InterruptManager`, and then we make a best-effort attempt to send a message to the `ddprof_time` channel to wake it up, so it can synchronize that a request is being served.

When the request finishes, we remove the VM interrupt info from the set, but do not wake the helper thread. The thread state will synchronize on its next event, which is fine.

This PR also removes the separate fork channels, and puts new cases into the `ProfilingMessage` and `UploadMessage` types instead. This has little impact on the profiling message channel, but it does have a little bit more of an impact on the upload one. Before, if messages were available on both channels, they would be selected at random. So, we may have to send more than one HTTP request before it processes the fork. Now, there's only one channel so it's more likely that multiple requests need to be sent at once (since they get added to the queue at the same time). Note, though, that it's rare for us to have more than one HTTP request at a time. This mostly only happens if they set a different `DD_SERVICE`, `DD_ENV`, `DD_VERSION` for that request using something like Apache's `.htaccess` file.

Supersedes #2025.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/build.rs
```diff
@@ -99,7 +99,7 @@ fn build_zend_php_ffis(php_config_includes: &str, preload: bool) {
     let files = ["src/php_ffi.c", "../ext/handlers_api.c"];
     let preload = if preload { "1" } else { "0" };
 
-    #[cfg(any(feature = "stack_walking_tests"))]
+    #[cfg(feature = "stack_walking_tests")]
```
morrisonlevi, line 102: We don't need `any` if there's only one feature.

profiling/src/profiling/interrupts.rs
```diff
@@ -22,39 +26,194 @@ impl std::fmt::Display for VmInterrupt {
...
+
+                                Err(err) => {
+                                    log::warn!("{err}");
+                                    break;
+                                },
+                            }
+                        }
+                    } else {
+                        // It's not great that we duplicate this code, but it
+                        // was the only way I could get it to compile.
```
bwoebi, line 119: Isn't it possible to do `if active_interrupts {.recv(wall_timer) } else { Future::pending() }` in that arm, so that it's never entered in the case interrupts are disabled?

morrisonlevi: This is inside of a crossbeam macro, so I'm not sure. I will poke around a bit here again to see if I can come up with something better.

morrisonlevi: I paired with Colin O'Brien and he noticed a trick in the crossbeam docs I didn't. We can select on a `crossbeam_channel::never` when we're on an iteration which shouldn't use the timer.

morrisonlevi:
When spot-checking the configurations that had possible sigsegvs, the profiler was present in all of them. This is surprising, for sure, so I'll have to dig into it.

morrisonlevi:
The shutdown sequence was subtly changed due to changes in refcounts on the uploader's channel. I've fixed this in a recent commit. This seems to have fixed the issue. However, a new sigsegv was discovered. Florian is investigating. It looks similar to past issues with JIT. As soon as we rule out this PR being the root of it, we can merge (from my perspective, still need approval).

profiling/src/profiling/mod.rs
```diff
@@ -522,6 +574,15 @@ impl Profiler {
             }
         };
         self.should_join.store(sent, Ordering::SeqCst);
+
+        // Drop the sender to the uploader channel to reduce its refcount. At
+        // this state, only the ddprof_time thread will have a sender to the
+        // uploader. Once the sender over there is closed, then the uploader
+        // can quit.
+        // The sender it is replaced with has a disconnected receiver, so it
+        // can't be used to send any messages.
```
realFlowControl, line 583:
```suggestion
        // The sender is replaced with a disconnected receiver, so it can't
        // be used to send any messages.
```

realFlowControl (approved):
Looks good to me 🚀 
I still want to fully understand the dependencies between these threads and especially the shutdown procedure, but I ran this PR for hours now in some of the Docker containers and they are stable, so WHOOP WHOOP!

morrisonlevi:
The other sigsegv is a JIT issue, so moving ahead with the merge.

realFlowControl:
Btw: thanks for the additional comments, that explains it very well

PR #2169, opened 2023-07-17
feat(profiling): attach fiber's function name as a label

realFlowControl:
### Description

This PR will attach the fiber's function name as a label to the PHP profiler. This PR will also be accompanied be some backend and ui PRs 

PROF-7402

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/profiling/mod.rs
```diff
@@ -905,6 +908,42 @@ impl Profiler {
                 sample_values.extend_from_slice(&values[3..5]);
             }
 
+            let fiber = unsafe { ddog_php_prof_get_active_fiber() };
+            if !fiber.is_null() {
+                unsafe {
+                    let functionname = Some(String::from_utf8_lossy(
+                        ddog_php_prof_zend_string_view(
+                            (*(*fiber).fci_cache.function_handler)
+                                .op_array
```
bwoebi, line 917: This needs a check whether it's actually an op_array... We have a helper function `extract_function_name()` in stack_walking.rs. He Handling of fci_cache.object then probably is unnecessary too?

realFlowControl: Thanks for pointing that out, I hated this part 😉

profiling/src/profiling/mod.rs
```diff
@@ -966,6 +972,21 @@ impl Profiler {
                 sample_values.extend_from_slice(&values[3..5]);
             }
 
+            #[cfg(php_has_fibers)]
+            {
+                let fiber = unsafe { ddog_php_prof_get_active_fiber() };
+                if !fiber.is_null() {
+                    if let Some(functionname) =
+                        unsafe { extract_function_name(&(*(*fiber).fci_cache.function_handler)) }
```
morrisonlevi, line 980: @bwoebi Can you double-check the expectation here? Is the `fci_cache` always going to hold a pointer to a valid `zend_function`?

bwoebi: Yes. The fcc is set by Fiber::__construct as part of zpp, which will always set the function_handler on success, and there's nothing changing that value in all of fibers afterwards, from start to destruction of the fiber itself.

realFlowControl: Thanks for checking 🙇

morrisonlevi (approved):
Looks good to me. Thanks for this, and thanks @bwoebi for double-checking the safety concern I asked about. @realFlowControl , check my refactoring + docs commit before merging.

PR #2173, opened 2023-07-19
run prof-correctness on PR

realFlowControl:
### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


.github/workflows/prof_correctness.yml
```diff
@@ -0,0 +1,70 @@
...
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Run tests
+        run: |
+          php -d extension=target/release/libdatadog_php_profiling.so -v
+          mkdir profiling/tests/correctness/allocations/
```
r1viollet, line 55: As we still have the nightly that tests the delivered binaries, this is fine.
As discussed we could think of ways to factorize some of the test setups.

profiling/tests/correctness/time.json
```diff
@@ -0,0 +1,47 @@
+{
+    "test_name":"clang_pie",
```
r1viollet, line 2: minor: test name was not adjusted (though we don't really use this info)

profiling/tests/correctness/time.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
r1viollet, line 1: As discussed, structuring the scenarios so they can be fetched would be nice.
I would suggest:
- correctness/{scenario}/expected.json
- correctness/{scenario}/main.php (or whatever can be used as a standard entry point ?)

realFlowControl: I would like to do this when we are working on implementing the counter part in https://github.com/DataDog/prof-correctness

r1viollet: Sorry I was not clear maybe, I just was suggesting adding an extra directory after correctness. 
Though I agree we can wait for the counter part in prof-correctness.
Generally I still think prof-correctness will eventually contain a number of different images on which you should run the test. Then you would give it a script to run, and an expected. That would run on the prof-correctness repository.

I'm fine with adjusting when needed :-)

r1viollet (approved):
LGTM

.github/workflows/prof_correctness.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
+
+      - name: Restore cached Primes
```
morrisonlevi, line 22: Cached primes? Is this a term I don't know or is this a leftover from copy/paste?

.github/workflows/prof_correctness.yml
```diff
@@ -0,0 +1,70 @@
...
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Run tests
+        run: |
+          php -d extension=target/release/libdatadog_php_profiling.so -v
+          mkdir profiling/tests/correctness/allocations/
+          DD_PROFILING_OUTPUT_PPROF=profiling/tests/correctness/allocations/test.pprof php -d extension=target/release/libdatadog_php_profiling.so profiling/tests/correctness/allocations.php
```
morrisonlevi, line 56: I've had issues with relative paths for `extension` in the past. I recommend throwing in `$PWD/` in there e.g. `-d extension=$PWD/target/...`.

There's another one of these two lines below as well.

morrisonlevi (approved):
Looks good to me other than my comments.

Issue #2172, opened 2023-07-19
[Bug]: _generated_tracer.php crashes with php-amqplib integration enabled


CatoTH:
### Bug report

We tried upgrading dd-trace-php and the PHP extension from an older version (0.79.0) to one of the more recent versions (we tried 0.88.1, 0.89.0 and 0.90.0). After this update, our php-amqplib rabbitmq-workers are crashing once a message actually comes in. We are getting the following stacktrace:

```
[2023-07-10T10:23:59+02:00] [ddtrace] [warning] - The span with name 'console' was closed without being finished first.
Argument 1 passed to DDTrace\Span::__construct() must be an instance of DDTrace\SpanData, null given, called in /opt/datadog/dd-library/0.89.0/dd-trace-sources/bridge/_generated_tracer.php on line 873
#0 /opt/datadog/dd-library/0.89.0/dd-trace-sources/bridge/_generated_tracer.php(873): DDTrace\Span->__construct(NULL, Object(DDTrace\SpanContext))
#1 /opt/datadog/dd-library/0.89.0/dd-trace-sources/bridge/_generated_tracer.php(813): DDTrace\ScopeManager->reconcileInternalAndUserland()
#2 /opt/datadog/dd-library/0.89.0/dd-trace-sources/bridge/_generated_tracer.php(1061): DDTrace\ScopeManager->getActive()
#3 /app/src/Application/Monolog/DatadogProcessor.php(32): DDTrace\Tracer->getActiveSpan()
#4 [internal function]: Application\Monolog\DatadogProcessor->__invoke(Array)
#5 /app/vendor/monolog/monolog/src/Monolog/Logger.php(340): call_user_func(Object(Application\Monolog\DatadogProcessor), Array)
#6 /app/vendor/monolog/monolog/src/Monolog/Logger.php(642): Monolog\Logger->addRecord(200, 'Executed comman...', Array)
#7 /app/src/#####CommandRunner.php(61): Monolog\Logger->info('Executed comman...', Array)
#8 /app/src/#####Consumer.php(29): #####CommandRunner->run('#####...', Array, '{"id":"90000000...', 60)
#9 /app/src/#####Consumer.php(50): #####Consumer->handle(Object(PhpAmqpLib\Message\AMQPMessage))
#10 [internal function]: #####Consumer->execute(Object(PhpAmqpLib\Message\AMQPMessage))
[...]

PR #2177, opened 2023-07-25
(Legacy OpenTracing API) Check if the active span has a parent

PROFeNoM:
### Description

While experimenting with ways to reproduce #2172, I stumbled on an issue that would lead to the tracer crashing with the error message `Attempt to read property "id" on null`. The latter would happen when using `Tracer::getActiveSpan` after `start_root_span()` was called.

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors, the reviewer is in charge of this task.

### DIFF
--- a/src/DDTrace/ScopeManager.php
+++ b/src/DDTrace/ScopeManager.php
@@ -129,8 +129,8 @@ final class ScopeManager implements ScopeManagerInterface
 
         $currentSpanId = $topScope->getSpan()->getSpanId();
         $newScopes = [];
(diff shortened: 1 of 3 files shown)


bwoebi (approved):
That's not surprising - that code was never taking stacked root spans into account.

The fix looks good enough: I wouldn't go as far as ensuring that Tracer::getRootScope() really looks at the current traces root. Let's just merge this :-)

PR #2180, opened 2023-07-29
chore: readonly not suppose to have default value

Ferror:
I wanted to use the stubs, and it seems that not all are defined correctly.

Btw. why not to move them to https://github.com/JetBrains/phpstorm-stubs?

### DIFF
--- a/ext/ddtrace.stub.php
+++ b/ext/ddtrace.stub.php
@@ -98,7 +98,7 @@ namespace DDTrace {
         /**
          * @var string The unique identifier of the span
          */
-        public readonly string $id = "";
+        public readonly string $id;
 
         /**
          * @var SpanLink[] $spanLinks An array of span links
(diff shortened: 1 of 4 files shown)


ext/ddtrace.stub.php
```diff
@@ -154,12 +154,12 @@ public function getLink(): SpanLink {}
         /**
          * @var SpanStack|null The parent stack, or 'null' if there is none
          */
-        public readonly SpanStack|null $parent = null;
```
bwoebi, line 157: Is there a reason you removed that readonly here?

bwoebi:
@Ferror We need these stubs in our own repository as the arginfo.h are generated from these.

But good finds that they cannot have a default value.

ext/ddtrace.stub.php
Ferror: Yes. You may notice that like [153](https://github.com/DataDog/dd-trace-php/pull/2180/files#diff-61e6285b55a2a83eebb2ae49717564d1a5f0307aee1d30d7df2bbf598826c5d1L153) defines whole class as a readonly, so it's invalid to duplicate it to property too.

bwoebi:
I'll merge it next week, have to also regenerate stubs and test whether it then still works propertly.

bwoebi (approved):
SpanStack itself is not supposed to be readonly, that was a mistake, removed it there instead.

PR #2185, opened 2023-08-01
Only enable allocation profiling if profiling is enabled at all

realFlowControl:
### Description

Due to a missing check on `locals.profiling_enabled` the allocation profiler was enabled as soon as the extension was loaded. We did a check on the `locals.profiling_allocation_enabled` when creating the sample message and omited everything that was disable but `wall-time`, sending empty (and invalid) `wall-time` samples.

This PR adds the missing check and cleans up with the indention.

PROF-8032

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/profiling/src/allocation.rs
+++ b/profiling/src/allocation.rs
@@ -89,64 +89,69 @@ thread_local! {
     static ALLOCATION_PROFILING_STATS: RefCell<AllocationProfilingStats> = RefCell::new(AllocationProfilingStats::new());
 }
 
-/* If Failure is returned the VM will do a C exit; try hard to avoid that,
- * using it for catastrophic errors only.
- */
 pub fn allocation_profiling_rinit() {
-    let profiling_allocation_enabled = unsafe { config::profiling_allocation_enabled() };
+    let (profiling_enabled, profiling_allocation_enabled) = unsafe {
+        (
+            config::profiling_enabled(),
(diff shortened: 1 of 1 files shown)


profiling/src/allocation.rs
```diff
@@ -89,122 +89,126 @@ thread_local! {
...
-                // already. This is unexpected, therefore we will not touch the ZendMM handlers
-                // anymore as resetting to prev handlers might result in segfaults and other
-                // undefined behaviour.
-                if !is_zend_mm() {
-                    let mut custom_mm_malloc: Option<zend::VmMmCustomAllocFn> = None;
-                    let mut custom_mm_free: Option<zend::VmMmCustomFreeFn> = None;
-                    let mut custom_mm_realloc: Option<zend::VmMmCustomReallocFn> = None;
+        if !locals.profiling_enabled || !locals.profiling_allocation_enabled {
+            return;
+        }
```
morrisonlevi, line 164: I think this is okay but I suspect that we _may_ allow toggling this at runtime in the future, in which case we'd still want to undo our heap custom handlers. Since the code already detects whether the installed handlers are ours, maybe we're good to just run it anyway?

morrisonlevi (approved):
Approved, though I left a comment for you to think about.

PR #2189, opened 2023-08-02
fix(profiling): lower timeline overhead when disabled

realFlowControl:
### Description

This PR will fix the same issue as #2185 (emitting samples when `datadog.profiling.enabled = Off` and `datadog.profiling.experimental_timeline_enabled = On`) and additionally we also register and unregister our hooks in `RINIT`/ `RSHUTDOWN` which makes sure we are not hooking into anything as long as the timeline feature is disabled, therefore lowering the overhead when timeline is disabled.

We hit a sigsegv during the development of this branch due to multiple concurrent borrows. This refactors our thread-local-storage with a few goals:

- Make mutable borrows as short-lived as possible.
- Make mutable borrows as narrowly scoped as possible, meaning they don't borrow more fields than they need.

So there are now globals for `TAGS` and `CLOCKS`. This caused a small performance regression, though we were not able to identify why. Instead, we brought performance back in line by borrowing the string table once for the entire stack collection, rather that re-borrowing it on every frame.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/lib.rs
```diff
@@ -908,7 +908,14 @@ fn notify_trace_finished(local_root_span_id: u64, span_type: Cow<str>, resource:
...
     REQUEST_LOCALS.with(|cell| {
-        let mut locals = cell.borrow_mut();
+        // try to borrow and bail out if not successful
+        let mut locals = match cell.try_borrow_mut() {
+            Ok(locals) => locals,
+            Err(_) => {
+                return;
+            }
+        };
+
```
realFlowControl, line 918: @morrisonlevi can you check this change?
The reason to this change was that the `REQUEST_LOCALS` could have already been borrowed in `datadog_php_profiling::timeline::ddog_php_prof_gc_collect_cycles` which wrapped the call to the original `gc_collect_cylces()` function. I changed that part to not have the `REQUEST_LOCALS` borrowed anymore when making the call to `gc_collect_cycles()`, so we might not need this anymore.

profiling/tests/phpt/phpinfo_01.phpt
```diff
@@ -46,9 +46,9 @@ assert(isset($values["Version"]));
 // Check exact values for this set
 $sections = [
     ["Profiling Enabled", "false"],
-    ["Experimental CPU Time Profiling Enabled", "true"],
-    ["Allocation Profiling Enabled", "true"],
-    ["Endpoint Collection Enabled", "true"],
+    ["Experimental CPU Time Profiling Enabled", "false"],
+    ["Allocation Profiling Enabled", "false"],
+    ["Endpoint Collection Enabled", "false"],
```
morrisonlevi, line 51: Think we should try to enhance these messages? "disabled because profiling as a whole is disabled" or something?

realFlowControl: I went with a short version: `false (profiling disabled)`

profiling/tests/phpt/phpinfo_02.phpt
morrisonlevi, line 1: Oops, looks like you copy/pasted and overwrite an existing test ^_^

morrisonlevi (approved):

profiling/src/lib.rs
```diff
@@ -386,10 +400,10 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
...
         (
             config::profiling_enabled(),
-            config::profiling_endpoint_collection_enabled(),
-            config::profiling_experimental_cpu_time_enabled(),
-            config::profiling_allocation_enabled(),
-            config::profiling_experimental_timeline_enabled(),
+            config::profiling_endpoint_collection_enabled() && config::profiling_enabled(),
+            config::profiling_experimental_cpu_time_enabled() && config::profiling_enabled(),
+            config::profiling_allocation_enabled() && config::profiling_enabled(),
+            config::profiling_experimental_timeline_enabled() && config::profiling_enabled(),
```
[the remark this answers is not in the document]
realFlowControl: Could you have a look at https://github.com/DataDog/dd-trace-php/pull/2189/commits/516a4bd8895e7c26939cd62c7f8bbb5eeaf277d9
This will only call the `config::profiling_...()` functions if `profiling_enabled` is true.

profiling/src/lib.rs
```diff
@@ -398,12 +398,13 @@ extern "C" fn rinit(r#type: c_int, module_number: c_int) -> ZendResult {
...
-            config::profiling_enabled(),
-            config::profiling_endpoint_collection_enabled() && config::profiling_enabled(),
-            config::profiling_experimental_cpu_time_enabled() && config::profiling_enabled(),
-            config::profiling_allocation_enabled() && config::profiling_enabled(),
-            config::profiling_experimental_timeline_enabled() && config::profiling_enabled(),
+            profiling_enabled,
+            profiling_enabled && config::profiling_endpoint_collection_enabled(),
+            profiling_enabled && config::profiling_experimental_cpu_time_enabled(),
+            profiling_enabled && config::profiling_allocation_enabled(),
+            profiling_enabled && config::profiling_experimental_timeline_enabled(),
```
morrisonlevi, line 407: Visually, I love how much cleaner it looks too!

profiling/src/lib.rs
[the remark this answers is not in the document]
morrisonlevi: I think what the commit has is good enough. I would _hope_ the optimizer can handle it anyway but even if it can't, it's better than before the commit.

PR #2176, opened 2023-07-21
Handle killed workers and change root span initialization in the integrations

PROFeNoM:
### Description

Fixes #2171 by:
- Force-flushing the traces before the process is [killed](https://github.com/illuminate/queue/blob/99e275f601f17b187324f10d8349476bf046b733/Worker.php#L771) by Laravel queue (`MaxAttemptsExceededException` handling, see screenshot below)
- Allowing the Laravel integration's hooked to be installed even without a root span (`report` helper function usage handling, see screenshot below); the latter is also applied to other integrations

**Screenshots**

- `report(new \Exception("Reported Exception in SendVerificationEmail"));` 
image

- Job timeout
image

### Readiness checklist
- [X] (only for Members) Changelog has been added to the release document.
- [ ] Tests added for this feature/bug.

### Reviewer checklist
- [X] Appropriate labels assigned.
- [X] Milestone is set.
- [X] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #2194, opened 2023-08-04
fix: add Symfony command's exception to the root span

### LINKED ISSUE #2193: [Bug]: Symfony Console Exceptions Not Being Reported As Errors
### Bug report

Whenever a Symfony console command throws an uncaught exception the `symfony.console.commmand.run` span is correctly being marked with an error, but that doesn't bubble up to the parent `console` span. This then means that the trace isn't being linked to an error in APM.

This was observed in 0.90.0 and rolling back to 0.88.1 fixes the issue, I didn't get a chance to test 0.89.0

**Symfony version**: 5.4

### PHP version

8.1

0.90.0

### Installed extensions

_No response_

### Output of `phpinfo()`

_No response_

### Upgrading from

_No response_

PROFeNoM:
### Description

Fixes #2193

Retrieves the uncaught exception thrown in `symfony.console.command.run` and sets it on the root span by explicitly adding the hooks on `Symfony\Component\Console\Application::renderException` and `Symfony\Component\Console\Application::renderThrowable` in the Symfony integration; hooks that were previously set from the Laravel integration, but weren't set anymore since #2153 because all integrations moved to deferred loading.

This PR also fixes the CI bug that appeared a few days ago. 
- [UnitOfWork::executeInserts](https://github.com/doctrine/orm/blob/a616914887ea160db4158d2c67752e99624f7c8a/lib/Doctrine/ORM/UnitOfWork.php#L1164) doesn't seem to have any args. 
- This condition then evaluates to true because there aren't any args, which means `\datadog\appsec\track_user_signup_event` isn't called, eventually leading to the tests failing. 
- It wasn't failing before because it seems like this was a change introduced in a somewhat recent [commit](https://github.com/doctrine/orm/commit/7fc359c2bb7d92ac0c4088986307fd793141e458), but it is impact-less.

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.
- [X] Changelog has been added to the release document. For community contributors, the reviewer is in charge of this task.


PROFeNoM:
Note that this doesn't change the fact that **I don't know** why would reverting to 0.88.1 fix this issue

morrisonlevi:
>  Note that this doesn't change the fact that I don't know why would reverting to 0.88.1 fix this issue

I think we should understand that before we merge. There may be something more to fix/change.

PROFeNoM:
(Still investigating)

image


Found that this issue appeared with #2153. The error message/stack is there, but the `error` tag is set to zero, hence the issue.

PROFeNoM:
The real issue came from the fact that the laravel integration moved to deferred loading AND that the laravel integration has the hooks on `Symfony\Component\Console\Application::renderException` and `Symfony\Component\Console\Application::renderThrowable`, setting the error on the root span. Since the laravel integration is no longer loaded every time, these methods weren't instrumented, and the error wasn't set 🙃

PROFeNoM:
Alternatively, we can/should merge #2176 before considering this PR. These hooks weren't installed because the integration wasn't fully loaded since there was no root span at initialization. It should be fixed with #2176.

bwoebi (approved):

PR #2197, opened 2023-08-07
feat(profiling): add exception profiling

realFlowControl:
### Description

This will add exception profiling to the profiler.

The sampling distance (`EXCEPTION_PROFILING_INTERVAL`) is to be evaluated, the current value is set to 100.
- Laravel Demo app throws 1 exception per request
- Symfony Demo app throws 0 exceptions per request (in `prod`mode, ~70 in `dev` mode)
- more to be checked

The sampling distance is configurable via `DD_PROFILING_EXPERIMENTAL_EXCEPTION_SAMPLING_DISTANCE` or `datadog.profiling.experimental_exception_sampling_distance` INI.

The accompanying PR's are behind a feature flag:
- https://github.com/DataDog/profiling-backend/pull/4375
- https://github.com/DataDog/profiling-backend/pull/4398
- https://github.com/DataDog/logs-backend/pull/55512

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/exception.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    EXCEPTION_PROFILING_INTERVAL.store(sampling_distance, Ordering::Relaxed);
+
+    trace!("Exception profiling enabled with sampling disance of {sampling_distance}");
+
+    unsafe {
+        PREV_ZEND_THROW_EXCEPTION_HOOK = zend::zend_throw_exception_hook;
+        zend::zend_throw_exception_hook = Some(exception_profiling_throw_exception_hook);
+    }
+}
```
realFlowControl, line 95: move to `minit`

profiling/src/exception.rs
```diff
@@ -0,0 +1,114 @@
...
+unsafe extern "C" fn exception_profiling_throw_exception_hook(
+    #[cfg(php7)] exception: *mut zend::zval,
+    #[cfg(php8)] exception: *mut zend::zend_object,
+) {
+    #[cfg(php7)]
+    let exception_name =
+        ddog_php_prof_zend_string_view((*(*(*exception).value.obj).ce).name.as_mut()).to_string();
+    #[cfg(php8)]
+    let exception_name =
+        ddog_php_prof_zend_string_view((*(*exception).ce).name.as_mut()).to_string();
```
realFlowControl, line 108: unify fetching the class name from `ce`

profiling/src/exception.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    EXCEPTION_PROFILING_STATS.with(|cell| {
+        let mut exceptions = cell.borrow_mut();
+        exceptions.track_exception(exception_name)
+    });
```
realFlowControl, line 113: call prev handler ;-)

profiling/src/profiling/mod.rs
```diff
@@ -290,6 +300,20 @@ impl TimeCollector {
             }
         }
 
+        #[cfg(feature = "exception_profiling")]
+        if let Some(exception_samples_offset) = exception_samples_offset {
+            let upscaling_info = UpscalingInfo::Proportional {
+                scale: EXCEPTION_PROFILING_INTERVAL.load(Ordering::Relaxed) as f64,
+            };
+            let values_offset: Vec<usize> = vec![exception_samples_offset];
+            match profile.add_upscaling_rule(values_offset.as_slice(), "", "", upscaling_info) {
```
morrisonlevi, line 309: No need for a vec here, just use an array, something like:

```suggestion
            let values_offset = [exception_samples_offset];
            match profile.add_upscaling_rule(&values_offset, "", "", upscaling_info) {
```

Probably can do the same to allocation profiling, where this was likely copy/pasted from.

profiling/src/exception.rs
```diff
@@ -0,0 +1,130 @@
...
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::Ordering;
+
+use rand_distr::{Distribution, Poisson};
+
+/// The engine's previous throw exception hook
+static mut PREV_ZEND_THROW_EXCEPTION_HOOK: Option<zend::VmZendThrowExceptionHook> = None;
+
+/// take a sample every 100 exceptions
+pub static EXCEPTION_PROFILING_INTERVAL: AtomicI64 = AtomicI64::new(100);
```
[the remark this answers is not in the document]
realFlowControl: Yes, this is read from from the `ddprof_time` thread in `handle_sample_message`.

profiling/src/config.rs
```diff
@@ -220,6 +228,20 @@ pub(crate) unsafe fn profiling_experimental_timeline_enabled() -> bool {
...
+/// This function must only be called after config has been initialized in
+/// rinit, and before it is uninitialized in mshutdown.
+pub(crate) unsafe fn profiling_experimental_exception_enabled() -> bool {
+    get_bool(ProfilingExperimentalExceptionEnabled, false)
+}
+
+/// # Safety
+/// This function must only be called after config has been initialized in
+/// rinit, and before it is uninitialized in mshutdown.
+pub(crate) unsafe fn profiling_experimental_exception_sampling_distance() -> i64 {
```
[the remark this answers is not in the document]
realFlowControl: Nope, thinking more about this, it should be unsigned and `> 0`, so I'll add some checks. I do not think that there is a situation where you'll set the sampling distance to something that might not fit into a `u16`, but this might also be too limiting OTOH.
I'll add some checks and we can go with `u32` WDYT?

profiling/src/exception.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+thread_local! {
+    static EXCEPTION_PROFILING_STATS: RefCell<ExceptionProfilingStats> = RefCell::new(ExceptionProfilingStats::new());
+}
+
+pub fn exception_profiling_minit() {
+    unsafe {
+        PREV_ZEND_THROW_EXCEPTION_HOOK = zend::zend_throw_exception_hook;
+        zend::zend_throw_exception_hook = Some(exception_profiling_throw_exception_hook);
```
bwoebi, line 71: @realFlowControl I'm not really persuaded that the way this hook is used, is the right way.
After all, this hook is used for every single zend_throw_exception_internal.

What you are effectively sampling for is hence the number of frames plus the number of try+catch/finally constructs the exception falls through.

This will skew the exception profiling heavily towards exceptions which fall through more functions. I.e. an exception which is thrown and caught just in the function below is sampled much more seldom than an exception which comes 10 frames deep from within a library and then is caught by the caller.

To guard against that, I'd propose to:
a) take more often an exception sample
b) store the pointer to the last exception as well as the pointer to the last exception stacktrace [reduce pointer collision probability].

bwoebi (commented):
I'm wondering whether we are interested - at all - in exception profiling for the internal UnwindExit and GracefulExit exceptions. It's a detail of the implementation, which isn't even documented.

profiling/src/exception.rs
realFlowControl: > What you are effectively sampling for is hence the number of frames plus the number of try+catch/finally constructs the exception falls through.

As in:
```php
function first() {
    try {
        second();
    } catch (\TypeError $e) {
        // let the \RuntimeException bubble
    }
}

function second() {
    throw new \RuntimeException("foobar");
}

try {
    first();
} catch (\Exception $e) {
    // catched
}
```

would call the hook twice? As I do not observe this. This example PHP code will call my hook exactly once.

bwoebi: It's 7.1 behaviour actually, I thought it's been that way for longer.

realFlowControl: I could reproduce this on PHP 7.1 and it is identifiable by doing a `NULL` check on the `exception` given to that hook. So I added this check and commit explaining why this exist.

.github/workflows/prof_correctness.yml
```diff
@@ -51,6 +51,7 @@ jobs:
 
       - name: Run tests
         run: |
+          export DD_PROFILING_LOG_LEVEL=trace
```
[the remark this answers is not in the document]
realFlowControl: Well yes and no. The exception prof correctness failed and I could not see what was going on. Having that would have helped a bit. So if you don't mind I would leave this set to `trace`

profiling/src/exception.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    });
+
+    if !exception_profiling {
+        return;
+    }
+
+    EXCEPTION_PROFILING_INTERVAL.store(sampling_distance, Ordering::Relaxed);
+
+    trace!("Exception profiling enabled with sampling disance of {sampling_distance}");
```
morrisonlevi, line 92: Once this is moved to first rinit, this can be `info`.

profiling/src/lib.rs
```diff
@@ -391,6 +399,8 @@ thread_local! {
         profiling_experimental_cpu_time_enabled: true,
         profiling_allocation_enabled: true,
         profiling_experimental_timeline_enabled: true,
+        profiling_experimental_exception_enabled: true,
```
morrisonlevi, line 402: I think this should be `false` to match the setting in the config.rs.

PR #2203, opened 2023-08-10
Refactor zai_string_view APIs and use them more

morrisonlevi:
### Description

This adds new macros and functions for working with zai_string_view. They are prefixed with `zai_str_` or `ZAI_STR_` in preparation for renaming `zai_string_view` to simply `zai_str`. By naming these with the new prefix, it will reduce the size of the review for that PR.

This changes the signatures of many functions taking pairs of pointers and lengths to using zai_string_views instead, and the functions were renamed because they had `literal` in their names which didn't apply anymore. A small number of these already had versions that took views, so the versions with literals were outright removed.

### Readiness checklist
- [x]  Changelog has been added to the release document.
- [x] ~Tests added for this feature/bug.~ Existing tests are sufficient.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zend_abstract_interface/zai_string/string.h
+++ b/zend_abstract_interface/zai_string/string.h
@@ -69,7 +69,10 @@ static inline zai_string_view zai_string_from_zstr(zend_string *zstr) {
     return zstr ? ZAI_STRING_FROM_ZSTR(zstr) : ZAI_STRING_EMPTY;
 }
 
-static inline bool zai_string_stuffed(zai_string_view s) { return s.ptr && s.len; }
+/** Returns whether the string is empty. */
+static inline bool zai_str_is_empty(zai_string_view self) {
+    return self.len == 0 || self.ptr == NULL;
+}
 
 static inline
 bool zai_string_view_eq(zai_string_view a, zai_string_view b) {
--- a/zend_abstract_interface/zai_string/string.h
+++ b/zend_abstract_interface/zai_string/string.h
@@ -22,41 +22,41 @@ typedef struct zai_string_view_s {
 } zai_string_view;
 
 /** Use if data is known to be non-null, use zai_string_view_new otherwise. */
-#define ZAI_STRING_VIEW_NEW(data, size)   \
+#define ZAI_STR_NEW(data, size)   \
     (zai_string_view) {.len = (size), .ptr = (data)}
 
 #define ZAI_STRL_VIEW(literal) \
-    ZAI_STRING_VIEW_NEW("" literal, sizeof(literal) - 1) \
+    ZAI_STR_NEW("" literal, sizeof(literal) - 1) \
 
 #define ZAI_STRING_EMPTY \
-    ZAI_STRING_VIEW_NEW("", 0)
+    ZAI_STR_NEW("", 0)
 
-/** Use if cstr is known to be non-null, use zai_string_from_cstr otherwise. */
-#define ZAI_STRING_FROM_CSTR(cstr)  \
-    ZAI_STRING_VIEW_NEW((cstr), strlen(cstr))
+/** Use if cstr is known to be non-null, use zai_str_from_cstr otherwise. */
+#define ZAI_STR_FROM_CSTR(cstr)  \
+    ZAI_STR_NEW((cstr), strlen(cstr))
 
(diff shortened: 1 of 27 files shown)


zend_abstract_interface/zai_string/string.h
```diff
@@ -5,28 +5,80 @@
...
+ * Creates a zai_string_view from a possibly-null zend_string. Returns
+ * ZAI_STRING_EMPTY if the pointer is null.
+ *
+ * If the pointer is known to be non-null, use ZAI_STRING_FROM_ZSTR directly.
+ */
+static inline zai_string_view zai_string_from_zstr(zend_string *zstr) {
+    return zstr ? ZAI_STRING_FROM_ZSTR(zstr) : ZAI_STRING_EMPTY;
+}
 
 static inline bool zai_string_stuffed(zai_string_view s) { return s.ptr && s.len; }
```
bwoebi, line 72: May I vote for having (the inverse) `zai_string_empty` instead? I'm not using the function because of its name - I literally want to know "is not empty" in my mind, thus `!zai_string_empty()` is the most natural way to think about it for me. Just doesn't occur to me to use "stuffed".

morrisonlevi: I can add that, no problem!

morrisonlevi: I added `zai_string_empty` and removed `zai_string_stuffed` 👍🏻

zend_abstract_interface/zai_string/string.h
```diff
@@ -5,28 +5,82 @@
...
+#include <Zend/zend.h>
+
+/**
+ * Represents a non-owning view of a string.
+ *
+ * When initializing the struct, use one of the initialization macros or
+ * functions. Do not initialize the struct directly e.g. `{0, null}`.
+ *
+ * todo: move .ptr to come before .len to match ddtrace_string and Rust and
+ *       ensure it's never null.
```
bwoebi, line 17: Technically, an `Option<zai_string_view>` could be represented by ptr being null.

zend_abstract_interface/zai_string/string.h
```diff
@@ -5,28 +5,82 @@
 #include <stddef.h>
 #include <string.h>
 
+#include <Zend/zend.h>
+
+/**
+ * Represents a non-owning view of a string.
```
bwoebi, line 11: `zai_symbol_lookup_key` at least returns an owned string. ... And it has to case the `const` away on ptr to free it :-)
There likely should be `zai_string_owned` with a non-const ptr. And a possibility to alias the owned struct to the view.

bwoebi: E.g.
```
typedef union zai_string_owned_s {
    struct {
        size_t len;
        char *ptr;
    };
    zai_string_view view;
} zai_string_owned;
```

zend_abstract_interface/zai_string/string.h
```diff
@@ -5,28 +5,82 @@
...
 
-static inline bool zai_string_equals_literal(zai_string_view s, const char *str) {
-    return s.len == strlen(str) && (strlen(str) == 0 || strncmp(s.ptr, str, strlen(str)) == 0);
+static inline bool zai_str_eq(zai_string_view a, zai_string_view b) {
+    return a.len == b.len && (b.len == 0 || memcmp(a.ptr, b.ptr, b.len) == 0);
 }
 
-static inline bool zai_string_equals_literal_ci(zai_string_view s, const char *str) {
-    return s.len == strlen(str) && (strlen(str) == 0 || strncasecmp(s.ptr, str, strlen(str)) == 0);
+static inline bool zai_str_equals_ci_cstr(zai_string_view s, const char *str) {
```
bwoebi, line 81: Should be probably called zai_str_eq_ci_cstr for symmetry with the other function name.

bwoebi (approved):
Please check my comments, but other than that, looks good :-)

zend_abstract_interface/zai_string/string.h
morrisonlevi: I'll rename it after the renaming PR to avoid a conflict 👍🏻

PR #2204, opened 2023-08-10
Add http.route to laravel

estringana:
### Description

Implements `http.route` extraction for Laravel. This makes it impossible to get meaningful per-endpoint metrics or to build an accurate API inventory.

The implementation hooks into Laravel's router after route matching completes and reads the matched route's URI pattern (e.g. `/users/{id}`) from the route object. This is attached to the root span as `http.route`. The hook fires after the router has done its work but before the controller runs, so there's no additional route-resolution overhead.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


bwoebi:
We do have http.route for Laminas, but there it's a name given by the user to the framework, whereas this looks like a standardized (?) format (or why do you replace `{foo}` by `:foo`?) for formatting routes.

Are there any resources behind that?

estringana:
> We do have http.route for Laminas, but there it's a name given by the user to the framework, whereas this looks like a standardized (?) format (or why do you replace `{foo}` by `:foo`?) for formatting routes.
> 
> Are there any resources behind that?
> 
> Also the semantics of that tag should then probably streamlined for laminas.

tests/snapshots/tests.integrations.laminas.v1_4.common_scenarios_test.test_scenario_get_return_string.json
```diff
@@ -11,7 +11,7 @@
       "_dd.p.dm": "-1",
       "component": "laminas",
       "http.method": "GET",
-      "http.route": "simple",
+      "laminas.route.name": "simple",
```
PROFeNoM, line 14: Can we really plainly remove this `http.route` tag? Or is it gonna be set back in a future PR? (Applies to Laminas changes)

PROFeNoM:
Also, could you please merge master? 😃 Some CI issues were fixed in the recent commits :+1:

tests/snapshots/tests.integrations.laminas.v1_4.common_scenarios_test.test_scenario_get_return_string.json
estringana: Not sure about this. @bwoebi what's your view?

bwoebi: A future PR will give it a semantically different value, so it's fine to rename this.

estringana:
> Also, could you please merge master? 😃 Some CI issues were fixed in the recent commits 👍

Done

bwoebi (approved):
Looks good to me now, thanks :-)

PR #2206, opened 2023-08-11
Rename zai_string_view to zai_str

morrisonlevi:
### Description

It also renames macros and functions to use `zai_str` or `ZAI_STR` as a prefix.

There are a few motivations for this:

- This is C, and shorter names are generally better than longer names.
- We use Rust far more than C++ on this team, so matching Rust's naming makes more sense than matching C++'s naming.
- Eventually, I hope to make `zai_str` never hold null pointers and to ensure it has the same layout as Rust `&str`s do, which would improve performance (and not just for Rust, some functions can drop their null pointer checks which helps C as well). Admittedly, it's likely to be a small small improvement.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ This is a pure refactor, the existing tests are adequate.

### Reviewer checklist
- [x] Milestone is set.


bwoebi:
The plan to avoid NULLs for ptr sounds good (can be just the static "") then.

bwoebi (approved):
I hope this is literal find & replace, but no objections to the change itself :-)

morrisonlevi:
Yep, literal find-and-replace on these (just carefully, one rename at a time).

realFlowControl (approved):

PR #2214, opened 2023-08-16
Require `zai_str.ptr` to not be null

morrisonlevi:
### Description

Requires `zai_str.ptr` to not be null and also moves the `.ptr` to be first. This matches Rust's slice layout, as well as `struct iovec` on most platforms. Reordering also helped find some C++ locations which were using designated initializers, as that feature in C++ requires fields to be initialized in the same order as the source, and were thereby broken by the re-ordering.

This also fixes the profiler's equivalent structs (there were two of them 🤦🏻). It combines them into one struct since they had to be updated anyway.

This also fixes a potential heap-buffer-overflow when zai is used without an observer. I think only the tracer was using the particular set of functions which could have hit this, and the tracer is an observer, so in practice I don't think it could have been triggered by users. I hit this when running asan builds locally. I couldn't determine why the CI doesn't hit this case.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/bindings/mod.rs
+++ b/profiling/src/bindings/mod.rs
@@ -3,6 +3,7 @@ mod ffi;
 pub use ffi::*;
 
 use libc::{c_char, c_int, c_uchar, c_uint, c_ushort, c_void, size_t};
+use std::borrow::Cow;
 use std::ffi::CStr;
 use std::marker::PhantomData;
 use std::str::Utf8Error;
@@ -439,7 +440,7 @@ impl TryFrom<&mut zval> for String {
 
             // Safety: checked the pointer wasn't null above.
             let str =
-                unsafe { ddog_php_prof_zend_string_view(zval.value.str_.as_mut()).to_string() };
+                unsafe { ddog_php_prof_zend_string_view(zval.value.str_.as_mut()).into_string() };
             Ok(str)
         } else {
             Err(StringError::Type(r#type))
@@ -447,15 +448,20 @@ impl TryFrom<&mut zval> for String {
     }
 }
 
+/// A non-owning, not necessarily null terminated, not necessarily utf-8
+/// encoded, borrowed string.
+/// It must satisfy the requirements of [core::slice::from_raw_parts], notably
+/// it must not use the nul pointer even when the length is 0.
+/// Keep this representation in sync with zai_str.
 #[repr(C)]
-pub struct ZaiStringView<'a> {
-    len: size_t,
+pub struct ZaiStr<'a> {
     ptr: *const c_char,
+    len: size_t,
     _marker: PhantomData<&'a [c_char]>,
 }
 
-impl<'a> From<&'a str> for ZaiStringView<'a> {
-    fn from(val: &'a str) -> Self {
+impl<'a> From<&'a [u8]> for ZaiStr<'a> {
+    fn from(val: &'a [u8]) -> Self {
         Self {
             len: val.len(),
             ptr: val.as_ptr() as *const c_char,
@@ -464,8 +470,14 @@ impl<'a> From<&'a str> for ZaiStringView<'a> {
     }
 }
 
-impl<'a> ZaiStringView<'a> {
-    pub const fn new() -> ZaiStringView<'a> {
+impl<'a> From<&'a str> for ZaiStr<'a> {
+    fn from(value: &'a str) -> Self {
+        Self::from(value.as_bytes())
+    }
+}
+
+impl<'a> ZaiStr<'a> {
+    pub const fn new() -> ZaiStr<'a> {
         const NULL: &[u8] = b"\0";
         Self {
             len: 0,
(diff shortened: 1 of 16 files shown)


zend_abstract_interface/zai_string/string.h
```diff
@@ -11,25 +11,46 @@
...
-#define ZAI_STR_NEW(data, size)   \
-    (zai_str) {.len = (size), .ptr = (data)}
+/**
+ * ZAI_ASSERT is like ZEND_ASSERT and C assert that it will expand into a valid
+ * expression which returns true (if it fails, it will not return at all).
+ */
+#if NDEBUG
+#define ZAI_ASSERT(cond) true
+#else
+#define ZAI_ASSERT(cond) (assert(cond), true)
```
morrisonlevi, line 28: Should I maybe move this somewhere else, or rename it to be ZA_STR_ASSERT or something?

zend_abstract_interface/zai_string/string.h
```diff
@@ -11,25 +11,46 @@
...
+#endif
```
morrisonlevi, line 29: Should I move this somewhere else, or maybe call it `ZAI_STR_ASSERT`?

zend_abstract_interface/zai_string/string.h
```diff
@@ -11,25 +11,46 @@
...
-    (zai_str) {.len = (size), .ptr = (data)}
+/**
+ * ZAI_ASSERT is like ZEND_ASSERT and C assert that it will expand into a valid
+ * expression which returns true (if it fails, it will not return at all).
+ */
+#if ZEND_DEBUG
+#define ZAI_ASSERT(cond) true
+#else
+#define ZAI_ASSERT(cond) (assert(cond), true)
+#endif
```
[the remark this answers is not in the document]
bwoebi: Maybe not worth it here, but as a note, ZEND_ASSERT() generally leaves a `__builtin_assume()` (or similar depending on compiler) behind in non-debug, as a performance optimization hint for the compiler.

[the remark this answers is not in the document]
morrisonlevi: Ah, yes, I flipped from `NDBEUG` to `ZEND_DEBUG` and didn't flop the branches.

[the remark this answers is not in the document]
morrisonlevi: I added use for `__builtin_assume` in non-debug which will help `clang` compilers, but spending like 2 minutes, I didn't find/think of one that works for GCC in an expression.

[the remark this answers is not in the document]
bwoebi: See: https://github.com/php/php-src/blob/ee82c94208d28958c9133c30d6ac7acdecf54b57/Zend/zend_portability.h#L90-L101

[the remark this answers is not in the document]
bwoebi: Looking at ... https://github.com/php/php-src/blob/ee82c94208d28958c9133c30d6ac7acdecf54b57/Zend/zend_portability.h#L90-L101

I'd expect `(__builtin_expect(c, 0)) || (__builtin_unreachable(), false))` to work. or something similar

zend_abstract_interface/zai_string/string.h
```diff
@@ -1,35 +1,53 @@
...
+
+#include <zai_assert/zai_assert.h>
+
 #include <stdbool.h>
 #include <stddef.h>
 #include <string.h>
 
-#include <Zend/zend.h>
+// never write, only read
+extern char *ZAI_STRING_EMPTY_PTR;
```
bwoebi, line 15: What's this? It seems unused?

morrisonlevi: Ah, probably an accidental merge when I picked a fix from zai-string, I'll remove it.

bwoebi (approved):
Looks good to me :-)

profiling/src/bindings/mod.rs
```diff
@@ -439,23 +440,28 @@ impl TryFrom<&mut zval> for String {
...
         } else {
             Err(StringError::Type(r#type))
         }
     }
 }
 
+/// A non-owning, not necessarily null terminated, not necessarily utf-8
+/// encoded, borrowed string.
+/// It must satisfy the requirements of [core::slice::from_raw_parts], notably
+/// it must not use the nul pointer even when the length is 0.
```
r1viollet, line 454: nit: null

zend_abstract_interface/zai_string/string.h
```diff
@@ -1,35 +1,50 @@
 #ifndef ZAI_STRING_H
 #define ZAI_STRING_H
 
+// Include this before standard headers, or you may get strange errors like:
+// error: unknown type name 'siginfo_t'
+#include <Zend/zend.h>
+
+#include <zai_assert/zai_assert.h>
```
r1viollet, line 8: Do we have debug builds running in our CI checking these asserts ?

bwoebi: Yes.

PR #2217, opened 2023-08-17
Add `zai_string` and use it in `zai_config`

morrisonlevi:
### Description

This adds a `zai_string` struct and associated methods. Part of the `zai_config` code was storing owned pointers inside `zai_str`s, and since it did it more than once, I figured adding a dedicated struct with methods seemed prudent.

A few notes:
 - We don't want to allocate for empty strings. There is some effort made to return a canonical empty string `ZAI_STRING_EMPTY` which always uses `ZAI_STRING_EMPTY_PTR` for the pointer value. This pointer is not freed in `zai_string_dtor`.
 - There are tests for some `zai_string` and `zai_str` functions now.

This also uses the `zai_str_from_zstr` function in the profiler instead of rolling its own `ddog_php_prof_zend_string_view`. This means that at least for now, `zai_str_from_zstr` cannot be purely `static inline`. So I use the standardized C99 `inline` + `extern inline` instead. This probably should have been in its own PR, sorry.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/zend_abstract_interface/symbols/lookup.c
+++ b/zend_abstract_interface/symbols/lookup.c
@@ -122,7 +122,7 @@ static inline zend_class_entry *zai_symbol_lookup_class_impl(zai_symbol_scope_t 
         case ZAI_SYMBOL_SCOPE_NAMESPACE: {
             zai_string key = zai_symbol_lookup_key(scope, name, true);
             result = zai_symbol_lookup_table(EG(class_table),zai_string_as_str(&key),false, true);
-            zai_string_dtor(&key);
+            zai_string_destroy(&key);
         } break;
 
         default:
@@ -165,7 +165,7 @@ static inline zend_function *zai_symbol_lookup_function_impl(zai_symbol_scope_t 
         key.ptr == name->ptr ? zai_symbol_lookup_clean(key) : key,
         scope_type != ZAI_SYMBOL_SCOPE_NAMESPACE, true);
 
-    zai_string_dtor(&tmpkey);
+    zai_string_destroy(&tmpkey);
 
     return function;
 }
@@ -184,7 +184,7 @@ static inline zval* lookup_constant_global(zai_symbol_scope_t scope_type, void *
         key.ptr == name->ptr ? zai_symbol_lookup_clean(key) : key,
         false, true);
 
-    zai_string_dtor(&tmpkey);
+    zai_string_destroy(&tmpkey);
 
     if (!constant) {
         return NULL;
--- a/zend_abstract_interface/symbols/lookup.c
+++ b/zend_abstract_interface/symbols/lookup.c
@@ -77,41 +77,19 @@ static inline zai_str zai_symbol_lookup_clean(zai_str view) {
     return ZAI_STR_NEW(view.ptr + 1, view.len - 1);
 }
 
-static inline zai_str zai_symbol_lookup_key(zai_str *namespace, zai_str *name, bool lower) {
-    zai_str rv;
-    zai_str vns     = zai_symbol_lookup_clean(*namespace);
-    zai_str vn      = zai_symbol_lookup_clean(*name);
+static zai_string zai_symbol_lookup_key(zai_str *namespace, zai_str *name, bool lower) {
+    zai_str vns         = zai_symbol_lookup_clean(*namespace);
+    zai_str separator   = vns.len ? ZAI_STRL("\\") : ZAI_STR_EMPTY;
+    zai_str vn          = zai_symbol_lookup_clean(*name);
+    zai_string rv       = zai_string_concat3(vns, separator, vn);
 
-    char *result = NULL;
-#define CHAR_AT(n) (result[n])
-    if (vns.len) {
-        rv.len = vns.len + vn.len + 1;
-        result = pemalloc(rv.len + 1, 1);
-
-        memcpy(&CHAR_AT(0), vns.ptr, vns.len);
-
-        for (uint32_t c = 0; c < vns.len; c++) {
-            CHAR_AT(c) = tolower(CHAR_AT(c));
-        }
-
-        CHAR_AT(vns.len) = '\\';
-        memcpy(&CHAR_AT(vns.len + 1), vn.ptr, vn.len);
-    } else {
-        rv.len = vn.len;
-        result = pemalloc(rv.len + 1, 1);
-
-        memcpy(&CHAR_AT(0), vn.ptr, vn.len);
+    // Namespaces are never case-sensitive, so they are always lowered, even
+    // if `lower == false`, but do not lowercase the name segment unless
+    // `lower == true`.
+    size_t len = vns.len + separator.len + (lower ? vn.len : 0);
+    for (size_t c = 0; c < len; c++) {
+        rv.ptr[c] = tolower(rv.ptr[c]);
     }
-
-    if (lower) {
-        for (uint32_t c = vns.len; c < rv.len; c++) {
-            CHAR_AT(c) = tolower(CHAR_AT(c));
-        }
-    }
-
-    CHAR_AT(rv.len) = 0;
-#undef CHAR_AT
-    rv.ptr = result;
 
     return rv;
 }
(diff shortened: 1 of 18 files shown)


zend_abstract_interface/symbols/lookup.c
```diff
@@ -142,16 +120,16 @@ static inline zend_class_entry *zai_symbol_lookup_class_impl(zai_symbol_scope_t
         } break;
 
         case ZAI_SYMBOL_SCOPE_NAMESPACE: {
-            zai_str key = zai_symbol_lookup_key(scope, name, true);
+            zai_string key = zai_symbol_lookup_key(scope, name, true);
 
-            if (key.ptr) {
```
morrisonlevi, line 147: Based on the previous implementation of `zai_symbol_lookup_key`, this `key.ptr` should always evaluate to true. If so, we are able to remove this `if` check and just always do the body, right?

bwoebi: right :-)

zend_abstract_interface/zai_string/string.c
```diff
@@ -0,0 +1,7 @@
+#include "string.h"
+
+char *ZAI_STRING_EMPTY_PTR = "";
```
bwoebi, line 3: What's the goal of this variable? Compared to just `""` at usage sites?

zend_abstract_interface/zai_string/string.h
```diff
@@ -167,4 +172,82 @@ bool zai_option_str_get(zai_option_str self, zai_str *view) {
...
+ * outlive the zai_string.
+ */
+__attribute__((pure))
+static inline zai_str zai_string_as_str(const zai_string *string) {
+    zai_str str;
+    memcpy(&str, string, sizeof(zai_str));
+    return str;
+}
+
+static inline zai_string zai_string_ctor(zai_str str) {
```
bwoebi, line 210: I propose `zai_string_from_str()` (in case one wants a `zai_string_from_raw(const char *, size_t)` in future).
Similarly `zai_string_free()` instead of `zai_string_dtor()` for a better C terminology.

zend_abstract_interface/zai_string/string.c
bwoebi: Is this a hack to avoid allocating and freeing an empty string?
I don't like this as it doesn't allow for a trivial realloc(), but one does have to check whether it's ZAI_STRING_EMPTY_PTR.

zend_abstract_interface/zai_string/string.h
```diff
@@ -167,4 +172,82 @@ bool zai_option_str_get(zai_option_str self, zai_str *view) {
...
+    if (str.len == 0) {
+        return ZAI_STRING_EMPTY;
+    }
+
+    // plus 1 for the null byte
+    char *bytes = (char *)__zend_malloc(str.len + 1);
```
bwoebi, line 216: Why here __zend_malloc? Instead of pemalloc(, 1) like in zai_string_concat3()?

zend_abstract_interface/symbols/lookup.c
```diff
@@ -192,28 +165,26 @@ static inline zend_function *zai_symbol_lookup_function_impl(zai_symbol_scope_t
...
 
     zend_constant *constant = zai_symbol_lookup_table(
         EG(zend_constants),
         key.ptr == name->ptr ? zai_symbol_lookup_clean(key) : key,
         false, true);
 
-    if (key.ptr != name->ptr) {
-        pefree((char*) key.ptr, 1);
-    }
+    zai_string_dtor(&tmpkey);
```
bwoebi, line 187:
```suggestion
    if (scope_type == ZAI_SYMBOL_SCOPE_NAMESPACE) {
        zai_string_dtor(&tmpkey);
    }
```
and avoid initializing tmpkey.

zend_abstract_interface/zai_string/string.c
morrisonlevi: Avoiding an allocation for an empty string is better than having an extra check in things like realloc, which notably we do not even use at the moment in `zai_string`.

zend_abstract_interface/symbols/lookup.c
morrisonlevi: What's your thinking? It seems trickier that way, and I can't tell what gain it's supposed to have.

zend_abstract_interface/zai_string/string.h
morrisonlevi: I agree that `zai_string_from_str` makes sense over `zai_string_ctor`, but I think `dtor` should be kept because it's not freeing the pointer, it's freeing the interior contents of the string. You could also have `zai_string_free` and it would make sense:

```c
inline void zai_string_free(zai_string *string) {
    if (string != NULL) {
        zai_string_dtor(string);
        free(string);
    }
}
```

But we don't use that anywhere today, so it doesn't exist.

zend_abstract_interface/zai_string/string.c
r1viollet: `""` would be a string literal even without a variable. Also the compiler would optimize the storage to make sure you have a single string literal even if you have several places where you use `""`.
Or is this to compare the address of the variable ? I might be missing something, I have not checked usage yet.

r1viollet: To be clear I don't mind both, I just did not see where the allocation would be happening.

zend_abstract_interface/zai_string/string.h
bwoebi: You are probably right - it's destruction vs freeing. (Maybe `zai_string_destroy`?)

zend_abstract_interface/zai_string/string.c
bwoebi: @r1viollet This is to compare the address.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -77,42 +77,20 @@ static inline zai_str zai_symbol_lookup_clean(zai_str view) {
...
-        result = pemalloc(rv.len + 1, 1);
-
-        memcpy(&CHAR_AT(0), vn.ptr, vn.len);
+static zai_string zai_symbol_lookup_key(zai_str *namespace, zai_str *name, bool lower) {
+    zai_str vns         = zai_symbol_lookup_clean(*namespace);
+    zai_str separator   = vns.len ? ZAI_STRL("\\") : ZAI_STR_EMPTY;
+    zai_str vn          = zai_symbol_lookup_clean(*name);
+    zai_string rv       = zai_string_concat3(vns, separator, vn);
+
+    // Namespaces are never case-sensitive, so they are always lowered, even
```
r1viollet, line 86: You seem to be re-lowering them all the same.

zend_abstract_interface/symbols/lookup.c
```diff
@@ -77,42 +77,20 @@ static inline zai_str zai_symbol_lookup_clean(zai_str view) {
...
-        memcpy(&CHAR_AT(vns.len + 1), vn.ptr, vn.len);
-    } else {
-        rv.len = vn.len;
-        result = pemalloc(rv.len + 1, 1);
-
-        memcpy(&CHAR_AT(0), vn.ptr, vn.len);
+static zai_string zai_symbol_lookup_key(zai_str *namespace, zai_str *name, bool lower) {
+    zai_str vns         = zai_symbol_lookup_clean(*namespace);
+    zai_str separator   = vns.len ? ZAI_STRL("\\") : ZAI_STR_EMPTY;
+    zai_str vn          = zai_symbol_lookup_clean(*name);
```
r1viollet, line 83: Mostly curious: why are we choosing `vn` as variable name?

zend_abstract_interface/symbols/lookup.c
```diff
@@ -142,16 +120,9 @@ static inline zend_class_entry *zai_symbol_lookup_class_impl(zai_symbol_scope_t
...
-
-            if (key.ptr) {
-                result = zai_symbol_lookup_table(
-                    EG(class_table),
-                    key,
-                    false, true);
-
-                pefree((char*) key.ptr, 1);
-            }
+            zai_string key = zai_symbol_lookup_key(scope, name, true);
```
r1viollet, line 123: It feels strange to create / destroy a memory space at every lookup for concat, though I guess it would pop up if we ever need to optimize this.

zend_abstract_interface/symbols/lookup.c
morrisonlevi: I kept the convention, I don't know why they are what they are 🤷🏻

zend_abstract_interface/symbols/lookup.c
morrisonlevi: Yes, we always have to lower the name because they are always case insensitive. The user may lookup `"MyFoo\Bar"`, or `myfoo\Bar`, and it's the same.

zend_abstract_interface/zai_string/string.h
morrisonlevi: I renamed it to `zai_string_destroy`.

PR #2220, opened 2023-08-18
build: finish hooking up Tea::Php using PhpConfig

morrisonlevi:
### Description

This was apparently merged in a half-baked state. Tea uses the `cmake/Modules/FindPhpConfig.cmake` helper to locate `php-config` and gather a bunch of information about it. Instead of configuring with something like `-DPHP_CONFIG=$(which php-config)`, this uses `-DPhpConfig_ROOT=$(php-config --prefix)` which is a standard CMake idiom.

Then we do that work again but in a different way in `zend_abstract_interface/CMakeLists.txt`. This PR re-uses the work that's already done by `PhpConfig`.

These changes allow me to load up these projects in CLion. For example, for `zend_abstract_interface`, I set these variables with the locations where I have locally installed these things:

```
-DPhpConfig_ROOT=/opt/php/8.2
-DTea_ROOT=/opt/tea/8.2/nts
-DBUILD_ZAI_TESTING=ON
```

Unfortunately I can't set breakpoints on the lines I really want to in the tests thanks to the enormous amount of magic in catch2, but I can at least have a better editing experience with these changes.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ The CI builds are testing these paths.

### Reviewer checklist
- [x] Milestone is set.


bwoebi (approved):
Not an expert in CMake, but let me ack it - it seems to work for you and CI is still happy, so.. :-)

PR #2223, opened 2023-08-21
Implement agent sampling

bwoebi:
### Description

This implements getting the sampling rates back from the agent. The machinery is run through libdatadog, as that code is a) thread safe and will be b) reused for when the sidecar is the trace sender.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


ext/coms.c
```diff
@@ -158,6 +162,13 @@ bool ddtrace_coms_minit(size_t initial_stack_size, size_t max_stack_size, size_t
     atomic_store(&ddtrace_coms_globals.next_group_id, 1);
     atomic_store(&ddtrace_coms_globals.current_stack, stack);
 
+    ddog_Option_VecU8 writer_result = ddog_create_agent_remote_config_writer(&dd_agent_config_writer, &ddtrace_coms_agent_config_handle);
+    if (writer_result.tag == DDOG_OPTION_VEC_U8_SOME_VEC_U8) {
+        ddtrace_bgs_logf("error crating config writer: %.*s", (int)writer_result.some.len, writer_result.some.ptr);
```
morrisonlevi, line 167:
```suggestion
        ddtrace_bgs_logf("error creating config writer: %.*s", (int)writer_result.some.len, writer_result.some.ptr);
```

ext/ddtrace.c
```diff
@@ -262,20 +263,24 @@ static void ddtrace_activate(void) {
...
+        ddtrace_sidecar_ensure_active();
+    }
+
     zend_string *sampling_rules_file = get_DD_SPAN_SAMPLING_RULES_FILE();
     if (ZSTR_LEN(sampling_rules_file) > 0 && !zend_string_equals(get_global_DD_SPAN_SAMPLING_RULES_FILE(), sampling_rules_file)) {
         dd_save_sampling_rules_file_config(sampling_rules_file, PHP_INI_USER, PHP_INI_STAGE_RUNTIME);
     }
 
-    if (strcmp(sapi_module.name, "cli") == 0 && !get_DD_TRACE_CLI_ENABLED()) {
+    if (!DDTRACE_G(disable) && strcmp(sapi_module.name, "cli") == 0 && !get_DD_TRACE_CLI_ENABLED()) {
```
morrisonlevi, line 283: I'm curious why this bit had to change?

ext/coms.c
```diff
@@ -158,6 +162,13 @@ bool ddtrace_coms_minit(size_t initial_stack_size, size_t max_stack_size, size_t
     atomic_store(&ddtrace_coms_globals.next_group_id, 1);
     atomic_store(&ddtrace_coms_globals.current_stack, stack);
 
+    ddog_Option_VecU8 writer_result = ddog_create_agent_remote_config_writer(&dd_agent_config_writer, &ddtrace_coms_agent_config_handle);
```
morrisonlevi, line 165: Not the fault of this PR, but noticed that `MaybeError` is a leaky abstraction when translated to C, as this should be `ddog_MaybeError` and the tag should be something like `DDOG_MAYBEERROR_SOME` or something. Makes it a bit confusing, because when initially reading the code, I was thinking... why do they call `dog_MaybeError_drop` when they have `ddog_Option_VecU8` instead?

Anyway, not something that needs solved in this PR.

src/api/Configuration.php
```diff
@@ -92,8 +92,14 @@ public function isAutofinishSpansEnabled()
     public function getSamplingRate()
     {
         // DD_SAMPLING_RATE is deprecated and will be removed in 0.40.0
```
morrisonlevi, line 94: Looks like it didn't get removed in 0.40.0 after all! 😆

morrisonlevi (commented):
I've scanned through it and don't have much to say. I'll review it more later today when I'm done with meetings.

ext/ddtrace.c
bwoebi: To not promote a DDTRACE_G(disable) being 0 to 2. 0 means generally disabled, and then sidecar won't start at all for example.

src/api/Configuration.php
bwoebi: Yep, it's a simple alias currently :-)

ext/coms.c
bwoebi: I did not author that abstraction and am not too happy with it, but that's what we currently have :-D
I'm open to addressing it in a future change.

ext/ddtrace.c
morrisonlevi: Are the meanings of these values documented somewhere already? Maybe that would have been obvious if I worked on the tracer for my day to day, but if not, I recommend documenting this change or documenting these values somewhere.

bwoebi: It's been boolean in the past until I needed distinction because of CLI_ENABLED, which is evaluated much later than the DDTRACE_G(disabled).

An enum was my thought as well.

dockerfiles/services/request-replayer/src/index.php
morrisonlevi, line 1: So a client sends `next-response` with a payload, and under what circumstances does it get returned?

ext/priority_sampling/priority_sampling.c
```diff
@@ -81,6 +106,28 @@ static void dd_decide_on_sampling(ddtrace_span_data *span) {
...
+                zval *sample_rate_zv = NULL;
+                if (Z_TYPE_P(service) == IS_STRING && env && Z_TYPE_P(env) == IS_STRING) {
+                    zend_string *sample_key = zend_strpprintf(0, "service:%.*s,env:%.*s",(int) Z_STRLEN_P(service), Z_STRVAL_P(service),
+                                                              (int) Z_STRLEN_P(env), Z_STRVAL_P(env));
+                    sample_rate_zv = zend_hash_find(DDTRACE_G(agent_rate_by_service), sample_key);
+                    zend_string_release(sample_key);
+                }
+                if (!sample_rate_zv) {
+                    // Default rate if no service+env pair matches
+                    sample_rate_zv = zend_hash_str_find(DDTRACE_G(agent_rate_by_service), ZEND_STRL("service:,env:"));
```
morrisonlevi, line 123: Seems a bit odd to look up `"service:,env:" in a hash this way, but I'm not against it if that's considered easier or simpler for something.

morrisonlevi (approved):
I don't know how to test the feature, but it looks fine by reading the code and I confirmed it didn't break anything for me in some manual testing, so LGTM.

dockerfiles/services/request-replayer/src/index.php
bwoebi: Generally on the next request - hence the next response :-)

PR #2230, opened 2023-08-28
Windows support for tracing

bwoebi:
### Description

WIP.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


zend_abstract_interface/zai_string/string.h
```diff
@@ -11,13 +11,13 @@ typedef struct zai_string_view_s {
 } zai_string_view;
 
 #define ZAI_STRL_VIEW(cstr) \
-    (zai_string_view) { .len = sizeof(cstr) - 1, .ptr = (cstr) }
```
morrisonlevi, line 14: Microsoft's C support has always been bad, but this is super annoying >.<

bwoebi: Yes, it is. This branch is also still based on quite an old state of the repo, so I'll have to rebase once agent sampling is merged, which will include your newest zai changes...

bwoebi: Fwiw, it's getting better in current (vs 2022), but we need to build older VS versions for PHP.

morrisonlevi: Honestly, if we required a newer toolchain, which versions of PHP would be lost?

bwoebi: All, except the upcoming PHP 8.3 :-)

morrisonlevi: Hmm... I mean, that _is_ tempting. "Yeah, we support Windows now, but you gotta upgrade to PHP 8.3" ;)

But there may be a balance. According to this wiki entry, we could maybe use VS 2019 for PHP 8.X, and its C support does look a lot better:
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-160#c-standard-library-features-1

test.php
```diff
@@ -0,0 +1 @@
+<?php var_dump(\DDtrace\active_span() == true);
```
PROFeNoM, line 1: Not used, right? Or am I missing something?

bwoebi: You aren't missing anything and it shall be removed :-D

zend_abstract_interface/tsrmls_cache.h
```diff
@@ -2,13 +2,13 @@
 
 #include <php.h>
 
-#if PHP_VERSION_ID >= 80000 && PHP_VERSION_ID < 80101 && defined(ZTS)
+#if PHP_VERSION_ID >= 80000 && PHP_VERSION_ID < 80100 && defined(ZTS)
```
PROFeNoM, line 5: Currently, we would run the code for PHP 8.1.0, which is different with this change. I don't know all the implications of this change; I want to be sure we're not breaking something that is untested, and that could be tricky to debug down the road

bwoebi: It doesn't work like that, what's relevant is the PHP_VERSION_ID against which the code is compiled.
The output must be the same regardless of the patch version used.

PROFeNoM (approved):
_That's some excellent work!_ :smiley: 

I obviously don't quite understand all the intricacies, but considering that:
1. The tests are passing
2. Appears to work in customers environment
3. We want to have it released in a **beta state**

I believe we should now ship it and work from there :ship:

In a future that isn't so far, I'll certainly revise the CI's .yml to include code coverage for Windows extension tests.

I approve the PR so we can move forward, as there are no apparent blockers, and this PR has been long in the making now. It ought to be released :)

PR #2234, opened 2023-08-29
Implement sidecar tracing

bwoebi:
### Description

Adds tracing within the sidecar (requirement for windows), disabled by default, windows support will enable it by default on windows specifically.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


dockerfiles/services/request-replayer/src/index.php
```diff
@@ -59,13 +59,14 @@ function logRequest($message, $data = '')
         break;
     default:
         $headers = getallheaders();
-        if (isset($headers['X-Datadog-Diagnostic-Check'])) {
+        if (isset($headers['X-Datadog-Diagnostic-Check']) || isset($headers['x-datadog-diagnostic-check'])) {
```
morrisonlevi, line 62: Just noting libdatadog uses the fact these are case insensitive, but the request replayer isn't currently normalizing them, so it was quick and easy to just check the two variants we get.

ext/configuration.h
```diff
@@ -153,6 +154,7 @@ enum ddtrace_dbm_propagation_mode {
     CONFIG(BOOL, DD_TRACE_DEBUG_CURL_OUTPUT, "false", .ini_change = zai_config_system_ini_change)              \
     CONFIG(INT, DD_TRACE_BETA_HIGH_MEMORY_PRESSURE_PERCENT, "80", .ini_change = zai_config_system_ini_change)  \
     CONFIG(BOOL, DD_TRACE_SIDECAR_TRACE_SENDER, "false", .ini_change = zai_config_system_ini_change)           \
+    CONFIG(BOOL, DD_TRACE_BYPASS_AGENT, "false", .ini_change = zai_config_system_ini_change)                   \
```
morrisonlevi, line 157: I think there is some existing environment variable that's named differently. Ask Brett and other and see if you can find out.

morrisonlevi (approved):
Looks good to me, although I will not pretend I understand everything. I would maybe drop some comments in the stuff dealing with ShmHandle and the mapped ShmHandle, but that's it.

PR #2242, opened 2023-09-05
fix: `datadog-setup.php` when INI file is a symlink

realFlowControl:
### Description

When the INI file found by `datadog-setup.php` is a symlink, the script wont update the INI file correctly.

We found out by a customer complaint. They where installing the tracer via a Debian Package which creates the `/usr/local/etc/php/conf.d/98-ddtrace.ini` as a symlink to `/opt/datadog-php/etc/ddtrace.ini`. Now `sed` is being used with `-i` to update some parts of the INI file, but `sed -i` destroys the symlink and creates a regular file instead which leads to none of the changes are making it to the INI file.

The workaround in that specific case is to not install the tracer via a Debian package, but only use the `datadog-setup.php`,  but still the INI file being a symlink should be supported.

PROF-8145

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -595,16 +595,21 @@ function install($options)
                 }
                 echo "Created INI file '$iniFilePath'\n";
             } else {
-                echo "Updating existing INI file '$iniFilePath'\n";
+                echo "Updating existing INI file '$iniFilePath'";
+                if (is_link($iniFilePath)) {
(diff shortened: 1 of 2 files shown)


bwoebi:
I didn't mean to approve yet, wait with merging until CI is green, please.

Anilm3 (approved):
lgtm

PR #2243, opened 2023-09-05
Work around JITting when using HookData->overrideArguments

bwoebi:
### Description

This PR forcefully disallows inlining the hooked methods where overrideArguments is used.
I'll look at implementing a mechanism directly in php-src to avoid this issue.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


morrisonlevi (approved):
This is quite obviously a terrible hack, and I cannot think of any way to make this better other than just giving up and not enabling the functionality which conflicts. Definitely need to do our best to get something into PHP 8.3+ to avoid these hacks.

PR #2246, opened 2023-09-06
feat(profiling): Do not disable allocation profiling on fixed PHP version with active JIT

realFlowControl:
### Description

In #2088 we disabled allocation profiling on PHP Version >= 8.0.0 when an active JIT is detected. This PR will incorporate the fact that PHP 8.1.21 and 8.2.8 have a fix for this bug, so we will now disable allocation profiling only when JIT is active on PHP 8.0, 8.1.0 - 8.1.20 and 8.2.0 - 8.2.7.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -83,6 +83,21 @@ lazy_static! {
             .expect("Reflection's zend_module_entry to be found and contain a valid string")
     };
 
+    static ref PHP_VERSION_STRUCTURED: Option<(u32, u32, u32)> = {
+        let mut iter = PHP_VERSION.split('.').fuse();
+
+        let major = iter.next()?.parse().ok()?;
+        let minor = iter.next()?.parse().ok()?;
+        let patch = iter
+            .next()?
+            .split(|c: char| !c.is_numeric())
+            .next()?
+            .parse()
+            .ok()?;
+
+        Some((major, minor, patch))
+    };
+
     /// The Server API the profiler is running under.
     static ref SAPI: Sapi = {
         // Safety: sapi_module is initialized before minit and there should be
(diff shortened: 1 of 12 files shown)


profiling/src/lib.rs
```diff
@@ -83,6 +83,21 @@ lazy_static! {
...
+        let patch = iter
+            .next()?
+            .split(|c: char| !c.is_numeric())
+            .next()?
+            .parse()
+            .ok()?;
+
+        Some((major, minor, patch))
+    };
+
```
realFlowControl, line 100: I am wondering if we should try and go for something `PHP_VERSION_ID` like, so that we can just call `php_version_id()` from PHP 8.3 onwards?
https://github.com/php/php-src/pull/11875

realFlowControl: Ah yes, this looks nicer 😄

morrisonlevi (commented):
There should already be a `PHP_VERSION_ID` constant in the bindings, e.g. `zend::PHP_VERSION_ID`. It's the version at compile-time, but of course this simplified certain parts of the PR.

realFlowControl:
You are right about that, but in this case I need the `PHP_VERSION_ID` at runtime as we are building for example on PHP 8.2.8 the Profiler version for PHP 8.2. Someone might use it with PHP 8.2.7 which does not have the JIT patch applied.

morrisonlevi (approved):
Approved from me, but please test my changes out before merging.

realFlowControl:
Clippy complained with https://rust-lang.github.io/rust-clippy/master/index.html#/manual_range_contains
I "fixed" this, added a test and fixed the PHP 8.3 branch. Thanks for approving!

PR #2256, opened 2023-09-13
fix: Use DD_SERVICE if available (Laravel)

### LINKED ISSUE #2252: [Bug]: laravel.request traces/operations no longer being reported
### Bug report

After updating to the latest release, 0.91.0, we longer see any laravel.request operations in APM.  At the same time we now see a few different eloquent (eloquent.get, eloquent.update, etc.) operations coming thru where we had never seen them before.  We have a few different apps that are all exhibiting the same behavior and for now we are just pinning to the previous release as we seem to still get laravel.request traces on 0.90.0.

### PHP version

8.1.17,8.0.30

0.91.0

### Installed extensions

_No response_

### Output of `phpinfo()`

_No response_

### Upgrading from

0.90.0

PROFeNoM:
### Description

Fix #2252 

A mistake was made in #2235 as the value of `DD_SERVICE` wasn't used if available.

Also, test the proper use of `DD_SERVICE`, if available, in the Laravel 9x and 10x suites.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors, the reviewer is in charge of this task.


bwoebi (approved):
Looks good, DD_SERVICE always needs to take priority.

PR #2261, opened 2023-09-18
fix(profiling) store customized heap

realFlowControl:
### Description

In case another extension is installing it's own heap into the ZendMM we start calling this instead of the one that we installed our custom handlers.
This PR will store the heap that we installed our custom handlers in and reuse it.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.


profiling/src/allocation.rs
```diff
@@ -122,12 +125,17 @@ pub fn allocation_profiling_rinit() {
         return;
     }
 
+    unsafe {
+        HEAP = Some(zend::zend_mm_get_heap());
+    }
```
morrisonlevi, line 130: Should this get set to `None` in rshutdown after it's handlers have been reset?

morrisonlevi (approved):
Looks good to me, just one question.

profiling/src/allocation.rs
realFlowControl: You are right

PR #2275, opened 2023-09-21
Set _dd.base_service when overriding service name for a span

pablomartinezbernardo:
### Description

Add `_dd.base_service` to child spans when child service name is updated manually. [AIT-8051](https://datadoghq.atlassian.net/browse/AIT-8051)

Integration tests have been changed to by default have `DD_TRACE_GENERATE_ROOT_SPAN=0`, to minimize the amount of tests that would have needed to be modified to add by adding the `_dd.base_service` tag 

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

[AIT-8051]: https://datadoghq.atlassian.net/browse/AIT-8051?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bwoebi:
Just for the record, the API you are targeting is mostly the "legacy API". A common way to change the service name would be directly editing SpanData, it's going through DDTrace\SpanData.

E.g. (snippet from CurlIntegration):
```php
        \DDTrace\trace_function('curl_exec', [
            'posthook' => function (SpanData $span, $args, $retval) use ($integration) {
                $span->name = $span->resource = 'curl_exec';
                $span->type = Type::HTTP_CLIENT;
                $span->service = 'curl';
// ...
             },
        ]);
```

What you want to do is most likely checking the service of the root span during serialization (in ext/serializer.c) and dynamically set _dd.base_service, if the service is not equal to the root spans service.

ext/serializer.c
```diff
@@ -1256,6 +1256,25 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
+        zval *prop_root_service = &span->root->property_service;
+        ZVAL_DEREF(prop_root_service);
+        zval prop_root_service_as_string;
+        ddtrace_convert_to_string(&prop_root_service_as_string, prop_root_service);
+
+        zval *new_root_name = zend_hash_find(service_mappings, Z_STR(prop_root_service_as_string));
+        if (new_root_name) {
+            zend_string_release(Z_STR(prop_root_service_as_string));
+            ZVAL_COPY(&prop_root_service_as_string, new_root_name);
+        }
```
bwoebi, line 1270: This persists the mapping on the root service the first time a span is serialized.
Especially in future, when we gain partial trace flushing, this will be a quite the spooky action at a distance.

We anyway have to do the lookup each time (to check whether the root span service name is not an unmapped one), so doing that doesn't save us anything. I don't think we should copy the mapped service name back to the root span.

ext/serializer.c
```diff
@@ -1256,6 +1256,25 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
+
+        if (strcasecmp(Z_STRVAL(prop_service_as_string), Z_STRVAL(prop_root_service_as_string)) != 0) {
```
bwoebi, line 1272:
```suggestion
        if (zend_string_equals(Z_STR(prop_service_as_string), Z_STR(prop_root_service_as_string))) {
```
zend_string_equals is aware of the lengths directly and also does a pointer equality comparison (which likely will match if both are equal).

ext/serializer.c
```diff
@@ -1256,6 +1256,25 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
+            zend_array *meta = ddtrace_property_array(&span->property_meta);
+            zend_hash_str_add_new(meta, ZEND_STRL("_dd.base_service"), &prop_root_service_as_string);
```
bwoebi, line 1274:
```suggestion
            zend_hash_str_update(meta, ZEND_STRL("_dd.base_service"), &prop_root_service_as_string);
```
Don't use `_add_new` if you cannot be sure that there's no user-input you could overwrite. If there's already a manually added meta key from PHP code named _dd.base_service, it'll leak. And just `_add` needs to be checked for success or failure to potentially free the value.

ext/serializer.c
```diff
@@ -1256,6 +1256,25 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         }
 
         add_assoc_zval(el, "service", &prop_service_as_string);
+
+        // Add _dd.base_service if service name differs from mapped root service name
+        zval *prop_root_service = &span->root->property_service;
+        ZVAL_DEREF(prop_root_service);
+        zval prop_root_service_as_string;
+        ddtrace_convert_to_string(&prop_root_service_as_string, prop_root_service);
```
bwoebi, line 1264:
```suggestion
        zval prop_root_service_as_string;
        ddtrace_convert_to_string(&prop_root_service_as_string, &span->root->property_service);
```
ddtrace_convert_to_string is able to handle refs directly

bwoebi:
Generally such code just inserting into should be probably moved into `_serialize_meta` directly and avoid altering the meta of the original span itself.

ext/serializer.c
pablomartinezbernardo: 👍 modified the suggestion a bit by using `zend_string_equals_ci` instead of `zend_string_equals` to keep the behavior consistent with `strcasecmp`

bwoebi: Yeah, missed that it wasn't strcmp :-)

bwoebi (approved):
Looks good to me now :-)

ext/serializer.c
```diff
@@ -1090,6 +1090,25 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
...
+
+    zend_array *service_mappings = get_DD_SERVICE_MAPPING();
+    zval *new_root_name = zend_hash_find(service_mappings, Z_STR(prop_root_service_as_string));
+    if (new_root_name) {
+        zend_string_release(Z_STR(prop_root_service_as_string));
+        ZVAL_COPY(&prop_root_service_as_string, new_root_name);
+    }
+
+    if (!zend_string_equals_ci(Z_STR(prop_service_as_string), Z_STR(prop_root_service_as_string))) {
+        add_assoc_str(meta, "_dd.base_service", Z_STR(prop_root_service_as_string));
```
bwoebi, line 1107: Wait, this needs a refcount increase.

ext/serializer.c
```diff
@@ -1090,6 +1090,25 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
         add_assoc_str(meta, "_dd.p.tid", zend_strpprintf(0, "%" PRIx64, span->root->trace_id.high));
     }
 
+    // Add _dd.base_service if service name differs from mapped root service name
+    zval prop_service_as_string;
+    ddtrace_convert_to_string(&prop_service_as_string, &span->property_service);
```
bwoebi, line 1095: And this needs to be released.

bwoebi (approved):

PR #2280, opened 2023-09-22
fix(profiling): use-after-free with Closure trampolines

morrisonlevi:
### Description

A customer reported a crash which lead to the discovery of this use-after-free. Note that this is limited to Closure trampolines, and only when the profiler is enabled. It has been a potential issue for some time, but seems to be more likely to cause a crash since #1998 was merged.

### Readiness checklist
- [x] Tests added for this feature/bug. Yes, but the test only will fail with release asan builds. Currently in CI all ASAN builds are debug builds, which the profiler does not support. We should follow-up in another PR and do release ASAN runs with the profiler.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -1028,6 +1028,24 @@ extern "C" fn interrupt_function_wrapper(execute_data: *mut zend::zend_execute_d
     }
 }
 
+/// Returns true if the func tied to the execute_data is a trampoline.
+/// # Safety
+/// This is only safe to execute _before_ executing the trampoline, because the trampoline may
+/// free the `execute_data.func` _without_ setting it to NULL:
+/// https://heap.space/xref/PHP-8.2/Zend/zend_closures.c?r=af2110e6#60-63
+/// So no code can inspect the func after the call has been made, which is why you would call this function: find out before you
+/// call the function if indeed you need to skip certain code after it has been executed.
+unsafe fn execute_data_func_is_trampoline(execute_data: *const zend::zend_execute_data) -> bool {
+    if execute_data.is_null() {
+        return false;
+    }
+
+    if (*execute_data).func.is_null() {
+        return false;
+    }
+    return ((*(*execute_data).func).common.fn_flags & zend::ZEND_ACC_CALL_VIA_TRAMPOLINE) != 0;
+}
+
 /// Overrides the engine's zend_execute_internal hook in order to process pending VM interrupts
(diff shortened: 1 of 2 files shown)


profiling/tests/phpt/trampolines_02.phpt
```diff
@@ -0,0 +1,29 @@
+--TEST--
+[profiling] use-after-free for inspecting after closure trampoline is called.
+--DESCRIPTION--
+The code for Closure::__invoke will free the `execute_data->func` before it
+returns, but it does not set it to null, except in debug builds:
+https://heap.space/xref/PHP-8.2/Zend/zend_closures.c?r=af2110e6#60-63
+
+Our zend_execute_internal hook inspected the func after the call has been made,
+potentially triggering the issue. This test will likely only fail under asan.
+It's unclear how the customer also got a crash out of it.
```
bwoebi, line 10: It basically crashes if the 18th (ZEND_ACC_CALL_VIA_TRAMPOLINE) bit of the address is not set.
This may be different on different executions, depending on base address (depending on OS), amount of memory allocated before (that's just 256 KB), address randomization etc.

bwoebi (approved):
I wouldn't say that it's unclear how it crashed :-D

But the fix looks good and correct to me :-)

morrisonlevi:
> I wouldn't say that it's unclear how it crashed :-D

Well sure, use-after-frees are obviously wrong even if they don't lead to crashes ^_^

I just couldn't ever get it to crash myself, hence the notes about it only failing in ASAN. I suppose that's clear enough.

realFlowControl (approved):

profiling/src/lib.rs
```diff
@@ -1028,6 +1028,24 @@ extern "C" fn interrupt_function_wrapper(execute_data: *mut zend::zend_execute_d
     }
 }
 
+/// Returns true if the func tied to the execute_data is a trampoline.
+/// # Safety
+/// This is only safe to execute _before_ executing the trampoline, because the trampoline may
+/// free the `execute_data.func` _without_ setting it to NULL:
+/// https://heap.space/xref/PHP-8.2/Zend/zend_closures.c?r=af2110e6#60-63
+/// So no code can inspect the func after the call has been made, which is why you would call this function: find out before you
+/// call the function if indeed you need to skip certain code after it has been executed.
```
realFlowControl, line 1037: 👍 for those comments 🚀

PR #2291, opened 2023-09-28
fix(profiling): do not leak module entry on purpose

morrisonlevi:
### Description

This was done in the early days of the move from C to Rust. The goal at the time was to avoid unsafe code, and that was one way of working around specific requirements. This is a different way, still mostly safe code, but will not leak.

Note that this wasn't a leak that mattered to anything except tooling. The goal is to soon run an NTS+ASAN build for the profiler, and this would trigger the leak sanitizer (implied by address sanitizer on Linux). This is for PROF-8286.

### Readiness checklist
- [x] ~Tests added for this feature/bug.~ The ASAN infrastructure needed is not there yet.

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -29,7 +29,7 @@ use ddcommon::cstr;
 use lazy_static::lazy_static;
 use libc::c_char;
 use log::{debug, error, info, trace, warn, LevelFilter};
-use once_cell::sync::OnceCell;
+use once_cell::sync::{Lazy, OnceCell};
 use profiling::{LocalRootSpanResourceMessage, Profiler, VmInterrupt};
 use sapi::Sapi;
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


profiling/src/lib.rs
```diff
@@ -168,9 +165,10 @@ pub extern "C" fn get_module() -> &'static mut zend::ModuleEntry {
         post_deactivate_func: Some(prshutdown),
         deps: DEPS.as_ptr(),
         ..Default::default()
-    };
+    });
 
-    Box::leak(Box::new(module))
+    // SAFETY: well, it's as least as safe as what every single C extension does.
+    unsafe { &mut *MODULE }
```
realFlowControl, line 171: Just for my understanding:
`Box::new` allocated on the heap and `Box::leak` made sure, that Rust does not care anymore.
With this change to `Lazy` this will now return a mut pointer to this local static and Rust will take care and free this space in the shutdown of the program to not trigger ASAN.
Correct?

realFlowControl (approved):

profiling/src/lib.rs
morrisonlevi: Close. Since it's a local static, it's never freed. It just lives in static memory forever, but this isn't considered a leak, just like having a static `int` isn't a leak.

The main difference aside from the leaking is that before we'd return a new module for every call to `get_module` but now we return the same module. It's expected to only be called once anyway, and the engine will yell at you if you load a module more than once.

But this is doing what every C module does: returning a static module.

PR #2295, opened 2023-09-29
Report extensions used in telemetry

bwoebi:
### Description

Instead of just composer dependencies.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/telemetry.c
+++ b/ext/telemetry.c
@@ -33,6 +33,17 @@ void ddtrace_telemetry_finalize(void) {
 
     ddog_TelemetryActionsBuffer *buffer = ddog_sidecar_telemetry_buffer_alloc();
 
+    zend_module_entry *module;
+    char module_name[261] = { 'e', 'x', 't', '-' };
+    ZEND_HASH_FOREACH_PTR(&module_registry, module) {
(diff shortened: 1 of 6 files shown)


ext/telemetry.c
```diff
@@ -32,6 +32,17 @@ void ddtrace_telemetry_finalize(void) {
 
     ddog_TelemetryActionsBuffer *buffer = ddog_sidecar_telemetry_buffer_alloc();
 
+    zend_module_entry *module;
+    char module_name[261] = { 'e', 'x', 't', '-' };
+    ZEND_HASH_FOREACH_PTR(&module_registry, module) {
```
Anilm3, line 37: Should we report datadog extensions? I recall some discussions about this in other libraries, but I don't have a particularly strong opinion about this.  @Moutix what do you think?

Anilm3 (approved):
lgtm

ext/telemetry.c
bwoebi: Yeah, this pretty much includes all modules, including ext-Core etc. But I'm not sure whether blacklisting a couple ones specifically makes a lot of sense?

PR #2296, opened 2023-09-29
Add release notes for 0.92.1

bwoebi:
A small hotfix with a couple fixes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -678,7 +678,7 @@ dependencies = [
 
 [[package]]
 name = "datadog-php-profiling"
-version = "0.92.0"
+version = "0.92.1"
 dependencies = [
  "ahash 0.8.3",
  "anyhow",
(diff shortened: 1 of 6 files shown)


Anilm3:
I might add a small change to appsec on Monday morning. I'll update this PR once the patch is released in the appsec repository.

morrisonlevi (approved):
LGTM, thanks!

bwoebi:
Okay @Anilm3 - just do a PR to master and we'll push it here afterwards then. (or are you just re-releasing 0.14.1 on appsec? That's fine too)

Anilm3:
> Okay @Anilm3 - just do a PR to master and we'll push it here afterwards then. (or are you just re-releasing 0.14.1 on appsec? That's fine too)

Actually, I might not make it, so please proceed without me and I'll release another patch later on.

PR #2298, opened 2023-10-02
Fix ZAI config handling for .htaccess with php-fpm

bwoebi:
When connecting .htaccess via php-fpm, apache will set true environment variables, a circumstance where the handling explicitly calling ZaiConfigOnUpdateIni was skipped, leading to non-updated configs.

### DIFF
--- a/zend_abstract_interface/config/config_ini.c
+++ b/zend_abstract_interface/config/config_ini.c
@@ -369,37 +369,40 @@ void zai_config_ini_rinit(void) {
 
     for (uint8_t i = 0; i < zai_config_memoized_entries_count; ++i) {
         zai_config_memoized_entry *memoized = &zai_config_memoized_entries[i];
-        if (memoized->ini_change == zai_config_system_ini_change || memoized->original_on_modify) {
+        if (memoized->ini_change == zai_config_system_ini_change) {
             continue;
         }
 
-        for (uint8_t name_index = 0; name_index < memoized->names_count; name_index++) {
-            zai_str name = ZAI_STR_NEW(memoized->names[name_index].ptr, memoized->names[name_index].len);
-            zai_env_result result = zai_getenv_ex(name, buf, false);
+        // makes only sense to update INIs once, avoid rereading env unnecessarily
+        if (!env_to_ini_name || !memoized->original_on_modify) {
+            for (uint8_t name_index = 0; name_index < memoized->names_count; name_index++) {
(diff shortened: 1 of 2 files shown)


zend_abstract_interface/config/config_ini.c
```diff
@@ -369,37 +369,40 @@ void zai_config_ini_rinit(void) {
...
-                 * so we cannot compare the current environment value to the current configuration value
-                 * for the purposes of short circuiting decode
-                 */
-                if (env_to_ini_name) {
-                    zend_string *str = zend_string_init(buf.ptr, strlen(buf.ptr), in_startup);
-
-                    zend_ini_entry *ini = memoized->ini_entries[name_index];
-                    if (zend_alter_ini_entry_ex(ini->name, str, PHP_INI_USER, PHP_INI_STAGE_RUNTIME, 0) == SUCCESS) {
+        // makes only sense to update INIs once, avoid rereading env unnecessarily
+        if (!env_to_ini_name || memoized->original_on_modify) {
```
Anilm3, line 377: Having this condition seems to break some random appsec tests which are changing env variables, albeit not all of them which is somewhat confusing.

It also prevents non-shared variables `.htaccess` from updating.

zend_abstract_interface/config/config_ini.c
```diff
@@ -369,37 +369,40 @@ void zai_config_ini_rinit(void) {
...
+                    /*
+                     * we unconditionally decode the value because we do not store the in-use encoded value
+                     * so we cannot compare the current environment value to the current configuration value
+                     * for the purposes of short circuiting decode
+                     */
+                    if (env_to_ini_name) {
+                        zend_string *str = zend_string_init(buf.ptr, strlen(buf.ptr), in_startup);
+
+                        zend_ini_entry *ini = memoized->ini_entries[name_index];
+                        if (zend_alter_ini_entry_ex(ini->name, str, PHP_INI_USER, PHP_INI_STAGE_RUNTIME, 0) == SUCCESS) {
```
Anilm3, line 392: Prevent changing the setting if it had already been changed by the tracer.

```suggestion
                        if (!ini->modified && zend_alter_ini_entry_ex(ini->name, str, PHP_INI_USER, PHP_INI_STAGE_RUNTIME, 0) == SUCCESS) {
```

zend_abstract_interface/config/config_ini.c
bwoebi: Yes, an `!` is missing.

PR #2299, opened 2023-10-04
feat(profiling) upgrade to `libdatadog` v5

realFlowControl:
### Description

This PR will upgrade our dependencies, but especially upgrade `libdatadog` from v4 to v5, as there are some neat improvements that will especially help with timeline.

I am currently working on a benchmark, this is a scenario where I am sending HTTP Requests for around ~30 seconds to an HTTP endpoint:

|     | v4 | v5 | baseline (no profiling) |
|----|----|---|---|
| Max RSS | 109.527 KB | 64.389 KB | 43.360 KB |
| Max RSS (before serialization) | 60.432 KB | 59.168 KB | - |

PROF-7418

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -47,6 +47,9 @@ use crate::exception::EXCEPTION_PROFILING_INTERVAL;
 
 const UPLOAD_PERIOD: Duration = Duration::from_secs(67);
 
+#[cfg(feature = "timeline")]
+pub const NO_TIMESTAMP: i64 = 0;
+
 // Guide: upload period / upload timeout should give about the order of
 // magnitude for the capacity.
 const UPLOAD_CHANNEL_CAPACITY: usize = 8;
@@ -713,7 +716,7 @@ impl Profiler {
                     },
                     labels,
                     locals,
-                    0,
+                    NO_TIMESTAMP,
                 )) {
                     Ok(_) => trace!(
                         "Sent stack sample of {depth} frames, {n_labels} labels, {alloc_size} bytes allocated, and {alloc_samples} allocations to profiler."
@@ -757,7 +760,7 @@ impl Profiler {
                     },
                     labels,
                     locals,
-                    0,
+                    NO_TIMESTAMP,
                 )) {
                     Ok(_) => trace!(
                         "Sent stack sample of {depth} frames, {n_labels} labels with Exception {exception} to profiler."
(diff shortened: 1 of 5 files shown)


profiling/src/profiling/mod.rs
```diff
@@ -650,14 +651,12 @@ impl Profiler {
                 let depth = frames.len();
                 let (wall_time, cpu_time) = CLOCKS.with(|cell| cell.borrow_mut().rotate_clocks());
 
-                let mut labels = Profiler::message_labels();
+                let labels = Profiler::message_labels();
+                let mut timestamp = 0;
```
danielsn, line 655: Is `0` a valid value for a timestamp?

danielsn (approved):
LGTM.  

As a note outside the scope of this PR: Given that this is rust calling rust, I wonder if we can elide some of the API stuff we did to make things work for FFI and directly use internal structures.

profiling/src/profiling/mod.rs
realFlowControl: Yes and no ;-)
The `0` marks this as there is no timestamp. In https://github.com/DataDog/dd-trace-php/pull/2299/files#diff-3b610cb72989821b75d3c33815a7c7160a76e30418613cdeec640c855b9ca167R384 you can see that I just give this to `NonZeroI64::new()` which would return an Option (and in case it is `0`, that Option would be None). 
I could also have this an optional type (`Option<u64>`), but then I would have to make a check on `None` before calling `NonZeroI64::new()`, otherwise, that constructor would magically to the thing anyway.

realFlowControl:
> As a note outside the scope of this PR: Given that this is rust calling rust, I wonder if we can elide some of the API stuff we did to make things work for FFI and directly use internal structures.

Thanks for the heads up! This PR was meant to make the least amount of changes to make it working with the new libdatadog, but you are right, I'll draft a new PR with using internal structures

danielsn (approved):
One small ⛏️ but otherwise LGTM

profiling/src/profiling/mod.rs
```diff
@@ -1253,7 +1249,7 @@ mod tests {
         locals.profiling_experimental_exception_enabled = true;
 
         let message: SampleMessage =
-            Profiler::prepare_sample_message(frames, samples, labels, &locals);
+            Profiler::prepare_sample_message(frames, samples, labels, &locals, 0);
```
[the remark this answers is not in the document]
realFlowControl: I remove the magic number, thanks for the hint.

morrisonlevi (approved):
Approved, but I'm worried once again at how many tests are failing regularly :/

PR #2305, opened 2023-10-06
WIP: track profiler overhead

realFlowControl:
### Description

This PR aims to track the overhead of profiling.
This is currently in PoC phase 😉 

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


morrisonlevi (commented):
Why is there a new sample type rather than putting them into timeline?

realFlowControl:
I'm just playing with ideas currently. Instead of adding a new sample for every sample that we create we might be able to attach to the sample currently created. It has some drawbacks, as we do not measure the entire round trip and we are for example missing the time it takes to `try_send()` the sample to the other thread.
We are btw. still able to surface these in the timeline, even when the times are in another sample type

PR #2310, opened 2023-10-10
Explicitly drop dd.p.dm if provided in propagated tags when the incoming sampling priority is reject

bwoebi:
### Description

Beforehand, the tracer would pass _dd.p.dm unfiltered, if it was propagated and sampling priorities not changed. I do think this is the more reasonable behaviour, but complying with system tests here.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


PROFeNoM (approved):
:+1:

PR #2314, opened 2023-10-12
Refactor tightly coupled extraction state

bwoebi:
### Description

Moving:
- distributed state directly to the root span
- the extraction logic into its own file, reusable for different purposes (extracting to span, to globals, to span link)
- the trace_id to only the root span
- the parent id to only the root span; the parent id of any other span is simply the id of the parent span

The distributed tracing state is now fully separated between the root span-less (i.e. containing incoming distributed data) and the distributed data tied to a span, allowing it to properly use their own distributed tracing without information leaking across traces.

It also adds a small hexId() helper function on SpanData. SpanLink gains a static fromHeaders() function, which works exactly like DDTrace\consume_distributed_tracing_headers().

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/ext/ddtrace.stub.php
+++ b/ext/ddtrace.stub.php
@@ -55,6 +55,14 @@ namespace DDTrace {
          * @return mixed
          */
         public function jsonSerialize(): mixed {}
+
+        /**
+         * Consumes distributed tracing headers, from which a span link will be constructed.
+         *
+         * @param array|callable(string):mixed $headersOrCallback Either an array with a lowercase header to value mapping,
+         * or a callback, which given a header name for distributed tracing, returns the value it should be updated to.
+         */
+        public static function fromHeaders(array|callable $headersOrCallback): SpanLink {}
     }
 
     class SpanData {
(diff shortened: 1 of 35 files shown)


ext/ddtrace.stub.php
```diff
@@ -136,9 +144,59 @@ public function getStartTime(): int {}
...
+         * This variable cannot be accessed by reference.
+         */
+        public string $parentId;
+
+        /**
+         * @var string The unique identifier for the trace id, as a zero-padded 16 character hexadecimal string.
+         * Assignment of an invalid id will reset the trace id to the default trace id.
+         * This variable cannot be accessed by reference.
+         */
+        public string $traceId = "";
```
PROFeNoM, line 199: Let's say I have a span and want to retrieve the current 128-bit trace ID (zero-padded 32-character hexdec string). How would it be done?

💭 Else, maybe it makes sense to give the possibility to retrieve through a method (e.g., `hexTradeId`)?
```c
PHP_METHOD(DDTrace_SpanData, hexTraceId) {
	ddtrace_span_data *span = OBJ_SPANDATA(Z_OBJ_P(ZEND_THIS));
	RETURN_STR(ddtrace_trace_id_as_hex_string(span->root ? span->root->trace_id : (ddtrace_trace_id){ .low = span->span_id, .high = 0 }));
}
```
 which would prevent the code duplication associated with either:
- The concatenation of `_dd.p.tid` (if any, else zeros) and `RootSpanData->traceId`?
- The conversion from the decimal trace ID returned by `DDTrace\trace_id()` to a zero-padded 32-character hexdec string?

tests/Integrations/Laravel/V5_8/QueueTest.php
```diff
@@ -381,11 +381,7 @@ protected function spanProcessOneJob(
             'laravel_queue_test',
             'queue',
             $resourceDetails
-        )->withExactTags([
-            Tag::HTTP_URL => 'http://localhost:9999/queue/workOn',
-            Tag::HTTP_METHOD => 'GET',
-            Tag::HTTP_STATUS_CODE => 200
```
PROFeNoM, line 387: Why would these tags be removed? Is it because this is not an entrypoint root span?

bwoebi: Yes, exactly

bwoebi: It should not have been there in the first place, but it was there as it was a root span, even if it really is a continuation of something else. Only the root span directly tied to the request should have http info.

ext/ddtrace.stub.php
bwoebi: Eih, it actually is the 32-character id, it's a mistake in the description.

bwoebi:
Note: The system-tests failures are known and need a new appsec version. ddtrace_set_priority_sampling_on_root will be used by appsec instead, but we have a chicken-and-egg problem, whereby we'll need to merge this, before appsec can release theirs.

ext/distributed_tracing_headers.c
```diff
@@ -0,0 +1,400 @@
...
+                if (*ptr == ' ' || *ptr == '\t') {
+                    continue;
+                }
+
+                last_comma = *ptr == ',';
+                // preserve only up to 31 vendor specific values, excluding our own
+                if (last_comma && ++commas == 30) {
+                    --persist;
+                    break;
+                }
```
PROFeNoM, line 278: I know this is not a change associated with this PR, nor is it a blocker, but would you happen to know why this is the case?

PROFeNoM (approved):
lgtm 😃

ext/distributed_tracing_headers.c
bwoebi: Because w3c header consumers are allowed to discard any tracestate exceeding 32 vendor specific values.

PR #2315, opened 2023-10-13
feat(profiling) add phases of inactivity to timeline

realFlowControl:
### Description

This PR will add reporting of inactivity phases in PHP to the timeline:
- calling `sleep()`/`usleep()` in userland
- blocking on `stream_select()`, `uv_run()`, `event_base_loop()`, `\EventBase\loop()` and `\EvLoop\run()` like you'd do in the [ReactPHP](https://github.com/reactphp/event-loop/) or [Revolt](https://github.com/revoltphp/event-loop) event loop implementations (AMPHP uses Revolt as its event loop)
- idle phases between PHP requests like in `php-fpm`, Apache `mod_php` or the PHP build-in webserver (`php -S`)

A thing that is currently open in this PR is this problem:
image
As you can see at the end of the "Main" thread lane, there should also be a grey box indicating and idle phase between requests. Now this box shows up in the next pprof with the idle time being correct. When looking into real world data, this is less of a problem and we might just ignore this for now, especially as the UI handles this pretty well.

- [x] Needs this backend PR: https://github.com/DataDog/profiling-backend/pull/4548

PROF-8370

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/ext/handlers_api.c
+++ b/ext/handlers_api.c
@@ -10,3 +10,15 @@ void datadog_php_install_handler(datadog_php_zif_handler handler) {
         old_handler->internal_function.handler = handler.new_handler;
     }
 }
+
+void datadog_php_install_method_handler(datadog_php_zim_handler handler) {
+    zend_function *old_handler;
+    zend_class_entry *ce = zend_hash_str_find_ptr(CG(class_table), handler.class_name, handler.class_name_len);
+    if (ce != NULL) {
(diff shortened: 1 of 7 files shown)


morrisonlevi:
First thought on the gap: I would probably move the rshutdown to prshutdown.

morrisonlevi:
We are thinking about splitting this into 2 prs. One adds the inactivity between requests and maybe the sleeps, and then everything else goes into a new PR where we think a bit more critically about what we're really trying to do for those.

PR #2316, opened 2023-10-16
feat: add DD_PROFILING_WALLTIME_ENABLED

morrisonlevi:
### Description

This doesn't currently disable wall-time profiles completely, it only cancels the timer. You can still trigger them manually.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -488,9 +488,20 @@ pub enum UploadMessage {
 
 impl Profiler {
     pub fn new(output_pprof: Option<Cow<'static, str>>) -> Self {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "trigger_time_sample")] {
+                // We don't want giant profiles when benchmarking.
+                // We don't want it to block, either.
+                // 1 fits both goals.
+                let message_capacity = 1;
+            } else {
+                let message_capacity = 128;
+            }
+        }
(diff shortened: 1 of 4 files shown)


profiling/tests/overhead/web1/index.php
realFlowControl, line 1: Should this file be empty?

realFlowControl (approved):

profiling/src/profiling/mod.rs
```diff
@@ -488,9 +488,20 @@ pub enum UploadMessage {
...
 impl Profiler {
     pub fn new(output_pprof: Option<Cow<'static, str>>) -> Self {
+        cfg_if::cfg_if! {
+            if #[cfg(feature = "trigger_time_sample")] {
+                // We don't want giant profiles when benchmarking.
+                // We don't want it to block, either.
+                // 1 fits both goals.
+                let message_capacity = 1;
+            } else {
+                let message_capacity = 128;
```
realFlowControl, line 498: We might need to decouple the `message_capacity` from the `trigger_time_sample` feature. The prof-correctness tests are build with `trigger_time_sample` but some of the tests are now failing, as the capacity of the channel is too small and messages are being dropped.

morrisonlevi: Doh, that's obvious in retrospect 🤦🏻

PR #2323, opened 2023-10-23
fix(profiling) exception sampling distance

realFlowControl:
### Description

There were two problems in exception profiling that this PR fixes:
1. `checked_sub()` does not change the value, but return the new one, which we have only checked for `is_none()` and ignored the return value otherwise. We did not update the `self.next_sample` value, effectively walking the stack for every exception.
2. When moving the initialisation to the first `RINIT`, we moved to a too early position in `lib.rs` where `REQUEST_LOCALS` where not yet initialised, resulting in sampling distance always defaulting to `100`.

I also added a test to ensure this is not failing again in the future 😉 

PROF-6915

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/exception.rs
+++ b/profiling/src/exception.rs
@@ -41,9 +41,8 @@ impl ExceptionProfilingStats {
     }
 
     fn track_exception(&mut self, name: String) {
-        let next_sample = self.next_sample.checked_sub(1);
-        if next_sample.is_some() {
-            self.next_sample = next_sample.unwrap();
+        if let Some(next_sample) = self.next_sample.checked_sub(1) {
+            self.next_sample = next_sample;
             return;
         }
 
--- a/profiling/src/exception.rs
+++ b/profiling/src/exception.rs
@@ -41,7 +41,9 @@ impl ExceptionProfilingStats {
     }
 
     fn track_exception(&mut self, name: String) {
-        if self.next_sample.checked_sub(1).is_none() {
+        let next_sample = self.next_sample.checked_sub(1);
+        if next_sample.is_some() {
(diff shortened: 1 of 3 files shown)


profiling/src/exception.rs
```diff
@@ -34,7 +34,9 @@ impl ExceptionProfilingStats {
     }
 
     fn track_exception(&mut self, name: String) {
-        if self.next_sample.checked_sub(1).is_none() {
+        let next_sample = self.next_sample.checked_sub(1);
+        if next_sample.is_some() {
+            self.next_sample = next_sample.unwrap();
```
morrisonlevi, line 39: Obvious opportunity to not have to unwrap (didn't test, browser-written code):

```suggestion
        if let Some(next_sample_ = self.next_sample.checked_sub(1) {
            self.next_sample = next_sample;
```

profiling/tests/phpt/exceptions_01.phpt
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+}
+
+generateExceptions();
+
+echo 'Done.';
+
+?>
+--EXPECTREGEX--
```
morrisonlevi, line 42: Think this is going to be flaky because of the wall time profiler?

realFlowControl: It is stable, in fact the wall time profiler takes a sample, but there is enough `.*` to ignore it. This test is here for two things:
- seeing the initialisation message with the custom sampling rate
- seeing one Exception sample being sent to the profiler thread

morrisonlevi (approved):

PR #2332, opened 2023-10-27
OpenTelemetry API

PROFeNoM:
### Description

**Trace API**
- Span Events and Span Links are missing

**Metrics**
- Not supported

**Log (Appenders)**
- Logs correlation is properly done when using Monolog [log appender](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/supplementary-guidelines.md#how-to-create-a-log4j-log-appender). Still, please review HandlerTest.php to confirm behavior with 128-bit trace identifiers (i.e., trace id: hex, span id: dec)

**Context**
- Spans created using DD's API are **only** converted to OTel spans when the OTel API is used and a context update is triggered (e.g., `Context::getCurrent`). Otherwise, they aren't converted (Reason: Creating 1 OTel span for each DD Span could become expensive for massive traces, considering OTel's interface carries many instances).
- Creating a span using DD's API **assumes** that its scope becomes activated, and activates ancestors from the oldest non-activated ancestor to the newly created span
- Converted spans use a custom InstrumentationScope (named `datadog`) and default resource information

**Autoinstrumentation**
- OTel & DD Auto instrumentation worked concurrently as far as my testing could tell. (TODO: Add some CI tests)
- _Note to self: Look at web.request spans existence_

**Trace SDK**
- Since span events and span links are missing, span events/links limits are missing
- Default span processors and exporters are supported as far as my testing could tell (e.g., Jaeger, see below)
image

**Profiler**
- Works with OTel-created spans (See screenshots below - [Support Admin Link to Trace](https://support-admin.eu1.prod.dog/admin/switch_handle_get/org_id/1000105026?next_url=%2Fapm%2Ftraces%3Fquery%3D%2540_top_level%253A1%2520env%253Aprod%2520-service%253Apdo%26cols%3Dcore_service%252Ccore_resource_name%252Clog_duration%252Clog_http.method%252Clog_http.status_code%26graphType%3Dflamegraph%26historicalData%3Dtrue%26messageDisplay%3Dinline%26query_translation_version%3Dv0%26shouldShowLegend%3Dtrue%26sort%3Dtime%26spanID%3D1242244416008272672%26spanType%3Dservice-entry%26timeHint%3D1699450904217%26trace%3DAgAAAYuvKt6ZVHavhgAAAAAAAAAYAAAAAEFZdXZLdkkxQUFBQ2VJQTlaZmxGTGowUgAAACQAAAAAMDE4YmFmMzQtMTBhMy00MmQwLTkyYjYtZTViYTUwMTkzYjdh%26traceID%3D1242244416008272672%26traceQuery%3D%26view%3Dspans%26start%3D1699437136241%26end%3D1699523536241%26paused%3Dfalse))
image
image

------------------

Misc:
- Datadog sources (`src/` directory) will now be copied to the artifacts. It is necessary for the autoloading to be done when installing with `datadog-setup.php` (the problem doesn't exist with `.apk` or `.deb`)

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.


ext/serializer.c
```diff
@@ -1210,17 +1210,28 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
     add_assoc_long(el, "start", span->start);
     add_assoc_long(el, "duration", span->duration);
 
+    zend_array *meta = ddtrace_property_array(&span->property_meta);
+
     // SpanData::$name defaults to fully qualified called name (set at span close)
-    zval *prop_name = &span->property_name;
+    zval *operation_name = zend_hash_str_find(meta, ZEND_STRL("operation.name"));
+    if (operation_name) {
+        zend_str_tolower(Z_STRVAL_P(operation_name), Z_STRLEN_P(operation_name));
```
bwoebi, line 1218: Use `zend_string_tolower`, this makes a copy if necessary (which you'll have to release then, too). Lowering operation_name directly is problematic if the value isn't RC=1.

ext/serializer.c
```diff
@@ -1268,6 +1294,23 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
 
+    if (span_type) {
+        zend_hash_str_del(meta, ZEND_STRL("span.type"));
+    }
+
+    zval *analytics_event = zend_hash_str_find(meta, ZEND_STRL("analytics.event"));
+    if (analytics_event) {
+        zval analytics_event_as_double;
+        if (Z_TYPE_P(analytics_event) == IS_STRING) {
+            ZVAL_DOUBLE(&analytics_event_as_double, zend_is_true(analytics_event)); // 'true' => 1.0, false => 0.0
```
bwoebi, line 1305: Is `zend_is_true` what you want? The comment sounds like you want the string `"false"` to mean 0?

PROFeNoM: Well, I personally don't want to; the system tests want to (or at least, they want `"true"` to mean 1)

PROFeNoM: More specifically, they want it to mean `"1"`, but I personally don't agree with that as this is inconsistent with the other metrics

README.md
```diff
@@ -3,6 +3,7 @@
 [![CircleCI](https://circleci.com/gh/DataDog/dd-trace-php/tree/master.svg?style=svg)](https://circleci.com/gh/DataDog/dd-trace-php/tree/master)
 [![CodeCov](https://codecov.io/gh/DataDog/dd-trace-php/branch/master/graph/badge.svg?token=eXio8H7vwF)](https://codecov.io/gh/DataDog/dd-trace-php)
 [![OpenTracing Badge](https://img.shields.io/badge/OpenTracing-enabled-blue.svg)](http://opentracing.io)
+[![OpenTelemetry Badge](https://img.shields.io/badge/OpenTelemetry-enabled-blue.svg)](http://opentracing.io)
```
pierotibou, line 6:
```suggestion
[![OpenTelemetry Badge](https://img.shields.io/badge/OpenTelemetry-enabled-blue.svg)](https://opentelemetry.io/)
```

composer.json
```diff
@@ -53,7 +53,7 @@
             "DDTrace\\Tests\\": "./tests/"
         }
     },
-    "minimum-stability": "dev",
+    "minimum-stability": "RC",
```
pierotibou, line 56: no idea what this is. But if it's linked to the fact that otel was RC not so long ago, shall this change?

ext/serializer.c
```diff
@@ -1268,6 +1295,23 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         _add_assoc_zval_copy(el, "type", &prop_type_as_string);
     }
 
+    if (span_type) {
+        zend_hash_str_del(meta, ZEND_STRL("span.type"));
```
pierotibou, line 1299: So many strings repeated everywhere :p Why not constants? (comment can clearly be ignored as not particular to this PR)

tests/OpenTelemetry/Integration/API/BaggageTest.php
```diff
@@ -0,0 +1,108 @@
...
+use DDTrace\Tests\Common\SpanAssertion;
+use DDTrace\Tests\Common\SpanAssertionTrait;
+use DDTrace\Tests\Common\TracerTestTrait;
+use OpenTelemetry\API\Baggage\Baggage;
+use OpenTelemetry\API\Baggage\Propagation\BaggagePropagator;
+use OpenTelemetry\API\Trace\SpanKind;
+use OpenTelemetry\Context\Context;
+use OpenTelemetry\SDK\Trace\TracerProvider;
+
+final class BaggageTest extends BaseTestCase
```
pierotibou, line 16: we support Baggage in PHP?

tests/composer.json
```diff
@@ -115,6 +116,17 @@
                     "create-lockfile": false
                 }
             },
+            "opentelemetry1": {
+                "require": {
+                    "open-telemetry/sdk": "^1.0.0",
```
pierotibou, line 121: Is there a way to ask for latests? It would be great as we would have tests by default against the latest version

pierotibou (commented):
I clearly haven't had time to finish the review. I just browsed through it for now.
Nice work and nice extensive testing.

tests/OpenTelemetry/Integration/API/BaggageTest.php
PROFeNoM: The [Baggage](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/API/Baggage/Baggage.php#L12) API is entirely dependent on the Context API, which we support. The question is, do we fully support it? To that, I would answer that I haven't been able to determine how we wouldn't support it, whether it is in these tests or in an app with `OTEL_PROPAGATORS=baggage,tracecontext`  (which led to the trace in the PR Description).

ext/serializer.c
PROFeNoM: Do you mean that you'd want strings like "span.type", "service.name", "_dd1.sr.eausr", "_dd.span_sampling.rule_rate" (to name a few) to be constants? (Pretty much like the [Tag](https://github.com/DataDog/dd-trace-php/blob/a1f35f6952cff87f5e4c31ee5534eee76d91bcac/src/api/Tag.php#L5) class in PHP?)

tests/OpenTelemetry/Integration/API/BaggageTest.php
pierotibou: yup

ext/serializer.c
bwoebi: I'm not a particularly big fan of constants which are identical to the value they represent (`#define SPAN_KIND "span.kind"`) or such.
For strings which are more complex, sure.

I prefer optimizing the code for readability - I read the code, I immediately see the bare full name of the value.
At least, as long as one is testing the behaviours, it will be caught in testing, should it be mistyped. Constants are mostly a tool to immediately see the issue upon compilation instead of testing. But I think the tradeoff isn't worth it.

composer.json
PROFeNoM: You're right, it was being it was RC. Now that it isn't, it can be reverted :+1:

tests/composer.json
PROFeNoM: Changed it to use `@stable` instead

src/DDTrace/OpenTelemetry/SpanContext.php
```diff
@@ -0,0 +1,130 @@
...
+    public static function create(string $traceId, string $spanId, int $traceFlags = TraceFlags::DEFAULT, ?TraceStateInterface $traceState = null): SpanContextInterface
+    {
+        return API\SpanContext::create($traceId, $spanId, $traceFlags, $traceState);
+    }
+
+    /** @inheritDoc */
+    public static function getInvalid(): SpanContextInterface
+    {
+        return API\SpanContext::getInvalid();
+    }
```
bwoebi, line 118: These methods are not used at all, right?

Overall, it seems like this class is just decorating OpenTelemetry\API\Trace\SpanContext ... with the only goal of on-the-fly generation of the TraceState object?

Wouldn't it be more straightforward to Implement the TraceStateInterface instead and inject that into the API\SpanContext?

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -0,0 +1,212 @@
...
+            $traceFlags = $sampled ? '01' : '00';
+            $traceParent = "00-$traceId-$parentId-$traceFlags";
+            \DDTrace\consume_distributed_tracing_headers([
+                'traceparent' => $traceParent,
+                'tracestate' => (string) $samplingResultTraceState, // __toString() is implemented in TraceState
+            ]);
+        } else {
+            \DDTrace\consume_distributed_tracing_headers([
+                'tracestate' => (string) $samplingResultTraceState, // __toString() is implemented in TraceState
+            ]);
```
bwoebi, line 179: Does this even have any effect? If I read ddtrace_read_distributed_tracing_ids correctly, it does nothing without traceparent.

If you want to update the tracestate only, you should remove any `dd=` part and directly assign to RootSpanData->tracestate.

Also, if we implement my suggestion from another comment about decorating TraceStateInterface instead, this logic can probably just be skipped if we know it's just the unmodified tracestate of the parent.

src/DDTrace/OpenTelemetry/SpanContext.php
PROFeNoM: Let's say we do a `DDTrace\TraceState` class. Wouldn't it also need to generate trace states on the fly? Isn't that counter-intuitive for a class that should represent an immutable list of key/value pairs? I tried to get my head around this today but I feel like I misunderstood you here 😬 

Imo, we need to generate trace states on the fly since distributed attributes can be added at any point in a span's life, hence changing the trace state 🤔

src/DDTrace/OpenTelemetry/SpanBuilder.php
PROFeNoM: > Are we missing a test for this scenario?

I changed a bit that section (with an otherwise failing scenario as well)

src/DDTrace/OpenTelemetry/SpanContext.php
bwoebi: > Wouldn't it also need to generate trace states on the fly

Yes, it would. But, uh, that's sort of "how it is". Like, SpanContext is also supposed to return an immutable tracestate basically, from whatever angle you approach it :-)

ext/handlers_http.h
```diff
@@ -75,6 +75,11 @@ static inline zend_string *ddtrace_format_tracestate(zend_string *tracestate, ze
         smart_str_append(&str, tracestate);
     }
 
+    if (str.s && ZSTR_VAL(str.s)[ZSTR_LEN(str.s) - 1] == ',') {
+        ZSTR_VAL(str.s)[ZSTR_LEN(str.s) - 1] = '\0';
+        ZSTR_LEN(str.s)--;
+    }
+
```
bwoebi, line 82: Note to self: This is really caused by bad handling of the trailing comma in distributed tracing _parsing_ and has been fixed ... on the span links branch.

src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    public function getInstrumentationScope(): InstrumentationScopeInterface
+    {
+        return $this->instrumentationScope;
+    }
+
+    public function hasEnded(): bool
+    {
+        return $this->hasEnded;
```
bwoebi, line 158: Is this intentionally just checking for hasEnded and not getDuration() != null, which is only triggered when Context::current is called?

```
start_span();
$otelSpan = Context::current();
close_span();
var_dump($otelSpan->hasEnded()); // false?
```

PROFeNoM: Speaking about `toSpanData`, that just made me realized that... the attributes may be wrong if someone has the wonderful idea of _unsetting_ attributes from the DD-span...

PROFeNoM: Especially if you unset

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -0,0 +1,219 @@
...
+        $hexSpanId = $span->hexId();
+        $spanContext = DDTraceAPI\SpanContext::createFromLocalSpan($span, $sampled, $traceId, $hexSpanId);
+
+        if (!in_array($samplingDecision, [SamplingResult::RECORD_AND_SAMPLE, SamplingResult::RECORD_ONLY], true)) {
+            return Span::wrap($spanContext);
+        }
+
+        $span->resource = $this->spanName; // OTel.name => DD.resource
+
+        $attributesBuilder = clone $this->attributesBuilder; // According to OTel's spec, attributes can't be changed after span creation...
```
PROFeNoM, line 175: Note to self: cloning may not be necessary anymore

src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -0,0 +1,392 @@
...
+
+        // After closing the span, the context won't change, but would otherwise be lost
+        $this->context = API\SpanContext::create(
+            $this->context->getTraceId(),
+            $this->context->getSpanId(),
+            $this->context->getTraceFlags(),
+            $this->context->getTraceState()
+        );
+
+        $this->spanProcessor->onEnd($this);
```
bwoebi, line 376: Not sure about this:
- Given that endOTelSpan is only called on Context::current(), if there's no further access to Context::current() after terminating the span, the spanProcessor->onEnd() will not be called
- this only runs for datadog-only Spans which happen to be accessed at least once by a Context::current(), right?
- I'm not sure what the best behaviour is, maybe, for consistency _exclusively_ run it on OTel managed spans?

src/DDTrace/OpenTelemetry/SpanBuilder.php
bwoebi: You're still assigning sampling attributes, thereby having a side-effect... I'd leave it as is.

bwoebi (approved):
There are some minor tradeoffs in the code, like Span::endOTelSpan fetching getTraceState() - which incidentally is only called when the next Context::current() call happens.
There might be theoretical divergence, but as long as we acknowledge such edge cases, I think it's fine.

I've left one more point for consideration, but overall looks good I think :-)

src/DDTrace/OpenTelemetry/Span.php
PROFeNoM: > Given that endOTelSpan is only called on Context::current(), if there's no further access to Context::current() after terminating the span, the spanProcessor->onEnd() will not be called

Yes. One alternative I also tried would have been to do an OpenTelemetryIntegration, which traces a span's termination... but faced limitations. To this day, I still think this would have been the cleanest way overall (i.e., have sort of an event when a DD span is created/closed without the OTel API so that we could simply instrument them and start/end OTel spans as needed).

> this only runs for datadog-only Spans, which happen to be accessed at least once by a Context::current(), right?

Yes

> I'm not sure what the best behaviour is, maybe, for consistency exclusively run it on OTel managed spans?

For consistency, it may indeed make sense

PROFeNoM:
@bwoebi 

The changes made on the endEpochNanos/hasEnded 'flags' to simplify the code, although not impacting at all DD's end result, impact Jaeger traces.

I'll be running the span processors **after** the Datadog span termination. It may conflict with partial flushing in the future. Alternatively, I can just rollback the changes made and use some flags, but it fundamentally conflicts as well.

We can discuss this in our meeting later on today if you care about this.

**Before commit:**
image

**After commit:**
image

image


[Datadog Trace](https://support-admin.eu1.prod.dog/admin/switch_handle_get/org_id/1000105026?next_url=%2Fapm%2Ftraces%3Fquery%3D%2540_top_level%253A1%2520env%253Aprod%2520-service%253Apdo%26cols%3Dcore_service%252Ccore_resource_name%252Clog_duration%252Clog_http.method%252Clog_http.status_code%26graphType%3Dflamegraph%26historicalData%3Dfalse%26messageDisplay%3Dinline%26query_translation_version%3Dv0%26shouldShowLegend%3Dtrue%26sort%3Dtime%26spanID%3D5065466138805401797%26spanType%3Dservice-entry%26timeHint%3D1700130445881%26trace%3D1137989916655637208016674789262054930497%26traceID%3D11379899166556372080%26traceQuery%3D%26view%3Dspans%26start%3D1700129675081%26end%3D1700130575081%26paused%3Dfalse)

bwoebi:
I think that's fine. Even if we implement partial flushing in future, we probably shouldn't _immediately, irreversibly_ flush the span, but flush it slightly delayed, so that some data may still be set on the closed span, otherwise we risk breaking currently perfectly working code from other users, which nonchalantly add some meta data to the span after closing it.

PR #2343, opened 2023-11-01
feat(profiling): add idle phases to timeline

realFlowControl:
### Description

This PR will add idle phases between requests and idle phases due to the `sleep()`-class functions to the timeline view.
In detail this covers:
- idle phase between RSHUTDOWN and RINIT (PHP process not handling a request)
- and PHP process sleeping due to a sleep-class function call, namely
  - `sleep()`
  - `usleep()`
  - `time_nanosleep()`
  - and `time_sleep_until()`

Also includes one unrelated change: Moves the call to `exception::exception_profiling_mshutdown();` a few lines up, so that it happens before the config and the profiler is shutdown.

This PR superseeds #2315 and will only contain idle phases as in sleeping or waiting to handle a new request in synchronous PHP (as used in PHP-FPM or Apache mod_php). Everything else (like event loop idle phases) shall go to another PR.

image

PROF-8370 

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -363,6 +363,9 @@ extern "C" fn prshutdown() -> ZendResult {
 
     TAGS.with(|cell| cell.replace(Arc::default()));
 
+    #[cfg(feature = "timeline")]
+    timeline::timeline_prshutdown();
+
     ZendResult::Success
 }
 
@@ -747,9 +750,6 @@ extern "C" fn rshutdown(_type: c_int, _module_number: c_int) -> ZendResult {
 
     #[cfg(feature = "allocation_profiling")]
     allocation::allocation_profiling_rshutdown();
-
-    #[cfg(feature = "timeline")]
-    timeline::timeline_rshutdown();
 
     ZendResult::Success
 }
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -639,6 +639,9 @@ extern "C" fn rinit(_type: c_int, _module_number: c_int) -> ZendResult {
     #[cfg(feature = "allocation_profiling")]
     allocation::allocation_profiling_rinit();
 
+    #[cfg(feature = "timeline")]
+    timeline::timeline_rinit();
+
     ZendResult::Success
 }
 
@@ -744,6 +747,9 @@ extern "C" fn rshutdown(_type: c_int, _module_number: c_int) -> ZendResult {
 
     #[cfg(feature = "allocation_profiling")]
     allocation::allocation_profiling_rshutdown();
+
+    #[cfg(feature = "timeline")]
+    timeline::timeline_rshutdown();
 
     ZendResult::Success
 }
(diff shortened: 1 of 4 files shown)


morrisonlevi:
> Also includes one unrelated change: Moves the call to exception::exception_profiling_mshutdown(); a few lines up, so that it happens before the config and the profiler is shutdown.

It seems sensible, but could you expand on that a little bit more? I wonder if there's some pattern we should employ here.

realFlowControl:
> > Also includes one unrelated change: Moves the call to exception::exception_profiling_mshutdown(); a few lines up, so that it happens before the config and the profiler is shutdown.
> 
> It seems sensible, but could you expand on that a little bit more? I wonder if there's some correct-by-construction pattern we should employ here.

The profiler is not needed in the `exception::exception_profiling_mshutdown()`, but it is needed to be around and valid in the `timeline::timeline_mshutdown()` as in that function we close the idle time box that was started in the last `timeline::timeline_prshutdown()`. This is done in order to report the last idle time when shutting down the PHP process.

This involves also keeping the static `TAGS` around and not clean them in `prshutdown()`. This is something that might have side effects that I could not foresee, so please pay special attention to this change when doing the review.

We might not need this in case we opt-in to just ignore the "last" idle timebox, which might be a corner case anyway, as this is the special case where the process shuts down.

profiling/src/lib.rs
```diff
@@ -634,11 +635,16 @@ extern "C" fn rinit(_type: c_int, _module_number: c_int) -> ZendResult {
                 profiler.add_interrupt(interrupt);
             }
         });
+    } else {
+        TAGS.with(|cell| cell.replace(Arc::default()));
```
morrisonlevi, line 639: Did you need this for any particular reason? It makes sense theoretically, just wondering about specifics.

realFlowControl: I think we might not need it after all. I moved it from `RSHUTDOWN` to here for just the one reason of having the `TAGS` in a defined state when used. If we do not cleanup, `TAGS` leak from the last request that had profiling enabled and might be wrong. Now the thing is: we are not using the `TAGS` when profiling is disabled.
I am totally fine with removing this entirely, but I'd like to document this somewhere (at best in Rust itself, otherwise a comment on the `TAGS` variable should also work).
We can add those kind of information/documentation also in another PR.

realFlowControl (commented):
LGTM 🚢

profiling/src/lib.rs
morrisonlevi: I think it's consistent. Previously we cleared tags in prshutdown and conditionally re-initialized them in rinit. Now we leave them alone in prshutdown, and either re-initialize them or reset them in rinit. These are roughly equivalent, except for the case that if you go from `prshutdown` to `mshutdown`, they'll still be valid, which was your motiviation. Seeems good to me.

morrisonlevi (approved):
As long as you've checked that it actually still works after my commit (not just that the commits look okay) then I'm good with it! The prof-correctness tests pass which is a good sign, obviously, but I still like to check this kind of thing manually.

PR #2350, opened 2023-11-08
perf(profiling): increase allocation sampling distance

morrisonlevi:
### Description

This causes fewer stack walks due to allocation profiling.

A customer recently complained about performance overhead, and in their case, the allocation profile was more than twice as expensive as the wall-time/cpu-time profiles.

I did some testing with the symfony-demo application as well, and also came up with ~2x cost. So... doubling this number brings allocation profiling costs roughly in line with wall-time/cpu-time, at least for a few recent tests (overhead can vary substantially based on the test).

Part of PROF-8596.

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/allocation.rs
+++ b/profiling/src/allocation.rs
@@ -33,8 +33,8 @@ pub fn allocation_profiling_minit() {
     unsafe { zend::ddog_php_opcache_init_handle() };
 }
 
-/// take a sample every 2048 KB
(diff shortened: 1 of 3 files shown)


realFlowControl:
@morrisonlevi I fixed the prof correctness test

realFlowControl (approved):

piotr-oles (approved):

MaximeLaplace (approved):

PR #2352, opened 2023-11-10
fix(profiling): potential allocation profiling crashes with certain opcodes

morrisonlevi:
### Description

**TODO:** update the whole description for state of the current patch.

On PHP 7.4 and on PHP 8.0 before 8.0.26, the engine doesn't `SAVE_OPLINE();` for the `ZEND_GENERATOR_CREATE` opcode. If the allocation profiler triggers on an allocation in this state, then it will sigsegv because the opline is non-null but invalid.

Note that the `SAVE_OPLINE();` is missing on PHP < 7.4 as well, but on those versions `i_init_func_execute_data` initializes the opline. Coupled with the new test, we are confident that we don't need to install an opcode handler for those versions.

#### Who is affected by the crash?

 - If you are running PHP 8.0.0 up to and including 8.0.25, then you are affected if you use a generator and the allocation profiler is enabled, and it's been enabled by default since dd-trace-php 0.89.0 (todo: double-check).
 - On PHP 7, if you not running the tracer (only the profiler), then you are affected. This configuration is very rare, generally only Datadog engineers run this configuration.

Everything else is unaffected. To be clear:
 - On PHP 7 you are fine if you also run the tracer (highly likely).
 - On PHP 8, if you are on 8.0.26 or higher, then you are unaffected.
 - On MacOS you are fine

In summary, no known customers are affected. Nearly everyone on PHP 7 runs the tracer, and nearly everyone is on a patched version of PHP 8.0, or on PHP 8.1 or newer. If you are on an affected version of PHP 8.0, as a quick mitigation you can disable the allocation profile by setting the ini `datadog.profiling.allocation_enabled=0`. You can do this with config mode of the setup script:

```bash
# Do this after `php datadog-setup.php --enable-profiling ...`
php datadog-setup.php config set --php-bin all \
    -ddatadog.profiling.allocation_enabled=0
```

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/profiling/src/php_ffi.c
+++ b/profiling/src/php_ffi.c
@@ -1,12 +1,19 @@
 #include "php_ffi.h"
 
 #include <assert.h>
-#include <stdbool.h>
+#include <stdlib.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <string.h>
 
 #if CFG_STACK_WALKING_TESTS
 #include <dlfcn.h> // for dlsym
+#endif
+
+#if PHP_VERSION_ID >= 70400
+#define CFG_NEED_OPCODE_HANDLERS 1
+#else
+#define CFG_NEED_OPCODE_HANDLERS 0
 #endif
 
 const char *datadog_extension_build_id(void) { return ZEND_EXTENSION_BUILD_ID; }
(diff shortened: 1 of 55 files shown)


realFlowControl (approved):

profiling/src/php_ffi.c
```diff
@@ -75,9 +75,53 @@ static post_startup_cb_result ddog_php_prof_post_startup_cb(void) {
...
+     * so nearly no users will triggered this one.
+     */
+    if (php_version_id < 80026 && zend_get_user_opcode_handler(ZEND_GENERATOR_CREATE) == NULL) {
+        user_opcode_handler_t handler = (user_opcode_handler_t)ddog_php_prof_opcode_dispatch;
+        zend_set_user_opcode_handler(ZEND_GENERATOR_CREATE, handler);
+    }
+
+    /* Issue is fixed in 8.0.12:
+     * https://github.com/php/php-src/commit/ec54ffad1e3b15fedfd07f7d29d97ec3e8d1c45a
+     * The tracer didn't save us here. Fortunately, throwing destru
```
morrisonlevi, line 112: todo: fix this comment lol

realFlowControl: Oh I remember. It is not only about fixing the comment, but try and see if the fix is enough, as there is a `zend_array_dup()` call few lines before the `SAVE_OPLINE()` happens. I'll check if I can trigger this, I highly assume it will crash as well.

morrisonlevi: Yes, we need to adjust the version range too. If it's still on PHP 8.3, then we need to install the opcode handler on more versions too.

realFlowControl: Jup, we need to install up to latest, can you check [ad46787](https://github.com/DataDog/dd-trace-php/pull/2352/commits/ad467876e9bad7f9cb9e826db595619ce5ee911d)

morrisonlevi: Looks good. I add another commit that cleans things up and adds a pointless test for `ZEND_BIND_STATIC` with a todo.

morrisonlevi:
I moved this back to draft. I need to figure out:

- Does the JIT do anything with these specific opcodes?
- If not, then bypass the handler check by installing these opcodes after they are checked.
- If so, rework the phpt files or use skipifs or something to accommodate the optional JIT warning about getting disabled if there are user opcode handlers.

morrisonlevi:
`ZEND_GENERATOR_CREATE` is definitely referred to in the JIT a few times, but as far as I can tell, it's about determining exit/entry points, and nothing really about replacing the opcode itself. However, I am not the best at reading this JIT code.

`ZEND_BIND_STATIC` is never referred to in the JIT. I think we can safely install this opcode handler without affecting the JIT.

realFlowControl:
@morrisonlevi I added `ZEND_FUNC_GET_ARGS` and finalised the PHP versions, now that the PRs have been merged and released.

morrisonlevi:
@realFlowControl Thank you! It looks good to me. I merged in master and added your upstream phpt with a long-form description. Can you do a final code review?

bwoebi (approved):
Looks good to me.

I also like set_run_time_php_version_id being moved to C code, looks much more readable there.

realFlowControl (approved):

PR #2366, opened 2023-11-17
Fix orphans removal when 128-bit is enabled

PROFeNoM:
### Description

Complements #2358.

`\DDTrace\trace_id()`, with 128-bit trace IDs, is 32-character long, whereas `$span->id` was 16-character long, failing the comparison and the orphan removal. Instead, this PR uses `\DDTrace\root_span()->id` and adds some regression tests for the Redis & Predis integrations.

### Readiness checklist

### Reviewer checklist
- [X] Milestone is set.
- [X] Changelog has been added to the release document. For community contributors, the reviewer is in charge of this task.


bwoebi:
There's a difference in old vs new behaviour: Old behaviour only removed it if it was a _trace root_. Now, it also removes it for _local root_ spans.

I think the proper check would be `$span instanceof RootSpanData && empty($this->parentId)`.

bwoebi (approved):

PR #2367, opened 2023-11-17
Completely avoid instrumenting the PSR NullLogger

bwoebi:
### Description

This is implemented by adding the ability to remove a hook from selected classes directly.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

### DIFF
--- a/zend_abstract_interface/hook/hook.c
+++ b/zend_abstract_interface/hook/hook.c
@@ -17,6 +17,7 @@ typedef struct {
     bool is_abstract;
     zend_long id;
     int refcount; // one ref held by the existence in the array, one ref held by each frame
+    HashTable exclusions;
 } zai_hook_t; /* }}} */
 
 typedef struct _zai_hooks_entry {
@@ -106,6 +107,10 @@ static void zai_hook_data_dtor(zai_hook_t *hook) {
 
     if (hook->function) {
         zend_string_release(hook->function);
+    }
+
+    if (hook->exclusions.arData) {
+        zend_hash_destroy(&hook->exclusions);
     }
 }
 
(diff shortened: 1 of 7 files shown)


zend_abstract_interface/hook/hook.c
```diff
@@ -226,6 +231,16 @@ static void zai_hook_hash_destroy(zval *zv) {
     efree(hooks);
 }
 
+static inline bool zai_hook_is_excluded(zai_hook_t *hook, zend_class_entry *ce) {
+    if (hook->exclusions.arData && ce) {
+        zend_string *lower = zend_string_tolower(ce->name);
+        bool excluded = zend_hash_exists(&hook->exclusions, lower);
```
PROFeNoM, line 237: That means we only strictly check for the equality of names, which means a class extending `NullLogger` would also get hooked.

I think we do have to iterate through the interfaces `ce->interfaces` + the parents as well

PROFeNoM: In the .phpt, I think that if we add a class `GrandChild` inherits `Child`, it would get hooked

bwoebi: Okay, it's a bit more complicated with interfaces: Every class implements all the interfaces any of its parents implement.

The underlying problem is, given a hook on X::foo():
```
interface X { function foo(); }
class A implements X {  function foo() { ... } }
class B extends A { function foo() { ... } }
class C extends B implements X {}
```

I think we agree that if a class (here: B) is directly implementing A, it probably shouldn't be hooked if A::foo() is dropped.
However, now C implements X directly. Should or should it not inherit? The catch though here is that PHP does not expose whether the interface is inherited from the parent only or directly inherited by the child.

I think it's fine to ignore "implements directly".

PROFeNoM (approved):
Nice 😃

PR #2370, opened 2023-11-17
Use _dd.agent_psr instead of _dd.rule_psr for agent sampling

bwoebi:
### Description

Also specify default sampling mechanism when neither rule, nor agent rules apply.

Also fixing the fact that cached agent rules were freed after a request, but not re-read until they changed. They are properly persisted now.

Refactored persisting json data at the same time, to avoid unnecessary copies as much as possible, with code common for zai config and json.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


zend_abstract_interface/json/json.c
```diff
@@ -55,5 +135,140 @@ bool zai_json_setup_bindings(void) {
         php_json_serializable_ce = *tmp_json_serializable_ce;
     }
 
-    return zai_json_encode && zai_json_decode_ex;
+    return zai_json_encode != NULL;
```
PROFeNoM, line 138: Not totally sure why we don't check for zai_json_decode_ex for PHP < 70100, or zai_json_parse & zai_json_parser_init for the other

PROFeNoM (approved):
I have some doubts about a small thing, but I don't think it is impactful anyway :+1:

zend_abstract_interface/json/json.c
bwoebi: It's mostly - if one symbol is failing, the others will too, avoiding the extra check :-D.

PR #2372, opened 2023-11-20
Fix possible double-free with hooks on internal methods inherited by user classes

bwoebi:
### Description

Doing the due diligence whether I can reuse nNumOfElements in the HashTable, to increase and decrease it, so that all checks for zend_hash_num_elements() == 0  only are true when neither internal duplicates exist, nor actual hooks exist, I missed the tiny detail, that, while all direct usages of nNumOfElements were safe, there was an usage of nNumOfElements in the HT_IS_WITHOUT_HOLES() macro, checking whether nNumOfElements == nNumUsed.

Thus, when there were two hooks, one later removed, and one class inheriting the internal method, nNumOfElements happened to be equal to nNumUsed in shutdown. During hash destruction the smart Zend code would check this and unconditionally iterate over all elements of the array, skipping IS_UNDEF checks.

Eventually leading to a double free.
As a saving grace, this double free generally happened very late in the shutdown sequence, so the memory corruption would usually not surface...

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


PROFeNoM (approved):
Nice catch

Issue #2376, opened 2023-11-21
[Bug]:  0.94.0 build failure


remicollet:
### Bug report

For ZTS builds:

```
[...]

PR #2353, opened 2023-11-10
Workaround php bug #81634 in PHP 8.0-8.1.0

cataphract:
### Description

Workaround for php bug #81634 in PHP 8.0-8.1.0

### Readiness checklist
- [ ] (only for Members) Changelog has been added to the release document.
- [x] Tests added for this feature/bug.

### Reviewer checklist
- [ ] Appropriate labels assigned.
- [ ] Milestone is set.
- [ ] Changelog has been added to the release document. For community contributors the reviewer is in charge of this task.

PR #2383, opened 2023-11-21
Use global-dynamic for tsrmls cache

cataphract:
### Description

PR #2353 uses no attribute in ddtrace.c TSRMLS_CACHE, but ddtrace.h has a declaration with local-dynamic. This seems to generally work, with separate entries in the GOT for each case (only module id for the local one and both global id and offset for the second). However, it seems that LTO is not happy with this (see https://github.com/DataDog/dd-trace-php/issues/2376 ).

So change all definitions and declarations to global-dynamic.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


Anilm3 (commented):
To summarise our conversation:

- Currently we have both `local-dynamic` and `global-dynamic` versions of `TSRMLS_CACHE`, which in theory shouldn't be done but in practice works fine.
- The above somehow fails with a segfault when using LTO, but it's unclear why, perhaps some optimisation that is not expecting this mixed TLS model situation, which sounds reasonable.
- Setting the TLS model of the `TSRMLS_CACHE` definition to `local-dynamic` isn't respected by the compiler so we're forced to use `global-dynamic`.

Assuming that's accurate, it would be useful to have a comment somewhere explaining this.

bwoebi:
Should we also add the attribute in ddtrace.c for proper visibility?

cataphract:
> Should we also add the attribute in ddtrace.c for proper visibility?

I think I tried that in the buster image and it didn't work, at least in the super old compiler in the buster image. Your patch with local-dynamic generates a segfault (the one I posted on slack), and I think with global-dynamic the same. IIRC it attempted to import the variable instead of defining it in the module. I must say I didn't get those problems in gcc-12 though.

Anyway, global-dynamic should be the default (unless overridden on the command line), at least judging from the error message that remi posted and from the assembly generated.

bwoebi (approved):
Okay, then let me ack it, given that it doesn't break anything, and it apparently fixes the lto builds :-)

remicollet:
I c onfirm ZTS + LTO build is fixed (tested with 8.2)

bwoebi:
Thanks a lot for verifying @remicollet ! :-)

PR #2384, opened 2023-11-22
Implement extended sampling

bwoebi:
### Description

- Support for tags and resources in sampling rules
- Allows looking at arbitrary spans too, not only root spans
- Sampling rules now use glob pattern matching

**Sampling rules use glob pattern matching**
We used to have regexes, so it would be a breaking change. To work around this, we add a new `DD_TRACE_SAMPLING_RULES_FORMAT` parameter. This parameter allows `regex` and `glob`. `regex` is the default for retrocompatibility reasons. That said we will drop this parameter in 1.0.

**Glob matching algo**
It already existed for single spans sampling. So we're just factorizing some code. The previous algo had a small bug so there's one commit dedicated to this fix.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.


tests/ext/priority_sampling/020-rule-nonroot.phpt
```diff
@@ -0,0 +1,90 @@
...
+$child->service = "bar";
+
+var_dump(\DDTrace\get_priority_sampling() == \DD_TRACE_PRIORITY_SAMPLING_USER_KEEP);
+
+if (!isset($root->metrics["_dd.agent_psr"]) && $root->metrics["_dd.rule_psr"] == 0.7) {
+    echo "Rule OK\n";
+} else {
+    var_dump($root->metrics);
+}
+
```
PROFeNoM, line 50: Note to self: What if we then change child's service to foo.

ext/priority_sampling/priority_sampling.c
```diff
@@ -63,55 +61,159 @@ static bool dd_rule_matches(zval *pattern, zval *prop) {
...
+        if (!eval_root) {
+            zval *applies = zend_hash_str_find(Z_ARR_P(rule), ZEND_STRL("target_span"));
+            if (!applies || Z_TYPE_P(applies) != IS_STRING || !zend_string_equals_literal(Z_STR_P(applies), "any")) {
                 continue;
             }
+        }
 
-            bool rule_matches = true;
+        zval *sample_rate_zv;
+        if (dd_check_sampling_rule(Z_ARR_P(rule), span) && (sample_rate_zv = zend_hash_str_find(Z_ARR_P(rule), ZEND_STRL("sample_rate")))) {
```
PROFeNoM, line 148: Hum, I'm not sure that it is taking into account that the default value of `sample_rate` is `1.0`. From what I see/understand, currently, if a rule doesn't include the `sample_rate` tag, then even if `dd_check_sampling_rule` evaluates to true, the sample rate won't be changed. I believe case 2 or 4A from the RFC would fail.

bwoebi: I see, it defaults to 1, right.

bwoebi: We already required sample_rate to be present in the past, and it's not been an explicit change in the RFC, so it must have been always wrong :-)

PROFeNoM (approved):
LGTM with respect to the RFC :+1:

ext/ddshared.h
```diff
@@ -6,5 +6,7 @@
 extern zend_string *ddtrace_php_version;
 
 void ddshared_minit(void);
+bool dd_rule_matches(zval *pattern, zval *prop, int rulesFormat);
+bool dd_glob_rule_matches(zval *pattern, zend_string* value);
```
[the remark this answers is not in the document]
pierotibou: I understand the will not to do additional includes, but `compat_string` seems not to be the good placeholder for this. As rule matching as nothing to see with string compatibility. 
I could move to a dedicated file if we believe it's not meant in `ddshared`, but it wouldn't solve the extra incude part of the comment

PR #2392, opened 2023-11-27
feat(profiling) add allocation and exception count to `event.json`

realFlowControl:
### Description

Alongside the pprof file we can upload an `event.json` which this PR will add and deliver the absolute numbers for:
- allocations count
- allocations len
- exceptions count

To see them you'd need to enable the `profiling_custom_data_display` feature flag, so this would most likely help us in debugging.

image

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -702,6 +702,7 @@ dependencies = [
  "perfcnt",
  "rand 0.8.5",
  "rand_distr",
+ "serde_json",
  "uuid",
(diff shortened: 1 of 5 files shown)


morrisonlevi (approved):
It doesn't block this PR, but whenever we get around to having ZTS support, we'll have to figure out how another way to do this.

PR #2395, opened 2023-11-28
Provide path params to appsec

estringana:
### Description

Extracts path parameters from incoming requests and attaches them to the span.

The implementation extracts params from the framework's route match result, which already has the parameter map populated after routing and adds them to the span. Works across Laravel, Symfony, and WordPress using each framework's native param extraction.

### Readiness checklist

### Reviewer checklist
- [ ] Milestone is set.

APPSEC-14738


appsec/src/extension/tags.c
```diff
@@ -1138,6 +1140,60 @@ static PHP_FUNCTION(datadog_appsec_track_custom_event)
...
+                           "function while appsec is disabled");
+        return;
+    }
+
+    zval *parameters = NULL;
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "z", &parameters) == FAILURE ||
+        Z_TYPE_P(parameters) != IS_ARRAY) {
+        mlog(dd_log_warning, "Unexpected parameters type. Expected array");
+        return;
+    }
```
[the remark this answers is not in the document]
estringana: I did this because we are relying on some functions coming from the frameworks. In case these functions from the frameworks change their return on new versions, this approach will avoid getting php errors.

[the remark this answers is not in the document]
cataphract: what happens if it isn't an array on the helper side?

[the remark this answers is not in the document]
estringana: `path_params` is an array so we should check here. I think the best approach is the one used here. If I get the paramater as a `HashTable`, I would have to tranform it manually to `zval` to add it to `parameters_zv` so PHP core will do that for us.

PROFeNoM (approved):
LGTM 😃 I would rebase onto master/merge master, though, before merging into master, considering the Debug Logs PR was merged and could spot errors, if any :)

appsec/src/extension/tags.c
```diff
@@ -1138,6 +1144,46 @@ static PHP_FUNCTION(datadog_appsec_track_custom_event)
     dd_tags_set_sampling_priority();
 }
 
+static PHP_FUNCTION(datadog_appsec_push_params)
```
cataphract, line 1147: this has nothing to do with tags, so shouldn't be in this file

appsec/tests/extension/push_params.phpt
```diff
@@ -0,0 +1,53 @@
+--TEST--
+Push params ara sent on request_exec
+--INI--
+extension=ddtrace.so
+datadog.appsec.log_file=/tmp/php_appsec_test.log
```
cataphract, line 5: you don't need this

appsec/tests/extension/push_params.phpt
```diff
@@ -0,0 +1,53 @@
...
+datadog.appsec.waf_timeout=42
+datadog.appsec.log_level=debug
```
cataphract, line 7: same

appsec/src/extension/tags.c
```diff
@@ -1138,6 +1144,46 @@ static PHP_FUNCTION(datadog_appsec_track_custom_event)
...
+    zend_array *parameters_arr = zend_new_array(1);
+    ZVAL_ARR(&parameters_zv, parameters_arr);
+    zval *res = zend_hash_add(
+        Z_ARRVAL(parameters_zv), _key_server_request_path_params, parameters);
+    if (res == NULL) {
+        zval_ptr_dtor(&parameters_zv);
+        mlog_g(dd_log_debug, "Parameters could not be added");
+        return;
+    }
+    Z_ADDREF_P(parameters);
```
cataphract, line 1173: why is the combination addref/ptr_dtor needed here?

appsec/src/extension/tags.c
```diff
@@ -1138,6 +1144,46 @@ static PHP_FUNCTION(datadog_appsec_track_custom_event)
     dd_tags_set_sampling_priority();
 }
 
+static PHP_FUNCTION(datadog_appsec_push_params)
+{
+    UNUSED(return_value);
+    if (DDAPPSEC_G(enabled) != ENABLED) {
```
cataphract, line 1150: needs a rebase and use `DDAPPSEC_G(active)`

cataphract (changes requested):

appsec/src/extension/tags.c
estringana: Because `parameters_zv` contains `parameters_arr` which has an alloc

appsec/src/extension/tags.c
estringana: I have moved this to `ddappsec.c`

appsec/src/extension/ddappsec.c
```diff
@@ -451,6 +457,46 @@ static PHP_FUNCTION(datadog_appsec_testing_request_exec)
     RETURN_TRUE;
 }
 
+static PHP_FUNCTION(datadog_appsec_push_params)
```
Anilm3, line 460: You should make this generic and provide functionality to pass one or even many addresses to the WAF, of arbitrary type.

appsec/src/extension/tags.c
```diff
@@ -40,6 +42,7 @@
 #define DD_TAG_HTTP_RH_CONTENT_LANGUAGE "http.response.headers.content-language"
 #define DD_TAG_HTTP_CLIENT_IP "http.client_ip"
 #define DD_TAG_USER_ID "usr.id"
+#define DD_TAG_SERVER_REQUEST_PATH_PARAMS "server.request.path_params"
```
Anilm3, line 45: I'm not entirely sure why this is here, seems to be duplicated.

appsec/src/extension/ddappsec.c
```diff
@@ -115,6 +115,9 @@ static zend_extension ddappsec_extension_entry = {
 
 ZEND_GET_MODULE(ddappsec)
 
+#define DD_TAG_SERVER_REQUEST_PATH_PARAMS "server.request.path_params"
```
Anilm3, line 118: nit: it's not a tag, but rather an address.

appsec/src/extension/ddappsec.c
```diff
@@ -451,6 +457,46 @@ static PHP_FUNCTION(datadog_appsec_testing_request_exec)
...
+    Z_ADDREF_P(parameters);
+
+    dd_conn *conn = dd_helper_mgr_cur_conn();
+    if (conn == NULL) {
+        zval_ptr_dtor(&parameters_zv);
+        mlog_g(dd_log_debug, "No connection; skipping push_params");
+        return;
+    }
+
+    dd_request_exec(conn, &parameters_zv);
```
Anilm3, line 495: `dd_request_exec` will provide you with the WAF verdict so you need to handle it and block / redirect accordingly.

Anilm3 (changes requested):
Just did a quick pass over the extension code and left a few comments.

On the other hand, note that `request_init` already sends an inferred `path_params`. Do you have any thoughts on how we could avoid sending those unnecessary `path_params` ?

Anilm3:
There should also be *appsec* integration tests .

appsec/src/extension/tags.c
estringana: Deleted

appsec/src/extension/ddappsec.c
estringana: Done

appsec/src/extension/ddappsec.c
estringana: Done

appsec/src/extension/ddappsec.c
estringana: Done

estringana:
> Just did a quick pass over the extension code and left a few comments.
> 
> On the other hand, note that `request_init` already sends an inferred `path_params`. Do you have any thoughts on how we could avoid sending those unnecessary `path_params` ?

Right now there is nothing implemented that makes this easy. The options are:

- Implement a new feature on the tracer that once a integration is used, maps the request targeted script to a framework. From that moment any request to that script we know it's a specific framework and then avoid sending generic parameters. This approach makes that the first request a target script will either send only the generic params or boths.
- Send generic params at rshutdown if no framework has sent them

Anilm3:
> > Just did a quick pass over the extension code and left a few comments.
> > On the other hand, note that `request_init` already sends an inferred `path_params`. Do you have any thoughts on how we could avoid sending those unnecessary `path_params` ?
> 
> Right now there is nothing implemented that makes this easy. The options are:
> 
> * Implement a new feature on the tracer that once a integration is used, maps the request targeted script to a framework. From that moment any request to that script we know it's a specific framework and then avoid sending generic parameters. This approach makes that the first request a target script will either send only the generic params or boths.
> * Send generic params at rshutdown if no framework has sent them

`RSHUTDOWN` is already too late for blocking on `path_params`, perhaps the first approach could work but it's a relatively large effort so perhaps it would make sense to do that in the future.

appsec/src/extension/ddappsec.c
```diff
@@ -456,6 +456,49 @@ static PHP_FUNCTION(datadog_appsec_testing_request_exec)
...
+
+    zval parameters_zv;
+    zend_array *parameters_arr = zend_new_array(1);
+    ZVAL_ARR(&parameters_zv, parameters_arr);
+    zval *res = zend_hash_add(Z_ARRVAL(parameters_zv), key, value);
+    if (res == NULL) {
+        zval_ptr_dtor(&parameters_zv);
+        mlog_g(dd_log_debug, "Parameters could not be added");
+        return;
+    }
```
cataphract, line 482: the array is brand-new, `zend_hash_add` cannot fail

appsec/src/extension/ddappsec.c
```diff
@@ -456,6 +456,49 @@ static PHP_FUNCTION(datadog_appsec_testing_request_exec)
...
+
+    dd_result result = dd_request_exec(conn, &parameters_zv);
+    zval_ptr_dtor(&parameters_zv);
+
+    if (result == dd_should_block) {
+        dd_request_abort_static_page();
+    } else if (result == dd_should_redirect) {
+        dd_request_abort_redirect();
+    }
```
cataphract, line 499: This is not correct for roadrunner. It should have the same logic as `dd_find_and_apply_verdict_for_user` (perhaps it would be a good idea to centralize it, though where it's not immediately clear).

estringana: Thanks for pointing it out. Done.
Regarding the centralising this, I would do it in a separate PR

cataphract (approved):

Anilm3 (approved):

PR #2397, opened 2023-11-29
fix(profiler) prevent emitting empty profiles

realFlowControl:
### Description

This will make sure that empty profiles that we might send to the `ddprof_time` thread will be ignored. This should not happen anyway, but it happened in the past.

PROF-8749

### Readiness checklist

### Reviewer checklist
- [x] Milestone is set.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -1046,38 +1046,38 @@ impl Profiler {
             let len = 2 + locals.profiling_experimental_cpu_time_enabled as usize;
             sample_types.extend_from_slice(&SAMPLE_TYPES[0..len]);
             sample_values.extend_from_slice(&values[0..len]);
-        }
 
-        // alloc-samples, alloc-size
-        if locals.profiling_allocation_enabled {
-            sample_types.extend_from_slice(&SAMPLE_TYPES[3..5]);
-            sample_values.extend_from_slice(&values[3..5]);
-        }
+            // alloc-samples, alloc-size
+            if locals.profiling_allocation_enabled {
+                sample_types.extend_from_slice(&SAMPLE_TYPES[3..5]);
+                sample_values.extend_from_slice(&values[3..5]);
+            }
 
-        #[cfg(feature = "timeline")]
-        if locals.profiling_experimental_timeline_enabled {
-            sample_types.push(SAMPLE_TYPES[5]);
-            sample_values.push(values[5]);
-        }
+            #[cfg(feature = "timeline")]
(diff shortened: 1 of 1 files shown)


profiling/src/profiling/mod.rs
```diff
@@ -572,6 +572,12 @@ impl Profiler {
     }
 
     pub fn send_sample(&self, message: SampleMessage) -> Result<(), TrySendError<ProfilerMessage>> {
+        if message.key.sample_types.len() == 0 {
+            // profiling disabled, this should not happen!
+            warn!("You spot a bug in the profiler, please be so kind and report this do Datadog.");
+            // this return is technically not correct :-(
+            return Err(TrySendError::Disconnected(ProfilerMessage::Sample(message)));
+        }
```
realFlowControl, line 580: @morrisonlevi I am not really happy with having this check for every sample that we send (also the return value is incorrect). OTOH this is the perfect spot, as all samples are going through this. If you have some time, could we put some thought into this? One idea is to check in the other thread, so we move this out of the hot-path

realFlowControl: I moved the logic to the other thread

profiling/src/profiling/mod.rs
```diff
@@ -346,6 +346,12 @@ impl TimeCollector {
         profiles: &mut HashMap<ProfileIndex, InternalProfile>,
         started_at: &WallTime,
     ) {
+        if message.key.sample_types.len() == 0 {
+            // profiling disabled, this should not happen!
+            warn!("You spot a bug in the profiler, please be so kind and report this do Datadog.");
```
morrisonlevi, line 351: We should improve this message if we are going to emit it. Maybe...

```suggestion
            warn!("A sample with no sample types was recorded in the profiler. Please report this to Datadog.");
```

profiling/src/profiling/mod.rs
```diff
@@ -1034,43 +1040,44 @@ impl Profiler {
...
             sample_values.extend_from_slice(&values[0..len]);
+        }
 
-            // alloc-samples, alloc-size
-            if locals.profiling_allocation_enabled {
-                sample_types.extend_from_slice(&SAMPLE_TYPES[3..5]);
-                sample_values.extend_from_slice(&values[3..5]);
-            }
+        // alloc-samples, alloc-size
+        if locals.profiling_allocation_enabled {
```
morrisonlevi, line 1052: If profiling isn't enabled, then none of these following ifs need to be taken. I'm pretty sure that's why it was done the way it was before.

morrisonlevi (approved):

PR #2401, opened 2023-12-01
Reduce noise in composer package download

### LINKED ISSUE #2181: Lot of entries are missing in the .gitattributes files
Hi @bwoebi, @morrisonlevi, @PROFeNoM, @labbati

The following file avoid to download non-necessary files/folder when installing the library with composer
https://github.com/DataDog/dd-trace-php/blob/master/.gitattributes

Currently, I get
image

I assume a lot of them could be added to the gitattributes file too:
- .github/.gitlab folders and github-action-helpers
- docs folder
- example folder
- all `.*` files
- all `.md` files
- docker-composer.yml, phpstan.neon, Makefile, etc
- certainly some folder which I don't know what they are useful for (I currently only use the src one)

Would it be possible to add more entries to the gitattributes file ?
I'm ok doing the PR if needed, but I don't know which folder/files can be also added (other than the one I listed)

pierotibou:
### Description

Update `.gitattributes` file to reduce the number of files and folder that are packaged when using composer.

Fixes #2181

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,14 @@ RELEASING.md        export-ignore
 .circleci           export-ignore
 .clang-format       export-ignore
 ext/vendor/*        linguist-generated=true
+
+.github           export-ignore
+.gitlab           export-ignore
+.nuke             export-ignore
+benchmark         export-ignore
+docs              export-ignore
+examples          export-ignore
+github-actions-helpers         export-ignore
+
+.gitlab-ci.yml    export-ignore
+phpstan*.neon     export-ignore


realFlowControl:
AFAIK this only affects composer, so we might also `export-ignore` the `profiler` directory and maybe even the `ext` directory? WDYT?

pierotibou:
> WDYT?

I'm not entirely sure that's why I asked for 👀 . If we have any doubt I'd rather stay on the conservative side of things.

bwoebi:
I would also add tests there.

bwoebi (approved):

PR #2402, opened 2023-12-04
Implement http.route for Slim

pablomartinezbernardo:
### Description

Adds `http.route` to Slim integration

### Reviewer checklist
- [ ] Test coverage seems ok.


PROFeNoM (approved):
LGTM :smiley: Sorry for the delay :+1:

src/Integrations/Integrations/Slim/SlimIntegration.php
```diff
@@ -91,15 +91,29 @@ function ($errorMiddleware, $self, $args) use ($rootSpan, $integration) {
                         'lookupRoute',
                         null,
                         function ($router, $scope, $args, $return) use ($rootSpan) {
+                            /** @var \Slim\Interfaces\RouteInterface $route */
+                            $route = $return;
```
estringana, line 95: Why do you need the variable `route` instead of using `return` directly?

pablomartinezbernardo: Copied it from what was there before, my guess is the performance penalty is negligible and this way we can annotate the variable for the IDE to autocomplete. Let me know if you'd rather change it!

bwoebi: Just having `/** @var \Slim\Interfaces\RouteInterface $return */` should work too.

bwoebi: @pablomartinezbernardo I think you now have put it _outside_ the function. it needs to stay where it was to be in the right scope.

pablomartinezbernardo: Tested it before pushing and it works like this as well, but you're right it might look clearer inside 👍

PR #2405, opened 2023-12-04
fix(ci) fix xdebug

realFlowControl:
### Description

This will fix the failing xdebug tests with the new containers and also add

- `sapi/cli/tests/bug80092.phpt` and
- `Zend/tests/gh12073.phpt`

to the xfail list.  `bug80092.phpt` is already blacklisted for PHP 8.3 and might be unlisted with 8.3.1. `gh12073` sets the memory limit to 2MB which is not possible anymore when the `dd_wrap_autoloader.php' is included.

`gh12073` fails with:
```
========DIFF========
001+ Warning: Failed to set memory limit to 2097152 bytes (Current memory usage is 8388608 bytes) in /usr/local/src/php/Zend/tests/gh12073.php on line 10
001- Fatal error: Allowed memory size of %d bytes exhausted%s(tried to allocate %d bytes) in %s on line %d
```

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -4276,7 +4276,7 @@ workflows:
           requires: [ 'Prepare Code' ]
           name: "PHP 83 Xdebug tests"
           docker_image: "datadog/dd-trace-ci:php-8.3_buster"
-          xdebug_version_one: "3.3.0alpha2"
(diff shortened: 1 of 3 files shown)


Anilm3 (approved):
lgtm

PR #2407, opened 2023-12-05
[WIP - Experimental] feat: Add Span End Callback

PROFeNoM:
### Description

**Experimental**

Leverage span end callback to end the OTel Span (if any) when a span is closed using the DD Api. This affects both OTel-created & remapped OTel spans. The goal is to fix edge cases + properly handle trace ids (See Tests Cases).

_Sample Edge Case_:
> 1. Start a span using DD API
> 2. Retrieve the current span using the OTel API
> 3. Add a distributed tag, which subsequently changes the tracestate 
> 4. Close the span using the DD API
> 5. Add a distributed tag, which subsequently changes the tracestate
> 6. Create a read-only copy of the span and read its tracestate

At present, there are issues with the third and fifth steps where the trace states are not being updated correctly in accordance with the OTel-API. Once these changes are made, the trace states can be dynamically retrieved and distributed context changes will accurately be reflected in the OTel API.

TODO:
- [X] Fix crashing callback scenario
- [ ] Add intertwined, start trace interoperability tests

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/OpenTelemetry/SpanContext.php
+++ b/src/DDTrace/OpenTelemetry/SpanContext.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace DDTrace\OpenTelemetry\API\Trace;
 
 use DDTrace\OpenTelemetry\SDK\Trace\Span;
+use DDTrace\RootSpanData;
 use DDTrace\SpanData;
 use OpenTelemetry\API\Trace as API;
 use OpenTelemetry\API\Trace\SpanContextInterface;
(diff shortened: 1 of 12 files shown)


src/DDTrace/OpenTelemetry/SpanContext.php
```diff
@@ -23,35 +24,31 @@ final class SpanContext implements SpanContextInterface
...
     private ?string $spanId;
 
-    private bool $isValid = true;
+    private static function getRootSpan(SpanData $span): RootSpanData
+    {
+        while (!($span instanceof RootSpanData) && $span->parent) {
+            $span = $span->parent;
+        }
+        return $span;
+    }
```
PROFeNoM, line 35: Note to self: May be wrong when considering intertwined `start_trace_span` & OTel starts w/ `setParent(false)`

PROFeNoM:
-

PR #2409, opened 2023-12-05
Fix analytics.event override

PROFeNoM:
### Description

A system-test testing the behavior of `analytics.event` tag override was added with Datadog/system-tests#1818. The current implementation would fail :)

This PR simply explicitly checks for 'true' (not case sensitive) to return `1.0`, and _anything else_ would yield `0.0`

I set it as  `do-not-merge` as I need the pipeline to finish to test it properly, but the local tests otherwise pass.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/serializer.c
+++ b/ext/serializer.c
@@ -1151,14 +1151,14 @@ void ddtrace_shutdown_span_sampling_limiter(void) {
 }
 
 // ParseBool returns the boolean value represented by the string.
-// It accepts 1, t, T, True (case insensitive), 0, f, F, False (case insensitive).
-// Any other value returns 0
+// It accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False.
+// Any other value returns -1.
 static zend_always_inline double strconv_parse_bool(zend_string *str) {
     // See Go's strconv.ParseBool
     // https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/strconv/atob.go;drc=1f137052e4a20dbd302f947b1cf34cdf4b427d65;l=10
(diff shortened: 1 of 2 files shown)


PROFeNoM:
I confirm that the system tests would pass with this change (with respect to the [alex/otel-api](https://github.com/DataDog/system-tests/compare/alex/otel-api) branch)
image

ext/serializer.c
```diff
@@ -1202,6 +1202,11 @@ void ddtrace_shutdown_span_sampling_limiter(void) {
     zend_hash_destroy(&dd_span_sampling_limiters);
 }
 
+static zend_always_inline double ddtrace_convert_string_to_double(zend_string *str) {
```
pierotibou, line 1205: There's no possibility for that use case to be something else than `true`? I ask because [we do something more comprehensive in .NET](https://github.com/DataDog/dd-trace-dotnet/blob/dbee006bb81c3a5f724d39f9265ac754242f17fe/tracer/src/Datadog.Trace/Activity/OtlpHelpers.cs#L429)

PROFeNoM: Indeed. I re-read the RFC, and it must match the same behavior as [strconv.ParseBool](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/strconv/atob.go;l=10). I'll change this as soon as I can.

Note: It's not tested in the system tests right now. I may add them.

pierotibou: Thanks a lot for the thorough follow up!

pierotibou (approved):
Thank you!

PR #2413, opened 2023-12-07
Fix dddbs service mapping

PROFeNoM:
### Description

SDBM-729

Service mapping wasn't applied to the `dddbs` tag while it should.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/Integrations/Integrations/DatabaseIntegrationHelper.php
+++ b/src/Integrations/Integrations/DatabaseIntegrationHelper.php
@@ -63,10 +63,6 @@ class DatabaseIntegrationHelper
         $tags = [];
 
         if ($databaseService != "") {
-            $mapping = dd_trace_env_config('DD_SERVICE_MAPPING');
-            if (isset($mapping[$databaseService])) {
-                $databaseService = $mapping[$databaseService];
-            }
             $tags["dddbs"] = $databaseService;
         }
 
--- a/src/Integrations/Integrations/DatabaseIntegrationHelper.php
+++ b/src/Integrations/Integrations/DatabaseIntegrationHelper.php
@@ -63,6 +63,10 @@ class DatabaseIntegrationHelper
         $tags = [];
 
         if ($databaseService != "") {
+            $mapping = dd_trace_env_config('DD_SERVICE_MAPPING');
+            if (isset($mapping[$databaseService])) {
+                $databaseService = $mapping[$databaseService];
+            }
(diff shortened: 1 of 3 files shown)


src/Integrations/Integrations/DatabaseIntegrationHelper.php
```diff
@@ -63,6 +63,10 @@ public static function propagateViaSqlComments($query, $databaseService, $mode =
         $tags = [];
 
         if ($databaseService != "") {
+            $mapping = dd_trace_env_config('DD_SERVICE_MAPPING');
```
pierotibou, line 66: So I think it works but I find it weird that we need to do this here.
Line 47, we take either the `peer.service` or span service. 
IIUC, the peer.service is already remapped.
I assume the service isn't as we're doing this here. It feels weird we have to do the remapping here. 
Shouldn't the span service be the remapped one already?
If not, shouldn't we move this remapping to the span object so that we don't have to duplicate this logic here?

tests/Integration/DatabaseMonitoringTest.php
```diff
@@ -57,6 +58,67 @@ public function testInjection()
...
+    {
+        try {
+            $hook = \DDTrace\install_hook(self::class . "::instrumented", function (HookData $hook) {
+                $span = $hook->span();
+                Integration::handleInternalSpanServiceName($span, "pdo");
+                $span->name = "instrumented";
+                DatabaseIntegrationHelper::injectDatabaseIntegrationData($hook, 'mysql', 1);
+            });
+            self::putEnv("DD_TRACE_DEBUG_PRNG_SEED=42");
+            self::putEnv("DD_DBM_PROPAGATION_MODE=full");
```
pierotibou, line 71: Would it be useful to have a test with partial as well? As we do have the service in partial mode IIRC.

pierotibou (approved):
LGTM. One suggestion for improving the test coverage though

tests/Integration/DatabaseMonitoringTest.php
PROFeNoM: Added the tests for `none` & `service`; I couldn't find [documentation](https://datadoghq.atlassian.net/wiki/spaces/TS/pages/3183772859/Troubleshooting+DBM+and+APM+Correlation) (or [public doc](https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/?tab=go)) for `service`, though

PR #2422, opened 2023-12-13
fix: Allow hooks to be removed after class binding

PROFeNoM:
### Description

During the micro benchmarking process, I came across an unexpected behavior related to #2367. The goal of the latter was most notably to introduce minimal overhead when injecting logs while instrumenting `Psr\Log\NullLogger`. However, the results showed otherwise (refer to the Before/After comparison below).

The problem arises when the call to `\DDTrace\remove_hook` is made after a class bind (`zen_compile.c:do_bind_class`). As a result, the observers-related code (`zend_observer.c: _zend_observer_class_linked_notify`) does not get executed. This implies that `hook.c:zai_hook_resolve_class` and `hook.c:zai_hook_is_excluded` are not called either. Thus, when the supposedly removed class is called again, there is no check to see if the hook was removed or not and we directly go to `zend_observer.c:_zend_observe_fcall_begin` & `hook.c:zai_hook_continue`. _This pull request addresses this issue by adding the necessary check._

From a general perspective, it was only effective to remove a hook before a class binding occurs. Attempting to remove a hook at any other point was ineffective. 

While microbenchmarks are still being developed, the before/after comparison below shows the effect of logs injection (Injection) versus no logs injection (Baseline):
**Before:**
image

**After:**
image

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/ext/sandbox/install_hook/exclude_inherited_hook_after_binding.phpt
+++ b/tests/ext/sandbox/install_hook/exclude_inherited_hook_after_binding.phpt
@@ -1,0 +1,28 @@
+--TEST--
+remove_hook() with class argument
+--FILE--
+<?php
+
+interface Elder {
+    function foo();
+}
+
+$id = DDTrace\install_hook("Elder::foo", function () { print "HOOKED: " . static::class . "\n"; });
+
+if (time()) {
+    abstract class Child implements Elder {}
+
+    class GrandChild extends Child {
+        function foo() {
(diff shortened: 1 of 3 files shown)


realFlowControl (approved):
The tests are looking good and are ✅ 
@morrisonlevi knows more about the tracer, so if it is okay, let's wait for his review

bwoebi:
Superseded by #2449.

PR #2428, opened 2023-12-18
Parses schema extraction remote config configuration

estringana:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/helper/remote_config/client_handler.cpp
+++ b/appsec/src/helper/remote_config/client_handler.cpp
@@ -47,10 +47,11 @@ client_handler::ptr client_handler::from_settings(service_identifier &&id,
     }
 
     std::vector<remote_config::listener_base::shared_ptr> listeners = {};
-    if (dynamic_enablement) {
+    if (dynamic_enablement || eng_settings.schema_extraction.enabled) {
         listeners.emplace_back(
             std::make_shared<remote_config::asm_features_listener>(
-                service_config));
+                service_config, dynamic_enablement,
+                eng_settings.schema_extraction.enabled));
     }
 
     if (eng_settings.rules_file.empty()) {
--- a/appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
+++ b/appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
@@ -10,21 +10,24 @@
 #include "utils.hpp"
 #include <algorithm>
 
-double dds::remote_config::asm_features_listener::parse_api_security(
+void dds::remote_config::asm_features_listener::parse_api_security(
     const rapidjson::Document &serialized_doc)
 {
-    auto api_security_itr = json_helper::get_field_of_type(
-        serialized_doc, "api_security", rapidjson::kObjectType);
+    rapidjson::Value::ConstMemberIterator const api_security_itr =
+        serialized_doc.FindMember("api_security");
 
-    if (!api_security_itr) {
+    if (api_security_itr == serialized_doc.MemberEnd()) {
+        return;
+    }
+
+    if (rapidjson::kObjectType != api_security_itr->value.GetType()) {
         throw error_applying_config("Invalid config json encoded contents: "
-                                    "api_security key missing or invalid");
+                                    "api_security key invalid");
     }
 
     auto request_sample_rate_itr =
-        api_security_itr.value()->value.FindMember("request_sample_rate");
-    if (request_sample_rate_itr ==
-        api_security_itr.value()->value.MemberEnd()) {
+        api_security_itr->value.FindMember("request_sample_rate");
+    if (request_sample_rate_itr == api_security_itr->value.MemberEnd()) {
         throw error_applying_config("Invalid config json encoded contents: "
                                     "request_sample_rate key missing");
     }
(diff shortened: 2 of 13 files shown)


appsec/src/helper/remote_config/client_handler.cpp
```diff
@@ -47,10 +47,11 @@ client_handler::ptr client_handler::from_settings(service_identifier &&id,
     }
 
     std::vector<remote_config::listener_base::shared_ptr> listeners = {};
-    if (dynamic_enablement) {
+    if (dynamic_enablement || eng_settings.schema_extraction.enabled) {
```
Anilm3, line 50: Perhaps we could add the listener in all cases, the method `client::compute_client_status()` already takes into account whether the RC value should be used or not.

appsec/src/helper/sampler.hpp
```diff
@@ -64,10 +59,20 @@ class sampler {
 
     std::optional<scope> get()
     {
+        if (service_config_->get_asm_enabled_status() ==
```
Anilm3, line 62: Why do you need this at all? When is the sampler being called with ASM disabled?

appsec/src/helper/sampler.hpp
```diff
@@ -64,10 +59,20 @@ class sampler {
...
+            return std::nullopt;
+        }
+
         const std::lock_guard<std::mutex> lock_guard(mtx_);
 
         std::optional<scope> result = std::nullopt;
 
+        if (sample_rate_ !=
+            valid_sample_rate(service_config_->get_request_sample_rate())) {
+            set_sampler_rate(service_config_->get_request_sample_rate());
```
Anilm3, line 73: Setting the rate disregards the current number of requests, it should be possible to avoid discarding the current state entirely.

appsec/src/helper/sampler.hpp
```diff
@@ -82,10 +87,38 @@ class sampler {
         return result;
     }
 
+    static double valid_sample_rate(double sampler_rate)
+    {
+        // NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+        if (sampler_rate <= 0) {
```
Anilm3, line 93: This is okay, I suppose you could also check that the status is valid in the listener.

appsec/src/helper/sampler.hpp
```diff
@@ -82,10 +87,38 @@ class sampler {
...
+            sampler_rate = 1;
+        } else if (sampler_rate < min_rate) {
+            sampler_rate = min_rate;
+        }
+        // NOLINTEND(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+
+        return sampler_rate;
+    }
+
+    void set_sampler_rate(double sampler_rate)
```
Anilm3, line 105: This should be protected, even more so when it should be used behind a mutex.

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -36,18 +67,31 @@ void dds::remote_config::asm_features_listener::on_update(const config &config)
...
+    }
+
+    if (dynamic_enablement_) {
+        parse_asm(serialized_doc);
+    }
+
+    if (api_security_enabled_) {
+        parse_api_security(serialized_doc);
+    }
+}
```
Anilm3, line 97: I think it would make more sense to evaluate all configurations and then apply at the end, I would also change the parser to ensure that if, in the future, we add more overlapping configurations (e.g. a configuration with the `asm` key but not containing `enabled`) that existing libraries don't suddenly report errors.

appsec/src/helper/sampler.hpp
estringana: I actually did this on purpose. When a new rate is set, then its stats have to change or the rates wont match was neither was set or it is set

cataphract (changes requested):

appsec/src/helper/remote_config/client_handler.cpp
estringana: I rather leave it like this and not populate those products we know we are not going to use

appsec/src/helper/sampler.hpp
estringana: I don't think it's a massive issue but I can take it out

appsec/src/helper/sampler.hpp
estringana: Done

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
estringana: I would leave it as it is. Right now this listener is in charge of two different and independent things. The methods are actually doing what you say, first parse and then apply.
Regarding relaxing errors, I could but that would be risky. As of now, when this configuration comes with `asm` key, it needs to be an object and it needs to have the `enable` key. If we can't rely on that, we would enter on a paranoiac  approach. Also as of know if configurations don't get apply, we won't know as there are no errors.

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -9,25 +9,56 @@
 #include "remote_config/exception.hpp"
 #include "utils.hpp"
 #include <algorithm>
-#include <rapidjson/document.h>
 
-void dds::remote_config::asm_features_listener::on_update(const config &config)
+void dds::remote_config::asm_features_listener::parse_api_security(
+    const rapidjson::Document &serialized_doc)
```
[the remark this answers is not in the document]
estringana: Can't say unserialised fits better here, sorry :(

appsec/src/helper/remote_config/listeners/asm_features_listener.hpp
```diff
@@ -25,14 +29,32 @@ class asm_features_listener : public listener_base {
...
     }
 
     void init() override {}
     void commit() override {}
 
 protected:
+    static constexpr std::string_view asm_features = "ASM_FEATURES";
+    void parse_asm(const rapidjson::Document &serialized_doc);
+    void parse_api_security(const rapidjson::Document &serialized_doc);
     std::shared_ptr<service_config> service_config_;
```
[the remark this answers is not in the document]
estringana: I may be missing something. Service config is used in different places of the system. For example, here its values are set while on the client for example their values are read. It's because of that shared core responsability that it is shared

appsec/src/helper/sampler.hpp
```diff
@@ -19,96 +19,35 @@ namespace dds {
...
-            return;
+        bool result = false;
+        if (floor(request_ * sample_rate_) !=
+            floor((request_ + 1) * sample_rate_)) {
+            result = true;
         }
 
-        request_ = 0;
-        sample_rate_ = sampler_rate;
+        request_.fetch_add(1, std::memory_order_relaxed);
```
cataphract, line 43: you're doing three fetches of request_; each can yield different values. Instead, you should use the result of `fetch_add`

appsec/src/helper/service.hpp
```diff
@@ -75,6 +75,11 @@ class service {
         return schema_sampler_;
     }
 
+    void set_sampler_rate(double rate)
+    {
+        schema_sampler_ = std::make_shared<sampler>(rate);
```
cataphract, line 80: Because now `schema_sampler_` can be written to from different threads, you need to have a [`std::atomic<std::shared_ptr<sampler>>`](https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2) (for C++20).

estringana: Done

cataphract (approved):

Anilm3 (changes requested):
Sorry I should've mentioned this sooner but the RFC related to sampling was deprecated in favour of a new one. So the sampling isn't provided through remote configuration any longer.

I'm not sure if there's anything from this PR you'd like to keep but otherwise it can't be merged.

PR #2432, opened 2023-12-20
test(profiling): with address sanitizer on bookworm with GH actions

morrisonlevi:
### Description

This adds bookworm images which were heavily based on the buster image:

- Many dependencies are newer and did not have to be built from scratch, which is nice.
- This doesn't run all the tests on bookworm, just builds and uploads images.
- However, only some images can be built. More work is required for PHP 7.3 and below to get working. I'm not sure how much more work is needed, but the two known issues are with `intl` and I think PHP 7.0 requires OpenSSL v1. Additionally, the shared extension versions have failures, something about `ldap_connect` having a mismatched signature, possibly: https://github.com/curl/curl/issues/7004.

The key reason this PR was made is to add a PHP 8.3 NTS ASAN build, and then runs the `profiling/tests/phpt` tests with it. This is why I didn't continue working on the bookworm images so that all versions work; I really just need ASAN right now, and it wasn't working on Buster due to libasan5 vs libasan6 differences.

PROF-8286

### Reviewer checklist
- [x] Test coverage seems ok.


dockerfiles/ci/bookworm/.env
```diff
@@ -0,0 +1,2 @@
+BOOKWORM_CURRENT_VERSION=1
+BOOKWORM_NEXT_VERSION=2
```
bwoebi, line 2: I do not understand the point of this file?

dockerfiles/ci/bookworm/Dockerfile
```diff
@@ -33,24 +35,23 @@ ENV DEVLIBS \
     libxslt1-dev \
     libzip-dev \
     zlib1g-dev \
-    libasan5 \
+    libasan6 \
     gnupg \
     unixodbc-dev \
-    unixodbc
+    unixodbc \
+    valgrind
```
bwoebi, line 42: Can we please add lsof? It's the one thing I'm installing all the time.

bwoebi:
I'm not sure whether parts of this are already pushed, but https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/13970/workflows/b0aaa43d-7966-4958-a4c2-416367afda4c/jobs/3410571 looks somewhat related I suppose?

dockerfiles/ci/bookworm/.env
morrisonlevi: The current buster builds are "unversioned." This can cause issues (like the one you mention below with nts-asan failing). This file is a way to version the build, although it's not plumbed all the way through everything like the CI and such. We do this in the libdatadog images and it's worked nicely there.

morrisonlevi:
The push was from a different but related PR where I tried to do this on buster images. I have re-pushed images from that were built from master.

morrisonlevi:
I've added lsof and rebuild the relevant images. Once this is reviewed, I'll actually add back all the buster image sources which were deleted in this PR. Doing it in this way makes the diff much easier to review.

.github/workflows/prof_asan.yml
```diff
@@ -0,0 +1,58 @@
...
+          export CARGO_TARGET_DIR=/tmp/build-cargo
+          export CC=clang-16
+          export CFLAGS='-fsanitize=address  -fno-omit-frame-pointer'
+          export LDFLAGS='-fsanitize=address'
+          export RUSTC_LINKER=lld-16
+          triplet=$(uname -m)-unknown-linux-gnu
+          RUSTFLAGS='-Zsanitizer=address' cargo +nightly-2023-05-03 build -Zbuild-std --target $triplet --release
+          cd tests
+          cp -v $(php-config --prefix)/lib/php/build/run-tests.php .
+          php run-tests.php --asan -d extension=$CARGO_TARGET_DIR/$triplet/release/libdatadog_php_profiling.so phpt
```
realFlowControl, line 46: This is just a minor note and not blocking: we could extract this into a "Run tests" step after the "Cache build dependencies" step, so that build dependencies will be cached even in case the actual tests fail.

realFlowControl (approved):

.github/workflows/prof_asan.yml
morrisonlevi: Good point 👍🏻 .

realFlowControl:
The `Cache build dependencies` step did not cache anything due to:
```
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
Warning: Cache save failed.
```
See https://github.com/DataDog/dd-trace-php/actions/runs/7468310960/job/20347970685?pr=2432
I fixed this in https://github.com/DataDog/dd-trace-php/pull/2432/commits/e34c5f640defb1b01932bf41137cd4010fda0af1

PR #2434, opened 2023-12-22
feat(profiler) add exception message to exception sample

realFlowControl:
### Description

This will add the exception message to the sample that is taken. You'd need to opt-in to the collection of exception messages by setting the environment variable `DD_PROFILING_EXCEPTION_MESSAGE_ENABLED` or the INI setting `datadog.profiling.exception_message_enabled` to `true`/`1`/`on`.

Additionally I moved reading the exception class name (and the message) to after the sampling decision which lowers the overhead of exception profiling.

PROF-8838

### Reviewer checklist

- [x] Backend and UI support is implemented
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/build.rs
+++ b/profiling/build.rs
@@ -39,6 +39,7 @@ fn main() {
         run_time_cache,
         fibers,
         trigger_time_sample,
+        vernum,
     );
 
     cfg_php_major_version(vernum);
@@ -89,11 +90,18 @@ fn build_zend_php_ffis(
     run_time_cache: bool,
     fibers: bool,
     trigger_time_sample: bool,
+    vernum: u64,
 ) {
     println!("cargo:rerun-if-changed=src/php_ffi.h");
     println!("cargo:rerun-if-changed=src/php_ffi.c");
     println!("cargo:rerun-if-changed=../ext/handlers_api.c");
     println!("cargo:rerun-if-changed=../ext/handlers_api.h");
+
+    let sandbox = if vernum < 80000 {
+        "../zend_abstract_interface/sandbox/php7/sandbox.c"
+    } else {
+        "../zend_abstract_interface/sandbox/php8/sandbox.c"
+    };
 
     // Profiling only needs config and its dependencies.
     let zai_c_files = [
@@ -104,6 +112,7 @@ fn build_zend_php_ffis(
         "../zend_abstract_interface/env/env.c",
         "../zend_abstract_interface/exceptions/exceptions.c",
         "../zend_abstract_interface/symbols/lookup.c",
+        sandbox,
         "../zend_abstract_interface/json/json.c",
         "../zend_abstract_interface/zai_string/string.c",
     ];
--- a/profiling/build.rs
+++ b/profiling/build.rs
@@ -78,6 +78,7 @@ const ZAI_H_FILES: &[&str] = &[
     "../zend_abstract_interface/config/config_decode.h",
     "../zend_abstract_interface/config/config_ini.h",
     "../zend_abstract_interface/env/env.h",
+    "../zend_abstract_interface/exceptions/exceptions.h",
     "../zend_abstract_interface/json/json.h",
 ];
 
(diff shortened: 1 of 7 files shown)


profiling/build.rs
```diff
@@ -85,19 +87,29 @@ fn build_zend_php_ffis(
...
     println!("cargo:rerun-if-changed=../ext/handlers_api.c");
     println!("cargo:rerun-if-changed=../ext/handlers_api.h");
 
+    let sandbox = if vernum < 80000 {
+        "../zend_abstract_interface/sandbox/php7/sandbox.c"
+    } else {
+        "../zend_abstract_interface/sandbox/php8/sandbox.c"
+    };
+
     // Profiling only needs config and its dependencies.
```
morrisonlevi, line 103: This comment is out-of-date now.

profiling/src/exception.rs
```diff
@@ -45,12 +45,31 @@ impl ExceptionProfilingStats {
...
+
+        #[cfg(php7)]
+        let message = unsafe {
+            zend::zai_str_from_zstr(zend::zai_exception_message((*exception).value.obj).as_mut())
+                .into_string()
+        };
+        #[cfg(php8)]
+        let message = unsafe {
+            zend::zai_str_from_zstr(zend::zai_exception_message(exception).as_mut()).into_string()
+        };
```
morrisonlevi, line 71: Since the duplicated expression isn't simple, I'd prefer to deduplicate it like this. What do you think?

```suggestion
        let message = unsafe {
            #[cfg(php7)]
            let exception_obj = (*exception).value.obj;
            #[cfg(php8)]
            let exception_obj = exception;
            zend::zai_str_from_zstr(zend::zai_exception_message(exception_obj).as_mut())
                .into_string()
        };
```

profiling/src/php_ffi.h
```diff
@@ -18,6 +18,8 @@
 // Profiling needs ZAI config for INI support.
 #include <config/config.h>
 
+#include <exceptions/exceptions.h>
```
morrisonlevi, line 21: Can you leave a note about why this header is here like the other lines around it?

profiling/src/profiling/mod.rs
```diff
@@ -750,6 +751,12 @@ impl Profiler {
                     key: "exception type",
                     value: LabelValue::Str(exception.clone().into()),
                 });
+
+                labels.push(Label {
+                    key: "exception message",
+                    value: LabelValue::Str(message.clone().into()),
```
morrisonlevi, line 757: Is the clone necessary? `message` was added by value to the function signature, so it's owned already and I don't see any other uses of it in the function.

profiling/src/exception.rs
realFlowControl: Looks good!

profiling/src/profiling/mod.rs
realFlowControl: You see it right, the `clone()` is not needed, I removed it. Thanks for looking!

morrisonlevi (commented):
Are there any concerns about gathering the exception message? Any filtering or anything that needs done? I wonder if APM does anything.

realFlowControl:
I need to add a backend and ui PR to show them, for .NET you can see how this will be surfaced:
image

realFlowControl:
@morrisonlevi I added an option to disable the collection of messages for exceptions. It is default enabled

realFlowControl:
@morrisonlevi I am btw. open to set this default disabled

profiling/src/exception.rs
```diff
@@ -116,14 +148,9 @@ unsafe extern "C" fn exception_profiling_throw_exception_hook(
     // traversed. Fortunately, this behavior can be easily identified by checking for a NULL
     // pointer.
     if exception_profiling && !exception.is_null() {
-        #[cfg(php7)]
-        let exception_name = (*(*exception).value.obj).class_name();
-        #[cfg(php8)]
-        let exception_name = (*exception).class_name();
```
bwoebi, line 122: I would propose using
```
        #[cfg(php7)]
        let exception = (*exception).value.obj;
```
here and completely get rid of the #[cfg] switches in track_exception

realFlowControl: Nice idea, I could cleanup nicely. Thanks 🙇

bwoebi (approved):
Looks good to me now :-)

PR #2435, opened 2023-12-27
Add basic PHP MicroBenchmarks

PROFeNoM:
### Description

Add php microbenchmarks using [phpbench](https://github.com/phpbench/phpbench).

The current microbenchmarks are currently only for the profiler. However, they will soon be divided into separate scenarios: `profiler` and `tracer`. The profiler microbenchmarks will remain virtually unchanged. On the other hand, the tracer microbenchmarks will be triggered when any changes occur in the `bridge`, `ext`, `src`, or `zend_abstract_interface` directories. Once the tracer microbenchmarks are completed, a pull request comment containing the run results will be generated.

Currently, the following benchmark suites are being executed and separated into two categories: **API** and **Integrations**.

_**API**_
- **ContextPropagation**: Evaluate the performances of `consume_distributed_tracing_headers` and `generate_distributed_tracing_headers` with both 64 and 128-bit trace identifiers
- **Hook**: Evaluate the overhead introduced by the installation of a hook on a dummy function call
- **MessagePackSerialization**: Evaluate the performance of message pack serialization, which is part of the hot path
- **SamplingRuleMatching**: Evaluate & compare the performances of Glob & Regex sampling rule matching
- **Span**: Evaluate & compare the performances of the Datadog API, OTel API, and Datadog+OTel API (in the "worst case scenario") on a basic, "inverted pyramid" trace
- **TraceAnnotations**: Same as the **Hook** suite, but using a trace annotation to set up the hook. It should be the same.
- **TraceFlush**: Evaluate the flushing performances (span + message pack serialization + coms writer flush), which is part of the hot path
- **TraceSerialization**: Evaluate the performances of span serialization, which is part of the hot path

**_Integrations_**

Evaluate & compare the performance and overhead introduced by the tracer on: 
- [Web] **Laravel**, **Symfony**, and **WordPress**
- [Databases] **PDO** (+ DBM) and **PHPRedis**
- [Logging] **Monolog** (+ NullLogger)

This PR is part of a sequence of PRs whose orders matter:
- [X] DataDog/relenv-benchmark-analyzer-core#6
- [X] DataDog/relenv-benchmark-analyzer#29
- [X] Datadog/benchmarking-platform#39
- [ ] _(this)_ Datadog/dd-trace-php#2435
- [ ] (First, let's focus on the above) Datadog/dd-trace-php#2438

Although meant to be analyzed using the benchmarking platform, a non-aggregated console report is also generated. Here is a sample report for the `PDOBench` benchmark suite (`make benchmarks FILTER=PDOBench`):
image

### Reviewer checklist
- [ ] Test coverage seems ok.


tests/Benchmarks/README.md
```diff
@@ -0,0 +1,45 @@
+# Benchmarks
```
pierotibou, line 1: 🙇

tests/Benchmarks/API/HookBench.php
```diff
@@ -0,0 +1,45 @@
...
+    public function benchWithoutHook()
+    {
+        for ($i = 0; $i < 1000; $i++) {
+            $this->dummyFunction();
+        }
+    }
+
+    public function setUp()
+    {
+        \DDTrace\trace_method('Benchmarks\API\HookBench', 'dummyFunction', function () { });
```
pierotibou, line 24: Are there differences between `trace_method`, `trace_function` and `install_hook`? Should we test all 3?

pierotibou (approved):
Looks quite cool, thanks a lot for doing it.

tests/Benchmarks/API/HookBench.php
PROFeNoM: Since `trace_method` and `trace_function` use nearly the same function under the hook, I'd expect them to be very similar. Although `install_hook` doesn't use the same one, it should be in the same ballpark.

Anyway, this is a great idea, and it does not harm to compare them (considering I don't expect benchmark instability on these API calls).

PROFeNoM: Added the benchmarks. It appears that `install_hook` is non-negligeably slower, especially considering the latter _doesn't open a span_ 🤔 

```
\Benchmarks\API\HookBench

    benchWithoutHook........................I9 - Mo113.253μs (±2.99%)
    benchHookOverheadTraceMethod............I9 - Mo463.337μs (±1.77%)
    benchHookOverheadFunctionTraceFunction..R1 I4 - Mo452.150μs (±2.14%)
    benchHookOverheadInstallHookOnMethod....R1 I9 - Mo1,271.603μs (±2.30%)
    benchHookOverheadInstallHookOnFunction..R1 I9 - Mo1,235.958μs (±2.85%)
```

pierotibou: Hmm very interesting. Thanks for adding those. We shall look into it indeed then, especially as we have added it in more integrations recently.

Makefile
```diff
@@ -470,6 +470,10 @@ COMPOSER = $(if $(ASAN), ASAN_OPTIONS=detect_leaks=0) COMPOSER_MEMORY_LIMIT=-1 c
 COMPOSER_TESTS = $(COMPOSER) --working-dir=$(TESTS_ROOT)
 PHPUNIT_OPTS ?=
 PHPUNIT = $(TESTS_ROOT)/vendor/bin/phpunit $(PHPUNIT_OPTS) --config=$(TESTS_ROOT)/phpunit.xml
+PHPBENCH_OPTS ?=
+PHPBENCH_CONFIG ?= $(TESTS_ROOT)/phpbench.json
+PHPBENCH_OPCAHE_CONFIG ?= $(TESTS_ROOT)/phpbench-opcache.json
```
bwoebi, line 475: typo

PR #2443, opened 2023-12-29
Roadrunner appsec support

cataphract:
### Description

* Adds AppSec support for RoadRunner.
* Adds integration tests to AppSec
* Adds (limited) ability to suppress function calls, to throw exceptions from advice, and to do recursive calls in advice.

### Reviewer checklist
- [ ] Test coverage seems ok.

### Tickets

[APPSEC-14734] [APPSEC-12702] [APPSEC-12703] [APPSEC-12705]

[APPSEC-14734]: https://datadoghq.atlassian.net/browse/APPSEC-14734?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APPSEC-12702]: https://datadoghq.atlassian.net/browse/APPSEC-12702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APPSEC-12703]: https://datadoghq.atlassian.net/browse/APPSEC-12703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APPSEC-12705]: https://datadoghq.atlassian.net/browse/APPSEC-12705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zend_abstract_interface/interceptor/php7/interceptor.c
```diff
@@ -176,7 +176,15 @@ static inline int zai_interceptor_ext_nop_handler_no_prev(zend_execute_data *exe
...
 
-    return ZEND_USER_OPCODE_DISPATCH;
+    if (&execute_data->func->op_array == op_array) {
+        return ZEND_USER_OPCODE_DISPATCH;
+    } else {
+        // the code was changed, so instead of executing the original handler of
+        // opline->opcode (gotten via zend_vm_get_opcode_handler_func),
+        // we return ZEND_USER_OPCODE_CONTINUE so that user opcode handler
+        // of the new execute_data->opline is executed
+        return ZEND_USER_OPCODE_CONTINUE;
```
bwoebi, line 186: Let's move that check to after zai_hook_memory_table_insert to avoid changing the hot path.

ext/hook/uhook.c
```diff
@@ -250,6 +258,37 @@ static bool dd_uhook_begin(zend_ulong invocation, zend_execute_data *execute_dat
     }
     dyn->hook_data->execute_data = NULL;
 
+    if (dyn->hook_data->suppress_call) {
+        if (ZEND_USER_CODE(execute_data->func->type)) {
+            static union {
+                zend_op zop;
+                zval zv;
+            } retop[] = {[0].zop =
```
bwoebi, line 266: What's this union for? zv seems not used except indefining the zop here... Did you mean struct? (so that .op1.constant points to it)

ext/hook/uhook.c
```diff
@@ -250,6 +258,37 @@ static bool dd_uhook_begin(zend_ulong invocation, zend_execute_data *execute_dat
...
+                zend_vm_set_opcode_handler(&retop[0].zop);
+            }
+            struct {
+                zend_function new_func;
+                zend_function *orig_func; } *fs = emalloc(sizeof *fs);
+            memcpy(&fs->new_func, execute_data->func, sizeof fs->new_func);
+            fs->orig_func = execute_data->func;
+            fs->new_func.op_array.last = 1;
+            fs->new_func.op_array.opcodes = (zend_op *)retop;
+            execute_data->func = &fs->new_func;
```
bwoebi, line 285: Is this part even necessary at all? Doesn't it suffice to set opline appropriately?

ext/hook/uhook.c
```diff
@@ -789,6 +850,57 @@ ZEND_METHOD(DDTrace_HookData, overrideReturnValue) {
...
 }
 
+ZEND_METHOD(DDTrace_HookData, suppressCall) {
+    dd_hook_data *hookData = (dd_hook_data *)Z_OBJ_P(ZEND_THIS);
+
+    if (zend_parse_parameters_none() == FAILURE) {
+        return;
+    }
+
+    hookData->suppress_call = true;
```
bwoebi, line 860: I'm fairly certain this approach doesn't work with inlining in JIT.

You may have to unconditionally blacklist hooks from which you wish to use call suppressions from being inlined: search for ZAI_JIT_BLACKLIST_ACTIVE in this file.

I.e. you will have to make this function accept boolean, always call it and just suppress when actually true, but unconditionally call the blacklisting.

ext/hook/uhook.stub.php
```diff
@@ -75,6 +75,27 @@ public function overrideArguments(array $arguments): bool;
...
+    public function overrideException(\Throwable|null $exception): bool;
+
+    /**
+     * Suppresses the call to the hooked function. Must be called within a pre-hook.
+     * @return bool always 'true'
+     */
+    public function suppressCall(): bool;
+
+    /**
+     * By default, advice is not called when the instrumented function from its advice.
```
bwoebi, line 93: What is "advice"? I think you have the wrong english word.

Anyway, I suppose the goal is allowing the hook to run on a recursively called method?
Any reason to implement it through a method, instead of as a new flag on install_hook? Do we sometimes need to do that conditionally?

ext/hook/uhook.c
```diff
@@ -250,6 +258,37 @@ static bool dd_uhook_begin(zend_ulong invocation, zend_execute_data *execute_dat
...
+                zend_function new_func;
+                zend_function *orig_func; } *fs = emalloc(sizeof *fs);
+            memcpy(&fs->new_func, execute_data->func, sizeof fs->new_func);
+            fs->orig_func = execute_data->func;
+            fs->new_func.op_array.last = 1;
+            fs->new_func.op_array.opcodes = (zend_op *)retop;
+            execute_data->func = &fs->new_func;
+            execute_data->opline = &retop[0].zop;
+        } else {
+            // XXX: not supported yet
```
bwoebi, line 288: This one is going to be interesting to support at all... Because observers are called after the func is fetched in ZEND_DO_ICALL...

ext/hook/uhook.c
cataphract: `retop` has two elements. The first element is the zend_op and the second element is zval. The zval is the NULL that's the op1 constant for the zend_op. This way the address of the zval is after the address of the zend_op. If I used two statics it could happen that the zval would have a lower memory address, and then the offset would be negative. But zend_op.op1.constant is unsigned, so it cannot have negative values.

ext/hook/uhook.c
```diff
@@ -250,6 +258,37 @@ static bool dd_uhook_begin(zend_ulong invocation, zend_execute_data *execute_dat
...
+                zval zv;
+            } retop[] = {[0].zop =
+                             {
+                                 .opcode = ZEND_RETURN,
+                                 .op1_type = IS_CONST,
+                                 .op1 = {.constant = sizeof(retop[0])},
+                                 .op2_type = IS_UNUSED,
+                             },
+                         [1].zv = {.u1.type_info = IS_NULL}};
+            // the race condition doesn't matter
```
bwoebi, line 274: which race condition? Comment seems misleading

ext/hook/uhook.c
bwoebi: Still, did you mean _struct_ instead of _union_? Because currently zop and zv overlap.

ext/ddtrace.c
```diff
@@ -867,6 +873,13 @@ static PHP_MINIT_FUNCTION(ddtrace) {
 
     ddtrace_integrations_minit();
     dd_ip_extraction_startup();
+    ddtrace_user_request_startup();
+
+    // we need to trigger the slow path in the fcall handler in order to be able to suppress calls
+    // otherwise OPLINE is not refreshed after calling the begin observers
+    if (zend_execute_ex == execute_ex) {
+        zend_execute_ex = ddtrace_execute_ex;
+    }
```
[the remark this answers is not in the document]
bwoebi: There's another thing we could leverage though, and it's setting `EG(vm_interrupt) = true;` in the start handler.
This will ensure `zend_interrupt_helper` is called, which will do a `ZEND_VM_ENTER()` reloading the opline. Note that `zend_interrupt_function` must be non-NULL for that, but that's not a problem - just set it to a dummy empty function if it's not NULL yet in post_startup_cb.

ext/hook/uhook.c
bwoebi: In fact you probably need to temporarily replace `internal_function.handler` with your own function, which then will replace `internal_function.handler` with the original function again and call it. ... with a global refcount to avoid replacing it while another thread also has replaced it...

ext/hook/uhook.c
cataphract: I think `xdebug` crashed without it, but I'll confirm.

bwoebi: Makes sense. We can leave that as is.

ext/hook/uhook.c
cataphract: If we define it as a situation where two threads access shared data without synchronization and at least one is a write, then this code is clearly racy in ZTS (`zend_vm_set_opcode_handler(&retop[0].zop)` can be called twice, but even one thread writing the opcode handler and the other reading a non-NULL value would be a race condition). However, the only two possible values read for the opcode handler are a NULL value and the correct non-NULL value (at least on amd64/arm64, where aligned 64-bit writes cannot be partially seen).

bwoebi: On that note, is there a reason for this to not just simply be a true global initialized in zai_uhook_minit?

ext/ddtrace.c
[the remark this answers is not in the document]
bwoebi: Also note: needs to be PHP 8.0+ only.

ext/hook/uhook.c
cataphract: Is that enough though? Can't it be set in the startup hook of another zend extension?

ext/ddtrace.c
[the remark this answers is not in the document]
cataphract: it's a bit more complicated because of a SAVE_OPLINE before calling the interrupt function and the fact that the interrupt function is global, but I think it should be right now.

ext/hook/uhook.c
bwoebi: Let's leave it as is, it's good enough here. Let's not make it more complicated than it needs to be.

ext/hook/uhook.c
```diff
@@ -862,7 +917,7 @@ ZEND_METHOD(DDTrace_HookData, suppressCall) {
     RETURN_TRUE;
```
bwoebi, line 917: Can we please log when disableJitInlining was not called before, to avoid forgetting calling it?

ext/hook/uhook.c
```diff
@@ -211,6 +218,30 @@ static bool dd_uhook_match_filepath(zend_string *file, zend_string *source) {
...
+
+    if (ex != expected_ex) {
+        expected_ex = NULL;
+        goto call_orig;
+    }
+
+    expected_ex = NULL;
+    ex->opline = ex->func->op_array.opcodes;
+
+call_orig:
```
bwoebi, line 238:
```suggestion
    if (expected_ex) {
        if (ex == expected_ex) {
            ex->opline = ex->func->op_array.opcodes;
        }

        expected_ex = NULL;
    }
```

Simpler logic?

ext/hook/uhook.c
```diff
@@ -250,6 +282,50 @@ static bool dd_uhook_begin(zend_ulong invocation, zend_execute_data *execute_dat
...
+                zend_vm_set_opcode_handler(&retop.zop);
+            }
+            struct {
+                zend_function new_func;
+                zend_function *orig_func; } *fs = emalloc(sizeof *fs);
+            memcpy(&fs->new_func, execute_data->func, sizeof fs->new_func);
+            fs->orig_func = execute_data->func;
+            fs->new_func.op_array.last = 1;
+            fs->new_func.op_array.opcodes = &retop.zop;
+            fs->new_func.op_array.opcodes = &retop.zop;
```
bwoebi, line 309:
```suggestion
            fs->new_func.op_array.opcodes = &retop.zop;
```

ext/ddtrace.c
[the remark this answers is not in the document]
bwoebi: You're right, I missed that. But looks good to me now :-)

bwoebi (approved):
Ack for the uhook changes :-)

Maybe emit a LOG() line if suppressCall is attempted on an internal function, but that's it.

ext/hook/uhook.stub.php
```diff
@@ -145,5 +152,7 @@ function install_hook(
  * @param int $id The id to remove.
  * @param string $location A class name (which inherits this hook through inheritance), which to specifically remove
  * this hook from.
+ * @return void no return, not formally declared void because of a buggy debug assertion in PHP 7.1
+ *              ("return value must be of type void, null returned")
  */
-function remove_hook(int $id, string $location = ""): void {}
+function remove_hook(int $id, string $location = "") {}
```
bwoebi, line 158: Why would that happen? https://heap.space/xref/PHP-7.1/Zend/zend_execute.c?r=7f6387b5#1045 seems to properly check for it?

src/Integrations/Integrations/Roadrunner/RoadrunnerIntegration.php
```diff
@@ -30,6 +35,73 @@ public function requiresExplicitTraceAnalyticsEnabling()
         return false;
     }
 
+    public static function build_req_spec(\Spiral\RoadRunner\Http\Request $req) {
```
bwoebi, line 38: Now we essentially have two translations:
from Http\Request to superglobals, and from Http\Request to span attributes.

We essentially re-implemented everything from ddtrace_set_root_span_properties as PHP code in the hook. Now, you're anyway passing things in a format which the extensions understand to the extension.

Would it be possible to have notify_start call out to the internal logic (from `ddtrace_set_root_span_properties`) for setting data from the superglobals? And notify_commit to some logic which currently resides in `_serialize_meta` (the part guarded by `ddtrace_span_is_entrypoint_root(span)` there)?

src/Integrations/Integrations/Roadrunner/RoadrunnerIntegration.php
```diff
@@ -2,10 +2,15 @@
 
 namespace DDTrace\Integrations\Roadrunner;
 
-use DDTrace\Tag;
+use DDTrace\HookData;
 use DDTrace\Integrations\Integration;
+use DDTrace\Tag;
 use DDTrace\Type;
 use DDTrace\Util\Normalizer;
+use http\Exception\RuntimeException;
```
bwoebi, line 10:
```suggestion
```
Mistake?

ext/user_request.stub.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+/** @generate-class-entries */
```
bwoebi, line 3: We've had users complain about us having multiple stub files and it thus making it easy to miss that there are others - in the PHP ecosystem stub files are generally a single file per extension, and I was looking into merging them eventually.

Could you please put it into ddtrace.stub.php? Thanks!

ext/user_request.c
```diff
@@ -0,0 +1,168 @@
...
+    for (size_t i = 0; i < reg_listeners.size; i++) {
+        ddtrace_user_req_listeners *listener = reg_listeners.listeners[i];
+        zend_array *repl = listener->response_committed(listener, span, status, headers);
+        if (repl != NULL && replacement_resp == NULL) {
+            replacement_resp = repl;
+        } else if (repl != NULL) {
+#ifdef GC_TRY_DELREF
+            GC_TRY_DELREF(repl);
+#else
+            GC_DELREF(repl);
```
bwoebi, line 109: This doesn't sound safe to me? I think you should instead polyfill GC_TRY_DELREF() in compatibility.h and check for `GC_FLAGS() & IS_ARRAY_IMMUTABLE`.

bwoebi: Anyway, a better suggestion would be zend_array_release(), which does the delref and rc check for you. And we already polyfill that one.

ext/user_request.c
```diff
@@ -0,0 +1,168 @@
...
+#endif
+            if (GC_REFCOUNT(repl) == 0) {
+                zend_array_destroy(repl);
+            }
```
bwoebi, line 74:
```suggestion
            zend_array_release(repl);
```

Ditto here.

Anilm3 (commented):
Still a lot to look through...

ext/hook/uhook.stub.php
cataphract: The error is not in the condition that you point out. That check does indeed pass. The problem is that it's missing a return so it fails the check next. See in PHP 7.2: https://heap.space/xref/PHP-7.2/Zend/zend_execute.c?r=902d39a3#998

ext/user_request.stub.php
cataphract: I mean I can do it, but then I can't really include ddtrace_arginfo.h in user_request.c, otherwise I'll get a bunch of warning about all those unused statics + likely code bloat. Then if I don't I'll have global (non-static) function declarations for the PHP functions without prototypes, which is also a compiler warning (depending on the compiler and settings). Generally, it just seems like a less cohesive solution.

Are you sure about this?

appsec/src/extension/ddappsec.h
```diff
@@ -16,20 +16,19 @@
 #include <stdbool.h>
 
 typedef enum _enabled_configuration {
-    NOT_CONFIGURED = 0,
-    ENABLED,
-    DISABLED
+    APPSEC_ENABLED_VIA_REMCFG = 0,
```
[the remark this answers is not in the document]
cataphract: The point was not really to save bits, the problem was that the two variables before were not a canonical representation of the possible states, making it more difficult to reason about. IIRC, there were two variables each with three possible values. This means 3^2 = 9 possible combinations. This OTOH is an accurate representation of the possible configuration values: fully enabled, fully disabled, and enabled to be enabled via remote config.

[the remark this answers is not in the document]
Anilm3: Okay that's fair enough, as mentioned in a separate comment, having some comments explaining this would be very useful.

appsec/src/extension/ddappsec.c
```diff
@@ -438,31 +343,29 @@ __thread void *unspecnull TSRMLS_CACHE = NULL;
...
+    if ((!get_global_DD_APPSEC_TESTING() && !dd_trace_enabled()) ||
+        (strcmp(sapi_module.name, "cli") != 0 && sapi_module.phpinfo_as_text)) {
+        DDAPPSEC_G(enabled) = APPSEC_FULLY_DISABLED;
+        DDAPPSEC_G(active) = false;
+        DDAPPSEC_G(to_be_configured) = false;
+    } else if (!dd_cfg_enable_via_remcfg()) {
+        DDAPPSEC_G(enabled) = get_DD_APPSEC_ENABLED() ? APPSEC_FULLY_ENABLED
+                                                      : APPSEC_FULLY_DISABLED;
+        DDAPPSEC_G(active) = get_DD_APPSEC_ENABLED() ? true : false;
+        DDAPPSEC_G(to_be_configured) = false;
```
estringana, line 355: Not sure If I missed something but, when appsec is enabled dynamically by remote config, it status can change many times. So saying `it does not have to be configure` sounds confusing.

appsec/src/extension/ddappsec.c
```diff
@@ -438,31 +343,29 @@ __thread void *unspecnull TSRMLS_CACHE = NULL;
...
     } else {
-        DDAPPSEC_G(enabled_by_configuration) = NOT_CONFIGURED;
+        DDAPPSEC_G(enabled) = APPSEC_ENABLED_VIA_REMCFG;
```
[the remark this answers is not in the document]
estringana: Before and now(I believe) on every request, the extension was doing a `config_sync` when the app was not disabled by config and not enabled by RC. This was done on purpose. Why do you think it's a bug?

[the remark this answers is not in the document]
Anilm3: It's a bug because that's not what I defined in the protocol RFC. `config_sync` was only intended as a mechanism to enable a disabled extension through remote config, changing the state was meant to be done by responding to a `request_init` with a `config_features`.

IIRC we discussed at length the new requirement of being able to have multiple responses to a single message so that we could update the extension configuration during a call to `request_init`, while still being enabled (e.g. if we had the ability to change the extension's configuration through RC, which we currently don't).

In saying that, I guess it's fair to assume that there are other pieces missing from the current implementation.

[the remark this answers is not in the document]
estringana: > config_sync was only intended as a mechanism to enable a disabled extension through remote config
This is what it does, so on every request after the extension has been disabled by RC, it does a `config_sync` to check if it has to be enabled or not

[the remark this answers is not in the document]
Anilm3: No, it currently sends it on every request even if it's enabled.

ext/user_request.stub.php
bwoebi: You don't need a prototype to define a function? You need a prototype to use it. ddtrace.c should not care of whether it's defined there or not, that's the linkers problem. So you don't need to include ddtrace_arginfo.h at all.

ext/hook/uhook.stub.php
bwoebi: Your link doesn't help - because you show a fixed version.
https://github.com/php/php-src/commit/9b0df97a42749464069859c5ca77c0241a8b05ca is likely what you mean - all versions before 7.1.20 are broken.

cataphract: Yeah I was just pointing to return statement that was added later.

src/Integrations/Integrations/Roadrunner/RoadrunnerIntegration.php
cataphract: Can you check if https://github.com/DataDog/dd-trace-php/pull/2443/commits/04e4c6c8b8c8f6ea17007635fd6e1a00a7f1c182 is what you had in mind?

ext/user_request.stub.php
cataphract: what I mean is that gcc/clang will warn, depending on the settings:

```
glopes /tmp $ cat /tmp/a.c
int v() { return 2; }
int main() { return v(); }
glopes /tmp $ gcc  -Wmissing-prototypes /tmp/a.c
/tmp/a.c:1:5: warning: no previous prototype for ‘v’ [-Wmissing-prototypes]
    1 | int v() { return 2; }
      |     ^
glopes /tmp $ clang  -Wmissing-prototypes /tmp/a.c
/tmp/a.c:1:5: warning: no previous prototype for function 'v' [-Wmissing-prototypes]
int v() { return 2; }
    ^
/tmp/a.c:1:1: note: declare 'static' if the function is not intended to be used outside of this translation unit
int v() { return 2; }
^
static 
1 warning generated.
```

appsec/src/extension/msgpack_helpers.c
```diff
@@ -61,6 +61,47 @@ void dd_mpack_write_zval(mpack_writer_t *nonnull w, zval *nullable zv)
...
+    if (!arr) {
+        mpack_write_nil(w);
+    }
+
+    uint32_t num_elems = zend_hash_num_elements(arr);
+    dd_php_array_type arr_type = dd_php_determine_array_type(arr);
+    if (arr_type == php_array_type_sequential) {
+        mpack_start_array(w, num_elems);
+        zval *val;
+        ZEND_HASH_FOREACH_VAL((zend_array *)arr, val)
```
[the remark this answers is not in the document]
cataphract: I think it surely does, but quite out of scope for this PR.

appsec/CMakeLists.txt
```diff
@@ -30,6 +31,8 @@ option(DD_APPSEC_TESTING "Whether to enable testing" ON)
 
 add_subdirectory(third_party EXCLUDE_FROM_ALL)
 
+include("cmake/patchelf.cmake")
```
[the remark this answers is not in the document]
cataphract: I'm inclined to keep it in order to cut building times once I add back the alpine tests. That way appsec (and hopefully ddtrace too) need not be built twice.

appsec/src/extension/ddappsec.c
cataphract: "To be configured" means "yet to be configured" (=not configured yet") or "it is to be configured" (=it's arranged/programmed for it to be configured in the future"). The purpose of this setting is just to indicate in MINIT if we got a configuration setting for the state of remote config yet (if, for instance, we can't contact the daemon, then to_be_configured will stay with the value `true`). In this branch, appsec is explicitly enabled or disabled, so we're never waiting for a setting from remote config. Does this make sense? I also added some comments in bbdb00390

ext/user_request.stub.php
bwoebi: Do we have it active? If not, it's stupid, but trivial to write `ZEND_FUNCTION(foo); ZEND_FUNCTION(foo) { ... }`.

appsec/src/extension/msgpack_helpers.c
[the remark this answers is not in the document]
Anilm3: Fair enough, I'll add a separate ticket for this.

Anilm3 (approved):

PR #2451, opened 2024-01-04
Add span links capabilities to the OTel API

PROFeNoM:
### Description

Add the span links to the OpenTelemetry API.

Please note the following points about converting Datadog spans to OpenTelemetry (OTel) spans and creating span links:

- When converting a Datadog span to an OTel one, its current span links are also converted.
- When creating a span using the OTel API, the corresponding `\DDTrace\SpanLink` instances are also created.

It is important to note that span links can be added to a span at any point during its lifetime using the Datadog (DD) API, but such links can only be added using the OTel API during the build phase. This limitation is not a problem with the DD API, but it is with the OTel API. To address this, DD span links are actualized onto the OTel span when creating an immutable span from it (`toSpanData()`). This is the only way to access the links associated with an OTel span.

However, the proposed implementation does not work if we consider span link removals (which only exist using the DD API). To handle span link removals, we would need to check for the difference between currently associated links on the OTel span and on the DD span. This check is not necessarily difficult to do, as long as we acknowledge that the trace identifiers of span links are immutable. The issue is that this approach would introduce recurring, unnecessary overhead and complexity for a non-required capability. I'm leaning towards not adding unnecessary overhead over an already massive one when using the OTel API. 

### Reviewer checklist
- [ ] Test coverage seems ok.


src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -75,6 +92,29 @@ private function __construct(
...
+
+                $spanLink = new SpanLink();
+                $spanLink->traceId = $linkContext->getTraceId();
+                $spanLink->spanId = $linkContext->getSpanId();
+                $spanLink->traceState = (string)$linkContext->getTraceState(); // __toString()
+                $spanLink->attributes = $link->getAttributes()->toArray();
+                $spanLink->droppedAttributesCount = 0; // Attributes limit aren't supported/meaningful in DD
+
+                // Save the link
+                ObjectKVStore::put($spanLink, $this->uniqueIdentifier, $link);
```
bwoebi, line 114: Maybe I'm missing something, but why an uniqueIdentifier? Isn't the DD SpanLink to OTel SpanLink mapping 1:1? It would be needed if the span links were dependent on the specific span they're attached to, but they aren't? So just using `"OTelLink"` as key would be fine too?

src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -400,4 +444,39 @@ public function getDDSpan(): SpanData
...
+                    Attributes::create($datadogSpanLink->attributes ?? [])
+                );
+
+                // Save the link
+                ObjectKVStore::put($datadogSpanLink, $this->uniqueIdentifier, $link);
+                $otel[] = $link;
+            } else {
+                // Save the link
+                $otel[] = $link;
+            }
```
bwoebi, line 475: Nit: Move out of branch

bwoebi (approved):
Thanks! Looks good to me :-)

PR #2460, opened 2024-01-11
Handle async requests using Guzzle

### LINKED ISSUE #2444: [Bug]: seeing host-unparsable-host and "CURL request failure" in apm in curl_multi_exec context
### Bug report

I was waiting for the fix for https://github.com/DataDog/dd-trace-php/issues/1156 since some time and was happy to try out 0.95 and later 0.96. while the problem described in https://github.com/DataDog/dd-trace-php/issues/1156 looks fixed now, i see the following in dd-apm now even if the curl calls work fine like before as far as i can tell:
image
going back to pre 0.91.2 (the version i used before) brings back the old behavior (no curl_exec spans but also no errors in dd).
the guzzle transfer span right before the curl multi spans looks good, btw (no problem with the hostname):
image
image
the actual curl request url looks something like that: https://foo.bar.eu:8443/foo/bar/baz .
this code runs locally inside a docker container via docker-compose with configured extra-hosts like that: 
foo.bar.eu: 172.xxx.xxx.xxx

sadly i was not able to find where the ddtrace code brakes while debugging this issue. i did not find any other issue regarding this topic so i created this issue. do you need more info from my side?

thx in advance!

### PHP version

8.1.21

0.95.0 and 0.96.0

### Installed extensions

[PHP Modules]
apcu
bcmath
Core
ctype
curl
date
ddappsec
ddtrace
dom
fileinfo
filter
ftp
gd
hash
iconv
json
libxml
mbstring
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
pdo_sqlsrv
Phar
posix
readline
redis
Reflection
session
SimpleXML
sodium
SPL
sqlite3
sqlsrv
standard
tokenizer
xdebug
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib

[Zend Modules]
Xdebug
Zend OPcache
ddappsec
ddtrace

### Output of `phpinfo()`

image
image
image
image
image
image
image

### Upgrading from

0.91.2

PROFeNoM:
### Description

Fixes #2444

**TL;DR**: Set the curl information as soon as possible, and don't overwrite it.

<details>
  <summary><strong>Shortened Explanation</strong></summary>

The current implementation fundamentally assumes that `curl_multi_info_read` is called after `curl_multi_exec`'s completion, which is not the case in Guzzle, resulting in either:

1. Overwriting the information that was already set on a `curl_exec` span by dummy information;
2. Setting dummy information on the `curl_exec` span.

</details>

<details>
  <summary><strong>Explanation</strong></summary>

The existing implementation fundamentally assumes a specific execution flow, particularly that `curl_multi_info_read` is invoked after the completion of `curl_multi_exec`. However, in the context of Guzzle Promises and its `tick` method (which repeatedly checks only once for completed requests), this assumption is not valid, leading to unintended consequences.

The sequence of operations during a typical execution involves the following flow:

- [Repeat](https://github.com/guzzle/guzzle/blob/cd634c213de11e0f229651173428f5c27d398628/src/Handler/CurlMultiHandler.php#L180)
  - [Call CurlMultiHandler::tick](https://github.com/guzzle/guzzle/blob/cd634c213de11e0f229651173428f5c27d398628/src/Handler/CurlMultiHandler.php#L185)
  - [Call curl_multi_exec from CurlMultiHandler::tick](https://github.com/guzzle/guzzle/blob/cd634c213de11e0f229651173428f5c27d398628/src/Handler/CurlMultiHandler.php#L167)
  - [Call CurlMultiHandler::processMessages from CurlMultiHandler::tick](https://github.com/guzzle/guzzle/blob/cd634c213de11e0f229651173428f5c27d398628/src/Handler/CurlMultiHandler.php#L170)
  - [Call curl_multi_info_read from CurlMultiHandler::processMessages](https://github.com/guzzle/guzzle/blob/cd634c213de11e0f229651173428f5c27d398628/src/Handler/CurlMultiHandler.php#L229)
    - `curl_multi_info_read` can initiate the following course of action:
      - [Trigger the hook set on curl_multi_info_read from CurlIntegration](https://github.com/DataDog/dd-trace-php/blob/ebbd19d04182cd275678c7bab53f821438fb50d2/src/Integrations/Integrations/Curl/CurlIntegration.php#L177)
      - <strong>[Remove the handle](https://github.com/guzzle/guzzle/blob/cd634c213de11e0f229651173428f5c27d398628/src/Handler/CurlMultiHandler.php#L235)</strong>

Completion and `curl_getinfo`
  -  As the completion of `curl_multi_exec` occurs in `CurlMultiHandler::tick` under the `CurlMultiHandler::execute` loop, a final call to `CurlMultiHandler::tick` is triggered, as the queue might not be empty (array_shift left to do).
     - Therefore, an additional call to the already completed and processed handle may be done again. This handle may or may not lead to a furnished curl_getinfo, depending on whether it was previously read.
  - Alternatively, the `curl_multi_exec` call may complete, but Guzzle can retry handles, hence inducing the creation of an additional `curl_multi_exec` span. **This retry scenario is the root cause of the flakiness observed in the added test.**

To address these issues, the proposed changes in this pull request aim to set the curl information as soon as possible, avoiding overwriting or putting dummy information during the execution flow outlined above, i.e.:
- Set the information either from the `curl_multi_info_read` or `curl_multi_exec` hook, whichever comes first with the curl information
- Only overwrite the information if no network_destination is set (or the currently set information resulted in `unparsable-host`), which is theoretically always set when setting the curl attributes.

</details>

**Notes:**
- This PR also adds snapshot testing for non-web requests :)
  - A Memcache test was changed as a result, but I guess it shouldn't matter, considering the test is still valid; it's just another environment?

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Thanks, hopefully we now cover everything :-)

PR #2467, opened 2024-01-12
refactor(profiling): extract wall_time.rs file

morrisonlevi:
### Description

PROF-8912

Does two things:

1. Extracts bits from `lib.rs` and `capi.rs` and puts them in a new `wall_time.rs` file.
2. Uses `ddog_php_prof_` instead of `datadog_profiling_` for functions in `wall_time.rs`.  This function is used by the tracer so it updates `ext/engine_hooks.c` too.

### Reviewer checklist
- [x] Test coverage seems ok.


realFlowControl (approved):
Whoop Whoop

PR #2468, opened 2024-01-13
breaking-change: require system INIs for profiling features

morrisonlevi:
### Description

This is a WIP and is still subject to discussion. Creating these as system INIs means the `Profiler` object can store them and we can stop passing them around everywhere. This cleans up the code, avoids some potential edges with aliasing thread local variables, and paves the way for more refactoring.

Filtering the `SampleValues` as been moved into a new module `src/profiling/sample_type_filter.rs`. Related tests have also been moved there. These tests no longer depend on the `Profiler` at all.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl:
One thing that concernes me a bit is that we have the configuration in `REQUEST_LOCALS` and as invariants in the `Profiler`. Now from what I know the INI settings can not be changed anymore, but having the config in two places leaves me with the fear that there could be a split state at one time in the future.
We should try and have one source of truth for the state of the profiler. Granted, calling `PROFILER.lock()` everytime when I just need to know if for example exception profiling is enabled is too much.

realFlowControl (approved):
As discussed, let's address my last comment in a new PR.

realFlowControl (approved):

PR #2470, opened 2024-01-15
feat(profiling) PHP ZTS support for CPU- and walltime profiling

realFlowControl:
### Description

This PR will:
- make CPU- and Walltime profiling work in PHP ZTS
- build ZTS profiler in CI and add to the artefact we ship
- run tests for ZTS version
- disable allocation profiling in ZTS (would crash otherwise and will be shipped later)

PROF-8921

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/tests/phpt/jit_01.phpt
```diff
@@ -15,6 +15,15 @@ if (!extension_loaded('datadog-profiling'))
...
     echo "skip: JIT not available on aarch64 on PHP 8.0", PHP_EOL;
+
+// TODO: remove once ZTS support for allocation profiling is done
+ob_start();
+phpinfo(INFO_GENERAL);
+$phpinfo = ob_get_contents();
+ob_end_clean();
+if (strpos($phpinfo, 'Thread Safety => enabled') !== false) {
+    echo "skip: not support on ZTS builds at the moment";
+}
```
morrisonlevi, line 26: This can be simplified to use the `PHP_ZTS` constant. Applies to all 3 phpt files.

profiling/build.rs
```diff
@@ -6,6 +6,9 @@ use std::path::PathBuf;
 use std::process::Command;
 
 fn main() {
+    // This makes sure `build.rs` will be run on every `cargo build` command and helps when
+    // switching PHP versions in local dev environment and when building in CI
+    println!("cargo:rerun-if-changed=NULL");
```
realFlowControl, line 11: I need to remove this again. This slows down local rebuilds (and the local rust-analyzer that I use as LSP) heavily. Sadly I got lots of "interesting" build errors in CI if I remove it, even a `cargo clean` would not solve these, so clearly this is not a solution to the problem.

morrisonlevi (approved):
A few things to manually check:

 - Does `datadog-setup.php` need any changes? Based on how it detects support, it should Just Work but we may need to change comments or output or something.
 - Have you checked Apache in the worker modes that support threading? I'd also double-check `apachectl graceful`.

Aside from these manual testing concerns, it looks fine to me.

PR #2464, opened 2024-01-12
Re-enable code coverage using codecov

PROFeNoM:
### Description

This PR brings back code coverage using codecov.

- [X] `CODECOV_TOKEN` added to CircleCI > Project Settigns > [Environment Variables](https://app.circleci.com/settings/project/github/DataDog/dd-trace-php/environment-variables?return-to=https%3A%2F%2Fapp.circleci.com%2Fpipelines%2Fgithub%2FDataDog%2Fdd-trace-php%3Fbranch%3Dmaster)
- [X] Add the `CODECOV_TOKEN` project environment variable to the docker image
  - Also added `CIRCLE_PROJECT_USERNAME` and `CIRCLE_PROJECT_REPONAME`, as they are required when [setting the slug](https://github.com/codecov/uploader/blob/6ccf3c3b145904308d38cdef1ad615df2ff6f9e0/src/ci_providers/provider_circleci.ts#L33-L49) (\<org\>/\<repo\>)
  - Also added `CI` and `CIRCLECI`, as they are required to properly [detect CircleCI as the CI provider](https://github.com/codecov/uploader/blob/6ccf3c3b145904308d38cdef1ad615df2ff6f9e0/src/ci_providers/provider_circleci.ts#L5-L7)
- [X] Generate a codecov PR comment
- [X] Create different [flags](https://app.codecov.io/gh/DataDog/dd-trace-php/flags) for appsec & the tracer (extension)
- [X] Consider adding integrations (+ corresponding flag) coverage

You can see a sample report below.

One limitation of this coverage is that functions written in the extension but solely tested in PHP (I don't think of any out of my head, but _hypothetically_) would be considered uncovered.

### Reviewer checklist
- [ ] Test coverage seems ok.
- [ ] Appropriate labels assigned.

PR #2474, opened 2024-01-17
fix: Missing test coverage

PROFeNoM:
### Description

Oops, I guess I forgot about them in the initial PR #2464 

Two Issues:
1. There were no coverage reports for Integrations because of compiled files (_generated*.php). These, not being the source files, were not counted as coverage. To fix this issue, un-compiled files are used. 
2. Solving the latter issue resolved the problems related to `test_integrations_*,` but there was still no web test coverage. This is because a new `Process` is spawned in the web tests. Therefore, phpunit+xdebug doesn't have any visibility over what's happening, and no coverage is captured (despite it being where the valuable information lies). To fix this issue, a script (`save_code_coverage.php`) will be prepended to the queried file to collect the code coverage manually. These generated code coverage files will be saved in `reports/cov`, where they can be merged to a clover file using `make merge_coverage_reports.`

Notes:
- PHPUnit was removed from some composer.json, as it would sometimes conflict with the script (which uses phpunit's capabilities)
- `tests/Appsec/Mock.php` was modified to prevent its element from being loaded twice. This happened because of the introduction of `save_code_coverage.php`, which requires `vendor/autoload.php`
- The log `"Backing up auto_prepend_file '%s'"` is moved to a debug-type one, as it would 1. make the tests fail and 2. shouldn't be a warning.
- Some tests that were too flaky/failing under coverage mode were skipped. They are anyway tested when running in non-coverage mode.
- Coverage tests (integrations & web) are only tested on 7.4 & 8.2, to save time.
- The `tracer-integrations` and `tracer-php-api` flags have been merged into one flag, `tracer-php`. This makes it more comprehensive.

### Reviewer checklist
- [ ] Test coverage seems ok.


Makefile
```diff
@@ -1316,7 +1318,7 @@ test_scenario_%:
 	$(Q) $(COMPOSER_TESTS) scenario $*
 
 merge_coverage_reports:
-	$(PHPCOV) merge --clover reports/coverage.xml reports/cov
+	php -d memory_limit=-1 $(PHPCOV) merge --clover reports/coverage.xml reports/cov
```
PROFeNoM, line 1321: The reason behind `memory_limit=-1` here is the number of `.cov` files being processed. I'm talking about 100s of coverage files being merged (1 per request + 1 per test suite).

bwoebi (approved):
Looking good, let's merge it before the branches diverge :-)

PROFeNoM:
Just did a final rebase onto master to be sure everything is good :+1: Will merge afterwards

PR #2475, opened 2024-01-17
fix(profiler) zai tests

realFlowControl:
### Description

Fixes the failing zai test in `master` that came after merging #2467. Also enabled running the zai tests on profiler changes so that we do not fall for this again.

### Reviewer checklist
- [x] Test coverage seems ok.


bwoebi (approved):
Thanks :-)

PR #2477, opened 2024-01-17
Extract `http.route` in Symfony

estringana:
### Description

Extract `http.route` in Symfony using the given target `controller::method`, `locale` and `name`. This PR is only targeting PHP Annotations and Docblocks

APPSEC-10781

### Reviewer checklist
- [ ] Test coverage seems ok.


tests/Integrations/Symfony/V3_0/CommonScenariosTest.php
```diff
@@ -42,6 +42,8 @@ public function provideSpecs()
                     )->withExactTags([
                         'symfony.route.action' => 'AppBundle\Controller\CommonScenariosController@simpleAction',
                         'symfony.route.name' => 'simple',
+                        'http.route' => '/simple',
+                        'http.route' => '/simple',
```
bwoebi, line 46:
```suggestion
                        'http.route' => '/simple',
```

tests/Unit/Integrations/Route.php
```diff
@@ -0,0 +1,179 @@
+<?php
+
+namespace DDTrace\Tests\Unit\Integrations;
+
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
+class Route
```
estringana, line 6: This file was taken from Symfony 6.2. I haven't found another approach rather than copy to have it on our unit tests

src/Integrations/Integrations/Symfony/PathExtractor.php
```diff
@@ -0,0 +1,187 @@
...
+            $methodName = $exploded[1];
+        }
+
+        if (!class_exists($className)) {
+            return;
+        }
+        $class = new ReflectionClass($className);
+        if ($class->isAbstract()) {
+            return;
+        }
```
PROFeNoM, line 46: Why do we return if the class is abstract? Why could we even have an abstract class here? Does it mean `$request->attributes->get('_controller')` references an abstract class?

estringana: I based this algo on `https://github.com/symfony/routing/blob/6.2/Loader/AnnotationClassLoader.php` but you're right, since we are on runtime(as opposed to that class), we should not get an abstract here. I'll remove it. Thanks

PROFeNoM: Thanks for the link :)

tests/Unit/Integrations/Route.php
estringana: Sorted

PROFeNoM (approved):
The use of reflection indeed introduces some overhead, but it seems reasonable (+5%). I guess `getAnnotations`, particularly, with the creation of instances, is expensive.

bwoebi (changes requested):
1.5ms fixed overhead according to benchmarks for just reading the routes is way too much.

We do have some overhead as to provide our functionality at all, but this is just too much, sorry. The code looks nice and well crafted but too expensive :-/

If you want to have this functionality you'll have to chose a wholly different approach, like hooking into Symfony\Component\Routing\Matcher\UrlMatcher (and \Symfony\Component\Routing\Matcher\Dumper\CompiledUrlMatcherTrait). It also probably needs a bit more version specific code, but I suppose that's the price to pay for being faster.

PR #2479, opened 2024-01-18
build(profiler): update dependencies and pin ahash

realFlowControl:
### Description

This updates the cargo dependencies

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/Cargo.toml
+++ b/profiling/Cargo.toml
@@ -12,7 +12,9 @@ bench = false # disables cargo build in libtest bench
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-ahash = { version = "0.8" }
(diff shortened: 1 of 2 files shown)


profiling/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "datadog-php-profiling"
 version = "0.96.0"
 edition = "2021"
 license = "Apache-2.0"
-rust-version = "1.64"
+rust-version = "1.71"
```
bwoebi, line 6: While we currently say 1.71 in our docs, the code compiles perfectly fine with 1.70, which is the newest available via the debian package manager.

The differences between 1.70 and 1.71 are also really small...

realFlowControl: Oh yes, you are right, I reverted that change.

morrisonlevi (approved):

PR #2483, opened 2024-01-19
fix: handle Apache graceful restarts more accurately

morrisonlevi:
### Description

**Important note:** this does not guarantee that we are fully complaint with `apachectl graceful`. For instance, there could potentially be issues with weak symbols according to Bob. This just gets us one step closer.

In Apache, processes can be gracefully restarted with `apachectl graceful`. When this happens, after doing the PHP shutdown process, the processes will not end but instead start the startup process (minit, activate, etc) again. Really, when we do something on "first rinit," that's short-hand for "first rinit after each minit." Or at least, if you want to be technically correct.

Without doing this, some configuration changes wouldn't take place. That is probably harmless but definitely confusing. The hack of re-initializing the `Once` variables is not the prettiest but it should be safe within PHP's extension lifecycle model. So far, I have tested this manually on Apache + mod_php on PHP 8.3 without any issues.

https://datadoghq.atlassian.net/browse/PROF-9135

### Reviewer checklist
- [x] Test coverage seems ok.


profiling/src/lib.rs
```diff
@@ -290,6 +290,17 @@ extern "C" fn minit(_type: c_int, module_number: c_int) -> ZendResult {
...
+    // each minit/mshutdown cycle. In Apache, when doing `apachectl graceful`,
+    // there can be more than one of these cycles per process.
+    // Re-initializing these on each minit allows us to do it once per cycle.
+    // This is unsafe generally, but all SAPIs are supposed to only have one
+    // thread alive during minit, so it should be safe here specifically.
+    unsafe {
+        ZAI_CONFIG_ONCE = Once::new();
+        RINIT_ONCE = Once::new();
+    }
+
```
realFlowControl, line 303: I would think that resetting the `Once` should go to `mshutdown`. It should already be initialised in the `static` below, so this would be a second initialisation. Additionally we could check for the Apache SAPI and only do it if we detect Apache, but maybe that is not worth it, as a `Once::new()` seems cheap and other SAPI's (now or in the future) might do the same ;-)

realFlowControl (approved):

profiling/src/lib.rs
morrisonlevi: Technically, you are right. I'm not sure why, but it makes me more comfortable to put it in startup than shutdown. I'll think about it more and see if I can find a real justification or not.

realFlowControl: "more comfortable" as in better save than sorry?
I can totally see that, let's leave it as is.

ext/handlers_curl.c
```diff
@@ -176,7 +176,7 @@ static HashTable *ddtrace_curl_multi_get_gc(zend_object *object, zval **table, i
 }
 
 static pthread_once_t dd_replace_curl_get_gc_once = PTHREAD_ONCE_INIT;
-static zend_object_handlers *dd_curl_object_handlers;
+ZEND_TLS zend_object_handlers *dd_curl_object_handlers = NULL;
```
bwoebi, line 179: I don't understand this change?

morrisonlevi: Previously on every call to `curl_multi_init`, every thread would write to this address. This is obviously not thread-safe. Making it thread-local fixes the problem and required no further changes, because whichever thread wins the Once control will write to handler using its thread-local handlers. Does that make sense?

bwoebi: Yes, every thread would write to that address, but it would be the same value over and over again, making it unproblematic.
At least we assume that the curl handlers are the same for each curl_multi_handle, otherwise the whole thing doesn't work anyway.

bwoebi: But it doesn't do harm to change it - so I don't care.

PR #2487, opened 2024-01-23
refactor(profiling): use SystemSettings in locals

morrisonlevi:
### Description

PROF-8912

The recently introduced `SystemSettings` had its data duplicated in `RequestLocals` because there were multiple ways to handle that issue. This PR is one way to resolve this.

There's now one `SystemSettings` in memory, and it's defined in `src/config.rs`. The `RequestLocals` and `Profiler` objects hold a non-null pointer to it. The system settings are fetched from the zai config's memoized entries, rather than the thread-locals, and this is done on first rinit. These changes caused a fair bit of churn in the codebase.

There are a few side effects as well:

- The agent endpoint is no longer part of the profile index key, since that's part of system configuration. Less things to hash and compare, so this is a tiny perf reduction on each sample collected.
- The allocation profiler does some runtime checks and disables the allocation profiler under certain circumstances. One of these was moved to rinit, but the other is now a runtime panic, because the new invariants cannot be upheld per-request.
- The minfo for logging will now say "off (profiling disabled)" when profiling is disabled. Previously, it reported whatever the value would be if profiling was enabled but it was actually off.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation.rs
+++ b/profiling/src/allocation.rs
@@ -109,10 +109,25 @@ lazy_static! {
     static ref JIT_ENABLED: bool = unsafe { zend::ddog_php_jit_enabled() };
 }
 
+pub fn first_rinit_should_disable_due_to_jit() -> bool {
+    if NEEDS_RUN_TIME_CHECK_FOR_ENABLED_JIT
+        && allocation_profiling_needs_disabled_for_jit(unsafe { crate::PHP_VERSION_ID })
+        && *JIT_ENABLED
+    {
+        error!("Memory allocation profiling will be disabled as long as JIT is active. To enable allocation profiling disable JIT or upgrade PHP to at least version 8.1.21 or 8.2.8. See https://github.com/DataDog/dd-trace-php/pull/2088");
+        true
+    } else {
+        false
+    }
+}
+
 pub fn allocation_profiling_rinit() {
     let allocation_profiling: bool = REQUEST_LOCALS.with(|cell| {
         match cell.try_borrow() {
-            Ok(locals) => locals.profiling_allocation_enabled,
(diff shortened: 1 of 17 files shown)


morrisonlevi:
I've been manually testing this on different setups. So far I've tried all of these on PHP 8.3:

- Apache with mod_php 
- nginx + php-fpm
- built-in web cli server

No issues detected. I'm going to work on polishing this now.

profiling/src/allocation.rs
```diff
@@ -180,11 +183,9 @@ pub fn allocation_profiling_rinit() {
...
     // `is_zend_mm()` should be `false` now, as we installed our custom handlers
     if is_zend_mm() {
-        error!("Memory allocation profiling could not be enabled. Please feel free to fill an issue stating the PHP version and installed modules. Most likely the reason is your PHP binary was compiled with `ZEND_MM_CUSTOM` being disabled.");
-        REQUEST_LOCALS.with(|cell| {
-            let mut locals = cell.borrow_mut();
-            locals.profiling_allocation_enabled = false;
-        });
+        // Can't proceed with it being disabled, because that's a system-wide
+        // setting, not per-request.
+        panic!("Memory allocation profiling could not be enabled. Please feel free to fill an issue stating the PHP version and installed modules. Most likely the reason is your PHP binary was compiled with `ZEND_MM_CUSTOM` being disabled.");
```
realFlowControl, line 188: I have to say that I like this way more. It is very intrusive, but OTOH folks would not ask questions why they don't see allocation profiles in the backend.

realFlowControl (approved):
🎉

PR #2488, opened 2024-01-24
Reduce macrobenchmark flakiness

ddyurchenko:
### Description

Reduce flakiness of macrobenchmark.

Improves spread of p50 latency from 9.7ms to 0.02ms (by ~40x).

Downside is that all variants are run now sequentially on the same machine (which currently takes still under 30 minutes).

If we need to scale horizontally to reduce execution time, there should be done more investigation into flakiness on reruns (in particular, I would use same seed for [random data generation](https://github.com/PROFeNoM/laravel-realworld-example-app/blob/main/database/seeders/DummyDataSeeder.php#L29-L81) that happens on benchmark reruns, but totally possible that something more might be done there).

### Reviewer checklist
- [ ] Test coverage seems ok.


PROFeNoM (approved):
Thank you for looking into this 🙏  [Seeded](https://github.com/PROFeNoM/laravel-realworld-example-app/commit/4ece6545bf54f4812c267452e11bfdd761f8589f) the RNG on the app, as you recommended.

PR #2489, opened 2024-01-26
Provide server.response.body

cataphract:
### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/extension/commands/request_shutdown.c
```diff
@@ -189,3 +261,355 @@ static void _pack_headers_no_cookies_map(
...
+        }
+        if (zend_string_equals_literal_ci(key, "content-type")) {
+            if (Z_TYPE_P(val) == IS_STRING) {
+                *len = Z_STRLEN_P(val);
+                return Z_STRVAL_P(val);
+            }
+            if (Z_TYPE_P(val) == IS_ARRAY) {
+                zend_array *arr = Z_ARR_P(val);
+                HashPosition pos;
+                zend_hash_internal_pointer_end_ex(arr, &pos);
```
Anilm3, line 286: When is the content type an array and is the actual type always the last element?

appsec/src/extension/commands/request_shutdown.c
```diff
@@ -189,3 +261,355 @@ static void _pack_headers_no_cookies_map(
...
+    ZEND_HASH_FOREACH_END();
+
+    return NULL;
+}
+
+static zval _convert_json(char *nonnull entity, size_t entity_len)
+{
+    zval zv;
+    ZVAL_NULL(&zv);
+#define MAX_DEPTH 12
```
Anilm3, line 309: We did talk about using the same limits as the WAF so it might be good to start thinking about that as well...

appsec/src/extension/commands/request_shutdown.c
```diff
@@ -189,3 +261,355 @@ static void _pack_headers_no_cookies_map(
...
+        mlog(dd_log_debug, "Failed to set XML parser option");
+        zval_dtor(&function_name);
+        zval_dtor(&parser);
+        return null_zv;
+    }
+
+    /* skip whitespace */
+    ZVAL_LONG(&args[1], 4 /*PHP_XML_OPTION_SKIP_WHITE*/);
+    ZVAL_BOOL(&args[2], 1);
+    is_successful = call_user_function(
```
Anilm3, line 394: IIRC during opcache preload,  user functions can sometimes return garbage, do you think that would be a concern here?

appsec/src/extension/entity_body.c
```diff
@@ -0,0 +1,125 @@
...
+        zend_string_release(s);
+    }
+    ZEND_HASH_FOREACH_END();
+    zend_hash_destroy(&_all_buffers);
+}
+
+static typeof(zend_write(NULL, 0)) _dd_save_output_zend_write(
+    const char *str, size_t str_length)
+{
+    if (DDAPPSEC_G(enabled) && _buffer != NULL) {
```
Anilm3, line 54: Should it not be `DDAPPSEC_G(active)`?

appsec/src/extension/request_lifecycle.c
```diff
@@ -426,14 +433,146 @@ static zend_array *nullable _response_commit(
...
+    php_stream_from_zval_no_verify(stream, rbe_zv);
+    if (!stream) {
+        mlog(dd_log_debug,
+            "Response body entity is not a string or a stream; ignoring");
+        return NULL;
+    }
+
+    // TODO: support non-seekable streams. Needs replacing the stream
+    if (stream->flags & PHP_STREAM_FLAG_NO_SEEK) {
+        mlog(dd_log_info, "Response body entity is a stream, but it is "
```
Anilm3, line 486: Any idea when we'll have non-seekable streams? If common, setting this to info might not be a great idea.

appsec/src/extension/request_lifecycle.c
```diff
@@ -426,14 +433,146 @@ static zend_array *nullable _response_commit(
...
+
+    if (php_stream_seek(stream, start_pos, SEEK_SET) < 0) {
+        mlog(dd_log_error, "Failed to rewind response body entity stream; "
+                           "response stream is likely corrupted");
+        return NULL;
+    }
+
+    size_t effective_size = stream_size - (size_t)start_pos;
+    if (effective_size >= (size_t)get_DD_APPSEC_MAX_BODY_BUFF_SIZE()) {
+        mlog(dd_log_info,
```
Anilm3, line 528: This should probably be debug...

appsec/tests/extension/convert_xml_basic.phpt
```diff
@@ -0,0 +1,100 @@
...
+                        "my recipient"
+                    ],
+                    "attributes": {
+                        "attr": "x"
+                    }
+                }
+            },
+            {
+                "from": {
+                    "content": [
```
Anilm3, line 47: This layout won't do as the `content` key will shadow the `from` key and it won't allow scanners to properly function.  Something along these lines should preserve the `key:value` relationship

```
"from" [ "Jani", { "attributes": ...}]
```

appsec/src/extension/commands/request_shutdown.c
cataphract: It always be an array actually. In this case, the headers come from a "user request" (i.e., not the PHP lifecycle), and notify_commit should be given a map {header_name: [value1, ...], ...}.

The last element is taken, because while the rfc 2616 specifies that there should only be one value for content-type (it says only headers defined as comma separated values may have more than one value), I read somewhere that in the wild it's the last value that wins.

appsec/src/extension/commands/request_shutdown.c
cataphract: I'm calling C-implemented functions.

appsec/src/extension/commands/request_shutdown.c
cataphract: unfortunately, it seems that PHP's json_decode just gives up and returns nothing if the max depth is exceeded. So this should be fairly high.

appsec/src/extension/request_lifecycle.c
cataphract: at the end the stream code was not needed for roadrunner because we end up instrumenting the roadrunner specific class `\Spiral\RoadRunner\Http\HttpWorker`. That only only takes a string and a generator of strings (the latter in `respondStream`, for support of chunked responses).

PSR-7 does define an interface with streams, but its implementation in RoadRunner delegates to the instrumented class. I still think it's useful to have this here because there are many other implementations of PSR-17 that we may want to support in the future.

As for the log level, I can drop it, yes.

estringana (approved):

appsec/src/extension/commands/request_shutdown.c
```diff
@@ -189,3 +274,393 @@ static void _pack_headers_no_cookies_map(
...
+
+static zval _convert_json(char *nonnull entity, size_t entity_len)
+{
+    zval zv;
+    ZVAL_NULL(&zv);
+    if (!_json_decode_ex) {
+        return zv;
+    }
+
+#define MAX_DEPTH 30
```
Anilm3, line 326: Schema extraction has a depth limit of 18, while the scalar rules have one of 20, so 30 seems overkill.

appsec/src/extension/entity_body.c
Anilm3: You seem to have forgotten to change it...

appsec/src/extension/entity_body.c
```diff
@@ -0,0 +1,105 @@
...
+        size_t to_write = MIN(str_length, _buffer_size - _buffer->len);
+        memcpy(_buffer->val + _buffer->len, str, to_write);
+        _buffer->len += to_write;
+    }
+    return orig_zend_write(str, str_length);
+}
+
+void dd_entity_body_activate()
```
Anilm3, line 47: nit

```suggestion
void dd_entity_body_rinit()
```

appsec/src/extension/entity_body.c
```diff
@@ -0,0 +1,105 @@
...
+// (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+#include "entity_body.h"
+#include "ddappsec.h"
+#include "logging.h"
+#include <SAPI.h>
+#include <limits.h>
+
+static typeof(zend_write(NULL, 0)) _dd_save_output_zend_write(
+    const char *str, size_t str_length);
+
```
Anilm3, line 14: nit

```suggestion

// Response body buffer
```

Anilm3 (approved):
Looks good, although:
- There is one requested change (i.e. `DDAPPSEG_G(active)`) that wasn't applied.
- Max depth of 30 seems overkill.

appsec/src/extension/commands/request_shutdown.c
cataphract: I don't think it's overkill, because it will allow for more data to be partially sent to the waf. If the limit here is exceeded, no data will be sent.

Anilm3: Right, that's fair.

PR #2490, opened 2024-01-26
Enhance logging

bwoebi:
### Description

This adds span serialization, span creation and hook invocation logging as well as sidecar logging.

It also features two new config knobs: `DD_TRACE_LOG_FILE` and `DD_TRACE_LOG_LEVEL`.

If a log file is given (i.e. not stderr/syslog), sidecar logs will be included.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/components-rs/log.rs
+++ b/components-rs/log.rs
@@ -1,8 +1,14 @@
 use std::cell::RefCell;
 use std::collections::BTreeSet;
 use std::ffi::c_char;
-use std::slice;
+use std::fmt::Debug;
 use bitflags::bitflags;
+use tracing::Level;
+use tracing_core::{Event, Field, Subscriber};
+use tracing_subscriber::fmt::{FmtContext, FormatEvent, FormatFields};
+use tracing_subscriber::fmt::format::Writer;
+use tracing_subscriber::registry::LookupSpan;
+use tracing_subscriber::util::SubscriberInitExt;
 use ddcommon_ffi::CharSlice;
 use ddcommon_ffi::slice::AsBytes;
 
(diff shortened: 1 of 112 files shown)


pierotibou (commented):
I don't have time for a big review yet, so my first concern is about the performance impact. Do the benchmarks run with debug logs? If yes, they shouldn't and we should rerun the benchmarks without debug logs.

If they don't, I wonder if we have opportunities to avoid decreasing as much performances. Maybe we need a lot of: "if debugEnabled then ..."

ext/hook/uhook.c
```diff
@@ -238,6 +238,17 @@ static void dd_zend_interrupt_function(zend_execute_data *ex)
 }
 #endif
 
+void dd_uhook_log_invocation(void (*log)(const char *, ...), zend_execute_data *execute_data, const char *type, zend_object *closure) {
+    const zend_function *func = zend_get_closure_method_def(closure);
+    log("Running a %s hook function from %s:%d on %s %s%s%s",
```
PROFeNoM, line 243: Why are we using `log` here instead of `LOG`? Does it matter?

ext/logging.c
```diff
@@ -1,97 +1,163 @@
...
+    dd_log_set_level(get_global_DD_TRACE_DEBUG());
+}
+
+atomic_int ddtrace_error_log_fd = -1;
+atomic_uintmax_t dd_error_log_fd_rotated = 0;
 
-void ddtrace_bgs_log_minit(void) { atomic_store(&php_ini_error_log, (uintptr_t)NULL); }
+void ddtrace_log_minit(void) {
+    if (ZSTR_LEN(get_global_DD_TRACE_LOG_FILE())) {
+        int fd = VCWD_OPEN_MODE(ZSTR_VAL(get_global_DD_TRACE_LOG_FILE()), O_CREAT | O_RDWR | O_APPEND, 0666);
```
PROFeNoM, line 36: (Not necessarily related to this line) How annoying would it be to have a test using `DD_TRACE_LOG_FILE`?

ext/hook/uhook.c
bwoebi: the `log` variable has the log level included and is spawned into your scope when using LOGEV.

ext/logging.c
bwoebi: Not annoying :-)

ext/logging.c
```diff
@@ -1,97 +1,170 @@
...
+        ddog_set_error_log_level(once); // optimized handling without parsing
+    } else {
+        ddog_set_log_level(dd_zend_string_to_CharSlice(get_global_DD_TRACE_LOG_LEVEL()), once);
+    }
 }
 
-atomic_uintptr_t php_ini_error_log;
+void ddtrace_log_ginit(void) {
+    dd_log_set_level(get_global_DD_TRACE_DEBUG());
+}
```
PROFeNoM, line 32: Where are we calling this?

components-rs/log.rs
```diff
@@ -1,114 +1,189 @@
...
+        const Debug = 4;
+        const Trace = 5;
+        const Once = 1 << 3; // I.e. once per request
+        const _Deprecated = 3 | (1 << 4);
+        const Deprecated = 3 | (1 << 4) | (1 << 3) /* Once */;
+        const Startup = 3 | (2 << 4);
+        const Startup_Warn = 1 | (2 << 4);
+        const Span = 4 | (3 << 4);
+        const Span_Trace = 5 | (3 << 4);
+        const Hook_Trace = 5 | (4 << 4);
```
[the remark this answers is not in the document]
bwoebi: These are only error levels of error messages, not the error levels you can configure.

The default error level is Error. To turn all errors off you could even set it to Off.

ext/logging.c
bwoebi: To ensure the log level is initialized per thread (adding comment)

ext/logging.c
```diff
@@ -1,97 +1,171 @@
...
+        } else {
+            ddog_set_log_level(DDOG_CHARSLICE_C("debug,startup=error"), once);
+        }
+    } else if (runtime_config_first_init) {
+        ddog_set_log_level(dd_zend_string_to_CharSlice(get_DD_TRACE_LOG_LEVEL()), once);
+    } else if (zend_string_equals_literal_ci(Z_STR(zai_config_memoized_entries[DDTRACE_CONFIG_DD_TRACE_LOG_LEVEL].decoded_value), "error")) {
+        ddog_set_error_log_level(once); // optimized handling without parsing
+    } else {
+        ddog_set_log_level(dd_zend_string_to_CharSlice(get_global_DD_TRACE_LOG_LEVEL()), once);
+    }
```
PROFeNoM, line 27: So if I understand this correctly, the "least" amount of logs one can set is `error`, right?

bwoebi: If DD_TRACE_DEBUG is disabled, error is still present (current state). DD_TRACE_LOG_LEVEL=off would fully disable it.

ext/ddtrace.c
```diff
@@ -459,13 +459,89 @@ static PHP_GINIT_FUNCTION(ddtrace) {
...
+static pthread_key_t dd_cxa_thread_atexit_key; // fallback for sidecar
+
+// Note: this symbol is not public
+int __cxa_thread_atexit_impl(void (*func)(void *), void *obj, void *dso_symbol) {
+    while (glibc__cxa_thread_atexit_impl != CXA_THREAD_ATEXIT_PHP && glibc__cxa_thread_atexit_impl != CXA_THREAD_ATEXIT_UNAVAILABLE) {
+        if (glibc__cxa_thread_atexit_impl == CXA_THREAD_ATEXIT_UNINITIALIZED) {
+            glibc__cxa_thread_atexit_impl = DL_FETCH_SYMBOL(NULL, "__cxa_thread_atexit_impl");
+            if (glibc__cxa_thread_atexit_impl == NULL) {
+                glibc__cxa_thread_atexit_impl = CXA_THREAD_ATEXIT_UNAVAILABLE;
+                pthread_key_create(&dd_cxa_thread_atexit_key, NULL);
```
cataphract, line 521: any reason to use this rather than a thread-local variable?

ext/ddtrace.c
```diff
@@ -459,13 +459,89 @@ static PHP_GINIT_FUNCTION(ddtrace) {
...
+#define CXA_THREAD_ATEXIT_PHP ((void *)0)
+#define CXA_THREAD_ATEXIT_UNINITIALIZED ((void *)1)
+#define CXA_THREAD_ATEXIT_UNAVAILABLE ((void *)2)
+
+static int (*glibc__cxa_thread_atexit_impl)(void (*func)(void *), void *obj, void *dso_symbol) = CXA_THREAD_ATEXIT_UNINITIALIZED;
+static pthread_key_t dd_cxa_thread_atexit_key; // fallback for sidecar
+
+// Note: this symbol is not public
+int __cxa_thread_atexit_impl(void (*func)(void *), void *obj, void *dso_symbol) {
+    while (glibc__cxa_thread_atexit_impl != CXA_THREAD_ATEXIT_PHP && glibc__cxa_thread_atexit_impl != CXA_THREAD_ATEXIT_UNAVAILABLE) {
```
cataphract, line 516: not sure about this. glibc__cxa_thread_atexit_impl is a global variable that's read and written to from several threads without synchronization. the race doesn't seem benign if two threads enter the function with glibc__cxa_thread_atexit_impl == CXA_THREAD_ATEXIT_UNINITIALIZED at the same time (pthread_key_create is called twice, and one or both could use the first value written for dd_cxa_thread_atexit_key)

Also the loop seems to obscure more than to help readability. From what I can tell it would be the same to have

```c
if (glibc__cxa_thread_atexit_impl == CXA_THREAD_ATEXIT_UNINITIALIZED) {
  // set to symbol address or CXA_THREAD_ATEXIT_UNAVAILABLE
  if (glibc__cxa_thread_atexit_impl == CXA_THREAD_ATEXIT_UNAVAILABLE) {
    // call pthread_key_create
  }
}

if ((uintptr_t)glibc__cxa_thread_atexit_impl > 2) {
   // call the real thing
  return;
} 

if (glibc__cxa_thread_atexit_impl == CXA_THREAD_ATEXIT_UNAVAILABLE) {
   // pthread_setspecific ... 
}

// the rest
```

bwoebi: yes, but the function will be called the first time before any threads are forked. I may add a note about that.

ext/ddtrace.c
bwoebi: Stupid mistake. The point was having the destructor used, but obviously I set it to NULL :-)

ext/ddtrace.c
```diff
@@ -459,13 +459,89 @@ static PHP_GINIT_FUNCTION(ddtrace) {
...
+#ifdef CXA_THREAD_ATEXIT_WRAPPER
+struct dd_rust_thread_destructor {
+    void (*dtor)(void *);
+    void *obj;
+    struct dd_rust_thread_destructor *next;
+};
+ZEND_TLS struct dd_rust_thread_destructor *dd_rust_thread_destructors = NULL;
+ZEND_TLS bool dd_is_main_thread = false;
+
+void dd_run_rust_thread_destructors(void *unused) {
```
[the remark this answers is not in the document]
bwoebi: Yep.

ext/ddtrace.c
bwoebi: And agree, simplified now.

PROFeNoM (approved):
LGTM, as the performance degradations, as discussed internally, have no macro impact.

PR #2503, opened 2024-02-02
Change ip header parsing/header reporting

cataphract:
See [APPSEC-51370] [APPSEC-51531]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-51370]: https://datadoghq.atlassian.net/browse/APPSEC-51370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APPSEC-51531]: https://datadoghq.atlassian.net/browse/APPSEC-51531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appsec/src/extension/ip_extraction.c
```diff
@@ -239,27 +269,43 @@ static bool _parse_ip_address(
...
+        ipaddr cur;
+        bool succ =
+            _parse_ip_address_maybe_port_pair(value, end_cur - value, &cur);
         if (succ) {
-            succ = !_is_private(out);
+            if (!_is_private(&cur)) {
+                memcpy(out, &cur, sizeof *out);
+                return EXTRACT_SUCCESS_PUBLIC;
+            }
+            // else is private
```
estringana, line 290: I think the previous conditional is readable enough that this comment is not adding anything

appsec/src/extension/ip_extraction.c
```diff
@@ -17,6 +17,7 @@
 #include <arpa/inet.h>
```
estringana, line 17: We manually keep this script aligned with `ext/ip_extraction.c` so I think you should add this new header there  as well

estringana (changes requested):

appsec/src/extension/ip_extraction.c
cataphract: hm is there any reason to have this duplicated rather than in only one place?

Anilm3: IIRC when we implemented this we were still in separate repositories and APM needs to collect the client IP using the same strategy when ASM is disabled but `DD_TRACE_CLIENT_IP_ENABLED` is set to true. Feel free to change it to a common implementation.

appsec/cmake/ddtrace.cmake
```diff
@@ -0,0 +1,120 @@
+include(ExternalProject)
```
bwoebi, line 1: Can we please not have this?

I would like to not maintain that separately, but would prefer if there were a solution which just wraps around the "normal" phpize & configure & make process of ddtrace.

estringana (approved):

appsec/cmake/ddtrace.cmake
cataphract: I understand it, but this disabled by default and not used in any of the CI jobs. I don't want to impose it a maintenance burden on the tracer side.

It is only a debug/dev aid. I keep tripping on out-of-date builds of the tracer, plus I didn't find any way to builds with full debug symbols on the C and rust side, esp. on release builds. With this I can keep builds of several php versions without recompiling ddtrace all the time. I just need several directories where I configure cmake with a different php-config.

appsec/src/extension/ddtrace.c
```diff
@@ -89,6 +91,12 @@ static void dd_trace_load_symbols(void)
             dlerror()); // NOLINT(concurrency-mt-unsafe)
     }
 
+    _ddtrace_ip_extraction_find = dlsym(handle, "ddtrace_ip_extraction_find");
```
Anilm3, line 94: For the backlog but we're reaching the point where it would be useful to have some macros to do this and generate the relevant passthrough functions as well.

ext/ip_extraction.c
```diff
@@ -418,11 +458,11 @@ static bool dd_is_private_v6(const struct in6_addr *addr) {
     static const struct {
         union {
             struct in6_addr base;
-            uint64_t base_i[2];
+            unsigned __int128 base_i;
```
Anilm3, line 461: I believe @bwoebi had some concerns about the `__int128`:

https://github.com/DataDog/dd-trace-php/pull/1621#discussion_r903664382

ext/user_request.c
```diff
@@ -123,12 +123,17 @@ PHP_FUNCTION(DDTrace_UserRequest_notify_commit)
         }
     }
 
+    if (status < 100 || status > 599) {
+        zend_type_error("Status code must be between 100 and 599");
```
bwoebi, line 127: Can we have `zend_value_error` here and polyfill it in `compatibility.h` with `#define zend_value_error zend_type_error` - this was added with PHP 8 and existing "misuses" of TypeError replaced by it in php-src.

Anilm3 (approved):
If @bwoebi has no more concerns, lgtm

bwoebi (approved):
Nope, looks good to me too now :-)

Issue #2070, opened 2023-05-22
[Feature] profiling `ZTS` builds


sumerc:
Hi,

Is supporting `zts` builds something planned for the future, is there any issue for this? Sorry, if this is already discussed, but I can not seem to find an issue for that.

sumerc:
It was for profiling. Thanks for the info!

What is the main reason behind the complexity?. Maybe we might think to work on it if we understand the underlying problem?

morrisonlevi:
May I ask what you are using ZTS builds for? What libraries and such are you using?

Some of the complexity is that the PHP ABIs change for ZTS. The C macros that C extensions use don't always help since the profiler is primarily written in Rust. It's possible to support, I think, and is more of a matter of priority. Understanding what you are using ZTS builds for can help us understand the priority.

sumerc:
We are thinking of integrating continuous profiling for https://docs.platform.sh/languages/php.html.

And it currently uses ZTS.  I am wondering how hard this task may be?

piotrekkr:
I wanted to use [frankenphp](https://github.com/dunglas/frankenphp) as replacement for `php-fpm+nginx`. It is a ZTS build. Installer does not allow profiling. I liked how profiling worked with php-fpm and would be nice if it was available with ZTS.

```
$ docker build . -f Dockerfile-frankenphp                                            
...
------                                                                                                                                                                                                                                                                                                                      
 > [stage-0 4/4] RUN php /datadog-setup.php --php-bin=all --enable-profiling:                                                                                                                                                                                                                                               
0.536 Searching for available php binaries, this operation might take a while.                                                                                                                                                                                                                                              
0.666 Downloading installable archive from https://github.com/DataDog/dd-trace-php/releases/download/0.96.0/dd-library-php-0.96.0-x86_64-linux-gnu.tar.gz.                                                                                                                                                                  
0.666 This operation might take a while.                                                                                                                                                                                                                                                                                    
1.620 ...................                                                                                                                                                                                                                                                                                                   
8.662 Download completed
8.662 
12.51 Installed required source files to '/opt/datadog/dd-library/0.96.0'
12.57 Installing to binary: php (/usr/local/bin/php)
12.66 Copied '/tmp/dd-install/dd-library-php/trace/ext/20230831/ddtrace-zts.so' to '/usr/local/lib/php/extensions/no-debug-zts-20230831/ddtrace.so'
[...]

PR #2506, opened 2024-02-06
feat(profiling) allocation profiling for PHP ZTS

realFlowControl:
### Description

PROF-8922 / #2070 

This PR will bring allocation profiling to ZTS versions of PHP. In the "there is another custom handler installed" code path we are using an `UnsafeCell` for a thread local which is the cheapest way (evaluated at compile time and not runtime), as this is in the hot path (every allocation and free would borrow check at runtime).
We could argue that the case of another extension that has custom handlers installed is nearly non existing, so we could also be better safe then sorry, but OTOH I could not get it to crash.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation.rs
+++ b/profiling/src/allocation.rs
@@ -40,9 +40,10 @@ pub struct AllocationProfilingStats {
 }
 
 struct ZendMMState {
-    /// The heap installed in ZendMM at the time we install our custom handlers, this is also the
-    /// heap our custom handlers are installed in. We need this in case there is no custom handlers
-    /// installed prior to us, in order to forward our allocation calls to this heap.
+    /// The heap installed in ZendMM at the time we install our custom
+    /// handlers, this is also the heap our custom handlers are installed in.
+    /// We need this in case there is no custom handlers installed prior to us,
+    /// in order to forward our allocation calls to this heap.
     heap: Option<*mut zend::zend_mm_heap>,
     /// The engine's previous custom allocation function, if there is one.
     prev_custom_mm_alloc: Option<zend::VmMmCustomAllocFn>,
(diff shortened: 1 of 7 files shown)


realFlowControl:
@morrisonlevi So far I was not able to create crashes due to the`UnsafeCell`, but it feels wrong. Whats your opinion on it?

morrisonlevi (commented):
I believe `OnceCell` with `cell.get_or_init(|| unreachable!())` and re-creating the `OnceCell` on each request has good enough codegen to be able to use, and then we avoid the unsafe. Discussed this with Florian on Zoom and Slack.

profiling/src/allocation.rs
```diff
@@ -368,12 +368,18 @@ unsafe extern "C" fn alloc_profiling_malloc(len: size_t) -> *mut c_void {
     ptr
 }
 
+/// Safety: this function pointer is only allowed to point to `allocation_profiling_prev_alloc()`
+/// when at the same time the `ZEND_MM_STATE.prev_custom_mm_alloc` is initialised to a valid
+/// function pointer, otherwise there will be dragons.
 static mut ALLOCATION_PROFILING_ALLOC: unsafe fn(size_t) -> *mut c_void =
```
morrisonlevi, line 374: In ZTS this needs to be a thread-local, I think?

profiling/src/allocation.rs
```diff
@@ -145,47 +170,50 @@ pub fn allocation_profiling_rinit() {
...
-            ALLOCATION_PROFILING_PREPARE_ZEND_HEAP = prepare_zend_heap;
-            ALLOCATION_PROFILING_RESTORE_ZEND_HEAP = restore_zend_heap;
+
+        if !is_zend_mm() {
+            // Neighboring custom memory handlers found
+            debug!("Found another extension using the ZendMM custom handler hook");
+            unsafe {
+                zend::zend_mm_get_custom_handlers(
+                    // Safety: `unwrap()` is safe here, as `heap` is initialized just above
+                    (*zend_mm_state).heap.unwrap(),
```
morrisonlevi, line 185: Could probably just save the un-optioned heap from above and avoid the pointer deref + unwrap? Same goes for the `ddog_php_prof_zend_mm_set_custom_handlers` section below.

realFlowControl: Done, I also added more docs

profiling/src/allocation.rs
```diff
@@ -180,36 +178,34 @@ pub fn allocation_profiling_rinit() {
...
         let zend_mm_state = cell.get();
 
-        // Safety: `zend_mm_get_heap()` always returns a non null pointer to a valid heap structure
+        // Safety: `zend_mm_get_heap()` always returns a non-null pointer to a valid heap structure
         let heap = unsafe { Some(zend::zend_mm_get_heap()).unwrap() };
 
-        unsafe {
-            (*zend_mm_state).heap = Some(heap);
-        }
+        unsafe { ptr::addr_of_mut!((*zend_mm_state).heap).write(Some(heap)) };
```
realFlowControl: Just for my understanding: usafe of `ptr::addr_of_mut` removes the creation of an intermediate reference, is that the only reason why you use this?

profiling/src/allocation.rs
```diff
@@ -218,13 +214,13 @@ pub fn allocation_profiling_rinit() {
             zend::ddog_php_prof_zend_mm_set_custom_handlers(
                 heap,
                 Some(alloc_profiling_malloc),
-                Some(alloc_profiling_free),
-                Some(alloc_profiling_realloc),
+                Some(alloc_prof_free),
+                Some(alloc_prof_realloc),
```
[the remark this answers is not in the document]
realFlowControl: My bad, I was going through the commits on a per commit basis and you changed that later 😉

profiling/src/allocation.rs
morrisonlevi: Yes. I can't run this particular thing under MIRI, though, I so I can't be _sure_ it's necessary. Avoiding references to the thread-local means there's less chance we screw up the aliasing rules with our unsafe.

morrisonlevi (approved):

PR #2508, opened 2024-02-07
Have datadog.trace.db_client_split_by_instance also affect mysqli and sqlsrv

bwoebi:
### Description

Users rightfully expect the setting to apply everywhere, after all it's not called "pdo_split_by_instance".

Also apply a round of datadog.service_mapping before actually applying the split by domain, to allow distinguishing source services. E.g. mysqli can be mapped to webserver-mysqli, so that all connections from webservers go to the desired service, without having to individually map all possible target servers of the webserver.

### Reviewer checklist
- [ ] Test coverage seems ok.


PROFeNoM:
Should we refactor this piece of code to be in a `Util` or such? I guess that it could save maintenance, if any, down the road.

bwoebi:
Except that for sqlsrv, db_instance defaults to <default>, a concept which doesn't even exist for mysqli, which sort of is hardcoded logic, making it a bit more difficult to trivially extract.

src/Integrations/Integrations/PDO/PDOIntegration.php
```diff
@@ -316,6 +316,10 @@ public static function setCommonSpanInfo($source, SpanData $span)
         $span->meta[Tag::COMPONENT] = PDOIntegration::NAME;
         if (\DDTrace\Util\Runtime::getBoolIni("datadog.trace.db_client_split_by_instance")) {
             if (isset($storedConnectionInfo[Tag::TARGET_HOST])) {
+                $mapping = \dd_trace_env_config('DD_SERVICE_MAPPING');
+                if (isset($mapping[$span->service])) {
+                    $span->service = $mapping[$span->service];
+                }
```
PROFeNoM, line 322: The only thing bothering me about this piece of code is that, theoretically, it already does this by `Integration::handleInternalSpanServiceName($span, PDOIntegration::NAME);`. I believe this applies to the other integrations as well.

_Not sure that it's necessary_

bwoebi: Ah, interesting. It was changed in 0.96.0, I see :-)

bwoebi: Removed that part (and left the test), thanks.

PR #2511, opened 2024-02-08
perf(profiling): use an arena-based string table

morrisonlevi:
### Description

This updates libdatadog to use a branch which has an arena-based string table. See Datadog/libdatadog#306.

The arena capacity probably needs some tweaking, and should probably be set to a specific target set by the team after careful consideration. The memory is lazily populated by the OS, at least, so it's not automatically used (asking for 512 MiB won't allocate all 512 MiB immediately, it will grow up to that limit and then stop).

Part of [PROF-9123](https://datadoghq.atlassian.net/browse/PROF-9123).

#### Overhead

Using the [symfony/demo](https://github.com/symfony/demo) with PHP 8.3 using php-fpm, this netted me about 3-4% less total container memory. I expected there to be a difference because:

1. There is less memory allocator fragmentation going on as strings are variable size.
2. I'm packing the string size into a `u16`, so we save a few bytes per string compared to a `usize`.

I'm happy with 3-4% off the container, and don't feel a need to try to specifically calculate the reduction of _just_ libdatadog's memory.

I haven't attempted to measure CPU rigorously. Unscientifically, it looks on par with master. I would expect a very slight win with this PR because arena allocators are quick to allocate, and we can skip deallocation altogether (but deallocation is infrequent, roughly once per minute). Since eyeballing it looks on-par with master, which was the expectation, I won't dive into it deeper.

### Reviewer checklist
- [ ] Test coverage seems ok.

[PROF-9123]: https://datadoghq.atlassian.net/browse/PROF-9123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


morrisonlevi:
Superseded by #2668.

PR #2514, opened 2024-02-12
Record nested exceptions in the integrations

### LINKED ISSUE #2498: [Bug]: Nested exceptions are not recorded in `Integration::setError()`
### Bug report

Our application often catches and re-throws exceptions; recasting them with generic messages. The `$previous` argument is passed so that we maintain a chain of thrown exceptions. 

For example:

    try {
       $nestedException = new NestedException('A very detailed exception message');

        throw new CustomException('An error message', 456, $nestedException);
    } catch (\Throwable $e) {
        throw new \Exception('This is a generic exception message', 123, $e);
    }

The method `DDTrace\Integrations\Integration::setError()` currently records exception information in the following way.

    public function setError(SpanData $span, $exception)
    {
        $span->meta[Tag::ERROR_MSG] = $exception->getMessage();
        $span->meta[Tag::ERROR_TYPE] = get_class($exception);
        $span->meta[Tag::ERROR_STACK] = $exception->getTraceAsString();
    }

This logic prevents the nested exceptions from being recorded as there is no check of `$exception->getPrevious()`. The result of this is that only the 'parent' exception information is recorded in DataDog, meaning very generic and unhelpful error messages or traces. The real exception information of `$nestedException` is lost/not available.

To ensure there is specific recording of the `$nestedException`, we would need to modify `setError()` to recursively check `$e->getPrevious()` and build a chain of the thrown exceptions. If you could advise on how this data could be structured on `$span` and I can create a PR.

We are using the `SymfonyIntegration` (Symfony v5) if that is relevant.

### PHP version

8.1.0

0.97.0

### Installed extensions

_No response_

### Output of `phpinfo()`

_No response_

### Upgrading from

_No response_

PROFeNoM:
### Description

Fixes #2498

The [serialization process](https://github.com/DataDog/dd-trace-php/blob/fecdcfc44bd8c2c86b63e03e1c4775f5146e83e3/ext/serializer.c#L302-L340) should be able to handle nested exceptions recursively. To handle nested exceptions, directly assign the exception instance to the `exception` property of a span.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Thanks, looks good :-)

PR #2516, opened 2024-02-12
Mark config as immutable and interned to avoid refcounting race conditions

bwoebi:
### Description

Multiple threads manipulating the refcount at once is a possible source of race conditions (especially with many small requests increasing the likeness).

Properly mark the strings as interned / arrays as immutable to avoid this.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/zend_abstract_interface/config/config.c
+++ b/zend_abstract_interface/config/config.c
@@ -154,8 +154,15 @@ static void zai_config_intern_zval(zval *pzval) {
 #endif
     }
     if (Z_TYPE_P(pzval) == IS_ARRAY) {
+        Z_ADDREF_P(pzval);
         GC_ADD_FLAGS(Z_ARR_P(pzval), IS_ARRAY_IMMUTABLE);
-        Z_TYPE_FLAGS_P(pzval) = IS_ARRAY_IMMUTABLE;
+#if PHP_VERSION_ID < 70200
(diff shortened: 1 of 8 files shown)


zend_abstract_interface/config/config.c
```diff
@@ -140,11 +140,75 @@ void zai_config_mshutdown(void) {
...
-void zai_config_first_time_rinit(void) {
+#if PHP_VERSION_ID < 70300
+#define GC_ADD_FLAGS(c, flag) GC_FLAGS(c) |= flag
+#define GC_ADDREF(p) ++GC_REFCOUNT(p)
+#endif
+
+static void zai_config_intern_zval(zval *pzval) {
+    if (Z_TYPE_P(pzval) == IS_STRING) {
+#if PHP_VERSION_ID >= 70400
+        ZVAL_INTERNED_STR(pzval, zend_new_interned_string(Z_STR_P(pzval)));
```
morrisonlevi, line 151: During a request, `zend_new_interned_string` will store strings in an arena in `CG(interned_strings)` (well, at least on PHP 8.3). Is that long enough for the values being store during `zai_config_first_time_rinit`? I worry that it's not.

realFlowControl (approved):

zend_abstract_interface/config/config.c
morrisonlevi: Ah, I missed that. Thanks!

PR #2523, opened 2024-02-19
feat(profiling) add FrankenPHP to the list of recognised/supported SAPIs

realFlowControl:
### Description

PROF-9184

This PR aimes to add support for FrankenPHP SAPI, now that ZTS support exists.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
If we're going to support frankenphp in the profiler, the tracer probably should too.

I _think_ it's not too different from other sapis, so probably the tracer should just support it out of the box once added to components/sapi/sapi.c and dd_is_compatible_sapi in ddtrace.c.

morrisonlevi (approved):

PR #2524, opened 2024-02-19
Make PDO split-by coherent with other DB split-by with service name flattening

### LINKED ISSUE #2519: [Bug]: PDO/PHPRedis client split by instance inconsistency
# PDO/PHPRedis client split by instance inconsistency

The change https://github.com/DataDog/dd-trace-php/blob/346de4d66c7597ef7dc57cff7ceff8dc931c53b0/src/Integrations/Integrations/PDO/PDOIntegration.php#L319, which has been introduced in https://github.com/DataDog/dd-trace-php/pull/1498 is not behaving like expected by the [documentation](https://docs.datadoghq.com/tracing/trace_collection/library_config/php/) in conjunction with the env variable `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` being set to `true`.

Other split by instance integrations like for example phpredis, ignore this flattening rule.

## Environment

php: 8.2.15
dd-trace-php: 0.97.0

## Current behaviour

Call to a database host datadoghq.com from the service `my-test` has the service name `pdo-datadoghq.com` instead of the default service name of `pdo` or flattened service name `my-test-datadoghq.com` (when `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` is set to true).

## Expected behaviour

Independently of `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED`, the service name is `pdo-datadoghq.com`.

## Discussion points

- What is the expected impact of `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` on the split by instance feature on multiple integrations?
  1. Should `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` **have no effect** on the split by instance feature? Then the change introduced in https://github.com/DataDog/dd-trace-php/pull/1498 is not correct. Instead of a concatenation, it should overwrite the service name with the original integration name, same as done for the phpredis split by instance.
  2. Should `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` **have an effect** on the splitting of instances, then the phpredis integration at line https://github.com/DataDog/dd-trace-php/blob/master/src/Integrations/Integrations/PHPRedis/PHPRedisIntegration.php#L351, which has been introduced with https://github.com/DataDog/dd-trace-php/pull/2163, is not correct and needs to be fixed instead, by doing a concatenation instead of overwriting.
- Documentation for `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` is missing on the datadog docs. Only Ruby has mentioned this feature even though almost all SDKs have implemented it.

## Impact

Currently this causes misalignment across deployed services as for phpredis we are getting a global phpredis service, split by instances, but not for PDO even though we configured it to split by instance. As a result each deployed service has its own split by instance for pdo, which is kind of useless at the moment. On the other hand I see a use case for splitting by instance PER deployed service where we have multiple databases/redis instances, which is in favor of fixing the phpredis integration instead.

pablomartinezbernardo:
Split by should take priority in PDO integration over service name flattening

Fixes #2519.

### DIFF
--- a/src/Integrations/Integrations/Integration.php
+++ b/src/Integrations/Integrations/Integration.php
@@ -156,10 +156,10 @@ abstract class Integration
      * @param SpanData $span
      * @param string $fallbackName
      */
-    public static function handleInternalSpanServiceName(SpanData $span, $fallbackName)
+    public static function handleInternalSpanServiceName(SpanData $span, $fallbackName, $skipFlattening = false)
     {
         $flatServiceNames =
-            \PHP_MAJOR_VERSION > 5
+            !$skipFlattening && \PHP_MAJOR_VERSION > 5
                 && \dd_trace_env_config('DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED');
 
         if ($flatServiceNames) {
--- a/src/Integrations/Integrations/PDO/PDOIntegration.php
+++ b/src/Integrations/Integrations/PDO/PDOIntegration.php
@@ -317,8 +317,9 @@ class PDOIntegration extends Integration
             \DDTrace\Util\Runtime::getBoolIni("datadog.trace.db_client_split_by_instance") &&
                 isset($storedConnectionInfo[Tag::TARGET_HOST])
         ) {
-            $span->service = PDOIntegration::NAME . '-' .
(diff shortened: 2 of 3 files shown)


bwoebi (approved):
Thanks, looks good to me :-)

PR #2526, opened 2024-02-20
Change CI Workdir from `datadog` to `app`

PROFeNoM:
### Description

In the dev environment, we use `/home/circleci/app,` while on the CI, it is `/home/circleci/datadog.` While this is nothing spectacular, it isn't very pleasant when generating snapshots that include an absolute path, as these aren't the same strings.

So far, I don't see any particular reason for this difference, and it is easier to change it on the CI side rather than rebuilding all the docker images :)

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
While it has been a frequent annoyance to have that difference, it also has spotted times where the paths were hardcoded. Like people running testsuites outside of our container. Then CI fails.

bwoebi:
Regarding the snapshots, I would appreciate if it were possible to replace some fixed expressions with a placeholder upon snapshot generation?

PROFeNoM:
Although this doesn't exist on the test agent, this should be a quick addition (+ it is easier than doing some fancy regex stuff). I'll look into it.

I agree that we could do this in the Makefile, or with a dedicated `.sh` for regenerating snapshots... but I think it's better for it to be on the test-agent-side

PR #2527, opened 2024-02-20
Cleanup deferred loader and remove some PHP 5 compat code

bwoebi:
### Description

Drops configuration.php as well.

Loading Integrations directly from the classes. This allows the classes being autoloaded as well. (Still using the generated bridge, but we should autoload instead.)

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/configuration.h
```diff
@@ -95,7 +95,6 @@ enum ddtrace_sampling_rules_format {
     CONFIG(BOOL, DD_TRACE_URL_AS_RESOURCE_NAMES_ENABLED, "true")                                               \
     CONFIG(BOOL, DD_HTTP_SERVER_ROUTE_BASED_NAMING, "true")                                                    \
     CONFIG(SET, DD_INTEGRATIONS_DISABLED, "default")                                                           \
-    CONFIG(BOOL, DD_PRIORITY_SAMPLING, "true")                                                                 \
```
PROFeNoM, line 98: Why are we removing this config? Was it only used with PHP5??

bwoebi: Yes, it's completely unused.

bridge/_files_tracer_api.php
```diff
@@ -19,5 +19,14 @@
     __DIR__ . '/../src/api/Exceptions/InvalidSpanOption.php',
     __DIR__ . '/../src/api/GlobalTracer.php',
     __DIR__ . '/../src/api/Format.php',
+    // TODO check if we can fully remove logger or replace it by something vastly simpler. Doesn't need to be part of API?
```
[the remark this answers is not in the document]
bwoebi: Vastly simpler in this case means for example a function to call into the extension which would just forward with log level and message to the logging system of the extension. A problem is for example that PHP logs aren't aware of the log level nor the log file.

PROFeNoM (approved):
After looking at the native profiles of the WordPress benchmarks, they look good. Something is wrong and will be fixed in a later PR. 

Otherwise, LGTM, and the memory improvements are great, well done 😃

PR #2535, opened 2024-02-22
Skip the flaky unix socket test on CI environments

PROFeNoM:
### Description

This test is just too flaky. Retries don't seem effective, and the issue still persists. This issue may arise because of a connection issue with the request replayer. 

It is okay for us to mark this test as XFAIL, considering that we know it will sometimes fail for **external** reasons, and we seemingly have no actionable steps to take when this test is failing.

image

This time, it succeeded eight times in a row... This time will be the right one.

### Reviewer checklist
- [ ] Test coverage seems okay.

### DIFF
--- a/tests/ext/background-sender/default_unix_domain_socket_agent.phpt
+++ b/tests/ext/background-sender/default_unix_domain_socket_agent.phpt
@@ -1,11 +1,10 @@
 --TEST--
 If an agent unix domain socket exists it will try to connect to it
 --SKIPIF--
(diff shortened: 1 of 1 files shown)


bwoebi:
I have a maybe better idea: avoid running it in CI, but run it locally? At least I'm not seeing it fail locally.

PROFeNoM:
That's a good idea (although for me, it either consistently passes or consistently fail locally)

bwoebi (approved):

PR #2537, opened 2024-02-23
Use a global version and change installer tests to use the version

bwoebi:
### Description

If installer tests do not use the current version, changes to the directory structure in the final artifact may become incompatible with the datadog-setup used.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (commented):
I don't love that I get v0.0.0 in the profiler when building from source, but it's okay. I checked that it builds from source still, and the profiler version in the metadata from CircleCI artifacts is correct. I haven't reviewed the tracer nor testing bits.

bwoebi:
Hm @morrisonlevi when you build it from source (assuming the full repo is checked out, it should still find `../VERSION`, doesn't it?

datadog-setup.php
```diff
@@ -1413,12 +1413,30 @@ function download($url, $destination)
...
+        // But this only really matters for CircleCI artifacts, so not too bad.
+        if ($data == "") {
+            foreach ($http_response_header as $header) {
+                if (stripos($header, "location: ") === 0) {
+                    $data = file_get_contents(substr($header, 10));
+                    goto got_data;
+                }
+            }
+            if (PHP_VERSION_ID < 70400) {
+                goto next_method; // location redirects are may not be read on 7.3 and older
```
Anilm3, line 1428:
```suggestion
                goto next_method; // location redirects may not be read on 7.3 and older
```

Anilm3 (approved):
lgtm

PR #2541, opened 2024-02-27
Avoid remapping freed memory in trace sender fallback

bwoebi:
### Description

Fix faulty logic reusing a freed shm handle when sending fails.

Not sure how to test that as the precondition never should fail ... in theory.

### Reviewer checklist
- [ ] Test coverage seems ok.


PROFeNoM (approved):
No tests 😝

PR #2542, opened 2024-02-28
Raise API errors instead of debug logs when wrong arguments are passed

PROFeNoM:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/hook/uhook_legacy.c
+++ b/ext/hook/uhook_legacy.c
@@ -314,7 +314,7 @@ static void dd_uhook(INTERNAL_FUNCTION_PARAMETERS, bool tracing, bool method) {
         Z_PARAM_STR(method_name)
         Z_PARAM_OPTIONAL
 
-        if (_i == ZEND_NUM_ARGS()) {
+        if (_i >= 0 && (uint32_t)_i == ZEND_NUM_ARGS()) {
             break;
         }
 
--- a/ext/hook/uhook_legacy.c
+++ b/ext/hook/uhook_legacy.c
@@ -314,7 +314,7 @@ static void dd_uhook(INTERNAL_FUNCTION_PARAMETERS, bool tracing, bool method) {
         Z_PARAM_STR(method_name)
         Z_PARAM_OPTIONAL
 
-        if ((ZEND_NUM_ARGS() <= (uint32_t)(1 + method)) {
+        if (ZEND_NUM_ARGS() <= (uint32_t)(1 + method)) {
             break;
         }
 
--- a/ext/hook/uhook_legacy.c
+++ b/ext/hook/uhook_legacy.c
@@ -313,21 +313,15 @@ static void dd_uhook(INTERNAL_FUNCTION_PARAMETERS, bool tracing, bool method) {
         }
         Z_PARAM_STR(method_name)
         Z_PARAM_OPTIONAL
-        if (!tracing) {
+        zval *_current_arg = _arg + _i - method;
+        if (Z_TYPE_P(_current_arg) == IS_ARRAY) {
+            Z_PARAM_ARRAY(config_array)
+        } else if (!tracing) {
             Z_PARAM_OBJECT_OF_CLASS_EX(prehook, zend_ce_closure, 1, 0)
         }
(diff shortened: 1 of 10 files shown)


ext/hook/uhook_legacy.c
```diff
@@ -306,36 +306,27 @@ static void dd_uhook(INTERNAL_FUNCTION_PARAMETERS, bool tracing, bool method) {
...
         if (method) {
             Z_PARAM_STR(class_name)
         }
         Z_PARAM_STR(method_name)
         Z_PARAM_OPTIONAL
-        if (!tracing) {
+
+        if (_i >= 0 && (uint32_t)_i == ZEND_NUM_ARGS()) {
+            break;
+        }
```
bwoebi, line 319: Why are we checking _i >= 0? Also maybe just:
```suggestion
        Z_PARAM_PROLOGUE(0, 0)
```

ext/hook/uhook_legacy.c
```diff
@@ -306,36 +306,27 @@ static void dd_uhook(INTERNAL_FUNCTION_PARAMETERS, bool tracing, bool method) {
...
+    ZEND_PARSE_PARAMETERS_START(1 + method, 2 + method + !tracing)
         // clang-format off
         if (method) {
             Z_PARAM_STR(class_name)
         }
         Z_PARAM_STR(method_name)
         Z_PARAM_OPTIONAL
-        if (!tracing) {
+
+        if (_i >= 0 && (uint32_t)_i == ZEND_NUM_ARGS()) {
```
bwoebi, line 317: Why are we checking _i >= 0? Also maybe just:
```suggestion
        if (ZEND_NUM_ARGS() <= 1 + method)
```

ext/hook/uhook_legacy.c
```diff
@@ -306,36 +306,27 @@ static void dd_uhook(INTERNAL_FUNCTION_PARAMETERS, bool tracing, bool method) {
...
+            break;
+        }
+
+        zval *_current_arg = _arg + _i - method;
```
bwoebi, line 321: I think that's wrong?
```suggestion
        zval *_current_arg = _arg + _i;
```

PROFeNoM: Long story short: Yes, it was, but the `_i - method` would always be `1`, so it appeared to have fixed my issue in the first place, while this was just luck. I didn't initially realize `_arg` was being incremented on the fly.

Therefore, I believe `_arg + 1` should represent the current arg. At this point in the code, `_arg` is a string (the method name).

The check done right before is used because, at this point, the args are optional (although I don't understand why it makes any sense to be able to set up a hook without any callbacks associated with it...). Therefore, it mayn't make sense to check for the current arg (since it doesn't exist).

Honestly, I wouldn't say I like this code section, and I find it pretty ugly. If it works, I guess that's what ultimately matters, but it doesn't feel idiomatic at all :/

tests/ext/wrong-parameter-errors/DDTrace_hook_function.phpt
```diff
@@ -1,5 +1,7 @@
 --TEST--
 DDTrace_hook_function is passed wrong parameters
+--SKIPIF--
+<?php if (PHP_VERSION_ID < 70100) die('skip: A TypeError is thrown in PHP 7.0 when an invalid number of arguments is passed to a function in strict mode.')
```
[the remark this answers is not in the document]
PROFeNoM: I don't think we can use `class_alias` on `TypeError` (or anything "PHP-made" for that matter) :grimacing:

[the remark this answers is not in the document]
PROFeNoM: I mean, what I can do is just change the code being run in PHP 7.0 to check for TypeError instead of ArgumentCountError...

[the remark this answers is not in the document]
bwoebi: Oh, you are right, class_alias on internal classes only works since PHP 8.3. Then nevermind :-)

bwoebi (approved):
Looks good, thanks. I'd say the tests are unncessary, given that it's testing core PHP functionality.
However, they're there and we can drop them if there's an issue.

PR #2549, opened 2024-02-29
w3c phase 2: add last parent_id to tracestate

mabdinur:
### Description

With this change span_id will be encoded in both the tracestate and traceparent headers. This will allow us to re-connect datadog traces that contain non-datadog spans.

Ex: Host A: Datadog Spans ->tracecontext-> Host B: Otel Dynatrace Spans -->tracecontext-> Host C: Otel Datadog Spans
- By adding span_id to the tracestate Host C will be able to detect upstream Datadog spans from Host A. This span_id will be sent to Datadog intake services via `_dd.parent_id` tag. The Datadog internal services will use this tag to reconnect visualize spans.

Tests will be added in this PR: https://github.com/DataDog/system-tests/pull/2181

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/handlers_http.h
```diff
@@ -180,6 +180,7 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
 
                 zend_string *full_tracestate = ddtrace_format_tracestate(tracestate, origin, sampling_priority, propagated_tags, tracestate_unknown_dd_keys);
                 if (full_tracestate) {
+                    // Replace dd=.... with dd=p:<zero padded span_id>;.....
```
mabdinur, line 183: Here I am avoiding updating  `ddtrace_format_tracestate`. We only NEED to add `p:........` in the tracestate when generating distributed tracing headers (ex: we do not need this field in span_links).

ext/handlers_http.h
```diff
@@ -178,7 +185,15 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
-                zend_string *full_tracestate = ddtrace_format_tracestate(tracestate, origin, sampling_priority, propagated_tags, tracestate_unknown_dd_keys);
+                uint64_t propagated_span_id = 0;
+                zval *old_parent_id;
+                if (root) {
+                    propagated_span_id = span_id;
+                } else if ((old_parent_id = zend_hash_str_find(&DDTRACE_G(root_span_tags_preset), ZEND_STRL("_dd.parent_id")))) {
+                    propagated_span_id = ddtrace_parse_hex_span_id(old_parent_id);
+                }
+
+                zend_string *full_tracestate = ddtrace_format_tracestate(tracestate, propagated_span_id, origin, sampling_priority, propagated_tags, tracestate_unknown_dd_keys);
```
mabdinur, line 196: `p:<span_id_hex>;` should be set using the current `span_id`.

For example if the php span the following traceparent `00-0000000000000000b5a2814f70060771-7197677932a62370-01` then it's tracestate should contain `p:7197677932a62370`. 

```suggestion
                zend_string *full_tracestate = ddtrace_format_tracestate(tracestate, span_id, origin, sampling_priority, propagated_tags, tracestate_unknown_dd_keys);
```

ext/ddtrace.c
```diff
@@ -642,7 +642,7 @@ ZEND_METHOD(DDTrace_SpanLink, fromHeaders) {
     zend_hash_copy(Z_ARR(link->property_attributes), &result.meta_tags, NULL);
 
     zend_string *propagated_tags = ddtrace_format_propagated_tags(&result.propagated_tags, &result.meta_tags);
-    zend_string *full_tracestate = ddtrace_format_tracestate(result.tracestate, result.origin, result.priority_sampling, propagated_tags, &result.tracestate_unknown_dd_keys);
+    zend_string *full_tracestate = ddtrace_format_tracestate(result.tracestate, 0, result.origin, result.priority_sampling, propagated_tags, &result.tracestate_unknown_dd_keys);
```
mabdinur, line 645: Setting a value of zero here might be misleading. 

I think we have two options here:
  1) Encode the current_span_id in the tracestate. Storing the hex encoding of a span's span_id in the tracestate is redundant but it maintains consistency with the tracestate generated by distributing tracing headers.
  2) We can avoid setting `p:......` field in the link.tracestate. This will cause the tracestate generated by span link and distributed tracing headers to diverge.

I am leaning towards option 1. As long as tracestate headers generated by the datadog and otel api contain the `p:.....` field then we should be in good shape. We do not NEED to set this field in span links but I think it is best to be consistent.

```suggestion
    zend_string *full_tracestate = ddtrace_format_tracestate(result.tracestate, result.span_id, result.origin, result.priority_sampling, propagated_tags, &result.tracestate_unknown_dd_keys);
```

ext/handlers_http.h
bwoebi: But then the traces will be disconnected, defeating the point of the change in that scenario?

This particular case handles the scenario where a distributed trace is injected, but _no_ active span exists on the local side.

Let's say the injected trace has a trace parent span `1234567890123456` (which is also the locally active span id), and a tracestate `p` of `abcdefabcdefabcd`. If we blindly inject `p:1234567890123456`. Even though a span id `1234567890123456` may never reach datadog.

ext/ddtrace.c
bwoebi: I do not have a strong opinion here and defaulted to the approach doing less computations. If you prefer 2) we can do that.

mabdinur: I would prefer 2 over the current approach. In this scenario no value is better than an incorrect one

ext/handlers_http.h
mabdinur: Ahhh I was missing this context. This change looks good to me

bouwkast: I'm trying to make sure that I understand this correctly (not super familiar with the PHP tracer) I'm not sure if I fully understand the `else if`:

Essentially if we have an active span we'll inject that span ID as `p`, which makes sense.

But if we are only injecting and propagating context that is what this is attempting to handle correct?
And if so, it'll check for `_dd.parent_id` on the root span and use that value or it'll propagate `0`. I don't think I covered this specific scenario in the spec so just want to make sure I understand what is happening as it may require updating the spec.

bwoebi: Yes, that's right. (not quite "on the root span", but the tags which seed the root span whenever one is created, i.e. amongst other things, the extracted tags.)

mabdinur:
We will need to ignore/filter out the `p` value in snapshot tests

bwoebi:
@mabdinur Yep, needs a change in the test-agent to filter them there.

tests/OpenTelemetry/Integration/API/TracerTest.php
```diff
@@ -806,22 +806,25 @@ public function getDescription(): string
...
             $this->assertEquals(1, $childContext->getTraceFlags()); // RECORD_AND_SAMPLED ==> 01 (AlwaysOn sampler)
-            //$this->assertSame("dd=t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$childContext->getTraceState());
+            //$this->assertSame("dd=p:[0-9a-f]{16};t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$childContext->getTraceState());
 
             $tracer = self::getTracer();
             $grandChild = $tracer->spanBuilder("grandChild")
                 ->setParent(Context::getCurrent()->withContextValue($child))
                 ->startSpan();
-            $this->assertSame("dd=t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$child->getContext()->getTraceState());
+            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $child->getSpanId())
```
mabdinur, line 815:
```suggestion
            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $child->getSpanId());
```

tests/OpenTelemetry/Integration/API/TracerTest.php
```diff
@@ -806,22 +806,25 @@ public function getDescription(): string
...
             $this->assertSame($remoteContext->getTraceId(), $grandChildContext->getTraceId());
             $this->assertSame($child->getContext()->getSpanId(), $grandChild->getParentContext()->getSpanId());
             $this->assertFalse($grandChildContext->isRemote()); // "When creating children from remote spans, their IsRemote flag MUST be set to false."
             $this->assertEquals(1, $grandChildContext->getTraceFlags()); // RECORD_AND_SAMPLED ==> 01 (AlwaysOn sampler)
-            $this->assertSame("dd=t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$child->getContext()->getTraceState());
-            $this->assertSame("dd=t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$grandChildContext->getTraceState());
+            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $child->getSpanId())
+            $this->assertSame($expected_tracestate, (string)$child->getContext()->getTraceState());
+            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $grandChildContext->getSpanId())
```
mabdinur, line 826:
```suggestion
            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $grandChildContext->getSpanId());
```

tests/OpenTelemetry/Integration/API/TracerTest.php
```diff
@@ -806,22 +806,25 @@ public function getDescription(): string
...
             $grandChildScope = $grandChild->activate();
 
             $grandChildContext = $grandChild->getContext();
             $this->assertSame($remoteContext->getTraceId(), $grandChildContext->getTraceId());
             $this->assertSame($child->getContext()->getSpanId(), $grandChild->getParentContext()->getSpanId());
             $this->assertFalse($grandChildContext->isRemote()); // "When creating children from remote spans, their IsRemote flag MUST be set to false."
             $this->assertEquals(1, $grandChildContext->getTraceFlags()); // RECORD_AND_SAMPLED ==> 01 (AlwaysOn sampler)
-            $this->assertSame("dd=t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$child->getContext()->getTraceState());
-            $this->assertSame("dd=t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", (string)$grandChildContext->getTraceState());
+            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $child->getSpanId())
```
mabdinur, line 824:
```suggestion
            $expected_tracestate =sprintf("dd=p:%016X;t.dm:-0,remoteparent=yes,rojo=00f067aa0ba902b7,congo=t61rcWkgMzE", $child->getSpanId());
```

tests/OpenTelemetry/Integration/InteroperabilityTest.php
```diff
@@ -646,7 +646,8 @@ public function testW3CInteroperability()
             $OTelRootSpan->end();
 
             $this->assertSame("00-ff0000000000051791e0000000000041-$DDChildSpanId-01", $carrier[TraceContextPropagator::TRACEPARENT]);
-            $this->assertSame('dd=t.dm:-0', $carrier[TraceContextPropagator::TRACESTATE]); // ff00000000000517 is the high 64-bit part of the 128-bit trace id
+            $expected_tracestate = sprintf('dd=p:%016X;t.dm:-0', $DDChildSpanId)
```
mabdinur, line 649:
```suggestion
            $expected_tracestate = sprintf('dd=p:%016X;t.dm:-0', $DDChildSpanId);
```

mabdinur:
@bwoebi thanks for fixing the failing tests and updating the OpenTelemetry SpanContext. A few parametric tests are failing due to the changes we made to tracestate. Once those failures are addressed we can merge this PR 🥳

PROFeNoM (approved):
LGTM 🚢

PR #2550, opened 2024-03-01
fix: Orphans not being sampled-out

PROFeNoM:
### Description

:rotating_light: **Requires #2556** :rotating_light:

**What's the issue?**

While doing some tests locally on my Laravel App, some orphan spans wouldn't get sampled away while they were in previous tracer versions.

**What's happening?**

In real life (i.e., not in our tests), the default fallback rule `(service: , env: )` would get evaluated. This rule exists because of the default agent sampling rate of 10 traces per second. Most notably, when the rate per service would become `0`, then `\DDTrace\get_priority_sampling()` would evaluate to zero, and we wouldn't set the priority sampling to auto reject again.

**What's the fix?**

The fix adds the check `|| $prioritySampling == DD_TRACE_PRIORITY_SAMPLING_AUTO_REJECT` to the rejection condition. 

While still not entirely realistic, the tests aim to assess that the code used to handle orphans works across values of agent sampling that would be evaluated through the default fallback rule.

--- 

Additionally, there were two additional `mysqli_query` spans in [one test](https://circleci.com/gh/DataDog/dd-trace-php/3914710) from this PR. Let's just test what we want actually to test.

### Reviewer checklist
- [ ] Test coverage seems ok.


src/Integrations/Util/Common.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace DDTrace\Util;
+
+use DDTrace\RootSpanData;
+use DDTrace\SpanData;
+
+class Common
+{
+    // When modifying this method, please also pay attention to the tests/ext/orphans.phpt test and update it accordingly if needed.
```
bwoebi, line 10: Is there a reason why we cannot test that as part of an Integration test? I'd rather avoid duplicating logic like this.
It also doesn't really fit in tests/ext as it's not testing the extension either, but testing how this particular code here interacts with functionality the extension provides.

PROFeNoM: I tried reaching out to the `/next-response` endpoint of the Request Replayer from TracerTestTrait, but it wouldn't work for w/e reason. I'm tempted to just use [the capabilities offered by the test agent](https://github.com/DataDog/dd-apm-test-agent/tree/master?tab=readme-ov-file#optional-agent_sample_rate_by_service) instead tbh

bwoebi: Feel free to use that if it works :-)

bwoebi (approved):
looks good to me now, thanks :-)

PROFeNoM:
The coverage of [these files](https://app.codecov.io/gh/DataDog/dd-trace-php/pull/2550/indirect-changes) is not supposed to be removed. I'll re-run the pipeline in case there has been something wrong during the coverage processing. If it was just a one-time event, I'll remove the do-not-merge label, else I'll search for the issue.

PR #2553, opened 2024-03-04
Get extensions and packages versions to CI artifacts

pablomartinezbernardo:
### Description
Use `get_loaded_extensions()` and `composer show` in CI to write dependency versions to a CSV as an artifact in CI, so it can be later retrieved to populate support data in metabase.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):

Makefile
```diff
@@ -972,7 +972,9 @@ define run_composer_with_retry
 	for i in $$(seq 1 $(MAX_RETRIES)); do \
 		echo "Attempting composer update (attempt $$i of $(MAX_RETRIES))..."; \
 		$(COMPOSER) --working-dir=$1 update $2 && break || (echo "Retry $$i failed, waiting 5 seconds before next attempt..." && sleep 5); \
-	done
+	done \
+
+	$(COMPOSER) --working-dir=$1 show -f json -D | jq -r '.installed[] | "\(.name);\(.version)"' >> "/tmp/artifacts/web_versions.csv"
```
PROFeNoM, line 977: Do we have to use `jq`? Could this be done using basic shell commands?

As is, it would work on the CI, but not locally. We _could_ install jq in all of our buster containers, but they would have to all be rebuilt (which is a quite long and annoying task).

Additionally, I believe the `/tmp/artifacts` has to be created beforehand (e.g., mkdir -p). It exists on the CI because the volume is mounted, but not locally. Therefore writing to web_versions.csv wouldn't work (since  `artifacts` doesn't exists)

PROFeNoM: Testing locally, I believe this should do the trick:

```suggestion
	$(COMPOSER) --working-dir=$1 show -f json -D | grep -o '"name": "[^"]*\|"version": "[^"]*' | paste -d';' - - | sed 's/"name": //; s/"version": //' | tr -d '"' >> "/tmp/artifacts/web_versions.csv"
```

But you may have a better suggestion; I'm no bash expert, lol

bwoebi: I ... umm.. find the jq version much more readable :-D

PROFeNoM: Then I say it: I won't be the one re-building the containers solely for jq :p

pablomartinezbernardo: Removed the dependency to jq and added a comment trying to explain. I don't know what rebuilding the containers involves, but if Alexandre doesn't want to do it then I think I don't either 😄

PR #2557, opened 2024-03-05
Add DD_APPSEC_SCA_ENABLED new configuration variable

estringana:
### Description

It is required to create a new configuration variable `DD_APPSEC_SCA_ENABLED` so customers can enable SCA. This variable is reported to the backend via telemetry and used there.

### Reviewer checklist
- [ ] Test coverage seems ok.

APPSEC-14721


Anilm3 (commented):
Are there perhaps any tests showing that this variable is actually being sent?

bwoebi:
I think you should test `datadog.appsec.sca_enabled=1` in some of the tests instead of using the ENV there. I believe it needs to be handled in configuration.c so that it's parsed as `datadog.appsec.sca_enabled` instead of `datadog.appsec_sca_enabled`.

estringana:
> I think you should test `datadog.appsec.sca_enabled=1` in one of the tests instead of using the ENV there. I believe it needs to be handled in configuration.c so that it's parsed as `datadog.appsec.sca_enabled` instead of `datadog.appsec_sca_enabled`.

Thanks for pointing that out @bwoebi . If I understood correctly, I fixed it

ext/configuration.c
```diff
@@ -145,6 +145,10 @@ static void dd_ini_env_to_ini_name(const zai_str env_name, zai_config_name *ini_
         memcpy(ini_name->ptr, "datadog.", sizeof("datadog.") - 1);
         ini_name->len = env_name.len + DD_TO_DATADOG_INC;
 
+        if (env_name.ptr == strstr(env_name.ptr, "DD_APPSEC_")) {
+            ini_name->ptr[sizeof("datadog.appsec") - 1] = '.';
+        }
+
         if (env_name.ptr == strstr(env_name.ptr, "DD_TRACE_")) {
             ini_name->ptr[sizeof("datadog.trace") - 1] = '.';
         }
```
bwoebi, line 154: Can we reorder this and put an else if there? Just to avoid eval of strstr when unnecessary.

bwoebi (approved):
Overall LGTM :-)

ext/configuration.c
estringana: Makes sense. Done

PR #2562, opened 2024-03-11
ci(profiling) add tests for exception profiling in ZTS builds

realFlowControl:
### Description

This PR will add some tests for exception profiling in ZTS PHP builds:
- prof correctness test
- PHPT tests
- rebuild Buster and CentOS images to have `ext-parallel`

PROF-8923 / #2070 

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/src/exception.rs
+++ b/profiling/src/exception.rs
@@ -9,10 +9,14 @@ use std::sync::atomic::Ordering;
 
 use rand_distr::{Distribution, Poisson};
 
-/// The engine's previous throw exception hook
+/// The engine's previous throw exception hook.
+/// We need to occupy the `zend_throw_exception_hook` in MINIT which is before threads get started
+/// (in ZTS), so we do not need a thread local for this function pointer.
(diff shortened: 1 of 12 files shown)


profiling/src/exception.rs
```diff
@@ -102,9 +106,18 @@ pub fn exception_profiling_minit() {
...
+    let exception_profiling = REQUEST_LOCALS.with(|cell| {
+        cell.try_borrow()
+            .map(|locals| locals.system_settings().profiling_exception_enabled)
+            .unwrap_or(false)
+    });
+
+    if !exception_profiling {
+        return;
+    }
+
```
realFlowControl, line 120: As of recently the INI is a true INI_SYSTEM, which means it can not be changed anymore (not per user dir or any other fancy things), which is also intended behaviour. With this change, we also do not show the "Exception profiling initialized .." message anymore in case it is disabled

morrisonlevi (approved):

PR #2570, opened 2024-03-12
style(profiling): add eg! and sg! macros

morrisonlevi:
### Description

Adds the `eg!` and `sg!` macros that mirror `EG` and `SG` in PHP internals. Works for both NTS and ZTS.

I'm not sure yet if this is better. Opening the PR to see if it works across versions (worked for PHP 8.3 locally).

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi:
It seems not worth it. The `EG` macro (and some others) changed from TSRM* to TSRM_FAST* in PHP 7.4, and handling all these differences when we can do it in 3 simple lines of C and 3 more in Rust doesn't seem worth it. Especially when the cost of getting it wrong is sigsegvs.

PR #2597, opened 2024-03-26
Fix the ExecIntegration name

bwoebi:
For some reason this was still core. It should be exec.


iamluc (approved):
Shouldn't it be merge in `ddtrace-1.0.0` instead of `master`?

bwoebi:
@iamluc No, I'm targeting master, this is a bug (the shouldLoad check looks for the wrong thing for instance).

PR #2601, opened 2024-03-28
build: shrink profiling artifacts with `panic=abort`

morrisonlevi:
### Description

Recently the build artifacts got quite large due to the tracer including debug symbols for things. The profiler has also been creeping up in size, and so I thought I'd try to shrink the build artifacts. These all seem to individually give small decreases in size:

- `codegen-units=1`
- `lto="fat"`
- `debug="line-tables-only"`
- `panic=abort`
- Build std from source and do panic=abort: `-Z build-std="panic_abort,core,alloc,std"`. However, this requires nightly, to avoid this, we use the `RUSTC_BOOTSTRAP=1` hack.

Note that if you are buliding locally to try to reproduce savings, we also compress with `objcopy --compress-debug-sections` afterwards.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/dockerfiles/ci/centos/7/base.Dockerfile
+++ b/dockerfiles/ci/centos/7/base.Dockerfile
@@ -85,14 +85,22 @@ RUN source scl_source enable devtoolset-7; set -eux; \
     ../configure && make -j $(nproc) && make install; \
     cd - && rm -fr build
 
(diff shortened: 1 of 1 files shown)


realFlowControl:
I gave it a spin and the release binary on `aarch64-apple-darwin` shrunk from 6.8 MB to 3.6 MB 🎉

PR #2602, opened 2024-03-28
Support the frankenphp SAPI

bwoebi:
### Description

Builds the frankenphp SAPI in docker and creates a span for requests received on the handler.

The created span is pretty lightweight, because frankenphp (unlike roadrunner or Swoole) properly populates the superglobals, so we can rely on the internal default handling.

### Reviewer checklist
- [ ] Test coverage seems ok.


src/Integrations/Integrations/Frankenphp/FrankenphpIntegration.php
```diff
@@ -0,0 +1,61 @@
...
+                    $rootSpan->type = Type::WEB_SERVLET;
+                    $rootSpan->meta[Tag::COMPONENT] = FrankenphpIntegration::NAME;
+                    $rootSpan->meta[Tag::SPAN_KIND] = Tag::SPAN_KIND_VALUE_SERVER;
+                    unset($rootSpan->meta["closure.declaration"]);
+                    $integration->addTraceAnalyticsIfEnabled($rootSpan);
+
+                    consume_distributed_tracing_headers(null);
+                }, function (HookData $hook) {
+                    var_dump($hook->span());
+                }
```
PROFeNoM, line 53: I think this can be removed :-)

src/Integrations/Integrations/Frankenphp/FrankenphpIntegration.php
```diff
@@ -0,0 +1,61 @@
...
+            );
+        }, function (HookData $hook) {
+            \DDTrace\remove_hook($hook->data);
+        });
```
PROFeNoM, line 57: _I'm wondering out loud_

I understand that we want the hook on the callback to be set only once, the `frankenphp_handle_request` being called on each iteration of the main loop.

However, I ponder whether it is better to:
- Install and remove the hook on the callback after the request is dispatched, hence having this overhead for basically all requests (although this happens _outside_ a request)
  - _This is what's currently done_

OR

- Don't remove the callback hook; instead, only install it if it isn't already set. I guess we could check this through a WeakMap on the Closure. This way, I think we could also use `hook_function` on `frankenphp_handle_request`, which I believe to be less performance hungry...

However, it is also possible that this change would be so negligible that it isn't meaningful...

Have you considered this way?

src/Integrations/Integrations/Frankenphp/FrankenphpIntegration.php
```diff
@@ -0,0 +1,61 @@
...
+                    $rootSpan->service = \ddtrace_config_app_name('frankenphp');
+                    $rootSpan->type = Type::WEB_SERVLET;
+                    $rootSpan->meta[Tag::COMPONENT] = FrankenphpIntegration::NAME;
+                    $rootSpan->meta[Tag::SPAN_KIND] = Tag::SPAN_KIND_VALUE_SERVER;
+                    unset($rootSpan->meta["closure.declaration"]);
+                    $integration->addTraceAnalyticsIfEnabled($rootSpan);
+
+                    consume_distributed_tracing_headers(null);
+                }, function (HookData $hook) {
+                    var_dump($hook->span());
```
iamluc, line 52: debug?

bwoebi: yes.

src/Integrations/Integrations/Frankenphp/FrankenphpIntegration.php
bwoebi: Changed it. I don't think it makes a big difference, but anyway, why not :-)

PROFeNoM (approved):
Let's ship it! ⛵

PR #2610, opened 2024-04-03
feat(profiling) add uptime information to profiles

realFlowControl:
### Description

This PR will add the `info.json` with only the `start_time` in RFC3339 format
```json
{
    "application": {
        "start_time": "2024-12-31T23:59:59Z"
    }
}
```

In the profile view this will look like this:
image

The web-ui will magically calculate the uptime and show at the top and besides the time stamp. Next step is that we can create a metric out of this once we have this released

The feature used here was introduced in `libdatadog` with https://github.com/DataDog/libdatadog/pull/289

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/profiling/uploader.rs
```diff
@@ -27,12 +29,14 @@ impl Uploader {
...
         output_pprof: Option<Cow<'static, str>>,
         endpoint: AgentEndpoint,
+        start_time: DateTime<Utc>,
     ) -> Self {
         Self {
             fork_barrier,
             receiver,
             output_pprof,
             endpoint,
+            start_time: start_time.to_rfc3339_opts(chrono::SecondsFormat::Secs, true),
```
realFlowControl, line 39: The only thing I can think off where this might be wrong is if you are doing an `apachectl -k graceful` which reloads the config and could mark a new start, but technically it is still the same process. We might need to handle those cases once we add configuration to the `info.json` as it could then be changed.

AlexJF (approved):

profiling/src/profiling/uploader.rs
morrisonlevi: I believe this creates a new `Uploader`. If I am reading the PR correctly, this would get a value for uptime. Are you saying it _shouldn't_ because it's the same process?

I _think_ `apachectl -k graceful` should be considered to be graceful stop and start of a process so... no issue here?

realFlowControl: You are right, no issue here 👍 
I was thinking what would you as a user expect after a `apachectl -k graceful`, but I would expect the uptime to be reset which in this case happens.

morrisonlevi (approved):

PR #2614, opened 2024-04-04
Make the root Makefile macOS compatible

iamluc:
### Description

Make `make install` works out of the box on MacOS, like on Linux.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -656,6 +656,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "current_platform"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a74858bcfe44b22016cb49337d7b6f04618c58e5dbfdef61b06b8c434324a0bc"
+
+[[package]]
 name = "darling"
 version = "0.20.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2307,6 +2313,7 @@ name = "php_sidecar_mockgen"
 version = "0.1.0"
 dependencies = [
  "cc_utils",
+ "current_platform",
  "sidecar_mockgen",
 ]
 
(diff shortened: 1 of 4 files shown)


config.m4
```diff
@@ -306,7 +306,7 @@ EOT
     if test "$PHP_DDTRACE_SIDECAR_MOCKGEN" != "-"; then
       ddtrace_mockgen_invocation="HOST= TARGET= $PHP_DDTRACE_SIDECAR_MOCKGEN"
     else
-      ddtrace_mockgen_invocation="cd \"$ext_srcdir/components-rs/php_sidecar_mockgen\"; HOST= TARGET= CARGO_TARGET_DIR=\$(builddir)/target/ \$(DDTRACE_CARGO) run"
+      ddtrace_mockgen_invocation="cd \"$ext_srcdir/components-rs/php_sidecar_mockgen\"; HOST=$(rustc -vV | sed -n 's|host: ||p') TARGET=$(rustc -vV | sed -n 's|host: ||p') CARGO_TARGET_DIR=\$(builddir)/target/ \$(DDTRACE_CARGO) run"
```
bwoebi, line 309: We can't rely on rustc being in PATH (that's also why we do have e.g. `AC_CHECK_TOOL(DDTRACE_CARGO, cargo, [:])` for cargo).

If this is specifically a problem with our Makefile, it must be an interaction with the Makefile. After all, it works when run isolated. I'd rather fix the Makefile than changing config.m4 here.

iamluc: For the context, it seems the `host` & `target` settings require a non-empty string for non-Linux platform.
It works on macOS when run isolated because the `cc` crate is locked to version `1.0.83` where this restriction does not seem to exist.
The error occurs with our Makefile because the latest `cc` crate version is used instead (`1.0.92` at the time of writing).

But to avoid the dependency on the `rustc` binary, I moved the detection to the sidecar mockgen code directly.
WDYT?

bwoebi (approved):
Okay, looks good then :-)

PR #2618, opened 2024-04-08
feat: Add CakePHP 3+ Support

PROFeNoM:
### Description

Add support for CakePHP 3+ by replacing the hook arguments with the new CakePHP v3+ classes.

Documented in https://github.com/DataDog/documentation/pull/22569

### Reviewer checklist
- [ ] Test coverage seems ok.


src/Integrations/Integrations/CakePHP/CakePHPIntegration.php
```diff
@@ -35,115 +33,16 @@ public function init()
         $integration = $this;
 
         // Since "Dispatcher" and "App" are common names, check for a CakePHP signature before loading
+        /*
         if (!defined('CAKE_CORE_INCLUDE_PATH')) {
             return self::NOT_AVAILABLE;
         }
+        */
```
bwoebi, line 40: I suppose you will want to continue checking for these for CakePHP V2 (i.e. if the class_exists check fails)?

src/Integrations/Integrations/CakePHP/V3/CakePHPIntegrationLoader.php
```diff
@@ -0,0 +1,156 @@
...
+        \DDTrace\trace_method('Cake\Error\Middleware\ErrorHandlerMiddleware', 'handleException', [
+            'instrument_when_limited' => 1,
+            'posthook' => function (SpanData $span, array $args) use ($integration) {
+                $integration->setError($integration->rootSpan, $args[0]);
+                $span->meta[Tag::COMPONENT] = CakePHPIntegration::NAME;
+                return false;
+            },
+        ]);
+
+        \DDTrace\trace_method('Cake\Http\Response', 'getStatusCode', [
```
bwoebi, line 87: I suppose you copied from V2, but shouldn't we just use proper hook_method for these? I mean `return false` just unconditionally drops the span...

I think we can improve that one in all instances here.

PROFeNoM: Absolutely :) Copy-pasted as you said and didn't pay attention to this; I'll change these

src/Integrations/Integrations/CakePHP/V3/CakePHPIntegrationLoader.php
```diff
@@ -0,0 +1,156 @@
...
+use DDTrace\SpanData;
+use DDTrace\Tag;
+use DDTrace\Type;
+use DDTrace\Util\Normalizer;
+
+class CakePHPIntegrationLoader
+{
+    public function load($integration)
+    {
+        $integration->rootSpan = null;
```
bwoebi, line 19: Can we please move away from caching rootSpan and rather refetch \DDTrace\root_span(), as more and more frameworks are also working under Swoole (where the root span isn't global, but may change with the executing fiber).

bwoebi (approved):
Looks good to me :-)

Issue #2592, opened 2024-03-25
[Bug]: Memory leak in creating Tracer


iNviNho:
### Bug report

Hi,

After an ugprade of the library, we are experiencing memory leaks.

We are running a long running process that starts tracing every time a message is consumed from Kafka and we close the global trace every time message is successfully processed.

This is the code we run every time we consume a message to start tracing
```php
protected function startDatadogTracing(
        string $topicName,
        string $traceId = null,
        string $parentId = null
    ): void {
        if (!empty($traceId) && !empty($parentId)) {
            set_distributed_tracing_context($traceId, $parentId);
        }

        $tracer = GlobalTracer::get();
        GlobalTracer::set($tracer);

        $resourceName = $topicName . '::process';

        $options = ['start_time' => ((int)(microtime(true) * 1000 * 1000))];
        $rootSpan = $tracer->startRootSpan(
            'Jobs',
            $options
        )
            ->getSpan();

        $rootSpan->setTag(Tag::SPAN_TYPE, Type::CLI);
        $rootSpan->setTag(Tag::SERVICE_NAME, 'broker.xxx');
        $rootSpan->setTag(Tag::RESOURCE_NAME, $resourceName, true);
        $rootSpan->setTag('command_origin', $this->signature);

        if (!$this->isTracingStarted()) {
            $this->setupTracingOfClasses();
            $this->setTracingStarted(true);
        }
    }
```

and this is the whole code together with an example how we used to close root spans and send them to dd-agent./

```php
$this->startDatadogTracing(
    $message->topic_name . '-persist-offer-created',
    $message->headers['x-datadog-trace-id'] ?? null,
    $message->headers['x-datadog-parent-id'] ?? null
);
$this->offerService->processOfferCreatedToPersistenceCommand(
    $message
);
GlobalTracer::get()->getRootScope()?->getSpan()->setTag('result', 'OK');
GlobalTracer::get()->getRootScope()?->close();
GlobalTracer::get()->flush()
```

however when running the setup locally, after every processed message memory increases by a bit. I am tracking the used memory using this simple line of code

```php
dump(memory_get_usage());
```

Do you guys have any idea what we might be doing wrong?

Thank you.

### PHP version

8.1.27

### Tracer or profiler version

0.98.0

### Installed extensions

/var/www/app # php -m
[PHP Modules]
bcmath
calendar
Core
ctype
curl
datadog-profiling
date
ddappsec
ddtrace
dom
fileinfo
filter
[...]

PR #2624, opened 2024-04-15
fix: Host Scopes not being removed

PROFeNoM:
### Description

May address #2592 

Using the reproducer added as a test, the host scopes would keep on growing, hence slowly but surely increasing memory usage

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/ScopeManager.php
+++ b/src/DDTrace/ScopeManager.php
@@ -82,6 +82,8 @@ final class ScopeManager implements ScopeManagerInterface
         if ($span instanceof Span && ObjectKVStore::get($span->internalSpan, 'ddtrace_scope_activated') !== null) {
             ObjectKVStore::put($span->internalSpan, 'ddtrace_scope_activated', false);
         }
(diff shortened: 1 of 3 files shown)


src/DDTrace/ScopeManager.php
```diff
@@ -82,6 +82,8 @@ public function deactivate(Scope $scope)
         if ($span instanceof Span && ObjectKVStore::get($span->internalSpan, 'ddtrace_scope_activated') !== null) {
             ObjectKVStore::put($span->internalSpan, 'ddtrace_scope_activated', false);
         }
+
+        array_splice($this->hostRootScopes, array_search($scope, $this->hostRootScopes, true), 1);
```
PROFeNoM, line 86: Erratum: I may have to check for the index; will do it later as soon as I can

bwoebi (approved):
Ah, interesting. This is indeed suspicious.

Nice catch. Thanks!

PR #2626, opened 2024-04-15
Bump version to ddtrace 1.0.0beta1

bwoebi:
One point zero.


package.xml
```diff
@@ -56,64 +56,50 @@
...
+        <release>beta</release>
         <api>stable</api>
     </stability>
     <license uri="https://github.com/DataDog/dd-trace-php/blob/master/LICENSE">BSD 3-Clause</license>
     <notes>
         <![CDATA[
-## Tracer
-
-Windows x64 support is officially GA (generally available).
+This beta release contains all breaking changes planned for dd-trace-php 1.0. An upgrade guide can be found at https://github.com/DataDog/dd-trace-php/blob/1.0.0beta1/UPGRADE-1.0.md.
```
PROFeNoM, line 65: Question: Is 1.0.0beta1 the right blob to use? Will it get created later on?

bwoebi: Yes.

PR #2631, opened 2024-04-17
Do not check get_DD_TRACE_ENABLED() outside of a request

bwoebi:
get_DD_TRACE_ENABLED() must not be accessed outside of requests, given that it accesses a request local array of configs.

DDTRACE_G(active_stack) is supposed to be always properly reset between requests, so we can just put this first and then the calls to get_DD_TRACE_ENABLED() are always correct.

Fixes a crash in datadog_profiling::timeline::timeline_mshutdown doing collect_idle() and ultimately accessing the profiling_context here.

Not sure how to test this; it was reported with the pretty obvious backtrace by an user.

### DIFF
--- a/ext/profiling.c
+++ b/ext/profiling.c
@@ -8,7 +8,7 @@ ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
 DDTRACE_PUBLIC struct ddtrace_profiling_context ddtrace_get_profiling_context(void) {
     struct ddtrace_profiling_context context = {0, 0};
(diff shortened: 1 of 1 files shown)


realFlowControl (approved):

morrisonlevi (commented):
Thank you for this PR. We should definitely write down some of this knowledge on conditional checking order in a code comment so it is not forgotten.

morrisonlevi (commented):
Does it make any sense to fetch `DDTRACE_G(active_stack) && DDTRACE_G(active_stack)->root_span` from mshutdown? I think we may need to avoid any call to this function after post-request shutdown or so.

bwoebi:
These globals are valid until gshutdown actually, so it's fine.

PR #2635, opened 2024-04-19
ci(profiling): run clippy, fix lints

morrisonlevi:
### Description

Fixes a few clippy and compiler lints in profiling. Just trying to polish up a bit at the end of our profiling meetup.

Note that it only runs for `x86_64-unknown-linux-gnu` because there was a problem building some packages with clippy on ARM, and the alpine images don't have `clippy`. It also is only running for NTS builds of PHP.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -2928,8 +2928,11 @@ jobs:
             php run-tests.php -d "extension=/mnt/ramdisk/cargo/release/libdatadog_php_profiling.so" --show-diff -g "FAIL,XFAIL,BORK,WARN,LEAK,XLEAK,SKIP" "phpt"
 
       - run:
-          name: clippy NTS
+          name: clippy NTS (for select platforms)
           command: |
+            if [[ "x86_64-unknown-linux-gnu" != "<< parameters.triplet >>" ]] ; then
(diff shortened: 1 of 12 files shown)


realFlowControl (approved):

.circleci/continue_config.yml
```diff
@@ -2927,6 +2927,20 @@ jobs:
             export TEST_PHP_EXECUTABLE=$(which php)
             php run-tests.php -d "extension=/mnt/ramdisk/cargo/release/libdatadog_php_profiling.so" --show-diff -g "FAIL,XFAIL,BORK,WARN,LEAK,XLEAK,SKIP" "phpt"
 
+      - run:
+          name: clippy NTS (for select platforms)
+          command: |
+            if [[ "x86_64-unknown-linux-gnu" != "<< parameters.triplet >>" ]] ; then
```
bwoebi, line 2933: Do we need to run clippy for all PHP versions, or would a single compilation be enough?

morrisonlevi: There are PHP version differences, although mostly they are PHP 7 vs PHP 8.

PR #2639, opened 2024-04-23
Provide a DogStatsD API

iamluc:
### Description

Provide a DogStatsD API.

Related libdatadog PR => https://github.com/DataDog/libdatadog/pull/399

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/sidecar.c
+++ b/ext/sidecar.c
@@ -1,6 +1,7 @@
 #include "ddtrace.h"
 #include "auto_flush.h"
 #include "configuration.h"
+#include "dogstatsd_client.h"
 #include "logging.h"
 #include <components-rs/ddtrace.h>
 #include "sidecar.h"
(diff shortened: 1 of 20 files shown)


morrisonlevi:
FYI there is a tiny dogstatsd client that should probably be cleaned up or deleted at `src/dogstatsd`.

ext/sidecar.c
```diff
@@ -20,16 +23,23 @@ static void ddtrace_set_sidecar_globals(void) {
...
+        dogstatsd_endpoint = ddog_endpoint_from_api_key(dd_zend_string_to_CharSlice(get_global_DD_API_KEY()));;
     } else {
         char *agent_url = ddtrace_agent_url();
         ddtrace_endpoint = ddog_endpoint_from_url((ddog_CharSlice) {.ptr = agent_url, .len = strlen(agent_url)});
         free(agent_url);
+
+        char *dogstatsd_url = ddtrace_dogstatsd_url();
+        dogstatsd_endpoint = ddog_endpoint_from_url((ddog_CharSlice) {.ptr = dogstatsd_url, .len = strlen(dogstatsd_url)});
+        free(dogstatsd_url);
     }
```
PROFeNoM, line 38: Here, and this also applies to `ddtrace_endpoint`, I believe we don't respect the value of `DD_SITE`. Most specifically, I think that `ddog_endpoint_from_api_key` doesn't take this env var into consideration.

Maybe `ddog_endpoint_from_api_key_and_site` should be used? Maybe a new function needs to be created for `ddog_endpoint_from_url`?

Correct me if I'm wrong

bwoebi: That's correct, however we don't even read that config at all currently. We should support it for sure when we add official agentless support.

PROFeNoM: Hummmm, I'm thinking of multiple things atm:
- For instance, when using Docker, customers will typically only set `DD_SITE` on the datadog-agent container, not necessarily the instrumented container --> The tracer doesn't know the value of DD_SITE. Is there any communication of this value between the agent and the tracers (not to my knowledge)
- (Still using the Docker example) To use the DogStatsD API, does a customer MUST set DD_API_KEY in the instrumented container?
- If a European customer wants to use the DogStatsD API, then I believe this... wouldn't work, right?

PROFeNoM: Nevermind, quickly tested and it seems to work 🤔

bwoebi: I'm not following - the dogstatsd data still goes to the agent - and then agent will then send it to the correct site.

PR #2642, opened 2024-04-24
Improve login accuracy event on Laravel>5

estringana:
### Description

It was reported that success login event on Laravel was being reported on authenticated user instead on the login event. This PR fixes that.

### Reviewer checklist
- [ ] Test coverage seems ok.


tests/Common/WebFrameworkTestCase.php
```diff
@@ -21,12 +21,36 @@ abstract class WebFrameworkTestCase extends IntegrationTestCase
...
     protected $checkWebserverErrors = true;
+    protected $cookiesFile;
+    protected $maintainSession = false;
+
+    protected function ddSetUp()
+    {
+        parent::ddSetUp();
+    }
+
+    protected function enableSession()
```
estringana, line 39: This PRs introduces integration tests in which a session need to be maintained in multiple consecutive calls. That's the reason for this

src/DDTrace/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -350,7 +350,7 @@ function ($This, $scope, $args, $loginSuccess) use ($integration) {
 
         // Used by Laravel >= 5.0
         \DDTrace\hook_method(
-            'Illuminate\Auth\Events\Authenticated',
```
estringana, line 353: This is the only core change. Everything else is test code to ensure the fix is working

Anilm3 (approved):
LGTM, although double checking on the weblog before merging would be great.

PROFeNoM (approved):
👌

Issue #2636, opened 2024-04-19
[Bug]: Laravel Octane traces not showing in dashboard


s1rc:
### Bug report

We have a k8s deployment where we are turning Laravel Octane on instead of php-fpm. I have added the following`DD_TRACE_CLI_ENABLED: true`  to the deployment as per https://github.com/DataDog/dd-trace-php/issues/704#issuecomment-2049490652

We are not seeing any traces from testing calls on the server. We do see `artisan config:cache` and the nginx trace, which is configured in front of Laravel Octane. 

Aside from adding `DD_TRACE_CLI_ENABLED: true` is there anything else we need to do?

Here are the configured DD specific ENV variables:
```
DD_DOGSTATSD_URL=unix:///var/run/datadog/dsd.socket
DD_DBM_PROPAGATION_MODE=full
DD_INSTRUMENTATION_INSTALL_ID=5078c538-b063-476e-8d3f-21eaed4de329
DD_ENTITY_ID=8be3d3b3-91e5-49f9-8757-7f1773a250a6
DD_TRACE_CLI_ENABLED=1
DD_INSTRUMENTATION_INSTALL_TIME=1713534130
DD_INSTRUMENTATION_TELEMETRY_ENABLED=0
DD_SERVICE=staging-api
DD_LOGS_INJECTION=1
DD_TRACE_CLIENT_IP_ENABLED=1
DD_TRACE_AGENT_URL=unix:///var/run/datadog/apm.socket
DD_ENV=staging
DD_AGENT_HOST=10.0.38.5
DD_TRACE_CLIENT_IP_HEADER=x-forwarded-for
DD_VERSION=develop-9ac3f0c9
DD_QUEUE_TRACE_SERVICE_PROVIDER_ENABLED=false
```

### PHP version

8.2.17

### Tracer or profiler version

0.99.1

### Installed extensions
```
[PHP Modules]
bcmath
bz2
Core
ctype
curl
date
ddappsec
ddtrace
dom
exif
fileinfo
filter
gd
hash
iconv
imagick
intl
json
libxml
mbstring
mysqli
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
random
readline
Reflection
session
SimpleXML
sockets
sodium
SPL
sqlite3
standard
swoole
tokenizer
xml
xmlreader
xmlwriter
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
ddappsec
[...]

PR #2645, opened 2024-04-26
fix: Background Sender reset on Swoole Fork

PROFeNoM:
### Description

#2636

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/ddtrace.stub.php
```diff
@@ -977,4 +977,9 @@ function dd_untrace(string $functionName, string $className = null): bool {}
      * @param int $timeout Timeout in milliseconds to wait for the flush to complete
      */
     function dd_trace_synchronous_flush(int $timeout): void {}
+
+    /**
+     * @internal
+     */
+    function dd_handle_fork(): void {}
```
bwoebi, line 984: Can we please add no new functions to top-level namespace, but add to `DDTrace` or `DDTrace\Internal` as appropriate?

ext/ddtrace.c
```diff
@@ -2136,6 +2136,55 @@ static void dd_ensure_root_span(void) {
     }
 }
 
+/* {{{ proto void dd_handle_fork(): void */
+PHP_FUNCTION(dd_handle_fork) {
+    UNUSED(execute_data);
+    UNUSED(return_value);
+    // CHILD PROCESS
```
bwoebi, line 2143: I assume it's still todo to deduplicate with the code in handlers_pcntl.c?

PROFeNoM: What do you exactly mean by "deduplicate"? Do you mean some refactoring duplicated code (using this function in the pcntl code)?

bwoebi: Yes, refactoring the common code out into a single function.

bwoebi (approved):
Thanks, looks good :-)

PR #2647, opened 2024-04-26
Add all asm headers on user events

estringana:
### Description

Collect all ASM headers when a user event is triggered

APPSEC-52902

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/tags.c
+++ b/appsec/src/extension/tags.c
@@ -113,6 +113,7 @@ static zend_string *_server_zstr;
 static HashTable _relevant_headers;       // headers for requests with attacks
 static HashTable _relevant_basic_headers; // headers for all requests
 static THREAD_LOCAL_ON_ZTS bool _user_event_triggered;
+static THREAD_LOCAL_ON_ZTS bool _user_event_triggered_by_user;
 static THREAD_LOCAL_ON_ZTS bool _appsec_json_frags_inited;
 static THREAD_LOCAL_ON_ZTS zend_llist _appsec_json_frags;
 static THREAD_LOCAL_ON_ZTS zend_string *nullable _event_user_id;
@@ -122,8 +123,8 @@ static THREAD_LOCAL_ON_ZTS bool _force_keep;
 static void _init_relevant_headers(void);
 static zend_string *_concat_json_fragments(void);
 static void _zend_string_release_indirect(void *s);
-static void _add_basic_ancillary_tags(
-    zend_object *nonnull span, const zend_array *nonnull server);
+static void _add_basic_ancillary_tags(zend_object *nonnull span,
+    const zend_array *nonnull server, bool all_headers);
 static bool _add_all_ancillary_tags(
     zend_object *nonnull span, const zend_array *nonnull server);
 void _set_runtime_family(zend_object *nonnull span);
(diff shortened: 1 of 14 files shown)


appsec/src/extension/tags.c
```diff
@@ -490,7 +494,12 @@ static void _add_basic_tags_to_meta(
     zend_array *meta_ht = Z_ARRVAL_P(meta);
 
     _dd_http_client_ip(meta_ht);
-    _dd_request_headers(meta_ht, _server, &_relevant_basic_headers);
+
+    HashTable *headers = &_relevant_basic_headers;
+    if (_user_event_triggered) {
```
Anilm3, line 499: Hiding this inside `_add_basic_tags_to_meta` really changes the meaning of it, maybe it would make more sense to refactor these functions to allow the caller, through one of the arguments, to define what has to be sent and then keep the logic of what is being sent contained to `dd_tags_add_tags`. It doesn't make a practical difference, but it's cleaner than unexpected behaviours determined by globals.

Anilm3 (commented):
@estringana I added a new clarification, this should only be done on `extended` mode.

appsec/src/extension/tags.c
```diff
@@ -315,6 +319,12 @@ void dd_tags_add_appsec_json_frag(zend_string *nonnull zstr)
     zend_llist_add_element(&_appsec_json_frags, &zstr);
 }
 
+void dd_tags_set_user_event_triggered(bool trigger_by_user)
```
Anilm3, line 322: Do we need two variables to track a single action? i.e. we just want to track whether an event requiring all headers was produced.

appsec/src/extension/tags.c
```diff
@@ -330,6 +340,14 @@ void dd_tags_rshutdown()
     }
 }
 
+bool should_collect_all_headers()
+{
+    return zend_llist_count(&_appsec_json_frags) > 0 ||
```
Anilm3, line 345: `should_collect_all_headers()` is only called within an if statement which checks for `zend_llist_count(&_appsec_json_frags) == 0`, so this can never be true.

appsec/src/extension/tags.c
```diff
@@ -485,12 +504,17 @@ static bool _add_all_ancillary_tags(
 
 // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
 static void _add_basic_tags_to_meta(
-    zval *nonnull meta, const zend_array *nonnull _server)
+    zval *nonnull meta, const zend_array *nonnull _server, bool all_headers)
```
Anilm3, line 507: Another option, instead of a boolean, could be to pass a pointer to the header hash table.

appsec/src/extension/tags.c
estringana: Not now, but if there is an attack all headers should be collected

Anilm3: But this function is not used for that purpose.

appsec/src/extension/tags.c
estringana: I thought about that but I chose this approach

appsec/src/extension/tags.c
estringana: I have just refactored it

appsec/src/extension/tags.c
estringana: It's not now but then `should_collect_all_headers` would be half true

Anilm3 (approved):

PR #2651, opened 2024-04-30
Minor improvements

iamluc:
### Description

Update `CONTRIBUTING.md` + fix the `cbindgen` Makefile target not working without installing `cbindgen` manually.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -363,11 +363,23 @@ clang_format_fix:
 cbindgen: remove_cbindgen generate_cbindgen
 
 remove_cbindgen:
-	for h in components-rs/ddtrace.h components-rs/telemetry.h components-rs/sidecar.h components-rs/common.h; do if [ -f "$$h" ]; then rm "$$h"; fi; done
+	rm -f components-rs/ddtrace.h components-rs/telemetry.h components-rs/sidecar.h components-rs/common.h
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -363,53 +363,33 @@ clang_format_fix:
 cbindgen: remove_cbindgen generate_cbindgen
 
 remove_cbindgen:
-	for h in components-rs/ddtrace.h components-rs/telemetry.h components-rs/sidecar.h components-rs/common.h; do if [ -f "$$h" ]; then rm "$$h"; fi; done
+	rm -f components-rs/ddtrace.h components-rs/telemetry.h components-rs/sidecar.h components-rs/common.h
 
-generate_cbindgen: components-rs/ddtrace.h components-rs/telemetry.h components-rs/sidecar.h components-rs/common.h
+generate_cbindgen: cbindgen_binary # Regenerate components-rs/ddtrace.h components-rs/telemetry.h components-rs/sidecar.h components-rs/common.h
```
iamluc, line 368: Some context:
Currently in our dev Docker container, running `make cbindgen` deletes  the `components-rs/*.h` files. 
Then, because the `cbindgen` command is not installed, the generation of the files is skipped silently.
Finally it fails in the deduplication step because the files are missing.

As the `components-rs/*.h` files cannot be generated individually because their declarations need to be deduplicated, I believe it's simpler to generate them in the same Makefile target.

bwoebi (approved):

PR #2652, opened 2024-05-02
Fix rejected distributed sampling decision overwrite

bwoebi:
This particular issue only applied when the distributed headers were extracted onto an existing root span. Thus it worked as expected for normal http root spans, but just posed issues with messaging (e.g. laravelqueue).


PROFeNoM (approved):
It seems like this PR wasn't rebased on top of https://github.com/DataDog/dd-trace-php/pull/2642.

I'll approve as it LGTM, yet obviously, we should theoretically rebase :+1:

PR #2659, opened 2024-05-08
Instrument AMQP batched publishing

### LINKED ISSUE #2644: [Feature] Missing instrumentation for batched AMQP publishing
### Describe the feature you'd like

We use batch operations in our systems and I noticed that we have no visibility of what's happening there after we removed our custom instrumentation in favour of the one provided by `ext-ddtrace`.

Would it be possible to add instrumentation for the following methods on the `AMQPIntegration`:

* `PhpAmqpLib\Channel\AMQPChannel::batch_basic_publish()`
* `PhpAmqpLib\Channel\AMQPChannel::publish_batch()`
* `PhpAmqpLib\Channel\AMQPChannel::wait_for_pending_acks()`

### Is your feature request related to a problem?

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

lcobucci:
### Description

This fixes #2644 by configuring the extension to trace the necessary methods.

### Reviewer checklist

- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/AMQP/AMQPIntegration.php
+++ b/src/DDTrace/Integrations/AMQP/AMQPIntegration.php
@@ -169,6 +169,62 @@ class AMQPIntegration extends Integration
 
         trace_method(
             "PhpAmqpLib\Channel\AMQPChannel",
+            "batch_basic_publish",
+            [
+                'prehook' => function (SpanData $span, $args) use ($integration) {
+                    /** @var AMQPMessage $message */
+                    $message = $args[0];
+                    if (!is_null($message)) {
+                        $integration->injectContext($message);
+                    }
+                },
+                'posthook' => function (SpanData $span, $args, $exception) use ($integration) {
+                    /** @var AMQPMessage $message */
+                    $message = $args[0];
+                    /** @var string $exchange */
+                    $exchange = $args[1];
+                    /** @var string $routing_key */
+                    $routingKey = $args[2] ?? '';
+
+                    $exchangeDisplayName = $integration->formatExchangeName($exchange);
+                    $routingKeyDisplayName = $integration->formatRoutingKey($routingKey);
+
+                    $integration->setGenericTags(
+                        $span,
+                        'batch_basic.add',
+                        'producer',
+                        "$exchangeDisplayName -> $routingKeyDisplayName",
+                        $exception
+                    );
+
+                    $span->meta[Tag::RABBITMQ_ROUTING_KEY] = $routingKeyDisplayName;
+                    $span->meta[Tag::RABBITMQ_EXCHANGE] = $exchangeDisplayName;
+
+                    if (!is_null($message)) {
+                        $span->meta[Tag::MQ_MESSAGE_PAYLOAD_SIZE] = strlen($message->getBody());
+                        $integration->setOptionalMessageTags($span, $message);
+                    }
+                }
+            ]
+        );
+
+        trace_method(
+            "PhpAmqpLib\Channel\AMQPChannel",
+            "publish_batch",
+            function (SpanData $span, $args, $exception) use ($integration) {
+                $integration->setGenericTags(
+                    $span,
+                    'publish_batch',
+                    'producer',
+                    null,
+                    $exception
+                );
+                $span->meta[Tag::MQ_OPERATION] = 'send';
+            }
+        );
+
+        trace_method(
+            "PhpAmqpLib\Channel\AMQPChannel",
             "basic_consume",
             function (SpanData $span, $args, $retval, $exception) use ($integration) {
                 /** @var string $queue */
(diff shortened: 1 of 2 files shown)


bwoebi:
That's awesome! Thanks @lcobucci :-)

We'll review it soon and make sure to include it in 1.0.0.

@PROFeNoM Can you please have a look, at a glance it looks great to me, but maybe you have a comment?

lcobucci:
Thanks, happy to help.

Do let me know if there any change required :+1:

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
```diff
@@ -167,6 +167,63 @@ function ($This) use ($integration) {
...
+                    $exchange = $args[1];
+                    /** @var string $routing_key */
+                    $routingKey = $args[2] ?? '';
+
+                    $exchangeDisplayName = $integration->formatExchangeName($exchange);
+                    $routingKeyDisplayName = $integration->formatRoutingKey($routingKey);
+
+                    $integration->setGenericTags(
+                        $span,
+                        'batch_basic.add',
```
PROFeNoM, line 194: Considering that the span created from `basic_publish` is `basic.publish`, I believe that the operation name for this span should be `batch_basic.publish`.

Was there any reason to use `.add` instead of `.publish`?

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
```diff
@@ -167,6 +167,63 @@ function ($This) use ($integration) {
...
+                        'producer',
+                        "$exchangeDisplayName -> $routingKeyDisplayName",
+                        $exception
+                    );
+                    $span->meta[Tag::MQ_OPERATION] = 'batch';
```
PROFeNoM, line 199: For semantical and consistency reasons between traces, the `messaging.operation` tag must be one of:
- `send`
- `receive`
- `process`

You are not supposed to know this :) In this case, `send` should be used instead of `batch`.

The tag should be changed in the tests as well :+1:

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
```diff
@@ -167,6 +167,63 @@ function ($This) use ($integration) {
...
+            "publish_batch",
+            function (SpanData $span, $args, $exception) use ($integration) {
+                $integration->setGenericTags(
+                    $span,
+                    'publish_batch',
+                    'producer',
+                    null,
+                    $exception
+                );
+                $span->meta[Tag::MQ_OPERATION] = 'send_batch';
```
PROFeNoM, line 223: Same. `send` should be used instead of `batch`.

PROFeNoM (commented):
Thanks for your contribution 😃 It looks very good to me. There would only be some minor, semantical changes to be made ,and then it's all good to me :+1:

PS: Sorry for the delay reviewing, I was OOO last week ;-)

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
lcobucci: I wanted to differentiate the actions that accumulates data in memory from the ones that actually interact with the server - IMHO the naming in the underlying library is misleading as `batch_basic_publish()` doesn't publish anything until `publish_batch()` is called.

Let me know what's your preference and I can adjust it.

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
lcobucci: Should it be `process`, then? I mean, that operation isn't sending nor receiving anything :)

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
PROFeNoM: Makes sense :+1: That's good for me then :)

src/DDTrace/Integrations/AMQP/AMQPIntegration.php
PROFeNoM: Both `receive` and `process` are reserved for messages consumer/server, so we cannot use either in this specific case :( 

As you correctly pointed out, this operation isn't a producer _per se_ because it doesn't send anything... (Let's keep it as a producer though, as it makes sense with the [OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/9fa7c656b26647b27e485a6af7e38dc716eba98a/specification/trace/api.md#spankind)) There are two possibilities, I believe:
1. We use `send`, but it's misleading
2. We don't put a `messaging.operation` tag

I believe the second option makes the most sense for this operation, as this tag is not mandatory. If it doesn't make sense, then let's keep it simple imo

PROFeNoM (approved):
LGTM! 😃 Thanks a lot for your superb contribution 🤗

PR #2662, opened 2024-05-14
build: update profiling deps/libdatadog v9

morrisonlevi:
### Description

This bumps dependency versions to more recent releases, including libdatadog to v9.0.0.

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/profiling/mod.rs
```diff
@@ -1114,7 +1122,7 @@ mod tests {
     }
 
     #[test]
-    #[cfg(feature = "timeline")]
+    #[cfg(all(feature = "timeline", not(miri)))]
```
morrisonlevi, line 1125: This is unrelated, but I was using `cargo +nightly miri` locally to see if the specific nightly version was compatible with the changes, and this fails on miri:

> error: unsupported operation: `extern` static `bindings::ffi::datadog_php_profiling_get_profiling_context` from crate `datadog_php_profiling` is not supported by Miri

profiling/Cargo.toml
```diff
@@ -14,23 +14,24 @@ bench = false # disables cargo build in libtest bench
 [dependencies]
 # with ahash 0.8.7 we run into https://github.com/tkaitchuck/aHash/issues/195
 # should be fixed with updating Rust to 1.72
-ahash = { version = "=0.8.6" }
+ahash = { version = "=0.8.11" }
```
realFlowControl, line 17: We should update the comment above, thats not true anymore

realFlowControl (approved):

PR #2664, opened 2024-05-15
Peer Service Precursors for DBM

PROFeNoM:
### Description

Please take a look at [AIT-10031](https://datadoghq.atlassian.net/browse/AIT-10031) for the full details (contains the short RFC).

**Changed**
- `dddbs`: It should NOT be overridden by peer.service. There is some fake service thingy in the RFC, but I don't believe it applies there? Or there is something that I'm unaware of

**Added**
- `ddh`: The hostname of the database server
- `dddb`: The schema/database/namespace
- `ddprs`: Value of peer.service if and only if the customer explicitly sets this

### Reviewer checklist
- [ ] Test coverage seems ok.

[AIT-10031]: https://datadoghq.atlassian.net/browse/AIT-10031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/DDTrace/Integrations/DatabaseIntegrationHelper.php
+++ b/src/DDTrace/Integrations/DatabaseIntegrationHelper.php
@@ -44,9 +44,18 @@ class DatabaseIntegrationHelper
             }
 
             $span = $hook->span();
-            $databaseService = $span->meta['peer.service'] ?? $span->service;
+            $targetHost = $span->meta[Tag::TARGET_HOST] ?? '';
+            $dbName = $span->meta[Tag::DB_NAME] ?? $span->meta[Tag::DB_INSTANCE] ?? '';
+            $peerService = $span->meta['peer.service'] ?? '';
 
-            $query = self::propagateViaSqlComments($hook->args[$argNum], $databaseService, $propagationMode);
+            $query = self::propagateViaSqlComments(
+                $hook->args[$argNum],
+                $span->service,
+                $propagationMode,
+                $targetHost,
+                $dbName,
+                $peerService
+            );
             $hook->args[$argNum] = $query;
             $hook->overrideArguments($hook->args);
             if ($propagationMode == \DDTrace\DBM_PROPAGATION_FULL) {
@@ -55,12 +64,22 @@ class DatabaseIntegrationHelper
         }
     }
 
-    public static function propagateViaSqlComments($query, $databaseService, $mode = \DDTrace\DBM_PROPAGATION_FULL)
-    {
+    public static function propagateViaSqlComments(
+        $query,
+        $databaseService,
+        $mode = \DDTrace\DBM_PROPAGATION_FULL,
+        $targetHost = '',
+        $dbName = '',
+        $peerService = ''
+    ) {
         $rootSpan = \DDTrace\root_span();
 
         // Note: the order of the tags is relevant, they must be passed ordered alphabetically
         $tags = [];
+
+        if ($dbName != "") {
+            $tags["dddb"] = $dbName;
+        }
 
         if ($databaseService != "") {
             $tags["dddbs"] = $databaseService;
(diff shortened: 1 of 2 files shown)


src/DDTrace/Integrations/DatabaseIntegrationHelper.php
```diff
@@ -44,34 +44,61 @@ public static function injectDatabaseIntegrationData(HookData $hook, $backend, $
...
         $tags = [];
 
         if ($databaseService != "") {
-            $tags["dddbs"] = $databaseService;
+            $tags[Tag::DBM_DATABASE_SERVICE_NAME] = $databaseService;
+        }
+
+        if ($targetHost != "") {
+            $tags[Tag::DBM_PEER_HOSTNAME] = $targetHost;
+        }
```
bwoebi, line 86: This is not visible anymore given that you put these as constants, but this is "ddh", which is not respecting the order (before is "dddbs", next is "dddb"), see the // Note a few lines above.
Hence, to make this much more visible, I would recommend _not_ using Tag:: here, but just adding it as a comment, e.g.
```
// peer hostname
$tags["ddh"] = $targetHost;
```
The strength of constants is when they are reused, but here's no reuse at all, so I think the approach with the comment would be better.

PROFeNoM: > which is not respecting the order

_Wasn't aware there was an order_

PROFeNoM: I'll make the changes

bwoebi: It may also make sense to have a test actually verify these are ordered. At least you missed the comment in the source file :-D Others might too.

tests/Integration/DatabaseMonitoringTest.php
```diff
@@ -212,6 +214,39 @@ public function testInjectionPeerService()
...
+        } finally {
+            \DDTrace\remove_hook($hook);
+        }
+
+        $dbmComment = preg_replace('/^.*\/\*(.*)\*\/.*$/', '$1', $commentedQuery);
+        $dbmCommentParts = explode(',', $dbmComment);
+        $keys = array_map(function ($part) {
+            return explode('=', $part)[0];
+        }, $dbmCommentParts);
+        $this->assertSame(['dddb', 'dddbs', 'dde', 'ddh', 'ddprs', 'ddps', 'ddpv', 'traceparent'], $keys);
```
bwoebi, line 247:
```suggestion
        $sortedKeys = $keys;
        sort($sortedKeys);
        $this->assertSame($sortedKeys, $keys);
```
Actually testing the order, not the present keys. Otherwise anyone extending this will just add his key without thinking about the fact that it needs to be ordered.

src/DDTrace/Integrations/DatabaseIntegrationHelper.php
```diff
@@ -44,9 +44,18 @@ public static function injectDatabaseIntegrationData(HookData $hook, $backend, $
             }
 
             $span = $hook->span();
-            $databaseService = $span->meta['peer.service'] ?? $span->service;
-
-            $query = self::propagateViaSqlComments($hook->args[$argNum], $databaseService, $propagationMode);
+            $targetHost = $span->meta[Tag::TARGET_HOST] ?? '';
+            $dbName = $span->meta[Tag::DB_NAME] ?? $span->meta[Tag::DB_INSTANCE] ?? '';
+            $peerService = $span->meta['peer.service'] ?? '';
```
tabgok, line 49: nit: this seems like the only tag not specified as `Tag::...` and I'm not sure if that's intentional or not

tabgok (approved):
Approving from a "this does the right thing for DBM comments" perspective

src/DDTrace/Integrations/DatabaseIntegrationHelper.php
PROFeNoM: Yes, this was sort of intentional since there is no `Tag::PEER_SERVICE`. I can add it - won't make any harm

bwoebi (approved):
LGTM now :-)

PR #2665, opened 2024-05-15
Move env and version from meta to span properties

bwoebi:
As a fallback, direct assignments to $span->meta["env"] (or version) do still take precedence over the new properties. This was deprecated, and will be eventually removed.

Side note: There were tests for DD_TAGS vs DD_ENV and DD_VERSION. These tests started failing. I elected to remove the tests instead of adding special handling in ddtrace_set_global_span_properties() (if key equals env and property_env non-empty, then ignore). These tests came from a 4 year old migration, at which time these tags had to be set via DD_TAGS, thus prioritizing DD_ENV and DD_VERSION explicitly made sense. By now it would just be a waste of CPU cycles to check that.

This change is important for being able to track changes to service/env/version in the future context of remote configuration.

### DIFF
--- a/ext/ddtrace.stub.php
+++ b/ext/ddtrace.stub.php
@@ -83,6 +83,18 @@ namespace DDTrace {
         public string|null $service = "";
 
         /**
+         * @var string The environment you are tracing. Defaults to active environment at the time of span creation
+         * (i.e., the parent span), or datadog.env initialization settings if no parent exists
+         */
+        public string $env = "";
+
+        /**
+         * @var string The version of the application you are tracing. Defaults to active version at the time of
+         * span creation (i.e., the parent span), or datadog.version initialization settings if no parent exists
+         */
+        public string $version = "";
+
+        /**
          * @var string|null The type of request which can be set to: web, db, cache, or custom (Optional). Inherited
          * from parent.
          */
--- a/ext/ddtrace_arginfo.h
+++ b/ext/ddtrace_arginfo.h
@@ -1,5 +1,5 @@
 /* This is a generated file, edit the .stub.php file instead.
- * Stub hash: 206c9d9eefe4af1076007627bb6d5651e23202ce */
+ * Stub hash: cbd454866b5b3f34bee057c9a35d7807c6e718ca */
 
 ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(arginfo_DDTrace_trace_method, 0, 3, _IS_BOOL, 0)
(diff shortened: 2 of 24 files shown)


ext/ddtrace_arginfo.h
```diff
@@ -532,6 +525,18 @@ static zend_class_entry *register_class_DDTrace_SpanData(void)
 	zend_declare_typed_property(class_entry, property_service_name, &property_service_default_value, ZEND_ACC_PUBLIC, NULL, (zend_type) ZEND_TYPE_INIT_MASK(MAY_BE_STRING|MAY_BE_NULL));
 	zend_string_release(property_service_name);
 
+	zval property_env_default_value;
+	ZVAL_EMPTY_STRING(&property_env_default_value);
+	zend_string *property_env_name = zend_string_init("env", sizeof("env") - 1, 1);
```
iamluc, line 530: Nitpick: what about using `ZEND_STRL` macro?

ext/ddtrace.stub.php
```diff
@@ -82,6 +82,18 @@ class SpanData {
...
+         * @var string The environment you are tracing. Defaults to active service at the time of span creation
+         * (i.e., the parent span), or datadog.env initialization settings if no parent exists
+         */
+        public string $env = "";
+
+        /**
+         * @var string The version of the application you are tracing. Defaults to active service at the time of
+         * span creation (i.e., the parent span), or datadog.version initialization settings if no parent exists
+         */
+        public string $version = "";
```
morrisonlevi, line 95: These default to the "active service"? Is that right?

iamluc (commented):
There is a similar hack for the `service.name` tag, but handled in userland: https://github.com/DataDog/dd-trace-php/blob/d602b97b797a96ab254b1bd26aadf599068906c1/src/DDTrace/Span.php#L152-L154

would it be worthwhile to harmonize?

ext/ddtrace.stub.php
bwoebi: Copy paste is a powerful tool. Powerful tools should be handled with more care than I apparently did :-D

ext/ddtrace_arginfo.h
iamluc: 🤣

bwoebi:
@iamluc the SERVICE_NAME is less of a hack, and more of an OpenTracing thing. I.e. it comes from an API which never directly assigns to meta. So the equivalence isn't really applicable here.

iamluc (approved):

PR #2668, opened 2024-05-20
feat(profiling): keep string cache data alive longer

morrisonlevi:
### Description

PROF-9796

At the high-level, there are two wins here:
 1. The strings are no longer allocated with the system allocator, which was demonstrated to be slightly more efficient for both memory and CPU in libdatadog.
 2. The string set is no longer thrown away at the start of every request. Instead, we allow it to live to the next request if the total memory used is under some threshold.

#### Details

This uses the new allocators in libdatadog to build a `StringSet`, and drops the old string table from the code. Like the new `StringTable` in libdatadog v9, the string data for the `StringSet` is held in an arena. However, it doesn't track insert order, nor hand out ids. It returns references to interned data instead, like a regular set.

We store `ThinStr<'a>`s in the runtime cache, as `usize`s. This bit is unsafe because the lifetime of the cache doesn't fit nicely in the borrow checker's semantics. However, the implementation guarantees that the data lives long enough.

The run time cache is empty on each request, but the table data is still around. So we'll re-establish links to the cache to the data, but we don't have to rebuild the set structure as much compared to the previous implementation

If the allocators powering the string set report memory over 4 MiB at the end of a request, then the string set will start over. This particular number was chosen because it is the same size used for the chunk size of the chain allocator powering the set. The idea is that if we have more than one chunk, then maybe we're too big. The exact thresholds should be tweaked based on experimentation and more data.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi:
Now that the copy/paste issue and the unrelated sigsegvs have been fixed, this is looking much better. I plan to do some manual testing to be more confident, but I'm marking this as ready for review.

profiling/src/profiling/stack_walking.rs
```diff
@@ -119,19 +127,30 @@ mod detail {
             let hit_rate = stats.hit_rate();
             debug!("Process cumulative {stats:?} hit_rate: {hit_rate}");
         });
+
+        CACHED_STRINGS.with(|cell| {
+            let set: &StringSet = &cell.borrow();
+            // todo: pin down threshold and remove magic number
+            if set.arena_used_bytes() > 4 * 1024 * 1024 {
```
realFlowControl, line 134: Do we have any statistics about how often this would happen? Could we maybe also make this configurable for customers but with a "sane" default?

morrisonlevi: I don't have data yet. That's one thing I'm going to try to discover with some manual testing. For now I'll probably add a `debug!` log line while I test it manually.

morrisonlevi: On a Symfony php-fpm app, I was able to get it up to about 227 KiB. On a WordPress apache app, I couldn't even hit 64 KiB. I did this hammering multiple endpoints with `ab -t 600 -c 2 $url`.

So with two apps, I've seen a pretty big difference in behavior. Hard to say. It'd be nice to have real metrics.

realFlowControl (approved):

PR #2673, opened 2024-05-24
Add span field meta_struct

estringana:
### Description

Add new field `meta_struct` to the span. This new field is a map where its values are encoded to msgpack and then the resulting codification to msgpack binary

APPSEC-52612

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/serializer.c
```diff
@@ -1017,6 +1045,27 @@ static void dd_serialize_array_metrics_recursively(zend_array *target, zend_stri
...
+    mpack_writer_destroy(&writer);
+
+    if (size == 0 || result == 0) {
+        return;
+    }
+
+    zval serialised;
+    ZVAL_STRINGL(&serialised, data, size);
+
+    dd_serialize_array_recursively(target, str, &serialised, false);
```
bwoebi, line 1064: So, `serialised` is always a string, the only thing dd_serialize_array_recursively does ... is then adding it to the hashtable - it was confusing to read, so just simplify it:
```suggestion
    zend_hash_update(target, str, &serialised);
```

ext/serializer.c
```diff
@@ -1017,6 +1045,27 @@ static void dd_serialize_array_metrics_recursively(zend_array *target, zend_stri
...
+    free(data);
+    zval_ptr_dtor(&serialised);
```
bwoebi, line 1066:
```suggestion
```

And remove this now given that zend_hash_update() doesn't incref (the ddtrace_convert_to_string in dd_serialize_array_recursively was doing that though).

estringana: Done! Thanks

bwoebi (approved):
Looks correct to me

Anilm3 (approved):
lgtm

PR #2682, opened 2024-05-31
Fix exception profiling in Generator->throw()

bwoebi:
The prev_execute_frame of fake frame of the generator is not fixed up when throwing. Only when executing afterwards.

[SCP-456](https://datadoghq.atlassian.net/browse/SCP-456)

[SCP-456]: https://datadoghq.atlassian.net/browse/SCP-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ext/handlers_api.c
+++ b/ext/handlers_api.c
@@ -2,12 +2,23 @@
 
 // This file is compiled by both the tracer and profiler.
 
-void datadog_php_install_handler(datadog_php_zif_handler handler) {
-    zend_function *old_handler;
-    old_handler = zend_hash_str_find_ptr(CG(function_table), handler.name, handler.name_len);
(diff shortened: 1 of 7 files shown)


morrisonlevi:
As discussed in Slack, Bob was able to [find a reproducer for this crash in pure PHP](https://3v4l.org/aqFD6) (our extension isn't needed) and opened an upstream issue: [Crash when stack walking in destructor of yielded from values during Generator->throw()](https://github.com/php/php-src/issues/14387)

realFlowControl (approved):
Awesome work 🎉

PR #2684, opened 2024-05-31
Init the new `loader` extension

iamluc:
### Description

This extension will be used to find and load the right `ddtrace` extension to support PHP Single Step Instrumentation

### Reviewer checklist
- [ ] Test coverage seems ok.


.circleci/continue_config.yml
```diff
@@ -2941,6 +2944,83 @@ jobs:
...
+            cd loader
+            phpize && ./configure
+            make clean && make -j all ECHO_ARG="-e" CFLAGS="-std=gnu11 -O2 -g -Wall"
+      - run:
+          name: Download the dd-trace-php package
+          command: |
+            set -eo pipefail
+            cd loader
+            curl -Lo dd-library-php.tar.gz << parameters.dd_library_package_url >>
+            tar xzf dd-library-php.tar.gz
```
bwoebi, line 2981: I suppose that's temporary, and we're going to put this pipeline after _package extension_?
Like, having the test_loader job depend on compile_loader and package extension.

iamluc: Yes. 
I did this to debug the CI pipeline faster

loader/compat_php.c
```diff
@@ -0,0 +1,91 @@
...
+
+	ZEND_ASSERT(!(HT_FLAGS(ht) & HASH_FLAG_PACKED));
+
+	p = (Bucket*)zend_hash_find(ht, key);
+	if (UNEXPECTED(p)) {
+		return (p == b) ? &p->val : NULL;
+	}
+
+	if (!ddloader_zstr_is_interned(php_api_no, key)) {
+		zend_string_addref(key);
```
bwoebi, line 41: zend_string_addref() also does to an interned check.
Can we please do?
```suggestion
		GC_ADDREF(key);
```

using
```
#if PHP_VERSION_ID < 70300
#define GC_ADDREF(x) (++GC_REFCOUNT(x))
#define GC_DELREF(x) (--GC_REFCOUNT(x))
#endif
```

And later, in ddloader_zend_hash_set_bucket_key, instead of `if (php_api_no > 20170718) {` also do `if (!ddloader_zstr_is_interned(php_api_no, b->key)) { if (GC_DELREF(b->key) == 0) pefree(s, GC_FLAGS(s) & (php_api_no <= 20170718 ? 1 : IS_STR_PERSISTENT)); }`? Instead of this _leak_ thing?

bwoebi: Ideally, we'd just have api_no-aware `ddloader_zend_string_addref()`, `ddloader_zend_string_init()` and `ddloader_zend_string_init()`, with proper persistent strings too. I don't feel too comfortable with using non-persistent strings in the minit there.

iamluc: Fixed.

FI, I haven't put in the code these defines"
```
#if PHP_VERSION_ID < 70300
#define GC_ADDREF(x) (++GC_REFCOUNT(x))
#define GC_DELREF(x) (--GC_REFCOUNT(x))
#endif
```

IMHO, it's just fine to say that this extension can be compiled only on PHP >= 7.4. I added a README.md to mention that.

bwoebi: Okay, I'm fine with that.

.circleci/continue_config.yml
```diff
@@ -5073,3 +5145,63 @@ workflows:
...
+
+      - "test_loader":
+          requires:
+            - "Compile Linux"
+          name: "[Linux] Loader PHP << matrix.php_major_minor >>-<< matrix.switch_php_version >>"
+          docker_image: "datadog/dd-trace-ci:php-<< matrix.php_major_minor >>_buster"
+          matrix:
+            parameters:
+              php_major_minor:
+#                - "5.6"
```
bwoebi, line 5170: We still need a test that, on 5.6, that the loader will disable itself. Without doing anything or emitting warnings. (Change the url to 1.0.0 to test without PHP 5 binaries included).

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+    zend_module_entry *ddtrace = zend_hash_str_find_ptr(&module_registry, ZEND_STRL("ddtrace"));
+    if (ddtrace) {
+        LOG(INFO, "ddtrace is already loaded, unregister ddtrace_injected");
+        ddloader_unregister_module();
+
+        return SUCCESS;
+    }
+
+    LOG(INFO, "ddtrace is not loaded, checking the dependencies");
+
```
bwoebi, line 124:
```suggestion

    // Normally done by zend_startup_module_ex, but we temporarily replaced these to skip potential errors. Check it ourselves here.
```

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+static int ddloader_load_ddtrace(int php_api_no, char *module_build_id, bool is_zts, bool is_debug) {
+    char *ext_path = ddloader_find_ext_path("trace", "ddtrace", php_api_no, is_zts, is_debug);
+    if (!ext_path) {
+        LOG(ERROR, "Extension file not found");
+        return SUCCESS;
+    }
+
+    // The code below basically comes from the function "php_load_extension" in "ext/standard/dl.c",
+    // which does not allow loading an extension using a full path.
```
bwoebi, line 173: That's news to me. php_load_extension reads:
```
	if (strchr(filename, '/') != NULL || strchr(filename, DEFAULT_SLASH) != NULL) {
		/* Passing modules with full path is not supported for dynamically loaded extensions */
		if (type == MODULE_TEMPORARY) {
			php_error_docref(NULL, E_WARNING, "Temporary module name should contain only filename");
			return FAILURE;
		}
		libpath = estrdup(filename);
```
Just not allowed for MODULE_TEMPORARY, but otherwise it works perfectly fine.

bwoebi:
```suggestion
    // The code below basically comes from the function "php_load_extension" in "ext/standard/dl.c",
    // but we need to rename the extension before passing it into the module_registry.
```

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+
+    zend_module_entry *module_entry;
+    zend_module_entry *(*get_module)(void);
```
bwoebi, line 176: Just delare the symbols inline.
```suggestion
```

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+
+    LOG(INFO, "Found extension file: %s", ext_path);
+
+    void *handle = DL_LOAD(ext_path);
+    if (!handle) {
+        LOG(ERROR, "Cannot load the extension");
+        goto abort;
+    }
+
+    get_module = (zend_module_entry * (*)(void)) ddloader_dl_fetch_symbol(handle, "_get_module");
```
bwoebi, line 186:
```suggestion
    zend_module_entry *(*)(void) get_module = (zend_module_entry *(*)(void)) ddloader_dl_fetch_symbol(handle, "_get_module");
```

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+    if (!get_module) {
+        LOG(ERROR, "Cannot fetch the module entry");
+        goto abort_and_unload;
+    }
+
+    module_entry = get_module();
```
bwoebi, line 192:
```suggestion
    zend_module_entry *module_entry = get_module();
```

bwoebi:
Just adding a note so that we don't forget about it: if getenv("DD_TELEMETRY_FORWARDER_PATH") is set, then fork(), execve() to it with the telemetry payload telling why it wasn't loaded.

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+    DL_UNLOAD(handle);
+abort:
+    LOG(INFO, "Abort the loader");
+    free(ext_path);
+
+    return SUCCESS;
+}
+
+static inline void ddloader_configure() {
+    if (getenv("DD_LOADER_ENABLE_LOGS")) {
```
bwoebi, line 245:
```suggestion
    if (getenv("DD_TRACE_DEBUG")) {
```
as per the Guardrails RFC

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+static inline void ddloader_logf(const char *level, const char *format, ...) {
+    if (!debug_logs) {
+        return;
+    }
+    char buf[256];
+
+    va_list va;
+    va_start(va, format);
+    vsnprintf(buf, sizeof(buf), format, va);
+    php_printf("[dd_library_loader][%s] %s\n", level, buf);
```
bwoebi, line 40: Can you please call `_php_error_log(0, msg, NULL, NULL);`? Makes it go to the proper place instead of stdout.

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+    free(ext_path);
+
+    return SUCCESS;
+}
+
+static inline void ddloader_configure() {
+    if (getenv("DD_LOADER_ENABLE_LOGS")) {
+        debug_logs = true;
+    }
+    if (getenv("DD_LOADER_FORCE")) {
```
bwoebi, line 248:
```suggestion
    if (getenv("DD_INJECT_FORCE")) {
```

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+static const zend_module_dep ddtrace_injected_module_deps[] = {ZEND_MOD_OPTIONAL("json") ZEND_MOD_OPTIONAL("standard") ZEND_MOD_OPTIONAL("ddtrace") ZEND_MOD_END};
+static const zend_module_dep *orig_ddtrace_module_deps = NULL;
+static const zend_function_entry *orig_functions;
+
+static bool debug_logs = false;
+static bool force_load = false;
+
+static int php_api_no = 0;
+
+static void ddloader_error_handler(int error_num, zend_string *error_filename, const uint32_t error_lineno, zend_string *message) {}
```
bwoebi, line 24: Maybe forward to dd_loader_logf()?

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+        case 320190902:  // 7.4
+        case 420200930:  // 8.0
+        case 420210902:  // 8.1
+        case 420220829:  // 8.2
+        case 420230831:  // 8.3
+            break;
+
+        default:
+            LOG(WARN, "Unknown api no: %d", api_no);
+            if (!force_load) {
```
bwoebi, line 270:
```suggestion
            if (!force_load && api_no > 420230831) {
```
No point in trying to force_load PHP 5, this would probably simply crash the loader itself.

loader/dd_library_loader.c
```diff
@@ -0,0 +1,355 @@
...
+    if (!ddtrace_injected) {
+        return;
+    }
+
+    // Set the MSHUTDOWN function to NULL to avoid it being called by zend_hash_str_del
+    ddtrace_injected->module_shutdown_func = NULL;
+    zend_hash_str_del(&module_registry, ZEND_STRL("ddtrace_injected"));
+}
+
+static PHP_MINIT_FUNCTION(ddtrace_injected) {
```
bwoebi, line 114: How hard would it be, to not hardcode `"ddtrace"`, but in a single place?
I'd love to be able to later on easily just add `ddloader_load_extension("profiler", /* some more arg which is a hint for ddloader_find_ext_path */)` or similar, and it'll just load the proper extension.

iamluc: Done using a global array
ie.
```
static injected_ext injected_ext_config[] = {
    DECLARE_INJECTED_EXT("ddtrace", "trace", ((zend_module_dep[]){ZEND_MOD_OPTIONAL("json") ZEND_MOD_OPTIONAL("standard") ZEND_MOD_OPTIONAL("ddtrace") ZEND_MOD_END})),
    DECLARE_INJECTED_EXT("datadog-profiling", "profiling", ((zend_module_dep[]){ZEND_MOD_END})),
    DECLARE_INJECTED_EXT("ddappsec", "appsec", ((zend_module_dep[]){ZEND_MOD_END})),
};
```

The complexity is that  we must backup some data of the module when we load it, and restore them at MINIT.
And to avoid creating an MINIT function for each extension we want to load, I use the "module_number" to identify which extension is starting.

bwoebi: That's perfect :-)

loader/.gitignore
```diff
@@ -0,0 +1,43 @@
+*~
```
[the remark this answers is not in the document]
iamluc: I added the missing entries in the root .gitignore and remove the loader one

loader/dd_library_loader.c
```diff
@@ -0,0 +1,516 @@
...
+}
+
+static void ddloader_telemetryf(telemetry_reason reason, const char *format, ...) {
+    va_list va;
+    va_start(va, format);
+    ddloader_logv(reason == REASON_COMPLETE ? INFO : ERROR, format, va);
+    va_end(va);
+
+    if (!telemetry_forwarder_path) {
+        LOG(INFO, "Telemetry disabled: environment varirable 'DD_TELEMETRY_FORWARDER_PATH' is not set.")
```
bwoebi, line 72:
```suggestion
        LOG(INFO, "Telemetry disabled: environment variable 'DD_TELEMETRY_FORWARDER_PATH' is not set.")
```

loader/dd_library_loader.c
```diff
@@ -0,0 +1,516 @@
...
+}
+
+/**
+ * Try to load a symbol from a library handle.
+ * As some OS prepend _ to symbol names, we try to load with and without it.
+ */
+static void *ddloader_dl_fetch_symbol(void *handle, const char *symbol_name_with_underscoe) {
+    void *symbol = DL_FETCH_SYMBOL(handle, symbol_name_with_underscoe + 1);
+    if (!symbol) {
+        symbol = DL_FETCH_SYMBOL(handle, symbol_name_with_underscoe);
```
bwoebi, line 170:
```suggestion
static void *ddloader_dl_fetch_symbol(void *handle, const char *symbol_name_with_underscore) {
    void *symbol = DL_FETCH_SYMBOL(handle, symbol_name_with_underscore + 1);
    if (!symbol) {
        symbol = DL_FETCH_SYMBOL(handle, symbol_name_with_underscore);
```

loader/compat_php.c
```diff
@@ -0,0 +1,212 @@
...
+
+ZEND_API zval *ZEND_FASTCALL zend_hash_set_bucket_key(HashTable *ht, Bucket *b, zend_string *key) __attribute__((weak));
+
+ZEND_API zval *ZEND_FASTCALL zend_hash_update(HashTable *ht, zend_string *key, zval *pData) __attribute__((weak));
+ZEND_API zval *ZEND_FASTCALL _zend_hash_update(HashTable *ht, zend_string *key, zval *pData) __attribute__((weak));
+
+static bool ddloader_zstr_is_interned(int php_api_no, zend_string *key) {
+    if (php_api_no <= 20170718) {  // PHP 7.0 - 7.2
+        php70_71_72_zend_refcounted_h *gc = (php70_71_72_zend_refcounted_h*)&key->gc;
+        return (gc->u.v.flags & PHP_70_71_72_IS_STR_INTERNED);
```
bwoebi, line 32:
```suggestion
        return GC_TYPE_INFO(gc) & (PHP_70_71_72_IS_STR_INTERNED << 8);
```
Is probably the simpler way to express what you want. At least you're also 8-shifting for IS_STR_PERSISTENT below. Let's not have multiple ways to express the same thing.

loader/compat_php.c
```diff
@@ -0,0 +1,212 @@
...
+    UNUSED(error_lineno);
+    LOG(WARN, "Error while registering the module: %s)", ZSTR_VAL(message));
+}
+
+void (*old_zend_error_cb)(void);
+
+void ddloader_replace_zend_error_cb(int php_api_no) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wincompatible-pointer-types"
+    old_zend_error_cb = zend_error_cb;
```
bwoebi, line 182:
```suggestion
    old_zend_error_cb = (void *)zend_error_cb;
```
You may just cast the stuff to `void` on the assignments to avoid the warning IIRC (same below)

loader/dd_library_loader.c
```diff
@@ -0,0 +1,548 @@
...
+#include <stdbool.h>
+#include <ext/standard/basic_functions.h>
+
+#include "compat_php.h"
+#include "php_dd_library_loader.h"
+
+static bool debug_logs = false;
+static bool force_load = false;
+static char *telemetry_forwarder_path = NULL;
+static char *package_path = "/home/circleci/app/dd-library-php";  // FIXME
```
bwoebi, line 19: Is this something to hardcode, or can we look it up in configuration_hash (`cfg_get_string`)?
Like, from my perspective, we anyway always define an INI file to include the `zend_extension`. So at that place, just indicate the package_path as well I'd say.

bwoebi: I see you anyway have an env var. Can we just require it to be set?

iamluc: Done. Env var is now required.

bwoebi (approved):
Looks good by me now - nice work :-)

PR #2685, opened 2024-05-31
feat: OpenAI Integration

PROFeNoM:
### Description

AIDM-34

image
image
image
image
image

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
I'd like to discuss the DatadogLogger, and would like to know if this needs to be done this way.

At a glance, I'm not particularly fond of it. I don't mind having logging. But I think the extension should expose an API to log messages, rather than having a custom logger written in PHP. Which then get sent to the right location, can be forwarded to the sidecar, possibly for logging telemetry and such in future.

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1197 @@
...
+            $this->serviceName = \dd_trace_env_config('DD_OPENAI_SERVICE');
+        } elseif (\dd_trace_env_config('DD_SERVICE')) {
+            $this->serviceName = \dd_trace_env_config('DD_SERVICE');
+        } else {
+            $this->serviceName = null;
+        }
+
+        if (!empty($this->serviceName)) {
+            $span->service = $this->serviceName;
+        }
```
bwoebi, line 72: Is there an else missing? Like
```suggestion
        if (!empty($this->serviceName)) {
            $span->service = $this->serviceName;
        } else {
            if (\dd_trace_env_config('DD_OPENAI_SERVICE')) {
                $this->serviceName = \dd_trace_env_config('DD_OPENAI_SERVICE');
            } elseif (\dd_trace_env_config('DD_SERVICE')) {
                $this->serviceName = \dd_trace_env_config('DD_SERVICE');
            } else {
                $this->serviceName = null;
            }

            if (!empty($this->serviceName)) {
                $span->service = $this->serviceName;
            }
        }
```

Also, I'd consider not fetching DD_SERVICE. It should respect inheritance, i.e. let's just do:
```suggestion
        if ($service = \dd_trace_env_config('DD_OPENAI_SERVICE')) {
            $span->service = $service;
        }
```

And that's it. No caching, always correct and simple.

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1185 @@
...
+    }
+
+    public static function getLogTags(
+        SpanData $span,
+        string   $methodName
+    ): array
+    {
+        $tags = [
+            'env' => \dd_trace_env_config('DD_ENV'),
+            'version' => \dd_trace_env_config('DD_VERSION'),
```
bwoebi, line 544:
```suggestion
            'env' => $span->env,
            'version' => $span->version,
```
No point in fetching from config - it's inherited and that's the important source.

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1185 @@
...
+            $errorType = $span->exception->getErrorType() ?? $span->exception->getErrorCode() ?? null;
+        } elseif ($span->exception) {
+            $errorType = \get_class($span->exception);
+        }
+
+
+        $tags = [
+            'env' => \dd_trace_env_config('DD_ENV'),
+            'service' => $span->service,
+            'version' => \dd_trace_env_config('DD_VERSION'),
```
bwoebi, line 663:
```suggestion
            'env' => $span->env,
            'service' => $span->service,
            'version' => $span->version,
```

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1185 @@
...
+                            streamed: true
+                        );
+                    },
+                    'posthook' => function (\DDTrace\SpanData $span, $args, $response) use ($logger, $httpMethod, $endpoint) {
+                        /** @var \OpenAI\Responses\StreamResponse $response */
+                        OpenAIIntegration::handleStreamedResponse(
+                            span: $span,
+                            logger: $logger,
+                            headers: method_exists($response, 'meta') ? $response->meta()->toArray() : [],
+                            response: \is_string($response) ? $response : ($response ? $response->getIterator() : []),
```
bwoebi, line 206: This doesn't look right. handleStreamedResponse takes only \Generator, not string.

tests/Integrations/OpenAI/OpenAITest.php
```diff
@@ -0,0 +1,521 @@
...
+            $response = new Response(
+                headers: $metaHeaders,
+                body: new Stream($responseBodyArray)
+            );
+            $client = mockClient($response);
+            if ($openAIParameters) {
+                $client->{$resource}()->{$openAIFn}($openAIParameters);
+            } else {
+                $client->{$resource}()->{$openAIFn}();
+            }
```
bwoebi, line 103: Please verify the response can also be actually read back by the caller - the integration must not _consume_ it.

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
bwoebi: Also, $response->getIterator() is not rewindable - i.e. It reads the response once. And if it's read, the actual caller cannot read it again. (see comment on the test)

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1220 @@
...
+                        /** @var array{baseUri: string, headers: string, apiKey: ?string} $clientData */
+                        OpenAIIntegration::handleRequest(
+                            span: $span,
+                            methodName: $methodName,
+                            args: $args,
+                            basePath: $clientData['baseUri'],
+                            apiKey: $clientData['apiKey'],
+                            streamed: true
+                        );
+                    },
```
pablomartinezbernardo, line 201: nit: Almost all this (except `streamed`) is duplicated between streamed/non-streamed. May not even be worth to simplify, though

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1220 @@
...
+
+    public static function normalizeRequestPayload(
+        string $methodName,
+        array  $args
+    ): array
+    {
+        switch ($methodName) {
+            case 'listModels':
+            case 'listFiles':
+            case 'listFineTunes':
```
pablomartinezbernardo, line 240: nit: What do you think about creating some constants for these given that they are used in a bunch of places?

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1220 @@
...
+        array    $args,
+        string   $basePath,
+        string   $apiKey,
+        bool     $streamed = false,
+    )
+    {
+        $payload = self::normalizeRequestPayload($methodName, $args);
+
+        $span->name = 'openai.request';
+        $span->resource = isset($payload['model']) ? $methodName . '/' . $payload['model'] : $methodName;
```
pablomartinezbernardo, line 331: I think this doesn't match what the doc defines:

> Resource name:  the resource name shall be: `<endpoint-name>/<model-name>`. For example, using the chat/completions endpoint and the model gpt-3.5-turbo, the resource name would be: `chat.completions/gpt-3.5-turbo`. If the model cannot be retrieved for some reason, the resource name is `<endpoint-name>`.

For example, in the case of chat completions we would end up with `createChatCompletion/gpt-3.5-turbo` instead right?

PROFeNoM: Nice catch

src/DDTrace/Integrations/OpenAI/OpenAIIntegration.php
```diff
@@ -0,0 +1,1220 @@
...
+        $estTokens = 0;
+        if (is_string($prompt)) {
+            $est1 = strlen($prompt) / 4;
+            $est2 = preg_match_all('/[.,!?]/', $prompt) * 0.75;
+            return round((1.5 * $est1 + 0.5 * $est2) / 2);
+        } elseif (is_array($prompt) && is_int($prompt[0])) {
+            return count($prompt);
+        }
+        return $estTokens;
+    }
```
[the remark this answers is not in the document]
PROFeNoM: Tbh, I **blindly** based myself on [Python's estimate function](https://github.com/DataDog/dd-trace-py/blob/6bf5db3adb07b08b52f3c017419356386b36ed0a/ddtrace/contrib/openai/utils.py#L154-L173) (which is different from the v1 document).

bwoebi (approved):

PR #2689, opened 2024-06-03
Allow synchronous sidecar flushes and reduce flush limit

bwoebi:
- Synchronous flushing for testing.
- Reduced flush interval for sidecar.
- Reduced flush limit to avoid the sidecar ballooning up memory wise.
- Respect DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS with the sidecar active.

### DIFF
--- a/ext/auto_flush.c
+++ b/ext/auto_flush.c
@@ -66,7 +66,12 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
                                 .client_computed_top_level = false,
                                 .client_computed_stats = false,
                         };
-                        ddog_MaybeError send_error = ddog_sidecar_send_trace_v04_shm(&ddtrace_sidecar, ddtrace_sidecar_instance_id, shm, written, &tags);
+                        size_t size_hint = written;
+                        zend_long n_requests = get_global_DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS();
+                        if (n_requests) {
+                            size_hint = MAX(get_global_DD_TRACE_AGENT_MAX_PAYLOAD_SIZE() * get_global_DD_TRACE_BETA_HIGH_MEMORY_PRESSURE_PERCENT() / 100 / n_requests, size_hint);
+                        }
+                        ddog_MaybeError send_error = ddog_sidecar_send_trace_v04_shm(&ddtrace_sidecar, ddtrace_sidecar_instance_id, shm, size_hint, &tags);
                         do {
                             if (send_error.tag == DDOG_OPTION_ERROR_SOME_ERROR) {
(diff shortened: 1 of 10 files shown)


iamluc (approved):
Code LGTM.
But alternatively, what about passing the FLUSH_AFTER_N_REQUESTS config to the sidecar and let the trace flusher handles it?

ext/auto_flush.c
```diff
@@ -66,7 +66,12 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
                                 .client_computed_top_level = false,
                                 .client_computed_stats = false,
                         };
-                        ddog_MaybeError send_error = ddog_sidecar_send_trace_v04_shm(&ddtrace_sidecar, ddtrace_sidecar_instance_id, shm, written, &tags);
+                        size_t size_hint = written;
+                        zend_long n_requests = get_global_DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS();
+                        if (n_requests) {
+                            size_hint = MAX(get_global_DD_TRACE_AGENT_FLUSH_SIZE() / n_requests, size_hint);
```
bantonsson, line 72: I'm not too fond of this logic living in the tracer instead of having a flush after N request setting in the side car itself, but on the other hand I'm not sure if that setting would make sense for the general data pipeline.

ext/configuration.h
```diff
@@ -171,6 +174,7 @@ enum ddtrace_sampling_rules_format {
     CONFIG(CUSTOM(STRING), DD_TRACE_CLIENT_IP_HEADER, "", .parser = ddtrace_parse_client_ip_header_config)     \
     CONFIG(BOOL, DD_TRACE_FORKED_PROCESS, "true")                                                              \
     CONFIG(INT, DD_TRACE_HOOK_LIMIT, "100")                                                                    \
+    CONFIG(INT, DD_TRACE_AGENT_FLUSH_SIZE, "1048576", .ini_change = zai_config_system_ini_change)              \
```
pierotibou, line 177: In [.NET](https://github.com/search?type=code&q=org%3ADataDog+DD_TRACE_BUFFER_SIZE) we use `DD_TRACE_BUFFER_SIZE`. I'd recommend using the same for consistency

pierotibou (approved):
LGTM, one comment about the new variable name

ext/configuration.h
bantonsson, line 177: A more sane number than 40Mb 👍🏻

bantonsson (approved):

ext/auto_flush.c
bwoebi: Same here, that's why I did it this way.

PR #2692, opened 2024-06-05
feat(profiling): `thread id` label on every sample

realFlowControl:
### Description

Turns out the profiling-backend assumes that all pprof based profiles emit a `thread id` and `thread name` label with every sample, so this PR aims to do this. Edit: profiling-backend was updated to not require the thread name, but we should add at least `thread id` anyway. Since thread names are uncommon (we don't know of a single PHP program/library which names threads), we omit `thread name` for now.

Took this as an opportunity to clean up a few things:

- `message_labels` was renamed to `common_labels` and now sets the "fiber" tag (was in `prepare_message` before). It also accepts a parameter for how much extra space to allocate in the `Vec`, so callers can avoid reallocations.
- Avoided `lazy_static` labels, and removed un-necessary `vec!` usages to reduce allocations.
- We always call `send_message(prepare_message(...))`, so I made a helper for this called `prepare_and_send_message`.

PROF-9887

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -37,8 +37,6 @@ use std::thread::JoinHandle;
 use std::time::{Duration, Instant, SystemTime};
 
 #[cfg(feature = "timeline")]
-use lazy_static::lazy_static;
-#[cfg(feature = "timeline")]
 use std::time::UNIX_EPOCH;
 
 #[cfg(feature = "allocation_profiling")]
@@ -681,6 +679,7 @@ impl Profiler {
 
     /// Collect a stack sample with elapsed wall time. Collects CPU time if
     /// it's enabled and available.
+    #[inline]
     pub fn collect_time(&self, execute_data: *mut zend_execute_data, interrupt_count: u32) {
         // todo: should probably exclude the wall and CPU time used by collecting the sample.
         let interrupt_count = interrupt_count as i64;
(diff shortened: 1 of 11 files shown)


profiling/src/profiling/mod.rs
```diff
@@ -1000,6 +1000,24 @@ impl Profiler {
...
+                value: LabelValue::Num(unsafe { libc::pthread_self() as i64 }, "id".into()),
+            },
+            Label {
+                key: "thread name",
+                value: LabelValue::Str(
+                    thread::current()
+                        .name()
+                        .unwrap_or("Main")
+                        .to_string()
+                        .into(),
```
morrisonlevi, line 1016: I'm pretty sure PHP doesn't name its threads on any usable level in web SAPIs. In ZTS, this means we're going to get like 4+ threads called "Main".  Maybe we should call this "Thread {tid}"?

Second, I don't know if Rust's thread names are what we want. I know that Rust conveys thread names to the OS, but I don't think the reverse happens. If someone actually names a thread with `pthread_setname_np`, will Rust pick it up? We should check.

morrisonlevi: It doesn't.

realFlowControl:
@morrisonlevi I can not approve but I approve
image

morrisonlevi (approved):

PR #2694, opened 2024-06-05
Hack to massage sidecar v0.7 into what is expected by tests

bantonsson:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -969,7 +969,7 @@ TEST_WEB_83 := \
 	test_web_custom \
 	test_web_zend_1_21
 
-FILTER := .
+FILTER := $(if $(FILTER),$(FILTER),".")
 MAX_RETRIES := 3
 
 # Note: The "composer show" command below outputs a csv with pairs of dependency;version such as "phpunit/phpunit;9.6.17"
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -969,7 +969,7 @@ TEST_WEB_83 := \
 	test_web_custom \
 	test_web_zend_1_21
 
-FILTER := .
+FILTER := $(if $(FILTER),$(FILTER),".")
```
bantonsson, line 972: This allows me to run a single test this way:
```
TESTS="tests/Integration/TracerTest.php" FILTER="testResourceNormalizationWebDefault" make test_integration
```
I don't know if there is a better way 🤷🏼‍♂️

bantonsson:
So the state right now is that tests keep getting weird array failures like this:
```
1) DDTrace\Tests\Integration\TracerTest::testResourceNormalizationWebDefault
Undefined array key "resource"

/home/circleci/app/tests/Integration/TracerTest.php:187
```

But if I insert a print line before the assert
```
        error_log('Traces v07: ' . print_r($traces, 1));
```
Then the test works 🤷🏼‍♂️

Makefile
bantonsson: Even better. You'll be happy to know that you beat _copilot_ 😉

Issue #2599, opened 2024-03-26
[Bug]: 0.98.x tar.gz file size increased?


brandonburkett:
### Bug report

With version 0.97 and below, the tar.gz file size was around 135mb, but with 0.98 release, I see it is up to 400+mb.  Our elastic beanstalk default /tmp only has `924M` without attempting to give /tmp more space.  

In short, `php datadog-setup.php --php-bin=all` is failing because /tmp is out of disk space while trying to untar, which was not a problem with 0.97 and below.

### PHP version

8.2.15

### Tracer or profiler version

0.98.x

bwoebi:
Hey @brandonburkett,

We've decided to keep full debug symbols of the rust code to enhance debuggability in case there are issues. However, this comes, as you noticed at the cost of a size increase.

To avoid the particular issue with /tmp space being exceeded, you can specify a different temp directory: `php -d sys_temp_dir=/other/path datadog-setup.php --php-bin=all`.

I'll keep this issue open so that we invest into splitting the .tar.gz up or using other strategies.

awons-tqg:
Shouldn't this be an opt-in feature? I see the point, but our Docker images exploded from ~100 MB to over 600 MB. It's not the end of the world, but this could be something users could decide on their own.

morrisonlevi:
I slightly improved the situation in v1.1.0. Here are the sizes for the `aarch64-linux-gnu` artifacts over time, but there are similarly sized improvements for `x86_64-linux-gnu`:

```
dd-library-php-0.98.0-aarch64-linux-gnu.tar.gz  482 MB
dd-library-php-0.99.0-aarch64-linux-gnu.tar.gz  485 MB
dd-library-php-1.0.0-aarch64-linux-gnu.tar.gz   495 MB

dd-library-php-1.1.0-aarch64-linux-gnu.tar.gz   425 MB
```

It's still over 400 MiB, so we plan to continue working on this.

awons:
We worked around it by copying only the necessary files and using them in our final Docker image with this script:

```bash
cd $(mktemp -d)
curl -Lo datadog-php-tracer.apk https://github.com/DataDog/dd-trace-php/releases/download/${DATADOG_TRACER_VERSION}/datadog-php-tracer_${DATADOG_TRACER_VERSION}_$(uname -m).apk
apk add datadog-php-tracer.apk --allow-untrusted
[...]

PR #2696, opened 2024-06-07
build: shrink size of Rust artifacts

morrisonlevi:
### Description

This is an attempt to shrink artifacts, but unlike #2601,  this does not do `panic=abort`. For `panic=abort` to be really effective, you also need to rebuild std from source. That means rebuilding images and also installing Rust from rustup (because the packages we are using do not include std sources). We had issues with CI last time we tried it. We played whack-a-mole for a while, but since the issues just kept going and we wanted to release 1.0.0beta1, we decide to revert back.

But this PR without `panic=abort` will still shrink the size a fair bit, so it's worth splitting out. I wouldn't go so far as saying that it will fix the size issue reported in PR #2599, but it helps.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,9 +13,10 @@ license = "Apache-2.0"
 debug = 2 # full debug info
 
 [profile.release]
-debug = 1
-lto = "thin"
+codegen-units = 1
+debug = "line-tables-only"
 incremental = false
+lto = "fat"
 
 [profile.tracer-release]
 debug = 1 # line tables only


morrisonlevi:
Difference between the commit on master and this branch:

```
# This branch
-rw-r--r-- 1 levi.morrison staff 425M Jun  6 21:30 dd-library-php-1.0.0+76dd2bf00d17575ba6c41682e29f419dd7795668-aarch64-linux-gnu.tar.gz
-rw-r--r-- 1 levi.morrison staff 428M Jun  6 21:30 dd-library-php-1.0.0+76dd2bf00d17575ba6c41682e29f419dd7795668-x86_64-linux-gnu.tar.gz

# master
-rw-r--r-- 1 levi.morrison staff 496M Jun  6 20:14 dd-library-php-1.0.0+e63625078edc52eb92e4c6104f0daaab3ec8fa63-aarch64-linux-gnu.tar.gz
-rw-r--r-- 1 levi.morrison staff 500M Jun  6 20:11 dd-library-php-1.0.0+e63625078edc52eb92e4c6104f0daaab3ec8fa63-x86_64-linux-gnu.tar.gz
```

Roughly 15% savings.

realFlowControl (approved):

PR #2698, opened 2024-06-08
Ensure the msgpack contents include only valid utf8

bwoebi:
Ensure no utf8 is contained in msgpack strings.

The agent would properly sanitize this away, but the sidecar is not. All traces containing a single invalid utf-8 sequence were dropped.

### DIFF
--- a/ext/serializer.c
+++ b/ext/serializer.c
@@ -47,10 +47,10 @@ extern void (*profiling_notify_trace_finished)(uint64_t local_root_span_id,
                                                zai_str resource);
 
 static void mpack_write_utf8_lossy_cstr(mpack_writer_t *writer, const char *str, size_t len) {
-    char *utf8 = ddtrace_strip_invalid_utf8(str, &len);
-    if (utf8) {
-        mpack_write_str(writer, utf8, len);
-        ddtrace_drop_rust_string(utf8, len);
+    char *strippedStr = ddtrace_strip_invalid_utf8(str, &len);
+    if (strippedStr) {
+        mpack_write_str(writer, strippedStr, len);
+        ddtrace_drop_rust_string(strippedStr, len);
     } else {
         mpack_write_str(writer, str, len);
     }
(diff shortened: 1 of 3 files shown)


ext/serializer.c
```diff
@@ -46,6 +46,16 @@ extern void (*profiling_notify_trace_finished)(uint64_t local_root_span_id,
                                                zai_str span_type,
                                                zai_str resource);
 
+static void mpack_write_utf8_lossy_cstr(mpack_writer_t *writer, const char *str, size_t len) {
+    char *utf8 = ddtrace_strip_invalid_utf8(str, &len);
+    if (utf8) {
+        mpack_write_str(writer, utf8, len);
+        ddtrace_drop_rust_string(utf8, len);
+    } else {
+        mpack_write_str(writer, str, len);
```
pierotibou, line 55: We were doing `mpack_write_cstr` before the change, what's the diff with `mpack_write_str`?

ext/serializer.c
```diff
@@ -46,6 +46,16 @@ extern void (*profiling_notify_trace_finished)(uint64_t local_root_span_id,
                                                zai_str span_type,
                                                zai_str resource);
 
+static void mpack_write_utf8_lossy_cstr(mpack_writer_t *writer, const char *str, size_t len) {
+    char *utf8 = ddtrace_strip_invalid_utf8(str, &len);
+    if (utf8) {
+        mpack_write_str(writer, utf8, len);
+        ddtrace_drop_rust_string(utf8, len);
```
pierotibou, line 53:
```suggestion
    char *strippedStr = ddtrace_strip_invalid_utf8(str, &len);
    if (strippedStr) {
        mpack_write_str(writer, strippedStr, len);
        ddtrace_drop_rust_string(strippedStr, len);
```

pierotibou: Feel free to ignore given the short timeframe, I thought it was clearer though

ext/serializer.c
bwoebi: cstr just does strlen(str) as length - it's virtually identical.

pierotibou (approved):
Cool thanks for the offline explanations and for adding the change behind the sidecar flag.

PR #2709, opened 2024-06-11
perf(profiling): shrink the size of ZendFrame

morrisonlevi:
PROF-9123

### Description

Previously, each frame was 7 * 64 bits. They are now 3 * 64 bits, at the cost that every `String` gets copied to a `ThinString`. This PR is an attempt to see if this trade-off is worth it.

Currently stacked.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (commented):
A few notes about how the total number of copies doesn't change, despite using the `ThinString` type which allocates a str copy.

PR #2710, opened 2024-06-12
Add http.route tag to SymfonyIntegration.php

Julian-Louis:
Hello, this is draft pull request adds the http.route tag to the Symfony integration.
Noticing that [PR #2477](https://github.com/DataDog/dd-trace-php/pull/2477) was closed, I decided to take the initiative to implement this feature.

Can you please provide your feedback on this proposal/implementation? If approved, I'll proceed with implementing tests to further validate the changes.

Thank you for your review !

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
+++ b/src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
@@ -300,11 +300,12 @@ class SymfonyIntegration extends Integration
         \DDTrace\trace_method(
             'Symfony\Component\EventDispatcher\EventDispatcher',
             'dispatch',
-            function (DDTrace\SpanData $span, $args) {
+            function (SpanData $span, $args) {
         
                 list($event) = $args;
-        
-                if (!$event instanceof ControllerEvent) {
+
+                $controllerEventClass = '\Symfony\Component\HttpKernel\Event\ControllerEvent';
+                if (!$event instanceof $controllerEventClass) {
                     return;
                 }
                 
--- a/src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
+++ b/src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
@@ -332,7 +332,11 @@ class SymfonyIntegration extends Integration
         
                 // Replace route parameters in URL with placeholders
                 foreach ($parameters as $key => $value) {
-                    $url = str_replace($value, '{' . $key . '}', $url);
+                    $search = '/' . preg_quote($value, '/') . '/';
+                    if (empty($value)) {
+                        continue;
+                    }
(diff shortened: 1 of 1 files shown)


bwoebi:
Hey @Julian-Louis,

thanks for taking the initiative here! I think this approach may not be fully correct.

Like, normalization is basically trying to recognize ids, but if you have routes which contain a name, not matching these patterns, the parameter will be assigned wrongly, I think.

E.g. let's assume a route `/posts/{category}/{id}` - we do a request `/posts/cooking/12`.
We'll end up with a normalized form `/posts/cooking/?`.
The parameters being replaced, it will be `/posts/cooking/{category}`, which is not right.

Julian-Louis:
Hello, thank you for your feedback !
I understand your concerns and will explore alternative implementations. I'll update the code as soon as I can

Julian-Louis:
Hello @bwoebi,

I revisited my code and implemented a new approach! By leveraging a reflection class, I gained access to the Symfony container, allowing me to utilize the router matcher. This allows me to accurately retrieve the actual parameters and URL of the route. And I normalized the URL by replacing parameter keys with corresponding values in the URL.

What do you think of it ?

bwoebi:
Hmm @Julian-Louis, I think this code has another missing case:

Now a route `/posts/{category}/{id}` matching `/posts/1/1` will be converted to `/posts/{category}/{category}`, if I read that code correctly?

I've been looking through the Symfony code right now, and am finding Router::getRouteCollection(), which contains all routes with their names and paths. Either the $router->options['cache_dir'] is set and the Router::getRouteCollection() is loaded if uncached or it's not set and it's always present.
So, isn't it simply possible to fill our own cache file with our own name=>paths mapping and use this? (if cache is enabled, otherwise, we'll just walk the routecollection directly?

A direct mapping from a cache file should be quite fast then I suppose?

Note that I have no idea what I'm talking about though, not an expert in Symfony, just navigating through their codebase and trying to figure out what's going on there.

Julian-Louis:
@bwoebi 
I just updated my code and now `/posts/{category}/{id}` (`/posts/1/1`) will be converted to `/posts/{category}/{id}` instead of `/posts/{category}/{category}` anymore, sorry for the mistake.

I'm not a big symfony expert too, but I'll try to take a look on the way you suggest

bwoebi:
I think the current solution is probably _viable_. It's not 100% correct (e.g. `/posts/posts` with a route `/posts/{name}` will be now `/{name}/posts`).
I'm not sure if that's the only edge case, it's maybe edge case enough to be accepted.

I also have no idea of the performance characteristics; I'll leave that to somebody else from my team to evaluate :-)

cataphract:
Superseded by https://github.com/DataDog/dd-trace-php/pull/2992 . I based it on your work. Thank you.

PR #2714, opened 2024-06-14
perf(profiling) remove profiler mutex

realFlowControl:
### Description

This PR removes the `Mutex` around the `Profiler` instance. Additionally I moved the lifecycle management into the `Profiler` object itself. 

PROF-9531

### Reviewer checklist
- [x] Test coverage seems ok.


profiling/src/profiling/mod.rs
```diff
@@ -709,9 +714,32 @@ impl Profiler {
...
+    /// and threads are not copied when the process is forked.
+    /// Additionally, if we've hit certain other issues like not being able to
+    /// determine the return type of the pcntl_fork function, we don't know if
+    /// we're the parent or child.
+    /// So, we throw away the current profiler and forget it, which avoids
+    /// running the destructor. Yes, this will leak some memory.
+    ///
+    /// # Safety
+    /// Must be called when no other thread is using the PROFILER object. That
+    /// includes this thread in some kind of recursive manner.
```
realFlowControl, line 730: thanks for this very good comment 🙇

realFlowControl:
@morrisonlevi fixed the merge conflicts 🎉

bwoebi (approved):
I suppose that's fine as the global profiler never gets a mut reference to it, i.e. no unsoundness is being introduced. (apart from kill, which is unsafe itself)

realFlowControl:
@morrisonlevi can you have another look?

profiling/src/profiling/mod.rs
```diff
@@ -510,7 +516,40 @@ pub enum UploadMessage {
...
+static PROFILER_INIT: Mutex<()> = Mutex::new(());
+
 impl Profiler {
+    /// Will initialize the `PROFILER` OnceCell and makes sure that only one thread will do so.
+    pub fn init(system_settings: &mut SystemSettings) {
+        // Safety: there could be a race in which another thread already initializes `PROFILER`,
+        // this is guarded by the `PROFILER_INIT` mutex next.
+        if unsafe { PROFILER.get() }.is_some() {
+            return;
+        }
```
morrisonlevi, line 531: This is definitely race condition. Whether it's one of those harmless ones is a bit trickier for me to think about because I don't know Rust's internals of these things as well as I do C's, where I spent most of my time learning correct parallelism. I assume you wish to avoid touching the lock if it's already initialized for performance reasons?

realFlowControl: > I assume you wish to avoid touching the lock if it's already initialized for performance reasons?

Exactly

morrisonlevi: I think you don't need to worry about it, on Linux the `futex` uses an atomic relaxed load to determine if the lock is contended: https://github.com/rust-lang/rust/pull/95035/files#diff-823e207a741b9946be4e20738fd2cfe16e077beb001632aaa8623e3ea0803cc0. I think you can just delete these lines.

morrisonlevi: I guess there's still the mutex poisoning wrapper around it. I wonder if we should use `parking_lot`'s `Mutex` instead.

realFlowControl: I need to be more precise about this code maybe:
- this is called from `RINIT` just to make it explicit (we could also check in the `Profiler::get()` method if the `Profiler` is initialised and initialise it in case it is not, but this way is more explicit)
- in case the Profiler is already initialised (`PROFILER` is some), I do not need to to anything anymore
- in case it is not, I need to initialise it, this is where the `PROFILER_INIT` mutex comes into play, because I only want one thread to actually initialise the `Profiler` and all others to wait

PR #2722, opened 2024-06-19
experiment: does a missing zend_execute_internal cause a crash?

morrisonlevi:
### Description

In another PR I seemed to be getting crashes in the sequence of events:

1. `dd_uhook_end` on an internal function
2. `ddog_php_prof_interrupt_function`
3. `Profiler ::collect_time`

And then I OOM, sometimes in Zend Memory Manager but sometimes in Rust.

I'm trying to isolate this from the other changes to see if there's just an issue here somewhow.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi:
Seems like... no. Something else in the patch must be messing it up.

PR #2727, opened 2024-06-20
fix: Add `ddsource` attribute to json-formatted logs

PROFeNoM:
### Description

**Problem:** The PHP Log pipeline is not automatically used to process PHP logs ==> No Log Correlation by default, despite the trace identifiers being injected.

Two solutions:
1. We inject the trace identifiers outside of `context`
2. We use `ddsource`

Either way, it seems like we should have always been using ddsource + it is more elegant and is the actual way of doing it.

TBD - I gotta double test this on my sandbox. generating the artifact...

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/Logs/LogsIntegration.php
+++ b/src/DDTrace/Integrations/Logs/LogsIntegration.php
@@ -237,6 +237,17 @@ class LogsIntegration extends Integration
         );
         \DDTrace\remove_hook($hook, NullLogger::class);
 
+        \DDTrace\install_hook(
+            'Monolog\Formatter\JsonFormatter::toJson',
+            function (HookData $hook) {
+                /** @var array $normalized */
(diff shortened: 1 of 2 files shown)


src/DDTrace/Integrations/Logs/LogsIntegration.php
```diff
@@ -237,6 +237,17 @@ public function init(): int
         );
         \DDTrace\remove_hook($hook, NullLogger::class);
 
+        \DDTrace\install_hook(
+            'Monolog\Formatter\JsonFormatter::toJson',
+            function (HookData $hook) {
+                /** @var array $normalized */
+                $normalized = $hook->args[0];
+                $normalized['ddsource'] = 'php';
```
PROFeNoM, line 245:
```suggestion
                $normalized['source'] = 'php';
```

PR #2728, opened 2024-06-21
Fix global priority sampling

bwoebi:
The global priority sampling was not recognized as explicit sampling and thus ignored.


iamluc (approved):

pablomartinezbernardo (approved):

pablomartinezbernardo (approved):
I'm sure you have already seen it, but some snapshots need updating it seems

PR #2731, opened 2024-06-25
feat: Source Code Integration

PROFeNoM:
AIDM-195

### TODO
- [X] [Check system-tests](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/16486/workflows/e5c4ed5e-c6e8-490a-8304-cac1da821630/jobs/4642566?invite=true#step-109-33)
image

### Description

Add three environment variables:

- `DD_GIT_REPOSITORY_URL` string, default: empty
- `DD_GIT_COMMIT_SHA` string, default: empty
- `DD_TRACE_GIT_METADATA_ENABLED` bool, default: true

Add following steps:
1. Check if `DD_TRACE_GIT_METADATA_ENABLED` is True.
2. Check if git metadata is already provided by environment variables:
    - Check `DD_GIT_REPOSITORY_URL` and `DD_GIT_COMMIT_SHA`, if they exist, use metadata from them.
    - Check `DD_TAGS`, if it contains git.repository_url and git.commit.sha, use metadata from them.
3. Exract git metadata from the `git` ~binary~ HEAD/refs/config files

Add metadata tags to the first span of each chunk:
- `_dd.git.repository_url` for repository URL
- `_dd.git.commit.sha` for commit hash

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/git.c
```diff
@@ -0,0 +1,218 @@
...
+
+    zend_string_release(git_commit_sha);
+    zend_string_release(git_repository_url);
+
+    return result;
+}
+
+void ddtrace_inject_git_metadata(zval* meta, bool is_root_span) {
+    ddtrace_git_metadata git_metadata = DDTRACE_G(git_metadata);
+    if (git_metadata.called_once) {
```
bwoebi, line 207: Do I see read it correctly, that this is only ever called once for the whole thread?
I suppose having multiple virtual hosts with multiple git repositories on an apache server is not that uncommon. So I suppose we'll need some HashMap based caching here?

ext/serializer.c
```diff
@@ -1504,6 +1505,10 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
     zend_array *meta = ddtrace_property_array(&span->property_meta);
     zend_array *metrics = ddtrace_property_array(&span->property_metrics);
 
+    if (get_DD_TRACE_GIT_METADATA_ENABLED()) {
+        ddtrace_inject_git_metadata(&span->property_meta, is_root_span);
+    }
+
```
bwoebi, line 1511: I think we should inject these metadata on span creation. I.e. put it into ddtrace_set_root_span_properties. Children probably should inherit it from the root span (just like service etc. are also inherited) in ddtrace_inherit_span_properties then.

Or when programmatically assigning to root span.

ext/git.c
PROFeNoM: > By the way, called_once seems never set to true

Oops, I think I deleted the line 😬

ext/span.h
```diff
@@ -170,6 +171,21 @@ struct ddtrace_span_link {
...
+            zval property_commit;
+            zval property_repository;
+        };
+    };
+};
+
+static inline ddtrace_git_metadata *OBJ_GITMETADATA(zend_object *obj) {
+    return (ddtrace_git_metadata *)((char *)obj - XtOffsetOf(ddtrace_git_metadata, std));
+}
+
```
bwoebi, line 188:
```suggestion
```

Unused and not needed :-)

ext/ddtrace.stub.php
```diff
@@ -139,6 +151,11 @@ class SpanData {
          */
         public array $peerServiceSources = [];
 
+        /**
+         * @var GitMetadata|null The git metadata of the span
+         */
+        public GitMetadata|null $gitMetadata = null;
+
```
bwoebi, line 158: I think it should be on the trace local root span, i.e. RootSpanData only. No point in propagating.

ext/serializer.c
```diff
@@ -1308,6 +1316,18 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
...
+    zval *git_metadata = &span->property_git_metadata;
+    if (git_metadata && Z_TYPE_P(git_metadata) == IS_OBJECT) {
+        ddtrace_git_metadata *metadata = (ddtrace_git_metadata *)Z_OBJ_P(git_metadata);
+        if (is_root_span) {
+            add_assoc_str(meta, "_dd.git.commit.sha", zend_string_copy(Z_STR_P(&metadata->property_commit)));
+            add_assoc_str(meta, "_dd.git.repository_url", zend_string_copy(Z_STR_P(&metadata->property_repository)));
+        } else {
+            add_assoc_str(meta, "git.commit.sha", zend_string_copy(Z_STR_P(&metadata->property_commit)));
+            add_assoc_str(meta, "git.repository_url", zend_string_copy(Z_STR_P(&metadata->property_repository)));
+        }
```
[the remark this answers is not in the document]
PROFeNoM: The honest answer is that I don't know yet. The system tests aren't defining what should happen. I only _assumed_ they shouldn't be there as t[he Python tracer explicitly tests that these values shouldn't exist on the root span ](https://github.com/DataDog/dd-trace-py/blob/203e89489f1e698a127317dc57b01e9d6467e859/tests/tracer/test_gitmetadata.py#L68-L72).

I'll try to find someone to tell me whether this is an issue in the backend.

ext/git.c
```diff
@@ -115,104 +137,88 @@ void removeCredentials(zend_string* repo_url) {
...
+void ddtrace_inject_git_metadata(zval *git_metadata_zv) {
+    zend_string *cwd_zstr = NULL;
 
-    if (!git_commit_sha || !git_repository_url) {
-        if (git_commit_sha) zend_string_release(git_commit_sha);
-        if (git_repository_url) zend_string_release(git_repository_url);
-        return false;
+    if (SG(options) & SAPI_OPTION_NO_CHDIR) {
+        const char *script_filename = SG(request_info).path_translated;
+        cwd_zstr = zend_string_init(script_filename, strlen(script_filename), 1);
```
bwoebi, line 192: path_translated is the full path, not just the dir, you'll need to strip the filename.

ext/git.c
```diff
@@ -115,104 +137,88 @@ void removeCredentials(zend_string* repo_url) {
...
-        inject_from_global_tags(meta, is_root_span) ||
-        inject_from_git_files(meta, is_root_span)) {
+    if (entry) {
+        zend_string_release(cwd_zstr);
+        ZVAL_COPY(git_metadata_zv, entry);
         return;
     }
+
+    if (inject_from_env(git_metadata_zv) || inject_from_global_tags(git_metadata_zv) || inject_from_git_dir(git_metadata_zv, cwd_zstr)) {
+        zend_hash_add(&DDTRACE_G(git_metadata), cwd_zstr, git_metadata_zv);
```
bwoebi, line 215: It's highly forbidden to persist objects across requests. You may only persist raw strings across a request.

ext/git.c
```diff
@@ -198,27 +202,30 @@ void ddtrace_inject_git_metadata(zval *git_metadata_zv) {
...
-    if (inject_from_env(git_metadata_zv) || inject_from_global_tags(git_metadata_zv) || inject_from_git_dir(git_metadata_zv, cwd_zstr)) {
-        zend_hash_add(&DDTRACE_G(git_metadata), cwd_zstr, git_metadata_zv);
+        if (entry->available) {
+            add_git_info(git_metadata_zv, entry->property_commit, entry->property_repository);
+        }
+    } else if (inject_from_env(git_metadata_zv) || inject_from_global_tags(git_metadata_zv) || inject_from_git_dir(git_metadata_zv, cwd_zstr)) {
+        struct _git_metadata *git_metadata = pemalloc(sizeof(struct _git_metadata), 1);
+        git_metadata->available = true;
+        zend_string *property_commit = Z_STR_P(&((ddtrace_git_metadata *) Z_OBJ_P(git_metadata_zv))->property_commit);
+        zend_string *property_repository = Z_STR_P(&((ddtrace_git_metadata *) Z_OBJ_P(git_metadata_zv))->property_repository);
```
bwoebi, line 215:
```suggestion
        zend_string *property_commit = Z_STR(((ddtrace_git_metadata *) Z_OBJ_P(git_metadata_zv))->property_commit);
        zend_string *property_repository = Z_STR(((ddtrace_git_metadata *) Z_OBJ_P(git_metadata_zv))->property_repository);
```
These all have equivalents without _P too, saving the dereference.

ext/ddtrace.c
```diff
@@ -453,6 +453,7 @@ static void ddtrace_activate(void) {
     ddtrace_telemetry_rinit();
     zend_hash_init(&DDTRACE_G(traced_spans), 8, unused, NULL, 0);
     zend_hash_init(&DDTRACE_G(tracestate_unknown_dd_keys), 8, unused, NULL, 0);
+    zend_hash_init(&DDTRACE_G(git_metadata), 8, unused, NULL, 1);
```
bwoebi, line 456: I think this now requires a destructor to free the memory (release the zend string and the struct itself).

ext/git.c
```diff
@@ -0,0 +1,231 @@
...
+void ddtrace_inject_git_metadata(zval *git_metadata_zv) {
+    zend_string *cwd_zstr = NULL;
+
+    if (SG(options) & SAPI_OPTION_NO_CHDIR) {
+        const char *script_filename = SG(request_info).path_translated;
+        const char *last_slash = strrchr(script_filename, '/');
+        if (last_slash) {
+            cwd_zstr = zend_string_init(script_filename, last_slash - script_filename, 1);
+        } else {
+            cwd_zstr = zend_string_init(script_filename, strlen(script_filename), 1);
```
bwoebi, line 195:
```suggestion
            cwd_zstr = zend_string_init(ZEND_STRL("."), 1);
```
If there's no slash, it means it's relative to the current dir.

ext/git.c
```diff
@@ -0,0 +1,231 @@
...
+
+    return success;
+}
+
+void ddtrace_inject_git_metadata(zval *git_metadata_zv) {
+    zend_string *cwd_zstr = NULL;
+
+    if (SG(options) & SAPI_OPTION_NO_CHDIR) {
+        const char *script_filename = SG(request_info).path_translated;
+        const char *last_slash = strrchr(script_filename, '/');
```
bwoebi, line 191: You'll have to handle `\\` (in _addition_ to `/`) `#ifdef _WIN32`.

ext/serializer.c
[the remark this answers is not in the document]
PROFeNoM: FYI, I Removed it as it seemingly doesn't matter. _dd.git is mapped to git

ext/git.c
```diff
@@ -0,0 +1,339 @@
...
+    ZEND_HASH_FOREACH_STR_KEY_PTR(&DDTRACE_G(git_metadata), cwd, git_metadata) {
+        refresh_git_metadata_if_needed(cwd, git_metadata);
+    }
+    ZEND_HASH_FOREACH_END();
+
+    request_cache_updated = true;
+}
+
+void ddtrace_inject_git_metadata(zval *carrier) {
+    if (!request_cache_updated) {
```
bwoebi, line 285: When is this value ever reset?

PROFeNoM: Well, it was in `ddtrace_git_metadata_rinit`, but my fat finger must have acted again and removed this line.

ext/git.c
bwoebi: _in addition_, you may also see / in paths on Windows. Like both are valid separators. As far as I gather from the code, it's not normalized for you.

ext/serializer.c
```diff
@@ -1307,6 +1321,18 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span) {
         EG(exception) = current_exception;
     }
 
+    zval *git_metadata = &span->root->property_git_metadata;
+    if (git_metadata && Z_TYPE_P(git_metadata) == IS_OBJECT) {
+        ddtrace_git_metadata *metadata = (ddtrace_git_metadata *)Z_OBJ_P(git_metadata);
+        if (is_root_span) {
+            add_assoc_str(meta, "_dd.git.commit.sha", zend_string_copy(Z_STR_P(&metadata->property_commit)));
```
bwoebi, line 1328: Please use `ddtrace_convert_to_str` for safety when reading user-mutable data. It will just do a copy when it's a string, but handle the case where somebody assigned something else instead of crashing.
```suggestion
            add_assoc_str(meta, "_dd.git.commit.sha", ddtrace_convert_to_str(&metadata->property_commit));
```

tests/Integrations/Custom/Autoloaded/GitMetadataTest.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    protected static function getAppIndexScript()
+    {
+        return __DIR__ . '/../../../Frameworks/Custom/Version_Autoloaded/public/index.php';
+    }
+
+    protected static function getEnvs()
+    {
+        return array_merge(parent::getEnvs(), [
+            'DD_TRACE_GIT_METADATA_ENABLED' => 'true',
```
bwoebi, line 18: Can you please extend this test to test re-loading of the .git metadata? I.e. put a fake `.git` dir in the folder of the script during the testing, change the HEAD and observe whether the changes are picked up?

ext/serializer.c
```diff
@@ -966,6 +976,10 @@ void ddtrace_set_root_span_properties(ddtrace_root_span_data *span) {
             ZVAL_DOUBLE(&sample_rate, web_integration->get_sample_rate());
             zend_hash_str_add_new(metrics, ZEND_STRL("_dd1.sr.eausr"), &sample_rate);
         }
+
+        if (get_DD_TRACE_GIT_METADATA_ENABLED() && !is_opcache_preload()) {
```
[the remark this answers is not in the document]
PROFeNoM: The [opcache preload test](https://github.com/DataDog/dd-trace-php/blob/44a07ba98de2062986a33cc1592a2b87533441a0/tests/opcache/preload.phpt#L4) was otherwise having a memory leak.

The leaked memory was the allocated memory by the cached zend strings initialized during the inject_from_git_dir procedure.

Why was it leaking?
The order of events was:
- The request is initialized
- The git metadata procedure is made and the strings allocated and cached
- _Something happens, resetting the content of the git_metadata global_
  - A root span is generated (cli.command), but it is not flushed because it doesn't get serialized since `DDTRACE_G(top_closed_stack)` is `NULL` during `ddtrace_serialize_closed_spans_with_cycle` 
- The request is re-initialized
- The git metadata procedure is made, and other strings are allocated and cached
- The request is made and spans serialized
- The git_metadata structure is cleaned
==> The "first" allocated git_metadata content wasn't freed.

From my point of view, I had two choices:
- Ensure the memory is freed so that we don't have a leak
- Don't inject the git metadata, since it anyway won't get serialized

Considering I couldn't find out how to do the first option, I defaulted to the second one.

ext/git.c
```diff
@@ -0,0 +1,339 @@
...
+
+    while (true) {
+        char git_dir[PATH_MAX];
+#ifndef _WIN32
+        snprintf(git_dir, sizeof(git_dir), "%s/.git", current_dir);
+#else
+        snprintf(git_dir, sizeof(git_dir), "%s\\.git", current_dir);
+#endif
+
+        if (access(git_dir, F_OK) == 0) {
```
bwoebi, line 139: You may choose to handle EPERM differently from other errors? Like short-circuit here. If you're not allowed to access, it means you're in a git dir already, just, no git metadata for you.

ext/git.c
```diff
@@ -0,0 +1,339 @@
...
+    return zend_string_init(buffer, len, persistent);
+}
+
+zend_string *get_commit_sha(const char *git_dir) {
+    char head_path[PATH_MAX];
+#ifndef _WIN32
+    snprintf(head_path, sizeof(head_path), "%s/HEAD", git_dir);
+#else
+    snprintf(head_path, sizeof(head_path), "%s\\HEAD", git_dir);
+#endif
```
bwoebi, line 73: You don't have to distinguish Windows for constructed paths, Windows perfectly understands forward-slashes in paths. You just have to handle the forward slashes _and_ backslashes when you receive a path as input from somewhere.
```suggestion
    snprintf(head_path, sizeof(head_path), "%s/HEAD", git_dir);
```

ext/git.c
```diff
@@ -0,0 +1,339 @@
...
+        source = inject_from_git_dir(carrier, cwd);
+    }
+
+    if (source != GIT_SOURCE_NONE) {
+        ddtrace_git_metadata *git_metadata_obj = (ddtrace_git_metadata *) Z_OBJ_P(carrier);
+        zend_string *property_commit = Z_STR(git_metadata_obj->property_commit);
+        zend_string *property_repository = Z_STR(git_metadata_obj->property_repository);
+        zend_string_addref(property_commit);
+        zend_string_addref(property_repository);
+        cache_git_metadata(cwd, property_commit, property_repository, source);
```
bwoebi, line 233: You may only cache if you're extracting from directory. Env and tags may change anytime, and you'd not be able to invalidate it.

ext/git.c
bwoebi: May I propose `DDTRACE_G(git_object)` being a `zend_object *` (which you release at request end if non-NULL)? If it's NULL, then we proceed as usual. If it isn't, you do a ZVAL_OBJ_COPY to carrier? That we we actually persist the git metadata for a whole request.

PROFeNoM: And so you'd remove both `git_metadata` and `git_request_cache_updated` from ddtrace_globals? right?

ext/serializer.c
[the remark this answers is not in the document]
bwoebi: Okay thanks, I'll check this out later.

ext/git.c
bwoebi: We still need `git_metadata` for _cross-request_ caching. Storing the `zend_object *` is only allowed for within a request.

ext/git.c
```diff
@@ -0,0 +1,308 @@
...
+        char *last_slash = find_last_dir_separator(current_dir);
+        if (!last_slash) {
+            return NULL;
+        }
+
+        *last_slash = '\0';
+    }
+}
+
+void remove_credentials(zend_string *repo_url) {
```
bwoebi, line 143: This must create a copy if `at != NULL`, to prevent possible race conditions and invalidating the global data (e.g. `get_DD_GIT_REPOSITORY_URL` may point to a persistent global)

PROFeNoM:
Converting to draft until I add the new tests.

ext/git.c
```diff
@@ -0,0 +1,388 @@
...
+void ddtrace_clean_git_metadata(void) {
+    git_metadata_t *val;
+    ZEND_HASH_FOREACH_PTR(&DDTRACE_G(git_metadata), val) {
+        if (val->property_commit) zend_string_release(val->property_commit);
+        if (val->property_repository) zend_string_release(val->property_repository);
+        pefree(val, 1);
+    }
+    ZEND_HASH_FOREACH_END();
+    zend_hash_destroy(&DDTRACE_G(git_metadata));
+}
```
iamluc, line 369: Is there a reason to not passing a destructor function when initializing the `DDTRACE_G(git_metadata)` with `zend_hash_init` instead of this function?

ext/git.c
```diff
@@ -0,0 +1,388 @@
...
+
+    if (DDTRACE_G(git_object)) {
+        ddtrace_git_metadata *git_metadata = (ddtrace_git_metadata *) DDTRACE_G(git_object);
+        if (Z_TYPE(git_metadata->property_repository) == IS_STRING) {
+            zend_string_release(Z_STR(git_metadata->property_repository));
+        }
+#if PHP_VERSION_ID < 70300 || PHP_VERSION_ID >= 70400
+        zend_object_release(DDTRACE_G(git_object));
+#endif
+        efree(DDTRACE_G(git_object));
```
bwoebi, line 385:
```suggestion
        zend_object_release(DDTRACE_G(git_object));
```

This is how it _should_ be. Is there something failing otherwise? This version check comes weird to me.

PROFeNoM: Ok. The issue was that `ddtrace_clean_git_object` shouldn't be called during `zend_post_deactivate_modules`, but `zend_deactivate_modules`. The refcount of git_object was otherwise incremented from 1 to two (for w/e reason), hence resulting in a memory leak. 

FWIW, the version-specific check was used in 7.3 as there was a failure when calling `zend_objects_store_del` since `EG(objects_store).object_buckets[handle]` wasn't a valid object. Regardless, changing when the object is cleaned seems to address the initial issue.

ext/git.c
PROFeNoM: There was no particular reason besides that I wasn't aware of it. Added one :+1:

bwoebi (approved):
Seems to work now :-)

PR #2735, opened 2024-06-27
WAF telemetry

cataphract:
### Description

See individual commits for descriptions of the changes.

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/extension/commands_helpers.c
```diff
@@ -648,6 +678,104 @@ bool dd_command_process_metrics(mpack_node_t root, zend_object *nonnull span)
...
+    zend_string_release(zstr_new);
+}
+
+// NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
+void _handle_telemetry_metric(const char *nonnull key_str, size_t key_len,
+    double value, const char *nonnull tags_str, size_t tags_len)
+{
+#define HANDLE_METRIC(name, type)                                              \
+    do {                                                                       \
+        if (key_len == sizeof(name "") - 1 &&                                  \
```
Anilm3, line 752:
```suggestion
        if (key_len == LSTRLEN(name) &&                                  \
```

appsec/src/helper/metrics.cpp
```diff
@@ -0,0 +1,11 @@
+// Unless explicitly stated otherwise all files in this repository are
+// dual-licensed under the Apache-2.0 License or BSD-3-Clause License.
+//
+// This product includes software developed at Datadog
+// (https://www.datadoghq.com/). Copyright 2022 Datadog, Inc.
+
+#include "metrics.hpp"
+
+namespace dds::metrics {
+TelemetrySubmitter::~TelemetrySubmitter() = default;
```
Anilm3, line 10: This could be moved to the header as `inline`.

appsec/src/helper/remote_config/client_handler.cpp
```diff
@@ -57,13 +65,36 @@ std::unique_ptr<client_handler> client_handler::from_settings(
...
+        const bool result = rc_client_->poll();
+
+        auto now = std::chrono::steady_clock::now();
+        last_success_ = now;
+
+        if (result && creation_time_ != empty_time) {
+            auto elapsed =
+                std::chrono::duration_cast<std::chrono::milliseconds>(
+                    now - creation_time_);
+            msubmitter_->submit_metric(metrics::rc_first_pull,
```
Anilm3, line 94: Interesting, you added the only metrics no one else has implemented (`appsec.remote_config`) and which I was deprecating... I might suggest removing them for now as, if they're not removed, the namespace will change from `appsec` to something else.

appsec/src/helper/service.hpp
```diff
@@ -21,13 +24,109 @@ namespace dds {
...
+            meta_[std::string{name}] = std::move(value);
+        }
+
+        // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
+        void submit_legacy_meta_copy_key(
+            std::string name, std::string value) override
+        {
+            const std::lock_guard<std::mutex> lock{meta_mutex_};
+            meta_[std::move(name)] = std::move(value);
+        }
```
Anilm3, line 75: I don't think you need two different names for these methods as if copy-semantics are used both methods are equivalent and move-semantics can only result in one of the two being called. On the other hand, shouldn't both log?

appsec/src/helper/service.hpp
```diff
@@ -21,13 +24,109 @@ namespace dds {
...
+
+        template <typename Func> void drain_metrics(Func &&func)
+        {
+            std::vector<tel_metric> metrics;
+            {
+                const std::lock_guard<std::mutex> lock(pending_metrics_mutex_);
+                metrics.swap(pending_metrics_);
+            }
+            for (auto &metric : metrics) {
```
Anilm3, line 84:
```suggestion
            for (auto &[name, value, tags] : metrics) {
```

appsec/src/helper/subscriber/waf.cpp
```diff
@@ -124,9 +124,111 @@ void log_cb(DDWAF_LOG_LEVEL level, const char *function, const char *file,
...
+    } else {
+        tags += ":false"sv;
+    }
+    if (rules_version) {
+        tags += ",event_rules_version:"sv;
+        tags += rules_version.value();
+    }
+    tags += ",waf_version:"sv;
+    tags += ddwaf_get_version();
+    return tags;
```
Anilm3, line 142: Yeah, this is what I meant before with leaking the telemetry API specifics, I think you could solve this by creating a small abstraction on top of the tags which then takes care of creating the final string.

Anilm3 (approved):
I left a number of comments, everything looks correct albeit there could be some improvements around the helper telemetry API and the naming conventions.

appsec/src/extension/commands_helpers.c
```diff
@@ -648,6 +678,104 @@ bool dd_command_process_metrics(mpack_node_t root, zend_object *nonnull span)
...
+            memcmp(key_str, name, key_len) == 0) {                             \
+            static zend_string *_Atomic key_zstr;                              \
```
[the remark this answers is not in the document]
cataphract: The variable is not thread-local, it's a true global.

appsec/src/extension/ddtrace.h
```diff
@@ -81,3 +81,28 @@ bool dd_trace_user_req_add_listeners(
 zend_string *nullable dd_ip_extraction_find(zval *nonnull server);
 
 const char *nullable dd_trace_remote_config_get_path(void);
+
+typedef enum {
+    DDTRACE_METRIC_TYPE_GAUGE,
+    DDTRACE_METRIC_TYPE_COUNT,
+    DDTRACE_METRIC_TYPE_DISTRIBUTION,
```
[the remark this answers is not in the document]
cataphract: I don't know. Doesn't seem exposed through sidecar, but I'm not an authority on the rust codecase. In any case, it's not metrics.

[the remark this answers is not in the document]
bwoebi: It does, but the components-rs does not yet expose a helper function for that. I.e. you'd have to add logic to send TelemetryActions::AddLog() to components-rs/telemetry.rs.
IDM seems to be adding support for that soon, though.

appsec/src/helper/client.cpp
```diff
@@ -501,4 +498,78 @@ void client::run(worker::queue_consumer &q)
...
+    RequestMetricsSubmitter() = default;
+    ~RequestMetricsSubmitter() override = default;
+    RequestMetricsSubmitter(const RequestMetricsSubmitter &) = delete;
+    RequestMetricsSubmitter &operator=(
+        const RequestMetricsSubmitter &) = delete;
+    RequestMetricsSubmitter(RequestMetricsSubmitter &&) = delete;
+    RequestMetricsSubmitter &operator=(RequestMetricsSubmitter &&) = delete;
+
+    void submit_metric(
+        std::string_view name, double value, std::string tags) override
```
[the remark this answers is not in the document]
cataphract: I don't see the point of adding overhead with such an intermediate representation. I guess I could add something with no overhead like

```cpp
class Tags {
void add(std::string_view key, std::string_view value);
std::string consume() { return std::move(tags_str_); }
private:
std::string tags_str_;
}
```

but even that seems unnecessary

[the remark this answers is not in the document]
Anilm3: So in your view is better to have to create this string-of-tags in an undefined and barely documented format than add some small abstraction with minimal overhead to simplify adding tags?

appsec/src/extension/commands_helpers.c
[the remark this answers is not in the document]
Anilm3: I can read, but I was just asking if you were just trying to have a copy per thread by using a single global.

appsec/src/helper/client.cpp
```diff
@@ -501,4 +498,78 @@ void client::run(worker::queue_consumer &q)
...
+    {
+        SPDLOG_TRACE("submit_legacy_meta_copy_key [req]: name={}, value={}",
+            name, value);
+        meta[name] = value;
+    }
+
+    std::map<std::string, std::string> meta;
+    std::map<std::string_view, double> metrics;
+    std::map<std::string_view, std::vector<std::pair<double, std::string>>>
+        tel_metrics;
```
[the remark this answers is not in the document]
cataphract: the version we have supports it. I can change it to unordered

appsec/src/helper/subscriber/waf.cpp
cataphract: see above, but I don't really find it necessary to abstract away such a simple convention

appsec/src/helper/service.hpp
cataphract: That's technically true, but I'm making a distinction where std::string_view is used for static storage constants, so this helps reinforce that, even if the implementations don't then use separate maps for owned and non-owned keys (they could though).

appsec/src/helper/service.hpp
Anilm3: Structured bindings are more readable though and are less sensitive to change, compared to... for example a hypothetical string of tags with an undefined format, as the compiler will easily let you know.

appsec/src/helper/client.cpp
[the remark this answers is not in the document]
cataphract: it's a comma separated key:value list, not exactly the .doc format. To answer your rhetorical question, I think the weighted value of the added abstraction in terms of added total complexity vs clarity and unification/hiding of the "format" implementation is a wash. But I'll happily oblige

appsec/src/helper/service.hpp
cataphract: besides what i said already, it may boil down to personal preference, but for me

```
for (auto &metric : metrics) {
  std::invoke(std::forward<Func>(func), metric.name, metric.value,
    std::move(metric.tags));
 }
 ```

reads better than

```
for (auto &[name, value, tags] : metrics) {
  std::invoke(std::forward<Func>(func), name, value, std::move(tags));
 }
 ```

because it's immediately obvious that name, value and tags refer to the metric.

Anilm3: That's okay then, it's not a big deal.

appsec/src/extension/ddtrace.h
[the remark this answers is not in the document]
Anilm3: We can add support later on then.

appsec/src/extension/commands_helpers.c
```diff
@@ -648,6 +678,103 @@ bool dd_command_process_metrics(mpack_node_t root, zend_object *nonnull span)
     return true;
 }
 
+static void _handle_telemetry_metric(const char *nonnull key_str,
```
estringana, line 681: Not sure if it is a good practice or I made it up but, should this early declaration be at the top?

bwoebi: You put these at the top if you use these multiple times at various places. For a single use like here, there's no reason to.

appsec/src/extension/ddtrace.c
```diff
@@ -123,6 +147,7 @@ void dd_trace_startup()
 
     if (get_global_DD_APPSEC_TESTING()) {
         _register_testing_objects();
+        _setup_testing_telemetry_functions();
```
estringana, line 150: I coudln't get why this is within `APPSEC_TESTING`. I'm probably missing something

estringana: Ok. Makes sense

appsec/src/helper/network/proto.hpp
```diff
@@ -89,6 +89,9 @@ template <typename T> struct base_response_generic : public base_response {
     }
 };
 
+using telemetry_metrics =
```
estringana, line 92: I think this is not used

appsec/src/extension/commands_helpers.c
```diff
@@ -648,6 +678,103 @@ bool dd_command_process_metrics(mpack_node_t root, zend_object *nonnull span)
...
+            ddtrace_metric_add_point(key_zstr, value, tags_zstr);              \
+            zend_string_release(tags_zstr);                                    \
+            mlog_g(dd_log_debug,                                               \
+                "Telemetry metric %.*s added with tags %.*s and value %f",     \
+                (int)key_len, key_str, (int)tags_len, tags_str, value);        \
+            return;                                                            \
+        }                                                                      \
+    } while (0)
+
+    HANDLE_METRIC("waf.requests", DDTRACE_METRIC_TYPE_COUNT);
```
estringana, line 767: I saw you added some integration tests checking this metrics but I am missing some PHPT checking the extension.

estringana: It would also allow you to test unknown telemetry metrics

PR #2742, opened 2024-07-03
Respect the active service/env/version for metric tags

bwoebi:
I just noticed that while testing live-debugger metrics - the actual service still ought to be reported.
It's just php-dogstatsd which isn't aware of ddtrace state.

### DIFF
--- a/ext/sidecar.c
+++ b/ext/sidecar.c
@@ -8,6 +8,7 @@
 #include <components-rs/ddtrace.h>
 #include "sidecar.h"
 #include "telemetry.h"
+#include "serializer.h"
 
 ddog_SidecarTransport *ddtrace_sidecar;
 ddog_Endpoint *ddtrace_endpoint;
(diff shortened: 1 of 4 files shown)


ext/sidecar.c
```diff
@@ -136,15 +137,23 @@ static inline void ddtrace_sidecar_dogstatsd_push_tag(ddog_Vec_Tag *vec, ddog_Ch
 
 static void ddtrace_sidecar_dogstatsd_push_tags(ddog_Vec_Tag *vec, zval *tags) {
     // Global tags (https://github.com/DataDog/php-datadogstatsd/blob/0efdd1c38f6d3dd407efbb899ad1fd2e5cd18085/src/DogStatsd.php#L113-L125)
+    ddtrace_span_data *span = ddtrace_active_span();
     zend_string *env = get_DD_ENV();
+    if (span) {
+        env = ddtrace_convert_to_str(&span->property_version);
```
iamluc, line 143: `property_env` .
Also `get_DD_ENV();` could be called in the "else" branch to avoid an unnecessary call when span is not null. :shrug: 
Same for `get_DD_VERSION()` below

iamluc (approved):
Expect the issue with `property_env`, :+1:

Issue #2481, opened 2024-01-19
[Feature] Symfony Messenger Integration


BeyerJC:
### Describe the feature you'd like

We already have an integration for symfony but it lacks the tracing for the Symfony messenger component.

I imagine it to be like the LaravelQueueIntegration and see when a worker picked up a message and all traces that happend while processing (e.g. guzzle integration may add some calls).

### Is your feature request related to a problem?

No

### Describe alternatives you've considered

Keep using my own solution.

PROFeNoM:
Hello @BeyerJC 👋,

Thank you for submitting your feature request. While we have yet to estimate a time for implementing this library's instrumentation, we agree that it would be a valuable addition.

I noticed that Symfony auto-instrumentation is already available through [OTel](https://github.com/open-telemetry/opentelemetry-php-contrib), which records spans for Symfony Messenger (open-telemetry/opentelemetry-php-contrib#215). In the meantime, you can use the OTel instrumentation with dd-trace-php support to obtain spans from this library.

Additionally, we encourage external contributions and would be happy to assist anyone with implementing this feature 😊

natepage:
Hi @PROFeNoM 👋 ,

We recently started to use datadog to monitor our PHP applications, we are mainly using Symfony as a framework to build those. 

I came across this issue after wondering where were my queue messages traces 😄 
I'm completely new to dd-trace-php but gave it a go and tried to instrument Symfony Messenger, having a look at what was done for Laravel Queue.

It's far from done but I have managed to:

- instrument the message bus for both dispatching a message to the remote queue, and consuming it within the worker
- hook into the sending logic to add a custom claim onto the envelope to assure distributed tracing

I've tried to instrument the Messenger MiddlewareInterface::handle method in order to have a clear picture of all middleware executed as part of the stack within the message bus, but it's not working as expected, and I can see a span only for the very first middleware and that's it. 

In order to experiment within our applications, I created a small composer package, and you can see my attempt in that instrumentation file directly: [symfony-messenger-integration.php](https://github.com/natepage/dd-trace-symfony-messenger/blob/main/files/symfony-messenger-integration.php)

This is very interesting and it would be of great value to us to have this integration working, and I'm happy to keep working on it, but... To be completely honest, I have no idea what I'm doing! 😄 I've managed to put this POC together by copying what other integrations are doing but some help/guidance from experts would be much appreciated! Thank you!

PROFeNoM:
Hey @natepage :wave:

[...]

PR #2749, opened 2024-07-08
feat: Symfony Messenger Integration

PROFeNoM:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
+++ b/src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
@@ -1,0 +1,296 @@
+<?php
+
+namespace DDTrace\Integrations\SymfonyMessenger;
+
+use DDTrace\HookData;
+use DDTrace\Integrations\Integration;
+use DDTrace\SpanData;
+use DDTrace\Tag;
+use DDTrace\Util\DDTraceStamp;
+use DDTrace\Util\ObjectKVStore;
+use Symfony\Component\Messenger\Bridge\AmazonSqs\Transport\AmazonSqsReceivedStamp;
+use Symfony\Component\Messenger\Envelope;
+use Symfony\Component\Messenger\Stamp\BusNameStamp;
+use Symfony\Component\Messenger\Stamp\DelayStamp;
+use Symfony\Component\Messenger\Stamp\ReceivedStamp;
+use Symfony\Component\Messenger\Stamp\RedeliveryStamp;
+use Symfony\Component\Messenger\Stamp\TransportMessageIdStamp;
+
+use function DDTrace\install_hook;
+use function DDTrace\trace_method;
+
+class SymfonyMessengerIntegration extends Integration
+{
+    const NAME = 'symfonymessenger';
+
+    public function init(): int
+    {
+        $integration = $this;
+
+        trace_method(
+            'Symfony\Component\Messenger\MessageBusInterface',
+            'dispatch',
+            function (SpanData $span, array $args, $returnValue, $exception = null) use ($integration) {
+                $integration->setSpanAttributes($span, null, $args[0], $exception);
+            }
+        );
+
+        // Attach current context to Envelope before sender sends it to remote queue
+        install_hook(
+            'Symfony\Component\Messenger\Transport\Sender\SenderInterface::send',
+            function (HookData $hook) {
+                /** @var \Symfony\Component\Messenger\Envelope $envelope */
+                $envelope = $hook->args[0];
+
+                if (\ddtrace_config_distributed_tracing_enabled()) {
+                    $ddTraceStamp = $envelope->last(DDTraceStamp::class);
+
+                    // Add distributed tracing stamp only if not already on the envelope
+                    if ($ddTraceStamp === null) {
+                        $hook->overrideArguments([
+                            $envelope->with(new DDTraceStamp(\DDTrace\generate_distributed_tracing_headers()))
+                        ]);
+                    }
+                }
+            }
+        );
+
+        install_hook(
+            'Symfony\Component\Messenger\Transport\Sender\SenderInterface::send',
+            function (HookData $hook) {
+                /** @var \Symfony\Component\Messenger\Envelope $envelope */
+                $envelope = $hook->args[0];
+
+                if (\ddtrace_config_distributed_tracing_enabled()) {
+                    $ddTraceStamp = $envelope->last(DDTraceStamp::class);
+
+                    // Add distributed tracing stamp only if not already on the envelope
+                    if ($ddTraceStamp === null) {
+                        $hook->overrideArguments([
+                            $envelope->with(new DDTraceStamp(\DDTrace\generate_distributed_tracing_headers()))
+                        ]);
+                    }
+                }
+            }
+        );
+
+        trace_method(
+            'Symfony\Component\Messenger\Worker',
+            'handleMessage',
+            [
+                'prehook' => function (SpanData $span, array $args) use ($integration, &$newTrace) {
+                    /** @var \Symfony\Component\Messenger\Envelope $envelope */
+                    $envelope = $args[0];
+                    /** @var string $transportName */
+                    $transportName = $args[1];
+
+                    $integration->setSpanAttributes(
+                        $span,
+                        $transportName,
+                        $envelope,
+                        null,
+                        true
+                    );
+
+                    $ddTraceStamp = $envelope->last(DDTraceStamp::class);
+                    if ($ddTraceStamp instanceof DDTraceStamp) {
+                        if (\dd_trace_env_config('DD_TRACE_SYMFONY_MESSENGER_DISTRIBUTED_TRACING')) {
+                            $newTrace = \DDTrace\start_trace_span();
+                            $integration->setSpanAttributes(
+                                $newTrace,
+                                $transportName,
+                                $envelope,
+                                null,
+                                true
+                            );
+
+                            \DDTrace\consume_distributed_tracing_headers($ddTraceStamp->getHeaders());
+
+                            $span->links[] = $newTrace->getLink();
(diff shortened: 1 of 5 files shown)


natepage:
@PROFeNoM I was about to follow up on #2481 and then seen this!

Thank you very much for taking the time to do all this work! Super exciting 😄 

Please feel free to reach out if you need anything, testing, etc.

src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
```diff
@@ -32,6 +32,13 @@ public function requiresExplicitTraceAnalyticsEnabling(): bool
         return false;
     }
 
+    public function isSymfonyMessengerCommand()
+    {
+        $consoleCommand = $_SERVER['argv'][1] ?? '';
+
+        return $consoleCommand === 'messenger:consume';
+    }
```
Jean85, line 40: This may not work: Symfony has an "autocomplete" feature for commands, so `mess:cons` would work too, if it's not ambiguous with other registered commands.

src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
```diff
@@ -0,0 +1,348 @@
...
+                $envelope = $hook->args[0];
+
+                if (\ddtrace_config_distributed_tracing_enabled()) {
+                    $ddTraceStamp = $envelope->last(DDTraceStamp::class);
+
+                    // Add distributed tracing stamp only if not already on the envelope
+                    if ($ddTraceStamp === null) {
+                        $tracingHeaders = \DDTrace\generate_distributed_tracing_headers();
+                        $hook->overrideArguments([
+                            $envelope->with(new DDTraceStamp($tracingHeaders))
```
Jean85, line 67: This may be wrong. `Envelope` [is immutable](https://github.com/symfony/symfony/blob/9a0266029867201e5637af634eaa8a263433dc07/src/Symfony/Component/Messenger/Envelope.php#L57-L66), so this has to be done like this:
```suggestion
                            $envelope = $envelope->with(new DDTraceStamp($tracingHeaders))
```

The issue here is that you're not changing `$envelope` by reference, so we may be out of luck :(

src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
```diff
@@ -0,0 +1,348 @@
...
+                    }
+
+                    $integration->setSpanAttributes(
+                        $span,
+                        'symfony.messenger.consume',
+                        null,
+                        $envelope,
+                        $transportName,
+                        'receive'
+                    );
```
Jean85, line 102: If we set `DD_TRACE_REMOVE_ROOT_SPAN_SYMFONY_MESSENGER=true`, would this be the root span, and hence appear as a resource? That would be my desired outcome, so that message handlers appear alike to controllers for HTTP traffic.

Jean85 (commented):
Thank you so much for this! I'm a new DD customer and this may save me a huge chunk of manual instrumentation work.

Is there a timeframe for this? I'm open to test this on my code if needed.

src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
PROFeNoM: Yes, that's indeed the goal, and also one of the reasons this env var will be enabled by default :+1: If the root span isn't removed, it will most likely be the `messenger:consume` command span; however, this operation will certainly be a long-running job, consuming multiple messages.

src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
PROFeNoM: In this piece of code, we're _overriding_ the argument of `SenderInterface::send` (which only takes one argument: an `Envelope` object) by another `Envelope` object with a custom stamp containing the distributed trace context information. This is done using a custom method from our API (`\DDTrace\HookData::overrideArguments`). As a result of this function call, the newly provided envelope will be used by the method call instead of the original one.

Doing `$envelope = $envelope->with(new DDTraceStamp($tracingHeaders))` would effectively replace the envelope within the scope of the prehook, but not the actual method call :+1:

src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
PROFeNoM: Thanks for the info; I wasn't aware of that. That's a bit annoying :p

PROFeNoM:
> Is there a timeframe for this? I'm open to test this on my code if needed.

The PR is overall ready for review. However, it'd be super valuable if you guys (@Jean85 & @natepage) have some time to try it out! We're really eager to any feedback 😃 After that, it could hopefully be included in the next release :-)

To try it out, you can use the link to the [datadog-setup.php artifact](https://output.circle-artifacts.com/output/job/d917eba0-7488-4155-a648-29e57209c0d5/artifacts/0/datadog-setup.php) (generated from this [CircleCI Job](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/16800/workflows/066960b8-df09-4bfc-ae20-7f0efbeabbd3/jobs/4807388/artifacts)) and use it instead of the one you're currently using when [installing the extension](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/php/#install-the-extension).

Thanks 😃

natepage:
@PROFeNoM Thank you very much for that! We will test this out early next week 😉

Jean85:
Thanks @PROFeNoM! I'm not fully available these weeks due to the summer holidays, but I'll try to test it out ASAP!

natepage:
@PROFeNoM just an update, I'm sorry but last week didn't go as planned, still on my list tho, will come back asap

Jean85:
@PROFeNoM I just had time to test this out, and it seems to work very well! 

My setup is the following:
 * I have a container with a web app, where I produced messages both from HTTP calls and (manually) via a CLI command for testing purposes
 * the messages reach a first transport (`aws_writer`) where I upload stuff to S3, then I dispatch a second message to a different transport (`notification`)
 * everything is consumed by a single (autoscaled) consumer container

The operations are populated correctly, marking producer and consumers correctly, even if the roles are sometimes overlapped:
image

Producer traces are shown correctly:
image

Traces are correlated correctly through both dispatching (app on the left, first message in the middle, second on the right): 
image

Do you need me to test/verify something else? Can't wait to deploy this! :smile:

src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
```diff
@@ -0,0 +1,348 @@
...
+        );
+
+        // Attach current context to Envelope before sender sends it to remote queue
+        install_hook(
+            'Symfony\Component\Messenger\Transport\Sender\SenderInterface::send',
+            function (HookData $hook) {
+                /** @var \Symfony\Component\Messenger\Envelope $envelope */
+                $envelope = $hook->args[0];
+
+                if (\ddtrace_config_distributed_tracing_enabled()) {
```
pablomartinezbernardo, line 60: Should this be governed by `DD_TRACE_SYMFONY_MESSENGER_DISTRIBUTED_TRACING`? Which triggers another question: Why add a new setting for this instead of using the global distributed tracing setting? Do we expect customers to have it enabled for one and not the other? Is it just for better customization?

pablomartinezbernardo (approved):
Just one small comment, LGTM otherwise

src/DDTrace/Integrations/SymfonyMessenger/SymfonyMessengerIntegration.php
PROFeNoM: `DD_TRACE_SYMFONY_MESSENGER_DISTRIBUTED_TRACING` will state whether the produce <-> consume relationship happens in the traditional distributed tracing manner - i.e., within the same trace - or if it should leverage span links. This decision is made on the consuming end of the operation.

In retrospect, `ddtrace_config_distributed_tracing_enabled` will state whether the produce <-> consume relationship will exist or not. This decision is made on the producing end of the operation.

We expect customers to have the global distributed tracing setting enabled. Regarding `DD_TRACE_SYMFONY_MESSENGER_DISTRIBUTED_TRACING`, we also expect it to be enabled by default - i.e., by default, we don't change the UX customers are used to. However, we offer the possibility to use span links instead, which has proved to be particularly effective to un-bloat flamegraphs. In the future, this could even be the default behavior... but that's TBD.

This comment just reminded me that I've to update the public doc though.

PROFeNoM:
I'll rebase the branch on master first before merging to make sure everything's good

radykal-com:
Hi, I have a problem I think it's originated by this feature.

If you send a message through Symfony messenger from an application with dd-trace-php extension loaded and consume the message from an application without dd-trace-php extension loaded, the application just crashes with

```Could not decode stamp: Could not denormalize object of type "DDTrace\Integrations\SymfonyMessenger\DDTraceStamp"```

dayallnash:
I also have the same issue as @radykal-com

PROFeNoM:
Thanks for reporting the issue @radykal-com, we're looking into it right now.

@dayallnash, do you strictly have the same error message as reported above?

radykal-com:
Apparently the merged fix should work as it gets rid of the DDTraceStamp class and uses a generic stamp shipped with symfony.

dayallnash:
> Thanks for reporting the issue @radykal-com, we're looking into it right now.
> 
> @dayallnash, do you strictly have the same error message as reported above?

Sorry for the late reply. Yes it was exactly the same issue. Having reviewed the merged fix, I am confident that that should work to fix that specific error - but I'm not currently in a position to test it to know if the entire feature works.

PR #2750, opened 2024-07-09
[loader] Add a libc check at runtime

iamluc:
### Description

Avoid trying to load an incompatible lib if we know it will fail.

Check code taken from https://github.com/DataDog/auto_inject/pull/266/files#diff-612d425717135d74863f5b8ef8a0c02d214976ad4a5d8b1690ee4bf2297459f0R47-R60

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Looks fine to me - after all it's just a sanity check with the new loader mechanism and shouldn't happen anyway.

PR #2754, opened 2024-07-09
[Tracer] Adds SpanEvents Support for DD and OTEL

link04:
### Description
Adding support for SpanEvents, related [RFC](https://docs.google.com/document/d/1b8GLnbC4oPrHoksmUOhy66iEJto1FjnVivr_jsZlSNw/edit), currently, added a SpanEvent class and ExceptionSpanEvent which inherits from it to add specialized exception events to the Span. 

Related PRs:
- System Test:
https://github.com/DataDog/system-tests/pull/2841

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/ddtrace.stub.php
```diff
@@ -23,6 +23,23 @@
      */
     const DBM_PROPAGATION_FULL = UNKNOWN;
 
+    class SpanEvent implements \JsonSerializable {
```
bwoebi, line 26: We don't need Json serialization I think - that was as far as I'm aware an implementation artifact of span links.

bwoebi, line 26:
```suggestion
    class SpanEvent implements \JsonSerializable {
        public function __construct(string $name, array $attributes = []) {}
```

ext/ddtrace.stub.php
```diff
@@ -23,6 +23,23 @@
...
+        /**
+         * @var string The event name
+         */
+        public string $name;
+
+        /**
+         * @var int The event start time in nanoseconds, if not provided set the current Unix timestamp
+         */
+        public int $timeUnixNano;
```
bwoebi, line 35: I dislike the name, but don't have a good suggestion right now.

link04: Sameee!!! I keep having to look at the changes to remember it, I've only been following the RFC so used that name to keep it consistent, I am open to any recommendations

src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -136,7 +160,8 @@ public static function startSpan(
         array $attributes,
         array $links,
         int $totalRecordedLinks,
-        bool $isRemapped = true // Answers the question "Was the span created using the OTel API?"
+        bool $isRemapped = true, // Answers the question "Was the span created using the OTel API?"
+        array $events
```
bwoebi, line 164: Doesn't really make sense to put the optional arg before the required one:
```suggestion
        array $events,
        bool $isRemapped = true // Answers the question "Was the span created using the OTel API?"
```

src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -69,7 +78,8 @@ private function __construct(
         ResourceInfo $resource,
         array $links = [],
         int $totalRecordedLinks = 0,
-        bool $isRemapped = true
+        bool $isRemapped = true,
+        array $events = [],
```
bwoebi, line 82:
```suggestion
        array $events = [],
        bool $isRemapped = true
```
The trailing comma is not allowed on old PHP versions, also, ordering.

ext/ddtrace.stub.php
```diff
@@ -24,6 +24,8 @@
     const DBM_PROPAGATION_FULL = UNKNOWN;
 
     class SpanEvent implements \JsonSerializable {
+        public function __construct(string $name, int $timestamp = null, array $attributes = []) {}
```
bwoebi, line 27: Let's put timestamp as third arg.

zacharycmontoya: Since the OpenTelemetry SDK has the parameter orders of $name, $timestamp, $attributes I wonder if it would make sense to keep this code as-is to match that order: https://github.com/open-telemetry/opentelemetry-php/blob/17ad56a3f3a64481b6764bb46b5584d7bbc5b138/src/SDK/Trace/Event.php#L12

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -92,6 +95,21 @@ public function addLink(SpanContextInterface $context, iterable $attributes = []
...
     }
 
+    public function addEvent(string $name, int $timestamp = null, iterable $attributes = []): SpanBuilderInterface
+    {
+        $this->events[] = new Event(
+            $name, 
+            $timestamp == null ? microtime(true) * 1e9 : $timestamp,
+            $this->tracerSharedState
+                ->getSpanLimits()
+                ->getLinkAttributesFactory()
```
zacharycmontoya, line 105: Based on [SpanLimits.php](https://github.com/open-telemetry/opentelemetry-php/blob/17ad56a3f3a64481b6764bb46b5584d7bbc5b138/src/SDK/Trace/SpanLimits.php#L28), I think it should be this instead:

```suggestion
                ->getEventAttributesFactory()
```

tests/OpenTelemetry/Integration/InteroperabilityTest.php
```diff
@@ -1149,4 +1150,60 @@ public function testSpanLinksInteroperabilityAddDuplicates()
...
+            $OTelSpanEvent = $OTelSpan->toSpanData()->getEvents()[0];
+
+            // Check the first event
+            $this->assertSame('event-name', $OTelSpanEvent->getName());
+
+            close_span();
+        });
+
+        $this->assertCount(1, $traces[0]);
+        $this->assertSame("[{\"name\":\"event-name\",\"time_unix_nano\":1720037568765201300,\"attributes\":{\"arg1\":\"value1\"}}]", $traces[0][0]['meta']['events']);
```
zacharycmontoya, line 1179: One note here: If an event has an _array_ as one of the attributes, we aren't exploding it out into indices. For example, if there is an attribute `'stringArr' => ['value1', 'value2']` then we should expect the JSON to look like `\"stringArr\":[\"value1\",\"value2\"]`.

Can you make sure this test case gets added to the unit tests?

ext/ddtrace.stub.php
bwoebi: But it has no defaults though. Their actual user-facing API has the expected order:
https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Trace/Span.php#L172

ext/ddtrace.c
```diff
@@ -651,6 +651,35 @@ PHP_METHOD(DDTrace_SpanLink, jsonSerialize) {
...
+
+    Z_TRY_ADDREF(event->property_name);
+    zend_hash_add(array, name, &event->property_name);
+    Z_TRY_ADDREF(event->property_timestamp);
+    zend_hash_add(array, timestamp, &event->property_timestamp);
+
+    zend_string_release(name);
+    zend_string_release(timestamp);
+
+    if (Z_TYPE(event->property_attributes) != IS_NULL && Z_ARRVAL(event->property_attributes)->nNumOfElements > 0) {
```
bwoebi, line 673:
```suggestion
    if (Z_TYPE(event->property_attributes) == IS_ARRAY && Z_ARRVAL(event->property_attributes)->nNumOfElements > 0) {
```

ext/ddtrace.c
```diff
@@ -651,6 +651,35 @@ PHP_METHOD(DDTrace_SpanLink, jsonSerialize) {
...
+PHP_METHOD(DDTrace_SpanEvent, jsonSerialize) {
+    ddtrace_span_event *event = (ddtrace_span_event*)Z_OBJ_P(ZEND_THIS);
+
+    zend_array *array = zend_new_array(3);
+
+    zend_string *name = zend_string_init("name", sizeof("name") - 1, 0);
+    zend_string *timestamp = zend_string_init("time_unix_nano", sizeof("time_unix_nano") - 1, 0);
+
+    Z_TRY_ADDREF(event->property_name);
+    zend_hash_add(array, name, &event->property_name);
```
bwoebi, line 666: You can use `zend_hash_str_add(array, ZEND_STRL("name"), &event->property_name)` instead of allocating zend_string manually.

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -92,6 +96,38 @@ public function addLink(SpanContextInterface $context, iterable $attributes = []
...
+        return $this;
+    }
+
+    public function recordException(Throwable $exception, iterable $attributes = []): SpanBuilderInterface
+    {
+        // Standardized exception attributes
+        $exceptionAttributes = [
+            'exception.message' => $attributes['exception.message'] ?? $exception->getMessage(),
+            'exception.type' => $attributes['exception.type'] ?? get_class($exception),
+            'exception.stacktrace' => $attributes['exception.stacktrace'] ?? $exception->getTraceAsString(),
```
bwoebi, line 113:
```suggestion
            'exception.stacktrace' => $attributes['exception.stacktrace'] ?? \DDTrace\get_sanitized_exception_trace($exception),
```

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -92,6 +96,38 @@ public function addLink(SpanContextInterface $context, iterable $attributes = []
...
+            'exception.escaped' => false
+        ];
+
+        // Update span metadata based on exception attributes
+        $this->setAttribute(Tag::ERROR_MSG, $exceptionAttributes['exception.message']);
+        $this->setAttribute(Tag::ERROR_TYPE, $exceptionAttributes['exception.type']);
+        $this->setAttribute(Tag::ERROR_STACK, $exceptionAttributes['exception.stacktrace']);
```
bwoebi, line 120: Are you sure about this? Exceptions generally should only be added on spans if they escape their span.

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -92,6 +96,38 @@ public function addLink(SpanContextInterface $context, iterable $attributes = []
...
+
+        // Merge additional attributes
+        $allAttributes = array_merge($exceptionAttributes, iterator_to_array($attributes));
```
bwoebi, line 123: iterator_to_array does not accept arrays in PHP 8.1. Only starting PHP 8.2, but Otel supports PHP 8.1. too.

bwoebi, line 123:
```suggestion
        $allAttributes = array_merge($exceptionAttributes, \is_array($attributes) ? $attributes : iterator_to_array($attributes));
```
iterator_to_array does not accept arrays in PHP 8.1. Only starting PHP 8.2, but Otel supports PHP 8.1. too.

src/DDTrace/OpenTelemetry/SpanBuilder.php
```diff
@@ -92,6 +96,38 @@ public function addLink(SpanContextInterface $context, iterable $attributes = []
         return $this;
     }
 
+    public function addEvent(string $name, int $timestamp = null, iterable $attributes = []): SpanBuilderInterface
+    {
+        $attributesArray = Attributes::create($attributes);
+        $nanoTimestamp = ($timestamp !== null && $timestamp < 1e9) ? $timestamp * 1000 : ($timestamp ?? (int)(microtime(true) * 1e9)); // Convert microseconds to nanoseconds if needed
```
bwoebi, line 102: That's a weird condition. What is this `$timestamp < 1e9` check for? As far as I see times in OTel are generally nanoseconds.

src/DDTrace/OpenTelemetry/SpanBuilder.php
zacharycmontoya: This is by design, to match what our Datadog Agent does for OTEL Events when it receives OTLP traces. If there are any `exception` events on a span, then the last event that is added will be used to populate the error tagging fields on the span.

Note: We do not update the `span.Error` field when calling this method. To update the `span.Error` field, the user would have to call the OpenTelemetry `span.SetStatus` API with the Error status.

bwoebi: `error` is set automatically when `error.message` or `error.type` are present. So, at least that assumption is not correct here.

When manually adding errors, users are supposed to only add `error.message` and `error = 1` is automatically added for them.

bwoebi: Also IIRC that behaviour of populating error.message is a fallback behaviour from the ages where no span event support existed in the UI.

ext/ddtrace.c
```diff
@@ -617,6 +617,69 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+        Z_TRY_ADDREF(event->property_attributes);
+        zend_hash_add(array, attributes, &event->property_attributes);
+        zend_string_release(attributes);
+    }
+
+    RETURN_ARR(array);
+}
+
+PHP_METHOD(DDTrace_SpanEvent, __construct) {
+    (void)return_value;  // Suppress unused parameter warning
```
bwoebi, line 650:
```suggestion
    UNUSED(return_value);
```

ext/ddtrace.c
```diff
@@ -617,6 +617,69 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+    if (timestamp == 0) {
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        ZVAL_LONG(&event->property_timestamp, ts.tv_sec * 1e9 + ts.tv_nsec);
+    } else {
+        ZVAL_LONG(&event->property_timestamp, timestamp * 1e9);
+    }
+
+    if (attributes && Z_TYPE_P(attributes) == IS_ARRAY) {
+        ZVAL_DUP(&event->property_attributes, attributes);
```
bwoebi, line 677:
```suggestion
        ZVAL_COPY(&event->property_attributes, attributes);
```

ext/ddtrace.c
```diff
@@ -617,6 +617,69 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+
+    // Convert the Unix timestamp to nanoseconds
+    if (timestamp == 0) {
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
+        ZVAL_LONG(&event->property_timestamp, ts.tv_sec * 1e9 + ts.tv_nsec);
+    } else {
+        ZVAL_LONG(&event->property_timestamp, timestamp * 1e9);
+    }
+
```
bwoebi, line 675:
```suggestion

    zval garbage;
    ZVAL_COPY_VALUE(garbage, &event->property_attributes);
```
And then later free (after replacing): `zval_ptr_dtor(&garbage);`. Same for the other values.

ext/ddtrace.c
```diff
@@ -621,6 +636,69 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+    ZEND_PARSE_PARAMETERS_END();
+
+    ddtrace_span_event *event = (ddtrace_span_event*)Z_OBJ_P(ZEND_THIS);
+
+    ZVAL_STR_COPY(&event->property_name, name);
+
+    // Use the provided timestamp or the current time in nanoseconds
+    if (timestamp == 0) {
+        struct timespec ts;
+        clock_gettime(CLOCK_REALTIME, &ts);
```
bwoebi, line 689: You may use
```
    timespec_get(&ts, TIME_UTC);
```
for out-of-the box compatibility with windows rather than hacking a custom clock_gettime together.

src/DDTrace/OpenTelemetry/SpanBuilder.php
link04: Hey @bwoebi, I've been looking into this ask here since I thought for us to work around this I could check if the span was an error or not, if not then reset the `error` tag but I am not sure how to accomplish this, for example I saw that the error can be removed from the `metrics` object using ` unset($this->span->metrics[Tag::ERROR]);` but couldn't find a way to do so for the body of the Span and wanted to check what you'd recommend.

bwoebi: What I would recommend is not doing this, honestly.
Like, it's a bad idea which goes against the defined semantics of the tracer.

link04: I see what you mean, will go ahead with only adding the `stack` tag for now and later connect with Error Tracking to  try getting the details from the span event instead, does that sounds better at the moment?

bwoebi: Yes.

ext/ddtrace.c
```diff
@@ -621,6 +621,139 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+        array_init(&parent_construct_params[1]);
+    }
+
+    // Call the parent constructor with the prepared parameters
+    zval parent_construct_retval;
+    zend_fcall_info fci;
+    zend_fcall_info_cache fci_cache;
+
+    zend_class_entry *parent_ce = Z_OBJCE_P(ZEND_THIS)->parent;
+    zend_function *parent_constructor = zend_hash_str_find_ptr(
```
bwoebi, line 723: Why so complicated and not simply extract a static function `void dd_span_event_construct(zend_string *name, zend_array *attributes, zend_long timestamp)` which handles setting the name and the attributes.

Also the code here doesn't assign the exception property at all, right?

ext/ddtrace.stub.php
```diff
@@ -23,6 +23,57 @@
...
+        
+        /**
+         * @var \Throwable
+         */
+        private \Throwable $exception;
+
+        /**
+         * @var string[] $attributes
+         */
+        public array $attributes;
```
bwoebi, line 74: Redundant, because inherited.
```suggestion
```

ext/ddtrace.c
```diff
@@ -621,6 +621,139 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+
+    Z_TRY_ADDREF(event->property_name);
+    zend_hash_add(array, name, &event->property_name);
+    Z_TRY_ADDREF(event->property_timestamp);
+    zend_hash_add(array, timestamp, &event->property_timestamp);
+
+    zend_string_release(name);
+    zend_string_release(timestamp);
+
+    if (Z_TYPE(event->property_attributes) != IS_NULL && Z_ARRVAL(event->property_attributes)->nNumOfElements > 0) {
```
bwoebi, line 643:
```suggestion
    if (Z_TYPE(event->property_attributes) == IS_ARRAY && Z_ARRVAL(event->property_attributes)->nNumOfElements > 0) {
```

bwoebi (changes requested):
The code also misses phpt tests (in `tests/ext` folder), for manual usage of DDTrace\SpanEvent and DDTrace\ExceptionSpanEvent.

ext/ddtrace.c
```diff
@@ -621,6 +621,117 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
 }
 
+static void dd_span_event_construct(ddtrace_span_event *event, zend_string *name, zval *attributes, zend_long timestamp)
+{
+    zval garbage_name, garbage_attributes;
+
+    // Initialize garbage zvals to avoid double free errors
+    ZVAL_UNDEF(&garbage_name);
+    ZVAL_UNDEF(&garbage_attributes);
+
```
bwoebi, line 631:
```suggestion
```

Redundant, get's overwritten immediately anyway.

ext/ddtrace.c
```diff
@@ -621,6 +621,117 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+    ZVAL_STR_COPY(&event->property_name, name);
+
+    // Use the provided timestamp or the current time in nanoseconds
+    if (timestamp == 0) {
+        struct timespec ts;
+        timespec_get(&ts, TIME_UTC);
+        ZVAL_LONG(&event->property_timestamp, ts.tv_sec * ZEND_NANO_IN_SEC + ts.tv_nsec);
+    } else {
+        ZVAL_LONG(&event->property_timestamp, timestamp);
+    }
```
bwoebi, line 646: Nit:
```suggestion
    if (timestamp == 0) {
        struct timespec ts;
        timespec_get(&ts, TIME_UTC);
        timestamp = ts.tv_sec * ZEND_NANO_IN_SEC + ts.tv_nsec;
    }
    ZVAL_LONG(&event->property_timestamp, timestamp);
```

ext/ddtrace.c
```diff
@@ -621,6 +621,117 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+
+PHP_METHOD(DDTrace_ExceptionSpanEvent, __construct)
+{
+    UNUSED(return_value);
+
+    zval *exception;
+    zval *attributes = NULL;
+
+    ZEND_PARSE_PARAMETERS_START(1, 2)
+        Z_PARAM_OBJECT_OF_CLASS(exception, zend_ce_throwable)
```
bwoebi, line 720: Parameter is still unused.

ext/ddtrace.stub.php
```diff
@@ -23,6 +23,52 @@
...
+         *
+         * @param \Throwable $exception exception to record.
+         * @param array $attributes Optional attributes for the event.
+         */
+        public function __construct(\Throwable $exception, array $attributes = []) {}
+        
+        /**
+         * @var \Throwable
+         */
+        private \Throwable $exception;
```
bwoebi, line 69:
```suggestion
        public \Throwable $exception;
```
Is there a particular reason it's private?

ext/ddtrace.c
link04: On the latest commit moved the code for the attribute handling in here instead. Let me know what you think of that please.

bwoebi: Can you move the attribute handling to the jsonSerialize function instead? So that changes to the exception property after object construction get reflected there.

link04: So I've been trying saving it to the ExceptionSpanEvent and then in the serializer make use of that but then my `phpt` fail where my `Interopability` one doesn't, not sure if you can spot why or provide any suggestions: https://github.com/DataDog/dd-trace-php/pull/2754/commits/ef1341aa1a84643d6f830d7fa2bd2bedf0dcfc72

ext/ddtrace.c
```diff
@@ -621,6 +621,145 @@ static PHP_GSHUTDOWN_FUNCTION(ddtrace) {
...
+            }
+            add_assoc_str_ex(&combined_attributes, ZEND_STRL("exception.type"), zend_string_copy(Z_OBJCE_P(exception)->name));
+
+            // Get the exception stack trace using zai_get_trace_without_args_from_exception
+            zend_string *stacktrace = zai_get_trace_without_args_from_exception(Z_OBJ_P(exception));
+            add_assoc_str_ex(&combined_attributes, ZEND_STRL("exception.stacktrace"), stacktrace);
+        }
+    }
+
+    if (attributes && Z_TYPE_P(attributes) == IS_ARRAY) {
```
bwoebi, line 685:
```suggestion
    if (Z_TYPE_P(attributes) == IS_ARRAY) {
```
Nit: redundant.

src/DDTrace/OpenTelemetry/Span.php
```diff
@@ -475,4 +517,32 @@ private function updateSpanLinks()
...
+        $otel = [];
+        foreach ($datadogSpanEvents as $datadogSpanEvent) {
+            // Check if the event relationship exists
+            $event = ObjectKVStore::get($datadogSpanEvent, "event");
+            if ($event === null) {
+                // Create the event
+                $event = new Event(
+                    $datadogSpanEvent->name,
+                    (int)$datadogSpanEvent->timestamp,
+                    Attributes::create((array)$datadogSpanEvent->attributes ?? [])
```
bwoebi, line 535: This creates an OTel-Event, but it won't include the exception. Can you generate the right exception-attributes here on the fly? if $datadogSpanEvent instanceof ExceptionSpanEvent? (i.e. exception.stacktrace, exception.type, exception.message)?

bwoebi: I.e. what you do in SpanBuilder::recordException

bwoebi:
I see you updated libdatadog. Can you please also run `make cbindgen` in the repository root then, to keep the bindings up to date?

PR #2761, opened 2024-07-16
Improve sampler

estringana:
### Description

There have been error when more than one request was running on parallel due to the sampler not managing it properly. This PR refactor the sampler to stop these errors.

APPSEC-53993

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/helper/sampler.hpp
+++ b/appsec/src/helper/sampler.hpp
@@ -28,31 +28,19 @@ public:
         // NOLINTEND(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
     }
 
-    bool get()
+    bool picked()
     {
-        bool result = false;
         if (sample_rate_ == 1) {
             return true;
         }
 
-        const std::lock_guard<std::mutex> lock_guard(mtx_);
-
-        if (floor(request_ * sample_rate_) !=
-            floor((request_ + 1) * sample_rate_)) {
-            result = true;
-        }
-
-        if (request_ < std::numeric_limits<unsigned>::max()) {
-            request_++;
-        } else {
-            request_ = 1;
-        }
-
-        return result;
+        auto old_request = request_.fetch_add(1, std::memory_order_relaxed);
+        return floor(old_request * sample_rate_) !=
+               floor((request_)*sample_rate_);
     }
 
 protected:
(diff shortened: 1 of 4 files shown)


appsec/src/helper/sampler.hpp
```diff
@@ -29,48 +27,19 @@ class sampler {
...
-
-            return *this;
-        }
 
-        ~scope()
-        {
-            if (concurrent_ != nullptr) {
-                concurrent_->store(false, std::memory_order_relaxed);
-            }
+    bool get()
```
Anilm3, line 31: nit: Now that this only returns a verdict / sampling decision, perhaps it should have a more meaningful name?

appsec/src/helper/sampler.hpp
```diff
@@ -29,48 +27,19 @@ class sampler {
...
         const std::lock_guard<std::mutex> lock_guard(mtx_);
 
-        std::optional<scope> result = std::nullopt;
-
-        if (!concurrent_ && floor(request_ * sample_rate_) !=
-                                floor((request_ + 1) * sample_rate_)) {
-            result = {scope{concurrent_}};
+        if (floor(request_ * sample_rate_) !=
+            floor((request_ + 1) * sample_rate_)) {
+            result = true;
```
Anilm3, line 42: All this code (including the lines below) can now be rewritten as (note that `request_ + 1` in your current version can overflow):

```cpp
auto old_request = request_;
request_ = (request_ + 1) < std::numeric_limits<unsigned>::max() ? request_ + 1 : 1;
return floor(old_request * sample_rate_) != floor((old_request + 1) * sample_rate_);

File: appsec/src/helper/sampler.hpp [hunk 3]
```diff
@@ -29,63 +27,21 @@ class sampler {
...
-        if (request_ < std::numeric_limits<unsigned>::max()) {
-            request_++;
-        } else {
-            request_ = 1;
+        if (sample_rate_ == 1) {
+            return true;
         }
 
-        return result;
+        auto old_request = request_++;
```
cataphract, line 37: this is ok, but it would be better to do `request_.fetch_add(1, std::memory_order_relaxed)`

cataphract (approved):

PR #2773, opened 2024-07-23
Move appsec helper to sidecar

cataphract:
### Description

Runs appsec helper inside sidecar.

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/extension/helper_process.c
```diff
@@ -226,6 +159,63 @@ bool dd_on_runtime_path_update(zval *nullable old_val, zval *nonnull new_val)
...
+    ddog_CharSlice log_level =
+        to_char_slice(get_global_DD_APPSEC_HELPER_LOG_LEVEL());
+
+    enable_appsec(helper_path, socket_path, lock_path, log_path, log_level);
+}
+
+void dd_helper_close_conn()
+{
+    if (!dd_conn_connected(&_mgr.conn)) {
+        mlog(dd_log_debug, "Not connected; nothing to do");
```
Anilm3, line 201: As good as logging nothing.

appsec/src/extension/network.c
```diff
@@ -423,3 +465,31 @@ dd_result dd_conn_set_timeout(
 
     return dd_success;
 }
+
+static void _timespec_add_ms(struct timespec *ts, long num_ms)
+{
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+    long seconds = num_ms / 1000;
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
```
Anilm3, line 473: You can use begin-end around the function:

```cpp
// NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)

// NOLINTEND(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)

File: appsec/src/helper/remote_config/http_api.cpp [hunk 3]
```diff
@@ -75,16 +90,41 @@ std::string execute_request(const std::string &host, const std::string &port,
...
     }
 
-    return result;
+    co_return result;
 }
 
+std::string execute_request_sync(const std::string &host,
+    const std::string &port, const http::request<http::string_body> &req)
+{
+
```
Anilm3, line 99:
```suggestion
```

appsec/src/helper/remote_config/http_api.cpp
```diff
@@ -75,16 +90,41 @@ std::string execute_request(const std::string &host, const std::string &port,
...
+    net::co_spawn(ioc, std::move(client_coroutine),
+        [&](const std::exception_ptr &eptr, std::string body) {
+            if (eptr) {
+                promise.set_exception(eptr);
+            } else {
+                promise.set_value(std::move(body));
+            }
+        });
+
+    ioc.run();
```
Anilm3, line 116: This is all very interesting... but what did we gain aside from having a timeout?

Anilm3 (approved):
Overall looks good, a bit hard to follow but otherwise makes sense, some extra comments:
- The helper fuzzer has to be updated, as it uses classes that have been changed (.e.g config)
- There are some ASAN leaks from ddtrace which should be suppressed before merging
- Some of the Rust changes are failing to compile.
-

appsec/src/helper/main.cpp
```diff
@@ -3,81 +3,156 @@
...
+            SPDLOG_INFO("AppSec helper finished");
+            return 0;
+        }
+        if (std::chrono::steady_clock::now() >= deadline) {
+            SPDLOG_WARN("Could not finish AppSec helper before deadline");
+            return 1;
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds{10}); // NOLINT
+    }
+    spdlog::shutdown();
```
Anilm3, line 157: Missing return statement

appsec/src/helper/remote_config/http_api.cpp
cataphract: The whole point is having a timeout. Having no timeout in a network operation, in a daemon on top of it, is unacceptable, as the operation can block indefinitely, preventing further communication.

appsec/src/extension/ddappsec.c
```diff
@@ -254,7 +258,7 @@ static PHP_RINIT_FUNCTION(ddappsec)
     // Safety precaution
     DDAPPSEC_G(during_request_shutdown) = false;
 
-    pthread_once(&_rinit_once_control, _rinit_once);
+    dd_appsec_rinit_once();
```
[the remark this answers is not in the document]
cataphract: I guess you could put it behind a check for testing mode (though even there it only rely on what happens in ddtrace, which introduces coupling), but you would gain little in performance, as `pthread_once` does the name (load a value and does a comparison against it). So you only save a function call.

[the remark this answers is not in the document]
Anilm3: I wasn't very clear, I was speaking about the extension in general not being required under the circumstances on which this call would be used (except testing).

appsec/src/helper/remote_config/http_api.cpp
Anilm3: I understand the purpose of the timeout of course, I was more interested in any other potential benefits if any.

appsec/src/extension/php_compat.c
```diff
@@ -6,6 +6,13 @@
 #include "php_compat.h"
 
 #if PHP_VERSION_ID < 70300
+static zend_string _zend_empty_string_st = {
+    .gc.refcount = 1,
```
bwoebi, line 10: Are you sure it's fine? IIRC interned strings need at least RC=2. But maybe on 7.2 and prior RC=1 was fine too.

PR #2776, opened 2024-07-25
Simplify appsec tracer tests

estringana:
### Description

Reduce duplication on `appsec` integration tests on the tracer suite

Summary:
On appsec we had two main tests `AutomatedLoginEventsTest` and `PathParamsTest` per framework version. This two test files were 99% identical given a framework. This PR extract a parent clase per framework and then each version extends it and when necessary, change implementation

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/Appsec/Mock.php
+++ b/tests/Appsec/Mock.php
@@ -63,7 +63,7 @@ if (!class_exists('datadog\appsec\AppsecStatus')) {
             $this->getDbPdo()->exec(sprintf("INSERT INTO appsec_events VALUES ('%s')", json_encode($event)));
         }
 
-        public function getEvents()
+        public function getEvents(array $names = [], array $addresses = [])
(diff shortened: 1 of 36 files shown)


tests/Appsec/Mock.php
```diff
@@ -63,7 +63,7 @@ public function addEvent(array $event, $eventName)
             $this->getDbPdo()->exec(sprintf("INSERT INTO appsec_events VALUES ('%s')", json_encode($event)));
         }
 
-        public function getEvents()
```
estringana, line 66: Changing the signature of this method not necessary on this PR but on a incoming PR(which is the one which triggered this refactor), this new signature is required. It's not changing any behaviour on this PR

PROFeNoM (approved):

PR #2758, opened 2024-07-12
PHP 8.4 support

realFlowControl:
### Description

This PR will bring support for PHP 8.4 🎉 

PROF-10137

### Reviewer checklist
- [ ] Test coverage seems ok.
- [x] Appropriate labels assigned.

PR #2781, opened 2024-07-31
Optimize circleci build times

bwoebi:
This is not the final incantation (i.e. more to come in future PRs), but containing all changes affecting multiple types of jobs at once:

- package extension is parallelized, now taking only 10 min; also uses `ln` instead of `cp` to save time copying things around.
  - There also was a redundant tar -> untar cycle in profiling, where the profiler artifacts were first tar-ed together, then untared at the expected location for the packaging script. This was a historic relic from the past and has now been unified.
  - This needs tar 1.35 to avoid tar exiting with exit code 1 if any inode link count changes in parallel... I'm also using that occasion to move the fpm packaging docker image to datadog/dd-trace-ci (instead of datadog/docker-library)
- all jobs depending on package extension fetch the artifact via API taking tens of seconds instead of attaching the workspace, taking minutes; also saving the package extension step from needing to persist its workspace
- Prepare Code is no longer a dependency of compiling-only tasks (the only relevant step was `append_build_id`, but that one is nearly zero-cost and can just be done in the individual jobs too), giving an 1 minute headstart compared to before.
- I think the ramdisk usage disappeared when we introduced setup_docker instead of the docker image executor. Putting it back into setup_docker. And using it for libdatadog compilation too.
- Using the cargo cache based on the root `Cargo.lock` now, rather than having it profiler specific.
- Compiling the rust (in an xlarge executor - but needs to be done only 3 times and just for 2.5 min, so...) and extension archives in separate jobs and (very quickly) linking it together in the individual tests (having a job in between to link the archives together would increase run time too much, so we just do that in the test jobs).
  - Given that the rust compilation runs in parallel with Prepare Code here, we only have about 1 min 50 sec overhead due to rust compilation. This is quite faster than the 3 min 45 sec it took before. So, it essentially means the jobs are completed 2 min earlier, as well as shaving off some cost in every single test.
- Windows compilation now compiles the rust code only once instead of twice (i.e. nts and zts), saving 3 minutes.
- Windows avoids fetching appsec git submodules, saving about 15 seconds. The filesystem I/O itself seems to be slow on windows. Mostly due to the expansive tests/ directory.

The windows docker image also has been slimmed down, which nearly halves fetching time. (But done in a different PR.)

Also noteworthy non-starters:
- git on windows takes more than a minute
  - Fetching --depth 1 git repository - saves nothing
  - Caching the git repository on windows - saving a few seconds

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -244,6 +244,24 @@ aliases:
         # Avoid intermittent DNS hangs: See https://github.com/curl/curl/issues/593#issuecomment-170146252
         echo "options single-request" | sudo tee -a /etc/resolv.conf
 
+  - &STEP_LINK_EXTENSION
+    run:
+      name: Link extension
+      background: true
+      command: |
+        echo $$ > /tmp/link_extension.pid
+        sed -i 's/-export-symbols .*\/ddtrace\.sym/-Wl,--retain-symbols-file=ddtrace.sym/g' ddtrace.ldflags
+        gcc -shared -Wl,-whole-archive ddtrace.a -Wl,-no-whole-archive $(cat ddtrace.ldflags) libddtrace_php.a -Wl,-soname -Wl,ddtrace.so -o ddtrace.so
+        mkdir -p tmp/build_extension/modules
+        mv ddtrace.so tmp/build_extension/modules/
+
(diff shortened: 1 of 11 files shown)


realFlowControl (approved):
I have a vision, I see lot's of merge conflicts with #2758 😉

PR #2785, opened 2024-08-08
Bump the version of the CI builds

bwoebi:
This ensures we can compare capabilities against e.g. system tests, distinguishing old and new release, otherwise everything would be the same version for them.


Anilm3 (approved):
Looks good to me, thanks!

PR #2790, opened 2024-08-14
Fix fatal error when reading uninit $meta on a span

bwoebi:
This seemed to happen in appsec preload shutdown.

I'm not sure _why_ meta was uninit in the first place, (yes a call to ddtrace_span_data_free_storage will make it undef, but why was that called - it's not a refcounting issue at least?!, I mean it wasn't an use after free?!).

### DIFF
--- a/appsec/src/extension/ddtrace.c
+++ b/appsec/src/extension/ddtrace.c
@@ -248,14 +248,14 @@ static zval *_get_span_modifiable_array_property(
 {
 #if PHP_VERSION_ID >= 80000
     zval *res =
-        zobj->handlers->get_property_ptr_ptr(zobj, propname, BP_VAR_R, NULL);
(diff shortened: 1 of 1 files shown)


appsec/src/extension/ddtrace.c
```diff
@@ -248,14 +248,14 @@ static zval *_get_span_modifiable_array_property(
 {
 #if PHP_VERSION_ID >= 80000
     zval *res =
-        zobj->handlers->get_property_ptr_ptr(zobj, propname, BP_VAR_R, NULL);
+        zobj->handlers->get_property_ptr_ptr(zobj, propname, BP_VAR_IS, NULL);
```
Anilm3, line 251: Looking at the code, when the span has been invalidated (which also isn't clear to me why that's happening), won't `get_property_ptr_ptr` return `NULL`? In which case this will crash instead.

Anilm3: Ah no, it should be fine.

PR #2793, opened 2024-08-15
feat(profiling) add timestamp to exception sample

realFlowControl:
### Description

This will add a timestamp to the exception sample in order to show it in the timeline view, see also:
- https://github.com/DataDog/profiling-backend/pull/5730
- https://github.com/DataDog/web-ui/pull/152074

image

PROF-10122

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -684,7 +684,9 @@ impl Profiler {
                 let (wall_time, cpu_time) = CLOCKS.with(|cell| cell.borrow_mut().rotate_clocks());
 
                 let labels = Profiler::common_labels(0);
-                let mut timestamp = 0;
+                let n_labels = labels.len();
+
+                let mut timestamp = NO_TIMESTAMP;
                 #[cfg(feature = "timeline")]
                 {
                     let system_settings = self.system_settings.load(Ordering::SeqCst);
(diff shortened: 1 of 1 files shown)


profiling/src/profiling/mod.rs
```diff
@@ -790,14 +790,26 @@ impl Profiler {
...
+                let mut timestamp = NO_TIMESTAMP;
+                #[cfg(feature = "timeline")]
+                {
+                    let system_settings = self.system_settings.load(Ordering::SeqCst);
+                    // SAFETY: system settings are stable during a request.
+                    if unsafe { *ptr::addr_of!((*system_settings).profiling_timeline_enabled) } {
+                        if let Ok(now) = SystemTime::now().duration_since(UNIX_EPOCH) {
+                            timestamp = now.as_nanos() as i64;
+                        }
+                    }
```
morrisonlevi, line 802: Do you think it's worth a helper routine for this? Or duplication is fine/better?

morrisonlevi (approved):
Looks good to me, just one question about potential Don't-Repeat-Yourself.

profiling/src/profiling/mod.rs
realFlowControl: I'm looking into this in another PR, good catch

PR #2802, opened 2024-08-20
Send x-datadog-test-session-token metric and send metrics to request-replayer

bwoebi:
This allows for clear test session separation without interference from other processes.

Fixing the currently flaky OpenAI tests.


iamluc (approved):
Nice!

PR #2822, opened 2024-08-29
Do not implicitly fall back to spl_autoload() if no autoloader is registered on PHP 7

bwoebi:
This could lead to accidental inclusion of files, breaking __autoload() and possibly including the wrong file.


tests/ext/autoload-php-files/legacy_autoloader.phpt
```diff
@@ -0,0 +1,23 @@
+--TEST--
+Execute the default spl_autoload implementation if spl_autoload_register() is called without args
```
iamluc, line 2: Wrong comment

iamluc (approved):
Except the comment of one test, LGTM!

PR #2823, opened 2024-08-29
Improve on randomized tests flakiness

bwoebi:
Make randomized tests green again!

There may still be some I/O related issues in prepare step, but this should cover most of them.
Also collect strace in non-asan tests to have a bit of an insight what goes wrong, when things go wrong.


iamluc (approved):
What good news!

PR #2833, opened 2024-09-03
Fix DD_AUTOLOAD_NO_COMPILE

bwoebi:
This mode is supposed to bring some performance improvements, but it turns out it was always disabled, due to the type confusion.

### DIFF
--- a/ext/configuration.h
+++ b/ext/configuration.h
@@ -92,7 +92,7 @@ enum ddtrace_sampling_rules_format {
 
 #define DD_CONFIGURATION_ALL                                                                                   \
     CONFIG(STRING, DD_TRACE_SOURCES_PATH, DD_DEFAULT_SOURCES_PATH, .ini_change = zai_config_system_ini_change) \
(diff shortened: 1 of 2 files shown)


ext/configuration.h
```diff
@@ -92,7 +92,7 @@ enum ddtrace_sampling_rules_format {
 
 #define DD_CONFIGURATION_ALL                                                                                   \
     CONFIG(STRING, DD_TRACE_SOURCES_PATH, DD_DEFAULT_SOURCES_PATH, .ini_change = zai_config_system_ini_change) \
-    CONFIG(STRING, DD_AUTOLOAD_NO_COMPILE, "0", .ini_change = zai_config_system_ini_change)                    \
+    CONFIG(BOOL, DD_AUTOLOAD_NO_COMPILE, "0", .ini_change = zai_config_system_ini_change)                      \
```
iamluc, line 95: Default value of other BOOL settings are either `true` of `false`.
For consistency, we should probably set `false` here instead of `0`

iamluc (approved):

ext/configuration.h
bwoebi: Is more consistent, yep :-)

PR #2836, opened 2024-09-05
Blocking from a hook is not stopping code execution

estringana:
### Description

Blocking a request from Appsec should stop customer code execution. However, when this blocking happens within a tracer hook, it does not stop executing customer code execution.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
I see, the tracer sandboxing is sandboxing the bailout away :-)
I suppose some it would be ideal to signal the tracer "please bailout again after catching this" :-D

zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -218,8 +220,25 @@ inline bool zai_sandbox_timed_out(void) {
...
+    if (strcmp("Datadog blocked the request, but the response has already been partially committed", ZSTR_VAL(PG(last_error_message))) == 0) {
+        return true;
+    }
+
+    return false;
+}
+
 inline void zai_sandbox_bailout(zai_sandbox *sandbox) {
-    if (!zai_sandbox_timed_out()) {
+    if (!zai_sandbox_timed_out() && !zai_is_request_blocked()) {
```
cataphract, line 241: I don't know why sandbox.h is written this way, with just C inline functions, but I guess that in the context of its strangeness it's fine.

cataphract (approved):
I think it's fine although it's a mystery to me why sandbox.{h,c} are written the way they are.

zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -356,8 +375,25 @@ inline bool zai_sandbox_timed_out(void) {
     return false;
 }
 
+inline bool zai_is_request_blocked(void)
```
Anilm3, line 378: Both implementations of this function are exactly the same, perhaps you can define it only once?

zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -218,8 +220,25 @@ inline bool zai_sandbox_timed_out(void) {
...
+{
+    if (!PG(last_error_message)) {
+        return false;
+    }
+
+    if (strcmp("Datadog blocked the request and presented a static error page", ZSTR_VAL(PG(last_error_message))) == 0) {
```
Anilm3, line 229: My main concern with this approach is that if someone changes the error message in appsec, it would stop working. Can you make changes on the appsec side to:
- Make sure the error messages are validated at compile time?
- Add comments specifying that the error message must not be changed.

zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -218,8 +220,25 @@ inline bool zai_sandbox_timed_out(void) {
...
+        return true;
+    }
+
+    if (strcmp("Datadog blocked the request, but the response has already been partially committed", ZSTR_VAL(PG(last_error_message))) == 0) {
```
Anilm3, line 233: Alternatively you could check the prefix, which is always `Datadog blocked the request`

zend_abstract_interface/sandbox/sandbox.h
```diff
@@ -218,8 +220,25 @@ inline bool zai_sandbox_timed_out(void) {
     return false;
 }
 
+inline bool zai_is_request_blocked(void)
```
Anilm3, line 223: This is missing the redirection message as well.

Anilm3 (changes requested):

zend_abstract_interface/sandbox/sandbox.h
estringana: They are the same yes. However, this file is full of this duplication. I did it that way so follow the same approach

Anilm3: Unless there's a good technical reason, don't duplicate the function.

estringana: I get your point but have a look to the whole file. It done using this approach and I will follow it. In fact now the functions aren't the same anymore

zend_abstract_interface/sandbox/sandbox.h
estringana: Done

zend_abstract_interface/sandbox/sandbox.h
estringana: Done

bwoebi (approved):
Looks good to me.

Anilm3 (approved):

PR #2838, opened 2024-09-06
Loader: Use the same version number as for other extensions

iamluc:
### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Bit hacky, (I'd prefer a solution in config.m4), but I guess good enough.

PR #2840, opened 2024-09-06
Track peak memory usage in root span metrics

### LINKED ISSUE #2832: [Feature] Add memory_get_peak_usage to trace information
### Describe the feature you'd like

We're running our applications in containers on Kubernetes, the `memory_limit` in PHP's configuration (`php.ini`) is still set to legacy 1024MB (this is an old value, and has never changed).

Nobody dares to lower it, because we not have any statistics how much memory we're using.

When checking out a Datadog trace (from PHP) I'm seeing:

```json
{
  "php": {
    "compilation": {
      "total_time_ms": 0
    }
  }
}
```

Maybe it's an idea, to add : `memory_get_peak_usage()` to this `php` information, this could be really useful for us.

For example: 

```json
{
  "php": {
    "compilation": {
      "total_time_ms": 0
    },
    "memory": {
      "peak_usage_bytes": 1337
    }
  }
}
```

Also it would be really useful if we're able to add graphs on a Dashboard, with memory usage :).

### Is your feature request related to a problem?

It's hard lowering the value of `memory_limit` because it's not clear how much memory we're using.

### Describe alternatives you've considered

Writing these metrics to an OpenMetrics store, so Prometheus can scrape these.

### Additional context

_No response_

bwoebi:
Fixes #2832.

Adding `php.memory.peak_usage_bytes` and `php.memory.peak_real_usage_bytes` metrics.

### DIFF
--- a/appsec/tests/extension/client_init_record_span_tags.phpt
+++ b/appsec/tests/extension/client_init_record_span_tags.phpt
@@ -109,4 +109,6 @@ Array
     [_dd.agent_psr] => 1
     [_sampling_priority_v1] => 1
     [php.compilation.total_time_ms] => %f
+    [php.memory.peak_usage_bytes] => %f
+    [php.memory.peak_real_usage_bytes] => %f
 )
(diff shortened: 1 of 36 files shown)


PROFeNoM (approved):
Thanks for the quick addition :-)

PR #2843, opened 2024-09-10
Crashtracker: fix a use-after-free error

iamluc:
### Reviewer checklist
- [ ] Test coverage seems ok.


iamluc:
> Do you want to also toggle the default of the LOG_BACKTRACES setting if compiled with asan (via preprocessor directive)?

I find it easier to disable the whole signal handling. WDYT?

PR #2853, opened 2024-09-16
Loader: disable the tracer in case of a potential incompatibility (extension/jit)

iamluc:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/loader/dd_library_loader.c
+++ b/loader/dd_library_loader.c
@@ -8,6 +8,7 @@
 #include <php.h>
 #include <php_ini.h>
 #include <stdbool.h>
+#include <main/SAPI.h>
 #include <ext/standard/basic_functions.h>
 
 #include "compat_php.h"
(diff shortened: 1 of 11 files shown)


loader/dd_library_loader.c
```diff
@@ -328,6 +334,16 @@ static PHP_MINIT_FUNCTION(ddloader_injected_extension_minit) {
         return SUCCESS;
     }
 
+    char *incompatible_exts[] = {"Xdebug", "the ionCube PHP Loader", "newrelic", "blackfire", "pcov"};
```
bwoebi, line 337: They're not all incompatible? especially not xdebug. We even actively test for compatibility with it (tests/xdebug).

What I meant: add to https://github.com/DataDog/dd-trace-php/blob/master/ext/excluded_modules.c the extensions which are generally incompatible (ioncube).

Add here, extensions which have a risk to segfault when run alongside dd-trace. Afaik pcov, blackfire, xdebug and newrelic work nicely together.

iamluc: > They're not all incompatible? especially not xdebug

We have crash reports related to Xdebug, and we know that some versions contain bugs that can cause crashes. 
So IMHO, in the context of SSI, it's safer not to inject the tracer at all when any other extensions that might use the same internal hooks as we do are loaded, rather than disabling it.

That's why I made the changes in the loader instead of the tracer.

bwoebi: We do check the compatibility before actually setting any hooks.

loader/dd_library_loader.c
```diff
@@ -69,13 +139,36 @@ static void ddtrace_pre_minit_hook(void) {
...
+        if (force_load) {
+            LOG(WARN, "OPcache JIT is enabled and may cause instability. Ignoring as DD_INJECT_FORCE is enabled");
+        } else {
+            LOG(WARN, "OPcache JIT is enabled and may cause instability. ddtrace will be disabled unless the environment DD_INJECT_FORCE is set to '1', 'true', 'yes' or 'on'");
+            disable_tracer = true;
+        }
+    }
+
+    if (disable_tracer) {
+        ddloader_ini_set_configuration(ZEND_STRL("datadog.trace.enabled"), ZEND_STRL("0"));
```
bwoebi, line 171: You actually want `ddtrace.disable=1`. `datadog.trace.enabled=0` still prepares most of the process (setting hooks, etc., just skips the actual tracing part).

loader/dd_library_loader.c
```diff
@@ -460,26 +554,19 @@ static int ddloader_load_extension(unsigned int php_api_no, char *module_build_i
     config->module_number = module_entry->module_number;
     config->version = (char *)module_entry->version;
 
-    return SUCCESS;
+    goto ok;
```
bwoebi, line 557: Nice catch :-)

loader/valgrind.supp
```diff
@@ -0,0 +1,29 @@
+{
+   <leak from zai interceptor>
```
bwoebi, line 2: What's that zai interceptor leak? Sounds like a mistake in our extension?

loader/dd_library_loader.c
iamluc: Thanks Valgrind :laughing:

loader/valgrind.supp
bwoebi: It's really just `zend_hash_destroy(&zai_hook_static_inheritors)` missing in `zai_hook_mshutdown`.

bwoebi: The fact that it's in the appsec suppression file probably just means "not our problem" for them :-)

iamluc: Sure :stuck_out_tongue:  But I'm also surprised it hasn't be reported by our ddtrace test suite :thinking:

bwoebi (approved):
Looks good to me, thanks :-)

PR #2854, opened 2024-09-16
Fix improper symbol exports in appsec ext/helper

cataphract:
### Description

Both the extension and the helper need version scripts to prevent symbols from being exported. -fvisibility=hidden is not sufficient because it's not being applied to some of the static libraries they link against.

On the extensions side: the path to the version script was wrong, as this error was ultimately ignored because invalid ld options were being ingored. This resulted in some extra zai_* functions being exported.

On the helper side: symbols from libunwind, libc++ and possibly others were being exported.

### Reviewer checklist
- [ ] Test coverage seems ok.


Anilm3 (approved):
lgtm

bwoebi (approved):
Works for me, thanks :-)

PR #2855, opened 2024-09-16
Remove gid from socket/lock path of helper

cataphract:
### Description

Now that we're inside sidecar, we need to have the same uniqueness characteristics of sidecar; in particular, there's one instance of sidecar per uid, but not per (uid,gid) as was the case with the helper before.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Looks trivial and is going to fix the existing issue.

Anilm3 (approved):

PR #2857, opened 2024-09-16
add aws credentials to circleci system tests

wconti27:
### Description
add necessary auth for new aws system tests and add IDM scenarios of interest

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -4866,27 +4866,25 @@ workflows:
           build: |
             cd system-tests
             ./build.sh php
-          environment:
-            - AWS_ACCESS_KEY_ID: $SYSTEM_TESTS_IDM_AWS_ACCESS_KEY_ID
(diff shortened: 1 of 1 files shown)


bwoebi:
I sort-of don't really like having more secrets than needed in CI. What do we need _aws_ credentials for, in system tests?

wconti27:
> I sort-of don't really like having more secrets than needed in CI. What do we need _aws_ credentials for, in system tests? Can't these tests run locally, somehow? I'd strongly prefer having tests without external dependencies.

We need AWS tests for correctly testing propagation between Lambda and messaging systems (SQS, SNS, Kinesis). The previous 3rd party software that we were using (Localstack) was not cutting it. It was not a 1 to 1 matchup of feature support compared to actual AWS, and was providing both false negatives and false positives for our tests. This change has already been made in the other tracers.

bwoebi (approved):
Okay, let's merge it.
But definitely first merge conti/fix-php-errors on the system-tests repo.

PR #2859, opened 2024-09-17
Fix orderly helper appsec shutdown

cataphract:
The sent SIGUSR1 was just terminating the process.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/helper/main.cpp
+++ b/appsec/src/helper/main.cpp
@@ -10,6 +10,8 @@
 #include "config.hpp"
 #include "runner.hpp"
 #include "subscriber/waf.hpp"
+#include <csignal>
+#include <cstdlib>
 #include <spdlog/common.h>
 #include <spdlog/sinks/basic_file_sink.h>
 #include <spdlog/sinks/stdout_sinks.h>
@@ -17,7 +19,6 @@
 #include <string_view>
 
 extern "C" {
-#include <csignal>
 #include <fcntl.h>
 #include <pthread.h>
 #include <sys/file.h>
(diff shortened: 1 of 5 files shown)


appsec/src/helper/main.cpp
```diff
@@ -149,7 +159,12 @@ appsec_helper_shutdown() noexcept
             return 0;
         }
         if (std::chrono::steady_clock::now() >= deadline) {
-            SPDLOG_WARN("Could not finish AppSec helper before deadline");
+            // we need to call exit() to avoid a segfault in the still running
+            // helper threads after the helper shared library is unloaded by
+            // trampoline.c
+            SPDLOG_WARN("Could not finish AppSec helper before deadline. "
+                        "Calling exit().");
+            std::exit(EXIT_FAILURE); // NOLINT
```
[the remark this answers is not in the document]
cataphract: I've replaced it with __builtin_unreachable()

appsec/src/helper/network/acceptor.cpp
```diff
@@ -80,13 +80,10 @@ socket::ptr acceptor::accept()
...
-    int s;
-    do {
-        // NOLINTNEXTLINE
-        s = ::accept(sock_, reinterpret_cast<struct sockaddr *>(&addr), &len);
-    } while (s == -1 && errno == EINTR);
+    // NOLINTNEXTLINE
+    int s = ::accept(sock_, reinterpret_cast<struct sockaddr *>(&addr), &len);
     if (s == -1) {
-        if (errno == EAGAIN) {
+        if (errno == EINTR || errno == EAGAIN) {
```
Anilm3, line 86: nit: EINTR in this case isn't really caused by a timeout is it? So perhaps a different exception would make sense, although in a practical sense it doesn't matter.

Anilm3 (approved):

appsec/src/helper/network/acceptor.cpp
cataphract: actually I really don't like the exception here. I think the only reason it uses an exception is because an empty result is taken to have a different meaning. But I'll change it for a more complex return value

PR #2860, opened 2024-09-18
appsec ext: retry on connection refused

cataphract:
### Description

After sidecar is launched but before the helper has a chance to create and listen to the unix socket, the extension may try to  connect and fail. We retried already if the socket did not exist. But the there was a socket file from a previous run, the error is ECONNREFUSED. So retry in that case as well.

### Reviewer checklist
- [ ] Test coverage seems ok.


Anilm3 (approved):
lgtm

PR #2862, opened 2024-09-26
Fix crash with git metadata injection

iamluc:
### Description

It can crash because `property_commit` or `property_repository` can be NULL.
But the calls to `zend_string_addref` are not needed as `add_git_info` will do a copy.

### Reviewer checklist
- [ ] Test coverage seems ok.


PROFeNoM (approved):
You're right. It is indeed possible for one or the other to be NULL... Thanks :bow:

PR #2864, opened 2024-09-30
Move helper to sidecar remote config

cataphract:
### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/helper/remote_config/client.cpp
```diff
@@ -6,240 +6,186 @@
...
-    : api_(std::move(arg_api)), id_(dds::generate_random_uuid()),
-      sid_(std::move(sid)), settings_(std::move(settings)),
-      listeners_(std::move(listeners))
+client::client(remote_config::settings settings,
+    std::vector<std::shared_ptr<listener_base>> listeners)
+    : reader_{ddog_remote_config_reader_for_path(settings.shmem_path.c_str()),
+          ddog_remote_config_reader_drop},
+      settings_{std::move(settings)}, listeners_{std::move(listeners)}
 {
+    assert(settings.enabled == true); // NOLINT
```
Anilm3, line 69:
```suggestion
    assert(settings_.enabled == true); // NOLINT
```

Anilm3 (changes requested):
Still reviewing and testing, so far:

Some configurations are being ignored but I haven't checked why:

```
Ignoring config with key employee/ASM_DD/29.recommended.json/config; unsupported product
```

The capabilities are not correct, the list should be:

```
ASM_ACTIVATION
ASM_EXCLUSIONS
ASM_CUSTOM_BLOCKING_RESPONSE
ASM_REQUEST_BLOCKING
ASM_RESPONSE_BLOCKING
ASM_CUSTOM_RULES
ASM_TRUSTED_IPS
ASM_DD_RULES
ASM_IP_BLOCKING
ASM_USER_BLOCKING
```

And updates seem to performed on every poll even if no changes have been made to the actual configurations.

appsec/src/helper/remote_config/config.cpp
```diff
@@ -0,0 +1,118 @@
...
+        if (org_id_end != std::string_view::npos) {
+            sv.remove_prefix(org_id_end + 1);
+            auto product_end = sv.find('/');
+            if (product_end != std::string_view::npos) {
+                return known_products::for_name(sv.substr(0, product_end));
+            }
+        }
+    } else if (sv.starts_with("employee/"sv)) {
+        sv.remove_prefix("employee/"sv.length());
+        auto product_end = shm_path.find('/');
```
Anilm3, line 35:
```suggestion
        auto product_end = sv.find('/');
```

appsec/tests/integration/src/test/bin/enable_extensions.sh
```diff
@@ -24,6 +24,7 @@ if [[ -f /appsec/ddappsec.so && -d /project ]]; then
     echo datadog.appsec.helper_path=/appsec/libddappsec-helper.so
     echo datadog.appsec.helper_log_file=/tmp/logs/helper.log
     echo datadog.appsec.helper_log_level=info
+#    echo datadog.appsec.helper_log_level=debug
```
estringana, line 27: should this be not commented out?

appsec/src/helper/remote_config/client.cpp
```diff
@@ -6,240 +6,186 @@
...
+    std::vector<std::shared_ptr<listener_base>> listeners)
 {
-    return std::make_unique<client>(std::make_unique<http_api>(settings.host,
-                                        std::to_string(settings.port)),
-        std::move(sid), settings, std::move(listeners));
+    return std::unique_ptr<client>{new client(settings, std::move(listeners))};
 }
 
-[[nodiscard]] protocol::get_configs_request client::generate_request() const
+bool client::poll()
```
estringana, line 87: Maybe I missed it but is there any tests on helper tests checking this? I saw the client_test got removed

appsec/src/helper/remote_config/client.cpp
```diff
@@ -6,240 +6,186 @@
...
+    }
+    SPDLOG_DEBUG("Runtime id is {}", resp.substr(0, pos_lf));
 
-    const protocol::client_state cs{targets_version_, config_states,
-        !last_poll_error_.empty(), last_poll_error_, opaque_backend_state_};
-    std::vector<std::string> products_str;
-    products_str.reserve(products_.size());
-    for (const auto &[product_name, product] : products_) {
-        products_str.push_back(product_name);
+    std::string_view configs = resp.substr(pos_lf + 1);
```
Anilm3, line 109: This will throw `std::out_of_range` if `pos_lf + 1` is beyond `size()`, although that might be the intended behaviour.

appsec/src/helper/remote_config/client.cpp
```diff
@@ -6,240 +6,186 @@
...
+    while (!configs.empty()) {
+        auto pos_lf = configs.find('\n');
+        if (pos_lf == std::string_view::npos) {
+            break;
+        }
+        new_configs.emplace(config::from_line(configs.substr(0, pos_lf)));
+        configs = configs.substr(pos_lf + 1);
```
Anilm3, line 116: idem

appsec/src/helper/remote_config/config.cpp
```diff
@@ -0,0 +1,118 @@
...
+            // NOLINTNEXTLINE(bugprone-lambda-function-name)
+            SPDLOG_WARN(
+                "Failed to unmap shared memory: {}", strerror_ts(errno));
+        }
+    }};
+
+    std::string result;
+    result.resize(shm_stat.st_size);
+
+    std::copy_n(static_cast<char *>(shm_ptr), shm_stat.st_size, result.begin());
```
Anilm3, line 114: Perhaps for a future iteration, would it make sense to just abstract the mapped memory and just read it as needed? Or would that cause some sort of data race with the writer?

appsec/src/extension/ddappsec.c
```diff
@@ -378,6 +395,22 @@ static void _check_enabled()
     };
 }
 
+__attribute__((visibility("default"))) void dd_appsec_rc_conf(
+    bool *nonnull appsec_features, bool *nonnull appsec_conf) // NOLINT
+{
+    bool prev_enabled = DDAPPSEC_G(enabled);
+    bool prev_active = DDAPPSEC_G(active);
+    bool prev_to_be_configured = DDAPPSEC_G(to_be_configured);
+    _check_enabled();
```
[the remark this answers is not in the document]
cataphract: yeah right, the restoration should happen at the end...

appsec/src/helper/engine.cpp
```diff
@@ -19,17 +17,16 @@
 
 namespace dds {
 
-void engine::subscribe(const subscriber::ptr &sub)
+void engine::subscribe(std::unique_ptr<subscriber> sub)
 {
-    auto common = std::atomic_load(&common_);
-    common->subscribers.emplace_back(sub);
+    common_->subscribers.emplace_back(std::move(sub));
```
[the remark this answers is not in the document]
cataphract: Subscription is effectively part of the engine creation (in `engine::from_settings`), so this is not a problem. `common_->subscribers` is effectively final. What is changed on engine update is actually `common_`, at that one is updated atomically in `engine::update`.

appsec/src/helper/engine.cpp
```diff
@@ -116,21 +118,22 @@ void engine::context::get_meta_and_metrics(
     }
 }
 
-engine::ptr engine::from_settings(const dds::engine_settings &eng_settings,
+std::unique_ptr<engine> engine::from_settings(
```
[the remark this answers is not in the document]
cataphract: The function of `engine::from_settings` is to create an engine and give ownership to the caller. If the instance will have shared ownership afterwards doesn't really concern this factory method.

appsec/src/helper/remote_config/client.cpp
cataphract: we know that `pos_lf` is valid, so `substr(pos_lf +1)` can't be out of bounds ([example](https://godbolt.org/z/K4PqP3he7))

appsec/src/helper/remote_config/config.cpp
cataphract: wdym? when is it readed needlessly? `config.read()` calls are only made on aggregators' `on_update`

appsec/src/helper/remote_config/client.cpp
cataphract: You're right, I removed it. These are tested only on the integration tests. Reading of remote config is done by calling `ddog_remote_config_read`, which is a function implemented in Rust that we import.

Any test using the real Rust read function would writing in shared memory in the format expected by that rust code. By that time, we're having the test depend on implementation details of the rust code.

And by mocking `ddog_remote_config_read`, we're testing much useful stuff anymore.

appsec/src/helper/remote_config/config.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (::fstat(fd, &shm_stat) == -1) {
+        throw std::runtime_error{
+            "Call to fstat on memory segment failed: " + strerror_ts(errno)};
+    }
+    if (shm_stat.st_uid != ::getuid()) {
+        throw std::runtime_error{"Shared memory segment does not have the "
+                                 "expected owner. Expect our uid " +
+                                 std::to_string(::getuid()) + " but found " +
+                                 std::to_string(shm_stat.st_uid)};
+    }
```
bwoebi, line 94: Why are we checking that in the first place? Maybe do it within `MAP_FAILED` condition to give a better message, but no need to check that ahead of time.

cataphract: if you're referring to the owner, to check whether the file was not actually put there by another, possibly unprivileged, user.

bwoebi: I see, that makes sense. Then the tracer should have the same check actually. However you should use `geteuid`, not `getuid` then. As the sidecar is bound to the effective user id too.

appsec/tests/integration/src/test/bin/enable_extensions.sh
cataphract: Unfortunately debug is too verbose because of the WAF. It's good for debugging, but not more normal tests runs I think.

appsec/src/helper/remote_config/config.cpp
cataphract: fwiw I've elided the copy

bwoebi: Thanks, looks good :-)

appsec/tests/extension/push_params_block_02.phpt
```diff
@@ -0,0 +1,47 @@
+--TEST--
+Push address gets blocked even when within a hook
```
Anilm3, line 2: This won't work because blocking within a hook is broken. Do we need this test in this PR?

Anilm3 (approved):
After an extensive review and testing, the only thing that remains a mystery is the missing `ASM_TRUSTED_IPS` capability. 

Please fix that before merging.

Issue #2865, opened 2024-09-30
[Bug]: Unable to load module in PHP 8.1 on ARM


kjdev:
### Bug report

Download file: dd-library-php-1.3.2-aarch64-linux-musl.tar.gz.
Distribution: Alpine 3.19

```
$ php -v
PHP 8.1.29 (cli) (built: Jun  6 2024 17:00:13) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.29, Copyright (c) Zend Technologies
$ arch
aarch64
$ sed -i 's/@release_version@/1.3.2/' datadog-setup.php
$ php datadog-setup.php --php-bin php
Downloading installable archive from https://github.com/DataDog/dd-trace-php/releases/download/1.3.2/dd-library-php-1.3.2-aarch64-linux-musl.tar.gz.
This operation might take a while.
...................
Download completed

Installed required source files to '/opt/datadog/dd-library/1.3.2'
Installing to binary: php (/usr/bin/php81)
Installing extension to /usr/lib/php81/modules
Copied '/tmp/dd-install/dd-library-php/trace/ext/20210902/ddtrace.so' to '/usr/lib/php81/modules/ddtrace.so'
Copied '/tmp/dd-install/dd-library-php/profiling/ext/20210902/datadog-profiling.so' to '/usr/lib/php81/modules/datadog-profiling.so'
Copied '/tmp/dd-install/dd-library-php/appsec/ext/20210902/ddappsec.so' to '/usr/lib/php81/modules/ddappsec.so'
Created INI file '/etc/php81/conf.d/98-ddtrace.ini'
Installation to 'php (/usr/bin/php81)' was successful
--------------------------------------------------
SUCCESS
$ php -v
PHP Warning:  PHP Startup: Unable to load dynamic library 'ddtrace.so' (tried: /usr/lib/php81/modules/ddtrace.so (Error loading shared library /usr/lib/php81/modules/ddtrace.so: Exec format error), /usr/lib/php81/modules/ddtrace.so.so (Error loading shared library /usr/lib/php81/modules/ddtrace.so.so: No such file or directory)) in Unknown on line 0
PHP Warning:  PHP Startup: [ddappsec] Failed to register user request listeners in Unknown on line 0
PHP 8.1.29 (cli) (built: Jun  6 2024 17:00:13) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.29, Copyright (c) Zend Technologies
[...]

PR #2866, opened 2024-10-01
Allocate more memory for link_extension step 

bwoebi:
Root cause of #2865 - using a large instead of a medium circleci instance.

And changing the build step so that it will properly fail now if any command fails.


morrisonlevi (approved):
Surprising that this got through somehow. I think maybe we should do a little more to ensure the artifacts are installable, but I object to nothing in this PR, so approved.

bwoebi:
@morrisonlevi It looks like it just affected a single version on a single target. But yes :-X

PR #2871, opened 2024-10-02
Enable CLI by default

bwoebi:
And enable autoflushing by default in CLI environments.

Especially with SSI, this enables us to actually instrument CLI scripts transparently as well.


iamluc (commented):
I wonder if we should set the default value to `auto` (or create a new ini config), which would be paired with an allow/deny list.

For example, I don’t think we should enable it by default for `composer` :thinking:

iamluc (approved):

PR #2874, opened 2024-10-04
Fix failing appsec tests on pipeline

estringana:
### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Yep, that's right. We missed that due to CircleCI only running appsec tests if appsec files actually change...

estringana:
> Yep, that's right. We missed that due to CircleCI only running appsec tests if appsec files actually change...

That's an interesting point. The pipeline is still failing(https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/17765/workflows/2032cd3f-dccf-4421-a563-45c35b9354e6/jobs/5318766 and a few more). Should we run appsec always? We have many functionality depending on the tracer so tecnically any change to the tracer can break appsec

PR #2884, opened 2024-10-09
fix: MongoDBIntegration condition typo

PROFeNoM:
### Description

We want to use the new methods when they both exist (and hence, use deprecated methods when either of them _doesn't_ exist). 

I made a typo and forgot a `!`. My bad.

That said, we should look in another PR why our tests didn't catch this.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Awww. I also missed that.

PR #2887, opened 2024-10-09
Fix profiler ZTS build on alpine

bwoebi:
Make switch-php accept the same inputs than on centos build so that `build_profiler` works correctly on both targets.

Apparently `build_profiler` was trying to use `switch-php` before, and if it didn't exist ... just built NTS and pretended it was the ZTS target on alpine. That obviously does not work.


morrisonlevi (approved):
I haven't checked tests, but looks good to me. Not sure why they were ever different to begin with.

PR #2888, opened 2024-10-10
AppSec: improve behavior with empty DD_SERVICE/DD_ENV

cataphract:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/request_lifecycle.c
+++ b/appsec/src/extension/request_lifecycle.c
@@ -47,6 +47,7 @@ static THREAD_LOCAL_ON_ZTS zend_array *nullable _superglob_equiv;
 static THREAD_LOCAL_ON_ZTS zend_string *nullable _client_ip;
 static THREAD_LOCAL_ON_ZTS zval _blocking_function;
 static THREAD_LOCAL_ON_ZTS bool _shutdown_done_on_commit;
+static THREAD_LOCAL_ON_ZTS bool _empty_service_or_env;
 #define MAX_LENGTH_OF_REM_CFG_PATH 31
 static THREAD_LOCAL_ON_ZTS char
     _last_rem_cfg_path[MAX_LENGTH_OF_REM_CFG_PATH + 1];
@@ -104,6 +105,9 @@ void dd_req_lifecycle_rinit(bool force)
         return;
     }
 
+    _empty_service_or_env = zend_string_equals_literal(get_DD_ENV(), "") ||
+                            zend_string_equals_literal(get_DD_SERVICE(), "");
+
     _set_cur_span(dd_trace_get_active_root_span());
     if (!_cur_req_span) {
         mlog(dd_log_debug, "No root span available on request init");
@@ -128,8 +132,13 @@ static zend_array *nullable _do_request_begin_user_req(zval *nullable rbe_zv)
     return _do_request_begin(rbe_zv, true);
 }
 
-static bool _rem_cfg_path_changed()
+static bool _rem_cfg_path_changed(bool ignore_empty /* called from rinit */)
 {
+    if (ignore_empty && _empty_service_or_env &&
+        _last_rem_cfg_path[0] != '\0') {
+        return false;
+    }
+
     const char *cur_path = dd_trace_remote_config_get_path();
     if (!cur_path) {
         cur_path = "";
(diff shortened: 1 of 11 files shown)


appsec/src/helper/subscriber/waf.cpp
```diff
@@ -78,7 +78,7 @@ DDWAF_LOG_LEVEL spdlog_level_to_ddwaf(spdlog::level::level_enum level)
     case spdlog::level::trace:
         return DDWAF_LOG_TRACE;
     case spdlog::level::debug:
-        return DDWAF_LOG_DEBUG;
+    // libddwaf is too verbose at debug level
```
Anilm3, line 81: Not sure about this, the trace level with RC has just too much crap, which makes it difficult to analyse anything WAF-specific...

appsec/src/extension/request_lifecycle.c
```diff
@@ -253,6 +272,28 @@ void dd_req_lifecycle_rshutdown(bool ignore_verdict, bool force)
...
         // _rest_globals already called
     }
+
+    // if we don't have service/env, our only chance to update the remote config
+    // path is rshutdown because ddtrace's rinit is called before ours and it
+    // resets the path
+    if (_has_empty_service_or_env() && _rem_cfg_path_changed(false)) {
+        mlog_g(dd_log_debug, "No DD_SERVICE/DD_ENV; trying to sync remote "
+                             "config path on rshutdown");
+        dd_conn *conn =
```
Anilm3, line 282: If the connection isn't already available, you're effectively initialising a (helper) client which will be stuck waiting for a `client_init`, isn't that the case?

cataphract: right, `dd_helper_mgr_cur_conn` makes more sense here

appsec/src/helper/subscriber/waf.cpp
cataphract: the previous behavior resulted in 90% of the messages being from libddwaf, esp during initialization, with pearls like this:

```
[2024-10-10 14:01:39.863][debug][181] Parsed rule crs-932-180
[2024-10-10 14:01:39.863][debug][181] Found address server.request.headers.no_cookies
[2024-10-10 14:01:39.863][debug][181] Found address server.request.headers.no_cookies
[2024-10-10 14:01:39.863][debug][181] Found address server.request.headers.no_cookies
[2024-10-10 14:01:39.863][debug][181] Found address server.request.headers.no_cookies
```

makes it difficult to read the logs

appsec/src/extension/request_lifecycle.c
```diff
@@ -128,8 +128,19 @@ static zend_array *nullable _do_request_begin_user_req(zval *nullable rbe_zv)
     return _do_request_begin(rbe_zv, true);
 }
 
-static bool _rem_cfg_path_changed()
+static bool _has_empty_service_or_env()
 {
+    return zend_string_equals_literal(get_DD_ENV(), "") ||
+           zend_string_equals_literal(get_DD_SERVICE(), "");
+}
```
bwoebi, line 135: You should cache that at rinit, it's possible to change these via `ini_set()` too at runtime... Which will then not reflect the truth for you in rshtudown.

appsec/src/helper/subscriber/waf.cpp
cataphract: I've looked at the code of libddwaf and the only trace messages are about matches, which make them lower-volume, and actually look more important than most debug messages.

appsec/src/extension/request_lifecycle.c
cataphract: Can't I use `get_global_DD_ENV()` here for the same effect?

cataphract: I've moved it to startup with get_global

appsec/src/helper/subscriber/waf.cpp
Anilm3: We had to demote match logs a while ago because of PII concerns.

appsec/src/extension/request_lifecycle.c
bwoebi: No, get_global is definitely inferior. It will also bypass all perdir changes. Like vhost specific config.

bwoebi: Just cache the value in a thread-local variable after rinit.

bwoebi (approved):
Looks good to me now.

Anilm3 (approved):
Lgtm

PR #2896, opened 2024-10-15
Implement UST Consistency

khanayan123:
### Description

Make changes such that iff a span has service name set by DD_SERVICE, it also gets the version specified in DD_VERSION. This is done to better align with the UST spec https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/?tab=kubernetes

### DIFF
--- a/ext/ddtrace.c
+++ b/ext/ddtrace.c
@@ -1101,10 +1101,14 @@ static zval *ddtrace_root_span_data_write(zend_object *object, zend_string *memb
     } else if (zend_string_equals_literal(prop_name, "service")) {
         if (ddtrace_span_is_entrypoint_root(&span->span) && !zend_is_identical(&span->property_service, value)) {
             root_span_data_changed = true;
-            // As per unified service tagging spec if a span is created with a service name different from the global
-            // service name it will not inherit the global version value
-            ZVAL_NULL(&span->property_version);
-
+        }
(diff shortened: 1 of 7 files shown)


ext/ddtrace.c
```diff
@@ -1096,6 +1096,10 @@ static zval *ddtrace_root_span_data_write(zend_object *object, zend_string *memb
     } else if (zend_string_equals_literal(prop_name, "service")) {
         if (ddtrace_span_is_entrypoint_root(&span->span) && !zend_is_identical(&span->property_service, value)) {
             root_span_data_changed = true;
+            // As per unified service tagging spec if a span is created with a service name different from the global
```
bwoebi, line 1099: Nearly, it needs to be changed when the span is _not_ the root span.
And misses `zval_ptr_dtor(&span->property_version);` to first free to old value.

ext/ddtrace.c
```diff
@@ -1096,6 +1096,10 @@ static zval *ddtrace_root_span_data_write(zend_object *object, zend_string *memb
...
+            // service name it will not inherit the global version value
+            ZVAL_NULL(&span->property_version);
```
bwoebi, line 1101: This cannot be null, use `ZVAL_EMPTY_STRING` instead.

bwoebi (approved):

PR #2903, opened 2024-10-18
[DRAFT] ASM Standalone

estringana:
### Description

Work in progress

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/auto_flush.c
+++ b/ext/auto_flush.c
@@ -13,6 +13,24 @@
 #include "ddshared.h"
 #include <main/SAPI.h>
 
+static bool trace_contains_appsec_event(zval *trace) {
+    if (!trace || Z_TYPE_P(trace) != IS_ARRAY) {
+        return false;
+    }
+
+    zval *root_span = zend_hash_index_find(Z_ARR_P(trace), 0);
+    if (!root_span || Z_TYPE_P(root_span) != IS_ARRAY) {
+        return false;
+    }
+
+    zval *meta = zend_hash_str_find(Z_ARR_P(root_span), ZEND_STRL("meta"));
+    if (!meta || Z_TYPE_P(meta) != IS_ARRAY) {
+        return false;
+    }
+
+    return zend_hash_str_exists(Z_ARR_P(meta), ZEND_STRL("appsec.event"));
+}
+
 ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles) {
     bool success = true;
 
@@ -42,6 +60,11 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
     array_init(&traces);
     zend_hash_index_add(Z_ARR(traces), 0, &trace);
 
+    if (get_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED() && !trace_contains_appsec_event(&trace)) {
+        LOG(DEBUG, "Trace discarded because it has no appsec events and appsec standalone is on");
+        return SUCCESS;
+    }
+
     size_t limit = get_global_DD_TRACE_AGENT_MAX_PAYLOAD_SIZE();
     if (get_global_DD_TRACE_SIDECAR_TRACE_SENDER()) {
         if (ddtrace_sidecar) {
@@ -64,7 +87,7 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
                                 .tracer_version = DDOG_CHARSLICE_C_BARE(PHP_DDTRACE_VERSION),
                                 .lang_version = dd_zend_string_to_CharSlice(ddtrace_php_version),
                                 .client_computed_top_level = false,
-                                .client_computed_stats = false,
+                                .client_computed_stats = get_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED(),
                         };
                         size_t size_hint = written;
                         zend_long n_requests = get_global_DD_TRACE_AGENT_FLUSH_AFTER_N_REQUESTS();
--- a/ext/configuration.h
+++ b/ext/configuration.h
@@ -235,6 +235,7 @@ enum ddtrace_sampling_rules_format {
     CONFIG(BOOL, DD_REMOTE_CONFIG_ENABLED, "true", .ini_change = zai_config_system_ini_change)          \
     CONFIG(BOOL, DD_DYNAMIC_INSTRUMENTATION_ENABLED, "false", .ini_change = zai_config_system_ini_change)      \
     CONFIG(SET, DD_DYNAMIC_INSTRUMENTATION_REDACTED_IDENTIFIERS, "", .ini_change = zai_config_system_ini_change) \
+    CONFIG(BOOL, DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED, "false")                                           \
     CONFIG(SET, DD_DYNAMIC_INSTRUMENTATION_REDACTED_TYPES, "", .ini_change = zai_config_system_ini_change)     \
     DD_INTEGRATIONS
 
--- a/ext/serializer.c
+++ b/ext/serializer.c
@@ -1697,8 +1697,12 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
     array_init(&metrics_zv);
     zend_string *str_key;
     zval *val;
+    bool is_standalone_appsec_span = false;
     ZEND_HASH_FOREACH_STR_KEY_VAL_IND(metrics, str_key, val) {
         if (str_key) {
+            if (!is_standalone_appsec_span && get_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED() && strcmp("_dd.appsec.enabled", ZSTR_VAL(str_key)) == 0) {
+                is_standalone_appsec_span = true;
+            }
             dd_serialize_array_metrics_recursively(Z_ARRVAL(metrics_zv), str_key, val);
         }
     } ZEND_HASH_FOREACH_END();
@@ -1706,6 +1710,9 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
     if (is_root_span) {
         if (Z_TYPE_P(&span->root->property_sampling_priority) != IS_UNDEF) {
             add_assoc_double(&metrics_zv, "_sampling_priority_v1", zval_get_long(&span->root->property_sampling_priority));
+        }
+        if(is_standalone_appsec_span) {
+            add_assoc_bool(&metrics_zv, "_dd.apm.enabled", false);
         }
     }
 


ext/priority_sampling/priority_sampling.c
```diff
@@ -123,36 +123,38 @@ static ddtrace_rule_result dd_match_rules(ddtrace_span_data *span, bool eval_roo
...
     zval *rule;
-    ZEND_HASH_FOREACH_VAL(get_DD_TRACE_SAMPLING_RULES(), rule) {
-        if (++index >= skip_at) {
-            break;
-        }
-
-        if (Z_TYPE_P(rule) != IS_ARRAY) {
-            continue;
-        }
+    if (!get_global_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED()) {  // APPSEC_STANDALONE enabled, override sampling rules to be empty
```
estringana, line 126: This just wrap the code within `if (!get_global_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED()) {`

ext/auto_flush.c
```diff
@@ -11,8 +11,30 @@
...
 #include <main/SAPI.h>
 
+ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
+
+static bool trace_contains_appsec_event(zval *trace) {
+    if (!trace || Z_TYPE_P(trace) != IS_ARRAY) {
+        return false;
+    }
+
+    zval *root_span = zend_hash_index_find(Z_ARR_P(trace), 0);
```
bwoebi, line 25: I think by the way ddtrace is currently designed, the root span will (almost?) always be the first span.
But that's more by coincidence that the root span just happens to be the last span closed - and spans are serialized in reverse order.
This may or may not change in future via normal refactors. (?)
We should at least note somewhere that we rely on that being the case now.

estringana: Ok. How would you go approach this then? I'm ok to implement it properly if it does not require any major refactoring

bwoebi: I'm not suggesting to actually change it, but primarily to document that you rely on this now.

ext/standalone_limiter.c
```diff
@@ -0,0 +1,71 @@
...
+    ZEND_ASSERT(dd_limiter);
+
+    uint64_t timeval = zend_hrtime();
+    uint64_t old_time = atomic_load(&dd_limiter->window.last_hit);
+
+    if (timeval - old_time < 60000000000) {
+        return false;
+    }
+
+    atomic_store(&dd_limiter->window.last_hit, timeval);
```
bwoebi, line 61: I suppose it doesn't matter if there's a possible race condition where two requests hit this at the same time. The worst which will happen is that rarely two traces are sent.
(Otherwise one could do a cmpxchg here.)

appsec/tests/extension/inc/request_replayer.inc
```diff
@@ -0,0 +1,151 @@
+<?php
```
cataphract, line 1: The mock_helper already has an http server running on 1826 that can echo back traces: https://github.com/DataDog/dd-trace-php/blob/master/appsec/tests/mock_helper/mock_dd_agent.cc#L55

ext/tracer_tag_propagation/tracer_tag_propagation.c
```diff
@@ -79,15 +79,46 @@ void ddtrace_add_tracer_tags_from_array(zend_array *array, zend_array *root_meta
...
     }
 
+    return root_meta;
+}
+
+zval *ddtrace_propagated_tags_get_tag(const char *tag) {
+    zend_array *propagated = ddtrace_get_propagated();
+    zend_array *root_meta = ddtrace_get_root_meta();
+
+    if (!tag) {
```
iamluc, line 106: nit: could be done before the calls to ddtrace_get_propagated / ddtrace_get_root_meta to save a few CPU instructions

iamluc (commented):
It seems [the CI spotted a crash](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/18440/workflows/18e85e7e-c5a9-4d2a-8dd7-56da973c5943/jobs/5677199)
```
tests/ext/integrations/curl/distributed_tracing_curl_propagate_asm_event.out:Segmentation fault (core dumped)
```

ext/standalone_limiter.c
```diff
@@ -0,0 +1,71 @@
...
+
+bool ddtrace_standalone_limiter_allow() {
+    ZEND_ASSERT(dd_limiter);
+
+    uint64_t timeval = zend_hrtime();
+    uint64_t old_time = atomic_load(&dd_limiter->window.last_hit);
+
+    if (timeval - old_time < 60000000000) {
+        return false;
+    }
```
cataphract, line 59: I'm not sure about this logic. This all but guarantees that more than one minute will pass between two events, which doesn't fit the requirement of sending at least one event per minute.

For instance, imagine a service that's hit every 45 seconds. Only one trace will be sent forward for every two, which means 90 seconds will pass between every trace.

tests/ext/integrations/curl/distributed_tracing_curl_propagate_asm_event.phpt
```diff
@@ -0,0 +1,40 @@
+--TEST--
+Distributed tracing header tags propagate asm events with curl_exec()
+--SKIPIF--
+<?php if (!extension_loaded('curl')) die('skip: curl extension required'); ?>
+<?php if (!getenv('HTTPBIN_HOSTNAME')) die('skip: HTTPBIN_HOSTNAME env var required'); ?>
```
cataphract, line 5: There's no reason to make the this test depend on external services being launched before... Just launch the webserver in the beginning of the test and kill it at the end, like in a destructor.

cataphract (changes requested):

ext/standalone_limiter.c
estringana: I sorted this. Can you have a look plz?

ext/ddtrace.stub.php
```diff
@@ -830,6 +830,10 @@ function set_blocking_function(\DDTrace\RootSpanData $span, callable $blockingFu
      * E_DEPRECATED, E_USER_DEPRECATED
      */
     function trigger_error(string $message, int $errorType): void {}
+    /**
+     * Simulate appsec has emit an asm event
```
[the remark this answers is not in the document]
estringana: Changed

ext/standalone_limiter.c
cataphract: Yeah, I think it's good now. Although like I said the logic should prob be moved to helper (maybe service class and returning a bool on req init) or sidecar.

ext/ddappsec.c
```diff
@@ -0,0 +1,31 @@
+#include "ddappsec.h"
```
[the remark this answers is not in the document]
estringana: Where would you suggest to move this functions to o how would you call this file?

appsec/src/extension/tags.c
```diff
@@ -344,6 +350,27 @@ void dd_tags_add_tags(
...
+        }
+    } else {
+        mlog(dd_log_info,
+            "Skipping adding " DD_TAG_P_APPSEC " as it was already added");
+    }
+
+    zval _1_zval;
+    ZVAL_STR(&_1_zval, _1_zstr);
+    dd_trace_span_add_propagated_tags(_dd_tag_p_appsec_zstr, &_1_zval);
+
```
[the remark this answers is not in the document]
estringana: Appsec add tags on shutdown. However, during the request there could be curl request to another services which will require header propagation. That's why we emit the event as early as detected

ext/standalone_limiter.c
```diff
@@ -0,0 +1,71 @@
...
+} ddtrace_standalone_limiter;
+
+static ddog_MappedMem_ShmHandle *dd_limiter_mapped_shm;
+static ddtrace_standalone_limiter *dd_limiter;
+
+void ddtrace_standalone_limiter_create() {
+    uint32_t limit = (uint32_t) 1;
+
+    ddog_ShmHandle *shm;
+    if (!ddtrace_ffi_try("Failed allocating shared memory", ddog_alloc_anon_shm_handle(limit, &shm))) {
```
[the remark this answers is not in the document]
estringana: This is cloning what the ddtrace limiter does

appsec/src/extension/tags.c
```diff
@@ -344,6 +350,27 @@ void dd_tags_add_tags(
...
+            return;
+        }
+    } else {
+        mlog(dd_log_info,
+            "Skipping adding " DD_TAG_P_APPSEC " as it was already added");
+    }
+
+    zval _1_zval;
+    ZVAL_STR(&_1_zval, _1_zstr);
+    dd_trace_span_add_propagated_tags(_dd_tag_p_appsec_zstr, &_1_zval);
```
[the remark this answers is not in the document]
estringana: Nice one. Done

appsec/tests/extension/inc/distributed_tracing.inc
```diff
@@ -0,0 +1,26 @@
...
+{
+    $data = json_decode($response, true);
+    if (!$data || !isset($data['headers'])) {
+        echo 'Invalid response:' . PHP_EOL;
+        var_dump($response);
+        return [];
+    }
+    $headers = [];
+    foreach ($data['headers'] as $key => $value) {
+        $headers[strtolower($key)] = mb_convert_encoding($value, 'ISO-8859-1', 'UTF-8');
```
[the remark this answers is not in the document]
estringana: This file has been removed now

tests/ext/integrations/curl/distributed_tracing_curl_propagate_asm_event.phpt
```diff
@@ -0,0 +1,40 @@
...
+
+DDTrace\set_priority_sampling(DD_TRACE_PRIORITY_SAMPLING_USER_KEEP);
+
+
+DDTrace\Testing\emit_asm_event();
+dt_dump_headers_from_httpbin(query_headers(), ['x-datadog-tags']);
+
+?>
+--EXPECT--
+x-datadog-tags: _dd.p.appsec=1,_dd.p.dm=-4
```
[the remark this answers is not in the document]
estringana: Could you expand why?

ext/auto_flush.c
```diff
@@ -38,6 +60,36 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
...
+                LOG(ERROR, "Metrics not found. Dropping trace");
+                return SUCCESS;
+            }
+
+            zval *sampling_priority = zend_hash_str_find(Z_ARR_P(metrics), ZEND_STRL("_sampling_priority_v1"));
+            if (!sampling_priority || (Z_TYPE_P(sampling_priority) != IS_DOUBLE && Z_TYPE_P(sampling_priority) != IS_LONG)) {
+                LOG(ERROR, "Invalid sampling priority. Dropping trace");
+                return SUCCESS;
+            }
+            ZVAL_LONG(sampling_priority, PRIORITY_SAMPLING_AUTO_REJECT);
```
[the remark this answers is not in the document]
estringana: No, we need to set them to auto reject

tests/ext/integrations/curl/distributed_tracing_curl_propagate_asm_event.phpt
estringana: I chose to follow the same approach other tests on the tracer

cataphract: ah sorry I thought this was on appsec

tests/ext/integrations/curl/distributed_tracing_curl_propagate_asm_event.phpt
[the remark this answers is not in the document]
cataphract: I didn't look too close, but the RFC says:

```
In case of one or several ASM events (IAST, WAF, custom events, etc.), the sampling priority of the local traces is set to “force keep” (_sampling_priority_v1:2 (MANUAL_KEEP)). The  propagated decision maker tag _dd.p.dm is also set to -5 (-5 is the ID for ASM). Note that _dd.p.dm must be only set once and never changed, even in downstream services.
```

In this test case, an ASM event is emitted on 1st request, so I guess that _dd.p.dm=-5 should be propagated to the next service, no?

ext/ddappsec.c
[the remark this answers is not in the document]
cataphract: idk "asm_event.c" for instance

ext/auto_flush.c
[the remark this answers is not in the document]
cataphract: How do you reconcile that statement with the text I quoted? Because in addition to what I quoted, the spec also says:

> Other sources of force-keep sampling priorities, such as errors, crucial to ASM’s backend side monitoring, are still allowed in standalone ASM mode.

So it seems to me that force_keep traces should not be rejected. If my understanding is wrong, please explain.

ext/tracer_tag_propagation/tracer_tag_propagation.c
estringana: Done

estringana:
@iamluc I did fix the segmentation fault and also address your suggestion. Can you review again plz? 🙏

ext/standalone_limiter.c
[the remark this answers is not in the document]
cataphract: ? Sorry, I don't understand your comment

cataphract (commented):
Replied on some the threads.

ext/auto_flush.c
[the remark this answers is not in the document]
estringana: This is the scenario when apm should behave as disable. There is no asm events, no appsec upstream services and it's not used as a heartbeat. In that case we need to let the trace drop. Does it makes sense?

cataphract (approved):

iamluc (approved):

PR #2909, opened 2024-10-21
Fixing DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP Empty Value Behavior

link04:
### Description

Updating what should happen if the value for `DD_TRACE_OBFUSCATION_QUERY_STRING_REGEXP` is set to an empty string since as of today it was redacting everything in system tests.

https://datadoghq.atlassian.net/browse/APMAPI-499

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
I think empty string handling is missing in Normalizer.php. I.e. it should also always match there with an empty string.

bwoebi:
In fact I'm just seeing https://github.com/DataDog/dd-trace-php/pull/2909/commits/5f1f193bd1c5672272cff6a96952c631ac1adb57, is there a specific reason you reverted that?

bwoebi (approved):
Looks right. Merging :-)

zend_abstract_interface/uri_normalization/uri_normalization.c
```diff
@@ -157,7 +157,7 @@ zend_string *zai_filter_query_string(zai_str queryString, zend_array *whitelist,
         zend_hash_get_current_key(whitelist, &str, &numkey);
         if (zend_string_equals_literal(str, "*")) {
             zend_string *qs = zend_string_init(queryString.ptr, queryString.len, 0);
-            if (pattern) {
+            if (ZSTR_LEN(pattern)) {
```
bwoebi, line 160:
```suggestion
            if (pattern && ZSTR_LEN(pattern)) {
```

PR #2922, opened 2024-11-05
Support Tag Keys for DD_TRACE_HEADER_TAGS

link04:
### Description
Adding Support for allowing customers to pass `:` separated values in the list of header tags which they would like to use the key instead of the current header name only.

### Reviewer checklist
- [ ] Test coverage seems ok.


zend_abstract_interface/config/config_decode.c
```diff
@@ -105,46 +105,78 @@ static bool zai_config_decode_int(zai_str value, zval *decoded_value) {
...
+                // Handle case without a colon (standalone key)
+                if (!has_colon && key_end) {
+                    size_t key_len = key_end - key_start + 1;
+                    zend_string *key = zend_string_init(key_start, key_len, persistent);
+                    if (lowercase) {
+                        zend_str_tolower(ZSTR_VAL(key), ZSTR_LEN(key));
+                    }
+
+                    zval val;
+                    ZVAL_NEW_STR(&val, zend_string_copy(key));
```
bwoebi, line 176: wouldn't ZVAL_EMPTY_STR() make more sense here? As in: no value = empty value? Instead of no value = the key again?

bwoebi:
Implementation looks good so far, but peer service seems to want that you ignore these cases without the colon.

zend_abstract_interface/config/config_decode.c
link04: Yeah that makes sense I wasn't sure if use the key as value to make use of the value in the serializer or empty, will go for the later as it does make more sense and seems to be better performance wise.

zend_abstract_interface/config/config_decode.c
```diff
@@ -164,16 +164,16 @@ static bool zai_config_decode_map(zai_str value, zval *decoded_value, bool persi
                     data++;
                 }
 
-                // Handle case without a colon (standalone key)
-                if (!has_colon && key_end) {
+                // Handle case without a colon (standalone key) only when lowercase is true which is current use case
+                if (lowercase && !has_colon && key_end) {
```
bwoebi, line 168: Let's add a second boolean parameter, `allow_key_only` or such, and set it to true too for this case. Otherwise this will probably be confusing.

link04: Aaaaahh I not me adding a boolean `set` but then changing my mind to this instead, got it clarity first always.

zend_abstract_interface/config/config_decode.c
```diff
@@ -127,59 +127,78 @@ static bool zai_config_decode_map(zai_str value, zval *decoded_value, bool persi
...
                     if (*data != ' ' && *data != '\t' && *data != '\n') {
                         key_end = data;
                     }
                     data++;
                 }
 
-                // Handle case without a colon (standalone key) only when lowercase is true which is current use case
-                if (map_keyless && !has_colon && key_end) {
+                // Handle standalone keys (without a colon) if map_keyless is enabled
+                if (map_keyless && !has_colon && key_end && key_start) {
```
bwoebi, line 182: note: key_start is never NULL here (nor above)

zend_abstract_interface/config/config_decode.c
```diff
@@ -127,59 +127,78 @@ static bool zai_config_decode_map(zai_str value, zval *decoded_value, bool persi
...
+                                data++;
                             }
-                            data++;
                         }
 
-                        size_t key_len = key_end - key_start + 1;
-                        size_t value_len = value_end - value_start + 1;
+                        if (key_end && key_start) {
+                            size_t key_len = key_end - key_start + 1;
+                            size_t value_len = (value_end && value_start) ? (value_end - value_start + 1) : 0;
```
bwoebi, line 150: move this line into the else part below and instead check `if (value_end)`, that's simpler

zend_abstract_interface/config/config_decode.c
```diff
@@ -105,49 +105,100 @@ static bool zai_config_decode_int(zai_str value, zval *decoded_value) {
...
+                        has_colon = true;
+                        data++;
+                        
+                        while (*data == ' ' || *data == '\t' || *data == '\n') data++;
 
-                        value_start = value_end = data;
-                        if (!*data || *data == ',') {
-                            --value_end;  // empty string instead of single char
+                        if (*data == ',' || *data == '\0') {
+                            value_start = value_end = NULL;
```
bwoebi, line 135:
```suggestion
                            value_end = NULL;
```
No need to null both.

zend_abstract_interface/config/config_decode.c
```diff
@@ -105,49 +105,100 @@ static bool zai_config_decode_int(zai_str value, zval *decoded_value) {
...
                         } else {
-                            while (*++data && *data != ',') {
+                            value_start = data;
+                            value_end = value_start;
+
+                            while (*data && *data != ',') {
```
bwoebi, line 140: you may convert this to do-while

zend_abstract_interface/config/config_decode.c
```diff
@@ -105,49 +105,100 @@ static bool zai_config_decode_int(zai_str value, zval *decoded_value) {
...
 
-                        value_start = value_end = data;
-                        if (!*data || *data == ',') {
-                            --value_end;  // empty string instead of single char
+                        if (*data == ',' || *data == '\0') {
+                            value_start = value_end = NULL;
                         } else {
-                            while (*++data && *data != ',') {
+                            value_start = data;
+                            value_end = value_start;
```
bwoebi, line 138:
```suggestion
                            value_start = value_end = data;
```
I think this communicates more clearly "we assign both to data", than start to data and end to start.

bwoebi:
Just styling notes, looks good now :-)

zend_abstract_interface/config/config_decode.c
link04: Thank you! Lol learned about this in C. Applied in: https://github.com/DataDog/dd-trace-php/pull/2922/commits/0de558f64ef4a156ba377c3ea35f7a69017bad86

zend_abstract_interface/config/config_decode.c
```diff
@@ -131,31 +131,29 @@ static bool zai_config_decode_map(zai_str value, zval *decoded_value, bool persi
...
 
                             zend_string *key = zend_string_init(key_start, key_len, persistent);
                             if (lowercase) {
                                 zend_str_tolower(ZSTR_VAL(key), ZSTR_LEN(key));
                             }
 
                             zval val;
-                            if (value_len > 0) {
+                            if (value_end) {
+                                size_t value_len = value_end ? (value_end - value_start + 1) : 0;
```
bwoebi, line 156: value_end is always true here :-)

link04: Ooops, my bad, updated. https://github.com/DataDog/dd-trace-php/pull/2922/commits/2f104b2484299e5d02cf979a9a3c3ae39bdf2ee6

bwoebi (approved):
LGTM, thanks Maximo :-)

PR #2927, opened 2024-11-05
#2926: Do not overwrite service in PSR18 integration

### LINKED ISSUE #2926: PSR18 integration service name
Please do not overwrite the service name in this integration. It should be consistent with other integrations like redis, pdo etc., respecting several env vars like `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED`.

With this kind of hardcoded service, we have no control over this component and it just appears as a separate service in the map even though its just a component.

Solution would be to simply remove this [line](https://github.com/DataDog/dd-trace-php/blob/master/src/DDTrace/Integrations/Psr18/Psr18Integration.php#L30).

TheLevti:
As described in the issue https://github.com/DataDog/dd-trace-php/issues/2926, for consistency and correctness, do not hardcode the service for this integration otherwise there is no way to control it and it forces that everyone has a service on their service map that is actually just a component.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/Psr18/Psr18Integration.php
+++ b/src/DDTrace/Integrations/Psr18/Psr18Integration.php
@@ -27,7 +27,6 @@ class Psr18Integration extends Integration
             function (SpanData $span, $args, $retval) use ($integration) {
                 $span->resource = 'sendRequest';
                 $span->name = 'Psr\Http\Client\ClientInterface.sendRequest';
-                $span->service = 'psr18';
                 $span->type = Type::HTTP_CLIENT;
                 $span->meta[Tag::SPAN_KIND] = 'client';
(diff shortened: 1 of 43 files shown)


PROFeNoM (commented):
Hey @TheLevti :wave: Functionally, this makes total sense to me. I'll go ahead and fix the tests if you're ok with that, so that we can merge the PR :shipit:

TheLevti:
> Hey @TheLevti 👋 Functionally, this makes total sense to me. I'll go ahead and fix the tests if you're ok with that, so that we can merge the PR :shipit:

Thank you very much, yes please.

PROFeNoM (approved):
Thanks for your recommendation! 😃

PR #2934, opened 2024-11-07
feat(profiling) add thread names

realFlowControl:
### Description

This PR will add a `thread name` label to collected samples. In most cases a dedicated thread name is either not set or even if it is not helpful. For NTS PHP we do just use the SAPI name as the thread name, this seems helpful. For ZTS we currently only collect a thread name for FrankenPHP as it sets meaningful names.
Additional notes:
- on Darwin a thread name might be empty if not set
- on Linux a thread name is the process name if not explicitly set (which is the case for most)

In the future we should try and detect Apache and parallel threads and name them accordingly.

PROF-10863

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/profiling/thread_utils.rs
```diff
@@ -69,3 +75,67 @@ pub fn join_timeout(handle: JoinHandle<()>, timeout: Duration, impact: &str) {
...
+                            name.as_mut_ptr() as *mut c_char,
+                            name.len(),
+                        )
+                    };
+
+                    if result == 0 {
+                        // If successful, convert the result to a Rust String
+                        let cstr =
+                            unsafe { std::ffi::CStr::from_ptr(name.as_ptr() as *const c_char) };
+                        let str_slice: &str = cstr.to_str().unwrap();
```
morrisonlevi, line 109: This will crash if it's not utf-8. I think we should log instead and use the SAPI name only if this happens. What do you think?

realFlowControl: I changed the `unwrap()` to a `unwrap_or_default()` (which makes it empty). So in that case we'd just have the SAPI name and nothing else which is okay I'd say

morrisonlevi (approved):

PR #2935, opened 2024-11-07
Disable helper when appsec is fully disabled. 

cataphract:
### Description

Also make datadog.appsec.enabled a system config. Note that the default value for datadog.appsec.enabled in ext/configuration.h (true) is different from that in datadog.appsec.disabled (false). This is intentional. appsec chooses between three states: explicitly enabled, explicitly disabled and controlled by remote config. Though a questionable decision, the config setting is still a boolean, and the third state is detected by a hack that determines whether the value was explicitly set. Because on the ddtrace side we want to suppress the helper only if explicitly disabled, we can set the default value to true, and disable the helper only the final is false (if the value is not the default true, then it was explicitly set to false).

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/extension/helper_process.c
```diff
@@ -194,7 +195,12 @@ __attribute__((visibility("default"))) void dd_appsec_maybe_enable_helper(
     ddog_CharSlice log_level =
         to_char_slice(get_global_DD_APPSEC_HELPER_LOG_LEVEL());
 
-    enable_appsec(helper_path, socket_path, lock_path, log_path, log_level);
+    ddog_MaybeError res =
+        enable_appsec(helper_path, socket_path, lock_path, log_path, log_level);
+    if (res.tag == DDOG_OPTION_ERROR_SOME_ERROR) {
```
bwoebi, line 200: you have to free this `ddog_MaybeError_drop(res)` is the API the rust code exposes.

bwoebi:
> Note that the default value for datadog.appsec.enabled in ext/configuration.h (true) is different from that in datadog.appsec.disabled (false). This is intentional.

As long as this works properly ... Both extensions will write to the same ini_entry in PHP... I think it would be better to keep them in sync and do the name_index < 0 check in the tracer too.

bwoebi:
I'm confused, dd_appsec_maybe_enable_helper still will sideload the appsec helper - when the tracer is launched with the sidecar, there's no check in dd_appsec_maybe_enable_helper for appsec disabled?

appsec/src/extension/helper_process.c
cataphract: I've made it return unit instead.

bwoebi (approved):
Looks great to me, thanks :-)

Anilm3 (approved):
lgtm

PR #2936, opened 2024-11-07
Fix macos build where ld will break with -Wl,--undefined-version

bwoebi:
It's clang, but a different clang.

It's better to do actual checks for the functionailty than checking for a specific compiler.


realFlowControl (approved):
Thanks, this is way nicer than what I did 🙇

PR #2941, opened 2024-11-11
support multiple `PHP_INI_SCAN_DIR`'s

realFlowControl:
### Description

In case a user sets the `PHP_INI_SCAN_DIR` environment variable to a special value with a leading or trailing `:` or have multiple directories separated by the default path separator, we need to handle this accordingly.

In case they use the trailing or leading `:` to just add to the default directory, there is no way for us to find that original value anymore, so we "just" always use the first explicitly defined path.

This is documented upstream at https://www.php.net/manual/en/configuration.file.php#configuration.file.scan

### Reviewer checklist
- [ ] Test coverage seems ok.


datadog-setup.php
```diff
@@ -827,6 +827,13 @@ function find_all_ini_files(array $phpProperties)
 function find_main_ini_files(array $phpProperties)
 {
     if (isset($phpProperties[INI_SCANDIR])) {
+
+        $pos = strpos($phpProperties[INI_SCANDIR], \PATH_SEPARATOR);
+        if ($pos !== false) {
+            // https://www.php.net/manual/en/configuration.file.php#configuration.file.scandir
+            $phpProperties[INI_SCANDIR] = current(array_filter(explode(':', $phpProperties[INI_SCANDIR])));
```
bwoebi, line 834:
```suggestion
            $phpProperties[INI_SCANDIR] = current(array_filter(explode(\PATH_SPARATOR, $phpProperties[INI_SCANDIR])));
```

bwoebi (approved):
Looks good, thanks.

PR #2943, opened 2024-11-12
feat(profiling) add `stream_select()`-type functions to timeline

realFlowControl:
### Description

This will wrap functions that are used to wait on blocking I/O, like `stream_select()` and others. It also moves the `frankenphp_handle_request()` to the `idle` visualisation because technically it is waiting on I/O, but it is used to wait for the next request, just like `RSHUTDOWN` -> `RINIT` so it should share the same vis.

In async PHP (using ReactPHP, AMPHP or others), spending time in a `stream_select()`-type function call means the event-loop is idle.

PPROF-10884

AMPHP event loop:
image

`ext-parallel`'s `\parallel\Events::poll()` method:
image

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/timeline.rs
+++ b/profiling/src/timeline.rs
@@ -131,6 +131,7 @@ create_sleeping_fn!(ddog_php_prof_uv_run, UV_RUN_HANDLER, State::Select);
 create_sleeping_fn!(ddog_php_prof_event_base_loop, EVENT_BASE_LOOP_HANDLER, State::Select);
 create_sleeping_fn!(ddog_php_prof_eventbase_loop, EVENTBASE_LOOP_HANDLER, State::Select);
 create_sleeping_fn!(ddog_php_prof_ev_loop_run, EV_LOOP_RUN_HANDLER, State::Select);
+create_sleeping_fn!(ddog_php_prof_parallel_events_poll, PARALLEL_EVENTS_POLL_HANDLER, State::Select);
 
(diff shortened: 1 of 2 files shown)


morrisonlevi (commented):
First glance looks good, will look a bit more later!

profiling/src/timeline.rs
```diff
@@ -241,12 +218,73 @@ pub unsafe fn timeline_startup() {
...
+            ptr::addr_of_mut!(STREAM_SELECT_HANDLER),
+            Some(ddog_php_prof_stream_select),
+        ),
+        zend::datadog_php_zif_handler::new(
+            CStr::from_bytes_with_nul_unchecked(b"socket_select\0"),
+            ptr::addr_of_mut!(SOCKET_SELECT_HANDLER),
+            Some(ddog_php_prof_socket_select),
+        ),
+        zend::datadog_php_zif_handler::new(
+            CStr::from_bytes_with_nul_unchecked(b"socket_accept\0"),
```
bwoebi, line 232: Not sure if socket_accept fits here, but if it does, stream_socket_accept should also fit here.

realFlowControl: I'll remove it from here and add it to my notes. Even though a call to `socket_accept()` might block (depending if the socket is blocking) it does not fit in this list of `select()`-type functions

morrisonlevi (approved):
I don't have any other feedback. Looks good to me.

PR #2951, opened 2024-11-14
Create many tiny artifacts for faster download

bwoebi:
And not escaping the constraints many systems have on /tmp size. If someone installs multiple php versions at once, we still download the full package.

The full package also remains available for offline installations of multiple targets for now.


datadog-setup.php
```diff
@@ -1444,12 +1492,18 @@ function download($url, $destination)
         curl_setopt($ch, CURLOPT_NOPROGRESS, false);
         $progress_counter = 0;
         $return = curl_exec($ch);
+
+        $httpCode = curl_getinfo($ch, CURLINFO_HTTP_CODE);
+        if ($httpCode == 404) {
```
iamluc, line 1497: This looks weird to me.
IIUC, at the first download attempt (for a specific version), if `curl` is installed, we are going to try only with curl.
Then at the second attempt (full package), this could create an infinite loop if github (or a proxy or whatever) returns a 404.

bwoebi: You're right, I messed that up.

bwoebi: Wrapped in $retry, looks good now?

PR #2956, opened 2024-11-15
fix: Symfony Messenger propagation to non-instrumented clients

PROFeNoM:
### Description

Related to the following [GH Comment](https://github.com/DataDog/dd-trace-php/pull/2749#issuecomment-2467409884).

There were two issues.

A first issue with `Serializer::decode`.
```
Could not denormalize object of type "DDTrace\Integrations\SymfonyMessenger\DDTraceStamp[]", no supporting normalizer found
```
It could have been addressed by the following [commit](https://github.com/DataDog/dd-trace-php/commit/dfa9ed30b4b098f0dc8c5cd568bcfc2a562e7846) of Bob's.

A second, more annoying issue, with `PhpSerializer::safelyUnserialize`. 
```
Message class "DDTrace\Integrations\SymfonyMessenger\DDTraceStamp" not found during decoding.
```
The native PHP serializer use `base64_encode` and `unserialize`. The latter will inevitably fail when `DDTraceStamp` is non-existent on non-instrumented clients.

---

The usage of `HandlerArgumentsStamp` was also considered, but it only was added from the 6.2 branch onwards.

---

The proposed alternative makes usage of `SerializerStamp` instead of `DDTraceStamp`. The former is shipped OOTB on supported versions. Injection/Extraction has been modified to accommodate for this new stamp.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Looks good to me, thanks for realizing we have multiple Serializers and writing the tests!

src/DDTrace/Integrations/SymfonyMessenger/DDTraceStamp.php
```diff
@@ -1,20 +0,0 @@
...
-    public function __construct(array $headers)
-    {
-        $this->headers = $headers;
-    }
-
-    public function getHeaders(): array
-    {
-        return $this->headers;
-    }
-}
```
Jean85, line 20: Please do not delete this, or a rolling update of the extension will break if the receiver is updated first.

Mark it simply as deprecated.

PR #2960, opened 2024-11-18
Keep DDTraceStamp

bwoebi:
While we no longer make use of it, immediately removing it may break existing applications. Let's keep it for a few versions and drop it later.

See: https://github.com/DataDog/dd-trace-php/pull/2956#pullrequestreview-2442766665

### DIFF
--- a/src/DDTrace/Integrations/SymfonyMessenger/DDTraceStamp.php
+++ b/src/DDTrace/Integrations/SymfonyMessenger/DDTraceStamp.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace DDTrace\Integrations\SymfonyMessenger;
+
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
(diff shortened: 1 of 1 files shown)


src/DDTrace/Integrations/SymfonyMessenger/DDTraceStamp.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace DDTrace\Integrations\SymfonyMessenger;
+
+use Symfony\Component\Messenger\Stamp\StampInterface;
+
+final class DDTraceStamp implements StampInterface
```
PROFeNoM, line 7:
```suggestion
/**
 * @deprecated since 1.5.0
 */
final class DDTraceStamp implements StampInterface
```

Shouldn't we mark it as deprecated?

PROFeNoM (approved):

PR #2980, opened 2024-12-03
ci(all): Pin Ubuntu version

realFlowControl:
### Description

GitHub [announced](https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/?#ubuntu-latest-upcoming-breaking-changes) to migrated `ubuntu-latest` to version 24.04 starting December 5th, so this is just to make sure this will work as expected.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/auto_add_pr_to_miletone.yml
+++ b/.github/workflows/auto_add_pr_to_miletone.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   add_to_milestone:
     if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.title, '[Version Bump]') == false
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.02
     permissions:
       contents: read
       pull-requests: write # need to modify existing PR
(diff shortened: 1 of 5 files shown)


.github/workflows/auto_add_pr_to_miletone.yml
```diff
@@ -10,17 +10,17 @@ on:
 jobs:
   add_to_milestone:
     if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.title, '[Version Bump]') == false
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.02
```
iamluc, line 13: Shouldn't it be `22.04` instead? (same below)

realFlowControl: Good catch, it should be `24.04` instead of `02`.
As of now, `ubuntu-latest` is 22.04, but starting December 5 it will be 24.04, so this will upgrade an pin.

iamluc (approved):

PR #2981, opened 2024-12-03
Fix PHP 8.4 deprecations

JeppeKnockaert:
### Description

This PR removes implicitly nullable parameters because these result in deprecation warnings for PHP 8.4.

https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.core.implicitly-nullable-parameter

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/AMQP/AMQPIntegration.php
+++ b/src/DDTrace/Integrations/AMQP/AMQPIntegration.php
@@ -486,7 +486,7 @@ class AMQPIntegration extends Integration
         SpanData $span,
         string $name,
         string $spanKind,
-        string $resourceDetail = null,
+        $resourceDetail = null,
         $exception = null
     ) {
         $span->name = "amqp.$name";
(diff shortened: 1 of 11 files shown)


bwoebi:
Hey @JeppeKnockaert,

Thanks for the contribution, however we still support 7.0, so that's a bit unfortunate, as this PR breaks that version.

I believe the only solution to keep it working on 7.0 and no incurring the deprecations is removing the types for the nullable values altogether.

JeppeKnockaert:
@bwoebi Thanks for reviewing, I updated the PR to remove the implicitly nullable types instead

bwoebi (approved):
Thank you for the changes.
I've moved the OpenTelemetry part back to nullable types as that code anyway isn't available for PHP 7.0.

realFlowControl:
Hey @JeppeKnockaert 👋 
thanks for this PR, it has been released with [`v1.6.0`](https://github.com/DataDog/dd-trace-php/releases/tag/1.6.0) just now 🎉

PR #2985, opened 2024-12-05
[loader] Assign an appropriate tag to each type of error

iamluc:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/loader/dd_library_loader.c
+++ b/loader/dd_library_loader.c
@@ -217,6 +217,9 @@ void ddloader_logf(log_level level, const char *format, ...) {
     va_end(va);
 }
 
+/**
+ * @param error The c-string this is pointing to must not exceed 150 bytes
+ */
 static void ddloader_telemetryf(telemetry_reason reason, const char *error, const char *format, ...) {
     log_level level = ERROR;
     switch (reason) {
--- a/loader/dd_library_loader.c
+++ b/loader/dd_library_loader.c
@@ -31,7 +31,7 @@ static bool already_done = false;
 # define OS_PATH "linux-gnu/"
 #endif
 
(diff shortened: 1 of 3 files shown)


loader/dd_library_loader.c
```diff
@@ -256,13 +261,15 @@ static void ddloader_telemetryf(telemetry_reason reason, const char *format, ...
...
     switch (reason) {
         case REASON_ERROR:
-            points =
-                "\
-                {\"name\": \"library_entrypoint.error\"}, \"tags\": [\"error_type:NA\"]}\
-            ";
+            snprintf(points_buf, sizeof(points_buf), "\
+                    {\"name\": \"library_entrypoint.error\", \"tags\": [\"error_type:%s\"]}\
+                ",
+                error ? error : "NA"
```
realFlowControl, line 271: In case `error` points to a char[] which is "too long", this would leave you with an invalid JSON in `points_buf`. I would like to somehow make this more explicit, I guess a comment for `ddloader_telemetryf` would already be okay.

loader/dd_library_loader.c
```diff
@@ -217,7 +217,8 @@ void ddloader_logf(log_level level, const char *format, ...) {
     va_end(va);
 }
 
-static void ddloader_telemetryf(telemetry_reason reason, const char *format, ...) {
+static void ddloader_telemetryf(telemetry_reason reason, const char *error, const char *format, ...) {
```
realFlowControl, line 220:
```suggestion
/**
 * @param error The c-string this is pointing to must not exceed 150 bytes
 */
static void ddloader_telemetryf(telemetry_reason reason, const char *error, const char *format, ...) {
```

See other comment

realFlowControl (approved):
Looks good to me, apart from one nitpick.

PR #2988, opened 2024-12-11
fix(profiling) invalid unicode handling in timeline

realFlowControl:
### Description

Fixes handling of invalid unicode for errors in timeline

### Reviewer checklist
- [ ] Test coverage seems ok.
- [ ] Appropriate labels assigned.

PR #2990, opened 2024-12-11
fix(profiling): mismatch in utf8 handling

morrisonlevi:
### Description

Happens in timeline filename for zend_error_observer.

I found this mismatch while doing something else. I asked Florian about it and he prepared a different way to fix it in #2988. I thought I'd push mine up too so we can compare.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/src/timeline.rs
+++ b/profiling/src/timeline.rs
@@ -204,21 +204,20 @@ unsafe extern "C" fn ddog_php_prof_zend_error_observer(
     }
 
     #[cfg(zend_error_observer_80)]
-    let file = unsafe {
-        let mut len = 0;
-        let file = file as *const u8;
-        while *file.add(len) != 0 {
-            len += 1;
-        }
-        std::str::from_utf8_unchecked(std::slice::from_raw_parts(file, len)).to_string()
+    let filename = unsafe {
+        let cstr = core::ffi::CStr::from_ptr(file);
+        cstr.to_string_lossy().into_owned()
     };
     #[cfg(not(zend_error_observer_80))]
-    let file = unsafe { zend::zai_str_from_zstr(file.as_mut()).into_string() };
+    let filename = unsafe {
+        let zstr = zai_str_from_zstr(file.as_mut());
+        zstr.into_string_lossy().into_owned()
+    };
 
     let now = SystemTime::now().duration_since(UNIX_EPOCH).unwrap();
     if let Some(profiler) = Profiler::get() {
         let now = now.as_nanos() as i64;
-        profiler.collect_fatal(now, file, line, unsafe {
+        profiler.collect_fatal(now, filename, line, unsafe {
             zend::zai_str_from_zstr(message.as_mut()).into_string()
         });
     }
--- a/profiling/src/timeline.rs
+++ b/profiling/src/timeline.rs
@@ -204,15 +204,11 @@ unsafe extern "C" fn ddog_php_prof_zend_error_observer(
     }
 
     #[cfg(zend_error_observer_80)]
-    let filename = unsafe {
-        let cstr = core::ffi::CStr::from_ptr(file);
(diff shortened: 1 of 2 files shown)


ekump (approved):

profiling/src/timeline.rs
```diff
@@ -204,21 +204,20 @@ unsafe extern "C" fn ddog_php_prof_zend_error_observer(
...
-        std::str::from_utf8_unchecked(std::slice::from_raw_parts(file, len)).to_string()
+    let filename = unsafe {
+        let cstr = core::ffi::CStr::from_ptr(file);
+        cstr.to_string_lossy().into_owned()
     };
     #[cfg(not(zend_error_observer_80))]
-    let file = unsafe { zend::zai_str_from_zstr(file.as_mut()).into_string() };
+    let filename = unsafe {
+        let zstr = zai_str_from_zstr(file.as_mut());
+        zstr.into_string_lossy().into_owned()
```
realFlowControl, line 214:
```suggestion
        zstr.into_string()
```
The `into_string()` for `ZaiStr` looks like it does the same, so I'd suggest we use it, but I do not have any hard feelings on this (aka: feel free to ignore)

profiling/src/timeline.rs
```diff
@@ -204,21 +204,20 @@ unsafe extern "C" fn ddog_php_prof_zend_error_observer(
...
-    let file = unsafe {
-        let mut len = 0;
-        let file = file as *const u8;
-        while *file.add(len) != 0 {
-            len += 1;
-        }
-        std::str::from_utf8_unchecked(std::slice::from_raw_parts(file, len)).to_string()
+    let filename = unsafe {
+        let cstr = core::ffi::CStr::from_ptr(file);
+        cstr.to_string_lossy().into_owned()
```
realFlowControl, line 209: I like the symmetry of the code when using `zai_str_from_cstr()` (like in the other PR) but it feels that it is just the same as this code with extra steps 😄

realFlowControl (approved):
LGTM 🎉

profiling/src/timeline.rs
morrisonlevi: Perhaps you would like this better:

```rs
    #[cfg(zend_error_observer_80)]
    let filename_str = unsafe { core::ffi::CStr::from_ptr(file) };
    #[cfg(not(zend_error_observer_80))]
    let filename_str = unsafe { zai_str_from_zstr(file.as_mut()) };
    
    let filename = filename_str.into_string_lossy().into_owned();
```

The type of `filename_str` varies which is why I didn't write it this way, but you can see how the operations are identical. The way I wrote it shows that, none of the lossy-bits are hidden away. They're doing the same thing on both branches.

Note that you can't save the `into_string_lossy()` into a variable like this because of lifetimes:

```rs
    #[cfg(zend_error_observer_80)]
    let filename_lossy = unsafe { core::ffi::CStr::from_ptr(file).into_string_lossy() };
    #[cfg(not(zend_error_observer_80))]
    let filename_lossy = unsafe { zai_str_from_zstr(file.as_mut()).into_string_lossy() };
    
    let filename = filename_str.into_owned();
```

This would have been my preferred way to write it.

realFlowControl: LGTM

morrisonlevi: To clarify: you like the PR as-is or do you like this better?

```rust
    #[cfg(zend_error_observer_80)]
    let filename_str = unsafe { core::ffi::CStr::from_ptr(file) };
    #[cfg(not(zend_error_observer_80))]
    let filename_str = unsafe { zai_str_from_zstr(file.as_mut()) };
    
    let filename = filename_str.into_string_lossy().into_owned();
```

realFlowControl: I like the change you proposed in the comment above better

PR #2957, opened 2024-11-15
Block remote config signals during ftp functions

bwoebi:
These are sensitive to EINTR, so block the signal during their execution.

Fixes #2952.

PR #2999, opened 2024-12-17
Block signals for more functions

iamluc:
### Description

Follow-up of https://github.com/DataDog/dd-trace-php/pull/2957

At least 2 customers reported random issues with database connections that seems to be fixed when remote config is disabled.
For one of the reports, the error was `EINPROGRESS` (`Operation now in progress`), which could seems unrelated, but actually [a bug in PHP](https://github.com/php/php-src/pull/13252#issuecomment-2438917586) (likely fixed by https://github.com/php/php-src/pull/16606) could return this error while the real one is `EINTR`

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/handlers_signal.c
+++ b/ext/handlers_signal.c
@@ -23,9 +23,15 @@ static void dd_handle_signal(zif_handler original_function, INTERNAL_FUNCTION_PA
 }
 
 #define BLOCKSIGFN(function) \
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
Looks good, thanks for the patch, Luc :-)

PR #3004, opened 2024-12-18
feat: Kafka Integration

PROFeNoM:
### Description

image

This integration's particularity lies in an internal function call substitution from `ProducerTopic::produce` to `ProducerTopic::producev`.

```
public produce ( integer $partition , integer $msgflags [, string $payload = NULL [, string $key = NULL [, string $opaque = NULL ]]] ) : void
public producev ( integer $partition , integer $msgflags , string $payload [, string $key = NULL [, array $headers = NULL [, integer $timestamp_ms = NULL [, string $opaque = NULL ]]]] ) : void
```

This substitution is a trick to work around the missing `$headers` parameters when using `ProducerTopic::produce`. Remember that this parameter injects distributed tracing headers.

Moreover, the consumer span is started _a posteriori_ - i.e., the start time of the consuming call is saved, and either a root span or child span is created depending on the retrieved headers from the message during the post hook. This is because not all consuming operation will receive a message with headers, and therefore, not all consuming operation ought to be a distributed span of its own.

### Reviewer checklist
- [ ] Test coverage seems ok.


pablomartinezbernardo (approved):

ext/handlers_kafka.c
```diff
@@ -0,0 +1,110 @@
+#ifndef _WIN32
+
+#include <php.h>
+#include <stdbool.h>
+#include "configuration.h"
+#include "handlers_api.h"
+#include "handlers_http.h" // For ddtrace_inject_distributed_headers
```
bwoebi, line 7:
```suggestion
```

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
```diff
@@ -0,0 +1,233 @@
...
+        $consumerMethods = [
+            'RdKafka\KafkaConsumer::consume',
+            'RdKafka\Queue::consume'
+        ];
+
+        foreach ($consumerMethods as $method) {
+            \DDTrace\install_hook(
+                $method,
+                function (HookData $hook) use ($integration) {
+                    $hook->data['start'] = \DDTrace\now();
```
bwoebi, line 110: Is there a specific reason you don't simply use `microtime(true)`? Which the `\DDTrace\start_span()` happily accepts?

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
```diff
@@ -0,0 +1,233 @@
...
+        //      array $headers = NULL [,
+        //      integer $timestamp_ms = NULL [,
+        //      string $opaque = NULL
+        // ]]]] ) : void
+        $argsCount = count($args);
+        if ($argsCount >= 5) {
+            $args[4] = array_merge($args[4] ?? [], $headers);
+        } elseif ($argsCount === 4) {
+            $args[] = $headers;
+        } else { // $argsCount === 3
```
bwoebi, line 90: Please actually add the if () here. Hooks are called before an ArgumentCountError would be emitted.
Just don't do anything if there are not enough args.

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
```diff
@@ -0,0 +1,233 @@
...
+
+    const METADATA_MAPPING = [
+        'metadata.broker.list' => Tag::KAFKA_HOST_LIST,
+        'group.id' => Tag::KAFKA_GROUP_ID,
+        'client.id' => Tag::KAFKA_CLIENT_ID
+    ];
+
+    public function init(): int
+    {
+        if (strpos(phpversion('rdkafka'), '6.') !== 0) {
```
bwoebi, line 24:
```suggestion
        if (strtok(phpversion('rdkafka'), '.') >= 6) {
```
Can we simply accept 6+ and adjust if needed?

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
```diff
@@ -0,0 +1,233 @@
...
+        ];
+
+        foreach ($consumerMethods as $method) {
+            \DDTrace\install_hook(
+                $method,
+                function (HookData $hook) use ($integration) {
+                    $hook->data['start'] = \DDTrace\now();
+                },
+                function (HookData $hook) use ($integration) {
+                    $integration->processConsumedMessage($hook);
```
bwoebi, line 113: To be honest, having this part in an extra method was confusing to me :-D
Mostly because I first saw close_span() and was confused where the start span was...

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
```diff
@@ -0,0 +1,233 @@
...
+    {
+        /** @var \RdKafka\Message $message */
+        $message = $hook->returned;
+
+        if ($message) {
+            $headers = array_filter(KafkaIntegration::extractMessageHeaders($message->headers ?? []), function($value) {
+                return $value !== null;
+            });
+
+            if (!empty($headers)) {
```
bwoebi, line 131:
```suggestion
            if ($message->headers && $link = SpanLink::fromHeaders($message->headers)) {
```
`SpanLink::fromHeaders()` returns null if there's nothing to extract or extraction failed. This is better than checking for specific headers existence (and you also forgot b3).

(you may then use `$link` directly in the `$span->links[] =` assignment too.

ext/handlers_kafka.c
```diff
@@ -0,0 +1,104 @@
...
+        Z_PARAM_STRING_OR_NULL(key, key_len)
+        Z_PARAM_STR_OR_NULL(opaque)
+    ZEND_PARSE_PARAMETERS_END();
+
+    zval args[MAX_PRODUCEV_ARGS];
+    dd_initialize_producev_args(args, partition, msgflags, payload, payload_len, key, key_len, opaque);
+
+    zval function_name;
+    ZVAL_STRING(&function_name, "producev");
+    call_user_function(NULL, getThis(), &function_name, return_value, 6 + opaque_param, args);
```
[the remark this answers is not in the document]
bwoebi: No, it will fatal error. We should check for function existence in the startup function and not install the handler in that case.

ext/handlers_kafka.c
```diff
@@ -0,0 +1,110 @@
...
+
+// True global - only modify during MINIT/MSHUTDOWN
+static bool dd_ext_kafka_loaded = false;
+static uint32_t opaque_param = 0;
+
+static zif_handler dd_kafka_produce_handler = NULL;
+
+static bool rdkafka_version_supported(void) {
+    zend_module_entry *rdkafka_me = zend_hash_str_find_ptr(&module_registry, ZEND_STRL("rdkafka"));
+    return rdkafka_me && strncmp(rdkafka_me->version, "6", 1) == 0;
```
bwoebi, line 21: Same here, please allow 6+.

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
PROFeNoM: It was simply for a matter of being as precise as possible (or at least, as precise as the way we are usually starting span) since `microtime(true)`, afaik, isn't nanosecond precision.

With that said, if you judge the difference to be negligible, then I can remove the usage of `\DDTrace\now` and replace it with microtime :+1:

ext/handlers_kafka.c
```diff
@@ -0,0 +1,104 @@
+#ifndef _WIN32
```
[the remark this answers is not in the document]
PROFeNoM: Relic of a previous implementation :+1: Removed.

src/DDTrace/Integrations/Kafka/KafkaIntegration.php
bwoebi: I do judge it negligible, but even in case we wouldn't, there's still gettimeofday() returning sec and usec.

bwoebi (approved):
Nicely done, thanks Alexandre :-)

ext/ddtrace.c
```diff
@@ -3,6 +3,7 @@
 #endif
 #include "ddtrace.h"
 #include <SAPI.h>
+#include <time.h>
```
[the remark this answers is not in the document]
PROFeNoM: yes, ty :+1:

iNviNho:
@bwoebi @PROFeNoM Is there a way to disable tracing `kafka.consumer` and `kafka.produce` resources?

We are manually starting traces in our php cli commands and therefore such traces are useless and increase the costs.

image

PROFeNoM:
Hey @iNviNho :wave: 

You can disable the kafka integration (and hence, the creation of these spans) by setting `DD_TRACE_KAFKA_ENABLED=0`.
Let me know if this works for you :+1:

iNviNho:
> Hey @iNviNho 👋
> 
> You can disable the kafka integration (and hence, the creation of these spans) by setting `DD_TRACE_KAFKA_ENABLED=0`. Let me know if this works for you 👍

I also tried to sample them but that seems not to work
```
ENV DD_TRACE_SAMPLING_RULES='[{"service": "broker.offer-service", "resource": "kafka.consume", "sample_rate": 0.0}, {"service": "broker.offer-service", "resource": "kafka.produce", "sample_rate": 0.0}]'
```
:(

PR #3005, opened 2024-12-19
Bump rust to 1.78 and add feature to Cargo.toml according to libdatadog changes

dubloom:
### Description

libdatadog removed the set_cgroup_file feature. However it is used in some dd-trace-php tests. As we are not sure they can be removed and without any changes the tracer cannot be built, they added back but under a feature flag.

This PR activates the feature flag and therefore enable the tracer to be built again.

### Reviewer checklist
- [ ] Test coverage seems ok.


iamluc:
@dubloom You probably have to update the reference of the libdatadog submodule to make it work

components-rs/Cargo.toml
```diff
@@ -8,7 +8,7 @@ crate-type = ["staticlib", "cdylib"]
 path = "lib.rs"
 
 [dependencies]
-ddcommon = { path = "../libdatadog/ddcommon" }
+ddcommon = { path = "../libdatadog/ddcommon", features = ["cgroup_testing"] }
```
ekump, line 11: If this is for tests, you probably don't want this feature as a regular dependency. You can leave the ddcommon dependency as it was here, and under dev-dependencies add it again with this feature.

dubloom: i might do it wrong, but when i try, the`make install` command will fail because set_cgroup_file is not found.

bwoebi: It's for tests, but for PHP extension tests. Which are (also) run on release builds.

bwoebi (approved):

morrisonlevi (approved):
Approved on the profiling side 👍🏻

PR #3006, opened 2024-12-19
feat: send integration error logs through telemetry

dubloom:
### Description

This PR will add logs telemetry to the php-tracer. With it, all integration error logs (which are warning at the moment) will be send through telemetry as the same time as metrics telemetry.

- Each time an integration error log is issued, it is stored. They are all retrieved in finalize_telemetry() and send to the sidecar.
- The DD_TELEMETRY_LOG_COLLECTION_ENABLED flag should be activated (by default it is) to enable this functionnality.
- The deduplication of logs is done by the sidecar, logs with the same identifier are sent only one time with a count value > 1. We use the hash of the message to identify a log.
- There is no stack trace in PHP but some file path can appear. Therefore, the file path stored are redacted, if it is a code in the DDTrace folder (should be happen all the time), we redact everything before this folder. Otherwise, we only keep the name of the PHP file.
- Two tests has been added in ext/telemetry: 
   - integration_load_error.phpt. It checks that an error happening during integration loading is sent to the telemetry.
   - integration_runtime_error.phpt. It checks that an error happening during the execution of an integration callback is sent to the telemetry 

Initiative documented [here](https://docs.google.com/document/d/1Wny7oRVKh19XQcrW-BtB9IQXXXs-pTCRnXYEDKLAHHM/edit?tab=t.0)

### Reviewer checklist
- [ ] Test coverage seems ok.


components-rs/telemetry.rs
```diff
@@ -166,3 +171,24 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_add_span_metric_point_buffer(
         tags,
     )));
 }
+
+#[no_mangle]
+pub unsafe extern "C" fn ddog_sidecar_telemetry_add_integration_log_buffer(
+    buffer: &mut SidecarActionsBuffer,
+    log: CharSlice
+) {
+    let mut hasher = DefaultHasher::new();
```
bwoebi, line 180: I think a stable hash would be better here? We use ZwoHasher for that purpose in libdatadog.

ext/configuration.h
```diff
@@ -126,6 +126,7 @@ enum ddtrace_sampling_rules_format {
     CONFIG(BOOL, DD_TRACE_ENABLED, "true", .ini_change = ddtrace_alter_dd_trace_disabled_config,               \
            .env_config_fallback = ddtrace_conf_otel_traces_exporter)                                           \
     CONFIG(BOOL, DD_INSTRUMENTATION_TELEMETRY_ENABLED, "true", .ini_change = zai_config_system_ini_change)     \
+    CONFIG(BOOL, DD_TELEMETRY_LOG_COLLECTION_ENABLED, "true", .ini_change = zai_config_system_ini_change)      \
```
bwoebi, line 129:
```suggestion
    CONFIG(BOOL, DD_TELEMETRY_LOG_COLLECTION_ENABLED, "true")      \
```
There's no need for having this system INI; just replace all checks with simple get_ instead of get_global_.

ext/telemetry.c
```diff
@@ -6,6 +6,7 @@
 #include "telemetry.h"
 #include "serializer.h"
 #include "sidecar.h"
+#include <string.h>
```
bwoebi, line 9:
```suggestion
```
Stray include?

components-rs/log.rs
```diff
@@ -211,3 +213,28 @@ pub extern "C" fn ddog_get_logs_count(level: CharSlice) -> u32 {
...
+        }
+    });
+}
+
+#[no_mangle]
+pub extern "C" fn ddog_add_integration_error_log(log: CharSlice) {
+    INTEGRATION_ERROR_LOGS.with(|logs| {
+        logs.borrow_mut().push(log.to_utf8_lossy().to_string());
+    });
+}
```
bwoebi, line 239: Why do we need this at all, and cannot simply call `ddog_sidecar_telemetry_add_integration_log_buffer` everywhere? Why do we buffer this in a Vec here? (you'd move the logic from components/telemetry/telemetry.c to ext/telemetry.c, and call `ddog_sidecar_telemetry_add_integration_log_buffer(ddtrace_telemetry_buffer(), log_slice)` directly)

ext/integrations/integrations.c
```diff
@@ -98,12 +99,18 @@ static void dd_invoke_integration_loader_and_unhook_posthook(zend_ulong invocati
                 zend_class_entry *ce = zend_lookup_class(aux->classname);
                 if (!ce) {
                     LOG(WARN, "Error loading deferred integration %s: Class not loaded and not autoloadable", ZSTR_VAL(aux->classname));
+                    if (get_global_DD_TELEMETRY_LOG_COLLECTION_ENABLED()) {
```
bwoebi, line 102: All logging here should also check for `get_global_DD_INSTRUMENTATION_TELEMETRY_ENABLED()`.

components-rs/telemetry.rs
dubloom: I did that because this is what is used [here](https://github.com/DataDog/libdatadog/blob/0957884d15dac4aa6df77cd258aea4fd192532d2/ddtelemetry/src/worker/mod.rs#L848). But i can switch to ZwoHasher

components/telemetry/telemetry.c
```diff
@@ -0,0 +1,44 @@
+#include "telemetry.h"
+#include "../components-rs/ddtrace.h"
+#include <stdio.h>
+#include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
+
+static void ddog_telemetryf_va(const char* format, va_list va, void (*telemetry_handler)(ddog_CharSlice)) {
```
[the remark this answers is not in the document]
dubloom: At the moment, its sole use is the function below but the structure allows reusability if we want to use telemetry logs for an application other than integration errors.

components-rs/log.rs
dubloom: My understanding is that we want to send the telemetry at the end of a request. Therefore the Vec vector is used to store the logs temporary to send it at the end. Adding directly to the log buffer may send the logs along the request.

If you think that is not the problem, your suggestion sounds indeed appropriate.

bwoebi: I think that's not the problem. It adds complexity in the code for no benefit - what we want is no unnecessary submissions, i.e. not start the submission machinery for every single thing, it has some fixed overhead. But if we flush anyways, that's fine too.

components/telemetry/telemetry.c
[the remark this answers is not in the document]
bwoebi: Alright, seems like a case of premature forward-thinking, but I don't insist here.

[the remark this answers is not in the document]
dubloom: To be honest, it kind is. But I think it does not add any real complexity and it could be useful

bwoebi (approved):
Thank you; fixed up some testing and windows stuff, but now it's great, merging :-)

PR #3008, opened 2024-12-19
appsec: fix recv/writev calls in the face of interrupting signals

cataphract:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -29,8 +29,6 @@ static inline ATTR_WARN_UNUSED mpack_error_t _omsg_finish(
 static inline void _omsg_destroy(dd_omsg *nonnull omsg);
 static inline dd_result _omsg_send(
     dd_conn *nonnull conn, dd_omsg *nonnull omsg);
-static inline dd_result _omsg_send_cred(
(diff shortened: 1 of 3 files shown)


appsec/src/extension/commands_helpers.c
```diff
@@ -42,16 +40,14 @@ typedef struct _dd_imsg {
     mpack_node_t root;
 } dd_imsg;
 
-// iif these two return success, _imsg_destroy must be called
-static inline dd_result _imsg_recv(
-    dd_imsg *nonnull imsg, dd_conn *nonnull conn);
-static inline ATTR_WARN_UNUSED dd_result _imsg_recv_cred(
+// iif this returns success, _imsg_destroy must be called
```
estringana, line 43: typo

estringana (approved):

appsec/src/extension/commands_helpers.c
cataphract: no, it's correct, `_imsg_destroy` should be called if and only if `_imsg_recv` returns success. I did spot a branch where it was not being called, though, so I added a new commit to use a gcc/clang extension to prevent this.

PR #3014, opened 2025-01-03
Implement SSRF

estringana:
### Description

Adds server-side request forgery (SSRF) detection on top of the RASP architecture introduced in #2770.

The functional change is wrapping outbound HTTP calls  through`file_get_contents`, `file_put_contents`, `fopen`, and `readfile` with HTTP URLs and pushing the target URL and protocol to the WAF before the request is made. If the WAF determines the target is suspicious, the request is blocked and the exploit is reported on the span.

The more significant change in this PR is an API refactor: `push_address()` is replaced by `push_addresses()`, which takes a map of `address:value` pairs and pushes them all in a single WAF call. This was required for SSRF because the WAF rule needs both the URL and the protocol to evaluate correctly. With the old API, you'd have to call it twice, which would trigger two independent WAF evaluations per request. Batching them into a single call is more efficient.

[APPSEC-52930]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-52930]: https://datadoghq.atlassian.net/browse/APPSEC-52930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appsec/src/extension/ddappsec.c
+++ b/appsec/src/extension/ddappsec.c
@@ -126,7 +126,7 @@ static void ddappsec_sort_modules(void *base, size_t count, size_t siz,
 
     // Reorder ddappsec to ensure it's always after ddtrace
     for (Bucket *module = base, *end = module + count, *ddappsec_module = NULL;
-        module < end; ++module) {
+         module < end; ++module) {
         zend_module_entry *m = (zend_module_entry *)Z_PTR(module->val);
         if (m->name == ddappsec_module_entry.name) {
             ddappsec_module = module;
--- a/appsec/src/extension/ddappsec.c
+++ b/appsec/src/extension/ddappsec.c
@@ -126,7 +126,7 @@ static void ddappsec_sort_modules(void *base, size_t count, size_t siz,
 
     // Reorder ddappsec to ensure it's always after ddtrace
     for (Bucket *module = base, *end = module + count, *ddappsec_module = NULL;
-         module < end; ++module) {
+        module < end; ++module) {
         zend_module_entry *m = (zend_module_entry *)Z_PTR(module->val);
         if (m->name == ddappsec_module_entry.name) {
             ddappsec_module = module;
@@ -474,7 +474,7 @@ static PHP_FUNCTION(datadog_appsec_testing_request_exec)
     RETURN_TRUE;
 }
 
-static PHP_FUNCTION(datadog_appsec_push_address)
+static PHP_FUNCTION(datadog_appsec_push_addresses)
 {
     struct timespec start;
     struct timespec end;
@@ -482,15 +482,14 @@ static PHP_FUNCTION(datadog_appsec_push_address)
     long elapsed = 0;
     UNUSED(return_value);
     if (!DDAPPSEC_G(active)) {
-        mlog(dd_log_debug, "Trying to access to push_address "
+        mlog(dd_log_debug, "Trying to access to push_addresses "
                            "function while appsec is disabled");
         return;
     }
 
-    zend_string *key = NULL;
-    zval *value = NULL;
+    HashTable *addresses = NULL;
     bool rasp = false;
-    if (zend_parse_parameters(ZEND_NUM_ARGS(), "Sz|b", &key, &value, &rasp) ==
+    if (zend_parse_parameters(ZEND_NUM_ARGS(), "h|b", &addresses, &rasp) ==
         FAILURE) {
         RETURN_FALSE;
     }
@@ -499,16 +498,32 @@ static PHP_FUNCTION(datadog_appsec_push_address)
         return;
     }
 
+    if (addresses == NULL) {
+        return;
+    }
+
+    zend_array *parameters_arr =
+        zend_new_array(zend_hash_num_elements(addresses));
     zval parameters_zv;
-    zend_array *parameters_arr = zend_new_array(1);
     ZVAL_ARR(&parameters_zv, parameters_arr);
-    zend_hash_add(Z_ARRVAL(parameters_zv), key, value);
-    Z_TRY_ADDREF_P(value);
+
+    zend_string *key;
+    zval *value;
+    ZEND_HASH_FOREACH_STR_KEY_VAL(addresses, key, value)
+    {
+        if (!key) {
+            continue;
+        }
+
+        zend_hash_add(Z_ARRVAL(parameters_zv), key, value);
+        Z_TRY_ADDREF_P(value);
+    }
+    ZEND_HASH_FOREACH_END();
 
     dd_conn *conn = dd_helper_mgr_cur_conn();
     if (conn == NULL) {
         zval_ptr_dtor(&parameters_zv);
-        mlog_g(dd_log_debug, "No connection; skipping push_address");
+        mlog_g(dd_log_debug, "No connection; skipping push_addresses");
         return;
     }
 
@@ -549,16 +564,16 @@ ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(request_exec_arginfo, 0, 1, _IS_BOOL, 0)
 ZEND_ARG_INFO(0, "data")
 ZEND_END_ARG_INFO()
 
-ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(push_address_arginfo, 0, 0, IS_VOID, 1)
-ZEND_ARG_INFO(0, key)
-ZEND_ARG_INFO(0, value)
+ZEND_BEGIN_ARG_WITH_RETURN_TYPE_INFO_EX(
+    push_addresses_arginfo, 0, 0, IS_VOID, 1)
+ZEND_ARG_INFO(0, addresses)
 ZEND_ARG_INFO(0, rasp)
 ZEND_END_ARG_INFO()
 
 // clang-format off
 static const zend_function_entry functions[] = {
     ZEND_RAW_FENTRY(DD_APPSEC_NS "is_enabled", PHP_FN(datadog_appsec_is_enabled), void_ret_bool_arginfo, 0, NULL, NULL)
-    ZEND_RAW_FENTRY(DD_APPSEC_NS "push_address", PHP_FN(datadog_appsec_push_address), push_address_arginfo, 0, NULL, NULL)
+    ZEND_RAW_FENTRY(DD_APPSEC_NS "push_addresses", PHP_FN(datadog_appsec_push_addresses), push_addresses_arginfo, 0, NULL, NULL)
     PHP_FE_END
 };
 static const zend_function_entry testing_functions[] = {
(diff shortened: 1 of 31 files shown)


cataphract (changes requested):
This doesn't seem sufficient to actually subscribe to the `ASM_RASP_*` products. In fact, it seems we're not subscribed even to `ASM_RASP_LFI` (see the logic in `ddog_init_remote_config`).

To avoid these situations, I would strong recommend that you implement an appsec/tests/integration test that submits the RASP configuration via remote config.

Leiyks (approved):

appsec/src/extension/ddappsec.c
```diff
@@ -499,16 +498,32 @@ static PHP_FUNCTION(datadog_appsec_push_address)
         return;
     }
 
+    if (addresses == NULL) {
+        return;
+    }
```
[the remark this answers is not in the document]
estringana: Good point :)

appsec/src/helper/remote_config/product.hpp
```diff
@@ -31,6 +31,8 @@ struct known_products {
         std::string_view{"ASM_FEATURES"}};
     static inline constexpr product ASM_RASP_LFI{
         std::string_view{"ASM_RASP_LFI"}};
+    static inline constexpr product ASM_RASP_SSRF{
+        std::string_view{"ASM_RASP_SSRF"}};
```
Anilm3, line 35: There are no `ASM_RASP_*` RC products...?

src/DDTrace/Integrations/Filesystem/FilesystemIntegration.php
```diff
@@ -51,20 +51,30 @@ public function init(): int
...
+           $uri_parsed = preg_match('/^([a-z]+)\:\/\//', $hook->args[0], $protocol);
+           $protocol = isset($protocol[1]) ? $protocol[1]: "";
+
+           $addresses = [];
+
+           if (empty($protocol) || $protocol === 'file') {
+                $addresses["server.io.fs.file"] = $hook->args[0];
+           }
+
+           if (in_array($variant, ['file_get_contents', 'fopen']) && (empty($protocol) || $protocol === 'http')) {
```
Anilm3, line 68: There are other protocols supported: `https`, `ftp`, `ftps`.

cataphract:
> This doesn't seem sufficient to actually subscribe to the ASM_RASP_* products. In fact, it seems we're not subscribed even to ASM_RASP_LFI (see the logic in ddog_init_remote_config).
> 
> To avoid these situations, I would strong recommend that you implement an appsec/tests/integration test that submits the RASP configuration via remote config.


Anil brought to me attention the fact that there are no new products related to RASP, only capabilities. So:

* https://github.com/DataDog/libdatadog/pull/814 should be reverted,
* the addition of `RemoteConfigProduct::AsmRaspLfi` should be reverted as well,
* this PR should not contain any references to these nonexistent products on the helper c++ files,
* `ddog_init_remote_config` should be modified to send these rasp capabilities,
* change integration tests so that it's tested that these capabilities are sent (there is already a test that checks the sent capabilities),
* add integration test for rasp rules, with configuration either via remote config or a local json file.

src/DDTrace/Integrations/Filesystem/FilesystemIntegration.php
estringana: Done

estringana:
@cataphract I have addressed all your points

appsec/src/extension/ddappsec.c
```diff
@@ -499,16 +498,28 @@ static PHP_FUNCTION(datadog_appsec_push_address)
...
+        zend_new_array(zend_hash_num_elements(addresses));
     zval parameters_zv;
-    zend_array *parameters_arr = zend_new_array(1);
     ZVAL_ARR(&parameters_zv, parameters_arr);
-    zend_hash_add(Z_ARRVAL(parameters_zv), key, value);
-    Z_TRY_ADDREF_P(value);
+
+    zend_string *key;
+    zval *value;
+    ZEND_HASH_FOREACH_STR_KEY_VAL(addresses, key, value)
```
Anilm3, line 508: I'm not sure if a copy of the array is needed, as we discussed it might be better to deal with the potential lack of keys during serialisation to avoid a copy.

estringana: Done

appsec/src/extension/ddappsec.c
```diff
@@ -500,15 +499,13 @@ static PHP_FUNCTION(datadog_appsec_push_address)
     }
 
     zval parameters_zv;
-    zend_array *parameters_arr = zend_new_array(1);
-    ZVAL_ARR(&parameters_zv, parameters_arr);
-    zend_hash_add(Z_ARRVAL(parameters_zv), key, value);
-    Z_TRY_ADDREF_P(value);
+    ZVAL_ARR(&parameters_zv, addresses);
+    Z_TRY_ADDREF_P(&parameters_zv);
```
cataphract, line 503: I don't think this is correct:

```c
#define ZVAL_ARR(z, a) do {					\
   zend_array *__arr = (a);				        \
   zval *__z = (z);						        \
   Z_ARR_P(__z) = __arr;					\
   Z_TYPE_INFO_P(__z) = IS_ARRAY_EX;		\
 } while (0)

#define IS_ARRAY_EX					(IS_ARRAY          | (IS_TYPE_REFCOUNTED << Z_TYPE_FLAGS_SHIFT) | (IS_TYPE_COLLECTABLE << Z_TYPE_FLAGS_SHIFT))

#define Z_TRY_ADDREF_P(pz) do {	\
  if (Z_REFCOUNTED_P((pz))) {		\
    Z_ADDREF_P((pz));			        \
   }								\
 } while (0)

#define Z_REFCOUNTED(zval)			((Z_TYPE_FLAGS(zval) & IS_TYPE_REFCOUNTED) != 0)

#define Z_ADDREF_P(pz)				zval_addref_p(pz)

static zend_always_inline uint32_t zval_addref_p(zval* pz) {
  ZEND_ASSERT(Z_REFCOUNTED_P(pz));
  return GC_ADDREF(Z_COUNTED_P(pz));
}

static zend_always_inline uint32_t zend_gc_addref(zend_refcounted_h *p) {
  ZEND_RC_MOD_CHECK(p);
  return ++(p->refcount);
}
```

Basically `Z_TRY_ADDREF_P` will increase the refcount unconditionally. zval and zend_refcounted (which zend_array is a type of) contain a copy of the (extended) type, but these will be out of sync.

cataphract (changes requested):

appsec/src/extension/ddappsec.c
estringana: Sorry, which function are you suggestion I should use? Without that increase on reference, old versions of php were failing

cataphract: Do you know the reason they were failing? The problem is not incrementing the reference, the problem is incrementing it through a zval that says it's safe to increment the refcount. Fixup the zval type (notice that a static array like one made with [`ZEND_EMPTY_ARRAY`](https://heap.space/xref/PHP-8.4/Zend/zend_hash.h?r=4b9e59b4#90) doesn't use `IS_ARRAY_EX` or increment the refcount (conditionally) directly on the `zend_array` with `GC_TRY_ADDREF`. The same applies to `zval_ptr_dtor` btw.

That is, unless you can show that non-modifiable arrays are copied when they're passed as arguments, but I don't think that's the fcase.

estringana: Sorted. I didn't have to use that extra reference. Thanks for pointing it out

appsec/src/extension/ddappsec.c
```diff
@@ -510,7 +509,6 @@ static PHP_FUNCTION(datadog_appsec_push_addresses)
     }
 
     dd_result res = dd_request_exec(conn, &parameters_zv, rasp);
-    zval_ptr_dtor(&parameters_zv);
```
cataphract, line 516: This is now sort of correct (you still have an invalid zval, but it has no ill-effects, provided that `dd_request_exec` is not changed in the future in a breaking way), but you chose the worst option. Instead of asking for a Hashtable* and converting it to a zval, you can ask zend_parse_parameters for a zval directly

cataphract (approved):

appsec/src/extension/ddappsec.c
estringana: Changed

src/DDTrace/Integrations/Filesystem/FilesystemIntegration.php
```diff
@@ -51,20 +51,30 @@ public function init(): int
...
         return static function (HookData $hook) use ($variant) {
-            if (count($hook->args) == 0 || !is_string($hook->args[0])) {
-                return;
-            }
+           if (count($hook->args) == 0 || !is_string($hook->args[0])) {
+               return;
+           }
+
+           $protocol = [];
+           $uri_parsed = preg_match('/^([a-z]+)\:\/\//', $hook->args[0], $protocol);
```
Anilm3, line 59: Not a big deal but an alternative could've been `parse_url`: https://www.php.net/manual/en/function.parse-url.php

Anilm3 (approved):
lgtm!

FabienLucini:
Hi, there is something we still do not understand here.
We tought that this PR would correct it but it doesnt, here is our problem : 

Each time the WAF detect a page with a blocked URL pattern, the method push_addresses throw a 500, resulting in an error log : 
Uncaught Symfony\Component\ErrorHandler\Error\FatalError: datadog\appsec\push_addresses(): Datadog blocked the request and presented a static error page in /opt/datadog/dd-library/1.7.0/dd-trace-sources/src/DDTrace/Integrations/Laravel/LaravelIntegration.php:142

So instead of presenting the dadatog waf page "you've been blocked", it throws a 500 and present our server 500 page.

How can we avoid this ?

estringana:
> Hi, there is something we still do not understand here. We thought that this PR would correct it but it doesn't, here is our problem :
> 
> Each time the WAF detect a page with a blocked URL pattern, the method push_addresses throw a 500, resulting in an error log : Uncaught Symfony\Component\ErrorHandler\Error\FatalError: datadog\appsec\push_addresses(): Datadog blocked the request and presented a static error page in /opt/datadog/dd-library/1.7.0/dd-trace-sources/src/DDTrace/Integrations/Laravel/LaravelIntegration.php:142
> 
> So instead of presenting the dadatog waf page "you've been blocked", it throws a 500 and present our server 500 page.
> 
> How can we avoid this ?

Thanks for reporting this @FabienLucini . We are discussing this internally and we will come back to you soon.

estringana:
@FabienLucini still on draft but you can follow the fix of these errors here https://github.com/DataDog/dd-trace-php/pull/3110

estringana:
@FabienLucini the PR with the fix got merged. It will be released on the next release

PR #3018, opened 2025-01-06
[ASM] Workaround for boost download issues

Anilm3:
### Description

For some unclear reason, hunter fails to download boost using the predefined URLs pointing to jfrog artifactory. There doesn't seem to be a mechanism to override the default URL used to download a package, consequently the change involves replacing the predefined URLs with a working one directly modifying the relevant cmake file before it's used by hunter.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Quite the hack, but if it works, it works :-D

PR #3020, opened 2025-01-07
fix: root span name in Slim V4

PROFeNoM:
### Description

I couldn't find a reason for the root span name to differ in Slim 3 compared to Slim 4. Furthermore, setting it to `slim.request` is consistent with all other integrations.

This behavior was identified in the linked escalation.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Integrations/Slim/SlimIntegration.php
+++ b/src/DDTrace/Integrations/Slim/SlimIntegration.php
@@ -34,6 +34,7 @@ class SlimIntegration extends Integration
                 // Overwrite root span info
                 $rootSpan = \DDTrace\root_span();
                 $integration->addTraceAnalyticsIfEnabled($rootSpan);
+                $rootSpan->name = 'slim.request';
                 $rootSpan->service = $appName;
                 $rootSpan->meta[Tag::SPAN_KIND] = 'server';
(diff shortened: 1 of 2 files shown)


bwoebi (approved):
Looks simple and correct, thanks :-)

src/DDTrace/Integrations/Slim/SlimIntegration.php
```diff
@@ -34,6 +34,7 @@ function ($app) use ($integration, $appName) {
                 // Overwrite root span info
                 $rootSpan = \DDTrace\root_span();
                 $integration->addTraceAnalyticsIfEnabled($rootSpan);
+                $rootSpan->name = 'slim.request';
```
mrtus, line 37: Q: Can this be overwritten?

A lot of our notebooks & dashboarding are configured on web.request, so it would be a shame if everything needs to be updated again

mrtus: Tagging @PROFeNoM and @bwoebi for visibility

Or if you require me to create an issue by all means I can do that :D

bwoebi: You can easily overwrite it with `\DDTrace\root_span()->name = 'web.request';` at the end of your entrypoint for example.

PR #3032, opened 2025-01-10
style: cargo clippy --fix and fmt on Rust v1.84

morrisonlevi:
### Description

I think there are some `static mut` uses and other things clippy should be complaining about in CI, but it's not. While investigating why we aren't getting more clippy errors, I figured I'd run `cargo +1.84 clippy fix` while I was at it.

We also fix some other small things with the build so that tests pass. Thanks Bob and Florian for contributing those.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl (approved):
Nice ❤️

bwoebi (approved):

PR #3044, opened 2025-01-17
Fall back to stderr if the log path contains colons

bwoebi:
This mitigates issues with logging being to special devices / pipes.


realFlowControl (approved):
looking good 🎉

PR #3050, opened 2025-01-20
Validate DD_TRACE_AGENT_URL

bwoebi:
This avoids panics in the sidecar and crashes in crashtracker initialization.


iamluc (approved):
LGTM, but shouldn't we fix `ddog_endpoint_from_url` instead?

PR #3058, opened 2025-01-28
fix(profiling) switch to `pthread_atfork()` for fork barrier handling

realFlowControl:
### Description

A customer reached out if we do support @swoole in the profiler. Technically we do, but we do not follow forks which @swoole does in `\Swoole\Http\Server` (also when used as a backend to @laravel Octane).

Because we do not observe the `fork()` that is happening in a Swoole HTTP server when it is forking new workers, the profiler is in an undefined state in the forked children which could ultimately lead to a worker that is dead locked and not serving requests.

This PR aims to observe and shutdown the profiler on any `fork()` call for the child, something we have so far only done for userland calls to `pcntl_fork()`, `pcntl_rfork()` and `pcntl_forkx()`. One thing that changes is that we now also "observe" Apache forks: Apache with `mpm_prefork` would go through `MINIT` before forking it's worker processes. This is not a problem, as in this case, there is no profiler at all, as we init in first-`RINIT`.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
IIRC this was done in the handlers as other fork() calls may not be in a valid state for the engine to continue / profiling is not even desired in the other process?

realFlowControl:
> IIRC this was done in the handlers as other fork() calls may not be in a valid state for the engine to continue / profiling is not even desired in the other process?

AFAIK the main reason was that Apache is going through MINIT/STARTUP and then forks. That's why we moved most profiler init things to RINIT / first-RINIT and we *should* be good, but I'll evaluate this.
The reason for this draft (and I'll update the PR description on that) is that `Swoole\Http\Server` (used in Laravel Octane) forks and we currently do not react to this fork in any way, as it is a native `fork()` call. Leaving the child in an undefined state, sometimes blocked and not serving requests anymore. I figured it might be better to handle all forks if we reliably can ;-)

morrisonlevi (approved):
I was initially very resistant to this change. Florian pointed out this line in the manual:

```
       After a fork(2) in a multithreaded process returns in the child,
       the child should call only async-signal-safe functions (see
       signal-safety(7)) until such time as it calls execve(2) to
       execute a new program.
```

And that it means from the point of `fork`, the child must _only_ call async-signal-safe functions until it calls `execve`. The restriction is not limited to the child handler portion of the callback, but any code from the fork until exec. This means it's not really any different from our `pcntl_fork` handler, at least at a distance.

So as long as this isn't messing up known forks in other locations, such Apache process manager doing a fork after the minit/startup process, then it's okay with me.

-----

There other ways to handle this:

 1. Shut down all threads pre-fork, and re-create them after forking. This would allow us to profile children as well. It would add latency because right now threads only have to reach a synchronization point, not complete their current work and then join.
 2. Avoid threads completely by sending everything out-of-process with the sidecar. I am not personally confident enough yet in the sidecar to go all-in here (but my confidence in it is increasing, fortunately).

realFlowControl:
While running a `fork()` loop overnight I found a deadlock on MacOS. It occurs when the `main` thread is in the fork-`prepare` handler, while at the same time we are sending profiles upstream in the `ddprof_upload` thread.

The following is going on:

- the `main` thread is in the `prepare` handler passed to the `pthread_atfork()` call
- to get there, MacOS libc [calls into `_pthread_atfork_prepare_handlers()`](https://github.com/apple-oss-distributions/libpthread/blob/c032e0b076700a0a47db75528a282b8d3a06531a/src/pthread_atfork.c#L93-L106) which looks like this

```C
void
_pthread_atfork_prepare_handlers(void)
{
	pthread_globals_t globals = _pthread_globals();

	_pthread_lock_lock(&globals->pthread_atfork_lock);
	size_t idx;
	for (idx = globals->atfork_count; idx > 0; --idx) {
		struct pthread_atfork_entry *e = &globals->atfork[idx-1];
		if (e->prepare != NULL) {
			e->prepare();
		}
	}
}
```
Notice: it is holding a lock on `globals->pthread_atfork_lock` *while* calling into the `prepare` handlers

Same time in another thread, let's call it `ddprof_upload`, we are uploading a profile to the backend. More specific, we are sending the serialised version already. Even more precise: we are doing a name server resolution before all the TCP/HTTP shenanigans. While doing the host name lookup, MacOS registers new `pthread_atfork()` handlers, look at the stack I found:

```
* thread #6, name = 'tokio-runtime-worker'
  * frame #0: 0x000000019ae78d9c libsystem_kernel.dylib`__ulock_wait2 + 8
    frame #1: 0x000000019aee0aac libsystem_platform.dylib`_os_unfair_lock_lock_slow + 180
    frame #2: 0x000000019aea77b0 libsystem_pthread.dylib`pthread_atfork + 88
    frame #3: 0x000000019af08898 libsystem_info.dylib`si_destination_compare_init_once + 140
    frame #4: 0x000000019aea80c0 libsystem_pthread.dylib`__pthread_once_handler + 76
    frame #5: 0x000000019aee0240 libsystem_platform.dylib`_os_once_callout + 32
    frame #6: 0x000000019aea8058 libsystem_pthread.dylib`pthread_once + 100
    frame #7: 0x000000019af075f4 libsystem_info.dylib`si_destination_compare + 116
    frame #8: 0x000000019ad4c708 libsystem_c.dylib`_isort + 136
    frame #9: 0x000000019aeedfc8 libsystem_info.dylib`_gai_sort_list + 216
    frame #10: 0x000000019aeed090 libsystem_info.dylib`si_addrinfo + 1280
    frame #11: 0x000000019aeecae8 libsystem_info.dylib`getaddrinfo + 168
    frame #12: 0x0000000106b36368 libdatadog_php_profiling.dylib`_$LT$std..sys_common..net..LookupHost$u20$as$u20$core..convert..TryFrom$LT$$LP$$RF$str$C$u16$RP$$GT$$GT$::try_from::_$u7b$$u7b$closure$u7d$$u7d$::h17928286472c438b at net.rs:207:25 [opt]
    frame #13: 0x0000000106b2ebd0 libdatadog_php_profiling.dylib`_$LT$$LP$$RF$str$C$u16$RP$$u20$as$u20$std..net..socket_addr..ToSocketAddrs$GT$::to_socket_addrs::h0a62b5997ae3d025 [inlined] std::sys::pal::common::small_c_string::run_with_cstr_stack::hf59ace9ebe7f5078 at small_c_string.rs:49:18 [opt]
...
```

Remember that the `main` thread is holding the lock because it is currently executing the the prepare handler? So here we are, the `ddprof_upload` thread waiting for the `main` thread to release the `pthread_atfork_lock` it is holding, while the `main` thread is waiting for the `ddprof_upload` thread to arrive at the barrier it is waiting for it 😉 

## Why is this not a problem on Linux*?

\* at least for glibc, I have not checked musl.

- the libc `fork()` will [determine if the env is multithreaded and call a helper](https://github.com/bminor/glibc/blob/8e86549d1417a4618ab98d10aaba427350b321c6/posix/fork.c#L48-L51) to iterate over prepare handlers

```C
  bool multiple_threads = !SINGLE_THREAD_P;
  uint64_t lastrun;

  lastrun = __run_prefork_handlers (multiple_threads);
```

- [the `__run_prefork_handlers()` get's called](https://github.com/bminor/glibc/blob/8e86549d1417a4618ab98d10aaba427350b321c6/posix/register-atfork.c#L113) with a truthy value
- it will [iterate over the prepare handlers and release the lock for the prepare handler list](https://github.com/bminor/glibc/blob/8e86549d1417a4618ab98d10aaba427350b321c6/posix/register-atfork.c#L136-L145) before calling the `prepare` handler and then accquire the lock again after its return:

```C
      if (runp->prepare_handler != NULL)
        {
          if (do_locking)
            lll_unlock (atfork_lock, LLL_PRIVATE);

          runp->prepare_handler ();

          if (do_locking)
            lll_lock (atfork_lock, LLL_PRIVATE);
        }
```

This allows other threads to add `pthread_atfork()` handlers while another thread is already in a prepare handler.

## How do we proceed with this PR?

We'll see ...

PR #3062, opened 2025-01-29
AIDM-455: Generate Markdown Table of Supported Versions

PROFeNoM:
### Description

Akin to what has been made on other tracing repositories, a full record of tested versions and a markdown table of min/max supported versions will be automatically kept up to date.

image

A new CircleCI Job `aggregate_tested_versions` has been added. The latter will run on every commit made to master. It will be in charge to collect tested_versions.json artifacts from CircleCI test jobs, aggregate them, and generate a PR if necessary.

A sample PR can be seen at the following link: https://github.com/DataDog/dd-trace-php/pull/3060 - ([CircleCI Job](https://app.circleci.com/jobs/github/DataDog/dd-trace-php/6059326)). 

Note that two environment variables have been added to CircleCI's settings: `CIRCLECI_TOKEN` and `GITHUB_TOKEN`. They are used to communicate with CircleCI's and Github's APIs, respectively.

As such, two "repository artifacts" will be added at the root of the repo: aggregated_tested_versions.json and integration_versions.md. The former contains all tested versions of target libraries, while the latter summarize min/max supported versions.

---

To record tested versions, a method `recordTestedVersion` will be called during tests' tear down.
- For composer-based library integrations (e.g., guzzle), nothing will have to be done. The logic will handle the retrieval of the target library and its currently tested version, and record it.
- For composer-based web integrations (e.g., laravel) and extensions (e.g., redis), the target library (e.g., `laravel/framework` or **ext-**redis) will have to be specified by overriding `getTestedLibrary`.
- For other cases (e.g., Wordpress), it is possible to override both `getTestedLibrary` and `getTestedVersion` with the tested library/version.

The `tested_versions` Makefie target will **always** be run at the end of `integration-snapshots` jobs, and generate the aforementioned `tested_versions.json` artifact.

### Reviewer checklist
- [ ] Test coverage seems ok.


integration_versions.md
```diff
@@ -0,0 +1,38 @@
...
+| predis/predis               | 1.1.10                 | 1.1.10                 |
+| shardj/zf1-future           | 1.21.4                 | 1.21.4                 |
+| slim/slim                   | 3.12.5                 | 4.14.0                 |
+| spiral/roadrunner           | 2.11.4                 | 2.11.4                 |
+| symfony/console             | 5.2.14                 | 7.2.1                  |
+| symfony/framework-bundle    | 2.3.42                 | 7.2.2                  |
+| symfony/messenger           | 4.4.49                 | 7.2.1                  |
+| wordpress                   | 4.8.10                 | 6.1.1                  |
+| yiisoft/yii2                | 2.0.49                 | 2.0.51                 |
+| zendframework/zf1           | 1.12.20                | 1.12.20                |
```
bouwkast, line 38: I see that there is a `Zend Framework` V2 https://docs.datadoghq.com/tracing/trace_collection/compatibility/php
But that is `Generic web tracing` would that mean that it is supported via HTTP call instrumentation?

tests/Integrations/Drupal/V10_1/CommonScenariosTest.php
```diff
@@ -15,4 +15,9 @@ protected static function getEnvs()
     {
         return array_merge(parent::getEnvs(), ['DD_SERVICE' => 'test_drupal_101']);
     }
+
+    protected static function getTestedVersion($testedLibrary)
+    {
+        return '10.1.3';
```
bouwkast, line 21: Question on these hardcoded versions, do those need to be updated manually?
If the `composer.json` has a greater version it'll be ignored and this `getTestedVersion` is returned?

Wondering if there is potential here in the future to try and bump these automatically.

tooling/tested_versions/create_or_update_supported_versions_pr.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+set -euox pipefail
+
+if [[ -z $(git status --porcelain) ]]; then
+  echo "No changes detected, exiting."
+  exit 0
+fi
+
+git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+git config --global user.name "github-actions[bot]"
```
bouwkast, line 10: Is it okay to use `--global` here? Just don't want to accidentally override some other script 😅

bouwkast (approved):
Nicely done and thanks for the PR description!

I think we can refine the generated table a bit at a later date as there seem to be some tricky integrations to report here.

integration_versions.md
PROFeNoM: The PHP Tracer will automatically generate a `web.request` root span for all incoming HTTP requests. Without further instrumentation, this is what is being called "Generic Web Tracing" for the PHP Tracer, meaning that transactions will be traced, as well as hypothetical Database calls (for instance).

On the other hand, "Framework-Level Instrumentation" offers greater information granularity based on the framework's specifics. For instance, for a MVC based framework, this would mean at least tracing the controller.

In the case of [ZF2](https://github.com/zendframework/zendframework/tree/release-2.4), we don't have a specific instrumentation. However, I believe that since it used to be a common PHP framework, it was included in the documentation so as to indicate that we provide some sort of support for it.

Does that answer your question? 🤔

tests/Integrations/Drupal/V10_1/CommonScenariosTest.php
PROFeNoM: For some tests, the version was hardcoded because it cannot be retrieved from the `composer.json`. This generally happens when we basically copy-paste the frameworks' repository.

For Drupal, there were three options:
1. We remove the current Drupal framework and start from scratch so that we can use composer.
2. We do a script that will go and retrieve the value of a constant that indicates which version is being used somewhere in the framework files
3. We hardcode it

The first approach is the most elegant and future-proof for Drupal, but also the most time-consuming. In other words, I don't think that it _needs_ to be in this PR. The first approach would also allow auto-bumping. The second approach is fundamentally a waste of time because the third option exists.

I believe that for Drupal, we should at one point pivot and rewrite the tests so that it can be auto-bumped.
However, for some other frameworks (e.g., WordPress)... I believe that it really has to be done manually 😬

tooling/tested_versions/create_or_update_supported_versions_pr.sh
PROFeNoM: I added these in [this commit](https://github.com/DataDog/dd-trace-php/pull/3062/commits/e9d7a17d0fe25b43b4ff6c6a33ff1fad7d8118d1). I made it because the following [error](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/19175/workflows/7c11417e-ff5f-44a4-932d-951d70716622/jobs/6057800?invite=true#step-106-145_31) happened in the [commit](https://github.com/DataDog/dd-trace-php/pull/3062/commits/e0640609e0e5438bcc26e03fdbd80f7b767c84fb) before:

```
Branch does not exist, creating it...
M       aggregated_tested_versions.json
M       integration_versions.md
Switched to a new branch 'update-supported-versions'

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: unable to auto-detect email address (got 'circleci@0be65266d252.(none)')

Exited with code exit status 128
```

I believe that this git config should be local to the job and shouldn't impact other jobs. Perhaps I can try without --global to see if it works to be on the safe side :+1:

bouwkast: All good then

tests/Integrations/Drupal/V10_1/CommonScenariosTest.php
bouwkast: Ahhh okay gotcha thanks for the explanation!
Sounds like we're all good then here for now.
I would agree that (2) isn't a huge win, I'm unsure of the effort required to rewrite tests, but that definitely doesn't need to be done at the moment and isn't a priority if the tests are functional so having it be done manually seems okay.

integration_versions.md
bouwkast: I'll bring that question up to figure out a path, but nothing to do at the moment unless you have ideas 😅

integration_versions.md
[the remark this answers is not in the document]
PROFeNoM: I had forgotten adding curl. My bad. However for zf2, it would indeed require manual effort.

PR #3066, opened 2025-02-03
fix: `ObjectKVStore::put` on throwing `__set`

PROFeNoM:
### Motivation

In `predis/predis`, starting v2.1.2 w/ https://github.com/predis/predis/pull/1049, it is no longer possible to use [ObjectKVStore::put](https://github.com/DataDog/dd-trace-php/blob/c11372f59b0be0dcd1fe9042ad83ec8080e40d6c/src/DDTrace/Util/ObjectKVStore.php#L140) in versions 7.4 and below since [Predis\Client::__set will be throwing an exception](https://github.com/predis/predis/blob/52489ea703aeb182b7ee7981eac69655d8f084ec/src/Client.php#L362-L365).

Currently, PredisIntegration is throwing an exception on versions 2.1.2+, using a PHP runtime that is strictly less than 7.4. Current estimates are reporting that [~5% of orgs](https://docs.google.com/spreadsheets/d/1JaQIu-UwshIep_mH5SIyYRfyE_dA49SumTQSbtwZ3VM/edit?usp=sharing) using the predis integration are impacted.

### Description

Fallback to [spl_object_hash](https://github.com/php/php-src/blob/caf5e8a1673392d08f58a8c89a5b257a9a6387ac/ext/spl/php_spl.c#L670-L673) when it is not possible to use the weakmap nor `__set`.

This should allow to proceed with https://github.com/DataDog/dd-trace-php/pull/3065.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/Util/ObjectKVStore.php
+++ b/src/DDTrace/Util/ObjectKVStore.php
@@ -110,6 +110,8 @@ class ObjectKVStore
 
     private static $weakmap;
 
+    private static $kv = [];
+
     /**
      * Put or replaces a key with a specific value.
      *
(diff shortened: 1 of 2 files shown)


bwoebi (changes requested):
Let's not do that. This leaks memory and is never cleaned up.
(And catching an exception is also relatively expensive)
We just need another object to attach our data to, and I think `$client->getConnection()` is perfectly suited for that.

PR #3075, opened 2025-02-06
fix(prof): avoid dlclose if threads did not join

morrisonlevi:
### Description

Null out the extension's handle if the ddprof_upload or ddprof_time threads timeout instead of joining. If the PHP engine dlclose's the handle and the shared object is unloaded while another thread is running, we've hit undefined behavior. This also probably results in a crash (on platforms that unload instead of no-op).

This may be the source of a crash when php-fpm does a log rotate. When doing the rotate, php-fpm shuts down all workers. If a worker is slow to process an upload and the timeout is hit, then we could hit this issue.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -702,7 +702,7 @@ dependencies = [
  "serde_json",
  "serde_repr",
  "serde_urlencoded",
- "thiserror",
+ "thiserror 1.0.69",
  "tokio",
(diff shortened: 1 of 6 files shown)


realFlowControl (approved):
Thanks @morrisonlevi ❤️

PR #3079, opened 2025-02-11
feat(appsec): add auto_user_instrum remote config capabilities

Leiyks:
### Description

Enable remote configs related to the `auto_user_instrumentation mode` according to the following [RFC](https://docs.google.com/document/d/19VHLdJLVFwRb_JrE87fmlIM5CL5LdOBv4AmLxgdo9qI/edit?pli=1&tab=t.0#heading=h.hidcz5iyug2k).

The following changes have been implemented:
- Helper: Add a new config aggregator for `ASM_FEATURES` to allow single file parsing in the listener.
- Helper/Extension: Add the new `settings` field in the requests_(init | exec | shutdown) responses to communicate parameters.
- Extension: Parsing methods for the additional `settings` field and new mechanism to get the auto_user_instrumentation mode. 

### Reviewer checklist
- [ ] Test coverage seems ok.

Related Jiras: [APPSEC-53750] 

[APPSEC-53750]: https://datadoghq.atlassian.net/browse/APPSEC-53750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appsec/src/extension/commands_helpers.c
```diff
@@ -553,6 +559,41 @@ static void _set_appsec_span_data(mpack_node_t node)
...
+
+    size_t count = mpack_node_map_count(root);
+
+    for (size_t i = 0; i < count; i++) {
+        mpack_node_t key = mpack_node_map_key_at(root, i);
+        mpack_node_t value = mpack_node_map_value_at(root, i);
+
+        if (mpack_node_type(key) != mpack_type_str) {
+            mlog(dd_log_warning,
+                "Failed to process setting: invalid type for key");
```
estringana, line 576: Jus a nitpick, but since this function is user_collection related, I would add something to the log pointing to user_collection so it's easier to debug

appsec/src/extension/commands_helpers.h
```diff
@@ -36,10 +36,11 @@ dd_result dd_command_proc_resp_verd_span_data(mpack_node_t root,
...
 /* Common helpers */
+void dd_command_process_settings(mpack_node_t root);
 void dd_command_process_meta(mpack_node_t root, zend_object *nonnull span);
 bool dd_command_process_metrics(mpack_node_t root, zend_object *nonnull span);
 bool dd_command_process_telemetry_metrics(mpack_node_t root);
 dd_result dd_command_process_config_features(
     mpack_node_t root, ATTR_UNUSED void *nullable ctx);
 dd_result dd_command_process_config_features_unexpected(
-    mpack_node_t root, ATTR_UNUSED void *nullable ctx);
+    mpack_node_t root, ATTR_UNUSED void *nullable ctx);
```
estringana, line 46: Missing empty line

estringana: Actyally, you removed it 😅

appsec/src/extension/commands_helpers.h
```diff
@@ -36,10 +36,11 @@ dd_result dd_command_proc_resp_verd_span_data(mpack_node_t root,
     void *unspecnull ctx /* actually struct req_info* nonnull */);
 
 /* Common helpers */
+void dd_command_process_settings(mpack_node_t root);
```
estringana, line 39: Is this called from outside of this module? Otherwise it could be private to it

appsec/src/extension/commands_helpers.c
```diff
@@ -553,6 +559,41 @@ static void _set_appsec_span_data(mpack_node_t node)
...
+        }
+
+        const char *key_str = mpack_node_str(key);
+        const char *value_str = mpack_node_str(value);
+        size_t key_len = mpack_node_strlen(key);
+        size_t value_len = mpack_node_strlen(value);
+
+        if (dd_string_equals_lc(
+                key_str, key_len, ZEND_STRL("auto_user_instrum"))) {
+            dd_parse_user_collection_mode_rc(value_str, value_len);
```
estringana, line 592: Neither function `dd_parse_user_collection_mode_rc` or here there is logging when something unexpected happen. Could you please add some?

appsec/src/extension/attributes.h
```diff
@@ -6,6 +6,8 @@
 #ifndef DD_ATTRIBUTES_H
 #define DD_ATTRIBUTES_H
 
+#include "configuration.h"
```
estringana, line 9: Is this required?

Leiyks: When attributes in included in a file without configuration.h included first it fails in the `build_packages` CI for some specific version. So it may be better to include it direcly in it to avoid issues, wdyt ?

estringana: if without it fails, then it's required. Not sure if @Anilm3 has any suggestion

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -52,4 +65,42 @@ void dds::remote_config::asm_features_listener::on_update(const config &config)
         throw error_applying_config(
             "Invalid config json encoded contents: enabled key invalid");
     }
+
+    if (!json_helper::field_exists(ruleset_, "auto_user_instrum")) {
+        service_config_->set_auto_user_instrum(
+            auto_user_instrum_mode::UNDEFINED);
+        return;
```
estringana, line 72: In order to be able to debug if something goes wrong, either exceptions or logging here would be helpful. Here and in the coming returns

appsec/src/extension/commands_helpers.c
Leiyks: The `settings` field is meant for generic use, currently it's only used for user_collection but it could perfectly used for other purposes if we need to.

I can modify the log, but we need to change it once the `settings` field is used for something else

appsec/src/extension/commands_helpers.h
Leiyks: I fixed the multiple missing empty lines, my bad :sweat_smile:

appsec/src/extension/commands_helpers.h
Leiyks: right, fixed it :+1:

appsec/src/helper/remote_config/listeners/config_aggregators/asm_features_aggregator.cpp
```diff
@@ -0,0 +1,64 @@
...
+            if (!value.IsObject()) {
+                throw error_applying_config(
+                    "Invalid type for " + std::string(key));
+            }
+
+            available_keys.emplace_back(key);
+        }
+    }
+
+    // All keys should be available so no need for extra checks
```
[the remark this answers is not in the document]
Leiyks: I think you misunderstood, it's because the available_keys vector is filled with found keys when they are found and with the correct type, so it only to say that we don't need to check the keys type again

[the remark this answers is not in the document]
Leiyks: I'll make the comment clearer

appsec/src/extension/commands_helpers.c
Leiyks: Changed the log for now :+1:

appsec/src/extension/attributes.h
Anilm3: I'm not sure about the issue described by @Leiyks, but the problem locally seems to be due to the introduction of `zai_str` into the `user_tracking.(c|h)` file, while simultaneously having removed `#include "configuration.h"`. Just add that header back to `user_tracking.h` and remove it from here.

Anilm3 (changes requested):
Some comments

appsec/src/extension/commands_helpers.c
```diff
@@ -553,6 +561,49 @@ static void _set_appsec_span_data(mpack_node_t node)
...
+        } else {
+            if (!get_global_DD_APPSEC_TESTING()) {
+                mlog(dd_log_warning,
+                    "Failed to process user collection setting: "
+                    "unknown key %.*s",
+                    (int)key_len, key_str);
+            }
+            continue;
```
Anilm3, line 602: The continue is unnecessary, just let it fall through.

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -7,22 +7,31 @@
...
+namespace dds::remote_config {
+
+void asm_features_listener::init()
 {
-    rapidjson::Document serialized_doc;
+    ruleset_ = rapidjson::Document(rapidjson::kObjectType);
+    aggregator_ = std::make_unique<asm_features_aggregator>();
+    aggregator_->init(&ruleset_.GetAllocator());
+    service_config_->unset_asm();
+    service_config_->unset_auto_user_instrum();
```
Anilm3, line 21: Thinking about this a bit more, perhaps it would be better to handle the absence of keys at the commit stage.

The reason is that this change may result in extensions temporarily disabled as the configurations are parsed.

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -7,29 +7,38 @@
...
+void asm_features_listener::commit()
+{
+    aggregator_->aggregate(ruleset_);
 
-    auto asm_itr = json_helper::get_field_of_type(
-        serialized_doc, "asm", rapidjson::kObjectType);
+    auto asm_itr =
+        json_helper::get_field_of_type(ruleset_, "asm", rapidjson::kObjectType);
     if (!asm_itr) {
+        service_config_->unset_asm();
```
Anilm3, line 34: Why throw here? I don't think the config is mandatory.

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -7,29 +7,38 @@
...
         throw error_applying_config("Invalid config json encoded contents: "
                                     "asm key missing or invalid");
     }
 
     auto enabled_itr = asm_itr.value()->value.FindMember("enabled");
     if (enabled_itr == asm_itr.value()->value.MemberEnd()) {
+        service_config_->unset_asm();
```
Anilm3, line 41: Same here, no need to throw in my opinion.

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
Leiyks: fixed it

appsec/src/helper/remote_config/listeners/asm_features_listener.cpp
```diff
@@ -49,7 +54,54 @@ void dds::remote_config::asm_features_listener::on_update(const config &config)
         // when appsec should not be enabled
         service_config_->disable_asm();
     } else {
-        throw error_applying_config(
-            "Invalid config json encoded contents: enabled key invalid");
+        service_config_->unset_asm();
+        return;
```
Anilm3, line 58: This return is unnecessary.

Anilm3 (approved):
lgtm

PR #3080, opened 2025-02-11
Implement option to Disable APM Tracing

estringana:
### Description

Implement `Disable APM Tracing`. This piece of work is a continuation of the previous #2903 done on previous quarter. This could be seen as a v2 of it.

[APPSEC-56460]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-56460]: https://datadoghq.atlassian.net/browse/APPSEC-56460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ext/trace_source.c
```diff
@@ -0,0 +1,67 @@
...
+        return NULL;
+    }
+    zend_string *encoded = zend_string_alloc(2, 0);
+    snprintf(ZSTR_VAL(encoded), 3, "%02x" PRIx64, DDTRACE_G(trace_source_bm));
+    return encoded;
+}
+
+bool ddtrace_trace_source_set_from_hexadecimal(zend_string *hexadecimal)
+{
+    if (!hexadecimal || ZSTR_LEN(hexadecimal) != 2) {
```
bwoebi, line 38: This violates the `MUST allow for parsing masks of at least 32 bits to ensure forward compatibility` requirement.

ext/tracer_tag_propagation/tracer_tag_propagation.c
```diff
@@ -44,9 +45,17 @@ void ddtrace_add_tracer_tags_from_header(zend_string *headerstr, zend_array *roo
             if (ZSTR_LEN(tag_name) >= sizeof("_dd.p.") &&
                 strncmp(ZSTR_VAL(tag_name), "_dd.p.", sizeof("_dd.p.") - 1) == 0) {
                 zval zv;
+                bool add = true;
                 ZVAL_STRINGL(&zv, valuestart, header - valuestart);
-                zend_hash_update(root_meta, tag_name, &zv);
-                zend_hash_add_empty_element(propagated_tags, tag_name);
+                if (strncmp(ZSTR_VAL(tag_name), DD_P_TS_KEY, sizeof(DD_P_TS_KEY) - 1) == 0) {
```
bwoebi, line 50: Use `zend_string_equals_cstr()` instead of strcmp

ext/ddtrace.h
```diff
@@ -147,6 +147,7 @@ ZEND_BEGIN_MODULE_GLOBALS(ddtrace)
     HashTable telemetry_spans_created_per_integration;
     ddog_SidecarActionsBuffer *telemetry_buffer;
 
+    uint8_t trace_source_bm;
```
bwoebi, line 150: This should not be maintained on the globals, but it's part of specific traces. Each trace can have its indvidual trace source.

estringana: I didn't add it to the root span until the last possible moment. The reason is that on the rootSpan it's a string and for performance purposes, I want to keep it as integer to avoid doing conversions back and forth. Does it makes sense? Not sure if you meant that

bwoebi: The problem is, that multiple traces may exist in parallel. However if _any_ trace has an asm event, it gets set on the trace and no further trace (within the same RINIT-lifecycle) will be applied the asm trace source.

ext/tracer_tag_propagation/tracer_tag_propagation.c
estringana: I changed for `zend_string_equals_literal`

bwoebi (changes requested):
I would recommend _not_ keeping the trace source as an integer, but maintain it directly as a string on the root meta.
It generally does not need to be read, except for updating.

A repeated strtol is not expensive, and it simplifies the API too, all we need is:
```
ddtrace_add_trace_source(ddtrace_root_span_data *root, int source);
```
This avoids all usage of a global, as well as flawlessly works with distributed tracing without special handling in injection nor extraction.
This strategy works well for the _dd.p.dm, and should work well for _dd.p.ts too.

ext/tracer_tag_propagation/tracer_tag_propagation.c
```diff
@@ -45,8 +46,13 @@ void ddtrace_add_tracer_tags_from_header(zend_string *headerstr, zend_array *roo
                 strncmp(ZSTR_VAL(tag_name), "_dd.p.", sizeof("_dd.p.") - 1) == 0) {
                 zval zv;
                 ZVAL_STRINGL(&zv, valuestart, header - valuestart);
-                zend_hash_update(root_meta, tag_name, &zv);
-                zend_hash_add_empty_element(propagated_tags, tag_name);
+                if (zend_string_equals_literal(tag_name, DD_P_TS_KEY)) {
+                    ddtrace_trace_source_set_from_hexadecimal(Z_STR_P(&zv), root_meta);
+                    zval_ptr_dtor(&zv);
+                } else {
```
bwoebi, line 52: This is now redundant and can be removed.

estringana: Why? The incoming `_dd.p.ts` needs to still be properly parsed and populated accordingly when `trace source` is ASM

bwoebi: Why do we need to parse it? The else branch takes care of populating. The only thing this essentially does is sanitization, which we don't really need here.

ext/handlers_http.h
```diff
@@ -136,12 +137,13 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
     zend_long sampling_priority = ddtrace_fetch_priority_sampling_from_root();
-    if (get_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED() && DDTRACE_G(asm_event_emitted) == true) {
+    if (!get_DD_APM_TRACING_ENABLED() && ddtrace_asm_event_emitted()) {
         sampling_priority = PRIORITY_SAMPLING_USER_KEEP;
     }
 
-    if (get_DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED() && DDTRACE_G(asm_event_emitted) == false &&
-        ddtrace_propagated_tags_get_tag(DD_TAG_P_APPSEC) == NULL) {
+    ddtrace_root_span_data *root_span = DDTRACE_G(active_stack) ? DDTRACE_G(active_stack)->root_span : NULL;
+    zend_array *meta = root_span ? ddtrace_property_array(&root_span->property_meta) : NULL;
```
bwoebi, line 145:
```suggestion
    zend_array *meta = root_span ? ddtrace_property_array(&root_span->property_meta) : &DDTRACE_G(root_span_tags_preset);
```
Otherwise propagation does not work when no trace was created yet.

estringana: 👏 thanks

estringana:
@bwoebi PR is ready with all changes spoken

tests/ext/background-sender/agent_sampling-standalone-asm_02.php
bwoebi, line 1: I suppose you committed that by accident?

estringana: :O

estringana: Deleted

bwoebi (approved):
Looks good to me now :-)

Anilm3 (approved):

PR #3089, opened 2025-02-14
fix(trace): crash when disabled by webserver config

morrisonlevi:
### Description

I found this looking through crash reports. Bob surmised the issue when I shared it with him.

We haven't tested this theory yet. I'm pushing this PR so CI can package it up nicely for us, so we have installable artifacts to try to A/B test.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
In any case, this PR is not doing any harm, even if it doesn't end up fixing the issue, just moving the condition to the start as well.
I thus merge this.

PR #3094, opened 2025-02-18
feat(prof): add 'tracing' features

morrisonlevi:
### Description

This adds the `tracing` and `tracing-subscriber` features to the profiler. Enabling `tracing-subscriber` requires `tracing`. These are not meant to be enabled for customer builds at this time. The goal is to use this data to enable benchmarking and other useful tools during development. Currently it's turned configured by the `RUST_LOG` env var.

We may want to move some of our logging of the PHP lifecycle to use this in the future.

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/lib.rs
```diff
@@ -240,6 +241,36 @@ extern "C" fn minit(_type: c_int, module_number: c_int) -> ZendResult {
...
+            let result = unsafe { libc::dup(libc::STDERR_FILENO) };
+            if result != -1 {
+                break result;
+            } else {
+                let error = std::io::Error::last_os_error();
+                if error.kind() != std::io::ErrorKind::Interrupted {
+                    error!("failed duplicating stderr to create tracing subscriber: {error}");
+                    return ZendResult::Failure;
+                }
+            }
```
realFlowControl, line 261: note: Could in theory run infinite, but as this is only meant to be used for dev and benchmarking, I do not care. Just in case we plan to release this to customer builds, we should define an exit condition of the loop.

realFlowControl (approved):
LGTM 🚀

profiling/src/lib.rs
morrisonlevi: It should only loop forever if the syscalls keep returning `Interrupted`, right?

realFlowControl: Yeah, which is highly unlikely, that's why I approved anyway.

PR #3095, opened 2025-02-19
feat: API Gateway Tracing

PROFeNoM:
### Description

This PR adds the possibility to create a new span on top of the existing request span based on received headers. This feature is opt-in through `DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED`.

The newly created span (the _"inferred span"_) will be a root span. The _original request span_ will also be a root span. The latter will be child of the former.

Runtime information (e.g., memory peak) will be set on the inferred span. It won't be set on both spans.

While the inferred span being the root span of the stack, it won't be returned by using `DDTrace\root_span`. Instead, the child will be returned to preserve the current behavior. The inferred span can still be accessed by using the parent property.

Errors set on the original request span will be set on the inferred span as well.

Implementation Doc: [link](https://docs.google.com/document/d/16sBdSPQexM3FVYzB4CP0282AjoH6kEGT-1Z6tXWYqIY/edit?tab=t.chr2gtt4x0e3#heading=h.3goslevnl8au)

Telemetry PR: https://github.com/DataDog/dd-go/pull/171137

image

### Reviewer checklist
- [ ] Test coverage seems ok.


wantsui (changes requested):

tests/Integrations/Laravel/APIGWTest.php
```diff
@@ -0,0 +1,69 @@
...
+                        'x-dd-proxy-httpmethod: GET',
+                        'x-dd-proxy-domain-name: example.com',
+                        'x-dd-proxy-stage: aws-prod',
+                    ]
+                )
+            );
+        });
+    }
+
+    public function testInferredProxyException()
```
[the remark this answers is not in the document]
PROFeNoM: Would the following phpt test be sufficient: [tests/ext/inferred_proxy/distributed_tracing.phpt](https://github.com/DataDog/dd-trace-php/pull/3095/files#diff-a637dda6424f8c2212dd896c3742b28dd49c2f3b44f89fbb34c3db2529b9d948) ?

Doing the same kind of test in our web test could be quite annoying. This is because they are 1. sort of "already distributed" due to the test setup, and 2. the trace_id is normalized to 0 in the JSONs. For instance, for the distributed aspect, this can be reflected by the presence of a parent_id on all the snapshot root spans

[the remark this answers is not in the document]
wantsui: oh yes! resolving this, the test you have with the headers and the root span looks good 👍

tests/Integrations/Laravel/APIGWTest.php
```diff
@@ -0,0 +1,69 @@
...
+
+    public function testInferredProxy()
+    {
+        $this->tracesFromWebRequestSnapshot(function () {
+            $this->call(
+                GetSpec::create(
+                    'A simple GET request returning a string',
+                    '/simple?key=value&pwd=should_redact',
+                    [
+                        'x-dd-proxy: aws.apigateway',
```
zarirhamza, line 38: The way we implemented it in other languages was to have a map w/the header value of `aws-apigateway` to map to a span name value of `aws.apigateway`. We want the header value to remain unchanged and we want the span name to be slightly different which can be represented by a key value map.

That way we can add more values in the future if we have more proxy headers that correspond to different operation names

zarirhamza (changes requested):

tests/Integrations/Laravel/APIGWTest.php
zarirhamza: Refer to this as an [example](https://github.com/DataDog/dd-trace-go/blob/fb51babc4beed2e0037c41df97848313e7e43af3/contrib/internal/httptrace/inferred_proxy.go#L74)

ext/ddtrace.c
```diff
@@ -374,20 +375,40 @@ bool ddtrace_alter_sampling_rules_file_config(zval *old_value, zval *new_value,
...
 
-static inline void dd_alter_prop(size_t prop_offset, zval *old_value, zval *new_value, zend_string *new_str) {
+static inline void dd_alter_prop_common(size_t prop_offset, zval *old_value, zval *new_value, zend_string *new_str, bool skip_inferred) {
     UNUSED(old_value, new_str);
 
     ddtrace_span_properties *pspan = ddtrace_active_span_props();
     while (pspan) {
+        if (skip_inferred) {
+            ddtrace_span_data *span = SPANDATA(pspan);
+            if (span && span->type == DDTRACE_INFERRED_SPAN) {
```
bwoebi, line 385:
```suggestion
            if (span->type == DDTRACE_INFERRED_SPAN) {
```
Nit, but SPANDATA() is just an offset calculation from pspan and will never be NULL.

ext/inferred_proxy_headers.c
```diff
@@ -0,0 +1,45 @@
...
+    ddtrace_add_proxy_info("aws-apigateway", "aws.apigateway", "aws-apigateway");
+
+    // Add more proxies using ddtrace_add_proxy_info
+}
+
+void ddtrace_free_proxy_info_map(void) {
+    ddtrace_proxy_info *info;
+    ZEND_HASH_FOREACH_PTR(&proxy_info_map, info) {
+        efree(info);
+    } ZEND_HASH_FOREACH_END();
```
bwoebi, line 26: Nit: the idiomatic way would be using:
```
static dd_proxy_info_dtor(zval *zv) {
    efree(Z_PTR_P(zv));
}
```
and passing it to zend_hash_init as destructor.

ext/inferred_proxy_headers.c
```diff
@@ -0,0 +1,45 @@
...
+
+void ddtrace_add_proxy_info(const char *system, const char *span_name, const char *component) {
+    ddtrace_proxy_info *info = emalloc(sizeof(ddtrace_proxy_info));
+    info->span_name = span_name;
+    info->component = component;
+    zend_hash_str_add_ptr(&proxy_info_map, system, strlen(system), info);
+}
+
+void ddtrace_init_proxy_info_map(void) {
+    zend_hash_init(&proxy_info_map, 8, NULL, NULL, 1);
```
bwoebi, line 15: This should be 0 as last parameter right? persistent=1 is for memory supposed to survive the request cycle.

PROFeNoM: Oops, indeed. I initially wanted to do the initialization/free in MINIT/MSHUTDOWN, but faced issues. Nice catch, ty.

bwoebi: this should be in minit/mshutdown though. No point in reinitializing that every request.

ext/serializer.c
```diff
@@ -1180,6 +1188,17 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *serv
...
+        ddtrace_root_span_data *root = ROOTSPANDATA(&span->std);
+        if (root->child_root) { // Can't check for DDTRACE_INFERRED_SPAN because the span is now closed
+            zval *child_exception_zv = &root->child_root->span.property_exception;
+            has_exception = Z_TYPE_P(child_exception_zv) == IS_OBJECT && instanceof_function(Z_OBJCE_P(child_exception_zv), zend_ce_throwable);
+            if (has_exception) {
+                ignore_error = false;
+                enum dd_exception exception_type = Z_PROP_FLAG_P(child_exception_zv) == 2 ? DD_EXCEPTION_CAUGHT : DD_EXCEPTION_UNCAUGHT;
+                ddtrace_exception_to_meta(Z_OBJ_P(child_exception_zv), service_name, span->start, meta, dd_add_meta_array, exception_type);
+            }
+        }
```
bwoebi, line 1201: What is that supposed to do?
If we have an inferred child root span... copy its exception to the current root span?
Why? What for?

ext/serializer.c
```diff
@@ -1180,6 +1188,17 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *serv
             exception_type = Z_PROP_FLAG_P(exception_zv) == 2 ? DD_EXCEPTION_CAUGHT : DD_EXCEPTION_UNCAUGHT;
         }
         ddtrace_exception_to_meta(Z_OBJ_P(exception_zv), service_name, span->start, meta, dd_add_meta_array, exception_type);
+    } else if (span->std.ce == ddtrace_ce_root_span_data) {
+        ddtrace_root_span_data *root = ROOTSPANDATA(&span->std);
+        if (root->child_root) { // Can't check for DDTRACE_INFERRED_SPAN because the span is now closed
+            zval *child_exception_zv = &root->child_root->span.property_exception;
+            has_exception = Z_TYPE_P(child_exception_zv) == IS_OBJECT && instanceof_function(Z_OBJCE_P(child_exception_zv), zend_ce_throwable);
```
bwoebi, line 1195: Can you move this before the if (has_exception) above?
```
// inherit exception from child root
if (!has_exception && span->std.ce == ddtrace_ce_root_span_data) {
        ddtrace_root_span_data *root = ROOTSPANDATA(&span->std);
        if (root->child_root) { // Can't check for DDTRACE_INFERRED_SPAN because the span is now closed
            exception_zv = &root->child_root->span.property_exception;
            has_exception = Z_TYPE_P(child_exception_zv) == IS_OBJECT && instanceof_function(Z_OBJCE_P(child_exception_zv), zend_ce_throwable);
    }
}
```
The duplication was confusing me

ext/span.h
```diff
@@ -86,6 +87,7 @@ struct ddtrace_span_data {
     bool notify_user_req_end;
     struct ddtrace_span_data *next;
     struct ddtrace_root_span_data *root;
+    bool is_child_of_inferred_span;
```
bwoebi, line 90: Why do we have this property at all? Isn't it redundant with child_root != NULL on the root_span_data?

bwoebi: I see only one usage of it, and it's trivially replaced.

ext/span.h
```diff
@@ -107,6 +109,7 @@ struct ddtrace_root_span_data {
     ddtrace_rule_result sampling_rule;
     bool explicit_sampling_priority;
     enum ddtrace_trace_limited trace_is_limited;
+    struct ddtrace_root_span_data *child_root; // Only used when inferring proxy services (type: DDTRACE_INFERRED_SPAN)
```
bwoebi, line 112: let's rename it to `inferred_root`

ext/span.c
```diff
@@ -796,3 +823,75 @@ zend_string *ddtrace_trace_id_as_hex_string(ddtrace_trace_id id) {
...
+        if (result.system) zend_string_release(result.system);
+        if (result.start_time_ms) zend_string_release(result.start_time_ms);
+        return NULL;
+    }
+
+    const ddtrace_proxy_info *proxy_info = ddtrace_get_proxy_info(result.system);
+    if (!proxy_info) {
+        zend_string_release(result.system);
+        zend_string_release(result.start_time_ms);
+        return NULL;
```
bwoebi, line 842: You're leaking the other strings in result here, there are more than two.

ext/inferred_proxy_headers.c
```diff
@@ -0,0 +1,41 @@
...
+
+ddtrace_inferred_proxy_result ddtrace_read_inferred_proxy_headers(ddtrace_read_header *read_header, void *data) {
+    ddtrace_inferred_proxy_result result = {0};
+
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY"), "x-dd-proxy", &result.system, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_REQUEST_TIME_MS"), "x-dd-proxy-request-time-ms", &result.start_time_ms, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_PATH"), "x-dd-proxy-path", &result.path, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_HTTPMETHOD"), "x-dd-proxy-httpmethod", &result.http_method, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_DOMAIN_NAME"), "x-dd-proxy-domain-name", &result.domain, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_STAGE"), "x-dd-proxy-stage", &result.stage, data);
```
bwoebi, line 34: you should probably put this in an `if (result.system || result.path)` to short-circuit this away.

ext/span.c
```diff
@@ -796,3 +823,75 @@ zend_string *ddtrace_trace_id_as_hex_string(ddtrace_trace_id id) {
...
+
+    zval_ptr_dtor(&span->property_name);
+    ZVAL_STR(&span->property_name, zend_string_init(proxy_info->span_name, strlen(proxy_info->span_name), 0));
+
+    zval_ptr_dtor(&span->property_resource);
+    if (result.http_method && result.path) {
+        ZVAL_STR(&span->property_resource, strpprintf(0, "%s %s", ZSTR_VAL(result.http_method), ZSTR_VAL(result.path)));
+    }
+
+    span->start = (zend_long)zend_strtod(ZSTR_VAL(result.start_time_ms), NULL) * 1000000;
```
bwoebi, line 855: This looks fishy. Either you should directly parse it as a long, or you need parenthesis (if the number is actually a decimal number). The (zend_long) cast is evaluated before the multiplication.
If the goal is string to zend_long, use `ZEND_ATOL()`.

ext/span.c
bwoebi: Maybe you should just `goto cleanup;` with the cleanup label at the end of the function here.

bwoebi:
How is this supposed to interact with sampling decisions and distributed tracing?
Currently it's going to use the api gateway root span for sampling decisions (e.g. DD_TRACE_SAMPLING_RULES is going to match against the inferred root rather the "normal" root). Is that intentional? Also, it's two root spans.
Same for distributed tracing, where are the propagated root tags to be taken from?

Also: ddtrace_get_root_span() is the function appsec uses to attach sampling info and root tags to. Is that supposed to be the inferred span or the proper root?

ext/span.c
```diff
@@ -796,3 +823,78 @@ zend_string *ddtrace_trace_id_as_hex_string(ddtrace_trace_id id) {
...
+    }
+
+    span->start = ZEND_ATOL(ZSTR_VAL(result.start_time_ms)) * 1000000;
+    span->duration_start = zend_hrtime() - (ddtrace_nanoseconds_realtime() - span->start);
+
+    zval zv;
+
+    if (result.domain) {
+        ZVAL_STR_COPY(&zv, result.domain);
+        ddtrace_assign_variable(&span->property_service, &zv);
```
bwoebi, line 870: Is this `if (result.domain)` rather a safeguard or something which will regularly happen? Because if it can happen that the domain is not set, the inferred span will have the generic service name (http.request), while frameworks will have the service name from the root span. Which will certainly be unexpected.

ext/span.c
```diff
@@ -796,3 +823,78 @@ zend_string *ddtrace_trace_id_as_hex_string(ddtrace_trace_id id) {
     snprintf(ZSTR_VAL(str), 33, "%016" PRIx64 "%016" PRIx64, id.high, id.low);
     return str;
 }
+
+void free_inferred_proxy_result(ddtrace_inferred_proxy_result *result) {
```
bwoebi, line 827: nit: make static.

PROFeNoM:
If I refer to the [following document](https://docs.google.com/document/d/1J6lDrk_qcD8zvQ0gJdR8oRabyQdz4RSPNFhLhjpKji4/edit?tab=t.0), it would appear that it is expected for the sampling rules to be matched against the new, inferred root span.

> Is that intentional?

Considering that they noted that sampling rules should be reset "_with AWS API Gateway service name instead of the former root service’s service name_", I assume this is intentional.

---

About the other questions, I don't know. At a glance, I am more concerned about the appsec question, as this is an internal behavior that the customer won't have power upon at all. I'd have to ask to know whether a decision was already made.

bwoebi:
@PROFeNoM Could you please ask whether that's supposed to affect all metadata (like, "tags" or "resource"), or only the service name? Because if it's the former it will be impossible to define sampling rules according to runtime behaviour (i.e. "if this tag is set, then apply only this reduced sampling rate), essentially.

PROFeNoM:
Superseded by https://github.com/DataDog/dd-trace-php/pull/3116

Issue #3097, opened 2025-02-20
[Bug]: Service override not working with the Predis integration


redpanda:
### Bug report

Hi,

I set the environment variable `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` to `true` and discovered it does not work with the Predis integration; I'm still seeing the service redis integration.
The changes to the Predis integration were forgotten during the implementation https://github.com/DataDog/dd-trace-php/pull/2163. Do you know if it is by choice, or has it been missed? Thank you in advance.

### PHP version

8.3

### Tracer or profiler version

1.5.1

PROFeNoM:
Hey @redpanda :wave:

Thanks for your report. I believe it was simply forgotten at the time of the implementation. I'll open a PR to make it work :+1:

redpanda:
Awesome! Thanks for the quick fix. 🙏

PR #3098, opened 2025-02-20
fix: Missing Predis Service Override

PROFeNoM:
### Description

Fixes https://github.com/DataDog/dd-trace-php/issues/3097

I think Predis was simply forgotten at the time of the implementation, so here is a proposition.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Thank you for the quick fix!

PR #3104, opened 2025-02-24
fix(prof): `fork()` call handling before rinit

realFlowControl:
### Description

In case a `fork()` happens before the allocation profiler got initialised, the atfork handler in the child will try and shutdown the allocation profiler in the child and crash.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/allocation_ge84.rs
+++ b/profiling/src/allocation/allocation_ge84.rs
@@ -245,8 +245,12 @@ pub fn alloc_prof_rshutdown() {
         let mut custom_mm_gc: Option<zend::VmMmCustomGcFn> = None;
         let mut custom_mm_shutdown: Option<zend::VmMmCustomShutdownFn> = None;
 
-        // Safety: `unwrap()` is safe here, as `heap` is initialized in `MINIT`
+        if unsafe { (*zend_mm_state).heap.is_null() } {
+            return;
+        }
+
         let heap = unsafe { (*zend_mm_state).heap };
+
         unsafe {
             zend::zend_mm_get_custom_handlers_ex(
                 heap,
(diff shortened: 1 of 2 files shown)


profiling/src/allocation/allocation_ge84.rs
```diff
@@ -245,8 +245,12 @@ pub fn alloc_prof_rshutdown() {
         let mut custom_mm_gc: Option<zend::VmMmCustomGcFn> = None;
         let mut custom_mm_shutdown: Option<zend::VmMmCustomShutdownFn> = None;
 
-        // Safety: `unwrap()` is safe here, as `heap` is initialized in `MINIT`
+        if unsafe { (*zend_mm_state).heap.is_null() } {
+            return;
+        }
+
         let heap = unsafe { (*zend_mm_state).heap };
```
bwoebi, line 252:
```suggestion
        let heap = unsafe { (*zend_mm_state).heap };
        if heap.is_null() {
            return;
        }
```
nicer?

profiling/src/allocation/allocation_le83.rs
```diff
@@ -181,7 +181,11 @@ pub fn alloc_prof_rshutdown() {
         let mut custom_mm_free: Option<zend::VmMmCustomFreeFn> = None;
         let mut custom_mm_realloc: Option<zend::VmMmCustomReallocFn> = None;
         // Safety: `unwrap()` is safe here, as `heap` is initialized in `RINIT`
```
bwoebi, line 183:
```suggestion
```

bwoebi (approved):
I'm not too happy about warn!() and trace!() calls here. (in the codeblock below each), as it's not signal safe (like in ZTS when you do fork+exec just to spawn a new process).

But the fix itself seems right.

realFlowControl:
> I'm not too happy about warn!() and trace!() calls here. (in the codeblock below each), as it's not signal safe (like in ZTS when you do fork+exec just to spawn a new process).

Yeah. There are some things that need cleanup. I have some ideas, I'll start a doc and note some of them and talk to Levi. For the time being, this should fix the issue at hand.

morrisonlevi (approved):
I adjusted for code style, added safety comments, and also documented why the thing can be null (although looking now I just realized I didn't copy that comment to the None case, meh).

PR #3112, opened 2025-03-02
[POC] Next-gen lambda extension coms

natepage:
### Description

This PR is only a proof-of-concept to experiment with the next-gen lambda extension communication issues.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
Hey @natepage,

I suppose this PR is for the `DD_EXTENSION | ERROR | Error processing traces: Missing Content-Length` error?
Does simply commenting this out resolve the issue for you?

(Sorry for the delayed reply, I thought this was a PR from within our organization.)

natepage:
@bwoebi hey! Yes you're right, I created this PR simply to be able to use the new extension created in CircleCI artifacts and see if it was that simple, truth is... I was so close and then something came up at work and I didn't get a chance to try it out... 😄 

I wasn't expecting any reply, so no stress, but I will make sure to get back to it asap and keep you posted, thank you for replying anyway!

natepage:
@bwoebi I've tried it out today, but it appears that I'm doing something wrong as I'm not seeing any difference in the way ddtrace behave so it looks like I'm not using my updated version.

I was under the impression that if I use the `datadog-setup.php` from the artifacts of CircleCI for this PR it would be using the updated logic but I can still see the `Transfer-Encoding: chunked` header being set but I've commented it out 😄 

Here is the URL I've used to download the artifact: `https://output.circle-artifacts.com/output/job/433594e6-fb83-417b-9d7f-96ddcdea78c2/artifacts/0/datadog-setup.php`

Perhaps it's not the right one, a little help would be appreciated to see if that simple change changes anything, thank you

bwoebi:
No, that's the absolutely right URL.
However curl defaults to chunked encoding unless it knows the body size ahead of time. This can, apparently, be set via `CURLOPT_INFILESIZE`.

natepage:
Ok this is not looking promising, I've managed to set the `Content-Length` header but it's not accurate so the trace agent from the AWS Lambda Extension is still complaining 😄 

Screenshot 2025-03-19 at 5 03 10 pm

natepage:
@bwoebi sorry me again, I'm really keen on finding a solution to that one so I wanted to keep experimenting,

But it appears that I'm doing something wrong, and I broke the CI, it can't checkout the code of the libdatadog submodule,
My thinking was to use an updated version of libdatadog to prefix the sidecar paths with `/tmp` so I:

- created a fork of [libdatadog](https://github.com/natepage/libdatadog)
- updated the logic there
- updated the .gitmodules file to point to my repo/branch
- ran `g submodule update --remote libdatadog` to bring my changes into ddtrace

But it looks like CircleCI is upset now 😄 
I'm not gonna lie I am way out of my comfort zone here, but once again very keen to help!

Maybe there is something obvious here and it's simply because I'm inexperienced with such setup, regardless it's taking me way longer than it should to go through the experiment that's for sure haha

Not sure if the help is wanted here, if you have an ongoing initiative for this internally, but the original issue is still labelled as "up-for-grabs" and it feels like we are very close to a working solution,

If yes, then would you or someone experienced in contributing to both ddtrace and libdatadog have time to chat to help me going? To accelerate the learning curve, it would be much appreciated!

bwoebi:
Yes, I would have time (next week, I'm off today and tomorrow).
Feel free to reach out to me in Datadog public slack (https://chat.datadoghq.com/) to communicate privately.

And yes, we're definitely appreciating help here (adding a content-length header is probably something I might be able to fix on myself; but the sidecar issue, I don't really understand what exactly even is not supported here / why it breaks in the first place).

PR #3115, opened 2025-03-04
Add a ratchet integration

bwoebi:
Implements server (cboden/ratchet) and clients (ratchet/pawl) of ratchet.

This is going to be the basis of further integrations in conjunction with event loops.

### DIFF
--- a/src/DDTrace/Integrations/Ratchet/RatchetIntegration.php
+++ b/src/DDTrace/Integrations/Ratchet/RatchetIntegration.php
@@ -268,10 +268,18 @@ class RatchetIntegration extends Integration
                     $span->meta[Tag::COMPONENT] = RatchetIntegration::NAME;
                     if ($rootTrace && \dd_trace_env_config("DD_TRACE_WEBSOCKET_MESSAGES_INHERIT_SAMPLING")) {
                         $span->samplingPriority = $rootSpan->samplingPriority;
-                        $span->meta["_dd.dm.inherited"] = 1;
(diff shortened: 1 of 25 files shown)


src/DDTrace/Integrations/Ratchet/RatchetIntegration.php
```diff
@@ -0,0 +1,435 @@
...
+                            return;
+                        }
+
+                        $rootSpan = ObjectKVStore::get($handler, "handshake_root", $handshake);
+                    }
+
+                    $message = $hook->args[0];
+                    if ($isClose && $message->getOpcode() !== Frame::OP_CLOSE) {
+                        return;
+                    }
```
PROFeNoM, line 258: Can you enlighten me on the reason behind this condition? Not saying by any means that it is wrong. Just trying to understand. Thanks!

bwoebi:
```
\DDTrace\install_hook($onControl, $hookFn(true), null, \DDTrace\HOOK_INSTANCE);
```
This hook - with $isClose === true:
$onControl is called for any control opcode, including pings and pongs for example. We are only interested in close frames.

bwoebi: I probably should rename this to `$isControl`, then it becomes more obvious.

PROFeNoM (approved):
I had totally forgotten to approve the PR, my bad! Nice work 😃

PR #3116, opened 2025-03-04
feat: API Gateway Tracing

PROFeNoM:
### Description

Based on the following implementation doc: [GDoc](https://docs.google.com/document/d/16sBdSPQexM3FVYzB4CP0282AjoH6kEGT-1Z6tXWYqIY/edit?tab=t.chr2gtt4x0e3#heading=h.3goslevnl8au)

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/ddtrace.c
```diff
@@ -2740,6 +2772,36 @@ PHP_FUNCTION(DDTrace_start_span) {
     dd_start_span(INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+/* {{{ proto InferredSpanData\null DDTrace\start_inferred_span(array $headers, SpanData $rootSpan) */
```
bwoebi, line 2775: I had the suggestion of getting rid of this function and instead putting the logic into consume_distributed_tracing_headers(). Did this not work out?

PROFeNoM: Oops, forgot about it. Will try it out :+1:

ext/span.c
```diff
@@ -225,6 +307,12 @@ ddtrace_span_data *ddtrace_open_span(enum ddtrace_span_dataype type) {
         LOG(SPAN_TRACE, "Starting new span: trace_id=%s, span_id=%" PRIu64 ", parent_id=%" PRIu64 ", SpanStack=%d", Z_STRVAL(span->root->property_trace_id), span->span_id, SPANDATA(span->parent)->span_id, span->stack->std.handle);
     }
 
+    if (get_DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED() && !DDTRACE_G(inferred_span_created)) {
```
bwoebi, line 310: What exactly is this flag supposed to solve? That nested traces don't get the inferred span data?

ext/span.c
```diff
@@ -548,6 +636,16 @@ void ddtrace_close_span(ddtrace_span_data *span) {
         ddtrace_switch_span_stack(span->stack);
     }
 
+    if (span->std.ce == ddtrace_ce_root_span_data) {
+        ddtrace_root_span_data *root = ROOTSPANDATA(&span->std);
+        zval *inferred_span_zv = &root->property_inferred_span;
+        if (Z_TYPE_P(inferred_span_zv) == IS_OBJECT) {
```
bwoebi, line 642: could you please add a condition that the inferred_span property is actually an instance of ddtrace_ce_inferred_span_data?

ext/span.c
PROFeNoM: I added this flag for this test case: [multiple local traces test](https://github.com/DataDog/dd-trace-php/blob/d2629ab7645eb44334da7b040676c61e050d2d74/tests/ext/inferred_proxy/multiple_traces.phpt)

The inferred span should only be created once.

Passed headers aren't really uniquely identifiable, or perhaps through their start time (but still, it's not unique; perhaps it's unique enough 🤷). This global variable was added to check whether an inferred span was already created within the request so that we don't create one again from $_SERVER headers. This shouldn't impact async frameworks (which are using `consume_distributed_tracing_headers`; so for the latter we can check whether an inferred span is already attached to the root span... if we consider this an internal behavior)

bwoebi: I think it's fine. Thanks for the response :-)

ext/serializer.c
```diff
@@ -1799,6 +1852,10 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         smart_str_free(&metrics_str);
     })
 
+    if (inferred_span) {
+        ddtrace_serialize_span_to_array(inferred_span, array);
```
bwoebi, line 1856: I would prefer if we could do all the transfer_data calls here (from serialized to inferred serialized), after the serializing step. But I won't insist.

ext/serializer.c
```diff
@@ -1808,6 +1788,28 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         }
     }
 
+    if (inferred_span) {
+        ddtrace_serialize_span_to_array(inferred_span, array);
```
bwoebi, line 1792: You could make `ddtrace_serialize_span_to_array` return the inserted array instead of refetching from `Z_ARR_P(array)`.

ext/serializer.c
```diff
@@ -1877,6 +1875,8 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
     }
 
     add_next_index_zval(array, el);
+
+    return el;
```
bwoebi, line 1879: noooo :-D
That returns a stack variable.

Use
```
return zend_hash_next_index_insert(Z_ARR_P(array), el);
```

PROFeNoM: Ok :-)

ext/span.c
```diff
@@ -150,6 +150,84 @@ uint64_t ddtrace_nanoseconds_realtime(void) {
...
+        zend_string_release(result->system);
+        zend_string_release(result->start_time_ms);
+        return NULL;
+    }
+
+    ddtrace_span_data *span = ddtrace_init_span(DDTRACE_INFERRED_SPAN, ddtrace_ce_inferred_span_data);
+    ZVAL_OBJ(&root->property_inferred_span, &span->std);
+
+    span->span_id = ddtrace_generate_span_id();
+    ddtrace_set_global_span_properties(span);
```
bwoebi, line 180: Move this to the end or e.g. `DD_TAGS=stage:1` will lead to a memory leak

ext/serializer.c
```diff
@@ -1117,9 +1117,15 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
 static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *service_name) {
     bool is_root_span = span->std.ce == ddtrace_ce_root_span_data;
-    zval meta_zv, *meta = &span->property_meta;
+    bool is_inferred_span = span->std.ce == ddtrace_ce_inferred_span_data;
+    zval meta_zv, *meta = &span->property_meta, *inferred_span_zv = NULL;
     bool ignore_error = false;
 
+    if (is_root_span) {
+        ddtrace_root_span_data *rsd = span->root;
+        inferred_span_zv = &rsd->property_inferred_span;
```
bwoebi, line 1126:
```suggestion
        inferred_span_zv = &span->root->property_inferred_span;
```
nit

ext/serializer.c
```diff
@@ -1117,9 +1117,15 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
 
 static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *service_name) {
     bool is_root_span = span->std.ce == ddtrace_ce_root_span_data;
-    zval meta_zv, *meta = &span->property_meta;
+    bool is_inferred_span = span->std.ce == ddtrace_ce_inferred_span_data;
+    zval meta_zv, *meta = &span->property_meta, *inferred_span_zv = NULL;
```
bwoebi, line 1121:
```suggestion
    zval meta_zv, *meta = &span->property_meta;
    ddtrace_span_data *inferred_span = NULL;
    if (is_root_span) {
        inferred_span = ddtrace_get_inferred_span(span->root);
    }
```
Let's check it consistently the same way.

ext/serializer.c
```diff
@@ -1756,6 +1788,26 @@ void ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
+        zend_array *serialized_metrics = Z_ARR(metrics_zv);
+
+        transfer_data(serialized_metrics, serialized_inferred_span_metrics, ZEND_STRL("_dd.agent_psr"), true);
+        transfer_data(serialized_metrics, serialized_inferred_span_metrics, ZEND_STRL("_dd.rule_psr"), true);
+        transfer_data(serialized_metrics, serialized_inferred_span_metrics, ZEND_STRL("_dd.limit_psr"), true);
+
+        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.message"), false);
+        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.type"), false);
+        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.stack"), false);
+        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.ignored"), false);
```
bwoebi, line 1806:
```suggestion
        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error"), false);
        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("track_error"), false);
```
I think `error.ignored` is not a thing after serialization, but `error` is.

ext/ddtrace.c
```diff
@@ -3079,70 +3111,91 @@ static bool dd_read_userspace_header(zai_str zai_header, const char *lowercase_h
...
-                _error_code = ZPP_ERROR_FAILURE;
-                break;
-            } else if (Z_TYPE_P(_arg) == IS_ARRAY) {
-                array = Z_ARR_P(_arg);
-                efree(_error);
-            } else {
-                _error_code = ZPP_ERROR_WRONG_CALLBACK;
-                break;
-            }
+            DD_PARAM_PROLOGUE(0, 0);
```
bwoebi, line 3120: what happened here with the indent?

ext/ddtrace.c
```diff
@@ -1785,6 +1816,7 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
     }
 
     ddtrace_clean_git_object();
+    DDTRACE_G(inferred_span_created) = false;
```
bwoebi, line 1819:
```suggestion
```
We already set it in init, that's enough

tests/snapshots/tests.integrations.roadrunner.apigw_test.test_inferred_proxy_exception.json
```diff
@@ -0,0 +1,56 @@
+[[
+  {
+    "name": "aws.apigateway",
+    "service": "example.com",
+    "resource": "GET /test",
+    "trace_id": 0,
+    "span_id": 1,
+    "parent_id": 3978455880403641907,
```
wantsui, line 8: Can this assert that `"error": 1,` for the aws api gateway span too?

tests/snapshots/tests.integrations.laravel.octane.apigw_test.test_inferred_proxy.json
```diff
@@ -0,0 +1,220 @@
+[[
+  {
+    "name": "aws.apigateway",
+    "service": "example.com",
+    "resource": "GET /test",
+    "trace_id": 0,
```
wantsui, line 6: Does PHP store the "start" time as field in the span attributes? If so, can it be asserted in this test that the start time is what you expect from 1739261376000?

ext/inferred_proxy_headers.c
```diff
@@ -0,0 +1,46 @@
...
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY"), "x-dd-proxy", &result.system, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_REQUEST_TIME_MS"), "x-dd-proxy-request-time-ms", &result.start_time_ms, data);
+
+    if (!result.system || !result.start_time_ms) {
+        return result;
+    }
+
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_PATH"), "x-dd-proxy-path", &result.path, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_HTTPMETHOD"), "x-dd-proxy-httpmethod", &result.http_method, data);
+    read_header((zai_str)ZAI_STRL("X_DD_PROXY_DOMAIN_NAME"), "x-dd-proxy-domain-name", &result.domain, data);
```
wantsui, line 38: What is the value of the service if `x-dd-proxy-domain-name` is not provided?

wantsui: (Asking because it's unclear to me from the test what happens)

bwoebi: If I read the code correctly, service then is never assigned and falls back to the empty string.

wantsui: If there's no fallback we should try to use DD_SERVICE where possible so we don't send the span with an empty service name.

tests/snapshots/tests.integrations.roadrunner.apigw_test.test_inferred_proxy_exception.json
PROFeNoM: That's a crazy catch. I see why.

tests/snapshots/tests.integrations.laravel.octane.apigw_test.test_inferred_proxy.json
PROFeNoM: The "start" field is removed from snapshots (since we do not compare it).
With commit [e75a791](https://github.com/DataDog/dd-trace-php/pull/3116/commits/e75a791318c1a6d8f322faea84564ca061a39835), I've slightly modified `basic_test.phpt` to also assert proper span duration. By "proper span duration", I mean within 0.01% of expected duration, which I believe to be fair.

ext/inferred_proxy_headers.c
PROFeNoM: Modified to fallback to the root span's service name. I can modify it to strictly be DD_SERVICE, but then we'd have to have some special handling when altering DD_SERVICE (+The service is sometimes inferred from the app config - e.g. Laravel).

ext/serializer.c
```diff
@@ -1802,8 +1802,8 @@ zval *ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.message"), false);
         transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.type"), false);
         transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error.stack"), false);
-        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("error"), false);
-        transfer_data(serialized_meta, serialized_inferred_span_meta, ZEND_STRL("track_error"), false);
+        transfer_data(Z_ARR_P(el), Z_ARR_P(serialized_inferred_span), ZEND_STRL("error"), false);
+        transfer_data(Z_ARR_P(el), Z_ARR_P(serialized_inferred_span), ZEND_STRL("track_error"), false);
```
bwoebi, line 1806: track_error is in meta though, right? Only array is top-level.

tests/snapshots/tests.integrations.laravel.octane.apigw_test.test_inferred_proxy.json
wantsui: sounds good, I'll mark this one as resolved ! :)

ext/serializer.c
PROFeNoM: Ok :+1: Will do in the next commit

PROFeNoM: Hummm. Thinking about it, the `track_error` cannot exist on a root span because the code that sets this value only runs when walking up the parent chain, which a root span doesn't have. I believe this tag has to be handled differently.

ext/inferred_proxy_headers.c
zarirhamza: LGTM

bwoebi (approved):

PR #3117, opened 2025-03-05
Implement API Security Sampling RFC

cataphract:
### Description

Implement API Security Sampling RFC. Removes previous sampling mechanism and its configuration setting.

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/helper/sampler.hpp
```diff
@@ -6,42 +6,285 @@
...
+                if (cur_data.last_accessed == cur_data.last_reported) {
+                    // then this one should not be a hit
+                    return false;
+                }
+
+                // the other thread did not register a hit
+                // we retry if our idea of time is ahead
+                if (cur_data.last_accessed < now) {
+                    goto retry;
+                }
```
Anilm3, line 167: I'm not so sure about this one, if you don't retry, you still return a hit (true), but you don't update...

cataphract: you're right, this is missing a return false

appsec/src/helper/sampler.hpp
```diff
@@ -6,42 +6,285 @@
...
+                // no space to add anything
+                table.size.fetch_add(-1);
+                return false;
+            }
+
+            std::uint64_t exp_number{};
+            if (!entry.key.compare_exchange_strong(
+                    exp_number, number, std::memory_order_relaxed)) {
+                if (exp_number == number) {
+                    // another thread inserted the same number
```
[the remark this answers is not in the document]
cataphract: yes, the decrement should move above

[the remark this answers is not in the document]
Anilm3: Moving it above the if statement should be enough I think.

appsec/src/helper/sampler.hpp
```diff
@@ -6,42 +6,285 @@
...
+    }
+
+    std::size_t approx_size(table &table) const { return table.size.load(); }
+
+private:
+    void rebuild_table(table &old_table, std::uint32_t report_threshold)
+    {
+        auto new_table = std::make_unique<table>();
+
+        struct copiable_entry {
```
Anilm3, line 243: nit: copyable?

appsec/src/helper/sampler.hpp
```diff
@@ -6,42 +6,285 @@
...
+        struct comp_entry {
+            bool operator()(
+                const copiable_entry &lhs, const copiable_entry &rhs)
+            {
+                return lhs.data.last_accessed < rhs.data.last_accessed;
+            }
+        };
+
+        // most recent at the top
+        std::priority_queue<copiable_entry, std::vector<copiable_entry>,
```
Anilm3, line 256: I think it would be faster to insert into a vector and then sort, since you don't need the data to be continuously sorted, you're incurring the cost of reordering the array on insertion.

cataphract: data in a heap is not fully sorted, and we don't need to sort all elements; we only need the top ones up to a certain number. That said, in these circumstances it's not necessarily faster than sorting the whole list at the end.

I'll look into the preallocation; it seems to require moving a preallocated vector into a different constructor overlaod.

appsec/src/helper/sampler.hpp
```diff
@@ -6,42 +6,285 @@
...
+            }
+        }
+
+        // insert up to MaxItems * 2 / 3
+        std::size_t count = 0;
+        while (!heap.empty() && count < MaxItems * 2 / 3) {
+            const auto &entry = heap.top();
+            auto [new_entry, _] = new_table->find_slot(entry.key);
+            new_entry.key.store(entry.key);
+            new_entry.data.store(entry.data);
```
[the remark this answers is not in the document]
cataphract: they could. I didn't bother because this is not a hot path, but they could be relaxed because the final exchange is has release ordering

appsec/src/helper/sampler.hpp
cataphract: "copiable" is the [canonical](https://en.wiktionary.org/wiki/copiable) form.

appsec/src/extension/configuration.h
```diff
@@ -68,7 +68,6 @@ extern bool runtime_config_first_init;
     CONFIG(BOOL, DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING_ENABLED, "true")                                                            \
     CONFIG(STRING, DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML, "")                                                                          \
     CONFIG(STRING, DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON, "")                                                                          \
-    CONFIG(DOUBLE, DD_API_SECURITY_REQUEST_SAMPLE_RATE, "0.1", .ini_change = zai_config_system_ini_change)                            \
```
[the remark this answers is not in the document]
cataphract: I've implemented as effectively a boolean (accepts only values starting with 0 -- presumably 0.0), because our resolution is only one second. Consequently, a straightforward implementation would mean the value of 0 would be equivalent to a value of 1.

Anilm3 (approved):

PR #3118, opened 2025-03-05
fix(profiling) I/O upscaling

realFlowControl:
### Description

This fixes some issues with I/O profiling, which was introduced in `v1.7.0`:
- `zend_thread` should be `true` when it is a zend thread and `false` otherwise, so we did not sample on Zend threads in `v1.7.0`
https://github.com/DataDog/dd-trace-php/blob/b38e9aed36475dba6985a90b76ce3037ac923767/profiling/src/io.rs#L613-L617
- upscaling the samples was missing and with introduction of higher sampling distances would result in wrong numbers
- `len` check was missing, if reading or writing fails, `read()`/`write()` would return `-1` which should not be sampled as bytes read/written
- Curl often uses `read()`/`write()` instead of `recv()`/`send()` even though operating on sockets, so Curl I/O would be wrongly accounted as file I/O instead of socket I/O (same is true for `file_get_contents()` when using it to fetch an HTTP URL)
- To not do a `fstat()` call to determine if `read()`/`write()` is operating on a socket or file, I've added a stats cache

PROF-11288

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/io.rs
```diff
@@ -424,20 +444,22 @@ unsafe extern "C" fn observed_fread(
     buf: *mut c_void,
     size: usize,
     n: usize,
-    fp: *mut libc::FILE,
+    fd: *mut libc::FILE,
 ) -> usize {
     let start = Instant::now();
-    let len = ORIG_FREAD(buf, size, n, fp);
+    let len = ORIG_FREAD(buf, size, n, fd);
```
bwoebi, line 450: I recommend reverting that, reading fp is better than fd for something that isn't actually a fd

realFlowControl: I think going with `stream` for `fread()`/`fwrite()` would match with[ Rust's `libc::fread()` and `libc::fwrite()` definition](https://docs.rs/libc/latest/libc/fn.fread.html). Okay with you?

bwoebi: Also fine, yes.

bwoebi (approved):
Looks good to me (though that macro add_io_profiling_rules! looks to me like it could have been a simple closure too).

realFlowControl:
> [...] though that macro add_io_profiling_rules! looks to me like it could have been a simple closure too

Yeah, I wen't a bit overboard with that 😉

PR #3121, opened 2025-03-06
feat(prof): add DD_TAGS support

morrisonlevi:
### Description

"Fun" fact: the profiler once had `DD_TAGS` support, and this is why
there are tag parsing functions in libdatadog. I think it was lost when
the profiler adopted ZAI config, which was quite a long time ago.

Another "fun" fact: the profiler also had an INI entry for
`datadog.tags`, it just didn't use it anywhere.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1099,7 +1099,6 @@ dependencies = [
  "ciborium",
  "clap",
  "criterion-plot",
- "csv",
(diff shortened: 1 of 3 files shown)


realFlowControl (approved):
LGTM, tested and works 🎉

PR #3123, opened 2025-03-07
Add span creation and close handlers

bwoebi:
The span creation handler is, in particular, needed to attach data dynamically to spans created within a context. E.g. attaching span links from a promises completion span to the promise awaiting span.

It is also beneficial for compression of spans; e.g. an eloquent integration may now opt to no longer create its own span, but manipulate the span created by the mysql integration instead - by first attaching a spanCreationObserver and then in that one attaching an onClose handler, gaining access to the final contents and able to manipulate it.

Span creation handlers are inherited via create_stack(). This can be essentially considered a form of context propagation without having a context with arbitrary metadata. These handlers can be globally removed by returning false (it resets the reference to null, effectively nulling it). That strategy ultimately allows full control over when the handlers are invoked. Also they are internally only called after the span was constructed with internal metadata, so that users never see spans in early construction stages, but before any user callbacks are invoked.

Span close handlers are executed in reverse order so that integrations (or users which ultimately end up calling integrations) may layer nicely (given that adding onClose handlers will generally happen with `$span->onClose[] = ...;` appending assignments).

It's also an alternative to an all-powerful post-processing handler.

### DIFF
--- a/tests/ext/span_create_observer.phpt
+++ b/tests/ext/span_create_observer.phpt
@@ -1,0 +1,99 @@
+--TEST--
+Test observing span creation
+--INI--
+datadog.trace.generate_root_span=0
+--FILE--
+<?php
+
+DDTrace\active_stack()->spanCreationObservers[] = function($span) {
+    $span->name = "hey";
+    echo "Observed\n";
+};
+
+DDTrace\start_span();
+$span = DDTrace\start_span();
+print "Got assigned a name: $span->name\n";
+DDTrace\close_span();
+DDTrace\close_span();
+
+DDTrace\create_stack();
+DDTrace\start_span();
+DDTrace\close_span();
+
+DDTrace\active_stack()->spanCreationObservers = [];
+DDTrace\start_span();
+DDTrace\close_span();
+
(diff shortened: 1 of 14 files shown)


tests/ext/span_create_observer.phpt
```diff
@@ -0,0 +1,99 @@
+--TEST--
+Test observing span creation
```
PROFeNoM, line 2: This test seems to be failing in, at least, PHP 7.4 ([link](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/19758/workflows/0398aad9-3e47-4a01-aac3-90f0cd32d296/jobs/6380399?invite=true#step-124-177435_86)). Is that expected? At a glance this appears to only be some typings-related things.

bwoebi: Fixed it, there was a wrong version constraint (the property is only typed on PHP 8.0+)

PROFeNoM (approved):
It looks good to me :+1: Thanks for adding the PHP snippet to the main code addition. It really helps 😃

PR #3131, opened 2025-03-14
Avoid using db.statement in favor of sql.query

bwoebi:
db.statement is not normalized while sql.query is. Also sending sql.query in SQLSRVIntegration is redundant.


ajgajg1134 (approved):
Looks good to me, are there any tests that need to be updated here?

ekump (approved):

PROFeNoM (approved):

PR #3142, opened 2025-03-26
Fix stable config precedence

iamluc:
### Description

The source (local or fleet) was not taken into account.

The precedence (as described in [this doc](https://docs.google.com/document/d/1MNI5d3g6R8uU3FEWf2e08aAsFcJDVhweCPMjQatEb0o/edit?pli=1&tab=t.0)) should be:
 "fleet managed stable config" > env vars > "customer/local managed stable config"

Issue caught by system-tests

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Looks good to me :-)

I'm not too happy that we now have two hashtable lookups, but I don't think _this_ is a significant performance issue, so approving.

PR #3145, opened 2025-03-27
Reduce OpenAI Flakiness & OTel PHP 7.4 Compat

PROFeNoM:
### Reviewer checklist
- [ ] Test coverage seems ok.


PROFeNoM:
According to the following [blog post](https://blog.packagist.com/deprecating-composer-1-support/), the package should be available for PHP 8.0 & 8.1 (using composer v1) tomorrow.

> This means new packages and old unused packages will only be visible to Composer 2 until someone installs them, at which point they will become visible to Composer 1 as well (within 24h as this process happens on a nightly basis).

bwoebi (approved):
Thank you :-)

PR #3146, opened 2025-03-27
fix(profiling) upscaling of I/O profiles

realFlowControl:
### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/profiling/mod.rs
```diff
@@ -292,6 +309,72 @@ impl TimeCollector {
...
+            .iter()
+            .position(|&x| x.r#type == "socket-read-time");
+        #[cfg(all(target_os = "linux", feature = "io_profiling"))]
+        let socket_read_time_samples_offset = sample_types
+            .iter()
+            .position(|&x| x.r#type == "socket-read-time-samples");
+        #[cfg(all(target_os = "linux", feature = "io_profiling"))]
+        let socket_write_time_offset = sample_types
+            .iter()
+            .position(|&x| x.r#type == "socket-write-time");
```
realFlowControl, line 324: Perhaps we can infer that if we have the `socket-read-time`, that for everything after the offset just has to be incremented ... maybe in a future cleanup

profiling/src/profiling/mod.rs
```diff
@@ -329,6 +412,150 @@ impl TimeCollector {
...
+                sampling_distance: SOCKET_READ_TIME_PROFILING_INTERVAL.load(Ordering::SeqCst),
+            };
+            let values_offset = [socket_read_time_offset, socket_read_time_samples_offset];
+            match profile.add_upscaling_rule(&values_offset, "", "", upscaling_info) {
+                Ok(_id) => {}
+                Err(err) => {
+                    warn!("Failed to add upscaling rule for socket read time samples, socket read time samples reported will be wrong: {err}")
+                }
+            }
+        }
```
realFlowControl, line 431: this could be put into a macro / function call to save on LOC and make it more maintainable

bwoebi:
Why are we hooking into readv()? At least a quick source search for `readv` finds nothing in php-src.

profiling/src/profiling/mod.rs
```diff
@@ -292,6 +309,72 @@ impl TimeCollector {
         #[cfg(feature = "allocation_profiling")]
         let alloc_size_offset = sample_types.iter().position(|&x| x.r#type == "alloc-size");
 
+        // check if we have the IO sample types
+        #[cfg(all(target_os = "linux", feature = "io_profiling"))]
```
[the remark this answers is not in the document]
bwoebi: Ah I see the "maybe in a future cleanup, so just for future reference, not planing to clean up in this PR" ... Maybe you really should do it in this PR, it makes it hard to spot mistakes.

morrisonlevi (approved):

profiling/src/profiling/mod.rs
```diff
@@ -329,6 +383,93 @@ impl TimeCollector {
...
+                        count_value_offset,
+                        sampling_distance,
+                    };
+                    let values_offset = [sum_value_offset, count_value_offset];
+                    match profile.add_upscaling_rule(&values_offset, "", "", upscaling_info) {
+                        Ok(_id) => {}
+                        Err(err) => {
+                            warn!("Failed to add upscaling rule for {metric_name}, {metric_name} reported will be wrong: {err}")
+                        }
+                    }
```
bwoebi, line 405:
```suggestion
                    if let Err(err) = profile.add_upscaling_rule(&values_offset, "", "", upscaling_info) {
                       warn!("Failed to add upscaling rule for {metric_name}, {metric_name} reported will be wrong: {err}")
                    }
```
nit

bwoebi (approved):
Thanks for the readability refactor :-)

PR #3147, opened 2025-03-27
Enable the sidecar sender by default on AWS lambda

bwoebi:
See comment in diff :-)

### DIFF
--- a/ext/configuration.c
+++ b/ext/configuration.c
@@ -181,11 +181,9 @@ bool ddtrace_config_minit(int module_number) {
     }
 
 #ifndef _WIN32
-    // Sidecar is currently broken - no traces sent. Investigation pending, background sender just works though.
+    // Background sender does not send a Content-Length header, but sidecar does. Force-enable it thus, as the background sender does not work at all.
     if (getenv("AWS_LAMBDA_FUNCTION_NAME")) {
-        config_entries[DDTRACE_CONFIG_DD_REMOTE_CONFIG_ENABLED].default_encoded_value = (zai_str) ZAI_STR_FROM_CSTR("false");
-        config_entries[DDTRACE_CONFIG_DD_TRACE_SIDECAR_TRACE_SENDER].default_encoded_value = (zai_str) ZAI_STR_FROM_CSTR("false");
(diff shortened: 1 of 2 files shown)


ext/configuration.c
```diff
@@ -181,11 +181,11 @@ bool ddtrace_config_minit(int module_number) {
     }
 
 #ifndef _WIN32
-    // Sidecar is currently broken - no traces sent. Investigation pending, background sender just works though.
+    // Background sender does not send a Content-Length header, but sidecar does. Force-enable it thus, as the background sender does not work at all.
     if (getenv("AWS_LAMBDA_FUNCTION_NAME")) {
```
iamluc, line 185: There is a similar condition in `appsec/src/extension/configuration.c`, shouldn't it be updated also?

ext/configuration.c
```diff
@@ -181,11 +181,11 @@ bool ddtrace_config_minit(int module_number) {
...
-        config_entries[DDTRACE_CONFIG_DD_REMOTE_CONFIG_ENABLED].default_encoded_value = (zai_str) ZAI_STR_FROM_CSTR("false");
-        config_entries[DDTRACE_CONFIG_DD_TRACE_SIDECAR_TRACE_SENDER].default_encoded_value = (zai_str) ZAI_STR_FROM_CSTR("false");
-        config_entries[DDTRACE_CONFIG_DD_INSTRUMENTATION_TELEMETRY_ENABLED].default_encoded_value = (zai_str) ZAI_STR_FROM_CSTR("false");
+        config_entries[DDTRACE_CONFIG_DD_REMOTE_CONFIG_ENABLED].default_encoded_value = (zai_str) ZAI_STR_FROM_CSTR("true");
```
iamluc, line 186: I understand that we force `DD_TRACE_SIDECAR_TRACE_SENDER` to `true`, but why are we doing it also for `DD_REMOTE_CONFIG_ENABLED` and `DD_INSTRUMENTATION_TELEMETRY_ENABLED`?
Shouldn't we honor the customer config?

ext/configuration.c
bwoebi: Oh, there is? Well, then yes, it should :-D

ext/configuration.c
bwoebi: But it's redundant because they anyway default to true.

ext/configuration.c
bwoebi: Done :-)

iamluc (approved):

PR #3148, opened 2025-03-28
Initialize baggage in RINIT

bwoebi:
Fixes segfault from system tests.

And properly supports numeric keys in serialization.


ext/ddtrace.c
```diff
@@ -1578,6 +1578,7 @@ static void dd_initialize_request(void) {
     zend_hash_init(&DDTRACE_G(root_span_tags_preset), 8, unused, ZVAL_PTR_DTOR, 0);
     zend_hash_init(&DDTRACE_G(propagated_root_span_tags), 8, unused, ZVAL_PTR_DTOR, 0);
     zend_hash_init(&DDTRACE_G(tracestate_unknown_dd_keys), 8, unused, ZVAL_PTR_DTOR, 0);
+    zend_hash_init(&DDTRACE_G(baggage), 8, unused, ZVAL_PTR_DTOR, 0);
```
iamluc, line 1581: I think the destroy is missing at shutdown

iamluc (approved):
Except my comment

ext/ddtrace.c
bwoebi: No, it's present in dd_clean_globals.

PR #3162, opened 2025-03-31
chore(configs): update DD_TAGS

mabdinur:
### Description

Ensure the parsing of the DD_TAGS is consistent with the Datadog Agent and other SDKs. This change is backwards compatible with existing DD_TAGS. The edges addressed by this fix will be tested in system tests: https://github.com/DataDog/system-tests/pull/4439

Note: This PR essentially translated the python implementation to c: https://github.com/DataDog/dd-trace-py/blob/v3.3.0/ddtrace/internal/utils/formats.py#L68. 

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/configuration.c
```diff
@@ -94,6 +94,68 @@ static bool dd_parse_sampling_rules_format(zai_str value, zval *decoded_value, b
...
+        array_init(decoded_value);
+        return true;
+    }
+
+    array_init(decoded_value);
+    
+    // Determine separator - prefer comma if present, otherwise use space
+    const char *sep = strchr(value.ptr, ',') ? "," : " ";
+    
+    char *str = strdup(value.ptr);
```
bwoebi, line 110: Why the extra string duplication? Just so that you can use strtok? That doesn't look clean to me.

bwoebi: Instead of strtok you may use strcspn.

bwoebi: I assume it wants to string duplicate too, so that it can pass null-terminated strings to add_assoc_string. Which can't be used to handle persistent strings.
Needs to use `zend_string_init(str, len, persistent)`.

mabdinur: Yes, I used cursor to translate some python snippets to C and then tweaked the output. I'll make those edits, thanks for the sanity check 😄

ext/configuration.c
```diff
@@ -94,6 +94,68 @@ static bool dd_parse_sampling_rules_format(zai_str value, zval *decoded_value, b
     return true;
 }
 
+static bool dd_parse_tags(zai_str value, zval *decoded_value, bool persistent) {
+    UNUSED(persistent);
```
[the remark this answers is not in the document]
mabdinur: It seems like `persistent` is ignored in all the parsers in `ext/configuration.c`. Is there an example on how it should be handled?

bwoebi (approved):

PR #3165, opened 2025-03-31
AppSec support for Frankenphp

cataphract:
### Reviewer checklist
- [ ] Test coverage seems ok.


ext/user_request.c
```diff
@@ -14,7 +13,7 @@ static struct {
 
 DDTRACE_PUBLIC bool ddtrace_user_req_add_listeners(ddtrace_user_req_listeners *listeners)
 {
-    if (strcmp(sapi_module.name, "cli") != 0) {
+    if (strcmp(sapi_module.name, "cli") != 0 && strcmp(sapi_module.name, "frankenphp") != 0) {
```
bwoebi, line 16: You know that the frankenphp SAPI has two different ways of working, one worker mode and one which essentially works like fpm? In the latter case you likely want this function to not be skipped.

cataphract: The condition does the opposite: it skips adding the listeners if not on cli (or frankenphp).

If we're running in classic mode, this has no ill effects: `frankenphp_handle_request` will never be called; the frankenphp integration won't be installed and `\DDTrace\UserRequest\notify_start` will never be called; the installed listeners will never be called.

I've added support and tests for classic mode; however, because I'm unable to detect it, it is being the option `DD_APPSEC_CLI_START_ON_RINIT`, which disables the "user request" lifecycle and delineates the request with RINIT/RSHUTDOWN

bwoebi: Ah right, this code runs in MINIT, and then you don't know yet. You can only detect that in RINIT via contents of $_SERVER.

appsec/src/extension/request_abort.c
```diff
@@ -601,13 +626,31 @@ void dd_request_abort_startup()
     _content_type_json_zstr =
         zend_string_init_interned(ZEND_STRL(JSON_CONTENT_TYPE), 1);
 
+#ifdef FRANKENPHP_SUPPORT
+    if (strcmp(sapi_module.name, "frankenphp") == 0) {
+        // prepare a noop zend_compile_file for req init blocking in frankenphp
+        _orig_zend_compile_file = zend_compile_file;
+        zend_compile_file = _req_init_block_zend_compile_file;
+    }
+#endif
```
bwoebi, line 635: This needs to happen in post_startup otherwise it won't be executed with opcache active.

cataphract: Thanks, fixed in [8355494](https://github.com/DataDog/dd-trace-php/pull/3165/commits/8355494de338c5c9f0c5f0d61b3e686c719048d3).

bwoebi (approved):
I don't look at tests, but overall that looks fine.

cataphract:
Thanks. Merging. Just added https://github.com/DataDog/dd-trace-php/pull/3165/commits/387d24cec64bc8148fc20a27e5c1e4ccf5bd2860 to try to fix flakiness I was seeing on the first frankenphp worker test.

PR #3167, opened 2025-03-31
chore(statsd): support DD_DOGSTATSD_HOST configuration

mabdinur:
### Description

Adds support for `DD_DOGSTATSD_HOST`. This configuration is supported by other ddtrace sdks

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/dogstatsd_client.c
+++ b/ext/dogstatsd_client.c
@@ -70,7 +70,12 @@ void ddtrace_dogstatsd_client_rinit(void) {
             host = url;
             port = NULL;
         } else {
-            host = ZSTR_VAL(get_DD_AGENT_HOST());
+            zend_string *dogstatsd_host = get_DD_DOGSTATSD_HOST();
+            if (ZSTR_LEN(dogstatsd_host) > 0) {
+                host = ZSTR_VAL(dogstatsd_host);
+            } else {
+                host = ZSTR_VAL(get_DD_AGENT_HOST());
+            }
             port = ZSTR_VAL(get_DD_DOGSTATSD_PORT());
 
             if (!*host) {
--- a/tests/ext/dogstatsd/metrics_host_config.phpt
+++ b/tests/ext/dogstatsd/metrics_host_config.phpt
@@ -1,0 +1,57 @@
+--TEST--
+Test DogStatsD configuration with DD_DOGSTATSD_HOST and DD_DOGSTATSD_PORT
+--SKIPIF--
+<?php if (!extension_loaded('sockets')) die('skip: the sockets extension is required for this test'); ?>
+--ENV--
+DD_DOGSTATSD_HOST=192.168.1.1
+DD_DOGSTATSD_PORT=9876
+--FILE--
+<?php
+
+class UDPServer {
+    private $socket;
+
+    public function __construct($addr, $port) {
+        if (!($this->socket = socket_create(AF_INET, SOCK_DGRAM, 0))) {
+            $errorcode = socket_last_error();
+            $errormsg = socket_strerror($errorcode);
+            die("Couldn't create socket: [$errorcode] $errormsg\n");
+        }
+
+        if (!socket_bind($this->socket, $addr, $port)) {
+            $errorcode = socket_last_error();
+            $errormsg = socket_strerror($errorcode);
+            die("Could not bind socket : [$errorcode] $errormsg\n");
+        }
+    }
+
+    public function dump($expected, $iter = 5000) {
+        $lines = [];
+        for ($i = 0; $i < $iter; ++$i) {
+            usleep(100);
+            if (socket_recvfrom($this->socket, $buf, 2048, MSG_DONTWAIT, $remote_ip, $remote_port)) {
+                $lines[] = "$buf\n";
+                if (count($lines) == $expected) {
+                    sort($lines, SORT_STRING);
+                    echo implode($lines);
+                    return;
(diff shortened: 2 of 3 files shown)


tests/ext/dogstatsd/metrics_host_config.phpt
```diff
@@ -0,0 +1,34 @@
...
+        return strlen($data);
+    }
+
+    public function stream_close() {
+        return true;
+    }
+}
+
+// Register our mock socket wrapper for udp connections
+stream_wrapper_register('udp', 'MockUDPSocket');
```
bwoebi, line 29: This does not work for connections initiated from the C code. You cannot mock this; you'll need an actual socket.

ext/dogstatsd_client.c
```diff
@@ -70,7 +70,12 @@ void ddtrace_dogstatsd_client_rinit(void) {
             host = url;
             port = NULL;
         } else {
-            host = ZSTR_VAL(get_DD_AGENT_HOST());
+            zend_string *dogstatsd_host = get_DD_DOGSTATSD_HOST();
+            if (ZSTR_LEN(dogstatsd_host) > 0) {
+                host = ZSTR_VAL(dogstatsd_host);
+            } else {
+                host = ZSTR_VAL(get_DD_AGENT_HOST());
+            }
```
bwoebi, line 78: This code should rather be refactored to use ddtrace_dogstatsd_url()... And this host being also checked there then.

mabdinur: Good point. I was going for a full integration test but that seems like overkill here. I just need to verify that the new configuration is being read.

bwoebi: You're conflating the comments - ddtrace_dogstatsd_url is a C function, I wasn't talking about the test here.
This url is not exposed directly to the PHP code.

bwoebi: I don't think you can easily.
Also, if you want to test against not-localhost, you might want to bind a socket on e.g. 172.0.0.2 instead.

bwoebi:
I think we can do without a proper test for this feature. The change is trivial.
Can we rely on system-tests to test it properly?

ext/dogstatsd.c
```diff
@@ -13,7 +13,11 @@ char *ddtrace_dogstatsd_url(void) {
         return zend_strndup(ZSTR_VAL(url), ZSTR_LEN(url) + 1);
     }
 
-    zend_string *hostname = get_global_DD_AGENT_HOST();
+    zend_string *hostname = get_DD_DOGSTATSD_HOST();
+    if (!hostname || ZSTR_LEN(hostname) == 0) {
```
bwoebi, line 17:
```suggestion
    if (ZSTR_LEN(hostname) == 0) {
```

ext/dogstatsd_client.c
```diff
@@ -34,10 +35,11 @@ void ddtrace_dogstatsd_client_rinit(void) {
...
     while (health_metrics_enabled) {
-        struct addrinfo *addrs;
-        char *url = ZSTR_VAL(get_DD_DOGSTATSD_URL());
-        char *host, *port;
-        if (*url) {
+        struct addrinfo *addrs = NULL;
+        char *url = ddtrace_dogstatsd_url();
+        char *host = NULL;
+        char *port = NULL;
+        if (url) {
```
bwoebi, line 42: unnecessary NULL check

ext/dogstatsd_client.c
```diff
@@ -46,60 +48,40 @@ void ddtrace_dogstatsd_client_rinit(void) {
                     LOG(WARN,
                         "Dogstatsd client encountered an invalid udp:// DD_DOGSTATSD_URL: %s, missing a colon followed by a port",
                         url);
+                    efree(url);
```
bwoebi: free(), not efree - it's not allocated via emalloc.

mabdinur: I used ZVal to store the url instead

ext/dogstatsd_client.c
```diff
@@ -34,72 +35,44 @@ void ddtrace_dogstatsd_client_rinit(void) {
...
+            int err;
+            if ((err = dogstatsd_client_getaddrinfo(&addrs, host, port))) {
+                LOG(WARN, "Dogstatsd client failed looking up %s:%s: %s", host, port,
+                                    (err == EAI_SYSTEM) ? strerror(errno) : gai_strerror(err));
+                efree(host);
+                free(port);
+                break;
             }
+            efree(host);
+            free(port);
```
mabdinur, line 62:
```suggestion
```

ext/dogstatsd_client.c
```diff
@@ -34,72 +35,44 @@ void ddtrace_dogstatsd_client_rinit(void) {
...
-                    break;
-                }
+            char *host = estrndup(url + 6, colon - url - 6);
+            char *port = colon + 1;
+            int err;
+            if ((err = dogstatsd_client_getaddrinfo(&addrs, host, port))) {
+                LOG(WARN, "Dogstatsd client failed looking up %s:%s: %s", host, port,
+                                    (err == EAI_SYSTEM) ? strerror(errno) : gai_strerror(err));
+                efree(host);
+                free(port);
```
mabdinur, line 58:
```suggestion
```

bwoebi (approved):
Looks right to me.

PR #3169, opened 2025-04-01
fix(prof) FrankenPHP idle phase

realFlowControl:
### Description

This PR fixes FrankenPHP idle phase detection in worker mode. Turns out, that `frankenphp_handle_request()` does also execute the closure given to it, so the current way of measuring the idle phase in worker mode is incorrect. So what I am doing now instead of hooking `frankenphp_handle_request()` is hooking into `sapi_module.activate`/ `sapi_module.deactivate` in case the detected SAPI is FrankenPHP. If one of those functions is being called while `frankenphp_handle_request()` is the current execute frame, we are indeed in the idle phase.

Why does this work?

FrankenPHP in `frankenphp_worker_request_shutdown()` calls into [`sapi_deactivate()`](https://github.com/dunglas/frankenphp/blob/d3589f97706eac585cdafddbc059cdd53711faf9/frankenphp.c#L141) and in `frankenphp_worker_request_startup()` it calls into [`sapi_activate()`](https://github.com/dunglas/frankenphp/blob/d3589f97706eac585cdafddbc059cdd53711faf9/frankenphp.c#L181).

There is also a small cleanup: I have extracted the idle phase start and stop code into their own functions.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -941,9 +941,7 @@ extern "C" fn mshutdown(_type: c_int, _module_number: c_int) -> ZendResult {
 
     // SAFETY: being called before [config::shutdown].
     #[cfg(feature = "timeline")]
-    unsafe {
-        timeline::timeline_mshutdown()
-    };
+    timeline::timeline_mshutdown();
 
     #[cfg(feature = "exception_profiling")]
     exception::exception_profiling_mshutdown();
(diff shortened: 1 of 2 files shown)


bwoebi (approved):
Looks great to me :-)

PR #3171, opened 2025-04-01
Fix stable config in profiling when ddtrace is not loaded

iamluc:
### Description

Fix stable config when only the `datadog-profiling` is loaded, and not `ddtrace`

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Looks right to me

PR #3175, opened 2025-04-01
follow php globals model

realFlowControl:
### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (commented):
I think this would still give a panic at the place we want, while not having string duplication. Linters (or maybe the compiler?) don't like panic having a constant as a format string, likes string literals.

```diff
diff --git a/profiling/src/allocation/mod.rs b/profiling/src/allocation/mod.rs
index 58437ebb9..0962d1f4b 100644
--- a/profiling/src/allocation/mod.rs
+++ b/profiling/src/allocation/mod.rs
@@ -78,36 +78,41 @@ struct ZendMMState {
     shutdown: unsafe fn(bool, bool),
 }
 
-unsafe fn alloc_prof_panic_alloc(_len: size_t) -> *mut c_void {
+
+#[cfg(track_caller)]
+fn initialization_panic() -> ! {
     panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
 }
+unsafe fn alloc_prof_panic_alloc(_len: size_t) -> *mut c_void {
+    initialization_panic();
+}
 
 unsafe fn alloc_prof_panic_realloc(_prev_ptr: *mut c_void, _len: size_t) -> *mut c_void {
-    panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
+    initialization_panic();
 }
 
 unsafe fn alloc_prof_panic_free(_ptr: *mut c_void) {
-    panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
+    initialization_panic();
 }
 
 #[cfg(php_zend_mm_set_custom_handlers_ex)]
 unsafe fn alloc_prof_panic_gc() -> size_t {
-    panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
+    initialization_panic();
 }
 
 #[cfg(php_zend_mm_set_custom_handlers_ex)]
 unsafe fn alloc_prof_panic_shutdown(_full: bool, _silent: bool) {
-    panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
+    initialization_panic();
 }
 
 #[cfg(not(php_zend_mm_set_custom_handlers_ex))]
 fn prepare_zend_heap_panic(_heap: *mut zend::_zend_mm_heap) -> c_int {
-    panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
+    initialization_panic();
 }
 
 #[cfg(not(php_zend_mm_set_custom_handlers_ex))]
 fn restore_zend_heap_panic(_heap: *mut zend::_zend_mm_heap, _custom_heap: c_int) {
-    panic!("Allocation profiler was not initialised properly. Please fill an issue stating the PHP version and the backtrace from this panic.");
+    initialization_panic();
 }
 
 impl ZendMMState {
```

realFlowControl:
I've committed your suggestion

morrisonlevi (commented):
This version of the patch is much easier to review, thank you! I'm going to push up some cleanup which shouldn't be controversial.

I _think_ I see a better/cleaner way to do this we should discuss over zoom (although you are welcome to try it). Basically, I think we can move from `UnsafeCell` to `Cell`, mark `ZendMMState` as `Copy`, and use a `Cell` for the pure global version too. Then, we drop the macros.

At a high level, this works because `LocalKey` has convenience methods since Rust 1.73 (`local_key_cell_methods`). This avoids needing to use the `with` dance:

```rs
thread_local! {
    static X: Cell<i32> = Cell::new(1);
}

// No `with` stuff needed!
assert_eq!(X.get(), 1);
```

morrisonlevi:
@realFlowControl, please review my recent commit. It implements the idea to use `Cell` instead of `UnsafeCell`. As I showed you in Slack, the compiler is good at not copying the whole struct just to use one field. At least, in the places I checked ^_^

It also deduplicates `initialization_panic`, `alloc_prof_panic_alloc`, `alloc_prof_panic_realloc`, and `alloc_prof_panic_free` and does some other cleanup.

morrisonlevi (approved):
I approve, but as I contributed to this, Florian should review my work. Also, probably do some manual testing of it too (just usual stuff, not asking for anything extreme).

profiling/src/allocation/allocation_le83.rs
```diff
@@ -48,30 +48,67 @@ struct ZendMMState {
...
+    };
+}
+
+#[cfg(not(php_zts))]
+/// Using a `Cell` here should be okay. There might not be any
+/// synchronization issues, as it is only mutated in RINIT and RSHUTDOWN.
+static mut ZEND_MM_STATE: Cell<ZendMMState> = Cell::new(ZendMMState::new());
+
+#[cfg(php_zts)]
+macro_rules! tls_zend_mm_state_copy {
```
realFlowControl, line 82: The naming is weird to my brain ...

PR #3176, opened 2025-04-01
Handle flushing of multiple traces at once

bwoebi:
Working around the agent complaining about "foreign_span". With the agent each array of spans must have the same trace id.
This is already naturally separated with the closed_stack logic, but it was never actually split into different arrays.

This needs to also move the asm logic as now there's definitely no longer one single trace returned from ddtrace_serialize_closed_spans.

The asm logic for event emitted is instead stored on the root span where the event happens (or inherited from global scope if there was no root span at emission time). And the priority sampling / rate limiting applied when the root span is closed. Moving it to the root span close also simplifies the logic and allows possibly straightforwardly dropping the whole trace in future. This also fixes the asm event accidentally persisting in worker modes.

### DIFF
--- a/ext/asm_event.c
+++ b/ext/asm_event.c
@@ -12,8 +12,12 @@ DDTRACE_PUBLIC void ddtrace_emit_asm_event() {
     if (!DDTRACE_G(active_stack)) {
         return;
     }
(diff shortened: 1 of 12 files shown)


estringana (approved):
Thanks for this PR @bwoebi

PR #3082, opened 2025-02-12
Enable RASP by default

estringana:
### Description

Enable RASP by default

[APPSEC-56457]

### Reviewer checklist
- [ ] Test coverage seems ok.
- [ ] Appropriate labels assigned.

[APPSEC-56457]: https://datadoghq.atlassian.net/browse/APPSEC-56457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3178, opened 2025-04-01
Revert "Enable RASP by default"

estringana:
Reverts DataDog/dd-trace-php#3082


bwoebi (approved):
Turning on rasp as a last minute event is not great, so let's revert :-)

PR #3181, opened 2025-04-02
Defer dropping the Symfony exec span to close time

bwoebi:
This avoids "There is no user-span on the top of the stack" errors.

Also make sure the tests watch CLI output for ddtrace errors.


PROFeNoM (approved):
Checked locally and confirmed that 
```
[ddtrace] [error] There is no user-span on the top of the stack. Cannot close.
```
is not present anymore when running
```
composer create-project symfony/skeleton:"7.2.x" skeleton --quiet && (skeleton/bin/console | grep ddtrace)
```
with this branch's artifact

PR #3182, opened 2025-04-03
fix(dd_tags): Update precedence of UST Configurations

mabdinur:
### Description

Ensure that configurations set in `DD_SERVICE`, `DD_ENV`, and `DD_VERSION` take precedence over values set in DD_TAGS.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/serializer.c
+++ b/ext/serializer.c
@@ -442,7 +442,6 @@ static void dd_add_post_fields_to_meta_recursive(zend_array *meta, const char *t
 
 void ddtrace_set_global_span_properties(ddtrace_span_data *span) {
     zend_array *meta = ddtrace_property_array(&span->property_meta);
-
     zend_array *global_tags = get_DD_TAGS();
     zend_string *global_key;
     zval *global_val;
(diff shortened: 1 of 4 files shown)


bwoebi:
@mabdinur This change would prevent env, service and version from being applied at all if coming from DD_TAGS. But as far as I know that's a supported use case?

`service` on the root span is _always_ set - if DD_SERVICE is empty it defaults to `ddtrace_default_service_name`.

ext/serializer.c
```diff
@@ -445,7 +446,17 @@ void ddtrace_set_global_span_properties(ddtrace_span_data *span) {
...
     zend_string *global_key;
     zval *global_val;
+    zval *prop_service = &span->property_service;
+    zval *prop_env = &span->property_env;
+    zval *prop_version = &span->property_version;
+
     ZEND_HASH_FOREACH_STR_KEY_VAL(global_tags, global_key, global_val) {
+        if ((Z_TYPE_P(prop_service) == IS_STRING && zend_string_equals_literal(global_key, "service")) ||
+            (Z_TYPE_P(prop_env) == IS_STRING && zend_string_equals_literal(global_key, "env")) ||
+            (Z_TYPE_P(prop_version) == IS_STRING && zend_string_equals_literal(global_key, "version"))) {
```
mabdinur, line 456:
```suggestion
        if ((zend_string_equals_literal(global_key, "service") && Z_TYPE_P(prop_service) == IS_STRING) ||
            (zend_string_equals_literal(global_key, "env") && Z_TYPE_P(prop_env) == IS_STRING) ||
            (zend_string_equals_literal(global_key, "version") && Z_TYPE_P(prop_version) == IS_STRING)) {
```

bwoebi:
If you ask me, I'd completely stop supporting env and version from DD_TAGS.
I would still skip the keys - but unconditionally.

Fwiw - service is already unsupported in DD_TAGS.

mabdinur:
> If you ask me, I'd completely stop supporting env and version from DD_TAGS. I would still skip the keys - but unconditionally. There's been a deprecation warning for a year now, if you used env or version in DD_TAGS.
> 
> Fwiw - service is already unsupported in DD_TAGS (as in: has no effect), I just realized.

I agree that service, version, and env should not be tags. They should be span properties. We just don't have the bandwidth to make this change across all client libraries this quarter. 

Since all tracing libraries (except for PHP) support setting service, version, and env via DD_TAG. We plan to just align the PHP tracer with all other sdks and plan to revisit this decision in a future quarter.

tldr: We are prioritizing consistency

bwoebi:
@mabdinur Let's keep the PR as is then, but drop the check for service. service is not doing anything.

The PHP tracer supports setting version and env, it's just deprecated.
It does _not_ support service. I also don't want to add this now, to remove it later.
I'd keep the PHP tracer temporarily inconsistent for this slight detail. Also: I've never had an user complaining about service not being supported in DD_TAGS.

PR #3192, opened 2025-04-08
chore(prof): update `libdatadog` to v17

realFlowControl:
### Description

This PR aims to update the `libdatadog` dependency to latest version 17. It does _not_ aim to use any of the new features, like the managed string storage introduced in https://github.com/DataDog/libdatadog/pull/725

Also this PR ran:
```sh
$ cargo update target-triple --precise 0.1.4
```

A `cargo update` does not work because otherwise the tracer `.so` file that we compile for Alpine fails on Alpine 3.7 on loading in PHP with `getrandom: symbol not found`.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/uploader.rs
+++ b/profiling/src/profiling/uploader.rs
@@ -7,7 +7,6 @@ use crate::profiling::{UploadMessage, UploadRequest};
 use crate::{PROFILER_NAME_STR, PROFILER_VERSION_STR};
 use chrono::{DateTime, Utc};
 use crossbeam_channel::{select, Receiver};
-use datadog_profiling::exporter::File;
 use ddcommon::Endpoint;
 use log::{debug, info, warn};
 use serde_json::json;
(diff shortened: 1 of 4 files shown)


profiling/src/profiling/uploader.rs
```diff
@@ -97,21 +96,12 @@ impl Uploader {
 
         let serialized =
             profile.serialize_into_compressed_pprof(Some(message.end_time), message.duration)?;
-        let endpoint_counts = Some(&serialized.endpoints_stats);
-        let start = serialized.start.into();
-        let end = serialized.end.into();
-        let files = &[File {
-            name: "profile.pprof",
-            bytes: serialized.buffer.as_slice(),
-        }];
```
morrisonlevi, line 106: I assume these things all moved into the internals?

morrisonlevi (approved):

profiling/src/profiling/uploader.rs
realFlowControl: Exactly. Just the line above (the call to `profile.serialize_into_compressed_pprof()`) returns an `EncodedProfile` which has all those values already and the new API with libdatadog 17 takes this `EncodedProfile` as the first argument to `exporter.build()` instead.

PR #3194, opened 2025-04-08
Use native rust spans instead of manually serializing to msgpack

Leiyks:
### Reviewer checklist
- [ ] Test coverage seems ok.


ext/serializer.c
```diff
@@ -1458,37 +1515,35 @@ zval *ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
     }
-    zval *el;
-    zval zv;
-    el = &zv;
-    array_init(el);
 
-    add_assoc_str(el, KEY_TRACE_ID, ddtrace_span_id_as_string(span->root->trace_id.low));
-    add_assoc_str(el, KEY_SPAN_ID, zend_string_copy(span->string_id));
+    ddog_set_span_trace_id(rust_span, span->root->trace_id.low);
+    ddog_set_span_span_id(rust_span, dd_parse_zend_string_id(span->string_id));
```
bwoebi, line 1520: We have also `span->span_id`, no need to compute the uint64_t here.

ext/serializer.c
```diff
@@ -319,11 +312,13 @@ static int dd_fatal_error_to_meta(zend_array *meta, dd_error_info error) {
     return error.type && error.msg ? SUCCESS : FAILURE;
 }
 
-static zend_result dd_add_meta_array(void *context, ddtrace_string key, ddtrace_string value) {
-    zval *meta = context, tmp = ddtrace_zval_stringl(value.ptr, value.len);
+static zend_result dd_add_meta_array(ddog_SpanBytes *span, ddtrace_string key, ddtrace_string value) {
+    ddog_ZendString *key_zstr = (ddog_ZendString *)zend_string_init(key.ptr, key.len, 0);
```
bwoebi, line 316: May I suggest to use
```
[export.rename]
"ZendString" = "_zend_string"
```
in cbindgen.toml, to avoid casting stuff around all the time?

components-rs/bytes.rs
```diff
@@ -0,0 +1,227 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn ddog_del_span_meta_(ptr: *mut SpanBytes, key: *mut ZendString) {
+    remove_hashmap!(ptr, key, meta);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn ddog_get_span_meta_(
+    ptr: *mut SpanBytes,
+    key: *mut ZendString,
+) -> *mut c_char {
```
bwoebi, line 153: We generally want to return ddcommon_ffi::CharSlice from ffi. We do not want to do any memory allocations or such inside getters.

components-rs/bytes.rs
```diff
@@ -0,0 +1,227 @@
...
+    exists_hashmap!(ptr, key, metrics);
+}
+
+pub unsafe extern "C" fn ddog_add_span_meta_struct_(
+    ptr: *mut SpanBytes,
+    key: *mut ZendString,
+    val: *mut ZendString,
+) {
+    insert_hashmap!(ptr, key, convert_to_bytes(val), meta_struct);
+}
```
bwoebi, line 227: Let's also add a variant taking a CharSlice as key. Many of the keys are hardcoded and don't exist as zend_string.
It's less expensive too, to use `impl From<&'static str> for BytesString { fn from(value: &'static str) -> Self {` that function.
`key.as_slice().into()`, then.

ext/serializer.c
```diff
@@ -1137,18 +1153,36 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
 }
 
-static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *service_name) {
+static void dd_set_entrypoint_root_rust_span_props_end(ddog_SpanBytes *span, int status, struct iter *headers, bool ignore_error) {
+    if (status) {
+        zend_string *status_str = zend_long_to_str((long)status);
+        ddog_add_CharSlice_span_meta_(span, dd_str_to_CharSlice("http.status_code"), status_str);
+
+        if (status >= 500 && !ignore_error) {
+            ddog_add_CharSlice_span_meta_(span, dd_str_to_CharSlice("error.type"), dd_str_to_zend_string("Internal Server Error"));
```
bwoebi, line 1162:
```suggestion
            ddog_add_span_meta(span, dd_str_to_CharSlice("error.type"), dd_str_to_CharSlice("Internal Server Error"));
```

components-rs/bytes.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    let key = convert_char_slice_to_bytes_string(key);
+
+    match span.meta.get(&key) {
+        Some(value) => {
+            CharSlice::from_raw_parts(value.as_str().as_ptr().cast(), value.as_str().len())
+        }
+        None => CharSlice::empty(),
+    }
+}
```
bwoebi, line 198: We have no need to duplicating these methods which do not contain ZendString here. libdatadog covers these.

components-rs/bytes.rs
```diff
@@ -0,0 +1,301 @@
...
+
+impl UnderlyingBytes for ZendStringWrapper {}
+
+unsafe fn convert_to_bytes(zend_str: *mut ZendString) -> Bytes {
+    (*zend_str).refcount += 1; // Increment the reference count to prevent double free
+
+    match String::from_utf8_lossy((*zend_str).as_ref()) {
+        Cow::Owned(s) => s.into(),
+        Cow::Borrowed(_) => Bytes::from_underlying(ZendStringWrapper(zend_str)),
+    }
```
bwoebi, line 54:
```suggestion
    Bytes::from_underlying(ZendStringWrapper(zend_str)),
```
We only need from_utf8_lossy if we actually need a BytesString. For raw Bytes that's not needed.

components-rs/bytes.rs
```diff
@@ -0,0 +1,301 @@
...
+
+#[no_mangle]
+pub unsafe extern "C" fn ddog_del_span_metrics_(ptr: *mut SpanBytes, key: *mut ZendString) {
+    remove_hashmap!(ptr, convert_zend_to_bytes_string(key), metrics);
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn ddog_del_CharSlice_span_metrics_(ptr: *mut SpanBytes, key: CharSlice) {
+    remove_hashmap!(ptr, convert_char_slice_to_bytes_string(key), metrics);
+}
```
bwoebi, line 239: but that one we definitely don't need - it's covered by ddog_del_span_metrics from data-pipeline for example.

ext/serializer.c
```diff
@@ -319,11 +312,13 @@ static int dd_fatal_error_to_meta(zend_array *meta, dd_error_info error) {
     return error.type && error.msg ? SUCCESS : FAILURE;
 }
 
-static zend_result dd_add_meta_array(void *context, ddtrace_string key, ddtrace_string value) {
-    zval *meta = context, tmp = ddtrace_zval_stringl(value.ptr, value.len);
+static zend_result dd_add_meta_array(ddog_SpanBytes *span, ddtrace_string key, ddtrace_string value) {
```
bwoebi, line 315: We should probably get rid of this function and just call ddog_add_span_meta and such variants at the appropriate places, and also avoid passing this around as function pointer.
This function had more uses in the past, but now it can be eliminated, when we're refactoring this anyway now.

ext/serializer.c
```diff
@@ -1159,40 +1193,41 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *serv
...
+            existing_env.ptr = "";
         }
     }
+
     if (!span->parent) {
         if (DDTRACE_G(last_flushed_root_env_name)) {
             zend_string_release(DDTRACE_G(last_flushed_root_env_name));
         }
-        DDTRACE_G(last_flushed_root_env_name) = zend_string_copy(Z_STR_P(existing_env));
+        DDTRACE_G(last_flushed_root_env_name) = dd_CharSlice_to_zend_string(existing_env);
```
bwoebi, line 1220: We need to refcount this. This must remain a zend_string - do the conversion to charslice where it's used.

ext/serializer.c
```diff
@@ -1159,40 +1193,41 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *serv
...
     }
```
bwoebi, line 1221: Can you please fetch this from the meta HashTable instead of rust? I.e. just leave these lines unchanged.

bwoebi: And below for the version too.

ext/serializer.h
```diff
@@ -22,6 +22,6 @@ void ddtrace_shutdown_span_sampling_limiter(void);
 
 void ddtrace_serializer_startup(void);
 
-typedef zend_result (*add_tag_fn_t)(void *context, ddtrace_string key, ddtrace_string value);
+typedef zend_result (*add_tag_fn_t)(ddog_SpanBytes *span, ddtrace_string key, ddtrace_string value);
```
bwoebi, line 26:
```suggestion
```
Obsolete.

ext/sidecar.h
```diff
@@ -60,3 +77,4 @@ bool ddtrace_exception_debugging_is_active(void);
 ddog_crasht_Metadata ddtrace_setup_crashtracking_metadata(ddog_Vec_Tag *tags);
 
 #endif // DD_SIDECAR_H
+//
```
bwoebi, line 80: hm.

ext/serializer.c
```diff
@@ -1159,26 +1184,27 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *serv
...
-
+ 
     zval *existing_env, new_env;
     if ((existing_env = zend_hash_str_find(Z_ARRVAL_P(meta), ZEND_STRL("env")))) {
         LOG(DEPRECATED, "Using \"env\" in meta is deprecated. Instead specify the env property directly on the span.");
     } else {
         ddtrace_convert_to_string(&new_env, &span->property_env);
         if (Z_STRLEN(new_env)) {
-            existing_env = zend_hash_str_add_new(Z_ARRVAL_P(meta), ZEND_STRL("env"), &new_env);
+            ddog_add_span_meta(rust_span, dd_str_to_CharSlice("env"), dd_zval_string_to_CharSlice(existing_env));
```
bwoebi, line 1199:
```suggestion
            ddog_add_CharSlice_span_meta_zstr(rust_span, dd_str_to_CharSlice("env"), Z_STR_P(existing_env));
```

ext/sidecar.h
```diff
@@ -42,6 +43,22 @@ static inline ddog_CharSlice dd_zai_string_to_CharSlice(zai_string str) {
...
+}
+
+static inline zend_string *dd_str_to_zend_string(char *str) {
+    return zend_string_init(ZEND_STRL(str), 0);
+}
+
+static inline zend_string *dd_zval_to_zend_string(zval *str) {
+    return Z_STR_P(str);
+}
+
```
bwoebi, line 61:
```suggestion
static inline ddog_CharSlice dd_str_to_CharSlice(char *str) {
    return (ddog_CharSlice){ .len = strlen(str), .ptr = str };
}

```

Let's cleanup what we don't - and should not - need.

ext/serializer.c
```diff
@@ -553,17 +561,17 @@ static zend_string *dd_get_user_agent(zend_array *_server) {
...
+    if (zend_hash_num_elements(peer_service_mapping) == 0 || !peer_service) {
         return false;
     }
 
     zval* mapped_service_zv = zend_hash_find(peer_service_mapping, peer_service);
     if (mapped_service_zv) {
         zend_string *mapped_service = zval_get_string(mapped_service_zv);
-        add_assoc_str(meta, "peer.service.remapped_from", peer_service);
-        add_assoc_str(meta, "peer.service", mapped_service);
+        ddog_add_CharSlice_span_meta_zstr(span, dd_str_to_CharSlice("peer.service.remapped_from"), peer_service);
```
bwoebi, line 573: We have `DDOG_CHARSLICE_C()` already as a macro for literals.

ext/exception_serialize.c
```diff
@@ -36,27 +38,21 @@ static zend_result dd_exception_to_error_msg(zend_object *exception, void *conte
...
+    asprintf(&error_text, "%s %s%s%s%s in %s:" ZEND_LONG_FMT, exception_type,
             ZSTR_VAL(exception->ce->name), status_line ? status_line : "", ZSTR_LEN(msg) > 0 ? ": " : "",
             ZSTR_VAL(msg), ZSTR_VAL(file), line);
 
     free(status_line);
 
-    ddtrace_string key = DDTRACE_STRING_LITERAL("error.message");
-    ddtrace_string value = {error_text, error_len};
-    zend_result result = add_tag(context, key, value);
+    ddog_add_span_meta(span, dd_str_to_CharSlice("error.message"), dd_str_to_CharSlice(error_text));
```
bwoebi, line 47:
```suggestion
    ddog_add_span_meta(span, DDOG_CHARSLICE_C("error.message"), (ddog_CharSlice) {.ptr = error_text, .len = error_len});
```
We don't have to recompute the len.

ext/exception_serialize.c
```diff
@@ -292,13 +280,13 @@ void ddtrace_create_capture_value(zval *zv, struct ddog_CaptureValue *value, con
...
-static ddog_DebuggerCapture *dd_create_frame_and_collect_locals(char *exception_id, char *exception_hash, int frame_num, ddog_CharSlice class_slice, ddog_CharSlice func_slice, zval *locals, zend_string *service_name, const ddog_CaptureConfiguration *capture_config, uint64_t time, void *context, add_tag_fn_t add_meta) {
+static ddog_DebuggerCapture *dd_create_frame_and_collect_locals(char *exception_id, char *exception_hash, int frame_num, ddog_CharSlice class_slice, ddog_CharSlice func_slice, zval *locals, zend_string *service_name, const ddog_CaptureConfiguration *capture_config, uint64_t time, ddog_SpanBytes *span) {
     char *snapshot_id = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena), uuid_len);
     ddog_snapshot_format_new_uuid((uint8_t(*)[uuid_len])snapshot_id);
 
     char msg[40];
-    int len = sprintf(msg, "_dd.debug.error.%d.snapshot_id", frame_num);
-    add_meta(context, (ddtrace_string){msg, len}, (ddtrace_string){snapshot_id, 36});
+    sprintf(msg, "_dd.debug.error.%d.snapshot_id", frame_num);
+    ddog_add_span_meta(span, dd_str_to_CharSlice(msg), dd_str_to_CharSlice(snapshot_id));
```
bwoebi, line 289: Same here, use ddog_CharSlice and provide len directly.

ext/serializer.c
```diff
@@ -1787,111 +1792,134 @@ zval *ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
-            ZEND_HASH_FOREACH_STR_KEY_VAL(Z_ARR(metrics_zv), key, tag_zv) {
+        size_t metrics_count = 0;
+
+        ddog_CharSlice *metrics_keys = ddog_span_metrics_get_keys(rust_span, &metrics_count);
+        if (metrics_keys && metrics_count > 0) {
+            for (size_t i = 0; i < metrics_count; i++) {
+                ddog_CharSlice key = metrics_keys[i];
+                if (!key.ptr || key.len == 0) {
+                    continue;
+                }
```
bwoebi, line 1875: Same here, don't hide empty keys

ext/serializer.c
```diff
@@ -1787,111 +1792,134 @@ zval *ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
...
         smart_str_free(&meta_str);
         smart_str_free(&metrics_str);
-    })
-
-    if (zend_hash_num_elements(Z_ARR(metrics_zv))) {
-        zend_hash_str_add_new(Z_ARR_P(el), ZEND_STRL("metrics"), &metrics_zv);
-    } else {
-        zend_array_destroy(Z_ARR(metrics_zv));
-    }
+    });
```
bwoebi, line 1910: Suggestion; maybe we should provide an API function in data-pipeline which returns that ("span 123: trace_id=4567, ..."), so that we can reuse that code, and fully goes to what's actually in the rust span.

cbindgen.toml
```diff
@@ -2,6 +2,11 @@ language = "C"
 tab_width = 2
 include_guard = "DDTRACE_PHP_H"
 style = "both"
+header = """// Copyright 2024-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
+struct _zend_string;
```
bwoebi, line 8:
```suggestion
header = """struct _zend_string;
```
We have the LICENSE file (which by the way dual-licenses BSD and apache2). Don't add an arbitrary license here.

ext/serializer.h
```diff
@@ -1,13 +1,13 @@
...
 #include "span.h"
-#include "ddtrace_string.h"
 
 int ddtrace_serialize_simple_array(zval *trace, zval *retval);
 int ddtrace_serialize_simple_array_into_c_string(zval *trace, char **data_p, size_t *size_p);
-size_t ddtrace_serialize_simple_array_into_mapped_menory(zval *trace, char *map, size_t size);
 
-zval *ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array);
+ddog_SpanBytes *ddtrace_serialize_span_to_rust_span(ddtrace_span_data *span, ddog_TraceBytes *trace);
+zval dd_serialiaze_rust_traces_to_zval(ddog_TracesBytes *traces);
```
bwoebi, line 10:
```suggestion
zval dd_serialize_rust_traces_to_zval(ddog_TracesBytes *traces);
```

typo

ext/serializer.c
```diff
@@ -1235,47 +1246,52 @@ static void _serialize_meta(zval *el, ddtrace_span_data *span, zend_string *serv
...
+                    ddtrace_convert_to_str(&metadata->property_repository)
+                );
+
             }
         }
     }
 
     if (get_DD_TRACE_PEER_SERVICE_DEFAULTS_ENABLED()) { // opt-in
         zend_array *peer_service_sources = ddtrace_property_array(&span->property_peer_service_sources);
-        if (zend_hash_str_exists(Z_ARRVAL_P(meta), ZEND_STRL("peer.service"))) { // peer.service is already set by the user, honor it
```
bwoebi, line 1260: Keep fetching from `meta`, this preserves the zend_string, avoiding conversion.

ext/serializer.c
```diff
@@ -958,10 +954,10 @@ static void dd_serialize_array_recursively(zend_array *target, zend_string *str,
             } ZEND_HASH_FOREACH_END();
 
             GC_UNPROTECT_RECURSION(arr);
+        } else if (convert_to_double) {
+            ddog_del_span_metrics_zstr(target, str);
         } else {
-            zval zv;
-            ZVAL_EMPTY_STRING(&zv);
-            zend_hash_update(target, str, &zv);
+            ddog_del_span_meta_zstr(target, str);
```
bwoebi, line 960: That's a behaviour change: Should insert an empty string instead.

ext/serializer.c
```diff
@@ -958,10 +954,10 @@ static void dd_serialize_array_recursively(zend_array *target, zend_string *str,
             } ZEND_HASH_FOREACH_END();
 
             GC_UNPROTECT_RECURSION(arr);
+        } else if (convert_to_double) {
+            ddog_del_span_metrics_zstr(target, str);
```
bwoebi, line 958: We should default this to 0 if empty rather than deleting.

PR #3199, opened 2025-04-09
fix(prof): disable allocation profiling when JIT is enabled on PHP 8.4

realFlowControl:
### Description

A customer reached out in #3197 with a crash while collecting a stack frame for an allocation that is done while hot function is compiled. This code seems to be there since the new JIT in PHP 8.4. So for now we are disabling allocation profiling when an active JIT is detected in PHP 8.4.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (approved):
Sad, but apparently necessary 😔

PR #3201, opened 2025-04-09
Add FakeApp for profiling benchmarking

morrisonlevi:
### Description

This adds an app for benchmarking purposes which is somewhat similar to real world things, but gathers stack samples at deterministic points.

The benchmarking script are here but not being run in CI. You'd roughly do:

```sh
cd 'profiling/benches/fakeapp'
bash +x driver.sh "$(git rev-parse --abbrev-ref HEAD)" "master"
```

The driver still needs to compare the output from `$TMPDIR/trigger-*.txt` to the ones in `./trigger-*.txt`.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/benches/fakeapp/bench.sh
+++ b/profiling/benches/fakeapp/bench.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# 3 is the number of unique samples triggered, it's just known statically for the app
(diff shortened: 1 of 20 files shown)


realFlowControl (approved):
LGTM 🎉

Issue #3208, opened 2025-04-13
[Bug]: SIGSEGV of coms-writer


junjihashimoto:
### Bug report

coms-writer causes SIGSEGV when exiting. It happens once every few days.

The backtrace of apache2+php is as follows:

```
$ gdb /usr/sbin/apache2 --core ./CoreDump
GNU gdb (Ubuntu 9.2-0ubuntu1~20.04.2) 9.2
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/sbin/apache2...
Reading symbols from /usr/lib/debug/.build-id/b9/d8d808bc1ceea80f6e55b07163298a6eb44c89.debug...
[New LWP 966983]
[New LWP 966752]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/sbin/apache2 -k start'.
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7f5153905700 (LWP 966983))]
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f5182f7e859 in __GI_abort () at abort.c:79
#2  0x00007f5182fe926e in __libc_message (action=action@entry=do_abort, fmt=0x7f5153904cf0 "", fmt@entry=0x7f5183113298 "%s\n") at ../sysdeps/posix/libc_fatal.c:156
[...]

PR #3209, opened 2025-04-13
fix: Avoid double free of stack data

junjihashimoto:
### Description

https://github.com/DataDog/dd-trace-php/issues/3208 

A double free may be occurring in ext/coms.
The cause of this seems to be that ownership of the stack is shared between the writer and the main thread, and it is suspected that the writer is freeing things that have been freed by other threads.
In this PR, we will avoid double frees by putting NULL into data for things that have been freed once, and checking this.
It's not a perfect fix, but it should avoid double frees.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi:
That might mitigate a problem a little, but the actual problem is that there are two writers. There should be only ever one single coms writer.

junjihashimoto:
@bwoebi Sorry, there is no two. It's just a duplicated command issue.('backtrace' and 'thread apply all bt')

bwoebi:
Oh, I see, I'm blind then - thanks! Looking again.

junjihashimoto:
@bwoebi 
What do you think of the reviews?
It's difficult to put valgrind into production here from a performance standpoint. On the other hand, it's possible to put this PR in and see if any problems occur.

junjihashimoto:
Hi @bwoebi ,
How can I install this patch library into production?
Is it uploaded to the CI artifacts?

bwoebi:
But the problem with your patch is that it may just mitigate the issue - ultimately an use-after-free may just segfault.

If stack is freed, accessing stack->data is not allowed.

junjihashimoto:
@bwoebi 
It is correct. It just prevents double-free from corrupting memory, and now it seems like the memory is being corrupted.
I think that the complete fix requires locks for ownership and has performance tradeoffs.

BTW, I can not see the artifact list in the artifact tab of circle ci. Am I looking in the wrong place?
image

realFlowControl:
Hey there @junjihashimoto,

build artifacts are removed from storage after 7 days. I restarted that job for you at https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/20558/workflows/71d751ac-61ea-4353-8ffd-7aec794284be so you can check this CI later for the build artifacts.

junjihashimoto:
@realFlowControl 
Thank you! I downloaded the packages as follows:
```
datadog-php-tracer_1.9.0+d475a605fd30217da25318f62e55f0de491b74a4_amd64.deb
datadog-setup.php
datadog-tracer.stubs.php
dd-library-php-1.9.0+d475a605fd30217da25318f62e55f0de491b74a4-aarch64-linux-gnu-20190902-debug.tar.gz
dd-library-php-1.9.0+d475a605fd30217da25318f62e55f0de491b74a4-aarch64-linux-gnu-20190902-zts.tar.gz
dd-library-php-1.9.0+d475a605fd30217da25318f62e55f0de491b74a4-aarch64-linux-gnu-20190902.tar.gz
```

PR #3223, opened 2025-04-28
Add http status error configuration

scott-shields-github:
### Description

This PR adds a set of configuration items for custom error flagging based on http status codes. This will allow users of the library to prevent tagging errors for things like a 4xx status code on a server, for example.

This would bring the library closer to the java tracer and some functionalities of the python and c# tracers.

The priority when evaluating spans is:
1. Manually configured error
2. Http status
3. Default - current exception based error tagging

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/configuration.c
```diff
@@ -156,6 +157,51 @@ static bool dd_parse_tags(zai_str value, zval *decoded_value, bool persistent) {
     return true;
 }
 
+static bool dd_parse_http_error_statuses(zai_str value, zval *decoded_value, bool persistent) {
+    UNUSED(persistent);
```
bwoebi, line 161: Please refer to the function above - it's forbidden to ignore the "persistent".

ext/configuration.c
```diff
@@ -156,6 +157,51 @@ static bool dd_parse_tags(zai_str value, zval *decoded_value, bool persistent) {
     return true;
 }
 
+static bool dd_parse_http_error_statuses(zai_str value, zval *decoded_value, bool persistent) {
```
bwoebi, line 160: Is these a reason to use the custom parser at all? The default parser for SET should cover all use cases here, I believe.

ext/configuration.c
scott-shields-github: taking a look

ext/configuration.c
bwoebi, line 160: Is there anything the default parser for set does not handle (which essentially simply splits on colon), that we need a custom parser here?

ext/serializer.c
```diff
@@ -1834,6 +1924,43 @@ zval *ddtrace_serialize_span_to_array(ddtrace_span_data *span, zval *array) {
         }
     }
 
+    // Figure out if an error should be tracked
```
bwoebi, line 1927: The existing code for marking entry spans as error is in dd_set_entrypoint_root_span_props_end.
Marking as error for the client integration should happen in CurlIntegration.php (in set_curl_attributes function).

bwoebi: Generally marking as error should be immediate, so that users may manually unmark as error, e.g. in onClose handlers or similar.

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -251,51 +291,61 @@ public function setup_curl_span($span) {
...
-                if (\substr_compare($key, '_time', -5) === 0) {
-                    $val *= 1000000000;
-                }
-                $span->meta["curl.{$key}"] = $val;
-            }
-        }
-
-        return $info;
-    }
+   public static function set_curl_attributes($span, $info) {
```
bwoebi, line 294: I think you messed up the indent here. (3 vs 4 spaces)

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -25,6 +25,46 @@ function addSpanDataTagFromCurlInfo($span, &$info, $tagName, $curlInfoOpt)
...
+            }
+        }
+
+        // The status code isn't in any defined error range
+        return false;
+    } else {
+        // Default behavior
+        return ($statusCode >= 400);
+    }
+}
```
bwoebi, line 66: Can we add this common code to DDTrace\Integrations\HttpClientIntegrationHelper please? Thanks!

ext/configuration.h
```diff
@@ -247,7 +247,9 @@ enum ddtrace_sampling_rules_format {
     CONFIG(SET, DD_DYNAMIC_INSTRUMENTATION_REDACTED_TYPES, "", .ini_change = zai_config_system_ini_change)     \
     CONFIG(INT, DD_TRACE_BAGGAGE_MAX_ITEMS, "64")                                                              \
     CONFIG(INT, DD_TRACE_BAGGAGE_MAX_BYTES, "8192")                                                            \
-    CONFIG(BOOL, DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED, "false")                                             \
+    CONFIG(BOOL, DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED, "false")                                            \
+    CONFIG(SET, DD_TRACE_HTTP_CLIENT_ERROR_STATUSES, "", .ini_change = zai_config_system_ini_change)           \
+    CONFIG(SET, DD_TRACE_HTTP_SERVER_ERROR_STATUSES, "", .ini_change = zai_config_system_ini_change)           \
```
bwoebi, line 252:
```suggestion
    CONFIG(SET, DD_TRACE_HTTP_CLIENT_ERROR_STATUSES, "500-599", .ini_change = zai_config_system_ini_change)           \
    CONFIG(SET, DD_TRACE_HTTP_SERVER_ERROR_STATUSES, "500-599", .ini_change = zai_config_system_ini_change)           \
```
Then we don't need a fallback behaviour at all.

ext/serializer.c
```diff
@@ -1152,10 +1152,77 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+        // Only check status codes if not ignoring errors
+        if (!ignore_error) {
+            bool is_error = false;
+            bool is_custom_error = false;
+
+            // Get server error configuration
+            zend_array *cfg = get_DD_TRACE_HTTP_SERVER_ERROR_STATUSES();
+
+            // If configuration exists and has entries, check against it
+            if (cfg && zend_hash_num_elements(cfg) > 0) {
```
bwoebi, line 1164:
```suggestion
```
We can get rid of that check, the configs are guaranteed to be non-NULL.

ext/serializer.c
```diff
@@ -1152,10 +1152,77 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+                } ZEND_HASH_FOREACH_END();
+
+                // If no custom match, still check for >=500 for backward compatibility
+                if (!is_error) {
+                    is_error = (status >= 500);
+                }
+            } else {
+                // Fallback to original behavior if no configuration is set
+                is_error = (status >= 500);
+            }
```
bwoebi, line 1201:
```suggestion
```
We don't need the fallback.

ext/serializer.c
```diff
@@ -1152,10 +1152,77 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+                    if (is_custom_error) {
+                        // New behavior for custom-configured errors
+                        ZVAL_STR(value, zend_string_init(ZEND_STRL("http_error"), 0));
+
+                        // Add error message and flag
+                        zval error_msg_zv = {0}, *msg_value;
+                        if ((msg_value = zend_hash_str_add(meta, ZEND_STRL("error.msg"), &error_msg_zv))) {
+                            char error_msg[50];
+                            snprintf(error_msg, sizeof(error_msg), "HTTP %d Error", status);
+                            ZVAL_STR(msg_value, zend_string_init(error_msg, strlen(error_msg), 0));
```
bwoebi, line 1215:
```suggestion
                            ZVAL_STR(msg_value, zend_strpprintf(0, "HTTP %d Error", status));
```
simpler :-)

ext/serializer.c
```diff
@@ -1152,10 +1152,77 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+                }
+            } else {
+                // Fallback to original behavior if no configuration is set
+                is_error = (status >= 500);
+            }
+
+            if (is_error) {
+                zval zv = {0}, *value;
+                if ((value = zend_hash_str_add(meta, ZEND_STRL("error.type"), &zv))) {
+                    if (is_custom_error) {
```
bwoebi, line 1206: I think we can just change this, and use `HTTP %d error` for everything.

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -251,51 +291,61 @@ public function setup_curl_span($span) {
...
+
+       $span->peerServiceSources = HttpClientIntegrationHelper::PEER_SERVICE_SOURCES;
+
+       addSpanDataTagFromCurlInfo($span, $info, Tag::HTTP_STATUS_CODE, 'http_code');
+
+       // Check if the status code should be marked as an error based on configuration
+       if (isset($info['http_code']) && !empty($info['http_code']) && !isset($span->meta[Tag::ERROR])) {
+           $statusCode = (int)$info['http_code'];
+           if (self::isClientError($statusCode)) {
+               $span->meta[Tag::ERROR] = 1;
```
bwoebi, line 322:
```suggestion
       if (isset($info['http_code']) && !empty($info['http_code']) && !isset($span->meta[Tag::ERROR_MSG])) {
           $statusCode = (int)$info['http_code'];
           if (self::isClientError($statusCode)) {
```
Same for the other integrations - we're not supposed to set Tag::ERROR manually.

src/DDTrace/Integrations/Guzzle/GuzzleIntegration.php
```diff
@@ -50,14 +50,38 @@ function (SpanData $span, $args, $retval) use ($integration) {
...
-                        $span->meta[Tag::HTTP_STATUS_CODE] = $response->getStatusCode();
+                        $statusCode = $response->getStatusCode();
+                        $span->meta[Tag::HTTP_STATUS_CODE] = $statusCode;
+
+                        // Mark as error if status code matches configuration
+                        if (self::isClientError($statusCode)) {
+                            $span->meta[Tag::ERROR] = 1;
+                            $span->meta[Tag::ERROR_TYPE] = 'http_error';
+                            $span->meta[Tag::ERROR_MSG] = "HTTP $statusCode: " . $response->getReasonPhrase();
+                        }
```
bwoebi, line 61: We should probably simplify this to `HttpClientIntegrationHelper::setClientError($span, $statusCode);`

ext/configuration.c
```diff
@@ -7,6 +7,7 @@
 #include "json/json.h"
 #include "sidecar.h"
 #include <components/log/log.h>
+#include <ctype.h>
```
bwoebi, line 10:
```suggestion
```
Leftover I think?

ext/serializer.c
```diff
@@ -1152,10 +1152,77 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+                            if (sscanf(s, "%d-%d", &start, &end) == 2) {
+                                if (status >= start && status <= end) {
+                                    is_error = true;
+                                    is_custom_error = true;
+                                    break;
+                                }
+                            }
+                        } else {
+                            // Single status code
+                            int code = atoi(s);
```
bwoebi, line 1184:
```suggestion
                            zend_long code = ZEND_ATOL(s);
```
We use the generic version from zend here

ext/serializer.c
```diff
@@ -1152,10 +1152,77 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+                        if (dash) {
+                            // Range like "500-599"
+                            int start, end;
+                            if (sscanf(s, "%d-%d", &start, &end) == 2) {
+                                if (status >= start && status <= end) {
+                                    is_error = true;
+                                    is_custom_error = true;
+                                    break;
+                                }
+                            }
```
bwoebi, line 1181:
```suggestion
                        // Range like "500-599"
                        int start, end;
                        if (sscanf(s, "%d-%d", &start, &end) == 2) {
                            if (status >= start && status <= end) {
                                is_error = true;
                                break;
                            }
                        }
```
This is simpler?

scott-shields-github:
> Thanks for the contribution! I've littered the code with suggestions, but it already looks quite good :-) Tests are obviously broken, but we can regenerate the snapshots.

Thanks so much for the review and suggestions!

ext/serializer.c
```diff
@@ -1152,10 +1152,48 @@ static void dd_set_entrypoint_root_span_props_end(zend_array *meta, int status,
...
+            } ZEND_HASH_FOREACH_END();
+
+            if (is_error) {
+                zval zv = {0}, *value;
+                if ((value = zend_hash_str_add(meta, ZEND_STRL("error.type"), &zv))) {
+                    ZVAL_STR(value, zend_string_init(ZEND_STRL("http_error"), 0));
+
+                    // Add error message and flag
+                    zval error_msg_zv = {0}, *msg_value;
+                    if ((msg_value = zend_hash_str_add(meta, ZEND_STRL("error.msg"), &error_msg_zv))) {
```
bwoebi, line 1193: It's called `error.message`, not error.msg (also misspelt in tests/ext/error_status_configuration/http_error_status_client_configuration.phpt)

bwoebi:
Thanks, the code looks fine now - we can have a look at the tests.

scott-shields-github:
> Thanks, the code looks fine now - we can have a look at the tests.

Anything I can do on my end? I've tried to build run the tests in the container, but I've been having trouble with getting it started with docker compose. I've only been able to run the 8.3-buster container and run the tests manually.

bwoebi:
`docker compose run --rm 8.4-buster bash` and then `make test_c` - and then the tests run for me.
Does that not work for you? What errors do you get? We definitely like to make it easy for everyone to contribute!

scott-shields-github:
> make test_c

I finally got the env up and running, by making no changes and just trying many times. I think it could be a local issue for me, I'll try on another machine this weekend to test. But, now that it's running, I see my tests pass and a few failures that do not look related (related to a container service it looks like). There's a couple other failures that look like they are common?

```
Fatal error: Uncaught Exception: wait for replay timeout
in request_replayer.inc:78
```

bwoebi:
They are not common and not supposed to happen.
The request_replayer is responsible for mirroring back sent traces to verify them.

Are you maybe memory constrained? These docker images tend to require quite a bit of memory.

scott-shields-github:
> They are not common and not supposed to happen. The request_replayer is responsible for mirroring back sent traces to verify them.
> 
> Are you maybe memory constrained? These docker images tend to require quite a bit of memory.

I'm at 22GB of memory for Rancher (dockerd engine - M2 macbook pro). I'll tinker with it a little more this weekend and see if I can find anything more helpful. Originally, I couldn't get all the dependencies to even attempt to load. It would always be different ones left out (mysql/kong/spanner/etc...) as well so I'm leaning towards and issue with my env.

scott-shields-github:
Locally, if I force an x86_64 build of 8.3-buster, I don't get those replay timeout errors anymore. I only get two failures seemingly unrelated to these changes:

```
(error_propagated.phpt)
001- "error.message": "Uncaught Exception (500): An exception occurred in %serror_propagated.php:6",
001+ "error.message": "Uncaught Exception (500): An exception occurred in \/...\/error_propagated.php:6",
```
and

```
(segfault_backtrace_disabled.phpt)
001- (Segmentation fault.*)|(Termsig=11)
001+ Continued after segfault?!
```

bwoebi:
I'm not sure what's wrong here with the CI, maybe try rebasing instead of merging?

scott-shields-github:
No luck with a rebase either. When I merged master at first I think I forgot to update the submodule before pushing. But it's up to date locally and i've tried forcing that a couple times. I was going to try to manually reinit it to see if it helps

scott-shields-github:
> I just ran CI myself and will merge it. The PR looks great, thanks for it!
> 
> Sorry for the delay, though the next release anyway is only tomorrow :-)

Great! Sorry, for the delay, I've been a little too busy to get back to it. Thanks so much for all the help!

PR #3225, opened 2025-04-28
Fix initial double remote config reading

bwoebi:
This also prevents an use-after-free given that only ddtrace_rinit_remote_config will initialize `DDTRACE_G(active_rc_hooks)`.

This use-after-free only manifests sometimes on consecutive requests where both requests end up using live-debugger resources.


morrisonlevi (approved):
Being honest, I don't know what's going on here, but it's supposed to fix a bug, so I'll trust you.

PR #3227, opened 2025-05-02
feat(prof): configurable allocation sampling distance

realFlowControl:
### Description

This makes the allocation profiling sampling distance (formerly fixed 4MB) configurable through the `DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE` environment variable or `datadog.profiling.allocation_sampling_distance` PHP INI setting.

> 4MB should be enough for anyone - 😉 

In most cases this is true, but if your application is heavily allocating lots of memory you'd see elevated overhead from the allocation profiler. This feature gives you a level to adjust in case you do want allocation profiling, but see elevated overhead because the application you are running is allocating a lot of memory.

### Reviewer checklist
- [ ] Test coverage seems ok.

PROF-11753


profiling/src/config.rs
```diff
@@ -942,6 +960,18 @@ pub(crate) fn minit(module_number: libc::c_int) {
                     displayer: None,
                     env_config_fallback: None,
                 },
+                zai_config_entry {
+                    id: transmute::<ConfigId, u16>(ProfilingAllocationSamplingDistance),
+                    name: ProfilingAllocationSamplingDistance.env_var_name(),
+                    type_: ZAI_CONFIG_TYPE_CUSTOM,
+                    default_encoded_value: ZaiStr::literal(b"4194304\0"), // 4194304 = 4 MB
```
morrisonlevi, line 967: Maybe leave the name of the extracted constant  in a comment here so if anyone changes it, they remember to change it here too.

morrisonlevi (approved):
I have some nits with the PR, but I do have a bigger concern: if we add dynamic sampling, how are these supposed to interact? I wonder how Python has done things recently, we should maybe align, or if they need to change something, better to do it now before it gets widely adopted. Maybe `PROFILING_ALLOCATION_SAMPLING_DISTANCE=dynamic`?

profiling/src/config.rs
```diff
@@ -27,6 +27,7 @@ pub struct SystemSettings {
     pub profiling_endpoint_collection_enabled: bool,
     pub profiling_experimental_cpu_time_enabled: bool,
     pub profiling_allocation_enabled: bool,
+    pub profiling_allocation_sampling_distance: u32,
```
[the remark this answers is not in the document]
realFlowControl: I assume so yes, thats why u32 is fine 😉 
In that one specific case they'd have got one sample

realFlowControl:
Thats exactly the idea, `DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE` for now is the distance in byte, in the future this will be `unsigned int` (distance in byte) or the string `dynamic` where `dynamic` would be the new default, once we have dynamic sampling implemented.

profiling/src/profiling/mod.rs
```diff
@@ -1582,6 +1582,7 @@ mod tests {
             profiling_endpoint_collection_enabled: false,
             profiling_experimental_cpu_time_enabled: false,
             profiling_allocation_enabled: false,
+            profiling_allocation_sampling_distance: 4194304,
```
morrisonlevi, line 1585: `DEFAULT_ALLOCATION_SAMPLING_INTERVAL`?

profiling/tests/phpt/allocation_bind_static_01.phpt
```diff
@@ -13,9 +13,8 @@ https://github.com/php/php-src/commit/ec54ffad1e3b15fedfd07f7d29d97ec3e8d1c45a
 But there's an `zend_array_dup` operation which can occur before this the call
 to `SAVE_OPLINE()`, so if the allocation profiler triggers there, it will
 access the invalid opline (non-null and invalid).
-
-TODO: note that this probably will not fail, because we have to take a sample
-at exactly the right time, and currently we don't have a way to force this.
+--ENV--
+DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE=1
```
morrisonlevi, line 17: ❤️

profiling/src/profiling/mod.rs
realFlowControl: Nice catch

PR #3228, opened 2025-05-02
fix(profiling) empty internal metadata

realFlowControl:
### Description

The `Uploader::create_internal_metadata()` would always return `None`...

PROF-11749

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Oops. Looks like a correct fix.

PR #3230, opened 2025-05-05
add a job that check all other job status

cbeauchesne:
### Description

Add a job in the CI that runs in PR. The job will be a success if all other jobs are skipped/success, and fails otherwise

### Motivation

While it's possible to enforce a green CI policy using GitHub's native "required status checks" feature, doing so requires explicitly listing all job names under branch protection rules. This approach has two key drawbacks:

- It does not support optional jobs
- It introduces ongoing maintenance overhead as the job list evolves

This jobs will check ALL other job, and we'll be able to set this as a requirement. The action used offers a `ignored-name-patterns` parameters`, I added few job that failed on that PR, let me know if others should be added.

the plan is to merge this PR, wait few days to be sure that everything is fine. Then add it as a requirement.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -19,3 +19,12 @@ jobs:
         uses: DataDog/ensure-ci-success@v1
         with:
           initial-delay-seconds: "100"  # what is the longest job that always run ?
+          max-retries: "60"
+          ignored-name-patterns: |
+            dd-gitlab/build
+            ci/circleci: test-test_extension_ci-8.4-medium+
+            ci/circleci: integration_snapshots-test_integrations-8.1
+            ci/circleci: integration_snapshots-test_integrations-8.2
+            ci/circleci: integration_snapshots-test_integrations-8.3
+            ci/circleci: integration_snapshots-test_integrations_coverage-8.2
+
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -1,0 +1,21 @@
+name: Check Pull Request CI Status
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+
+permissions:
+  checks: read
+  statuses: read
+
+jobs:
+  all-jobs-are-green:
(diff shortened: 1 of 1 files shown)


.github/workflows/all-green.yml
```diff
@@ -0,0 +1,30 @@
...
+permissions:
+  checks: read
+  statuses: read
+
+jobs:
+  all-jobs-are-green:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Ensure CI Success
+        uses: DataDog/ensure-ci-success@v1
```
marmichalski, line 19: We've built something similar to replace https://github.com/upsidr/merge-gatekeeper in our CI - https://github.com/lendable/sloth - maybe you would be interested in joining the effort? 😊

cbeauchesne: gosh, I crawl the web to find an action that fits my needs, and I missed yours 🤕 .

Unfortunately, I won't be able to join, as go is chineese to me.

Though, I'll put some suggestion easy to tackle that would have been a blocker for me. If at  some point it's a good fit, I'll consider using yours!

Thank's a lot for the head's up!

bwoebi (changes requested):
I'm not going to merge this currently.

This request changes is not a request for changes _today_, but for changes in the future, around ignored-name-patterns.

The reason is that we're in the middle of migration to gitlab CI, which currently has more flaky and non-working tests, which we might have to merge within the next two weeks, even with not all issues resolved.
This requirement will then be ugly and with possibly many exceptions.
I'd like to revisit this PR again within a month though.

.github/workflows/all-green.yml
marmichalski: No worries, ours is the latter one though (sloth), which is in TypeScript :)

bwoebi:
@cbeauchesne This is pointless though as within two weeks all ci/circleci will disappear, leaving this check effectively checking nothing then.
I'd rather consider this PR when the time comes to do it properly.

PR #3231, opened 2025-05-06
internal: add opcache tags in crash report

morrisonlevi:
[PROF-11759](https://datadoghq.atlassian.net/browse/PROF-11759)

### Description

When looking through the crash tracker reports, it would be helpful to know if the JIT is active. Since the JIT is not functional if the buffer is zero, ~and it also defaults to 0, this gives us a pretty good signal whether it's enabled.~ Edit: the documentation is not up-to-date for the values in PHP 8.4 😞

These tags are what they were set to at first request, and some of them are `INI_ALL` so it's possible (and even probable) that there are changes at runtime. With the `INI_SYSTEM` settings, we do short-cut to avoid adding tags when JIT isn't enabled. We could possibly revisit this to still tag that it's off. That way we can distinguish between "don't know" (such as an old version of the product) and "know it's off."

The crashtracking group says the cardinality of these tags is fine.

### Reviewer checklist
- [x] Test coverage seems ok.

[PROF-11759]: https://datadoghq.atlassian.net/browse/PROF-11759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iamluc (commented):
We should probably collect a few more, like `opcache.jit`, `opcache.enabled`, ... 
Since PHP 8.4,  `opcache.jit` is the one used to enable/disable JIT.

Here is what we do in the SSI loader: https://github.com/DataDog/dd-trace-php/blob/0697a7f4a137ce989d5a60a6f4ab09f9e4280b46/loader/dd_library_loader.c#L112-L145

morrisonlevi:
> We should probably collect a few more, like opcache.jit, opcache.enabled, ...
Since PHP 8.4, opcache.jit is the one used to enable/disable JIT.

The trouble is that `opcache.enable` and `opcache.jit` are `INI_ALL. The tags (at the moment) are applied during `ddtrace_signals_first_rinit`, so there's not really a good way at the moment to update them. I'd rather have trusted tags than probably-right-but-not-all-the-time ones. Since `opcache.jit_buffer_size` is `INI_SYSTEM`, we can have confidence here. Additionally, `opcache.jit` and `opcache.enable` default to something which would be considered on, whereas `opcache.jit_buffer_size` defaults to being off. That means if the jit_buffer_size is not zero, it's probably on.

This PR is a bit too simplistic, unfortunately, because PHP allows short-hand notation other than just bytes, for example `0M` is theoretically possible.

morrisonlevi:
I've adjusted it to parse quantities and restrict this to PHP 8+ (which is when it was added).

morrisonlevi:
Okay. So I'll do something similar to what @iamluc suggested afterall. Boo. Big sad.

ext/signals.c
```diff
@@ -102,6 +107,138 @@ static bool ddtrace_crashtracker_check_result(ddog_VoidResult result, const char
...
+        zend_long quantity = zend_ini_parse_quantity(value, &errstr);
+        if (errstr) zend_string_release(errstr);
+#else
+        zend_long quantity = zend_atol(ZSTR_VAL(value), ZSTR_LEN(value));
+#endif
+        if (quantity <= 0) {
+            return;
+        }
+        jit_buffer_size.ptr = ZSTR_VAL(value);
+        jit_buffer_size.len = ZSTR_LEN(value);
```
bwoebi, line 202:
```suggestion
        jit_buffer_size = dd_zend_string_to_CharSlice(value);
```
nit (same for enable_cli)

ext/signals.c
```diff
@@ -102,6 +107,138 @@ static bool ddtrace_crashtracker_check_result(ddog_VoidResult result, const char
...
+    }
+    return NULL;
+}
+#endif
+
+// Fetches certain opcache tags and adds them with the pattern of php.opcache.*.
+static void ddtrace_crasht_add_opcache_tags(ddog_Vec_Tag *tags) {
+#if PHP_VERSION_ID >= 80000
+    // If opcache isn't loaded, don't attach any opcache tags.
+    if (!ddtrace_crasht_find_zend_extension("Zend OPcache")) {
```
bwoebi, line 176: There's a function `zend_get_extension`. Is there a reason you do the iteration manually?

ext/signals.c
```diff
@@ -102,6 +107,138 @@ static bool ddtrace_crashtracker_check_result(ddog_VoidResult result, const char
...
+    ddog_CharSlice key
+) {
+    const ddog_CharSlice PREFIX = DDOG_CHARSLICE_C("php.");
+    ZEND_ASSERT(key.len > PREFIX.len && memcmp(key.ptr, PREFIX.ptr, PREFIX.len) == 0);
+
+    ddog_CharSlice ini = {
+        .ptr = key.ptr + PREFIX.len,
+        .len = key.len - PREFIX.len,
+    };
+    zend_string *value = ddtrace_crasht_ini_get_value(ini);
```
bwoebi, line 144:
```suggestion
    const char *value = zend_ini_string(key.ptr + PREFIX.len, key.len - PREFIX.len, 0);
```
(using zend_ini_str and polyfilling it in compatibility.h would be even better)

ext/signals.c
```diff
@@ -102,6 +107,138 @@ static bool ddtrace_crashtracker_check_result(ddog_VoidResult result, const char
...
+    return value;
+}
+
+// Pass in a key like "php.opcache.enable" and "php." will get stripped off,
+// and that's what the INI setting will be.
+static void ddtrace_crasht_add_opcache_tag(
+    ddog_Vec_Tag *tags,
+    ddog_CharSlice key
+) {
+    const ddog_CharSlice PREFIX = DDOG_CHARSLICE_C("php.");
```
bwoebi, line 137:
```suggestion
    const char *PREFIX = "php.";
```
Let's keep it simple and use strlen() for the length. We will just rely on the compiler to recognize that strlen(PREFIX) is 4.

ext/signals.c
```diff
@@ -102,6 +107,138 @@ static bool ddtrace_crashtracker_check_result(ddog_VoidResult result, const char
...
+        }
+    }
+
+    ddtrace_crasht_add_opcache_tag(tags, DDOG_CHARSLICE_C("php.opcache.jit"));
+
+    ddog_CharSlice key = DDOG_CHARSLICE_C("php.opcache.jit_buffer_size");
+    ddog_Vec_Tag_PushResult result = ddog_Vec_Tag_push(tags, key, jit_buffer_size);
+    if (UNEXPECTED(result.tag != DDOG_VEC_TAG_PUSH_RESULT_OK)) {
+        ddtrace_crasht_failed_tag_push(&result.err, key);
+    }
```
bwoebi, line 236: You always repeat these 4 lines ... maybe you should just do one `ddtrace_crasht_push_tag()` function (which includes the push failure handling).

ext/signals.c
morrisonlevi: Hmm, zend_ini_string returns a `char*`. Wish there was an API for passing a ptr + len for the INI but getting back a `zend_string *`. In any case, yeah, we can avoid the allocation.

morrisonlevi: I've polyfilled it for older versions. However, note that I'm not using it on PHP <8 so... it may not compile. It looks like it should, based on looking at API functions on heap.space.

bwoebi: Yes, that polyfill will work perfectly on PHP 7.

bwoebi (approved):

PR #3235, opened 2025-05-06
style(prof): fix tons of clippy lints

morrisonlevi:
### Description

Many clippy lints have accumulated. This fixes all of the lints on PHP 8.3/8.4 and on Rust 1.78. It also fixes many lints from Rust 1.86 and nightly. Here are some of the things fixed:

- Using Rust's CStr literal syntax `c""` and converting some engine members from `*const u8` to `*const c_char`.
- Using `core::ptr::eq` for certain pointer comparisons.
- Adjust use of mutable statics to use pointers, see [Disallow references to static mut](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html#disallow-references-to-static-mut).

And some non-clippy items:
- Converts `/*` style comments to `//`, which is by far more popular.
- Converts `Safety:` comments to `SAFETY:`, which is mostly what we use everywhere else.
- Updates some of the historical comments to be more accurate in the present.

There were historical reasons for not running clippy in CI. Over time think these have all been fixed, helped by advances in Rust. We should enable clippy soon for profiling. On the versions listed, the only lints on current nightly are [unpredictable function pointer comparisons](https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html). I'm not sure what to do here, other than maybe allow them. But since this is nightly, it may not land in its current state anyway, so I'm not going to sweat it.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/allocation_le83.rs
+++ b/profiling/src/allocation/allocation_le83.rs
@@ -383,7 +383,6 @@ unsafe extern "C" fn alloc_prof_free(ptr: *mut c_void) {
     tls_zend_mm_state_get!(free)(ptr);
 }
 
-#[export_name = "ddog_php_prof_alloc_prof_prev_free"]
 unsafe fn alloc_prof_prev_free(ptr: *mut c_void) {
     // Safety: `ZEND_MM_STATE.prev_custom_mm_free` will be initialised in
     // `alloc_prof_rinit()` and only point to this function when
--- a/profiling/src/allocation/allocation_le83.rs
+++ b/profiling/src/allocation/allocation_le83.rs
@@ -203,10 +203,10 @@ pub fn alloc_prof_rinit() {
 }
 
 pub fn alloc_prof_rshutdown() {
-    // If `is_zend_mm()` is true, the custom handlers have been reset to `None`
-    // already. This is unexpected, therefore we will not touch the ZendMM
+    // If `is_zend_mm()` is true, the custom handlers have already been reset
+    // to `None`. This is unexpected, therefore we will not touch the ZendMM
     // handlers anymore as resetting to prev handlers might result in segfaults
-    // and other undefined behaviour.
+    // and other undefined behavior.
     if is_zend_mm() {
         return;
     }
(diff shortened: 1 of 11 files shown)


profiling/src/profiling/stack_walking.rs
```diff
@@ -461,6 +461,8 @@ mod detail {
 
 pub use detail::*;
 
+// todo: this should be feature = "stack_walking_tests" but it seemed to
+//       cause a failure in CI to migrate it.
 #[cfg(all(test, stack_walking_tests))]
 mod tests {
```
morrisonlevi, line 467: Example of the error ([job](https://app.circleci.com/pipelines/github/DataDog/dd-trace-php/20870/workflows/6c0f6e63-502d-45d1-b234-c798f6c98fd1/jobs/6930919)):

```
  = note: /mnt/ramdisk/cargo/release/deps/datadog_php_profiling-a15e7ad77c649749.datadog_php_profiling.b94f957b4ddf90d7-cgu.0.rcgu.o: In function `{closure#0}':
          /home/circleci/datadog/profiling/src/profiling/stack_walking.rs:263: undefined reference to `zend_flf_functions'
          collect2: error: ld returned 1 exit status
          
  = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib)

error: could not compile `datadog-php-profiling` (lib test) due to 1 previous error
```

profiling/src/allocation/allocation_le83.rs
```diff
@@ -374,12 +377,13 @@ unsafe fn alloc_prof_orig_alloc(len: size_t) -> *mut c_void {
...
 /// This function exists because when calling `zend_mm_set_custom_handlers()`,
 /// you need to pass a pointer to a `free()` function as well, otherwise your
-/// custom handlers won't be installed. We can not just point to the original
+/// custom handlers won't be installed. We cannot just point to the original
 /// `zend::_zend_mm_free()` as the function definitions differ.
 unsafe extern "C" fn alloc_prof_free(ptr: *mut c_void) {
     tls_zend_mm_state_get!(free)(ptr);
 }
 
+#[export_name = "ddog_php_prof_alloc_prof_prev_free"]
```
realFlowControl, line 386:
```suggestion
```

Do we need this and if so, why do we not need to change the exported symbol name for the other functions?

realFlowControl (approved):

profiling/src/allocation/allocation_le83.rs
morrisonlevi: No, I don't even remember why I put it on there, it was during experimentation with some error I was getting. Removed.

PR #3237, opened 2025-05-08
fix trailing whitespace

realFlowControl:
### Description

AFAIK there is only a small number of files where we do want to have no final newline at the end of the file, so far I know it is the `VERSION` file, but all other files should have a final newline character.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,8 +3,11 @@ root = true
 [*]
 charset = utf-8
 indent_style = space
-insert_final_newline = false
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
I think that's right.

morrisonlevi (approved):
VERSION is the only one I'm aware of, may want to wait a bit to merge for Bob to also chime in.

PR #3239, opened 2025-05-09
Update SLO metrics

estringana:
### Description
Update SLO metrics:

- Adding rasp.error
- Updating waf_error
- Adding required tags

[[APPSEC-57288]]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-57288]: https://datadoghq.atlassian.net/browse/APPSEC-57288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appsec/src/helper/subscriber/waf.cpp
+++ b/appsec/src/helper/subscriber/waf.cpp
@@ -417,10 +417,11 @@ void instance::listener::call(
     dds::parameter_view &data, event &event, const std::string &rasp_rule)
 {
     ddwaf_result res;
+    DDWAF_RET_CODE code;
     auto run_waf = [&]() {
         dds::parameter_view *persistent = rasp_rule.empty() ? &data : nullptr;
         dds::parameter_view *ephemeral = rasp_rule.empty() ? nullptr : &data;
-        code_ = ddwaf_run(
+        code = ddwaf_run(
             handle_, persistent, ephemeral, &res, waf_timeout_.count());
     };
 
@@ -436,7 +437,7 @@ void instance::listener::call(
             parameter_to_json(parameter_view{res.events}),
             res.total_runtime / millis);
         SPDLOG_DEBUG("Waf response: code {} - actions {} - derivatives {}",
-            fmt::underlying(code_),
+            fmt::underlying(code),
             parameter_to_json(parameter_view{res.actions}),
             parameter_to_json(parameter_view{res.derivatives}));
     } else {
@@ -473,8 +474,10 @@ void instance::listener::call(
             rasp_metrics_[rasp_rule].timeouts++;
         }
         rasp_metrics_[rasp_rule].evaluated++;
-        if (code_ == DDWAF_MATCH) {
+        if (code == DDWAF_MATCH) {
             rasp_metrics_[rasp_rule].matches++;
+        } else if (code != DDWAF_OK) {
+            rasp_metrics_[rasp_rule].errors++;
         }
     }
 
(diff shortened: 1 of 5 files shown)


appsec/src/helper/subscriber/waf.cpp
```diff
@@ -417,11 +417,10 @@ void instance::listener::call(
     dds::parameter_view &data, event &event, const std::string &rasp_rule)
 {
     ddwaf_result res;
-    DDWAF_RET_CODE code;
     auto run_waf = [&]() {
         dds::parameter_view *persistent = rasp_rule.empty() ? &data : nullptr;
         dds::parameter_view *ephemeral = rasp_rule.empty() ? nullptr : &data;
-        code = ddwaf_run(
+        code_ = ddwaf_run(
```
Anilm3, line 423: This code will be replaced on every run, if one has an error (for some reason), and the next one succeeds, you'll send the success code instead of the error.

cataphract: You need to conceptually store a map from error code to a count instead. You can ignore DDWAF_OK and DDWAF_MATCH.

If there is more than one error code, you'll submit multiple metric counts.

cataphract (changes requested):

appsec/src/helper/subscriber/waf.cpp
cataphract: Also this should apply only to rasp runs

estringana: Sorted, I got confused with the part of the document talking about the span tag instead

Anilm3: The error code still has to be reported as a tag in non-RASP calls using `appsec.waf.error`.

estringana: > The error code still has to be reported as a tag in non-RASP calls using `appsec.waf.error`.

That one will be done on the non SLO ones

appsec/src/helper/subscriber/waf.cpp
```diff
@@ -527,7 +529,7 @@ void instance::listener::submit_metrics(
         tags.add("waf_timeout", "true");
     }
     if (waf_run_error_) {
-        tags.add("waf_run_error", "true");
+        tags.add("waf_error", "true");
```
cataphract, line 532: the doc says "waf_error" sould contain the numerical error returned by ddwaf_run, or -127 if the error happened in the binding. Which is why it's not enough to store a count earlier.

cataphract (changes requested):

appsec/src/helper/subscriber/waf.cpp
estringana: There are two different tags for `waf_error` if you check https://docs.google.com/document/d/1D4hkC0jwwUyeo0hEQgyKP54kM1LZU98GL8MaP60tQrA/edit?tab=t.0
This one is the one for `appsec.waf.requests` and therefore it's ok.

The one you're refering to(`appsec.waf.error`) is not implemented on this PR

Anilm3: You're right, but below you added `appsec.rasp.error` without including the code as a tag

estringana: Yes because that tag is not part of the SLO

cataphract (approved):

Anilm3 (approved):

PR #3248, opened 2025-05-13
style(prof): use local_key_cell_methods

morrisonlevi:
### Description

Rust added these in v1.73. Cleans up our code a bit. I didn't want mingle them in with a bug fix.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/io.rs
+++ b/profiling/src/io.rs
@@ -245,30 +245,19 @@ unsafe extern "C" fn observed_poll(fds: *mut libc::pollfd, nfds: u64, timeout: c
     let duration = start.elapsed();
 
     if !fds.is_null() {
+        let duration_nanos = duration.as_nanos() as u64;
         if (*fds).revents & 1 == 1 {
             // requested events contains reading
-            SOCKET_READ_TIME_PROFILING_STATS.with(|cell| {
-                let mut io = cell.borrow_mut();
-                io.track(duration.as_nanos() as u64)
-            });
+            SOCKET_READ_TIME_PROFILING_STATS.with_borrow_mut(|io| io.track(duration_nanos));
         } else if (*fds).revents & 4 == 4 {
             // requested events contains writing
-            SOCKET_WRITE_TIME_PROFILING_STATS.with(|cell| {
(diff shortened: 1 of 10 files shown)


profiling/src/io.rs
```diff
@@ -491,14 +433,10 @@ unsafe extern "C" fn observed_fread(
...
-    FILE_READ_TIME_PROFILING_STATS.with(|cell| {
-        let mut io = cell.borrow_mut();
-        io.track(duration.as_nanos() as u64)
-    });
-    FILE_READ_SIZE_PROFILING_STATS.with(|cell| {
-        let mut io = cell.borrow_mut();
-        io.track(len as u64)
-    });
+    FILE_READ_TIME_PROFILING_STATS.with_borrow_mut(|io| io.track(duration.as_nanos() as u64));
+    if len > 0 {
```
morrisonlevi, line 437: Most other place guard this with a `len > 0` check, but this one didn't. Added for clarity, let me know if it's not desired.

bwoebi: This seems to make sense :-)

profiling/src/io.rs
```diff
@@ -723,11 +649,7 @@ impl<C: IOCollector> IOProfilingStats<C> {
     }
 
     fn track(&mut self, value: u64) {
-        let zend_thread = REQUEST_LOCALS.with(|cell| {
-            let locals = cell.borrow();
-            !locals.vm_interrupt_addr.is_null()
-        });
-        if !zend_thread {
+        if !REQUEST_LOCALS.with_borrow(|locals| !locals.vm_interrupt_addr.is_null()) {
```
bwoebi, line 652: You may want to keep the variable for documentation purposes. !vm_interrupt_addr.is_null() does not convey that a "is zend_thread" check was meant.

bwoebi (approved):
Looks good :-)

profiling/src/io.rs
realFlowControl: Your patch is correct, thanks

PR #3250, opened 2025-05-14
Truncate input

estringana:
### Description

This PR adds SLO tag `input_truncated` and trucates input data sent to waf accordingly to the following limitations:
- Max string length by default 4096
- Max container depth by default20
- Max container size by default 256

[[APPSEC-57288](https://datadoghq.atlassian.net/browse/APPSEC-57288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ)]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-57288]: https://datadoghq.atlassian.net/browse/APPSEC-57288?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appsec/src/extension/msgpack_helpers.c
```diff
@@ -63,25 +77,40 @@ void dd_mpack_write_zval(mpack_writer_t *nonnull w, zval *nullable zv)
...
     }
 
-    _mpack_write_zval(w, zv);
+    _mpack_write_zval(w, zv, 0);
 }
 
 // NOLINTNEXTLINE(misc-no-recursion)
-void dd_mpack_write_array(
-    mpack_writer_t *nonnull w, const zend_array *nullable arr)
+void _dd_mpack_write_array(
```
Anilm3, line 84:
```suggestion
static void _mpack_write_array(
```

appsec/src/extension/msgpack_helpers.c
```diff
@@ -63,25 +77,40 @@ void dd_mpack_write_zval(mpack_writer_t *nonnull w, zval *nullable zv)
...
     dd_php_array_type arr_type = dd_php_determine_array_type(arr);
     if (arr_type == php_array_type_sequential) {
         mpack_start_array(w, num_elems);
         zval *val;
         ZEND_HASH_FOREACH_VAL((zend_array *)arr, val)
         {
-            _mpack_write_zval(w, val);
+            _mpack_write_zval(w, val, depth);
+            num_elems--;
+            if (num_elems == 0) {
```
Anilm3, line 111:
```suggestion
            if (--num_elems == 0) {
```

appsec/src/extension/msgpack_helpers.c
```diff
@@ -94,21 +123,32 @@ void dd_mpack_write_array(
...
+                dd_mpack_write_zstr(w, key_s);
             } else {
                 char buf[ZEND_LTOA_BUF_LEN];
                 ZEND_LTOA((zend_long)key_i, buf, sizeof(buf));
                 mpack_write(w, buf);
             }
-            _mpack_write_zval(w, val);
+            _mpack_write_zval(w, val, depth);
+            num_elems--;
+            if (num_elems == 0) {
```
Anilm3, line 134:
```suggestion
            if (--num_elems == 0) {
```

appsec/src/extension/msgpack_helpers.c
```diff
@@ -94,21 +123,32 @@ void dd_mpack_write_array(
...
             if (key_s) {
-                mpack_write_str(w, ZSTR_VAL(key_s), ZSTR_LEN(key_s));
+                dd_mpack_write_zstr(w, key_s);
             } else {
                 char buf[ZEND_LTOA_BUF_LEN];
                 ZEND_LTOA((zend_long)key_i, buf, sizeof(buf));
                 mpack_write(w, buf);
             }
-            _mpack_write_zval(w, val);
+            _mpack_write_zval(w, val, depth);
```
Anilm3, line 132: Shouldn't this be `depth + 1`?

appsec/src/extension/msgpack_helpers.c
```diff
@@ -63,25 +77,40 @@ void dd_mpack_write_zval(mpack_writer_t *nonnull w, zval *nullable zv)
...
+        num_elems = MAX_ARRAY_SIZE;
+    }
     dd_php_array_type arr_type = dd_php_determine_array_type(arr);
     if (arr_type == php_array_type_sequential) {
         mpack_start_array(w, num_elems);
         zval *val;
         ZEND_HASH_FOREACH_VAL((zend_array *)arr, val)
         {
-            _mpack_write_zval(w, val);
+            _mpack_write_zval(w, val, depth);
```
Anilm3, line 109: Shouldn't this be depth + 1?

appsec/src/extension/commands_helpers.c
```diff
@@ -749,9 +757,33 @@ bool dd_command_process_telemetry_metrics(mpack_node_t metrics)
             if (mpack_node_error(metrics) != mpack_ok) {
                 break;
             }
+            if (dd_msgpack_helpers_is_data_truncated() &&
+                strncmp(WAF_REQUEST_TAG, key_str, WAF_REQUEST_TAG_LEN) == 0) {
```
Anilm3, line 761: `strncmp` will compare _at most_ `WAF_REQUEST_TAG_LEN` characters, meaning that if `WAF_REQUEST_TAG` was a substring of `key_str`, it would consider them equally. 

You probably want something along the lines of: 
```suggestion
                WAF_REQUEST_TAG_LEN == key_len && memcmp(WAF_REQUEST_TAG, key_str, WAF_REQUEST_TAG_LEN) == 0) {
```

appsec/src/extension/commands_helpers.c
```diff
@@ -17,6 +17,13 @@
 #include <mpack.h>
 #include <stdatomic.h>
 
+static const char WAF_REQUEST_TAG[] = "waf.requests";
+static const size_t WAF_REQUEST_TAG_LEN = sizeof(WAF_REQUEST_TAG) - 1;
```
Anilm3, line 21: It's not a tag, it's a metric isn't it?

appsec/src/extension/commands_helpers.c
```diff
@@ -749,9 +757,33 @@ bool dd_command_process_telemetry_metrics(mpack_node_t metrics)
...
+                    if (separator > 0) {
+                        modified_tags_str[tags_len] = TAG_SEPARATOR;
+                    }
+                    memcpy(modified_tags_str + tags_len + separator,
+                        TRUNCATED_TAG, TRUNCATED_TAG_LEN);
+                    modified_tags_str[tags_len + TRUNCATED_TAG_LEN +
+                                      separator] = '\0';
+                    tags_len += TRUNCATED_TAG_LEN + separator;
+                }
+                tags_str = modified_tags_str;
```
Anilm3, line 779: If allocation fails, you still assign it? Why not keep the original instead?

Anilm3 (commented):
I'm not sure if there's a use-case, but maybe you want to be able to pass the limits as a structure rather than hardcoding them. This would allow you to restrict truncation to the actual data that is going to be provided to the WAF rather than the whole payload. In practice it probably doesn't matter as it is because our payloads are relatively small (I think), but maybe it's not true on every case... @cataphract can probably advise.

appsec/src/extension/msgpack_helpers.c
estringana: Same as before

appsec/src/extension/msgpack_helpers.c
```diff
@@ -14,9 +14,14 @@
 #include "php_helpers.h"
 #include "php_objects.h"
 
-static const int MAX_DEPTH = 32;
+static const size_t MAX_DEPTH_READING = 32;
+static const size_t MAX_DEPTH_WRITING = 20;
+static const size_t MAX_STR_LEN = 4096;
+static const size_t MAX_ARRAY_SIZE = 256;
```
cataphract, line 20: I know these limits are the "traditional" ones, but they don't make sense. Each array can have 256 elements, each element may be another array of 256 elements and so... 256^20 is a very large number. We should take this opportunity to fix this and limit the size of the payload, or at the very least the number of elements

appsec/src/extension/commands_helpers.c
```diff
@@ -749,9 +757,35 @@ bool dd_command_process_telemetry_metrics(mpack_node_t metrics)
...
+                    emalloc(tags_len + TRUNCATED_TAG_LEN + 1 + separator);
+                if (modified_tags_str) {
+                    memcpy(modified_tags_str, tags_str, tags_len);
+                    if (separator > 0) {
+                        modified_tags_str[tags_len] = TAG_SEPARATOR;
+                    }
+                    memcpy(modified_tags_str + tags_len + separator,
+                        TRUNCATED_TAG, TRUNCATED_TAG_LEN);
+                    modified_tags_str[tags_len + TRUNCATED_TAG_LEN +
+                                      separator] = '\0';
```
cataphract: NUL termination is not necessary

appsec/src/extension/msgpack_helpers.c
```diff
@@ -14,9 +14,14 @@
...
+static THREAD_LOCAL_ON_ZTS bool data_truncated_ = false;
```
cataphract, line 21: I'm not sure if this approach is the correct. This will mark all waf.requests in a request as truncated, whereas it probably only happened for one of them. I think the truncation flag should be instead sent to the helper, so that for that specific waf run it could generate a metric with the tag (or not).

Anilm3: I think the `appsec.waf.requests` is not sent per WAF call but per http request. On the other hand, there's also the `appsec.waf.input.truncated` tag which is expected per run with truncation.

appsec/src/extension/msgpack_helpers.c
Anilm3: You're right but this PR is moving from 0 limits to actually having limits so it's a net improvement in that respect. On the other hand, 256^20 would make PHP sweat long before even reaching this code. Generally I don't disagree with a better limit but discussions are moving towards a byte limit rather than a node limit (which is non-trivial to compute, but not impossible, but ultimately our limits are poor UX if they're ultimately exposed), so I'd suggest not implementing something that will ultimately deviate from every other implementation.

appsec/src/extension/msgpack_helpers.c
estringana: The metric @Anilm3 you're talking about is `appsec.waf.input_truncated` and that one is not SLO. This PRs truncated the data just for the purpose of sending the SLO tag missing. The implemenetation of that metric you mentioned will probably get done in the future as part of another PR

cataphract: @Anilm3 You're right, `appsec.waf.requests` is sent once per request. I'm OK with this implementation then.

PR #3263, opened 2025-05-20
Implement ATO v2 functions

estringana:
### Description

ATO v2 implementation following the approved RFC.

Extends the existing ATO integrations (Laravel, Symfony, WordPress) with the updated event schema: structured user fields, SDK-compatible event naming, and the additional signals required by the v2 spec. The framework-specific instrumentation changes are minimal, most of the work is in aligning the output with the new contract.

[APPSEC-57667]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-57667]: https://datadoghq.atlassian.net/browse/APPSEC-57667?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -16,6 +16,7 @@
 #include "user_tracking.h"
 #include <ext/standard/base64.h>
 #include <mpack.h>
+#include <stdatomic.h>
 
 static const char WAF_REQUEST_METRIC[] = "waf.requests";
 static const size_t WAF_REQUEST_METRIC_LEN = sizeof(WAF_REQUEST_METRIC) - 1;
@@ -790,6 +791,21 @@ bool dd_command_process_telemetry_metrics(mpack_node_t metrics)
     return true;
 }
 
+static void _init_zstr(
+    zend_string *_Atomic *nonnull zstr, const char *nonnull str, size_t len)
+{
+    zend_string *zstr_cur = atomic_load_explicit(zstr, memory_order_acquire);
+    if (zstr_cur != NULL) {
+        return;
+    }
+    zend_string *zstr_new = zend_string_init(str, len, 1);
+    if (atomic_compare_exchange_strong_explicit(zstr, &(zend_string *){NULL},
+            zstr_new, memory_order_release, memory_order_relaxed)) {
+        return;
+    }
+    zend_string_release(zstr_new);
+}
+
 // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
 void _handle_telemetry_metric(const char *nonnull key_str, size_t key_len,
     double value, const char *nonnull tags_str, size_t tags_len)
@@ -797,8 +813,14 @@ void _handle_telemetry_metric(const char *nonnull key_str, size_t key_len,
 #define HANDLE_METRIC(name, type)                                              \
     do {                                                                       \
         if (key_len == LSTRLEN(name) && memcmp(key_str, name, key_len) == 0) { \
-            dd_telemetry_add_metric(                                           \
-                name, LSTRLEN(name), value, tags_str, tags_len, type);         \
+            static zend_string *_Atomic key_zstr;                              \
+            _init_zstr(&key_zstr, name, LSTRLEN(name));                        \
+            zend_string *tags_zstr = zend_string_init(tags_str, tags_len, 1);  \
+            dd_telemetry_add_metric(key_zstr, value, tags_zstr, type);         \
+            zend_string_release(tags_zstr);                                    \
+            mlog_g(dd_log_debug,                                               \
+                "Telemetry metric %.*s added with tags %.*s and value %f",     \
+                (int)key_len, key_str, (int)tags_len, tags_str, value);        \
             return;                                                            \
         }                                                                      \
     } while (0)
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -12,8 +12,8 @@
 #include "msgpack_helpers.h"
 #include "request_abort.h"
 #include "tags.h"
-#include "user_tracking.h"
 #include "telemetry.h"
+#include "user_tracking.h"
 #include <ext/standard/base64.h>
 #include <mpack.h>
 
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -13,6 +13,7 @@
 #include "request_abort.h"
 #include "tags.h"
 #include "user_tracking.h"
+#include "telemetry.h"
 #include <ext/standard/base64.h>
 #include <mpack.h>
 #include <stdatomic.h>
@@ -790,21 +791,6 @@ bool dd_command_process_telemetry_metrics(mpack_node_t metrics)
     return true;
 }
 
-static void _init_zstr(
-    zend_string *_Atomic *nonnull zstr, const char *nonnull str, size_t len)
-{
-    zend_string *zstr_cur = atomic_load_explicit(zstr, memory_order_acquire);
-    if (zstr_cur != NULL) {
-        return;
-    }
-    zend_string *zstr_new = zend_string_init(str, len, 1);
-    if (atomic_compare_exchange_strong_explicit(zstr, &(zend_string *){NULL},
-            zstr_new, memory_order_release, memory_order_relaxed)) {
-        return;
-    }
-    zend_string_release(zstr_new);
-}
-
 // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
 void _handle_telemetry_metric(const char *nonnull key_str, size_t key_len,
     double value, const char *nonnull tags_str, size_t tags_len)
@@ -812,16 +798,8 @@ void _handle_telemetry_metric(const char *nonnull key_str, size_t key_len,
 #define HANDLE_METRIC(name, type)                                              \
     do {                                                                       \
         if (key_len == LSTRLEN(name) && memcmp(key_str, name, key_len) == 0) { \
-            static zend_string *_Atomic key_zstr;                              \
-            _init_zstr(&key_zstr, name, LSTRLEN(name));                        \
-            zend_string *tags_zstr = zend_string_init(tags_str, tags_len, 1);  \
-            ddtrace_metric_register_buffer(                                    \
-                key_zstr, type, DDTRACE_METRIC_NAMESPACE_APPSEC);              \
-            ddtrace_metric_add_point(key_zstr, value, tags_zstr);              \
-            zend_string_release(tags_zstr);                                    \
-            mlog_g(dd_log_debug,                                               \
-                "Telemetry metric %.*s added with tags %.*s and value %f",     \
-                (int)key_len, key_str, (int)tags_len, tags_str, value);        \
+            dd_add_telemetry_metric(                                           \
+                name, LSTRLEN(name), value, tags_str, tags_len, type);         \
             return;                                                            \
         }                                                                      \
     } while (0)
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -797,7 +797,7 @@ void _handle_telemetry_metric(const char *nonnull key_str, size_t key_len,
 #define HANDLE_METRIC(name, type)                                              \
     do {                                                                       \
         if (key_len == LSTRLEN(name) && memcmp(key_str, name, key_len) == 0) { \
-            dd_add_telemetry_metric(                                           \
+            dd_telemetry_add_metric(                                           \
                 name, LSTRLEN(name), value, tags_str, tags_len, type);         \
             return;                                                            \
         }                                                                      \
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -16,7 +16,6 @@
 #include "telemetry.h"
 #include <ext/standard/base64.h>
 #include <mpack.h>
-#include <stdatomic.h>
 
 static const char WAF_REQUEST_METRIC[] = "waf.requests";
 static const size_t WAF_REQUEST_METRIC_LEN = sizeof(WAF_REQUEST_METRIC) - 1;
--- a/appsec/src/extension/tags.c
+++ b/appsec/src/extension/tags.c
@@ -1065,6 +1065,8 @@ static PHP_FUNCTION(datadog_appsec_track_user_signup_event_automated)
     dd_trace_emit_asm_event();
 }
 
+void set_user_event_triggered(void) { _user_event_triggered = true; }
+
 static PHP_FUNCTION(datadog_appsec_track_user_signup_event)
 {
     UNUSED(return_value);
--- a/appsec/src/extension/tags.c
+++ b/appsec/src/extension/tags.c
@@ -1065,7 +1065,7 @@ static PHP_FUNCTION(datadog_appsec_track_user_signup_event_automated)
     dd_trace_emit_asm_event();
 }
 
-void set_user_event_triggered(void) { _user_event_triggered = true; }
+void dd_tags_set_user_event_triggered(void) { _user_event_triggered = true; }
 
 static PHP_FUNCTION(datadog_appsec_track_user_signup_event)
 {
--- a/appsec/src/extension/tags.c
+++ b/appsec/src/extension/tags.c
@@ -1181,7 +1181,8 @@ static PHP_FUNCTION(datadog_appsec_track_user_login_success_event_automated)
     }
 
     if (ZSTR_LEN(user_id) > 0) {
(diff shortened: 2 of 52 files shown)


ext/ddtrace_arginfo.h
```diff
@@ -441,6 +455,8 @@ static const zend_function_entry ext_functions[] = {
 	ZEND_RAW_FENTRY(ZEND_NS_NAME("DDTrace\\Testing", "emit_asm_event"), zif_DDTrace_Testing_emit_asm_event, arginfo_DDTrace_Testing_emit_asm_event, 0, NULL, NULL)
 	ZEND_RAW_FENTRY(ZEND_NS_NAME("DDTrace\\Internal", "add_span_flag"), zif_DDTrace_Internal_add_span_flag, arginfo_DDTrace_Internal_add_span_flag, 0, NULL, NULL)
 	ZEND_RAW_FENTRY(ZEND_NS_NAME("DDTrace\\Internal", "handle_fork"), zif_DDTrace_Internal_handle_fork, arginfo_DDTrace_Internal_handle_fork, 0, NULL, NULL)
+	ZEND_RAW_FENTRY(ZEND_NS_NAME("DDTrace\\ATO\\V2", "track_user_login_success"), zif_DDTrace_ATO_V2_track_user_login_success, arginfo_DDTrace_ATO_V2_track_user_login_success, 0, NULL, NULL)
```
Anilm3: This is a poor choice of namespace, this must be `\datadog\appsec\v2` instead.

appsec/src/extension/tags.h
```diff
@@ -19,6 +19,7 @@ void dd_tags_rinit(void);
 void dd_tags_rshutdown(void);
 void dd_tags_add_tags(zend_object *nonnull span, zend_array *nullable superglob_equiv);
 void dd_tags_add_blocked(void);
+void set_user_event_triggered(void);
```
cataphract: namespace the function with dd_tags_

appsec/src/extension/telemetry.h
```diff
@@ -0,0 +1,14 @@
...
+// dual-licensed under the Apache-2.0 License or BSD-3-Clause License.
+//
+// This product includes software developed at Datadog
+// (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+#pragma once
+
+#include "ddtrace.h"
+#include <stddef.h>
+
+void dd_add_telemetry_metric(const char *nonnull name, size_t name_len,
```
cataphract: convention would be `dd_telemetry_add_metric`

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,86 @@ static PHP_FUNCTION(set_user_wrapper)
     _ddtrace_set_user(INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+void dd_fire_user_event(void)
```
cataphract: this is a local function. Local convention is it should be static, prefixed with underscore and not starting with dd_, whioch is the convention for the global ones.

appsec/src/extension/commands_helpers.c
```diff
@@ -758,38 +758,15 @@ bool dd_command_process_telemetry_metrics(mpack_node_t metrics)
...
-            _init_zstr(&key_zstr, name, LSTRLEN(name));                        \
-            zend_string *tags_zstr = zend_string_init(tags_str, tags_len, 1);  \
-            ddtrace_metric_register_buffer(                                    \
-                key_zstr, type, DDTRACE_METRIC_NAMESPACE_APPSEC);              \
-            ddtrace_metric_add_point(key_zstr, value, tags_zstr);              \
-            zend_string_release(tags_zstr);                                    \
-            mlog_g(dd_log_debug,                                               \
-                "Telemetry metric %.*s added with tags %.*s and value %f",     \
-                (int)key_len, key_str, (int)tags_len, tags_str, value);        \
+            dd_add_telemetry_metric(                                           \
```
cataphract: All the names of the metrics are still known at compile time. You might as well keep the optimization that allocates the name of the metric only once and have instead `dd_add_telemetry_metric` take a (possibly persistent) `zend_string`

ext/ddtrace_arginfo.h
estringana: I disagree with your comment. The functionality added here is adding value even when appsec is not enabled or even when a customer not aware of ASM

Anilm3: - This is a feature that only works with appsec enabled, the tags that you add when appsec is disabled are ultimately not being taken advantage of other than for appsec purposes (at least for now).
- The namespace ATO is a bad choice as that's not the only purpose of these functions.
- It'll be more obvious for customers who are using `\datadog\appsec\...` that these are newer versions, I think.

If you still want to keep them in the `DDTrace` namespace, figure out a better sub-namespace than ATO.

estringana: The RFC is called `ATO SDK Consolidation` so I disagree with ATO not being a good name. Regarding not giving value to customer, I also disagree, when appsec is disabled, the tracer(as per the RFC says) is adding some valuable tags for customers.

Anilm3: It doesn't matter what the RFC is called, the functionality that this offers is business logic tracking with ATO being one of the many uses of this information.

estringana: Well, I still think ATO is a good one, can you propose one that suits better according to your point of view?

Anilm3: `\datadog\appsec\v2\` or just `\DDTrace\`

estringana: Changed to `\datadog\appsec\v2\`

appsec/src/extension/commands_helpers.c
estringana: Doing this! Thanks for the valuable comment

estringana: Done

appsec/src/extension/telemetry.c
```diff
@@ -0,0 +1,50 @@
...
+void dd_telemetry_add_sdk_event(char *nonnull event_type, size_t event_type_len)
+{
+    static zend_string *_Atomic key_zstr;
+    _init_zstr(&key_zstr, LSTRARG("sdk.event"));
+    char *tags = NULL;
+    size_t tags_len = asprintf(&tags, "event_type:%.*s,sdk_version:v2",
+        (int)event_type_len, event_type);
+    zend_string *tags_zstr = zend_string_init(tags, tags_len, 1);
+    dd_telemetry_add_metric(key_zstr, 1, tags_zstr, DDTRACE_METRIC_TYPE_COUNT);
+    mlog_g(dd_log_debug,
```
Anilm3: This log message is a duplicate of the one already available in `dd_telemetry_add_metric`.

appsec/src/extension/telemetry.c
```diff
@@ -0,0 +1,50 @@
...
+        "Telemetry metric %.*s added with tags %.*s and value %f",
+        (int)ZSTR_LEN(name_zstr), ZSTR_VAL(name_zstr), (int)ZSTR_LEN(tags_zstr),
+        ZSTR_VAL(tags_zstr), value);
+}
+
+void dd_telemetry_add_sdk_event(char *nonnull event_type, size_t event_type_len)
+{
+    static zend_string *_Atomic key_zstr;
+    _init_zstr(&key_zstr, LSTRARG("sdk.event"));
```
Anilm3: Why not use an interned string instead as we usually do (requires `*_startup()`)?

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,88 @@ static PHP_FUNCTION(set_user_wrapper)
     _ddtrace_set_user(INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+static void _fire_user_event(void)
```
Anilm3:
```suggestion
static void _emit_user_event(void)
```

To keep the naming consistent.

appsec/src/extension/telemetry.c
estringana: Ok

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,88 @@ static PHP_FUNCTION(set_user_wrapper)
...
+    } else if (user != NULL && Z_TYPE_P(user) == IS_ARRAY) {
+        zval *user_id_zv = zend_hash_str_find(Z_ARR_P(user), ZEND_STRL("id"));
+        if (user_id_zv == NULL) {
+            mlog(dd_log_warning,
+                "Id not found in user object in "
+                "datadog\\appsec\\v2\\track_user_login_success");
+            return;
+        }
+        if (Z_TYPE_P(user_id_zv) != IS_STRING) {
+            mlog(dd_log_warning,
```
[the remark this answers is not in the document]
Anilm3: Don't do this, of course...

appsec/src/extension/telemetry.c
estringana: Same approach used as on `_handle_telemetry_metric`

Anilm3: I understand it's used in a single other place, but that doesn't answer the question.

appsec/src/extension/tags.c
```diff
@@ -1248,7 +1251,8 @@ static PHP_FUNCTION(datadog_appsec_track_user_login_success_event)
     _user_event_triggered = true;
     zend_array *meta_ht = Z_ARRVAL_P(meta);
 
-    dd_find_and_apply_verdict_for_user(user_id, ZSTR_EMPTY_ALLOC());
+    dd_find_and_apply_verdict_for_user(
+        user_id, ZSTR_EMPTY_ALLOC(), user_event_login_success);
```
Anilm3: nit: for this and all cases below, I think you can just use `zend_empty_string`

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,88 @@ static PHP_FUNCTION(set_user_wrapper)
     _ddtrace_set_user(INTERNAL_FUNCTION_PARAM_PASSTHRU);
 }
 
+static void _fire_user_event(void)
+{
+    dd_tags_set_user_event_triggered();
+    dd_trace_emit_asm_event();
+    dd_tags_set_sampling_priority();
```
Anilm3, line 74: Just a reminder that on your other PR, these two are merged together (It confused me...)

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,88 @@ static PHP_FUNCTION(set_user_wrapper)
...
+        return;
+    }
+
+    if (_ddtrace_v2_track_user_login_failure == NULL) {
+        mlog(dd_log_debug, "Invalid DDTrace\\track_user_login_failure, "
+                           "this shouldn't happen");
+        return;
+    }
+
+    _emit_user_event();
```
Anilm3: This should happen right after calling `_ddtrace_v2_track_user_login_failure`, because you have exit conditions before this point but the sampling priority must be set for SDK events.

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,88 @@ static PHP_FUNCTION(set_user_wrapper)
...
+        if (Z_TYPE_P(user_id_zv) != IS_STRING) {
+            mlog(dd_log_warning,
+                "Unexpected id type in "
+                "datadog\\appsec\\v2\\track_user_login_success");
+            return;
+        }
+        user_id = Z_STR_P(user_id_zv);
+    }
+
+    _emit_user_event();
```
Anilm3: This should happen right after calling `_ddtrace_v2_track_user_login_success`, because you have exit conditions before this point but the sampling priority must be set for SDK events.

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,88 @@ static PHP_FUNCTION(set_user_wrapper)
...
+            mlog(dd_log_warning,
+                "Id not found in user object in "
+                "datadog\\appsec\\v2\\track_user_login_success");
+            return;
+        }
+        if (Z_TYPE_P(user_id_zv) != IS_STRING) {
+            mlog(dd_log_warning,
+                "Unexpected id type in "
+                "datadog\\appsec\\v2\\track_user_login_success");
+            return;
```
Anilm3: I'm not sure you should exit here or in the condition above, as we can still call `libddwaf` only with the login if the user ID is not valid.

ext/ddtrace.c
```diff
@@ -2069,39 +2113,121 @@ PHP_FUNCTION(DDTrace_set_user) {
...
+    } else if (user != NULL && Z_TYPE_P(user) == IS_ARRAY) {
+        // This is required to avoid writting to metadata if no id
+        zval *user_id_zv = zend_hash_str_find(Z_ARR_P(user), ZEND_STRL("id"));
+        if (user_id_zv == NULL) {
+            LOG_LINE(WARN, "Id not found in user object in datadog\\appsec\\v2\\track_user_login_success");
+            RETURN_NULL();
+        }
+        if (Z_TYPE_P(user_id_zv) != IS_STRING) {
+            LOG_LINE(WARN, "Unexpected id type in datadog\\appsec\\v2\\track_user_login_success");
+            RETURN_NULL();
```
Anilm3: Should not fail just because the user is not available, same as above.

ext/ddtrace.c
```diff
@@ -2069,39 +2113,121 @@ PHP_FUNCTION(DDTrace_set_user) {
...
+            zend_hash_update(target_table, key, &value_copy);
+
+            zend_string_release(key);
+        }
+        ZEND_HASH_FOREACH_END();
+    }
+
+    if (user != NULL && Z_TYPE_P(user) != IS_NULL && user_id == NULL) {
+        LOG_LINE(WARN, "Unexpected id in datadog\\appsec\\v2\\track_user_login_success");
+        RETURN_NULL();
```
Anilm3: Idem

ext/ddtrace.c
```diff
@@ -2111,7 +2237,84 @@ PHP_FUNCTION(DDTrace_set_user) {
...
+    if (user_id != NULL) {
+        _ddtrace_set_user(user_id, metadata, false);
+    }
+}
+
+PHP_FUNCTION(datadog_appsec_v2_track_user_login_failure) {
+    UNUSED(execute_data);
+
+    zend_string *login = NULL;
+    zend_bool exists;
```
Anilm3:
```suggestion
    zend_bool exists = false;
```

Not sure if it'll be initialised otherwise.

appsec/src/extension/user_tracking.c
estringana: I will adapt the second one merged

appsec/src/extension/user_tracking.c
estringana: This is because of the following sentence from the RFC:
>The user object, when provided, must contain an id field. The user object must match the type expected by the setUser method, If setUser only accepts an ID, the method variant using a user object should be ignored.

ext/ddtrace.c
estringana: Idem as above

ext/ddtrace.c
estringana: Idem as above

appsec/src/extension/user_tracking.c
Anilm3: You still have to call the WAF.

ext/ddtrace.c
Anilm3: Please check the comment in the RFC.

appsec/src/extension/user_tracking.c
estringana: Done here and in all other instances

ext/ddtrace.c
estringana: Done here and in all other instances

ext/ddtrace.c
estringana: Done here and in all other instances

appsec/src/extension/telemetry.c
estringana: Changed.

appsec/src/extension/user_tracking.c
```diff
@@ -61,6 +67,78 @@ static PHP_FUNCTION(set_user_wrapper)
...
+
+    zend_string *login;
+    zval *user = NULL;
+    zend_array *metadata = NULL;
+    zend_string *user_id = NULL;
+    if (zend_parse_parameters(
+            ZEND_NUM_ARGS(), "S|zh", &login, &user, &metadata) == FAILURE) {
+        return;
+    }
+    if (_ddtrace_v2_track_user_login_success == NULL) {
```
Anilm3: I'm a bit confused, why do you check this here, when you already called the function before? Same applies to the login failure function

Anilm3 (approved):
Just two things to fix, otherwise it seems okay.

PR #3269, opened 2025-05-27
Schema extraction with DD_APM_TRACING_ENABLED=false

estringana:
### Description

Generate schemas when the request has been picked by the sample even when standalone mode is on

[[APPSEC-57339]]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-57339]: https://datadoghq.atlassian.net/browse/APPSEC-57339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appsec/src/extension/commands_helpers.c
```diff
@@ -641,6 +648,11 @@ void dd_command_process_meta(mpack_node_t root, zend_object *nonnull span)
             return;
         }
     }
+
+    if (has_schemas && !get_DD_APM_TRACING_ENABLED()) {
+        dd_trace_emit_asm_event();
+        dd_tags_set_sampling_priority();
```
Anilm3: These two functions seem to always appear together, it might make sense to just merge them as the concept of emitting an ASM event is quite specific to your implementation of standalone ASM.

appsec/src/extension/commands_helpers.c
```diff
@@ -632,6 +633,12 @@ void dd_command_process_meta(mpack_node_t root, zend_object *nonnull span)
             key_len = INT_MAX;
         }
 
+        if (!has_schemas && dd_string_starts_with_lc(
+                                key_str, key_len, ZEND_STRL("_dd.appsec.s"))) {
```
Anilm3:
```suggestion
                                key_str, key_len, ZEND_STRL("_dd.appsec.s."))) {
```

appsec/src/extension/commands_helpers.c
cataphract: I second this. Emitting an asm event has several consequences: besides setting the priority, the _dd.p.ts should be set (and propagated) and propagation should be allowed to continue. It seems strange to single out setting the sample priority here.

appsec/src/extension/request_lifecycle.c
```diff
@@ -882,7 +882,7 @@ static uint64_t _calc_sampling_key(zend_object *root_span, int status_code)
     // check sampling priority: must be >=1. o/wise return 0
     zend_long sampling_priority =
         dd_trace_get_priority_sampling_on_span_zobj(root_span);
-    if (sampling_priority < 1) {
+    if (get_DD_APM_TRACING_ENABLED() && sampling_priority < 1) {
```
cataphract, line 885: This doesn't appear to make sense at first sight. IIRC you're saying that sampling should run in situations where it wouldn't if APM tracing were enabled (!).

I suspect this is here to workaround a problem with the standalone implementation that's setting the sampling priority to drop to early, but if that's the case then the solution should (in principle) be to set it later in the request, rather than setting it to drop and then later to user_keep if the schema extraction sampler says we should run.

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
```diff
@@ -146,7 +146,7 @@ trait CommonTests {
         }
 
         Span span = trace.first()
-        assert span.metrics._sampling_priority_v1 == 1.0d
+        assert span.metrics._sampling_priority_v1 < 2.0d
```
cataphract, line 149: this doesn't look related to this PR, as this test doesn't deal with standalone or schema extraction. Can you explain the motivation for it?

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/SamplingTestsInFpm.groovy
```diff
@@ -1,12 +1,17 @@
...
 import com.datadog.appsec.php.docker.AppSecContainer
+import com.datadog.appsec.php.model.Trace
 import org.junit.jupiter.api.Test
 
 import java.net.http.HttpResponse
 
 trait SamplingTestsInFpm {
 
+    void assertSamplingPriority(Trace trace, boolean schemaExtracted) {
+        //Nothing to do here. The work will be done on overrides
```
cataphract, line 12: I'd rename it something like `maybeAssertSamplingPriority` since this looks like an actual implementation and lose the second argument. Instead of `assertSamplingPriority(trace, false);`  you can put nothing and instead of `assertSamplingPriority(trace, true);` you could just do `maybeAssertSamplingPriority()`

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
estringana: The auth event should not set sampling priority to 1

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/SamplingTestsInFpm.groovy
estringana: I have done changes

appsec/src/extension/commands_helpers.c
estringana: Ok

appsec/src/extension/request_lifecycle.c
estringana: Changed

ext/asm_event.c
```diff
@@ -18,9 +18,8 @@ DDTRACE_PUBLIC void ddtrace_emit_asm_event() {
     } else {
         DDTRACE_G(asm_event_emitted) = true;
     }
-    if (!get_DD_APM_TRACING_ENABLED()) {
-        ddtrace_set_priority_sampling_on_root(PRIORITY_SAMPLING_USER_KEEP, DD_MECHANISM_ASM);
-    }
+
+    ddtrace_set_priority_sampling_on_root(PRIORITY_SAMPLING_USER_KEEP, get_DD_APM_TRACING_ENABLED() ? DD_MECHANISM_MANUAL : DD_MECHANISM_ASM);
```
Anilm3: Are you sure it makes sense to use `DD_MECHANISM_MANUAL` considering it sets `_dd.p.dm` to `-4` (which isn't accurate)?

estringana: Which one would you set then? There is an asm event so we should force to be kept

Anilm3: It's not clear to me if we should set `DD_MECHANISM_MANUAL(-4)` or `DD_MECHANISM_ASM(-5)` when APM is enabled... Maybe @daniel-romano-DD can clarify.

dromanol: The sampling mechanism should reflect the product that made the decision to keep the trace, so, as it is a schema extraction, it should be -5 (ASM)

Anilm3: After an offline conversation, the conclusion is that we should just change this to be `DD_MECHANISM_ASM` when we set manual keep.

estringana: So this line will change to the following if I understood correct
```
ddtrace_set_priority_sampling_on_root(PRIORITY_SAMPLING_USER_KEEP,  DD_MECHANISM_ASM);
```

Anilm3: Yes. that's it.

estringana: Done

PR #3277, opened 2025-05-28
feat(profiling) re-enable allocation profiling with JIT for PHP 8.4.7

realFlowControl:
### Description

With #3199 we've disable allocation profiling when an active JIT was detected on PHP 8.4. The [upstream fix](https://github.com/php/php-src/pull/18289) landed in PHP 8.4.7 and was confirmed to fix the issue, so we can allow allocation profiling even when an active JIT is detected on PHP >= 8.4.7  

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/allocation_ge84.rs
+++ b/profiling/src/allocation/allocation_ge84.rs
@@ -143,8 +143,8 @@ lazy_static! {
 }
 
 pub fn first_rinit_should_disable_due_to_jit() -> bool {
-    if *JIT_ENABLED && zend::PHP_VERSION_ID >= 80400 {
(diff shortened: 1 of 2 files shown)


profiling/src/allocation/allocation_ge84.rs
```diff
@@ -143,8 +143,8 @@ lazy_static! {
 }
 
 pub fn first_rinit_should_disable_due_to_jit() -> bool {
-    if *JIT_ENABLED && zend::PHP_VERSION_ID >= 80400 {
-        error!("Memory allocation profiling will be disabled as long as JIT is active. To enable allocation profiling disable JIT. See https://github.com/DataDog/dd-trace-php/pull/3199");
+    if *JIT_ENABLED && zend::PHP_VERSION_ID >= 80400 && zend::PHP_VERSION_ID < 80407 {
```
bwoebi, line 146: Don't use static version id checks on patch versions.

PR #3295, opened 2025-06-12
fix(prof): fewer borrows, less panics on borrows

realFlowControl:
### Description

We found a `panic_already_borrowed` in our RelEnv with I/O profiling. The following is happening:
- the profiler takes a sample
- during this, a signal arrives (PHP timed out)
- the timeout handler want’s to write it’s usual “Fatal error: Maximum execution time ..” error
- this calls into the I/O profiling write hook which decides to take a sample :boom:

So far we've assumed that we do not sample while we sample, turns out the signal handler for timeout can bypass our assumptions and we may actually sample while we sample 😬

There is no reason to hold onto the mutable borrow for those `RefCell`'s while we collect the stack trace, so this PR aims to only hold onto the mutable borrow for those `RefCell`'s while we change the stats data and make the sampling decision. After that we release the borrow before calling into stack trace collection.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/mod.rs
+++ b/profiling/src/allocation/mod.rs
@@ -56,25 +56,29 @@ impl AllocationProfilingStats {
         self.next_sample = self.poisson.sample(&mut self.rng) as i64;
     }
 
-    fn track_allocation(&mut self, len: size_t) {
+    fn should_collect_allocation(&mut self, len: size_t) -> bool {
         self.next_sample -= len as i64;
 
         if self.next_sample > 0 {
-            return;
+            return false;
         }
 
         self.next_sampling_interval();
 
-        if let Some(profiler) = Profiler::get() {
-            // Safety: execute_data was provided by the engine, and the profiler doesn't mutate it.
-            unsafe {
-                profiler.collect_allocations(
-                    zend::ddog_php_prof_get_current_execute_data(),
-                    1_i64,
-                    len as i64,
-                )
-            };
-        }
+        true
+    }
+}
+
+pub fn collect_allocation(len: size_t) {
+    if let Some(profiler) = Profiler::get() {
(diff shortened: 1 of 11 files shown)


profiling/src/allocation/mod.rs
```diff
@@ -56,25 +56,30 @@ impl AllocationProfilingStats {
...
-                    len as i64,
-                )
-            };
-        }
+        true
+    }
+}
+
+#[cold]
+pub fn collect_allocation(len: size_t) {
```
morrisonlevi, line 73: I was wondering why this got split out, took me a moment to see `#[cold]`! Definitely makes sense.

profiling/rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.78"
+channel = "1.84"
```
morrisonlevi: Bob is rebuilding images for 1.84 to match the libdatadog MSRV bump.

morrisonlevi (commented):
I pushed up a few protections. I added a new helper trait `RefCellExt` with `try_with_borrow` and `try_with_borrow_mut` helpers for `LocalKey<RefCell<T>>` so that we can coalesce multiple kinds of borrow errors at once. Then I used it across many local variables. Please review my commits!

This address the comment about `CACHED_STRINGS`. IMO there's no point in half-handling all the local variables needed to walk to _actually_ walk the stack. I think in all cases now, we don't collect the stack and emit an error. We could theoretically change this to instead have no stack and emit the sample without it.

realFlowControl (commented):
As the author I can't "formally approve", but I've reviewed your commits and LGTM!

PR #3300, opened 2025-06-17
ci: add PR performance quality gate

PROFeNoM:
### Description

[APMSP-2049](https://datadoghq.atlassian.net/browse/APMSP-2049)

Adds PR-level performance quality gate.
Now, if a PR adds more than 10% (defined [here](https://github.com/DataDog/benchmarking-platform/blob/dd-trace-php/bp-runner.fail-on-regression.yml)) overhead to any microbenchmark, a job `check-big-regressions` will fail and will prevent the PR to be merged

Gate will be skippable by setting the label `performance/ignore-performance-regression` - (Corresponding BP Runner PR: https://github.com/DataDog/benchmarking-platform-tools/pull/78)

[Sample success](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-py/-/jobs/963235691#L175)
image

[Sample failure](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-php/-/jobs/990099495)image

### Reviewer checklist
- [ ] Test coverage seems ok.

[APMSP-2049]: https://datadoghq.atlassian.net/browse/APMSP-2049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/Benchmarks/API/TraceFlushBench.php
```diff
@@ -24,7 +24,11 @@ public function setUp()
     {
         \dd_trace_serialize_closed_spans();
         Utils::putEnvAndReloadConfig([
-            'DD_TRACE_GENERATE_ROOT_SPAN=0'
+            'DD_TRACE_GENERATE_ROOT_SPAN=0',
+            'DD_TRACE_AUTO_FLUSH_ENABLED=0',
+            'DD_TRACE_SIDECAR_TRACE_SENDER=0',
+            'DD_REMOTE_CONFIG_ENABLED=0',
+            'DD_INSTRUMENTATION_TELEMETRY_ENABLED=0',
```
bwoebi: This doesn't work here, these are system INIs and won't have an effect _within_ a benchmark suite. You'd have to start the benchmarking suite with these envs in the first place.

PROFeNoM: That's actually what I was doing on my terminal 🤦 damn, I'm tired

bwoebi: Though disabling the sidecar is ... probably not a good idea, as we actually want to optimize flushing _for_ the sidecar scenario...

PROFeNoM: Yh, in my last commit, it seems that what actually had an effect is the auto_flushing.

I did initially disabled the sidecar because otherwise I would get some connection errors locally (as I do on other tests). Seems to have been resolved :shrug:

PROFeNoM: I will most certainly put all of these stability-related changes in another PR

bwoebi: Yeah, without the autoflushing disabled \DDTrace\flush() is essentially a no-op, probably the reason why that test was taking such a short time.

ddyurchenko (approved):
Looks good! :heavy_check_mark:

PR #3301, opened 2025-06-17
Implement new SSI configuration telemetry

iamluc:
### Description

Implement https://docs.google.com/document/d/1wydXj2fW0V0jaWU6W9pIbipT_AXD15tmbTtOPXFBEy4/edit?tab=t.nun42691l9k2 for PHP

https://datadoghq.atlassian.net/browse/INPLAT-577

System-tests PR: https://github.com/DataDog/system-tests/pull/4797

### Reviewer checklist
- [ ] Test coverage seems ok.


loader/dd_library_loader.c
```diff
@@ -228,9 +228,25 @@ static void ddtrace_pre_minit_hook(injected_ext *config) {
...
+    if (ddtrace_loaded_by_ssi) {
+        *ddtrace_loaded_by_ssi = true;
+    }
+    char **ddtrace_ssi_injection_enabled = (char **)DL_FETCH_SYMBOL(module->handle, "ddtrace_ssi_injection_enabled");
+    if (ddtrace_ssi_injection_enabled) {
+        *ddtrace_ssi_injection_enabled = getenv("DD_INJECTION_ENABLED");
+    }
+    char **ddtrace_ssi_forced_injection_enabled = (char **)DL_FETCH_SYMBOL(module->handle, "ddtrace_ssi_forced_injection_enabled");
+    if (ddtrace_ssi_forced_injection_enabled) {
+        *ddtrace_ssi_forced_injection_enabled = getenv("DD_INJECTION_FORCE");
```
bwoebi:
```suggestion
        *ddtrace_ssi_forced_injection_enabled = force_load ? "True" : "False";
```
At least according to the document it should be only one of those values, not the raw value?

loader/dd_library_loader.c
```diff
@@ -228,9 +228,25 @@ static void ddtrace_pre_minit_hook(injected_ext *config) {
...
     }
+
+    // Let ddtrace knows that it was loaded by the loader
+    bool *ddtrace_loaded_by_ssi = (bool *)DL_FETCH_SYMBOL(module->handle, "ddtrace_loaded_by_ssi");
+    if (ddtrace_loaded_by_ssi) {
+        *ddtrace_loaded_by_ssi = true;
+    }
+    char **ddtrace_ssi_injection_enabled = (char **)DL_FETCH_SYMBOL(module->handle, "ddtrace_ssi_injection_enabled");
+    if (ddtrace_ssi_injection_enabled) {
+        *ddtrace_ssi_injection_enabled = getenv("DD_INJECTION_ENABLED");
```
bwoebi: Stupid question, not related to this PR, but we don't make use of that variable at all currently, right?

loader/dd_library_loader.c
iamluc: I implemented it the same way as in Ruby: https://github.com/DataDog/dd-trace-rb/pull/4737/files#diff-106f688b0baf1218c65023cab312b14ccff2fbf78f07e09bb9c2dde05da93f95R145-R157

@mabdinur what do you think?

loader/dd_library_loader.c
iamluc: You're correct. We don't need it.

bwoebi: Then this configuration feels misleading, I think we only should report what we actually use. We don't use that config, so we should not report it.

loader/dd_library_loader.c
bwoebi: Well, at least it would not report it correctly when force_load is set via ini.

iamluc: Good catch!

loader/dd_library_loader.c
iamluc: cc @mabdinur WDYT? Again, it is similar to what's done in Ruby I think

loader/dd_library_loader.c
iamluc: Updated as you suggested

loader/dd_library_loader.c
mabdinur: This is the same approach we also used in Java and .NET. It was used to simplify the implementation and testing. These fields could also be used in debug logs or by other products (I believe it's used by the profiler in Python). 

I dont have a strong opinion here. As long as we capture the right telemetry names and values this change looks good to me.

loader/dd_library_loader.c
mabdinur: Yup it should ideally be a Boolean but this behavior is not consistent across libraries. The .NET tracer reports raw configuration values to the instrumentation telemetry platform. Python reports the processed values. Other libraries do a mix of the two. This issue should be resolved by the config chaining initiative. Here we should go with the simplest implementation.


The ruby implementation is incomplete. I'll update it later today.

bwoebi (approved):
All comments addressed, nice :-)

PR #3306, opened 2025-06-23
feat(prof): move allocation profiler stack collection to safe point

realFlowControl:
### Description

The allocation profiler currently collects a stack trace immediately when the sampling threshold (4MB) is reached. This can occur anywhere in the PHP VM, including outside safe points. While usually safe, there are rare cases where `opline` has been `free()`'d and not restored, leading to use-after-free issues during stack collection 💥

These are engine-level bugs, which we've addressed retroactively via upstream fixes and/or profiler-side workarounds. This PR takes a proactive approach by deferring stack trace collection until the next safe point, by storing allocation samples in TLS and raising the engine interrupt flag. At the next interrupt - where we already handle wall/CPU time sampling - we check for a pending allocation sample and collect the trace then.

Out of Scope / Follow-Up Work
1. Stack trace collection for I/O samples should also be deferred to safe points.
2. When multiple profiler events (e.g., wall time, CPU time, allocation) are pending, we should ideally collect only one stack trace to reduce overhead. This will require additional refactoring.
3. Investigate whether we can attach wall/CPU timing metrics to allocation samples when collected, potentially increasing resolution without introducing sampling bias.

Point 2 seems like a clear improvement, but due to its architectural implications, it's deferred for now. Importantly, this PR already avoids redundant stack walks in one case: when a single PHP function call performs multiple allocations and hits the sampling threshold multiple times, we previously performed multiple redundant stack walks. With this change, those are consolidated into a single stack trace collected at the next safe point.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/mod.rs
+++ b/profiling/src/allocation/mod.rs
@@ -65,22 +65,36 @@ impl AllocationProfilingStats {
 
         self.next_sampling_interval();
 
-        if let Some(profiler) = Profiler::get() {
-            // Safety: execute_data was provided by the engine, and the profiler doesn't mutate it.
-            unsafe {
-                profiler.collect_allocations(
-                    zend::ddog_php_prof_get_current_execute_data(),
-                    1_i64,
-                    len as i64,
-                )
-            };
-        }
+        // Queue the allocation for safe-point collection
+        REQUEST_LOCALS.with_borrow_mut(|locals| {
+            locals
+                .allocation_interrupt_count
+                .fetch_add(1, Ordering::SeqCst);
+        });
+
+        PENDING_ALLOCATION_SIZE.with_borrow_mut(|pending_size| {
+            // Trigger interrupt only when transitioning from zero to non-zero
+            if *pending_size == 0 {
+                *pending_size += len as u64;
+                if let Some(profiler) = Profiler::get() {
+                    // TODO: If this thread needs to take an allocation sample, calling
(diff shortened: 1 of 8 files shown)


morrisonlevi (commented):
This is fine for a proof of concept; it let's us see how the accuracy of the allocation profiler is impacted.

However, it will walk the stacks twice, which is wasteful. If you are trying to compare overhead, you'll want to fix that first, or disable wall time sampling when testing.

We should think carefully about this, but I _think_  it may make sense to gather time any time we walk the stack.

profiling/src/allocation/mod.rs
```diff
@@ -65,22 +65,36 @@ impl AllocationProfilingStats {
...
+            // Trigger interrupt only when transitioning from zero to non-zero
+            if *pending_size == 0 {
+                *pending_size += len as u64;
+                if let Some(profiler) = Profiler::get() {
+                    // TODO: If this thread needs to take an allocation sample, calling
+                    // `profiler.trigger_interrupt()` will not only trigger this threads interrupt,
+                    // but all other PHP ZTS threads interrupts as well. The interrupt handler is
+                    // pretty slim, and does not collect a stack trace if there is nothing pending,
+                    // yet we should only trigger an interrupt in the "current" thread.
+                    profiler.trigger_interrupt();
```
realFlowControl, line 85: Even though we could fix it here, I'd say we are also save to ignore this for the time being, as only very very very few people are using ZTS PHP.

PR #3312, opened 2025-06-24
Try fix appsec integration

cataphract:
### Reviewer checklist
- [ ] Test coverage seems ok.


cataphract:
> Have you intentionally _not_ put the images jobs `.gitlab/ci-images.yml` file?

@bwoebi Yes, I tried an approach that didn't involve 1) manually specifying your docker hub credentials and 2) having to submit PRs to another repository to update the hashes. At the end, while I _could_ get it to work, the lack of docker-in-docker:arm64 made the endeavor unviable. I would either have to use a runner that didn't have access to vault through emissary, or I'd have to use a runner where docker couldn't be run (and thus I would have to use other means to build the images like buildkit which makes the whole thing more difficult).

Another problem is the 60-minute time limit and the fact we'd need to split the jobs.

So at the end, while no obstacle was insurmountable, all together it become more practical to keep the status quo and build images ourselves. I left the jobs there so the effort can be resumed once there's docker-in-docker:arm64

PR #3316, opened 2025-06-25
feat: reduce telemetry sent

Leiyks:
Reduce telemetry volume sent to be a 30 min interval as precise application lifecycle tracking is not needed.
The goal is to improve performance as well as massively reduce the intake costs for PHP telemetry.

### Reviewer checklist
- [ ] Test coverage seems ok.

Related Jiras: [LANGPLAT-395]

[LANGPLAT-395]: https://datadoghq.atlassian.net/browse/LANGPLAT-395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components-rs/telemetry.rs
```diff
@@ -220,3 +226,50 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_add_integration_log_buffer(
...
+
+    let mut reader = OneWayShmReader::<NamedShmHandle, _>::new(mapped, shm_path.clone());
+    let (_, buf) = reader.read();
+
+    let Ok((seen_integrations, seen_composer_paths)) = bincode::deserialize::<(
+        std::collections::HashSet<String>,
+        std::collections::HashSet<std::path::PathBuf>,
+    )>(buf) else {
+        return std::ptr::null_mut();
+    };
```
bwoebi: Can we cache this (i.e. the reader and the deserialized data) please? I.e. store this in the DDTRACE_G(). E.g. store up to 10 readers in a HashMap or something. This is going to be accessed far more often than it will change.

ext/sidecar.c
```diff
@@ -402,7 +404,13 @@ void ddtrace_sidecar_submit_root_span_data_direct(ddtrace_root_span_data *root,
...
     }
     if (changed || !root) {
-        ddtrace_ffi_try("Failed sending remote config data", ddog_sidecar_set_remote_config_data(&ddtrace_sidecar, ddtrace_sidecar_instance_id, &DDTRACE_G(sidecar_queue_id), service_slice, env_slice, version_slice, &DDTRACE_G(active_global_tags)));
+        ddtrace_ffi_try("Failed sending remote config data", ddog_sidecar_set_universal_service_tags(&ddtrace_sidecar, ddtrace_sidecar_instance_id, &DDTRACE_G(sidecar_queue_id), service_slice, env_slice, version_slice, &DDTRACE_G(active_global_tags)));
+    }
+
+    ddog_SidecarActionsBuffer *filtered = ddog_sidecar_telemetry_buffer_filter_new(ddtrace_telemetry_buffer(), service_slice, env_slice, version_slice);
+    if (filtered) {
+        ddtrace_ffi_try("Failed flushing filtered telemetry buffer",
+            ddog_sidecar_telemetry_buffer_flush(&ddtrace_sidecar, ddtrace_sidecar_instance_id, &DDTRACE_G(sidecar_queue_id), filtered));
```
bwoebi: Can we combine that, and have the rust code flush directly?

ext/telemetry.c
```diff
@@ -202,38 +203,14 @@ void ddtrace_telemetry_finalize(void) {
     }
 
     dd_commit_metrics();
+
     ddtrace_ffi_try("Failed flushing telemetry buffer",
                     ddog_sidecar_telemetry_buffer_flush(&ddtrace_sidecar, ddtrace_sidecar_instance_id, &DDTRACE_G(sidecar_queue_id), buffer));
```
bwoebi: The flushing also needs to filter with shm.

components-rs/telemetry.rs
```diff
@@ -227,54 +231,97 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_add_integration_log_buffer(
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn ddog_sidecar_telemetry_cache_drop(_: Box<ShmCacheMap>) {}
+
+unsafe fn ddog_sidecar_telemetry_cache_get_or_update(
+    cache: &mut ShmCacheMap,
+    service: CharSlice,
+    env: CharSlice,
+) -> *const ShmCache {
```
bwoebi:
```suggestion
) -> Option<&ShmCache> {
```
is there a particular reason to return that as pointer?

components-rs/telemetry.rs
```diff
@@ -227,54 +231,97 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_add_integration_log_buffer(
     buffer.buffer.push(SidecarAction::Telemetry(action));
 }
 
+pub struct ShmCache {
+    pub last_updated: Instant,
+    pub integrations: std::collections::HashSet<String>,
+    pub composer_paths: std::collections::HashSet<PathBuf>,
```
bwoebi:
```suggestion
    pub integrations: HashSet<String>,
    pub composer_paths: HashSet<PathBuf>,
```
Nit, you've anyway `use`'d them at the top of the file.

ext/sidecar.c
```diff
@@ -402,9 +404,12 @@ void ddtrace_sidecar_submit_root_span_data_direct(ddtrace_root_span_data *root,
         }
     }
     if (changed || !root) {
-        ddtrace_ffi_try("Failed sending remote config data", ddog_sidecar_set_remote_config_data(&ddtrace_sidecar, ddtrace_sidecar_instance_id, &DDTRACE_G(sidecar_queue_id), service_slice, env_slice, version_slice, &DDTRACE_G(active_global_tags)));
+        ddtrace_ffi_try("Failed sending remote config data", ddog_sidecar_set_universal_service_tags(&ddtrace_sidecar, ddtrace_sidecar_instance_id, &DDTRACE_G(sidecar_queue_id), service_slice, env_slice, version_slice, &DDTRACE_G(active_global_tags)));
```
bwoebi: We changed this, but we need to remove the config check here `if (!ddtrace_sidecar || !get_global_DD_REMOTE_CONFIG_ENABLED()) {`

ext/ddtrace.h
```diff
@@ -1,5 +1,6 @@
 #ifndef DDTRACE_H
 #define DDTRACE_H
+#include "components-rs/common.h"
```
bwoebi: Minor nit, but code style: we're using <> includes for everything which is not a path relative to the current file.

components-rs/telemetry.rs
```diff
@@ -220,3 +231,127 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_add_integration_log_buffer(
...
+#[no_mangle]
+pub unsafe extern "C" fn ddog_sidecar_telemetry_cache_drop(_: Box<ShmCacheMap>) {}
+
+#[no_mangle]
+pub extern "C" fn ddog_sidecar_telemetry_config_sent(
+    cache: &ShmCacheMap,
+    service: CharSlice,
+    env: CharSlice,
+) -> bool {
+    let service_str = service.to_string();
```
bwoebi: You generally want `to_utf8_lossy` rather than to_string. And then you can also do `get((service_str, env_str))` Same in the function below.
Basically avoid creating new Strings. at all (reasonable) costs unless unavoidable.

components-rs/telemetry.rs
```diff
@@ -220,3 +231,127 @@ pub unsafe extern "C" fn ddog_sidecar_telemetry_add_integration_log_buffer(
...
+    pub config_sent: bool,
+    pub integrations: HashSet<String>,
+    pub composer_paths: HashSet<PathBuf>,
+    pub reader: OneWayShmReader<NamedShmHandle, CString>,
+}
+
+#[derive(Hash, Eq, PartialEq)]
+struct ShmCacheKey(String, String);
+
+impl Equivalent<ShmCacheKey> for (&String, &String) {
```
bwoebi:
```suggestion
impl Equivalent<ShmCacheKey> for (&str, &str) {
```
Not String - &str can be any string slice, &String means a reference to a string.

components-rs/telemetry.rs
bwoebi: Could you please use to_utf8_lossy() and not try_to_utf8?

Leiyks: I didn't use `to_utf8_lossy` because the equivalent trait does not apply to a String tuple if I use &str as mentioned in your other comment

bwoebi: It should return `Cow<str>` which you can call `.as_ref()` on to get a `&str`, or am I missing something?

Leiyks: Nevermind, I was using `into_owned()` ...

bwoebi (approved):
Nice - I fixed a windows issue, but now it compiles. Merging.

PR #3319, opened 2025-06-27
fix(profiling): validate `opline` before access

realFlowControl:
### Description

When collecting the stack trace outside of a vm interrupt, the `execute_data.opline` might be a dangling pointer. So far we had some mitigations in place for known cases, but there are some cases where this may still happen. Granted this is extremely rare, but annoying.

The root cause is that the `execute_data.opline` pointer has a different lifetime than the `execute_data.func.op_array` it points into.

This PR:
- adds an out-of-bounds check for accessing the `opline`
- removes all our mitigations
- runs the `allocation_generator_01.phpt` test only if PHP > 8.0.10 is used, as otherwise it crashes in PHP

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/stack_walking.rs
+++ b/profiling/src/profiling/stack_walking.rs
@@ -109,11 +109,10 @@ fn safely_get_opline(execute_data: &zend_execute_data) -> Option<&crate::binding
         return None;
     }
 
-    // Safety: `opcodes_start` is null checked and both point/pointed to the same allocated object.
-    let opline_offset = unsafe { execute_data.opline.offset_from(opcodes_start) };
-
-    // Check if `opline` is within the allocated opcodes array `op_array`
-    if opline_offset >= 0 && (opline_offset as u32) < op_array.last {
+    // Check if `opline` is within the allocated opcodes array `op_array`. `op_array.last` is the
+    // index of the last opcode, so valid range is [opcodes_start, opcodes_start + last]
+    let opcodes_end = unsafe { opcodes_start.add(op_array.last as usize) };
+    if execute_data.opline >= opcodes_start && execute_data.opline < opcodes_end {
         // Safety: we did our best we could to validate that this pointer is not NULL and not
         // dangling and actually pointing to the right kind of data. Otherwise this is the crash
         // site you are looking for.
--- a/profiling/src/profiling/stack_walking.rs
+++ b/profiling/src/profiling/stack_walking.rs
@@ -88,6 +88,42 @@ pub fn extract_function_name(func: &zend_function) -> Option<Cow<'static, str>> 
     }
 }
 
+/// Safely get opline reference with bounds checking to prevent segfaults on dangling pointers that
+/// have been observed when dereferencing `execute_data.opline` under some conditions.
+///
+/// Safety: Relies on the caller to only try and access an `opline` for a non internal function
+#[inline]
+fn safely_get_opline(execute_data: &zend_execute_data) -> Option<&crate::bindings::zend_op> {
+    if execute_data.opline.is_null() {
+        return None;
+    }
+
+    let func = unsafe { execute_data.func.as_ref()? };
+
+    // Safety: relies on the caller to only call for non internal functions. In that case, the
+    // `op_array` is used in the union.
+    let op_array = unsafe { &func.op_array };
+    let opcodes_start = op_array.opcodes;
+
+    if opcodes_start.is_null() {
+        return None;
+    }
+
+    // Safety: `opcodes_start` is null checked and both point/pointed to the same allocated object.
+    let opline_offset = unsafe { execute_data.opline.offset_from(opcodes_start) };
+
+    // Check if `opline` is within the allocated opcodes array `op_array`
+    if opline_offset >= 0 && (opline_offset as u32) < op_array.last {
+        // Safety: we did our best we could to validate that this pointer is not NULL and not
+        // dangling and actually pointing to the right kind of data. Otherwise this is the crash
+        // site you are looking for.
+        unsafe { Some(&*execute_data.opline) }
+    } else {
+        // Opline is out of bounds
+        None
+    }
+}
+
 unsafe fn extract_file_and_line(
     execute_data: &zend_execute_data,
 ) -> (Option<Cow<'static, str>>, u32) {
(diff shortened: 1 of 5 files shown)


profiling/src/profiling/stack_walking.rs
```diff
@@ -88,6 +88,42 @@ pub fn extract_function_name(func: &zend_function) -> Option<Cow<'static, str>>
     }
 }
 
+/// Safely get opline reference with bounds checking to prevent segfaults on dangling pointers that
+/// have been observed when dereferencing `execute_data.opline` under some conditions.
+///
+/// Safety: Relies on the caller to only try and access an `opline` for a non internal function
+#[inline]
+fn safely_get_opline(execute_data: &zend_execute_data) -> Option<&crate::bindings::zend_op> {
```
morrisonlevi: Not marked `unsafe` but probably should be.

bwoebi: Why should it? Assuming you have a valid zend_execute_data reference (which some other safe abstraction should have returned), it's a safe abstraction over it.

realFlowControl: I'd guess because of the safety comment I've added which is a precondition to safely call this function. It was there in the beginning, but I wanted to explicitly add and mark unsafe spots in the function to specifically describe why those calls are safe. I'd guess the safety comment there is enough from my POV, as the function is safe (if called for a non internal function 😜)

profiling/src/profiling/stack_walking.rs
```diff
@@ -88,6 +88,42 @@ pub fn extract_function_name(func: &zend_function) -> Option<Cow<'static, str>>
...
+    // `op_array` is used in the union.
+    let op_array = unsafe { &func.op_array };
+    let opcodes_start = op_array.opcodes;
+
+    if opcodes_start.is_null() {
+        return None;
+    }
+
+    // Safety: `opcodes_start` is null checked and both point/pointed to the same allocated object.
+    let opline_offset = unsafe { execute_data.opline.offset_from(opcodes_start) };
```
morrisonlevi: Is this true though? Isn't that why we have issues in the first place, because the opline doesn't point inside there?

morrisonlevi: > The requirement for pointers to be derived from the same allocated object is primarily needed for const-compatibility: the distance between pointers into different allocated objects is not known at compile-time. However, the requirement also exists at runtime and may be exploited by optimizations. If you wish to compute the difference between pointers that are not guaranteed to be from the same allocation, use `(self as isize - origin as isize) / size_of::<T>()`.

I think we need to do something like that instead, as the documentation recommends

bwoebi: The main problem is that we don't guarantee the alignment here.
I'd propose
```suggestion
    // Safety: `opcodes_start` is null checked and both point/pointed to the same allocated object.
    let opline_offset = unsafe { execute_data.opline.byte_offset_from(opcodes_start) / size_of::<T>() };
```
to avoid the casting. That function has no such requirements.

realFlowControl: Yes, it was pointing into it, but it might not anymore and then the alignment is not guaranteed. Thanks

profiling/src/profiling/stack_walking.rs
realFlowControl: @morrisonlevi I do get what you mean and even if we made this check anyway, it is better to not only document, but promote this to the calling site. I've update the call sites and marked the function unsafe.

profiling/src/profiling/stack_walking.rs
morrisonlevi: If I read this right, I don't think `byte_offset_from` is any better:

> Calculates the distance between two pointers within the same allocation. The returned value is in units of bytes.

They aren't in the same allocation. I think you need to do `(self as isize - origin as isize) / size_of::<T>()` as [offset_from](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset_from-1) recommends.

realFlowControl: We might not need to calculate an offset after all, just check the pointers:
```rust
    let opcodes_end = unsafe { opcodes_start.add(op_array.last as usize) };
    if execute_data.opline >= opcodes_start && execute_data.opline < opcodes_end {
```
This looks more like [PHP is doing it](https://heap.space/xref/PHP-8.0/ext/opcache/jit/zend_jit_trace.c?r=072dc3c857a00d687ba00a62e73d5051294bcdfb#7141-7142):
```c
ZEND_ASSERT(opline >= EX(func)->op_array.opcodes &&
	opline < EX(func)->op_array.opcodes + EX(func)->op_array.last);
```

morrisonlevi: In C, I believe it's implementation defined behavior to do this comparison. I'm not _sure_ what Rust says about it, but I think it's UB. I know you can use a pointer that is one-past the end of the end of the allocated object's range, and can do comparisons and arithmetic with it, but that doesn't mean you can do it with arbitrary pointers, like we're doing here.

realFlowControl: I've added a explicit cast to `usize`

profiling/tests/phpt/allocation_generator_01.phpt
```diff
@@ -13,17 +13,19 @@ restores the opline in a opcode handler!
 --SKIPIF--
 <?php
 if (!extension_loaded('datadog-profiling'))
-    echo "skip: test requires datadog-profiling", PHP_EOL;
+    die("skip: test requires datadog-profiling");
+if (PHP_VERSION_ID <= 80010)
+    die("skip: PHP is buggy");
```
morrisonlevi, line 18: What's the deal here for this test?

realFlowControl: The user opcode handlers have been removed, so this test in PHP <= 8.0.10 will segfault when printing the out of memory error message. Our user opcode handler "fixed" this crash and we'd collect stack traces in allocation profiling and see the error message.

morrisonlevi: Ah. I mean I guess this is a really rare condition because they have to OOM at a highly specific spot.

PR #3331, opened 2025-07-07
Fix rate limiter

estringana:
## Description
Fixes two bugs in the AppSec helper/extension pipeline.

The first is in ASM event propagation: when the WAF returns a `keep` decision and the rate limiter allows it, the helper sets `force_keep` in the response. The extension needs to act on that signal immediately to set trace priority and propagate headers to downstream services. Instead, the event was being emitted at shutdown — too late for header propagation.

The second is an engine hashing bug causing each FPM worker to create its own AppSec service instance instead of reusing the shared one, resulting in N parallel WAF instances where there should be one.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/commands_helpers.c
+++ b/appsec/src/extension/commands_helpers.c
@@ -461,9 +461,14 @@ dd_result dd_command_proc_resp_verd_span_data(
     mpack_node_t actions = mpack_node_array_at(root, RESP_INDEX_ACTION_PARAMS);
     dd_result res = _command_process_actions(actions, ctx);
 
-    if (res == dd_should_block || res == dd_should_redirect ||
-        res == dd_should_record) {
+    if (res == dd_should_block || res == dd_should_redirect) {
+        dd_tags_set_sampling_priority();
         dd_trace_emit_asm_event();
+        _set_appsec_span_data(
+            mpack_node_array_at(root, RESP_INDEX_APPSEC_SPAN_DATA));
+    }
+
+    if (res == dd_should_record) {
         _set_appsec_span_data(
             mpack_node_array_at(root, RESP_INDEX_APPSEC_SPAN_DATA));
     }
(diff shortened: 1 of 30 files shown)


appsec/src/extension/tags.c
```diff
@@ -960,6 +970,12 @@ static zval *nullable _root_span_get_meta(void)
     return meta;
 }
 
+static void _emit_auto_user_event(void)
+{
+    dd_tags_set_sampling_priority();
+    dd_trace_emit_asm_event();
```
cataphract: Isn't this duplicative? `dd_tags_set_sampling_priority` sets `force_keep` so that later we can call:

```c
dd_trace_set_priority_sampling_on_span_zobj(
    span, PRIORITY_SAMPLING_USER_KEEP, DD_MECHANISM_ASM);
```

But by calling `dd_trace_emit_asm_event()` we already do:

```c
if (DDTRACE_G(active_stack)->root_span) {
    DDTRACE_G(active_stack)->root_span->asm_event_emitted = true;
    ddtrace_trace_source_set_asm_source();
} else {
    DDTRACE_G(asm_event_emitted) = true;
}
 ddtrace_set_priority_sampling_on_root(PRIORITY_SAMPLING_USER_KEEP, DD_MECHANISM_ASM);
```

appsec/src/extension/commands_helpers.c
```diff
@@ -461,17 +461,23 @@ dd_result dd_command_proc_resp_verd_span_data(
...
         dd_trace_emit_asm_event();
         _set_appsec_span_data(
             mpack_node_array_at(root, RESP_INDEX_APPSEC_SPAN_DATA));
     }
 
+    if (res == dd_should_record) {
+        _set_appsec_span_data(
+            mpack_node_array_at(root, RESP_INDEX_APPSEC_SPAN_DATA));
+    }
+
```
cataphract: I'm not sure `dd_should_block` and `dd_should_redirect` should be treated differently, assuming the waf sends `keep` in the correct circumstances

cataphract (commented):
I think the logic should simply be that upon finding `force_keep` in helper messages (which was set when both 1) we had keep in the waf response and 2) the limiter says allow), we set the the thread local _force_keep (_force_keep = _force_keep || value_from_helper).

Then, at shutdown, we only need to call `ddtrace_emit_asm_event()`, as that sets the priority, the decision marker and the trace source.

cataphract (approved):
I think it's fine now, although you could add a test that verifies the tags when there's a keep response from the helper

PR #3339, opened 2025-07-11
Add code origins

bwoebi:
Tracking stacktraces of entry and exit spans.


ext/code_origins.h
```diff
@@ -0,0 +1,11 @@
+#ifndef DD_CODE_ORIGINS_H
+#define DD_CODE_ORIGINS_H
+
+#include "span.h"
+
+void ddtrace_add_code_origin_information(ddtrace_span_data *span, int skip_frames);
+void ddtrace_maybe_add_code_origin_information(ddtrace_span_data *span);
+void ddtrace_code_origins_minit();
+void ddtrace_code_origins_shutdown();
```
iamluc: unused?

ext/code_origins.c
```diff
@@ -0,0 +1,84 @@
...
+                    zend_string_release(key);
+                }
+                ++current_frame;
+            }
+        }
+        execute_data = EX(prev_execute_data);
+    }
+
+    // This typically happens when a file is the entrypoint.
+    if (current_frame != 0) {
```
iamluc, line 49: Shouldn't it be `current_frame == 0`?

ext/code_origins.c
```diff
@@ -0,0 +1,84 @@
...
+                    zend_array *parent_meta = ddtrace_property_array(&span->parent->property_meta);
+                    zval *parent_kind = zend_hash_str_find(parent_meta, ZEND_STRL("span.kind"));
+                    // span.kind must match, otherwise websocket has false positives for example, where both entry and exit span have type websocket
+                    if (!parent_kind || zend_is_identical(kind, parent_kind)) {
+                        return;
+                    }
+                } else {
+                    return;
+                }
+            }
```
iamluc: 🤯  I trust you 🙈

ext/span.c
```diff
@@ -552,6 +553,9 @@ void ddtrace_clear_execute_data_span(zend_ulong index, bool keep) {
     if ((Z_TYPE_INFO_P(span_zv) -= 2) == 1 || !keep) {
         if (!ddtrace_span_is_dropped(span)) {
             if (keep) {
+                if (&span->root->span != span) {
+                    ddtrace_maybe_add_code_origin_information(span);
```
iamluc, line 558: Since `ddtrace_close_span` is called just below, maybe we could move this check inside it?

It already calls `ddtrace_maybe_add_code_origin_information`, so centralizing the logic there might be cleaner.

iamluc (commented):
Except a few comments, LGTM 👍

ext/code_origins.c
Leiyks: I don't understand why the parent span would have the code origin information in all cases, can you please explain @bwoebi ? :thinking:

ext/configuration.h
```diff
@@ -251,6 +251,8 @@ enum ddtrace_sampling_rules_format {
     CONFIG(BOOL, DD_TRACE_INFERRED_PROXY_SERVICES_ENABLED, "false")                                            \
     CONFIG(SET, DD_TRACE_HTTP_CLIENT_ERROR_STATUSES, "500-599", .ini_change = zai_config_system_ini_change)    \
     CONFIG(SET, DD_TRACE_HTTP_SERVER_ERROR_STATUSES, "500-599", .ini_change = zai_config_system_ini_change)    \
+    CONFIG(BOOL, DD_CODE_ORIGIN_FOR_SPANS_ENABLED, "true")                                                     \
```
Leiyks, line 254: Should this feature be enabled by default, seeing benchmarks are all red ?

ext/code_origins.h
bwoebi: Yes, leftover.

ext/code_origins.c
bwoebi: No, this shall add the _dd.code_origin.type if any frames were added.

bwoebi: The comment is to say "it can sometimes be 0 when a file is the entrypoint", as to why the condition is necessary in the first place.

ext/span.c
bwoebi: The intention is to track root span closing and hook-spans, but nothing else. Do you have a suggestion how to centralize this nicely?

ext/code_origins.c
iamluc: Oh, OK. I thought it was a fallback in case no frames were added

ext/configuration.h
bwoebi: "all red" - it's a 2%-5% regression on a tight loop which has code origins handling now as a part of it.
This doesn't surprise me at all, but the macrobenches (wordpress, symfony) do not exhibit such large regressions.

ext/code_origins.c
bwoebi: Basically: we do not add it to child spans where the parent has the same attributes. So, when the parent has no such child, it gets added to the parent. Ending with the data being only on the parent.

bwoebi: Adjusted the comment to make the language a bit clearer.

PR #3352, opened 2025-07-22
Implement jwt

estringana:
### Description

[APPSEC-57285]

### Reviewer checklist
- [ ] Test coverage seems ok.

[APPSEC-57285]: https://datadoghq.atlassian.net/browse/APPSEC-57285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


appsec/src/helper/action.hpp
```diff
@@ -27,7 +27,7 @@ struct action {
 };
 
 struct event {
-    bool keep = false;
+    bool keep = true;
```
estringana: According to RFC this should default to true

appsec/tests/helper/engine_test.cpp
```diff
@@ -18,7 +18,7 @@
...
...th": [] } ], "regex": "^string.*" } } ] } ], "rules_compat": [{"id": "ttr-000-001", "name": "Trace Tagging Rule: Attributes, No Keep, No Event", "tags": {"type": "security_scanner", "category": "attack_attempt" }, "conditions": [{"operator": "match_regex", "parameters": {"inputs": [{"address": "arg4", "key_path": [] } ], "regex": "^string.*" } } ], "output": {"event": false, "keep": true, "attributes": {"_dd.appsec.trace.integer": {"value": 12345 }, "_dd.appsec.trace.string": {"value": "678" }, "_dd.appsec.trace.agent": {"address": "server.request.headers.no_cookies", "key_path": ["user-agent" ] } } }, "on_match": [] }, {"id": "ttr-000-002", "name": "Trace Tagging Rule: Attributes, No Keep, No Event", "tags": {"type": "security_scanner", "category": "attack_attempt" }, "conditions": [{"operator": "match_regex", "parameters": {"inputs": [{"address": "arg5", "key_path": [] } ], "regex": "^string.*" } } ], "output": {"event": false, "keep": false, "attributes": {"_dd.appsec.trace.integer": {"value": 12345 }, "_dd.appsec.trace.string": {"value": "678" }, "_dd.appsec.trace.agent": {"address": "server.request.headers.no_cookies", "key_path": ["user-agent" ] } } }, "on_match": [] } ] })";
```
estringana: Added `rules_compat`

appsec/src/helper/subscriber/waf.cpp
```diff
@@ -618,12 +619,34 @@ void instance::listener::call(
...
+                    SPDLOG_WARN("Schema for key {} is too large to submit",
+                        derivative.key());
+                }
             } else {
-                attributes_.emplace(derivative.key(), derivative);
+                if (derivative.is_signed()) {
+                    auto value =
+                        static_cast<double>(static_cast<int64_t>(derivative));
+                    metrics_attributes_.emplace(derivative.key(), value);
+                } else if (derivative.is_unsigned()) {
```
Anilm3, line 643: It could also be a floating point value.

appsec/src/helper/subscriber/waf.cpp
```diff
@@ -618,12 +619,34 @@ void instance::listener::call(
...
+                    auto value =
+                        static_cast<double>(static_cast<uint64_t>(derivative));
+                    metrics_attributes_.emplace(derivative.key(), value);
+                } else {
```
Anilm3: For compatibility reasons, explicitly check that the derivative is a string before adding it to meta and log an error / warning if it's an unexpected type.

appsec/src/helper/subscriber/waf.cpp
```diff
@@ -644,8 +644,16 @@ void instance::listener::call(
...
+                    auto value = static_cast<double>(derivative);
+                    metrics_attributes_.emplace(derivative.key(), value);
                 } else {
-                    meta_attributes_.emplace(derivative.key(), derivative);
+                    if (derivative.is_string()) {
+                        meta_attributes_.emplace(derivative.key(), derivative);
+                    } else {
+                        SPDLOG_WARN("Schema for key {} is not a valid type",
+                            derivative.key());
+                    }
```
Anilm3: I meant something along the lines of 

```cpp
} else if (derivative.is_string()) {
   ...
} else {
  // log error
}
```

Also why does the warning say `Schema for key`? I'd change it to "Unsupported attribute type for key {}"

appsec/src/helper/engine.cpp
```diff
@@ -119,7 +119,7 @@ std::optional<engine::result> engine::context::publish(
         }
     }
 
-    if (event_.actions.empty() && event_.data.empty()) {
+    if (!event_.keep && event_.actions.empty() && event_.data.empty()) {
```
cataphract: I don't get this logic. It we have no actions and triggers:

* if keep is false, we just send a single action with verdict OK (force_keep false)
* If keep is true, we send force_keep == limiter_.allow(), but an empty set of actions / triggers.

Is this right? At the very least this requires some comment.

(Also can we do something about the fact that event.data is results.events is response.triggers (we're calling the same thing three different names))

appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/mock_agent/rem_cfg/Capability.groovy
```diff
@@ -30,7 +30,9 @@ enum Capability {
     ASM_RASP_NOSQLI(27),
     ASM_RASP_XSS(28),
     APM_TRACING_SAMPLE_RULES(29),
-    CSM_ACTIVATION(30)
+    CSM_ACTIVATION(30),
+    ASM_DD_MULTICONFIG(42),
```
cataphract, line 34: You declare this new capability, but I see no other changes indicating that this is supported (and no tests).

cataphract (commented):
Also missing integration tests for the new functionality

appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/mock_agent/rem_cfg/Capability.groovy
estringana: I did change te test which checks capabilities on appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/RemoteConfigTests.groovy

appsec/src/helper/engine.cpp
estringana: The logic is ok. Now rules can override the output and set the keep to true

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/TelemetryTests.groovy
```diff
@@ -281,7 +281,7 @@ class TelemetryTests {
                 ]
         ])
 
-        def messages = waitForTelemetryLogs(30) { List<TelemetryHelpers.Logs> logs ->
+        def messages = waitForTelemetryLogs(50) { List<TelemetryHelpers.Logs> logs ->
```
estringana: With 30 it was randomly failing because it was getting 0 messages

appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/mock_agent/rem_cfg/Capability.groovy
cataphract: By supported I mean actually attempt to use the declared functionality (multiple ASM_DD configurations), not just testing that it's declared.

estringana: Can you give me another test example using other of those?

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/RemoteConfigTests.groovy
```diff
@@ -426,4 +426,132 @@ class RemoteConfigTests {
         applyRemoteConfig(target, [:])
     }
 
+    @Test
+    void 'test asm_dd_multiconfig'() {
```
cataphract, line 430: You could simplify this a bit and test only two user agents. I don't think there's much value in the third.

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/TelemetryTests.groovy
estringana: It's still even failing randomly... This started failing since we I pointed to the latest libdatadog

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/RemoteConfigTests.groovy
estringana: Ok

estringana: Done

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/TelemetryTests.groovy
estringana: It's failing on master. So it's not introduced here

PR #3356, opened 2025-07-24
fix(prof): Do not call `zend_jit_status()` on affected versions

realFlowControl:
### Description

There is a [use-after-free crash](https://github.com/php/php-src/pull/19212) which we need to work around.
I've already prepared for the upstream fix to be merged and added some version checks that are all commented out at this point in time and should have the comment removed and version number fixed once the upstream PR has landed.

PROF-12148

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/php_ffi.c
+++ b/profiling/src/php_ffi.c
@@ -482,19 +482,39 @@ void ddog_php_opcache_init_handle() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


morrisonlevi (approved):
Looks good. We don't parse units but in practice it probably works, since `strcmp(str, "0") == 0` would also match `0M`, `0K`, etc.

Issue #3360, opened 2025-07-25
[Bug]: Allocation profiling causes the issue on PHP 8.4.*


mharacewiat:
### Bug report

# Description

I'm observing issues with `dd-profiling v1.10.0` on `PHP 8.4.10` (supposedly on all `8.4.*` versions). Tested both CLI and FPM.

Function I use is `memory_get_peak_usage`, and it seems to behave as expected when extension is disabled and/or when either `DD_PROFILING_ENABLED` and `DD_PROFILING_ALLOCATION_ENABLED` are `false`.

# Observations

```shell
$  php -v
PHP 8.4.10-dev (cli) (built: Jul 25 2025 09:59:45) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.4.10-dev, Copyright (c) Zend Technologies
    with datadog-profiling v1.10.0, Copyright Datadog, by Datadog

$ php -r '$x = str_repeat("a", 1024*1024); var_dump(memory_get_peak_usage(false));'
int(0)

$  php -n -r '$x = str_repeat("a", 1024*1024); var_dump(memory_get_peak_usage(false));'
int(1449752)

$  DD_PROFILING_ENABLED=false php -r '$x = str_repeat("a", 1024*1024); var_dump(memory_get_peak_usage(false));'
int(1458040)

$  DD_PROFILING_ALLOCATION_ENABLED=false php -n -r '$x = str_repeat("a", 1024*1024); var_dump(memory_get_peak_usage(false));'
int(1449840)
```

### PHP version

8.4.10

### Tracer or profiler version

1.10.0

### Installed extensions

[PHP Modules]
Core
datadog-profiling
date
hash
json
pcre
random
Reflection
SPL
standard

[Zend Modules]
datadog-profiling

### Output of `phpinfo()`

datadog-profiling

Version => 1.10.0
Profiling Enabled => true
Profiling Experimental Features Enabled => false
Experimental CPU Time Profiling Enabled => true
Allocation Profiling Enabled => true
Timeline Enabled => true
Exception Profiling Enabled => true
I/O Profiling Enabled => false
Endpoint Collection Enabled => true
Platform's CPU Time API Works => true
Profiling Log Level => off
Profiling Agent Endpoint => http://localhost:8126/
Application's Environment (DD_ENV) =>  
Application's Service (DD_SERVICE) => Standard input code
Application's Version (DD_VERSION) =>  

Directive => Local Value => Master Value
[...]

PR #3364, opened 2025-07-28
fix(prof): Add running PHP language tests for profiler and fix discovered bugs

realFlowControl:
### Description

In the light of #3360 it came to my attention that we are not running the PHP language tests for the profiler.

This PR wires up the profiler **"PHP language tests"** job so the upstream PHP test suite actually runs with the profiling extension loaded, across the supported PHP versions and for both the `nts` and `zts` builds. Getting that suite green surfaced several real bugs (including genuine profiler crashes) plus a few environment issues, all addressed here. The xfail lists are documented in `profiling/tests/README.md`.

#### Profiler bugs fixed

**Leaked stderr fd → hang.** The profiler duplicates `STDERR_FILENO` to keep a logging target alive after PHP closes stderr, but used a plain `dup()` without `O_CLOEXEC`, so the duplicate leaked across `proc_open()`/`exec()` into child processes. `ext/curl/tests/curl_setopt_ssl.phpt` spawns an `openssl s_server`, which inherited the leaked fd and held a `run-tests.php` worker's pipe open, so the worker blocked on `pipe_read` until the 60-minute CI timeout. Fixed by using `fcntl(F_DUPFD_CLOEXEC)`.

**NULL file dereference in the timeline error observer (PHP 8.0 segfault).** The error observer called `CStr::from_ptr(file)` with no NULL check. An uncaught exception whose `file` property is NULL is reported as a fatal error with `file=NULL`, and `strlen(NULL)` segfaulted. PHP 8.0 only, where the observer receives a raw C string. Fixed with a `NULL` guard, regression test added.

**Async signals delivered to profiler helper threads (ZTS segfault).** The helper threads (`ddprof_time`, `ddprof_upload`) only masked the fixed set of signals the Zend Engine registers. A script using `pcntl_async_signals(true)` + `pcntl_signal(SIGCHLD, …)` could have `SIGCHLD` delivered to a helper thread, where pcntl's handler ran without a PHP/TSRM context and dereferenced the thread-local `PCNTL_G` → segfault (ZTS only). Fixed by blocking all non-fault signals on the helper threads so async signals go to a PHP thread, regression test added.

#### CI / harness fixes

**Wrong conf.d path.** `run_php_language_tests.sh` hardcoded `/opt/php/debug/conf.d` when removing the `rdkafka`/`memcached` extensions, but the profiler job runs `nts`/`zts` (different paths), so `rdkafka` stayed loaded and its `CONFWARN` output broke `Zend/tests/instantiate_all_classes.phpt`. The scan dir is now derived from the active PHP (`PHP_CONFIG_FILE_SCAN_DIR`).

**`SKIP_ONLINE_TESTS` typo.** The job exported `SKIP_ONLINE_TEST` (singular) while the tests check `SKIP_ONLINE_TESTS`, so online tests were never skipped (and had to be xfailed). Fixed, and the two online tests removed from the xfail list.

#### parallel, and why we disable it for the ZTS language tests

The CI image auto-loads `ext-parallel` on the `zts` build. parallel replaces `sapi_module.ub_write` at MINIT, and as a side effect, on ZTS the engine's `PHP Startup` diagnostics (deprecation/warning notices) are flushed *after* the script's first output instead of before it. ~56 upstream tests assert the normal ordering and therefore fail — purely because parallel is loaded; this reproduces on a vanilla ZTS PHP with no Datadog extension. Rather than xfail 56 tests, we remove `parallel.ini` for the language-test run (alongside `rdkafka`/`memcached`), which restores the correct ordering and keeps full coverage; these tests don't need parallel.

The former is not a problem for the tracer, because the tracer does not run ZTS upstream tests.

Also: UBSAN found a parallel bug when throwing exceptions from a parallel thread and was reported upstream and a fixed **parallel 1.2.14** has since been released — the profiler UBSAN job now `pecl install`s it over the image's version until the CI images are rebuilt.

#### xfails (not profiler bugs)

- **opcache optimizer dumps, PHP < 8.4** (`prop_types`, `gh11170`, `nullsafe_002`): the wall-time profiler hooks `zend_execute_internal` on < 8.4, so internal calls compile to `DO_FCALL` instead of `DO_ICALL` and the optimized-opcode dumps differ. xfailed for < 8.4 only.
- **`bug60634`** (`die()` in a session save handler): a parallel-run session-collision flake, not the profiler (passes in isolation with the profiler enabled); xfailed.

### Reviewer checklist
- [ ] Test coverage seems ok.

### AI Usage
99% of the PR have been written by Claude Opus 4.8, Claude Sonnet 4.6 and Kimi K2.5


.gitlab/generate-profiler.php
```diff
@@ -113,3 +125,38 @@
...
     - cargo test --all-features
+
+"PHP language tests":
+  stage: test
+  tags: [ "arch:${ARCH}" ]
+  image: registry.ddbuild.io/images/mirror/datadog/dd-trace-ci:php-${PHP_MAJOR_MINOR}_bookworm-8
+  variables:
+    KUBERNETES_CPU_REQUEST: 5
+    KUBERNETES_MEMORY_REQUEST: 3Gi
+    KUBERNETES_MEMORY_LIMIT: 4Gi
```
morrisonlevi: You want REQUEST to equal LIMIT and to specify all of them. Otherwise they are just suggestions and you won't be assigned the QoS which actually enforces these (which also ensures you are actually given what you are requested, as well).

.github/workflows/prof_asan.yml
```diff
@@ -127,6 +127,25 @@ jobs:
           cargo build --profile profiler-release
           cp -v "$CARGO_TARGET_DIR/profiler-release/libdatadog_php_profiling.so" "$(php-config --extension-dir)/datadog-profiling.so"
 
+      # TODO(parallel): the php-8.5_bookworm-8 image ships parallel 1.2.13, which
+      # has a bug that intermittently trips UBSAN. Install the fixed 1.2.14 over
+      # it (ZTS-only; parallel requires ZTS). Remove this step once the CI images
+      # are rebuilt with parallel >= 1.2.14.
```
morrisonlevi, line 133: You can add it to the Dockerfile now, so that it's one less thing to remember when they are rebuilt.

profiling/src/profiling/thread_utils.rs
```diff
@@ -20,25 +20,38 @@ where
...
+                // Hardware/synchronous fault signals: keep them deliverable to
+                // this thread.
+                const KEEP_UNBLOCKED: [libc::c_int; 6] = [
+                    libc::SIGSEGV,
+                    libc::SIGBUS,
+                    libc::SIGFPE,
+                    libc::SIGILL,
+                    libc::SIGABRT,
+                    libc::SIGTRAP,
                 ];
```
morrisonlevi, line 51: I think the comments here are overly verbose but yes, makes sense.

morrisonlevi (commented):
I did a pass over the code, have a few comments.

.github/workflows/prof_asan.yml
realFlowControl: The `Dockerfile`'s for our containers just do `pecl install parallel`, so once we rebuild them they'll pick up the latest release. Once 8.5.8 is released, we need to rebuild anyway and will magically get it.

bwoebi (approved):

PR #3366, opened 2025-07-29
feat(tracer): handle native HTTP requests

Leiyks:
### Description

This PR implement the instrumentation of native HTTP/HTTPS requests such as this:
```php
file_get_contents(“http[s]://“)
```

### Reviewer checklist
- [ ] Test coverage seems ok.

Related Jiras: [LANGPLAT-398]

[LANGPLAT-398]: https://datadoghq.atlassian.net/browse/LANGPLAT-398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ext/handlers_http.h
+++ b/ext/handlers_http.h
@@ -219,18 +219,18 @@ static inline zend_string *ddtrace_extract_header_key(char *header) {
     return key;
 }
 
-static inline HashTable *ddtrace_init_header_keys_set(zend_array *array, header_format format) {
+static inline void ddtrace_init_header_keys_set(zend_array *array, header_format format, HashTable *header_keys) {
     if (format == HEADER_MODE_KV_PAIRS) {
-        return NULL;
+        return;
     }
 
-    HashTable *set = zend_new_array(0);
+    zend_hash_init(header_keys, 8, NULL, NULL, 0);
 
     if (format != HEADER_MODE_ARRAY) {
-        return set;
+        return;
     }
 
-    ulong idx = 0;
+    uint32_t idx = 0;
     ZEND_HASH_FOREACH_VAL(array, zval *entry) {
         zend_string *header_key = ddtrace_extract_header_key(Z_STRVAL_P(entry));
         if (!header_key) {
@@ -240,12 +240,10 @@ static inline HashTable *ddtrace_init_header_keys_set(zend_array *array, header_
 
         zval index_zv;
         ZVAL_LONG(&index_zv, idx);
-        zend_hash_update(set, header_key, &index_zv);
+        zend_hash_update(header_keys, header_key, &index_zv);
         zend_string_release(header_key);
         idx++;
     } ZEND_HASH_FOREACH_END();
-
-    return set;
 }
 
 static inline void ddtrace_add_or_update_header_array(zend_array *array, HashTable *keys, const char *header_key, size_t header_key_len, zend_string *header) {
@@ -292,27 +290,6 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array, 
     zval headers;
     ZVAL_ARR(&headers, array);
     smart_str stream_header = {0};
-    HashTable *header_keys = ddtrace_init_header_keys_set(array, format);
-
-#define ADD_HEADER(header, ...) \
-    if (format == HEADER_MODE_KV_PAIRS) { \
-        add_assoc_str_ex(&headers, ZEND_STRL(header), zend_strpprintf(0, __VA_ARGS__)); \
-    } else if (format == HEADER_MODE_ARRAY) { \
-        ddtrace_add_or_update_header_array(Z_ARRVAL(headers), header_keys, ZEND_STRL(header), zend_strpprintf(0, header ": " __VA_ARGS__)); \
-    } else { \
-        if (stream_header.s) { \
-            smart_str_appendl(&stream_header, "\r\n", 2); \
-        } \
-        zend_string *formatted = zend_strpprintf(0, header ": " __VA_ARGS__); \
-        smart_str_append(&stream_header, formatted); \
-        \
-        zend_string *key = zend_string_init(ZEND_STRL(header), 0); \
-        zend_str_tolower(ZSTR_VAL(key), ZSTR_LEN(key)); \
-        zend_hash_add_empty_element(header_keys, key); \
-        zend_string_release(key); \
-        \
-        zend_string_release(formatted); \
-    }
 
     bool send_datadog = zend_hash_str_exists(inject, ZEND_STRL("datadog"));
     bool send_tracestate = zend_hash_str_exists(inject, ZEND_STRL("tracecontext"));
@@ -328,8 +305,30 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array, 
     ddtrace_root_span_data *root_span = DDTRACE_G(active_stack) ? DDTRACE_G(active_stack)->root_span : NULL;
     zend_array *meta = root_span ? ddtrace_property_array(&root_span->property_meta) : &DDTRACE_G(root_span_tags_preset);
     if (!get_DD_APM_TRACING_ENABLED() && !ddtrace_asm_event_emitted() && !ddtrace_trace_source_is_meta_asm_sourced(meta)) {
-        if (header_keys) zend_array_destroy(header_keys);
         return;
+    }
+
+    HashTable header_keys;
+    ddtrace_init_header_keys_set(array, format, &header_keys);
+
+#define ADD_HEADER(header, ...) \
+    if (format == HEADER_MODE_KV_PAIRS) { \
+        add_assoc_str_ex(&headers, ZEND_STRL(header), zend_strpprintf(0, __VA_ARGS__)); \
+    } else if (format == HEADER_MODE_ARRAY) { \
+        ddtrace_add_or_update_header_array(Z_ARRVAL(headers), &header_keys, ZEND_STRL(header), zend_strpprintf(0, header ": " __VA_ARGS__)); \
+    } else { \
+        if (stream_header.s) { \
+            smart_str_appendl(&stream_header, "\r\n", 2); \
+        } \
+        zend_string *formatted = zend_strpprintf(0, header ": " __VA_ARGS__); \
+        smart_str_append(&stream_header, formatted); \
+        \
+        zend_string *key = zend_string_init(ZEND_STRL(header), 0); \
+        zend_str_tolower(ZSTR_VAL(key), ZSTR_LEN(key)); \
+        zend_hash_add_empty_element(&header_keys, key); \
+        zend_string_release(key); \
+        \
+        zend_string_release(formatted); \
     }
 
     if (sampling_priority != DDTRACE_PRIORITY_SAMPLING_UNKNOWN) {
@@ -456,7 +455,7 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array, 
 
                 zend_string *key = ddtrace_extract_header_key(cursor);
                 if (key) {
-                    if (!zend_hash_exists(header_keys, key)) {
+                    if (!zend_hash_exists(&header_keys, key)) {
                         smart_str_appendl(&stream_header, "\r\n", 2);
                         smart_str_appendl(&stream_header, cursor, len);
                     }
@@ -474,8 +473,8 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array, 
         zend_hash_str_update(array, ZEND_STRL("header"), &result);
     }
 
-    if (header_keys) {
-        zend_array_destroy(header_keys);
+    if (format != HEADER_MODE_KV_PAIRS) {
+        zend_hash_destroy(&header_keys);
     }
 
 #undef ADD_HEADER
--- a/ext/handlers_httpstreams.c
+++ b/ext/handlers_httpstreams.c
@@ -1,7 +1,6 @@
 #include "configuration.h"
 #include "handlers_http.h"
 #include "serializer.h"
-#include "handlers_httpstreams.h"
 
 static php_stream_wrapper_ops traced_http_wrapper_wops;
 static php_stream_wrapper_ops traced_https_wrapper_wops;


ext/ddtrace.c
```diff
@@ -1367,6 +1367,108 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+        http_context_zv = zend_hash_str_find(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
+    }
+
+    if (http_context_zv && Z_TYPE_P(http_context_zv) == IS_ARRAY) {
+        HashTable *http_context = Z_ARRVAL_P(http_context_zv);
+        ddtrace_inject_distributed_headers(http_context, HEADER_MODE_CONTEXT);
+    }
+
+    // Open internal span
+    ddtrace_span_data *span = ddtrace_open_span(DDTRACE_INTERNAL_SPAN);
```
bwoebi: You should actually use `ddtrace_alloc_execute_data_span` and ddtrace_clear_execute_data_span() here, so that manual instrumentation (or instrumentation from appsec tracing these things) shares a span with this.

Use index `(zend_ulong)-2`.

bwoebi: Also, note that you then cannot use `_add_new`, because it already might exist on the span.

ext/ddtrace.c
```diff
@@ -1367,6 +1367,108 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+
+    if (!wrapper_zv) {
+        return;
+    }
+
+    php_stream_wrapper *orig = (php_stream_wrapper *)Z_PTR_P(wrapper_zv);
+    memcpy(original_wrapper, orig, sizeof(php_stream_wrapper));
+    memcpy(traced_wrapper, orig, sizeof(php_stream_wrapper));
+
+    traced_wrapper->wops = emalloc(sizeof(*orig->wops));
```
bwoebi: No emalloc here, it's a persistent allocation. You probably should just add another global for the php_stream_wrapper_ops, avoiding having to alloc and free it at all. (e.g. traced_wrapper_ops arg).

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+        context = php_stream_context_alloc();
+    }
+
+    // Inject distributed tracing headers
+    zval *http_context_zv = zend_hash_str_find(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
+    if (!http_context_zv) {
+        zval tmp;
+        array_init(&tmp);
+        zend_hash_str_update(Z_ARRVAL(context->options), "http", sizeof("http") - 1, &tmp);
+        http_context_zv = zend_hash_str_find(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
```
bwoebi:
```suggestion
        http_context_zv = zend_hash_str_find_deref(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
```
Might be reference, we need to guard against that. Also in the find above.

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+    if (!context) {
+        context = php_stream_context_alloc();
+    }
+
+    // Inject distributed tracing headers
+    zval *http_context_zv = zend_hash_str_find(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
+    if (!http_context_zv) {
+        zval tmp;
+        array_init(&tmp);
+        zend_hash_str_update(Z_ARRVAL(context->options), "http", sizeof("http") - 1, &tmp);
```
bwoebi: When altering an array, we need to make sure it's RC=1. See SEPARATE_ARRAY().

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+        http_context_zv = zend_hash_str_find(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
+    }
+
+    if (http_context_zv && Z_TYPE_P(http_context_zv) == IS_ARRAY) {
```
bwoebi:
```suggestion
    if (Z_TYPE_P(http_context_zv) == IS_ARRAY) {
```
We ensure that it's non-null on all paths.

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+        HashTable *http_context = Z_ARRVAL_P(http_context_zv);
```
bwoebi: Separate the array here as well.

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+            size_t host_len = host_end ? (size_t)(host_end - host_start) : strlen(host_start);
+            char *host = estrndup(host_start, host_len);
+            ZVAL_STRING(&zv, host);
+            zend_hash_str_update(meta, ZEND_STRL("network.destination.name"), &zv);
+            efree(host);
+        }
+    }
+
+    php_stream_wrapper *original_wrapper =
+        strncmp(filename, "https://", 8) == 0 ? &ddtrace_original_https_wrapper : &ddtrace_original_http_wrapper;
```
bwoebi: It would be cleaner to do two functions `ddtrace_stream_opener_http()` and `ddtrace_stream_opener_https()` which forward the original_wrapper to a common function. I'm not sure either whether that comparison universally holds.

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+    zend_string **opened_path,
+    php_stream_context *context,
+    int flags,
+    const char *persistent_id,
+    uint32_t persistent_id_len,
+    const char *wrapper_data,
+    uint32_t wrapper_data_len
+) {
+    if (!context) {
+        context = php_stream_context_alloc();
```
bwoebi: If you allocated your own, you'll also have to free it later.

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+            zend_hash_str_update(meta, ZEND_STRL("http.method"), method_zv);
+        }
+
+        const char *host_start = strstr(filename, "://");
+        if (host_start) {
+            host_start += 3; // skip "://"
+            const char *host_end = strchr(host_start, '/');
+            size_t host_len = host_end ? (size_t)(host_end - host_start) : strlen(host_start);
+            char *host = estrndup(host_start, host_len);
+            ZVAL_STRING(&zv, host);
```
bwoebi: this copies it twice. just do `ZVAL_STRINGL(&zv, host_start, host_len);`

ext/ddtrace.c
```diff
@@ -1367,6 +1367,131 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+        http_context_zv = zend_hash_str_find(Z_ARRVAL(context->options), "http", sizeof("http") - 1);
+    }
+
+    if (http_context_zv && Z_TYPE_P(http_context_zv) == IS_ARRAY) {
+        HashTable *http_context = Z_ARRVAL_P(http_context_zv);
+        ddtrace_inject_distributed_headers(http_context, HEADER_MODE_CONTEXT);
+    }
+
+    // Open internal span
+    ddtrace_span_data *span = ddtrace_alloc_execute_data_span(-2, NULL);
```
bwoebi: NULL? You mean `EG(current_execute_data)`?

ext/handlers_http.h
```diff
@@ -356,17 +370,22 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
     }
-    
+
     if (propagated_tags) {
         zend_string_release(propagated_tags);
     }
 
+    if (format == HEADER_MODE_CONTEXT && stream_header.s) {
+        smart_str_0(&stream_header);
+        add_assoc_str(&headers, "header", stream_header.s);
```
bwoebi: If I read that code correctly, you are unconditionally replacing it, rather than concatenating or pushing to (depending whether the original is string or array) the existing header value in the "http" context array.

Leiyks: Indeed, fixed it in the last commit :ok_hand:

tests/ext/integrations/http_streams/distributed_tracing_http_streams_with_headers_string.phpt
```diff
@@ -0,0 +1,82 @@
...
+            'method' => 'GET',
+            'header' => $headerString,
+        ],
+    ]);
+    return file_get_contents($GLOBALS['url'], false, $ctx);
+}
+
+$responses = [];
+
+$responses[] = fetch_with_header_string("x-foo: one\r\nx-bar: alpha");
```
bwoebi, line 37: can you actually test that 
```suggestion
$responses[] = fetch_with_header_string("x-foo: one\r\nx-bar: alpha\r\nx-datadog-trace-id: 123");
```
gets properly overwritten?

ext/handlers_http.h
```diff
@@ -229,12 +301,27 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
+        if (stream_header.s) { \
+            smart_str_appendl(&stream_header, "\r\n", 2); \
+        } \
+        zend_string *formatted = zend_strpprintf(0, header ": " __VA_ARGS__); \
+        smart_str_append(&stream_header, formatted); \
+        \
+        zend_string *key = zend_string_init(ZEND_STRL(header), 0); \
+        zend_str_tolower(ZSTR_VAL(key), ZSTR_LEN(key)); \
+        zend_hash_add_empty_element(header_keys, key); \
+        zend_string_release(key); \
```
bwoebi, line 322: This is unnecessary, you aren't ever going to insert a same header twice within this function, so we never need to insert these.

ext/handlers_http.h
```diff
@@ -229,12 +301,27 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
     ZVAL_ARR(&headers, array);
+    smart_str stream_header = {0};
+    HashTable *header_keys = ddtrace_init_header_keys_set(array, format);
 
 #define ADD_HEADER(header, ...) \
-    if (key_value_pairs) { \
+    if (format == HEADER_MODE_KV_PAIRS) { \
         add_assoc_str_ex(&headers, ZEND_STRL(header), zend_strpprintf(0, __VA_ARGS__)); \
+    } else if (format == HEADER_MODE_ARRAY) { \
+        ddtrace_add_or_update_header_array(Z_ARRVAL(headers), header_keys, ZEND_STRL(header), zend_strpprintf(0, header ": " __VA_ARGS__)); \
```
bwoebi: Can we provide the to-lower form directly in the macro to avoid lowercasing at runtime?
Then you also can use zend_hash_str_find() directly in ddtrace_add_or_update_header_array

ext/handlers_http.h
```diff
@@ -201,7 +201,79 @@ static inline zend_string *ddtrace_serialize_baggage(HashTable *baggage) {
...
+    zval *existing_idx_zv = zend_hash_find(keys, key);
+    if (existing_idx_zv && Z_TYPE_P(existing_idx_zv) == IS_LONG) {
+        zend_hash_index_del(array, Z_LVAL_P(existing_idx_zv));
+    }
+
+    // Update keys set
+    zend_long new_idx = zend_hash_num_elements(array);
+    zval idx_zv;
+    ZVAL_LONG(&idx_zv, new_idx);
+    zend_hash_update(keys, key, &idx_zv);
```
bwoebi: you may wrap this in `if (zend_hash_num_elements(keys) > 0)` to avoid some computations in a common case without pre-existing headers

bwoebi: Updating the keys set is unnecessary, you aren't ever going to insert a same header twice within this function, so we never need to insert these.

ext/handlers_http.h
```diff
@@ -356,17 +444,60 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
+                zend_string *line = zend_string_init(cursor, len, 0);
+                zval tmp;
+                ZVAL_STR(&tmp, line);
+
+                zend_string *key = ddtrace_extract_header_key(&tmp);
+                if (key) {
+                    if (!zend_hash_exists(header_keys, key)) {
+                        smart_str_appendl(&stream_header, "\r\n", 2);
+                        smart_str_appendl(&stream_header, ZSTR_VAL(line), ZSTR_LEN(line));
+                        zend_hash_add_empty_element(header_keys, key);
```
bwoebi: That's definitely not right. Headers are _generally_ allowed to exist multiple times. Just some headers, including these headers defined by this function must not be repeated.
I think removing this line of code is enough.

ext/handlers_http.h
```diff
@@ -356,17 +444,60 @@ static inline void ddtrace_inject_distributed_headers_config(zend_array *array,
...
+                if (len == 0) {
+                    cursor += eol ? 2 : 0;
+                    continue;
+                }
+
+                zend_string *line = zend_string_init(cursor, len, 0);
+                zval tmp;
+                ZVAL_STR(&tmp, line);
+
+                zend_string *key = ddtrace_extract_header_key(&tmp);
```
bwoebi: I'd recommend passing `cursor, len` directly to ddtrace_extract_header_key, avoiding the extra allocation of line.

ext/handlers_http.h
```diff
@@ -201,7 +201,79 @@ static inline zend_string *ddtrace_serialize_baggage(HashTable *baggage) {
...
+    }
+
+    size_t key_len = (size_t)(colon - val);
+    zend_string *key = zend_string_init(val, key_len, 0);
+    zend_str_tolower(ZSTR_VAL(key), key_len);
+    return key;
+}
+
+static inline HashTable *ddtrace_init_header_keys_set(zend_array *array, header_format format) {
+    HashTable *set = zend_new_array(0);
```
bwoebi: We do not need the extra allocation at all in HEADER_MODE_KV_PAIRS.

ext/handlers_http.h
```diff
@@ -201,7 +201,67 @@ static inline zend_string *ddtrace_serialize_baggage(HashTable *baggage) {
...
+
+    return set;
+}
+
+static inline void ddtrace_add_or_update_header_array(zend_array *array, HashTable *keys, const char *header_key, size_t header_key_len, zend_string *header) {
+    // Check for duplicates
+    zval *existing_idx_zv = zend_hash_str_find(keys, header_key, header_key_len);
+    if (existing_idx_zv && Z_TYPE_P(existing_idx_zv) == IS_LONG) {
+        zend_hash_index_del(array, Z_LVAL_P(existing_idx_zv));
+    }
```
bwoebi:
```suggestion
    // Check for duplicates
    if (zend_hash_num_elements(keys) > 0) {
        zval *existing_idx_zv = zend_hash_str_find(keys, header_key, header_key_len);
        if (existing_idx_zv) {
           zend_hash_index_del(array, Z_LVAL_P(existing_idx_zv));
       }
    }
```

ext/handlers_http.h
```diff
@@ -201,7 +201,67 @@ static inline zend_string *ddtrace_serialize_baggage(HashTable *baggage) {
...
+
+    // Insert header
+    zval arr_zv;
+    ZVAL_ARR(&arr_zv, array);
+    add_next_index_str(&arr_zv, header);
```
bwoebi:
```suggestion
    zval zv;
    ZVAL_STR(&zv, header);
    zend_hash_next_index_insert(array, &zv);
```
Makes more sense to use the lower-level API, rather than wrapping and having that api unwrap again

ext/ddtrace.c
```diff
@@ -1367,6 +1367,160 @@ static void dd_disable_if_incompatible_sapi_detected(void) {
...
+
+    ddtrace_instrument_stream_wrapper(
+        "https",
+        &ddtrace_original_https_wrapper,
+        &ddtrace_traced_https_wrapper,
+        &traced_https_wrapper_wops,
+        ddtrace_stream_opener_https
+    );
+}
+
```
bwoebi: Could you please move this into a new handlers_httpstream.c? Let's avoid bloating ddtrace.c when it cna be easily self-contained :-D

PR #3373, opened 2025-08-05
Fix submission of telemetry logs from appsec

cataphract:
### Description

Depends on https://github.com/DataDog/libdatadog/pull/1172

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/extension/commands/config_sync.c
```diff
@@ -45,8 +45,14 @@ static dd_result _request_pack(mpack_writer_t *nonnull w, void *nonnull ctx_)
     // 1. rem_cfg_path
     dd_mpack_write_nullable_cstr(w, data->rem_cfg_path);
 
-    // 2. queue_id
-    mpack_write_u64(w, dd_trace_get_sidecar_queue_id());
+    // 2. telemetry_settings
+    struct telemetry_rc_info tel_rc_info = dd_trace_get_telemetry_rc_info();
```
estringana: Maybe extract this to its own method so it can be used from both places?

estringana (approved):

ext/sidecar.c
```diff
@@ -22,6 +23,14 @@ ddog_Endpoint *ddtrace_endpoint;
 struct ddog_InstanceId *ddtrace_sidecar_instance_id;
 static uint8_t dd_sidecar_formatted_session_id[36];
 
+#ifdef ZTS
+static __thread zend_string *last_service_name;
+static __thread zend_string *last_env_name;
+#else
+static zend_string *last_service_name;
+static zend_string *last_env_name;
+#endif
```
bwoebi: I'm not too happy with that, by now we duplicate the last service name quite a bit. Can we just use `DDTRACE_G(last_flushed_root_service_name)` and `DDTRACE_G(last_flushed_root_env_name)` and (maybe with a rename to `active_root_service_name` or something like that) just use that everywhere (and assign it next to ddog_sidecar_set_universal_service_tags as you're doing in this PR).

One question: do you actually need to access that one outside the request lifecycle or why did you make it persistent?

bwoebi: Otherwise, nice PR, I like it :-)

cataphract: > One question: do you actually need to access that one outside the request lifecycle or why did you make it persistent?

well the assignment depends at least on `ddog_sidecar_set_universal_service_tags` not failing, so at least for that reason the pointer may not be updated on rinit. To avoid a dangling pointer in those situations, I made it persistent. I did forget to free it though

bwoebi: Can we not have it dependent on the success of ddog_sidecar_set_universal_service_tags? This should really never fail unless the sidecar connection is fucked, at which point it doesn't matter anyway.

PR #3379, opened 2025-08-12
Fix SQLSRVIntegration resource handling

bwoebi:
There's no object for the ObjectKVStore here.

Despite being broken it just happened to happily for the simplest case of just one connection and one statement being active at the same time.

### DIFF
--- a/ext/ddtrace.stub.php
+++ b/ext/ddtrace.stub.php
@@ -804,6 +804,24 @@ namespace DDTrace {
      * @param array $tags A list of tags associated to the metric
      */
     function dogstatsd_set(string $metric, int $value, array $tags = []): void {}
+
+    /**
+     * Store data tied to a resource. Behaves like a weakmap, i.e. data is freed when the associated resource is freed.
+     *
+     * @param resource $resource Some resource
+     * @param string $key An arbitrary string key to uniquely match
+     * @param mixed $value The data to be stored
+     */
+    function resource_weak_store(mixed $resource, string $key, mixed $value): void {}
+
+    /**
+     * Get data tied to the resource.
+     *
+     * @param resource $resource Some resource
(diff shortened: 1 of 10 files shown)


ext/weak_resources.h
```diff
@@ -0,0 +1,11 @@
+#ifndef DD_WEAK_RESOURCES_H
+#define DD_WEAK_RESOURCES_H
+
+#include <Zend/zend_types.h>
+
+void ddtrace_weak_resouces_rinit(void);
+void ddtrace_weak_resouces_rshutdown(void);
```
PROFeNoM: Small typo: It's missing an `r` (`...resouces...` --> `...resources...`) in header and C files.

ext/ddtrace.stub.php
```diff
@@ -804,6 +804,24 @@ function dogstatsd_histogram(string $metric, float $value, array $tags = []): vo
...
     function dogstatsd_set(string $metric, int $value, array $tags = []): void {}
+
+    /**
+     * Store data tied to a resource. Behaves like a weakmap, i.e. data is freed when the associated resource is freed.
+     *
+     * @param resource $resource Some resource
+     * @param string $key An arbitrary string key to uniquely match
+     * @param mixed $value The data to be stored
+     */
+    function resource_weak_store(mixed $resource, string $key, mixed $value): void {}
```
PROFeNoM, line 815: Nit: We are calling the files (and C functions within those files) `weak_resource...`, but the PHP functions `resource_weak...`. Is there any reason for that? But that's really a nit; only because it's exposed in the API.

PROFeNoM (approved):
This looks good to me 👍 I think the integration would be even stronger with coverage for an interleaved 2-conn/2-stmt case (to fully lock in the fix), but that can come later. For now I’ll go ahead and approve, only leaving a couple of naming nits.

PR #3380, opened 2025-08-13
Reduce integrations overhead

bwoebi:
- Using static for integrations (avoiding carrying an Integration instance around).
- static Closures
- Simplify sandboxed calls, reduce them to minimum overhead
- Only allocate new runtime cache objects when actually needed

### DIFF
--- a/ext/hook/uhook_legacy.c
+++ b/ext/hook/uhook_legacy.c
@@ -17,6 +17,8 @@ typedef struct {
     bool run_if_limited;
     bool active;
     bool allow_recursion;
+    bool begin_hook_active;
+    bool end_hook_active;
 } dd_uhook_def;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


marmichalski:
Hey @bwoebi!

I did not investigate too much, but I think this PR is a cause of a change in behaviour in hooks using static methods.

We are using `Instrumentation` class that installs all hooks on specific classes/interfaces, something like:
```php
<?php 

final class Instrumentation
{
    public static function commandExecutors(): void
    {
        install_hook(
            CommandExecutor::class.'::execute',
            static function (HookData $hook): void {
                $span = $hook->span();
                // ...
                $span->meta['command_executor.class'] = self::class;
                // ...
            },
        );
    }
}
```

Previously `self` would refer to the actual CommandHandler object class, just like `$this` in non-static closure. Now `self` returns `Instrumentation` class.

Is this expected?

bwoebi:
@marmichalski Yes, that behaviour is now expected, if you want to use `self::` (and have it refer to the other class), just remove `static`.
There was quite some overhead from rebinding Closures to class scopes, so static Closures are no longer rebound - with the option to have non-static closures bind.

PR #3387, opened 2025-08-25
Properly cache the telemetry cache

bwoebi:
Have the sidecar alone govern the lifetime of the cache.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1401,6 +1401,7 @@ dependencies = [
  "backtrace",
  "blazesym",
  "cc",
+ "cc_utils",
  "chrono",
  "ddcommon 0.0.1",
  "ddtelemetry",
--- a/components-rs/common.h
+++ b/components-rs/common.h
@@ -441,6 +441,10 @@ typedef struct ddog_Tag {
   const struct ddog_DslString *value;
 } ddog_Tag;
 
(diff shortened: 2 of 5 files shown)


components-rs/telemetry.rs
```diff
@@ -261,60 +259,66 @@ pub extern "C" fn ddog_sidecar_telemetry_cache_new() -> Box<ShmCacheMap> {
...
             }
         }
+    }
+
+    let service_str = service.to_utf8_lossy();
+    let env_str = env.to_utf8_lossy();
 
-        return None;
+    // I hate you, borrow checker, you get an unsafe from me!
+    if let Some(cached_entry) = (&mut *(cache as *mut ShmCacheMap)).get_mut(&(service_str.as_ref(), env_str.as_ref())) {
```
bwoebi, line 306: @morrisonlevi If you know how to make that safe, I'm happy (without materializing the key in the case the entry already exists, that is). I absolutely do not get why the borrow checker tells me that I'd take a mut ref twice to the same value here. This if ... ought to be scoping the &mut ref, and allow me to access cache.entry() afterwards.

bwoebi: Ah, I found https://blog.polybdenum.com/2024/12/21/four-limitations-of-rust-s-borrow-checker.html - it's just a shortcoming of the borrow checker.

Leiyks (approved):

PR #3399, opened 2025-09-01
Exclude /vendor/ from code origins

bwoebi:
This also prevents our testsuite from being overly sensitive to dependency updates :-D


Leiyks (approved):

tests/snapshots/tests.integrations.symfony.v6_2.messenger_test.test_async_failure_consumer.json
```diff
@@ -309,38 +279,8 @@
...
-         "_dd.code_origin.frames.5.type": "Doctrine\\DBAL\\Connection",
-         "_dd.code_origin.frames.6.file": "{path}/tests/Frameworks/Symfony/Version_6_2/vendor/doctrine/dbal/src/Connection.php",
-         "_dd.code_origin.frames.6.line": "782",
-         "_dd.code_origin.frames.6.method": "update",
-         "_dd.code_origin.frames.6.type": "Doctrine\\DBAL\\Connection",
-         "_dd.code_origin.frames.7.file": "{path}/tests/Frameworks/Symfony/Version_6_2/vendor/symfony/doctrine-messenger/Transport/Connection.php",
-         "_dd.code_origin.frames.7.line": "263",
-         "_dd.code_origin.frames.7.method": "reject",
-         "_dd.code_origin.frames.7.type": "Symfony\\Component\\Messenger\\Bridge\\Doctrine\\Transport\\Connection",
+         "_dd.code_origin.frames.0.file": "{path}/tests/Frameworks/Symfony/Version_6_2/bin/console",
```
shatzi: we skip 3rd party frames. but the frame index should be the same so the UI can put it at the right location.
For example, if the error start with 4 3rd-party frames and only the 5 one is 1st-party, then it should be frames.4.... (starting at 0).

bwoebi: Does it look good now?

ext/code_origins.c
```diff
@@ -3,63 +3,86 @@
...
+        }
 
-                if (EX(func)->op_array.function_name) {
-                    key = zend_strpprintf(0, "_dd.code_origin.frames.%d.method", current_frame);
-                    ZVAL_STR_COPY(&zv, EX(func)->op_array.function_name);
-                    zend_hash_update(meta, key, &zv);
-                    zend_string_release(key);
-                }
+        // Heuristically exclude code outside of the git repository, essentially
+        const char *vendor = zend_memnstr(ZSTR_VAL(EX(func)->op_array.filename), ZEND_STRL("vendor"), ZSTR_VAL(EX(func)->op_array.filename) + ZSTR_LEN(EX(func)->op_array.filename));
```
jamisonbryant: Note, some older PHP applications still use "Vendor" (capitalized) instead of the more-modern "vendor" (lowercase), and `zend_memnstr` is case-sensitive as far as I am aware.

I am not a C/PHP internals expert, so I must simply ask: does the code here account for this edge case?

bwoebi: I never heard about /Vendor/. Is this an ancient version of composer which did that?

jamisonbryant: Actually, I just now learned that the _vendor dir in composer is configurable!_ I had no idea composer supported this. It's been this way in our code since 2014:

CleanShot 2025-09-05 at 08 55 58

I definitely acknowledge that Datadog can't support every possible value that people may want to name their vendor dir, but at least supporting both "vendor" and "Vendor" or using a case-insensitive comparison function would be greatly appreciated.

jamisonbryant: For a little more context, if it's helpful. This is direct from a CakePHP Core engineer:

> Yeah it is from the pre-composer times where cakephp had a 'Vendor' directory.
> where you could put your vendor files (as composer didn't exist right)
> you can't have both Vendor and vendor on windows/macos, and composer was still 'new' so we chose to reconfigure composer 🤷

So this is not something we can change easily on our end, for this application anyway.

PR #3403, opened 2025-09-03
Fix crash in header injection with invalid user headers

bwoebi:
Observed in crashtracker - type confusion.

### DIFF
--- a/ext/handlers_http.h
+++ b/ext/handlers_http.h
@@ -232,6 +232,11 @@ static inline void ddtrace_init_header_keys_set(zend_array *array, header_format
 
     uint32_t idx = 0;
     ZEND_HASH_FOREACH_VAL(array, zval *entry) {
+        if (Z_TYPE_P(entry) != IS_STRING) {
+            idx++;
+            continue;
+        }
+
         zend_string *header_key = ddtrace_extract_header_key(Z_STRVAL_P(entry));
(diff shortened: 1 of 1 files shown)


realFlowControl (commented):
Not related to this PR, but it feels like the `zend_hash_init()` in `ddtrace_init_header_keys_set()` and the `zend_hash_destroy()` are not symmetric?

```c
    if (format == HEADER_MODE_KV_PAIRS) {
        return;
    }
    zend_hash_init(header_keys, 8, NULL, NULL, 0);
    if (format != HEADER_MODE_ARRAY) {
        return;
    }
```

vs.

```c
    if (format != HEADER_MODE_ARRAY) {
        return;
    }
    zend_hash_init(header_keys, 8, NULL, NULL, 0);
```

And then call `zend_hash_destroy()` in `ddtrace_inject_distributed_headers_config()` in the same if conditobn

bwoebi:
@realFlowControl I dont get your question:
```
    if (format == HEADER_MODE_KV_PAIRS) {
        return;
    }

    zend_hash_init(header_keys, 8, NULL, NULL, 0);
```

```
    if (format != HEADER_MODE_KV_PAIRS) {
        zend_hash_destroy(&header_keys);
    }
```

the destroy and init are perfectly symmetric here?

PR #3418, opened 2025-09-17
feat(prof): add source code integration

realFlowControl:
### Description

This adds [source code integration](https://docs.datadoghq.com/integrations/guide/source-code-integration/?tab=php#embed-git-information-in-your-build-artifacts) feature to the PHP profiler via the environment variables `DD_GIT_COMMIT_SHA` and `DD_GIT_REPOSITORY_URL` or the equivalent INI settings `datadog.git_commit_sha` and `datadog.git_repository_url`. You can already use this feature via [the `DD_TAGS` environment variable](https://docs.datadoghq.com/integrations/guide/source-code-integration/?tab=php#configure-telemetry-tagging). 

image

### Reviewer checklist
- [x] Test coverage seems ok.

PROF-12551

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -701,14 +701,14 @@ extern "C" fn rinit(_type: c_int, _module_number: c_int) -> ZendResult {
             TAGS.set({
                 // SAFETY: accessing in RINIT after config is initialized.
                 let globals = GLOBAL_TAGS.deref();
-                let unified_service_tags_len = locals.service.is_some() as usize
+                let extra_tags_len = locals.service.is_some() as usize
                     + locals.env.is_some() as usize
                     + locals.version.is_some() as usize
                     + locals.git_commit_sha.is_some() as usize
                     + locals.git_repository_url.is_some() as usize;
 
                 let mut tags = Vec::new();
-                tags.reserve_exact(globals.len() + unified_service_tags_len + locals.tags.len());
+                tags.reserve_exact(globals.len() + extra_tags_len + locals.tags.len());
                 tags.extend_from_slice(globals.as_slice());
                 add_optional_tag(&mut tags, "service", &locals.service);
                 add_optional_tag(&mut tags, "env", &locals.env);


morrisonlevi (changes requested):
This doesn't integrate with the config module, so users cannot use INIs to configure this. That also means our `datadog-setup.php config` mode won't work with it either.

If we agree it should be done, then add these two variants to the `ConfigId` enum:

```rs
    GitCommitSha,
    GitRepositoryUrl,
```

Add some functions like:

```rs

/// # Safety
/// This function must only be called after config has been initialized in
/// rinit, and before it is uninitialized in mshutdown.
pub(crate) unsafe fn git_commit_sha() -> Option<String> {
    get_str(Env)
}

/// # Safety
/// This function must only be called after config has been initialized in
/// rinit, and before it is uninitialized in mshutdown.
pub(crate) unsafe fn git_repository_url() -> Option<String> {
    get_str(Env)
}
```

Add it to the ENTRIES array:

```rs
                zai_config_entry {
                    id: transmute::<ConfigId, u16>(GitCommitSha),
                    name: Env.env_var_name(),
                    type_: ZAI_CONFIG_TYPE_STRING,
                    default_encoded_value: ZaiStr::new(),
                    aliases: ptr::null_mut(),
                    aliases_count: 0,
                    ini_change: None,
                    parser: Some(parse_utf8_string),
                    displayer: None,
                    env_config_fallback: None,
                },
                zai_config_entry {
                    id: transmute::<ConfigId, u16>(GitRepositoryUrl),
                    name: Env.env_var_name(),
                    type_: ZAI_CONFIG_TYPE_STRING,
                    default_encoded_value: ZaiStr::new(),
                    aliases: ptr::null_mut(),
                    aliases_count: 0,
                    ini_change: None,
                    parser: Some(parse_utf8_string),
                    displayer: None,
                    env_config_fallback: None,
                },
```

Add it to the `env_var_name` match:

```rs
            GitCommitSha => b"DD_GIT_COMMIT_SHA\0",
            GitRepositoryUrl => c"DD_GIT_REPOSITORY_URL\0",
```

And add it to the test cases for names:

```rs
            (b"DD_GIT_COMMIT_SHA\0", "datadog.git_commit_sha"),
            (b"DD_GIT_REPOSITORY_URL\0", "datadog.git_repository_url"),
```

morrisonlevi (changes requested):
Also, I think maybe it should be moved to the part where service, env, and version are handled. Theoretically you could change files and reconfigure, yeah? At least the tracer didn't register this as a system INI. What do you think?

realFlowControl:
I did the changes towards this, especially that it makes the feature symmetric with the tracer

PR #3419, opened 2025-09-18
Support curl_multi_exec root spans

bwoebi:
Instead of updating the duration of the first invocation, close the span upon the final invocation.

### DIFF
--- a/tests/Integrations/Guzzle/V5/GuzzleIntegrationTest.php
+++ b/tests/Integrations/Guzzle/V5/GuzzleIntegrationTest.php
@@ -225,7 +225,6 @@ class GuzzleIntegrationTest extends IntegrationTestCase
             },
             __DIR__ . '/guzzle_in_distributed_web_request.php'
         );
-        print_r($traces);
 
         $this->assertOneSpan(
             $traces,


tests/Integrations/Guzzle/V5/GuzzleIntegrationTest.php
```diff
@@ -225,6 +225,7 @@ function ($execute) use (&$headers1, &$headers2) {
             },
             __DIR__ . '/guzzle_in_distributed_web_request.php'
         );
+        print_r($traces);
```
Leiyks: Leftover
```suggestion
```

Leiyks (approved):

src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -90,22 +110,30 @@ public static function init(): int
             Integration::handleInternalSpanServiceName($span, self::NAME);
             $span->meta[Tag::COMPONENT] = self::NAME;
             $span->peerServiceSources = HttpClientIntegrationHelper::PEER_SERVICE_SOURCES;
+
+            \DDTrace\collect_code_origins(1);
```
morrisonlevi, line 114: Code origins weren't enabled for this span before, right? Should we call this out in the title/description?

bwoebi: They were. Code origins are enabled for spans started via $hook->span(), but not start_span(), so we need to do it manually.

PR #3422, opened 2025-09-19
feat(prof): source code integration

morrisonlevi:
### Description

This builds on #3418. It's built on master with Florian's work cherry-picked.

This does a few things:
 - Organizes tags into 3 groups:
    1. Global tags (language, profiler version, etc).
    2. Unified service tags (service, env, and version).
    3. "Local" tags which come from `DD_TAGS` (which is where their name comes from, a copy of the tag is stored in the thread-local variables struct).
 - The `git.commit.sha` and `git.repository_url` tags are part of globals.
 - Adds `DD_GIT_COMMIT_SHA` and `DD_GIT_REPOSITORY_URL` properly to the config module so you can configure it with INI settings and our `datadog-setup.php config` mode. Florian just used the environment variable. The tracer supports it this way and I didn't want a discrepancy here either.
 - Fixes a crash which occurred if `DD_GIT_COMMIT_SHA` or `DD_GIT_REPOSITORY_URL` were set but had empty values.
 - Enhances tag warnings when they occur so we know which key is wrong (all of our keys are static anyway).
 - Adds tests for removing credentials from the git repository URL.

### Things to discuss
There are discrepancies between Florian's implementation and the [one recently added to the tracer](https://github.com/DataDog/dd-trace-php/blob/e4f6598187be7142001d036409e6ebb187c47748/ext/git.c#L142-L166). Florian's seems right to me because it better handles SCP-like URLs. I do not know yet if the tracer based its implementation on an RFC or other specification. Perhaps the choice to not handle SCP style URLs was intentional. In any case, Florian's implementation cuts off the protocol if `@` is used whereas the tracer's keeps it. This is another difference we should scrutinize.

Should these be request locals or globals? I suppose if you a reconfigure you may have changed the values, and you'd want it to get picked up. As-written with it being a global, it won't.

Also, the INI names by default are `datadog.git_commit_sha` and `datadog.git_repository_url`.  I think that's probably fine but I don't think it was ever discussed when the tracer added it, so throwing it out there in case anyone has concerns or opinons about it.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi:
Closing because I don't want to hijack the work Florian started.

PR #3423, opened 2025-09-19
refactor(prof): tag handling

morrisonlevi:
### Description

Tags are now divided into three groups:

 1. Globals which are language, profiler version, process ID, etc.
 2. Unified service tags which are service, env, and version.
 3. Locals, which are request-local tags where DD_TAGS are stored. Unified service tag are also request-local but stored separately.

Globals are initialized lazily. This moves more tags into global.

The warning for an invalid tag now includes the key name. All our tag keys are static, so there's not privacy concern to include it.

The `Vec<Tag>` now reserves exactly how much memory it needs. I don't expect this to be a meaningful performance change, just doing it while I'm here.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/lib.rs
+++ b/profiling/src/lib.rs
@@ -76,15 +76,35 @@ static mut RUNTIME_PHP_VERSION: &str = {
 };
 
 lazy_static! {
-    static ref LAZY_STATICS_TAGS: Vec<Tag> = {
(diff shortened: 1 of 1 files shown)


realFlowControl (commented):
✅

PR #3430, opened 2025-09-23
Optimize PDOIntegration::parseDsn()

bwoebi:
Avoid manual string parsing in a loop, just use a single regex.

And remove some redundant empty branch.

### DIFF
--- a/src/DDTrace/Integrations/PDO/PDOIntegration.php
+++ b/src/DDTrace/Integrations/PDO/PDOIntegration.php
@@ -206,10 +206,27 @@ class PDOIntegration extends Integration
         $span->meta[Tag::ERROR_TYPE] = get_class($pdoOrStatement) . ' error';
     }
 
+    const DSN_REGEX = <<<'REGEX'
+(\A
+    (?<engine>[^:]++):
+    (?:
+        (?:
+             (?:server|unix_socket|host(?:name)?)=(?<host>(?:[^;]*+(?:;;)?)++)
+            |port=(?<port>(?&host))
+            |charset=(?<charset>(?&host))
+            |(?:database|dbname)=(?<db>(?&host))
+            |driver=(?<driver>(?&host))
+            |(?&host) # host can actually be empty, supporting repeated or trailing semicolons
+        )
+        (?:;|\Z)
+    )++
+)xi
+REGEX;
+
     private static function parseDsn($dsn)
     {
-        if (\preg_match('(\A(?<engine>[^:]++)(?::(?:(?:(?:server|unix_socket|host(?:name)?)=(?<host>(?:[^;]*+(?:;;)?)++)|port=(?<port>(?&host))|charset=(?<charset>(?&host))|(?:database|dbname)=(?<db>(?&host))|driver=(?<driver>(?&host))|(?&host))(?:;|\Z))++)?)i', $dsn, $m)) {
-            $engine = $m['engine'];
+        if (\preg_match(self::DSN_REGEX, $dsn, $m)) {
+            $engine = $m['engine']; // If uri: is used we'll also land here, but it's deprecated and we don't support it
             $db = $m['db'] ?? "";
             $charset = $m['charset'] ?? "";
             $host = $m['host'] ?? "";
@@ -239,7 +256,7 @@ class PDOIntegration extends Integration
         } elseif ($iniDsn = ini_get("pdo.dsn.$dsn")) {
             $tags = self::parseDsn($iniDsn);
         } else {
-            // There's technically also uri: but we don't support it and it's anyway deprecated
+            // If we cannot find the ini
             $tags = [
                 Tag::DB_SYSTEM => 'other_sql',
                 Tag::DB_TYPE => 'other_sql',


src/DDTrace/Integrations/PDO/PDOIntegration.php
```diff
@@ -208,41 +208,34 @@ public static function detectError($pdoOrStatement, SpanData $span)
 
     private static function parseDsn($dsn)
     {
-        $engine = substr($dsn, 0, strpos($dsn, ':'));
-        $tags = ['db.engine' => $engine];
+        \preg_match('(\A(?<engine>[^:]++)(?::(?:(?:charset=(?<charset>[^;]++)|(?:database|dbname)=(?<db>[^;]++)|(?:server|unix_socket|host(?:name)?)=(?<host>[^;]++)|port=(?<port>[^;]++)|driver=(?<driver>[^;]++)|[^;]*+)(?:;|\Z))++)?)i', $dsn, $m);
```
cataphract: This looks very limited:

* does not support DSNs without colon (ini lookup for pdo.dsn.{value}
* does not support uri:
* the driver specific part does not support colon escapes with ;; (on the generic path for mysql, oci, etc.) and does not properly parse sqlite or psql (sqlite:/path sqlite::memory: , psql:host=localhost port=5432 dbname=foo)

bwoebi: Indeed, `(?:;|\Z)` should be `;*+\Z` instead.
Everything else was not supported in the existing code though. I'll have a look.

src/DDTrace/Integrations/PDO/PDOIntegration.php
```diff
@@ -208,41 +208,43 @@ public static function detectError($pdoOrStatement, SpanData $span)
...
+                $tags[Tag::DB_SYSTEM] = strtolower($driver);
+            }
+        } elseif ($iniDsn = ini_get("pdo.dsn.$dsn")) {
+            $tags = self::parseDsn($iniDsn);
+        } else {
+            // There's technically also uri: but we don't support it and it's anyway deprecated
+            $tags = [
+                Tag::DB_SYSTEM => 'other_sql',
+                Tag::DB_TYPE => 'other_sql',
+            ];
```
PROFeNoM, line 263: In the fallback where the DSN can't be parsed (and there's no `ini_get("pdo.dsn.$dsn")` hit), we currently return only `db.system`/`db.type`.
Previously, we also set `db.engine` from the prefix (substring before the first `:`). Shouldn't we keep `db.engine` to avoid a tagging regression in metrics & dashboards that key off `db.engine`?

src/DDTrace/Integrations/PDO/PDOIntegration.php
```diff
@@ -208,41 +208,43 @@ public static function detectError($pdoOrStatement, SpanData $span)
...
-        $engine = substr($dsn, 0, strpos($dsn, ':'));
-        $tags = ['db.engine' => $engine];
-        $dbSystem = self::DB_DRIVER_TO_SYSTEM[$engine] ?? 'other_sql';
-        $tags[Tag::DB_SYSTEM] = $dbSystem;
-        $tags[Tag::DB_TYPE] = $dbSystem;  // db.type is DD equivalent to db.system in OpenTelemetry, used for SQL spans obfuscation
-        $valStrings = explode(';', substr($dsn, strlen($engine) + 1));
-        foreach ($valStrings as $valString) {
-            if (!strpos($valString, '=')) {
-                continue;
+        if (\preg_match('(\A(?<engine>[^:]++)(?::(?:(?:(?:server|unix_socket|host(?:name)?)=(?<host>(?:[^;]*+(?:;;)?)++)|port=(?<port>(?&host))|charset=(?<charset>(?&host))|(?:database|dbname)=(?<db>(?&host))|driver=(?<driver>(?&host))|(?&host))(?:;|\Z))++)?)i', $dsn, $m)) {
```
PROFeNoM: I believe the regex tail `(?:;|\Z)` is a bit too string, as in that it only handles one trailing `;`. If the DSN (for whatever reason) ends with multiples semicolons, like:
```
mysql:host=127.0.0.1;port=3306;dbname=foo;;
```
then I believe the match would either fail (or backtrack a bunch). Perhaps the easiest fix would be to just allow "zero or more" trailing semicolons?
```
- ... (?:;|\Z)
+ ... ;*+\z
```

---

PS: a nit of mine would be to put the a top-level constant (e.g., `DSN_REGEX`), and make it multiline... Just for readability.

src/DDTrace/Integrations/PDO/PDOIntegration.php
bwoebi: The only case where the DSN cannot be parsed is if there's no colon in the dsn. So actually the case when you use a plain name, but the ini doesn't exist.

src/DDTrace/Integrations/PDO/PDOIntegration.php
bwoebi: Also trailing semicolons are supported thanks to (?&host), but it's not obvious. Made multiline and added a comment.

PROFeNoM: much clearer, thanks!

PROFeNoM (approved):

PR #3436, opened 2025-10-01
feat: add process tags to tracing payloads

PROFeNoM:
### Reviewer checklist
- [ ] Test coverage seems ok.


ext/process_tags.c
```diff
@@ -0,0 +1,356 @@
...
+        ptr += value_len;
+        
+        if (i < process_tags.count - 1) {
+            *ptr++ = ',';
+        }
+    }
+    *ptr = '\0';
+
+    // Create a persistent zend_string
+    process_tags.serialized = zend_string_init(buffer, total_len, 1);
```
bwoebi: The better way would be

```
    process_tags.serialized = zend_string_alloc(total_len, 1);
    char *ptr = ZSTR_VAL(process_tags.serialized);
```
rather than creating a temporary buffer, to copy and free it; right away allocate the zend_string and manipulate its contents.

ext/process_tags.c
```diff
@@ -0,0 +1,356 @@
...
+        free(normalized_value);
+    }
+}
+
+/**
+ * Comparison function for qsort to sort process tags by key
+ */
+static int compare_tags(const void *a, const void *b) {
+    const process_tag_entry_t *tag_a = (const process_tag_entry_t *)a;
+    const process_tag_entry_t *tag_b = (const process_tag_entry_t *)b;
```
bwoebi:
```suggestion
    const process_tag_entry_t *tag_a = a;
    const process_tag_entry_t *tag_b = b;
```
Needing an explicit cast of void pointers is so C++ :-D

ext/process_tags.c
```diff
@@ -0,0 +1,356 @@
...
+        entrypoint_name = get_basename(exe_path);
+        entrypoint_basedir = get_dirname(exe_path);
+    } else {
+        // Fallback: use argv[0] or executable_location if available
+        if (sapi_module.executable_location) {
+            entrypoint_name = get_basename(sapi_module.executable_location);
+            entrypoint_basedir = get_dirname(sapi_module.executable_location);
+        }
+    }
+#endif
```
bwoebi: This is not what we want to actually collect. We are never interested in /proc/self/exe, which always is the php (or webserver) executable.
For PHP specifically the entrypoint is the primary script executed, and actually not a process-global in non-CLI cases. ... which makes no sense for non-cli process tags, so they should be omitted then (???).
This one though, annoyingly, is not available at minit time (but at first_rinit, which is good enough).

PROFeNoM: You are absolutely right. Changed it to use `SG(request_info).path_translated` for the actual script, moved everything to `first_rinit()` like you suggested, and ditched the MINIT hook entirely.

But then, I went down a rabbit hole on your 'makes no sense for non-CLI" comment. I looked a bit deeper at [the Java Implementation](https://github.com/DataDog/dd-trace-java/pull/8698) and the [Collected Tags Table](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?tab=t.0#bookmark=id.7ebm0iear0d), and they do collect process tags for web servers like Tomcat and JBoss: they get `server.type` (the vendor) and other metadata. So I figured PHP-FPM etc. is kind of similar - one pool usually serves one app, and the SAPI name is basically our version of `server.type: tomcat`.

So right now, it collects full entrypoint info for CLI (script name, basedir, etc.) but for web SAPIs it just gets `server.type` (the SAPI name) and `entrypoint.workdir`. No script name for web since different requests = different scripts, like you said.

My thinking was that most production PHP is not CLI, so excluding web would mean this only helps a small portion of workloads. 

But if you think we should keep it CLI-only for simplicity, I'm happy to revert that part.
This is, anyway, something that will have to be perhaps discussed w/ the folks knowing more about that feature.

PR #3444, opened 2025-10-13
Print block_id

estringana:
### Description

Display the new `block_id` provided by the latest libddwaf version. This block id is provided on blocks and redirects under the name `security_response_id`. It is displayed to the user on html, json and redirections.

The placeholder for block id on the template is `[security_response_id]`. There can be multiple placeholders in the same template

Additionally, messaging around requests blocked/redirected has been reworded as some customer found them confusing.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/request_abort.c
+++ b/appsec/src/extension/request_abort.c
@@ -62,7 +62,8 @@ static const char static_error_html[] =
 static const char static_error_json[] =
     "{\"errors\": [{\"title\": \"You've been blocked\", \"detail\": \"Sorry, yo"
     "u cannot access this page. Please contact the customer service team. Secur"
-    "ity provided by Datadog.\", \"security_response_id\": \"[security_response_id]\"}]}";
+    "ity provided by Datadog.\", \"security_response_id\": "
+    "\"[security_response_id]\"}]}";
 
 static zend_string *_initial_cwd;
 static zend_string *_body_error_html_def;
@@ -540,14 +541,16 @@ static bool _abort_prelude(void)
         if (DDAPPSEC_G(during_request_shutdown)) {
             mlog(dd_log_info,
                 "Datadog blocked the request, but the response has already "
-                "been partially committed. No action required. Security Response ID: %s",
+                "been partially committed. No action required. Security "
+                "Response ID: %s",
                 _block_parameters->block_id
                     ? ZSTR_VAL(_block_parameters->block_id)
                     : "");
         } else {
             _emit_error(
                 "Datadog blocked the request, but the response has already "
-                "been partially committed. No action required. Security Response ID: %s",
+                "been partially committed. No action required. Security "
+                "Response ID: %s",
                 _block_parameters->block_id
                     ? ZSTR_VAL(_block_parameters->block_id)
                     : "");


appsec/src/extension/request_abort.c
```diff
@@ -75,6 +76,8 @@ static zend_string *_content_type_json_zstr;
 static THREAD_LOCAL_ON_ZTS int _response_code = DEFAULT_BLOCKING_RESPONSE_CODE;
 static THREAD_LOCAL_ON_ZTS dd_response_type _response_type =
     DEFAULT_RESPONSE_TYPE;
+static THREAD_LOCAL_ON_ZTS zend_string *_block_id = NULL;
+static THREAD_LOCAL_ON_ZTS zend_string *_block_id_default = NULL;
```
cataphract: This is assigned in dd_request_abort_startup, so it should not be thread-local. In fact, I'd say you don't need this at all, you can just use `zend_empty_string`

appsec/src/extension/request_abort.c
```diff
@@ -75,6 +76,8 @@ static zend_string *_content_type_json_zstr;
 static THREAD_LOCAL_ON_ZTS int _response_code = DEFAULT_BLOCKING_RESPONSE_CODE;
 static THREAD_LOCAL_ON_ZTS dd_response_type _response_type =
     DEFAULT_RESPONSE_TYPE;
+static THREAD_LOCAL_ON_ZTS zend_string *_block_id = NULL;
```
cataphract: Reset this variable first thing in every request (or during shutdown, ensuring it's always executed), just in case the variable is not cleaned up on on the normal paths. Otherwise, you may get a dangling pointer here.

appsec/src/extension/request_abort.c
```diff
@@ -214,15 +219,71 @@ void dd_set_block_code_and_type(int code, dd_response_type type)
...
+
+        const char *p = src;
+        size_t count = 0;
+        while ((p = strstr(p, placeholder)) != NULL) {
+            required_size += (p - src) + replacement_len;
+            p += placeholder_len;
+            src = p;
+            count++;
+        }
+        required_size += ZSTR_VAL(target) + old_len - src;
```
cataphract: a bit confusing, no? Just count the number of occurrences, then do `original_len - count * placeholder_len + count * replacement_len`

appsec/src/extension/request_abort.c
```diff
@@ -214,15 +219,71 @@ void dd_set_block_code_and_type(int code, dd_response_type type)
...
+
+    if (target_ptr_ptr && *target_ptr_ptr) {
+        zend_string *nonnull target = *target_ptr_ptr;
+        const char *placeholder = "{block_id}";
+        size_t placeholder_len = strlen(placeholder);
+        size_t replacement_len = ZSTR_LEN(block_id);
+
+        const char *src = ZSTR_VAL(target);
+        size_t old_len = ZSTR_LEN(target);
+        size_t required_size = 0;
```
cataphract: don't use `size` and `len` interchangeably. Reserve size for allocation sizes, including the NUL terminator

appsec/src/extension/request_abort.c
```diff
@@ -214,15 +219,71 @@ void dd_set_block_code_and_type(int code, dd_response_type type)
...
+        size_t count = 0;
+        while ((p = strstr(p, placeholder)) != NULL) {
+            required_size += (p - src) + replacement_len;
+            p += placeholder_len;
+            src = p;
+            count++;
+        }
+        required_size += ZSTR_VAL(target) + old_len - src;
+
+        if (count > 0) {
```
cataphract: prefer `if (count == 0) { return; }`

appsec/src/extension/request_abort.c
```diff
@@ -214,15 +219,71 @@ void dd_set_block_code_and_type(int code, dd_response_type type)
...
+            zend_string *new_zstr = zend_string_alloc(required_size, 0);
+
+            const char *src2 = ZSTR_VAL(target);
+            char *dst = ZSTR_VAL(new_zstr);
+
+            while ((p = strstr(src2, placeholder)) != NULL) {
+                size_t bytes = p - src2;
+                memcpy(dst, src2, bytes);
+                dst += bytes;
+                memcpy(dst, ZSTR_VAL(block_id), replacement_len);
```
cataphract: you could be a bit more consistent. Sometimes you refer tro block_id, then to replacement... Either do sth like:

```c
memcpy(dst, ZSTR_VAL(block_id), ZSTR_LEN(block_id));
```

or

```c
const char *replacement = ZSTR_VAL(block_id);
const size_t replacement_len = ZSTR_LEN(block_id);
// ...
memcpy(dst, replacement, replacement_len);
```

appsec/src/extension/request_abort.c
```diff
@@ -214,15 +219,71 @@ void dd_set_block_code_and_type(int code, dd_response_type type)
...
+                dst += replacement_len;
+                src2 = p + placeholder_len;
+            }
+            size_t remaining = ZSTR_VAL(target) + old_len - src2;
```
cataphract: This doesn't read well, again because of the mixing styles:

```c
const size_t src_end = ZSTR_VAL(target) + ZSTR_LEN(target);
size_t remaining = src_end - src2;
```

appsec/src/extension/request_abort.c
```diff
@@ -51,15 +51,16 @@ static const char static_error_html[] =
     "4842b7}footer{width:100%;text-align:center}footer "
     "p{font-size:16px}</style></head><body><main><p>Sorry, you cannot access "
     "this page. Please contact the customer service "
-    "team.</p></main><footer><p>Security provided by <a "
+    "team.</br> Block ID: {block_id}</p></main><footer><p>Security provided by "
```
Anilm3: This template hasn't been standardised, AFAIK?

estringana: Not yet. But when it gets done, all I will have to do is to change this

appsec/src/extension/request_abort.c
```diff
@@ -72,6 +73,8 @@ static zend_string *_content_length_zstr;
 static zend_string *_location_zstr;
 static zend_string *_content_type_html_zstr;
 static zend_string *_content_type_json_zstr;
+static zend_string *_block_id = NULL;
+static zend_string *_block_id_default = NULL;
```
cataphract: This is fine, but you could replace it so you don't have to assign it and waste storage with it:

```c
static inline zend_string *_block_id_get_default(void)
{
    return zend_empty_string;
}
```

or, depending on your feelings on macros

```c
#define BLOCK_ID_DEFAULT (zend_empty_string)
```

appsec/src/extension/request_abort.c
```diff
@@ -72,6 +73,8 @@ static zend_string *_content_length_zstr;
 static zend_string *_location_zstr;
 static zend_string *_content_type_html_zstr;
 static zend_string *_content_type_json_zstr;
+static zend_string *_block_id = NULL;
```
cataphract: `block_id` must remain thread local on ZTS.

Perhaps you could refactor this a bit and put all thread local related stuff inside its own struct. You could also save some TLS space with a pointer indirection:

```c
struct block_parameters {
    zend_string *nonnull block_id;
    dd_response_type response_type;
    int redirection_response_code;
    zend_string *nullable redirection_location;
};

static THREAD_LOCAL_ON_ZTS struct block_parameters *nullable _block_parameters;

static struct block_parameters *nonnull block_parameters_new(/* parameters */)
{
    struct block_parameters *p = emalloc(sizeof(*p));
    // set fields...
    return p;
}

// then at the end/start of the request, reset the pointer _block_parameters.
```

PR #3451, opened 2025-10-17
Reset ddtrace_endpoint properly

bwoebi:
Make sure that dogstatsd_endpoint and ddtrace_endpoint are always freed together and the existence governed by ddtrace_endpoint.

Follow-up to #3407.

### DIFF
--- a/ext/sidecar.c
+++ b/ext/sidecar.c
@@ -20,7 +20,7 @@
 ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 
 ddog_Endpoint *ddtrace_endpoint;
-ddog_Endpoint *dogstatsd_endpoint;
+ddog_Endpoint *dogstatsd_endpoint; // always set when ddtrace_endpoint is set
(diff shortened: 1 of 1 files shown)


ext/sidecar.c
```diff
@@ -57,6 +57,12 @@ static void ddtrace_set_resettable_sidecar_globals(void) {
     ddtrace_sidecar_instance_id = ddog_sidecar_instanceId_build(session_id, runtime_id);
 }
 
+static void dd_free_endpoints(void) {
+    ddog_endpoint_drop(ddtrace_endpoint);
+    ddog_endpoint_drop(dogstatsd_endpoint);
+    ddtrace_endpoint = NULL;
```
Leiyks, line 63: Isn't `dogstatsd_endpoint` supposed to be set to `NULL` as well here ?

bwoebi: No, and it's never checked against NULL either.

Leiyks (approved):

PR #3452, opened 2025-10-17
Upload SSI symbols to our backend

bwoebi:
Takes the debug syms from the builds and uploads them.


.gitlab/generate-package.php
```diff
@@ -1445,6 +1445,42 @@
...
+      else
+        echo "DATADOG_API_KEY_STAGING not set, skipping Datadog symbol upload"
+      fi
+    - |
+      if [ -n "${DATADOG_API_KEY_PROD:-}" ]; then
+        echo "Uploading debug symbols to Datadog..."
+        DATADOG_API_KEY=$DATADOG_API_KEY_PROD DATADOG_SITE=datadoghq.com DD_BETA_COMMANDS_ENABLED=1 datadog-ci elf-symbols upload --disable-git ./build
+      else
+        echo "DATADOG_API_KEY_PROD not set, skipping Datadog symbol upload"
+      fi
```
Leiyks: Don't we want an proper exit error when the API keys are not set ? 
That would avoid having a silently failing job

Leiyks (approved):

PR #3464, opened 2025-10-28
When helper is unavailable, avoid very long waits (> 7s)

cataphract:
While there is a backoff mechanism, in the appsec extension, it is not effective. If the helper goes away:

* All the workers will simultaneously try to connect
* They will wait 7.5 s before giving up

This means no requests at all will be served by the server at each point. This PR changes this in two ways:

* The timeouts are significantly tightened.
* After a failure is registered, only worker will try to connect.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/helper_process.c
+++ b/appsec/src/extension/helper_process.c
@@ -7,7 +7,9 @@
 // NOLINTNEXTLINE(misc-header-include-cycle)
 #include <components-rs/ddtrace.h>
 #include <php.h>
+#include <stdatomic.h>
 #include <stdbool.h>
+#include <sys/mman.h>
 
 #define HELPER_PROCESS_C_INCLUDES
 #include "compatibility.h"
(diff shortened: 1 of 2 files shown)


appsec/src/extension/helper_process.c
```diff
@@ -53,12 +66,22 @@ static void _register_testing_objects(void);
...
 
 void dd_helper_startup(void)
 {
+    _shared_state = mmap(NULL, sizeof(dd_helper_shared_state),
+        PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+
+    if (_shared_state == MAP_FAILED) {
+        _shared_state = NULL;
+        mlog_err(dd_log_error, "Failed to mmap shared state");
+    }
```
Anilm3, line 83: No per-process backoff if the mmap fails?

cataphract: My first version had a fallback to pemalloc(8, 1). But then I had to have another variable to determine if I should free with pefree munmap and gave up on the idea. It's very unlikely that the nmap fails anyway

PR #3470, opened 2025-11-05
chore!: upgrade to libdatadog v23, profiling uses zstd now

morrisonlevi:
### Description

This upgrades libdatadog to v23. Well,  to a version after v23.0.0 due to an issue in the cc_utils crate. Unfortunately it has a half-migration to the new naming and versioning scheme of the libdatadog crates, but that's what the tracer has migrated to, and felt it was better to match.

### Motivation

This means the profiler will emit profiles in zstd and not lz4.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/prof_correctness.yml
+++ b/.github/workflows/prof_correctness.yml
@@ -79,8 +79,8 @@ jobs:
               mkdir -p profiling/tests/correctness/"$test_case"/
               export DD_PROFILING_OUTPUT_PPROF=$PWD/profiling/tests/correctness/"$test_case"/test.pprof
               php -d extension=$PWD/target/release/libdatadog_php_profiling.so profiling/tests/correctness/"$test_case".php
-              if [ -f "$DD_PROFILING_OUTPUT_PPROF".1.lz4 ]; then
-                echo "File $DD_PROFILING_OUTPUT_PPROF.1.lz4 should not exist!"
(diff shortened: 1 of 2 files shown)


profiling/src/profiling/mod.rs
```diff
@@ -342,7 +342,8 @@ impl TimeCollector {
                 },
                 value: period.min(i64::MAX as u128) as i64,
             }),
-        );
+        )
+        .expect("failed to create a new InternalProfile object");
```
realFlowControl, line 346: I'd guess that before this PR we'd also panic in case creating a new profile fails, just what we panicked in libdatadog. If that's the case I'm fine with this `expect` here

realFlowControl: For future reference, we should at least make sure we see this in crash tracking to get an idea if that actually happens and then handle it gracefully

morrisonlevi: Yes, the panic was merely implicit before.

realFlowControl (approved):
I've added https://github.com/DataDog/dd-trace-php/pull/3470/commits/df8daa00c6fa3cffd1de0bd9572d7d1dd4b7f6df to fix the filenames for filesystem export.

Our `.so` file size grows from 10.5 MiB to 12.3 MiB according to the GitLab job artefacts. AFAIK size stopped to be an issue since we've change how we distribute, but we should definitely have a look afterwards.

PR #3474, opened 2025-11-07
feat(prof): switch panics to abort

realFlowControl:
### Description

Crashtracking showed us a crash during shutdown with the following stack trace:
```
#0   0x000073438a0969fc pthread_kill 
#1   0x000073438a042476 gsignal 
#2   0x000073438a0287f3 abort 
#3   0x0000734381cbe11a std::sys::pal::unix::abort_internal::h3363667ac34a1b25 (sys/pal/unix/mod.rs:373:14)
#4   0x0000734381cecc23 std::panicking::rust_panic_with_hook::h370ee1901241e459 (std/src/panicking.rs)
#5   0x0000734381ceca95 std::panicking::begin_panic_handler::{{closure}}::heabfe92676d6d073 (std/src/panicking.rs:667:13)
#6   0x0000734381ceca29 std::sys::backtrace::__rust_end_short_backtrace::h6e22d229d4fdf49e (src/sys/backtrace.rs:170:18)
#7   0x0000734381ceca1c rust_begin_unwind (std/src/panicking.rs:665:5)
#8   0x0000734381a674fc core::panicking::panic_nounwind_fmt::runtime::h9e8ab7853c82c295 (src/panicking.rs:119:22)
#9   0x0000734381a674fc core::panicking::panic_nounwind_fmt::h34b0e56fb1476b87 (intrinsics/mod.rs:3535:9)
#10  0x0000734381a6755c core::panicking::panic_nounwind::h10b530ef2064232a (src/panicking.rs:223:5)
#11  0x0000734381a6750c core::panicking::panic_cannot_unwind::h9e6fc13f1a787f17 (src/panicking.rs:315:5)
#12  0x0000734381abbb87 shutdown (profiling/src/lib.rs:1024:1)
#13  0x00005b6efdf0984e zend_llist_apply 
#14  0x00005b6efdeb47b2 php_module_shutdown 
#15  0x000073438a029e40 __libc_start_main 
#16  0x00005b6efdd80ab5 _start 
```

`profiling/src/lib.rs:1024:1` is just this line:

https://github.com/DataDog/dd-trace-php/blob/303d859a692dc5c7b7228f7abcabf6da22927670/profiling/src/lib.rs#L1023

This makes no sense and is not the crash site, so what is happening is the following: During PHP's `MSHUTDOWN` somewhere in the Rust code there is a `panic()` raised that tries to unwind the call stack until it reaches the `extern "C"` function from where it can't unwind further, as unwinding over the FFI boarder would be UB, so instead the panic panics as `panic_cannot_unwind` and finally `abort`'s. This leaves crashtracking with the stack trace we get that is not helpful.

So far we have not tried to recover from any `panic` and treated panics as fatal. Thats why we can switch building the profiler from `panic=unwind` (default) to `panic=abort`, which makes every panic abort and crashtracker will get more meaningful stack traces.

As a nice side effect binary size will go down by ~9% 🎉  

### Reviewer checklist
- [ ] Test coverage seems ok.

PROF-12991

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -73,19 +73,19 @@ services:
   # --- Alpine ---
   '8.0-alpine': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-8.0_alpine' }
   # --- Bookworm ---
-  '7.0-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-7.0_bookworm-5' }
-  '7.1-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-7.1_bookworm-5' }
-  '7.2-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-7.2_bookworm-5' }
-  '7.3-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-7.3_bookworm-5' }
-  '7.4-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-7.4_bookworm-5' }
-  '7.4-bookworm-shared-ext': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-7.4-shared-ext-5' }
-  '8.0-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-8.0_bookworm-5' }
-  '8.0-bookworm-shared-ext': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-8.0-shared-ext-5' }
-  '8.1-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-8.1_bookworm-5' }
-  '8.2-bookworm': { <<: *linux_php_service, image: 'datadog/dd-trace-ci:php-8.2_bookworm-5' }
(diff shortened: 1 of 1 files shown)


bwoebi:
I'm not exactly sure here, but I believe the root Cargo.toml (i.e. not the one in profiling/) is actually used for build-configuration.

At least, that's IIRC the reason why there's a dedicated tracer-release configuration, to not mess with profiling. (So: please check.)

.github/workflows/prof_asan.yml
```diff
@@ -45,8 +45,8 @@ jobs:
           triplet=$(uname -m)-unknown-linux-gnu
           RUST_NIGHTLY_VERSION="-2025-01-03"
           rustup component add rust-src --toolchain nightly${RUST_NIGHTLY_VERSION}
-          RUSTFLAGS='-Zsanitizer=address' cargo +nightly${RUST_NIGHTLY_VERSION} build -Zbuild-std --target $triplet --release
-          cp -v "$CARGO_TARGET_DIR/$triplet/release/libdatadog_php_profiling.so" "$(php-config --extension-dir)/datadog-profiling.so"
+          RUSTFLAGS='-Zsanitizer=address' cargo +nightly${RUST_NIGHTLY_VERSION} build -Zbuild-std=std,panic_abort --target $triplet --profile profiler-release
```
morrisonlevi, line 48: Why does the asan test build std with panic_abort but not others? Maybe because it's already rebuilding std or something?

bwoebi (approved):
Please merge this, it brings size down too.

PR #3477, opened 2025-11-08
build: add rust-src to base containers

morrisonlevi:
@### Description

This adds the `rust-src` component to our build containers. We do the convenient thing on alpine and rustup platforms, and do the manual install bits for other platforms.

I've rebuilt the images I know of except Buster, which wouldn't build. I assume that we've fully migrated to Bookworm by now?

### Motivation

We want to be able to build the profiler with `-Z build-std=std,panic_abort`. To be able to do this, we have to have the `rust-src` component in the images.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/dockerfiles/ci/buster/Dockerfile
+++ b/dockerfiles/ci/buster/Dockerfile
@@ -318,25 +318,6 @@ RUN mkdir -p -v "${CARGO_HOME}" "${RUSTUP_HOME}" \
 
 ENV PATH="/rust/cargo/bin:${PATH}"
 
-# Install rust-src manually, since it's not included in the offline installer.
-# Levi figured this out through reading the rustup script and trial and error.
-RUN rustver="1.84.1" \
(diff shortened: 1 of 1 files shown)


dockerfiles/ci/centos/7/base.Dockerfile
```diff
@@ -239,6 +239,25 @@ RUN source scl_source enable devtoolset-7 \
     && cd - \
     && rm -fr "$FILENAME" "${FILENAME%.tar.gz}"
 
+# Install rust-src manually, since it's not included in the offline installer.
+# Levi figured this out through reading the rustup script and trial and error.
+RUN rustver="1.84.1" \
```
bwoebi: Can we avoid hardcoding twice here and just use `${RUST_VERSION}` (defined a few lines above)?

morrisonlevi: Good point, oops.

PR #3479, opened 2025-11-10
Collect runtime crash frames

bwoebi:
Now based on newest libdatadog

### DIFF
--- a/tests/ext/crashtracker_segfault.phpt
+++ b/tests/ext/crashtracker_segfault.phpt
@@ -28,7 +28,7 @@ posix_setrlimit(POSIX_RLIMIT_CORE, 0, 0);
 
 $php = getenv('TEST_PHP_EXECUTABLE');
 $args = getenv('TEST_PHP_ARGS')." ".getenv("TEST_PHP_EXTRA_ARGS");
-$cmd = $php." ".$args." -r 'posix_kill(posix_getpid(), 11);'";
+$cmd = $php." ".$args." -r 'spl_autoload_register(function() { posix_kill(posix_getpid(), 11); }); class_exists(Test::class);'";
 system($cmd);
 
 $rr->waitForRequest(function ($request) {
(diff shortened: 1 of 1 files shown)


ext/crashtracking_frames.c
```diff
@@ -0,0 +1,189 @@
...
+#if PHP_VERSION_ID >= 80300
+    if (EG(filename_override)) {
+        // Add the current execution point to the frame so we don't lose it
+        zend_string *filename_override = EG(filename_override);
+        zend_long lineno_override = EG(lineno_override);
+        EG(filename_override) = NULL;
+        EG(lineno_override) = -1;
+
+        zend_string *filename = EG(filename_override);
+        zend_long lineno = zend_get_executed_lineno();
```
realFlowControl: The `zend_get_executed_lineno()` has no in-bounds check. For the profiler we found that sometimes the `opline` can be a dangling pointer and we found that in the Zend Engine (in some places) an access to the `opline` is guarded by an in-bounds check. I think we could crash again here, while collecting the runtime crash frames for a crash.

The profiler implementation is in https://github.com/DataDog/dd-trace-php/pull/3319

ext/crashtracking_frames.c
```diff
@@ -0,0 +1,189 @@
...
+                    if (zai_is_mapped(op, sizeof(zend_op))) {
+                        frame.line = op->lineno;
+                    } else {
+                        frame.column = -2;
+                    }
+                } else {
+                    frame.line = func->op_array.line_end;
+                }
+            } else {
+                frame.line = call->opline->lineno;
```
realFlowControl, line 113: I do not understand the `zai_is_mapped` call above, but this could also crash when the `opline` is dangling, see comment above.

ext/crashtracking_frames.c
realFlowControl: https://heap.space/xref/PHP-8.0/ext/opcache/jit/zend_jit_trace.c?r=072dc3c857a00d687ba00a62e73d5051294bcdfb#7141-7142
and
https://heap.space/xref/PHP-8.4/ext/zend_test/observer.c?r=8d30ed4f808da939f2fd648fcab6307921c787ad#71-75

bwoebi: I'm just dumb and completely misread that code when copy pasting. I'll fix this.

ext/crashtracking_frames.c
bwoebi: zai_is_mapped calls mincore() which also handles dangling pointers - worst case you get some garbage frame, but no crash.

ext/crashtracking_frames.c
bwoebi: Fixed.

PR #3487, opened 2025-11-13
feat: support OpenTelemetry Metrics

rachelyangdog:
### Description
- Adds support for OpenTelemetry Metrics integration with Datadog.

Changes: 
- Added new configurations `DD_METRICS_OTEL_ENABLED `(enable OTel metrics support) and `DD_HOSTNAME`
- Added OTel endpoint configuration according to the [OTel Metrics RFC](https://docs.google.com/document/d/1Zkpk6QdlRQ67459fqHk8_SlJKb3-ZqRg-3juLHtWBtU/edit?usp=sharing)
- Sets default metrics temporality to "delta" (according to [RFC](https://docs.google.com/document/d/1Zkpk6QdlRQ67459fqHk8_SlJKb3-ZqRg-3juLHtWBtU/edit?usp=sharing) 
 - Added auto configuring of OTLP endpoints using `DD_TRACE_AGENT_URL` and/or  `DD_AGENT_HOST`
 - Added gRPC and HTTP protocol and port selection 
- Added resource attribute mapping
- Added functions to support collection of telemetry metrics for configuration (actual implementation will be added in a separate PR)

Please refer to this [document](https://docs.google.com/document/d/1HjBpz26oSlbvJSh72VFy9ccbkgd00NHCRF9JPGp-boE/edit?usp=sharing) containing known discrepancies of the PHP OTel Metrics implementation 

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/src/DDTrace/OpenTelemetry/CompositeResolver.php
+++ b/src/DDTrace/OpenTelemetry/CompositeResolver.php
@@ -17,7 +17,7 @@ class DatadogResolver implements ResolverInterface
     private const DEFAULT_HOST = 'localhost';
     private const DEFAULT_SCHEME = 'http';
 
-    public function retrieveValue(string $name): mixed
+    public function retrieveValue(string $name)
     {
         if (!$this->isMetricsEnabled($name)) {
             return null;
--- a/src/DDTrace/OpenTelemetry/Configuration.php
+++ b/src/DDTrace/OpenTelemetry/Configuration.php
@@ -22,17 +22,21 @@ const OTEL_CONFIG_WHITELIST = [
 ];
 
 // Helper function to track config access
-function track_otel_config_if_whitelisted(string $name, mixed $value): void
+function track_otel_config_if_whitelisted(string $name, $value): void
 {
     if (in_array($name, OTEL_CONFIG_WHITELIST, true)) {
         // Convert value to string for telemetry
-        $value_str = match (true) {
-            is_bool($value) => $value ? 'true' : 'false',
-            is_null($value) => '',
-            is_array($value) => json_encode($value),
-            is_object($value) => get_class($value),
-            default => (string)$value,
-        };
+        if (is_bool($value)) {
+            $value_str = $value ? 'true' : 'false';
+        } elseif (is_null($value)) {
+            $value_str = '';
+        } elseif (is_array($value)) {
+            $value_str = json_encode($value);
+        } elseif (is_object($value)) {
+            $value_str = get_class($value);
+        } else {
+            $value_str = (string)$value;
+        }
 
         \dd_trace_internal_fn('track_otel_config', $name, $value_str);
     }


PROFeNoM:
@codex review

src/DDTrace/OpenTelemetry/Configuration.php
```diff
@@ -0,0 +1,150 @@
...
+    },
+    function (\DDTrace\HookData $hook) {
+        if (isset($hook->data) && $hook->returned !== null) {
+            track_otel_config_if_whitelisted($hook->data, $hook->returned);
+        }
+    }
+);
+
+// Hook Configuration::getInt
+\DDTrace\install_hook(
```
PROFeNoM: (Major)
I believe that we could prevent a lot of duplication that is happening in this whole section (from line 40 to 150). **For instance** (just a proposition, take it or not :+1:), we could replace the whole section by something like:
```php
function install_config_tracking_hook(string $methodName): void
{
    \DDTrace\install_hook(
        "OpenTelemetry\\SDK\\Common\\Configuration\\Configuration::$methodName",
        function (\DDTrace\HookData $hook) {
            $name = $hook->args[0] ?? null;
            if ($name && is_string($name)) {
                $hook->data = $name;
            }
        },
        function (\DDTrace\HookData $hook) {
            if (isset($hook->data) && $hook->returned !== null) {
                track_otel_config_if_whitelisted($hook->data, $hook->returned);
            }
        }
    );
}

foreach (['getString', 'getInt', 'getBoolean', 'getMixed', 'getMap', 'getList', 'getEnum'] as $method) {
    install_config_tracking_hook($method);
}
```

Disclaimer: Not tested.

src/DDTrace/OpenTelemetry/Configuration.php
```diff
@@ -0,0 +1,150 @@
...
+// Helper function to track config access
+function track_otel_config_if_whitelisted(string $name, mixed $value): void
+{
+    if (in_array($name, OTEL_CONFIG_WHITELIST, true)) {
+        // Convert value to string for telemetry
+        $value_str = match (true) {
+            is_bool($value) => $value ? 'true' : 'false',
+            is_null($value) => '',
+            is_array($value) => json_encode($value),
+            default => (string)$value,
```
PROFeNoM: (Minor)
What happens if an object is passed? I don't recall what would happen on `(string)$value` if `$value` is an object. 

Or perhaps it's not possible to have an object passed to this function?

Otherwise, I'd cool w/ somehting like `is_object($value)` => `get_class($value)`

src/DDTrace/OpenTelemetry/CompositeResolver.php
```diff
@@ -0,0 +1,125 @@
...
+use OpenTelemetry\Contrib\Otlp\OtlpUtil;
+use OpenTelemetry\API\Signals;
+
+\DDTrace\install_hook(
+    'OpenTelemetry\SDK\Common\Configuration\Resolver\CompositeResolver::__construct',
+    null,
+    function (\DDTrace\HookData $hook) {
+
+        $this->addResolver(new class () implements \OpenTelemetry\SDK\Common\Configuration\Resolver\ResolverInterface {
+            public function retrieveValue(string $name): mixed
```
PROFeNoM: (Major)
This method (`retrieveValue`) is... quite complex, and is handling multiple responsibilities.

For instance, I believe that extracting this into helper methods could make it more maintainable long-term:
```php
private const DEFAULT_PROTOCOL = 'http/protobuf';
private const GRPC_PORT = '4317';
private const HTTP_PORT = '4318';
private const DEFAULT_HOST = 'localhost';

private function resolveProtocol(bool $metricsSpecific): string 
{ 
    if ($metricsSpecific && Configuration::has('OTEL_EXPORTER_OTLP_METRICS_PROTOCOL')) {
        return Configuration::getEnum('OTEL_EXPORTER_OTLP_METRICS_PROTOCOL');
    }
    
    if (Configuration::has('OTEL_EXPORTER_OTLP_PROTOCOL')) {
        return Configuration::getEnum('OTEL_EXPORTER_OTLP_PROTOCOL');
    }
    
    return self::DEFAULT_PROTOCOL;
}

private function resolveAgentEndpoint(): array 
{ 
    // ... agent resolution logic
}

private function buildEndpoint(string $protocol, bool $isMetricsEndpoint): string 
{ 
    // ... endpoint building logic
}
```

What do you think?

src/DDTrace/OpenTelemetry/CompositeResolver.php
```diff
@@ -0,0 +1,125 @@
...
+                        }
+                    }
+
+                    // Build endpoint: {scheme}://{host}:{port}
+                    if ($host === '') {
+                        $host = 'localhost';
+                    }
+
+                    // Determine port based on protocol if not already set
+                    $port = ($protocol === 'grpc') ? '4317' : '4318';
```
PROFeNoM: (~Minor)
Are we sure `$protocol` is gonna be lower-cased? What if the protocol comes back with different casing (e.g., `gRPC`)? **I don't know if this is possible**, but perhaps it could be safer doing `strtolower($protocol)`? 

Also wondering: what if the scheme from `DD_TRACE_AGENT_URL` already includes a port? Should we parse and respect that? idk the impact

src/DDTrace/OpenTelemetry/Detectors/Environment.php
```diff
@@ -0,0 +1,29 @@
...
+    function (\DDTrace\HookData $hook) {
+        $attributes = [];
+
+        if (\dd_trace_env_config('DD_ENV')!='') {
+            $attributes['deployment.environment.name'] = \dd_trace_env_config('DD_ENV');
+        }
+
+        if (\dd_trace_env_config('DD_VERSION') != '') {
+            $attributes['service.version'] = \dd_trace_env_config('DD_VERSION');
+        }
```
PROFeNoM: (nit)
We're calling `dd_trace_env_config` twice here for both `DD_ENV` and `DD_VERSION`. I believe we could extract to a variable to avoid the double call

```php
$ddEnv = \dd_trace_env_config('DD_ENV');
if ($ddEnv !== '') {
    $attributes['deployment.environment.name'] = $ddEnv;
}

$ddVersion = \dd_trace_env_config('DD_VERSION');
if ($ddVersion !== '') {
    $attributes['service.version'] = $ddVersion;
}
```

it's rly a nit though.

src/DDTrace/OpenTelemetry/Detectors/Environment.php
```diff
@@ -0,0 +1,29 @@
...
+
+        foreach (\dd_trace_env_config('DD_TAGS') as $key => $value) {
+            $attributes[$key] = $value;
+        }
+
+        $builder = (new AttributesFactory)->builder($attributes);
+        $newResource = ResourceInfo::create($builder->build());
+        $resource = $hook->returned;
+        $resource = $resource->merge($newResource);
+        $hook->overrideReturnValue($resource);
```
PROFeNoM: (Nit/Minor)
This exact 5-line pattern for building and merging resources appears in all three detector files.

For maintainability in the future and not having to _think_ about changing at three different places, perhaps it could make sense to have a single source of truth for the resource merging logic? 

for instance, we could extract this to a shared helper
```php 
namespace DDTrace\OpenTelemetry\Detectors;

// [...imports...]

class DetectorHelper
{
    public static function mergeAndOverrideResource(\DDTrace\HookData $hook, array $attributes): void
    {
        if (empty($attributes)) {
            return; // No changes needed
        }
        
        $builder = (new AttributesFactory())->builder($attributes);
        $newResource = ResourceInfo::create($builder->build());
        $mergedResource = $hook->returned->merge($newResource);
        
        $hook->overrideReturnValue($mergedResource);
    }
}
```

and then in each detector:
```php
DetectorHelper::mergeAndOverrideResource($hook, $attributes);
```

or smthg like that. WDYT?

src/DDTrace/OpenTelemetry/Detectors/Service.php
```diff
@@ -0,0 +1,27 @@
...
+        $attributes = [];
+
+        $rootSpan = \DDTrace\root_span();
+        if ($rootSpan) {
+            $attributes['service.name'] = $rootSpan->service;
+        } else {
+            if (ddtrace_config_app_name() === '') {
+                return;
+            }
+            $attributes['service.name'] = \ddtrace_config_app_name();
```
PROFeNoM: (Nit)
We're calling `ddtrace_config_app_name`  twice here. Could extract to avoid the double call=
```php
} else {
    $appName = \ddtrace_config_app_name();
    if ($appName === '') {
        return;
    }
    $attributes['service.name'] = $appName;
}
```

tests/OpenTelemetry/Integration/MetricsTest.php
```diff
@@ -0,0 +1,198 @@
...
+    use TracerTestTrait;
+
+    private static function getOtelVersion(): array
+    {
+        if (!class_exists('OpenTelemetry\API\Common\Version')) {
+            return [0, 0, 0];
+        }
+        $version = \OpenTelemetry\API\Common\Version::VERSION;
+        $parts = explode('.', $version);
+        return [
```
PROFeNoM: (Nit/Minor)
Hummm. I believe it _might_ happen that the package includes pre-release suffixes (see https://packagist.org/packages/open-telemetry/opentelemetry). Perhaps we should consider handling that in case it ever happens again?
```php
$version = \OpenTelemetry\API\Common\Version::VERSION;
// Strip any pre-release suffixes
$parts = explode('.', $version);
```

ext/ddtrace.c
```diff
@@ -2851,6 +2851,25 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
+            }
+        } else if (params_count == 3 && FUNCTION_NAME_MATCHES("track_telemetry_metrics")) {
+            zval *metric_name = ZVAL_VARARG_PARAM(params, 0);
+            zval *metric_value = ZVAL_VARARG_PARAM(params, 1);
+            zval *tags = ZVAL_VARARG_PARAM(params, 2);
+            if (Z_TYPE_P(metric_name) == IS_STRING && Z_TYPE_P(tags) == IS_STRING) {
+                ddtrace_metric_register_buffer(Z_STR_P(metric_name), DDOG_METRIC_TYPE_COUNT, DDOG_METRIC_NAMESPACE_TRACERS);
+                ddtrace_metric_add_point(Z_STR_P(metric_name), zval_get_double(metric_value), Z_STR_P(tags));
+                RETVAL_TRUE;
+            }
```
PROFeNoM, line 2872: (Major)
I believe we should type-validate `metric_value` here, just as we did for `metric_name` and `tags`, because currently we are calling `zval_get_double(metric_value)` without verifying it's numeric.

This may or may not cause a crash if someone passes a non-numeric vakue.

src/DDTrace/OpenTelemetry/Configuration.php
rachelyangdog: Good catch! I think in PHP, `(string)$object` throws a fatal error if the object doesn't have a `__toString()` method. Since `getMixed` is one of the hooked methods, an object return value is definitely possible.
I added your suggestion of `is_object($value) => get_class($value)`. Thank you!

src/DDTrace/OpenTelemetry/CompositeResolver.php
rachelyangdog: The OTel spec defines protocols as lowercase, so I expect that it will always be `'grpc'`. But, I agree that to be cautious we could add `strtolower($protocol)`.

rachelyangdog: > Also wondering: what if the scheme from `DD_TRACE_AGENT_URL` already includes a port? Should we parse and respect that? idk the impact

We don't parse the port from `DD_TRACE_AGENT_URL`. We want to use the default OTLP ports. We are using `DD_TRACE_AGENT_URL` to extract the scheme and host.

src/bridge/_files_opentelemetry.php
```diff
@@ -3,8 +3,13 @@
 return [
     __DIR__ . '/../DDTrace/OpenTelemetry/Context.php',
     __DIR__ . '/../DDTrace/OpenTelemetry/Convention.php',
+    __DIR__ . '/../DDTrace/OpenTelemetry/CompositeResolver.php',
+    __DIR__ . '/../DDTrace/OpenTelemetry/Configuration.php',
     __DIR__ . '/../DDTrace/OpenTelemetry/SpanContext.php',
     __DIR__ . '/../DDTrace/OpenTelemetry/Span.php',
     __DIR__ . '/../DDTrace/OpenTelemetry/SpanBuilder.php',
     __DIR__ . '/../DDTrace/OpenTelemetry/CachedInstrumentation.php',
```
PROFeNoM, line 9: (Major++)
I believe we need to add `__DIR__ . '/../DDTrace/OpenTelemetry/Detectors/DetectorHelper.php',` there as well :+1:

ext/ddtrace.c
PROFeNoM: Bob just told me that "zval_get_double() is infallible and always safe to use", so perhaps I made you do this change for nothing, oops 😬

PR #3493, opened 2025-11-19
Amend security_response_id being release before displaying it

estringana:
### Description
This PR fixes two errors related to security_response_id memory managment:
- When emitting an error, shutdown handlers are called. Then security_response_id is free before the error message string is formed. When it is formed then security_response_id is pointing to garbage
- When headers are already sent, security_response_id was not freed

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/request_abort.c
+++ b/appsec/src/extension/request_abort.c
@@ -61,9 +61,9 @@ static const char static_error_html[] =
     "target=\"_blank\">Datadog</a></p></footer></body></html>";
 
 static const char static_error_json[] =
-    "{\"errors\": [{\"title\": \"You've been blocked\", \"detail\": \"Sorry, yo"
+    "{\"errors\":[{\"title\":\"You've been blocked\",\"detail\":\"Sorry, yo"
     "u cannot access this page. Please contact the customer service team. Secur"
-    "ity provided by Datadog.\"}], \"security_response_id\": "
+    "ity provided by Datadog.\"}],\"security_response_id\":"
     "\"[security_response_id]\"}";
 
 static zend_string *_initial_cwd;
@@ -471,8 +471,10 @@ zend_array *nonnull dd_request_abort_static_page_spec(
     }
 
     {
-        char buf[sizeof("18446744073709551615") - 1];
-        size_t len = sprintf(buf, "%zu", body_len);
+        // This magic number is the string representation of SIZE_MAX on 64 bit
+        // systems
+        char buf[sizeof("18446744073709551615")];
+        size_t len = snprintf(buf, sizeof(buf), "%zu", body_len);
         zend_string *s = zend_string_init(buf, len, 0);
         zval cont_len_zv;
         ZVAL_STR(&cont_len_zv, s);
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -71,18 +71,21 @@ workflow:
   - when: manual
     interruptible: true
 
+.on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible: &on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible
+  if: '($CI_COMMIT_REF_NAME == "master" && ($NIGHTLY_BENCHMARKS || $CI_PIPELINE_SOURCE != "schedule")) || $CI_COMMIT_REF_NAME =~ /^ddtrace-/'
+  when: always
+  interruptible: false
+
 .macrobenchmarks-rules: &macrobenchmarks-rules
-  - if: $CI_COMMIT_REF_NAME == "master" && $NIGHTLY_BENCHMARKS
-    when: always
-    interruptible: false
-  - if: $CI_COMMIT_REF_NAME == "master" && $CI_PIPELINE_SOURCE != "schedule"
-    when: always
-    interruptible: false
-  - if: $CI_COMMIT_REF_NAME =~ /^ddtrace-/
-    when: always
-    interruptible: false
+  - <<: *on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible
   - when: manual
     interruptible: true
+
+.pre-release-performance-quality-gates-rules: &pre-release-performance-quality-gates-rules
+  - <<: *on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible
+  # Unlike on "master", on nightly benchmarks, or on release branches there's
+  # no strict requirement to run the check unconditionally with "when: always".
+  - interruptible: true
 
 .microbenchmarks:
   stage: benchmarks
@@ -182,13 +185,13 @@ check-big-regressions:
 
 .macrobenchmarks:
   stage: benchmarks
-  rules: *macrobenchmarks-rules
-  tags: ["runner:apm-k8s-same-cpu"]
   needs:
     - job: "package extension: [amd64, x86_64-unknown-linux-gnu]"
       artifacts: true
     - job: datadog-setup.php
       artifacts: true
+  rules: *macrobenchmarks-rules
+  tags: ["runner:apm-k8s-same-cpu"]
   timeout: 1h
   image: $MACROBENCHMARKS_CI_IMAGE
   script:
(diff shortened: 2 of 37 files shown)


appsec/src/extension/request_abort.c
```diff
@@ -251,8 +251,15 @@ void dd_set_block_code_and_type(
 
 void dd_request_abort_rinit(void)
 {
-    _block_parameters_set(SECURITY_RESPONSE_ID_DEFAULT, DEFAULT_RESPONSE_TYPE,
-        DEFAULT_BLOCKING_RESPONSE_CODE, NULL);
+    if (_block_parameters) {
+        mlog(dd_log_warning,
+            "_block_parameters is not NULL when calling dd_request_abort_rinit "
+            "(shutdown did not run on prev request?)");
+        _block_parameters_free();
```
cataphract: should just set to NULL

appsec/src/extension/request_abort.c
```diff
@@ -434,7 +442,7 @@ void _request_abort_static_page(int response_code, int type)
         content_type = JSON_CONTENT_TYPE;
         body = _get_json_blocking_template();
     } else {
-        mlog(dd_log_error, "unknown response type (bug) %d", response_type);
+        mlog(dd_log_error, " response type (bug) %d", response_type);
```
cataphract: looks accidental

appsec/src/extension/request_abort.c
```diff
@@ -653,17 +689,15 @@ static void _emit_error(const char *format, ...)
...
 
     // not enough: EG(error_handling) = EH_SUPPRESS;
     _suppress_error_reporting();
-    php_verror(NULL, "", E_ERROR, format, args);
-
-    va_end(args);
+    _php_verror(E_ERROR, "%s", msg);
+    if (free_msg) {
+        efree(msg);
+    }
```
cataphract: this will never be reachable

appsec/src/extension/request_abort.c
```diff
@@ -600,6 +617,24 @@ static void _emit_error(const char *format, ...)
...
+        msg = safe_emalloc(len + 1, 1, 0);
+        vsnprintf(msg, len + 1, format, args2);
+        free_msg = true;
+    } else {
+        msg = buf;
+    }
+    va_end(args2);
+    va_end(args);
+
+    _block_parameters_free();
```
cataphract: this should not be needed, and to the extent that is (because somehow the rshutdown hook is skipped) we'd need to free a bunch of other stuff as well

appsec/src/extension/request_abort.c
```diff
@@ -600,6 +617,23 @@ static void _emit_error(const char *format, ...)
...
     va_list args;
     va_start(args, format);
+    va_list args2;
+    va_copy(args2, args);
+    // NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers,readability-magic-numbers)
+    char buf[512];
+    int len = vsnprintf(buf, sizeof(buf), format, args);
+    char *msg = NULL;
+    bool free_msg = false;
+    if (len >= (int)sizeof(buf)) {
```
cataphract, line 577: you should also check if vsnprintf returned a negative value

appsec/src/extension/request_abort.c
```diff
@@ -600,6 +617,23 @@ static void _emit_error(const char *format, ...)
...
+        msg = safe_emalloc(len + 1, 1, 0);
```
cataphract: `safe_emalloc(len, 1, 1)`. if you add 1 yourself you're skipping the overflow check of safe_emalloc

appsec/src/extension/request_abort.c
```diff
@@ -600,6 +617,23 @@ static void _emit_error(const char *format, ...)
...
+        vsnprintf(msg, len + 1, format, args2);
```
cataphract: check for a negative return here too

appsec/src/extension/request_abort.c
```diff
@@ -616,16 +650,17 @@ static void _emit_error(const char *format, ...)
...
+            }
             // fpm doesn't try to run the script if it sees this null
             SG(request_info).request_method = NULL;
             return;
         }
 #ifdef FRANKENPHP_SUPPORT
         if (strcmp(sapi_module.name, "frankenphp") == 0) {
-            php_verror(NULL, "", E_WARNING, format, args);
-            va_end(args);
+            _php_verror(E_WARNING, "%s", msg);
```
cataphract, line 620: you're not freeing msg here

appsec/src/extension/request_abort.c
```diff
@@ -653,17 +688,12 @@ static void _emit_error(const char *format, ...)
      * be a possibility, but it bypasses the value of error_reporting and is
      * always logged */
     {
-        va_list args2;
-        va_copy(args2, args);
-        php_verror(NULL, "", E_COMPILE_WARNING, format, args2);
-        va_end(args2);
+        _php_verror(E_COMPILE_WARNING, "%s", msg);
```
cataphract, line 651: same

appsec/src/extension/request_lifecycle.c
```diff
@@ -363,6 +362,8 @@ static void _do_request_finish_php(bool ignore_verdict)
         dd_trace_close_all_spans_and_flush();
         dd_request_abort_redirect();
     }
+
+    dd_request_abort_rshutdown();
```
cataphract: you're calling on the top as well. I don't think the call on top is correct, because the abort functions need block_parameters. This also makes `_reset_globals()` too early.

So maybe after this block:

```c
    if (_enabled_user_req) {
        if (_cur_req_span) {
            mlog_g(dd_log_info,
                "Finishing user request whose corresponding "
                "span is presumably still unclosed on rshutdown");
            _do_request_finish_user_req(true, _superglob_equiv, 0, NULL, NULL);
            _reset_globals();
        }
    } else {
        _do_request_finish_php(ignore_verdict);
        // _rest_globals already called
    }
    // ADDED:
    _block_parameters_free()
```

OTOH this will not be reached if we call _emit_error from `_do_request_finish_php`... so I guess sth like:

```diff
--- a/appsec/src/extension/request_abort.c
+++ b/appsec/src/extension/request_abort.c
@@ -375,11 +375,13 @@ void dd_request_abort_redirect(void)
                 ? ZSTR_VAL(_block_parameters->security_response_id)
                 : "");
     } else {
+        __auto_type *block_parameters = _block_parameters;
+        _block_parameters = NULL;
         _emit_error("Datadog blocked the request and attempted a redirection "
                     "to %s. No action required. Security Response ID: %s",
-            ZSTR_VAL(_block_parameters->redirection_location),
-            _block_parameters->security_response_id
-                ? ZSTR_VAL(_block_parameters->security_response_id)
+            ZSTR_VAL(block_parameters->redirection_location),
+            block_parameters->security_response_id
+                ? ZSTR_VAL(block_parameters->security_response_id)
                 : "");
     }
 }
@@ -462,10 +464,12 @@ void _request_abort_static_page(int response_code, int type)
                 ? ZSTR_VAL(_block_parameters->security_response_id)
                 : "");
     } else {
+        __auto_type *block_parameters = _block_parameters;
+        _block_parameters = NULL;
         _emit_error("Datadog blocked the request and presented a static error "
                     "page. No action required. Security Response ID: %s",
-            _block_parameters->security_response_id
-                ? ZSTR_VAL(_block_parameters->security_response_id)
+            block_parameters->security_response_id
+                ? ZSTR_VAL(block_parameters->security_response_id)
                 : "");
     }
 }
@@ -552,12 +556,14 @@ static bool _abort_prelude(void)
                     ? ZSTR_VAL(_block_parameters->security_response_id)
                     : "");
         } else {
+            __auto_type *block_parameters = _block_parameters;
+            _block_parameters = NULL;
             _emit_error(
                 "Datadog blocked the request, but the response has already "
                 "been partially committed. No action required. Security "
                 "Response ID: %s",
-                _block_parameters->security_response_id
-                    ? ZSTR_VAL(_block_parameters->security_response_id)
+                block_parameters->security_response_id
+                    ? ZSTR_VAL(block_parameters->security_response_id)
                     : "");
         }
         return false;
```

appsec/src/extension/request_abort.c
estringana: I can't as on this case, the same msg is used for the next `_php_verror`

appsec/src/extension/request_lifecycle.c
estringana: I don't get this last comment

PR #3494, opened 2025-11-20
fix: code_origin_for_spans_enabled naming inconsistency

morrisonlevi:
### Description

This fixes a typo where the name sometimes uses plural "origins" instead of singular "origin":

- `datadog.code_origins_for_spans_enabled` (incorrect - has typo, used in `ext/configuration.c`)
- `datadog.code_origin_for_spans_enabled` (correct, used in `components-rs/remote_config.rs`)

It's difficult for me to explain what happens when some paths are using plural while some are using singular. I am not familiar with how this remote config code works or is supposed to work. This  is definitely a bug, and _may_ be the reason for some "Out of memory" and "zend_mm_heap corrupted".

### Additional Notes

This is not based on the latest master, but on tag 1.14.0, so that build artifacts are simply 1.14.0 with this change.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl (approved):
nice catch

PR #3497, opened 2025-11-24
feat(CI): add service healthcheck and better retry policy

Leiyks:
### Description

This PR add some stability to the CI by:

- Adding proper healthcheck when starting external services and test agent
- Improving retry policy on failing jobs
- Fix some timeout test that were successful but waiting for background command to finish
- Make jobs uninterruptible on master branch 
- Fix System test missing dependencies

Jiras: [LANGPLAT-643]

[LANGPLAT-643]: https://datadoghq.atlassian.net/browse/LANGPLAT-643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/generate-common.php
+++ b/.gitlab/generate-common.php
@@ -52,6 +52,7 @@ default:
       - api_failure
       - script_failure
       - stuck_or_timeout_failure
+      - job_execution_timeout
 
 .all_targets: &all_minor_major_targets
 <?php
@@ -153,8 +154,8 @@ foreach ($arch_targets as $arch_target) {
       KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
       KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
       KAFKA_AUTO_CREATE_TOPICS_ENABLE: true
-      KAFKA_ZOOKEEPER_CONNECTION_TIMEOUT_MS: 180000
-      KAFKA_ZOOKEEPER_SESSION_TIMEOUT_MS: 180000
+      KAFKA_ZOOKEEPER_CONNECTION_TIMEOUT_MS: 120000
+      KAFKA_ZOOKEEPER_SESSION_TIMEOUT_MS: 120000
 
   redis:
     name: registry.ddbuild.io/images/mirror/datadog/dd-trace-ci:php-redis-5.0
--- a/.gitlab/generate-tracer.php
+++ b/.gitlab/generate-tracer.php
@@ -148,20 +148,14 @@ stages:
     # Set test environment variables
     docker exec ${CONTAINER_NAME} powershell.exe "setx DD_AUTOLOAD_NO_COMPILE true; setx DATADOG_HAVE_DEV_ENV 1; setx DD_TRACE_GIT_METADATA_ENABLED 0"
 
-    # Run extension tests (use -q for quiet mode to reduce log output)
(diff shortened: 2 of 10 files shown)


.gitlab/generate-common.php
```diff
@@ -50,6 +50,9 @@ function unset_dd_runner_env_vars() {
       - runner_system_failure
       - scheduler_failure
       - api_failure
+      - script_failure
```
realFlowControl, line 53: Could this retry any script error and potentially mask real bugs that the tests uncover?

.gitlab/generate-tracer.php
```diff
@@ -496,7 +521,26 @@ function before_script_steps() {
...
+            test-agent) service_type="test-agent" ;;
+            mysql-integration) service_type="mysql" ;;
+            elasticsearch*) service_type="elasticsearch" ;;
+            kafka*) service_type="kafka" ;;
+            redis*) service_type="redis" ;;
+            httpbin*) service_type="httpbin" ;;
+          esac
+
+          .gitlab/wait-for-service-ready.sh "$host" "$port" "$service_type" 30 5
+        done
```
realFlowControl: I am not too deep into the rest of this script, but this is duplicated from above (lines 215 to 234). Would it makes sense to move this into the `.gitlab/wait-for-service-ready.sh`? This way we'd get rid of the duplication and also don't need to keep it in sync in the future.

.gitlab/generate-tracer.php
```diff
@@ -513,7 +557,7 @@ function before_script_steps() {
 $services["deferred_loading"] = "mysql";
 $services["deferred_loadin"] = "redis";
 $services["pdo"] = "mysql";
-$services["kafk"] = ["kafka", "zookeeper"];
+$services["kafka"] = ["kafka", "zookeeper"];  // Overwrite auto-generated entry
```
realFlowControl, line 522: Nice catch, seems like this should have had been a problem in the past, but it wasn't? 🤯

.gitlab/wait-for-service-ready.sh
```diff
@@ -0,0 +1,73 @@
...
+set -euo pipefail
+
+HOST=$1
+PORT=$2
+SERVICE_TYPE=${3:-generic}
+MAX_ATTEMPTS=${4:-30}
+RETRY_DELAY=${5:-5}
+
+echo "Waiting for $HOST:$PORT to be reachable..."
+wait-for "$HOST:$PORT" --timeout=180
```
realFlowControl:
```suggestion
if ! wait-for "$HOST:$PORT" --timeout=180; then
    echo "ERROR: Could not reach $HOST:$PORT" >&2
    exit 1
fi
```
Should we early exit here?

.gitlab/generate-common.php
Leiyks: From what I have seen so far, the tests are failing because of an external reason (a service like kafka is somehow not reachable for 1 particular test, or the language tests failing because of an public endpoint not being reachable, ect...)

Not really an error on our side most of the time, and when the error is then the tests are failing more than 2 times in a row

.gitlab/generate-tracer.php
Leiyks: I think it was causing multiple instance of kafka to run in parallel and because they have the same the same port sometimes the tests were waiting for one instance to boot properly but failed because the port were already taken by the other instance. 

Also the 2 instance put more load on the runner as well, but it is not failing the tests most of the time :D

.gitlab/generate-tracer.php
Leiyks: Yes indeed, let me change that

.gitlab/wait-for-service-ready.sh
Leiyks: Yes indeed :ok_hand:

PR #3499, opened 2025-11-24
fix(tracer): add null guard clause in sidecar reconnect callback

Leiyks:
### Description

Fixes a use-after-free crash in sidecar shutdown that occurs when the reconnect callback is invoked after endpoints have been freed.

image

Related Jiras: [APMS-17794]

[APMS-17794]: https://datadoghq.atlassian.net/browse/APMS-17794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -71,18 +71,21 @@ workflow:
   - when: manual
     interruptible: true
 
+.on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible: &on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible
+  if: '($CI_COMMIT_REF_NAME == "master" && ($NIGHTLY_BENCHMARKS || $CI_PIPELINE_SOURCE != "schedule")) || $CI_COMMIT_REF_NAME =~ /^ddtrace-/'
+  when: always
+  interruptible: false
+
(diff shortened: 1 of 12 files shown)


ext/sidecar.c
```diff
@@ -115,6 +116,10 @@ static void dd_sidecar_post_connect(ddog_SidecarTransport **transport, bool is_f
 }
 
 static void dd_sidecar_on_reconnect(ddog_SidecarTransport *transport) {
+    if (!ddtrace_endpoint || !dogstatsd_endpoint) {
+        return;
+    }
+
```
realFlowControl, line 122: The `dd_sidecar_connection_factory_ex` function uses another pattern:
https://github.com/DataDog/dd-trace-php/blob/6122feb8e17f47680f6269bb30af35b20936f251/ext/sidecar.c#L161-L165

I am not 100% certain why a `ZEND_ASSERT` is sufficient below, but most likely because `ddtrace_endpoint` and `dogstatsd_endpoint` should be kept in sync anyway.

So maybe this is just fine ..

Leiyks: Yeah this is the original reason this was put there i believe

PR #3505, opened 2025-11-28
perf(prof): speedup hot path in allocator

realFlowControl:
### Description

We got some crash reports with stacks like this:
```
#14  0x00006101e274a1e3 execute_ex 
#15  0x00006101e26ca4f6 zend_call_function 
#16  0x00006101e26ca8ad zend_call_known_function 
#17  0x00006101e26c9647 zend_lookup_class_ex 
#18  0x00006101e26cac0e zend_fetch_class 
```

None of these are actually helpful, because this would just look like a crash in the engine's `execute_ex` function somewhere. But now we got a new crash trace adding some stack frames that look like stack corruption, but reveal something interesting nonetheless:
```
#0   0x00006101e26a8712 zend_mm_gc 
#1   0x00007b434a0a82e1 get_or_init<core::cell::RefCell<datadog_php_profiling::RequestLocals>, (), fn() -> core::cell::RefCell<datadog_php_profiling::RequestLocals>> (native/lazy.rs:53:15)
#2   0x00007b434a0a82e1 {closure#0} (native/mod.rs:94:25)
#3   0x00007b434a0a82e1 call_once<datadog_php_profiling::REQUEST_LOCALS::{constant#0}::{closure_env#0}, (core::option::Option<&mut core::option::Option<core::cell::RefCell<datadog_php_profiling::RequestLocals>>>)> (src/ops/function.rs:250:5)
#4   0x00007b434a0a82e1 try_with<core::cell::RefCell<datadog_php_profiling::RequestLocals>, datadog_php_profiling::{impl#2}::try_with_borrow::{closure_env#0}<datadog_php_profiling::RequestLocals, datadog_php_profiling::allocation::alloc_prof_rshutdown::{closure_env#0}, bool>, core::result::Result<bool, core::cell::BorrowError>> (src/thread/local.rs:282:37)
#5   0x00007b434a0a82e1 try_with_borrow<datadog_php_profiling::RequestLocals, datadog_php_profiling::allocation::alloc_prof_rshutdown::{closure_env#0}, bool> (profiling/src/lib.rs:498:12)
#6   0x00007b434a0a82e1 alloc_prof_rshutdown (allocation/mod.rs:152:30)
#7   0x00007b434a0a81be get_or_init<core::cell::RefCell<datadog_php_profiling::allocation::AllocationProfilingStats>, (), fn() -> core::cell::RefCell<datadog_php_profiling::allocation::AllocationProfilingStats>> (native/lazy.rs:56:40)
#8   0x00007b434a0a81be {closure#0} (native/mod.rs:94:25)
#9   0x00007b434a0a81be call_once<datadog_php_profiling::allocation::ALLOCATION_PROFILING_STATS::{constant#0}::{closure_env#0}, (core::option::Option<&mut core::option::Option<core::cell::RefCell<datadog_php_profiling::allocation::AllocationProfilingStats>>>)> (src/ops/function.rs:250:5)
#10  0x00007b434a0a81be try_with<core::cell::RefCell<datadog_php_profiling::allocation::AllocationProfilingStats>, datadog_php_profiling::{impl#2}::try_with_borrow_mut::{closure_env#0}<datadog_php_profiling::allocation::AllocationProfilingStats, datadog_php_profiling::allocation::allocation_le83::alloc_prof_realloc::{closure_env#0}, bool>, core::result::Result<bool, core::cell::BorrowMutError>> (src/thread/local.rs:282:37)
#11  0x00007b434a0a81be try_with_borrow_mut<datadog_php_profiling::allocation::AllocationProfilingStats, datadog_php_profiling::allocation::allocation_le83::alloc_prof_realloc::{closure_env#0}, bool> (profiling/src/lib.rs:507:12)
#12  0x00007b434a0a81be borrow_mut_or_false<std::thread::local::LocalKey<core::cell::RefCell<datadog_php_profiling::allocation::AllocationProfilingStats>>, datadog_php_profiling::allocation::AllocationProfilingStats, datadog_php_profiling::allocation::allocation_le83::alloc_prof_realloc::{closure_env#0}> (profiling/src/lib.rs:489:9)
#13  0x00007b434a0a81be alloc_prof_realloc (allocation/allocation_le83.rs:418:10)
#14  0x00006101e274a1e3 execute_ex 
#15  0x00006101e26ca4f6 zend_call_function 
#16  0x00006101e26ca8ad zend_call_known_function 
#17  0x00006101e26c9647 zend_lookup_class_ex 
#18  0x00006101e26cac0e zend_fetch_class 
```

~~There is just no way that `alloc_prof_realloc` would call into `alloc_prof_rshutdown` and from there, there is no way to call into `zend_mm_gc`...~~
The stack trace is still weird, but `alloc_prof_realloc` in case ZendMM can't allocate, will bailout (`longjmp()` to `EG(bailout)` which usually leads to `RSHUTDOWN` and as such to `alloc_prof_rshutdown`. I can't see a path though from `alloc_prof_rshutdown` to `zend_mm_gc`...

But there are a few ideas and things to this stack anyway: according to the runtime stack trace, it is in `Composer\Autoload\ClassLoader::findFileWithExtension()` [line 505](https://github.com/composer/composer/blob/1aed0a57ae5407446cda2e98a52484b4e8c1490b/src/Composer/Autoload/ClassLoader.php#L505) which could trigger a reallocation, which on the way down, could call into `zend_mm_gc`. So at least this makes sense, not 100% with the stack trace we see, but ...

So anyway, I was thinking:
- we might not need to do all of this `get_or_init()` dance with the `RefCell` in `ALLOCATION_PROFILING_STATS` at runtime, when we know there is no reentrancy
- we also do not need to do the runtime lazy init for the thread local `ALLOCATION_PROFILING_STATS` itself when we make it a `const`
- `SeqCst` memory ordering for stats and intervals is not needed, we only need atomicity, there is no happen-before relation to other memory, so we can use `Relaxed` and also move all the stats behind a feature flag because we are the only ones using it

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/allocation_ge84.rs
+++ b/profiling/src/allocation/allocation_ge84.rs
@@ -1,4 +1,6 @@
-use crate::allocation::{collect_allocation, ALLOCATION_PROFILING_STATS};
+use crate::allocation::{
+    allocation_profiling_stats_mut, allocation_profiling_stats_should_collect, collect_allocation,
+};
 use crate::bindings::{self as zend};
 use crate::{RefCellExt, PROFILER_NAME};
 use core::{cell::Cell, ptr};
@@ -366,14 +368,7 @@ unsafe extern "C" fn alloc_prof_malloc(len: size_t) -> *mut c_void {
         return ptr;
     }
 
-    let should_collect = ALLOCATION_PROFILING_STATS
-        .try_with(|stats| {
-            // Safety: ALLOCATION_PROFILING_STATS is thread-local and initialized in GINIT.
-            unsafe { (*(*stats.get()).as_mut_ptr()).should_collect_allocation(len) }
-        })
-        .unwrap_unchecked();
-
-    if should_collect {
+    if allocation_profiling_stats_should_collect(len) {
         collect_allocation(len);
     }
 
@@ -442,14 +437,7 @@ unsafe extern "C" fn alloc_prof_realloc(prev_ptr: *mut c_void, len: size_t) -> *
         return ptr;
     }
 
-    let should_collect = ALLOCATION_PROFILING_STATS
-        .try_with(|stats| {
-            // Safety: ALLOCATION_PROFILING_STATS is thread-local and initialized in GINIT.
-            unsafe { (*(*stats.get()).as_mut_ptr()).should_collect_allocation(len) }
-        })
-        .unwrap_unchecked();
-
-    if should_collect {
+    if allocation_profiling_stats_should_collect(len) {
         collect_allocation(len);
     }
 
--- a/profiling/src/allocation/mod.rs
+++ b/profiling/src/allocation/mod.rs
@@ -1,3 +1,12 @@
+mod tls_allocation_profiling_stats;
+
+#[cfg(php_zend_mm_set_custom_handlers_ex)]
+pub mod allocation_ge84;
+#[cfg(not(php_zend_mm_set_custom_handlers_ex))]
+pub mod allocation_le83;
+
+pub use tls_allocation_profiling_stats::*;
+
 use crate::bindings::{self as zend};
 use crate::profiling::Profiler;
 use crate::{RefCellExt, REQUEST_LOCALS};
@@ -5,15 +14,8 @@ use libc::size_t;
 use log::{debug, error, trace};
 use rand::rngs::ThreadRng;
 use rand_distr::{Distribution, Poisson};
-use std::cell::UnsafeCell;
 use std::ffi::c_void;
-use std::mem::MaybeUninit;
 use std::sync::atomic::{AtomicU64, Ordering};
-
-#[cfg(php_zend_mm_set_custom_handlers_ex)]
-pub mod allocation_ge84;
-#[cfg(not(php_zend_mm_set_custom_handlers_ex))]
-pub mod allocation_le83;
 
 /// Default sampling interval in bytes (4MB)
 pub const DEFAULT_ALLOCATION_SAMPLING_INTERVAL: u64 = 1024 * 4096;
@@ -84,37 +86,6 @@ pub fn collect_allocation(len: size_t) {
             )
         };
     }
-}
-
-thread_local! {
-    pub(crate) static ALLOCATION_PROFILING_STATS: UnsafeCell<MaybeUninit<AllocationProfilingStats>> =
-        const { UnsafeCell::new(MaybeUninit::uninit()) };
-}
-
-pub fn alloc_prof_ginit() {
-    unsafe {
-        // Eagerly initialize the allocation profiling stats before handling first request
-        ALLOCATION_PROFILING_STATS
-            .try_with(|slot| {
-                // SAFETY:
-                // - `thread_local!` guarantees `slot` is per-thread, and `UnsafeCell::get()`
-                //   gives us a mutable pointer to the inner `MaybeUninit`.
-                // - We must not call this twice on the same thread.
-                let slot_ptr: *mut MaybeUninit<AllocationProfilingStats> = slot.get();
-                (*slot_ptr).write(AllocationProfilingStats::new());
-            })
-            .unwrap_unchecked();
-    }
-
-    #[cfg(not(php_zend_mm_set_custom_handlers_ex))]
-    allocation_le83::alloc_prof_ginit();
-    #[cfg(php_zend_mm_set_custom_handlers_ex)]
-    allocation_ge84::alloc_prof_ginit();
-}
-
-pub fn alloc_prof_gshutdown() {
-    #[cfg(php_zend_mm_set_custom_handlers_ex)]
-    allocation_ge84::alloc_prof_gshutdown();
 }
 
 #[cfg(not(php_zend_mm_set_custom_handlers_ex))]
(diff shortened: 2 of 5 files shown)


morrisonlevi:
> There is just no way that alloc_prof_realloc would call into alloc_prof_rshutdown and from there, there is no way to call into zend_mm_gc...

I don't think this quite right. The function `zend_mm_safe_error` has `zend_bailout` which jumps to `EG(bailout)`.

> I was thinking: we might not need to do all of this get_or_init() dance and use a RefCell at runtime, when we can instead initialize the ALLOCATION_PROFILING_STATS in GINIT and then just "know" that this is initialized and use it.

RefCell is not about being initialized or not, that is the thread-local lazy initialization. The RefCell is about enforcing Rust's borrowing rules at runtime.

-----

I think there's possibly some optimization though, we could use `MaybeUninit` and initialize it in ginit, then nothing else needs to check for initialization, it can know it's initialized. It still needs to try to borrow though, I think.

morrisonlevi (commented):
> There is just no way that alloc_prof_realloc would call into alloc_prof_rshutdown and from there, there is no way to call into zend_mm_gc...

I don't think this quite right. The function `zend_mm_safe_error` is called by various allocation functions such as `zend_mm_realloc_huge`, and `zend_mm_safe_error` has `zend_bailout` which jumps to `EG(bailout)`. It could jump to shutdown, yeah?

> I was thinking: we might not need to do all of this get_or_init() dance and use a RefCell at runtime, when we can instead initialize the ALLOCATION_PROFILING_STATS in GINIT and then just "know" that this is initialized and use it.

RefCell is not about being initialized or not, that is the thread-local lazy initialization. The RefCell is about enforcing Rust's borrowing rules at runtime.

-----

I think there's possibly some optimization though. We could use `MaybeUninit<RefCell<AllocationProfilingStats>>` and make it `const { MaybeUninit::uninit() }` so it doesn't need lazy-initialization checks at runtime. We initialize it in ginit. The RefCell still is there to enforce borrow rules at runtime (cover our butts, yes, it causes crashes, but that's better than undefined behavior).

realFlowControl:
With b0ce8e9 I've guarded the allocation (and exception) stats we are sending with the metadata JSON behind a feature flag because it looks like we are spending nearly ⅓ of our time just in incrementing these counters:

image

They are also only needed by us as a measurement to validate the upscaling algorithm. I am gonna measure again once the binary is ready to benchmark.

profiling/Cargo.toml
```diff
@@ -62,6 +62,7 @@ harness = false
 
 [features]
 default = ["io_profiling"]
+debug_stats = []
```
morrisonlevi, line 65: Could we rename this to `debug_alloc_stats` or more specific?

profiling/src/allocation/allocation_ge84.rs
```diff
@@ -353,7 +355,9 @@ pub fn alloc_prof_rshutdown() {
 }
 
 unsafe extern "C" fn alloc_prof_malloc(len: size_t) -> *mut c_void {
+    #[cfg(feature = "debug_stats")]
     ALLOCATION_PROFILING_COUNT.fetch_add(1, SeqCst);
```
morrisonlevi: These can be Relaxed, I think there's no impact without happens-before relationships.

profiling/src/allocation/mod.rs
```diff
@@ -84,11 +87,24 @@ pub fn collect_allocation(len: size_t) {
 }
 
 thread_local! {
-    static ALLOCATION_PROFILING_STATS: RefCell<AllocationProfilingStats> =
-        RefCell::new(AllocationProfilingStats::new());
+    pub(crate) static ALLOCATION_PROFILING_STATS: MaybeUninit<UnsafeCell<AllocationProfilingStats>> =
```
morrisonlevi: I was wrong before, this should be an `UnsafeCell<MaybeUninit<T>>` and not `MaybeUninit<UnsafeCell<T>>`. The reason is your initialization code below is undefined behavior (I'll leave a comment there too), from [MaybeUninit::as_ptr](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.as_ptr):

> Writing to memory that this pointer (non-transitively) points to is undefined behavior (except inside an `UnsafeCell<T>`).

But we have the opposite. We have to invert.

profiling/src/allocation/mod.rs
```diff
@@ -84,11 +87,24 @@ pub fn collect_allocation(len: size_t) {
...
+    // Eagerly initialize the allocation profiling stats before handling first request
+    ALLOCATION_PROFILING_STATS.with(|slot| {
+        unsafe {
+            // SAFETY:
+            // - `thread_local!` guarantees `slot` is per-thread, so this cast is fine
+            //   as long as we respect “init exactly once per thread”.
+            // - We must not call this twice on the same thread.
+            let slot_mut: *mut MaybeUninit<UnsafeCell<AllocationProfilingStats>> =
+                slot as *const _ as *mut _;
+            (*slot_mut).write(UnsafeCell::new(AllocationProfilingStats::new()));
```
morrisonlevi: This write is undefined behavior, see my review.

profiling/Cargo.toml
morrisonlevi: Nevermind, it includes exception stats as well, all good.

morrisonlevi:
I had a thought: when an Apache graceful restart is done, the process goes through mshutdown and back into minit. What is the story there with ginit and gshutdown?

PR #3507, opened 2025-11-28
appsec: fix duration_ext metric

cataphract:
It had three problems:
* Value in ns instead of microseconds
* It would effectively be the time of the first rasp call only
* Did not fetch the span properly where to store the tags

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -71,18 +71,21 @@ workflow:
   - when: manual
     interruptible: true
 
+.on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible: &on-master-or-nightly-benchmarks-or-release-always-and-not-interruptible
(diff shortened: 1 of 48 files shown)


appsec/src/extension/rasp.c
```diff
@@ -0,0 +1,66 @@
...
+    zval zv;
+    ZVAL_DOUBLE(&zv, duration);
+    zval *nullable res =
+        zend_hash_add(Z_ARRVAL_P(metrics_zv), _dd_rasp_duration_ext, &zv);
+    if (res == NULL) {
+        mlog(dd_log_warning, "Failed to add metric %.*s",
+            (int)ZSTR_LEN(_dd_rasp_duration_ext),
+            ZSTR_VAL(_dd_rasp_duration_ext));
+    }
+}
```
estringana, line 66: I would keep this part on tags.c since everything related to metrics is there. Just an opinion

PR #3510, opened 2025-11-28
test: add missing CI image to mirror

Leiyks:
### Description

This PR modify our CI to use our private registry instead of dockerhub to pull the public images we are using.
The goal is to stop having image pull rate limits errors when we are running too many pipeline.

### DIFF
--- a/dockerfiles/frameworks/flow.yml
+++ b/dockerfiles/frameworks/flow.yml
@@ -14,7 +14,7 @@ services:
     cap_add:
       - SYS_PTRACE
   mysql:
-    image: mysql:5.7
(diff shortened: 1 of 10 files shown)


Leiyks:
> Have we considered authenticating against Docker Hub to bypass (or widen) pull limits?

Indeed, this could be a much better solution under the condition that we have a Pro/Team/Business service account that we can use to do the authentication.
Currently to upload docker images we are launching a manual pipeline with our account name and access token from dockerhub.
After discussing with Pawel we figured that just using our private registry would be fine, but I agree that it would cause an issue from a maintenance perspective.

PR #3514, opened 2025-12-02
feat(Symfony): Check cache.app is reachable before use 

Leiyks:
### Description

This PR add a readyness check on the Symfony cache app in case the service is unavailable.
The absence of this check can be the cause for significant latency increase when the cache app is not available as we are fetching all the routes in that case.

### DIFF
--- a/src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
+++ b/src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
@@ -439,6 +439,57 @@ class SymfonyIntegration extends Integration
                 if (!\method_exists($cache, 'getItem')) {
                     return;
                 }
+
+                // Check if the cache adapter is actually accessible before trying to use it
+                if (\method_exists($cache, 'getAdapter')) {
+                    try {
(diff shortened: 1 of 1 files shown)


src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
```diff
@@ -439,6 +439,57 @@ static function() {
...
+                    try {
+                        $adapter = $cache->getAdapter();
+
+                        if (
+                            $adapter instanceof \Symfony\Component\Cache\Adapter\RedisAdapter ||
+                            $adapter instanceof \Symfony\Component\Cache\Adapter\RedisTagAwareAdapter
+                        ) {
+                            $reflection = new \ReflectionClass($adapter);
+                            $redisProperty = $reflection->getProperty('redis');
+                            $redisProperty->setAccessible(true);
```
bwoebi, line 454: Needs a `PHP_VERSION_ID < 80000` check to avoid deprecations

tabgok:
approved from IDM's standpoint

PR #3515, opened 2025-12-03
feat(prof): detect parallel threads

realFlowControl:
### Description

This PR adds code to detect and label threads that have been created [by `ext-parallel`](https://github.com/krakjoe/parallel). It is actually 80% comments and just a few lines of code 😉

So in the end it turns out we can access the thread locals variables from other extensions. We do not need anything from that, just check if it is `NULL`. In case it is, it is not a `ext-parallel` thread, otherwise it is one 🎉 

Technically I am checking for the `php_parallel_scheduler_context` TLS being non-NULL, it is set in the [`php_parallel_scheduler_setup()` function](https://github.com/krakjoe/parallel/blob/3955f21f5a6b74692f305e2ca9401a0ffb042ade/src/scheduler.c#L81-L82) right from the start of the thread main loop and before calling `ts_resource(0);` which triggers `GINIT`.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/php_ffi.c
+++ b/profiling/src/php_ffi.c
@@ -631,14 +631,8 @@ __attribute__((weak)) zend_write_func_t zend_write;
  * This function is meant to be called in the GINIT phase of a PHP request, it is also safe to be
  * called in RINIT or during request processing, but its outcome won't change anymore after GINIT.
  * That being said, it being a costly function (module registry lookup, `dlsym()`,
- * `__tls_get_addr()`), the best usage pattern is to call this in GINIT and cache the result in a
- * thread local.
- *
- * Why not just `__attribute__((weak)) php_parallel_scheduler_context;`?
- * This would work if we could enforce the order of `dlopen()` calls for extensions (which we
- * can't). If the parallel extension is loaded before the profiler extension it works just nice
- * but if the profiler gets loaded first this will not resolve correct. Luckily `dlsym()` behaves
- * as a safe wrapper around this.
+ * `__tls_get_addr()` in the fallback branch), the best usage pattern is to call this in GINIT and
+ * cache the result in a thread local.
  */
 bool ddog_php_prof_is_parallel_thread() {
     // Check if parallel extension is loaded to retrieve it's dl handle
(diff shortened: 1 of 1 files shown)


morrisonlevi (commented):
Before I dive into the code, which I've already seen a bit because of R&D week: since you are a maintainer of ext/parallel, should we instead release a new version of it which does something to make this simpler and easier to know? This PR is evidence there is value here, so rather than committing to platform specific code, should we "just" release a new version of ext/parallel?

profiling/src/php_ffi.c
```diff
@@ -624,3 +624,75 @@ zval *ddog_php_prof_get_memoized_config(uint16_t config_id) {
...
+        unsigned long offset;
+    } tls_descriptor;
+
+    tls_descriptor *desc = (tls_descriptor*)tls_symbol;
+
+    if (desc->thunk == NULL) {
+        return false;
+    }
+
+    tls_ptr = (void**)desc->thunk(desc);
```
morrisonlevi, line 682: Help me check my understanding. The link has this comment, `really void* (*ThunkFunc)(Thunk*)` for the `thunk`. It returns the address of the thread-local variable. But what lies at that address is another pointer, so that's why you use the cast `(void**)desc->thunk(desc)`. Correct? More accurately this would be `php_parallel_runtime_t**`, except we don't have the definition for `php_parallel_runtime_t`, so hence `void**`, right?

realFlowControl: Exactly

realFlowControl:
> [...] since you are a maintainer of ext/parallel, should we instead release a new version of it which does something to make this simpler and easier to know? [...]

I thought of this as well but: this won't allow us to detect "old" versions of parallel that would not export that new symbol (so we'd be left with either this code or being blind for those old versions). 

> [...] so rather than committing to platform specific code [...]

The code is fairly simple IMHO, it boils down to a `dlsym()`, a cast, a pointer dereference and a `NULL` compare, it won't get much simpler and honestly the MacOS code path only exists because we are using MacOS for local development.

I get your point though and it would be "nice" to have something explicit, so I've added 78caed3 and opened https://github.com/krakjoe/parallel/pull/352

PR #3520, opened 2025-12-08
fix: Protect sidecar globals access with mutex

realFlowControl:
### Description

The `last_service_name` and `last_env_name` globals could result in a use-after-free when the calling site to `ddtrace_get_telemetry_rc_info()` does not finish usage before those get cycled in `ddtrace_sidecar_submit_root_span_data_direct()`:
https://github.com/DataDog/dd-trace-php/blob/1dff083b79a234a66a76637069c1c2a9b63d2525/ext/sidecar.c#L508-L517

Thanks @yuandesu for reaching out regarding this race condition.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/commands/client_init.c
+++ b/appsec/src/extension/commands/client_init.c
@@ -121,6 +121,13 @@ static dd_result _pack_command(
 
     mpack_finish_map(w); // telemetry settings
 
+    if (tel_rc_info.service_name) {
+        zend_string_release(tel_rc_info.service_name);
+    }
+    if (tel_rc_info.env_name) {
+        zend_string_release(tel_rc_info.env_name);
+    }
+
     // Sidecar settings
     mpack_start_map(w, 2);
     {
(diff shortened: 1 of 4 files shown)


appsec/src/extension/commands/client_init.c
```diff
@@ -121,6 +121,13 @@ static dd_result _pack_command(
 
     mpack_finish_map(w); // telemetry settings
 
+    if (tel_rc_info.service_name) {
+        zend_string_release(tel_rc_info.service_name);
+    }
+    if (tel_rc_info.env_name) {
+        zend_string_release(tel_rc_info.env_name);
+    }
```
bantonsson, line 129: Don't know enough about what's in the struct or what a zend string really is, but can't this repeated code be in a helper function?

estringana: I agree on this. What about encapsulating the free of the type `telemetry_rc_info` within a function? Maybe `telemetry_rc_info_free` or similar. That way it shows that 2 of the 3 pointers in that struct are owned

cataphract:
I don't believe that this change is necessary (see slack thread). The worker thread only reads from these strings, and the writing thread (the PHP worker thread) is the same that appsec runs on when fetching/reading these strings. You can reopen if you think I'm mistaken.

PR #3536, opened 2025-12-17
feat(CI): add `docker login` to before_script in all CI jobs

Leiyks:
### Description

This PR add a dockerhub login step on jobs pulling docker images to avoid image pull rate limits errors from dockerhub on non-authenticated requests.


realFlowControl (commented):
Do we know if `jq` and/or `vault` are not installed? In case they are, we should remove the code to check and install them and rely on them being around in the GBI.

PR #3548, opened 2025-12-24
Collect framework endpoints

estringana:
### Description

Implements endpoint discovery on the PHP tracer side. This PR extracts route information from Laravel, Symfony, and WordPress and pushes it to the Rust sidecar for downstream use by API Security.

The implementation hooks into the existing framework route integrations (the same instrumentation points used for `http.route` extraction more or less) and collects the full endpoint inventory. This data is pushed to the sidecar via the shared IPC channel on each request, where it's deduplicated and aggregated across the process lifetime.

The PHP side of this is relatively straightforward, the routes are already being extracted, this just adds a second consumer of that data. The more complex part is the Rust sidecar implementation in [libdatadog#1182](https://github.com/DataDog/libdatadog/pull/1182), which handles aggregation, deduplication, and forwarding to the backend. The two PRs together close the full data flow: PHP extension -> IPC -> Rust sidecar -> Backend.

Endpoint discovery is what allows API Security to build a map of all exposed endpoints in a customer's application without requiring any manual configuration.

### Reviewer checklist
- [ ] Test coverage seems ok.


components-rs/common.h
```diff
@@ -264,15 +264,19 @@ typedef struct _zend_string _zend_string;
...
-typedef enum ddog_ConfigurationOrigin {
-  DDOG_CONFIGURATION_ORIGIN_ENV_VAR,
-  DDOG_CONFIGURATION_ORIGIN_CODE,
-  DDOG_CONFIGURATION_ORIGIN_DD_CONFIG,
-  DDOG_CONFIGURATION_ORIGIN_REMOTE_CONFIG,
-  DDOG_CONFIGURATION_ORIGIN_DEFAULT,
-  DDOG_CONFIGURATION_ORIGIN_LOCAL_STABLE_CONFIG,
-  DDOG_CONFIGURATION_ORIGIN_FLEET_STABLE_CONFIG,
-} ddog_ConfigurationOrigin;
+typedef enum ddog_Log {
```
cataphract, line 267: This diff moves a lot of code around, making it difficult to determine what changed. Commit only the relevant parts.

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laravel8xTests.groovy
```diff
@@ -103,4 +104,28 @@ class Laravel8xTests {
         assert span.meta."_dd.appsec.event_rules.version" != ''
         assert span.meta."appsec.blocked" == "true"
     }
+
+    @Test
+    void 'Endpoints are sended'() {
```
cataphract: how does work? why is this test able to get the endpoints? I think test should be the first test to run

also, "sent"

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Symfony62Tests.groovy
```diff
@@ -134,4 +135,28 @@ class Symfony62Tests {
             assert res.exitCode == 0
         }
     }
+
+    @Test
+    void 'Endpoints are sended'() {
```
cataphract: same considerations

ext/ddtrace.c
```diff
@@ -2796,6 +2796,91 @@ PHP_FUNCTION(DDTrace_dogstatsd_set) {
...
+
+    RETURN_BOOL(ddog_sidecar_telemetry_are_endpoints_collected(ddtrace_telemetry_cache(), service_name, env_name));
+}
+
+zend_string *get_env() {
+    if (get_DD_ENV() == NULL) {
+        return zend_string_init(ZEND_STRL("env"), 0);
+    }
+    return get_DD_ENV();
+}
```
cataphract: Is this used?

ext/ddtrace.c
```diff
@@ -2796,6 +2796,91 @@ PHP_FUNCTION(DDTrace_dogstatsd_set) {
...
+
+ddog_Method dd_string_to_method(zend_string *method) {
```
cataphract: could be static

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laravel8xTests.groovy
estringana: It does a call to `/` so it generates the endpoints.
I have fixed the typo

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Symfony62Tests.groovy
estringana: Same as above

components-rs/common.h
estringana: I've just generated this file again basing it on the master one with `make cbindgen` and it generated the same output

cataphract (approved):

ext/compat_getrandom.c
```diff
@@ -0,0 +1,31 @@
+// Minimal getrandom(2) compatibility for older libcs without the symbol
+#include <sys/types.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <errno.h>
+
+// Provide a weak, local fallback so that:
+// - On older libcs without getrandom, we satisfy the symbol at runtime.
+// - On newer libcs, libc's strong symbol may be preferred by the runtime loader.
+__attribute__((weak)) ssize_t getrandom(void *buf, size_t buflen, unsigned int flags) {
```
estringana, line 10: @bwoebi would you mind to have a look at this? I had to add it because the version of libdatadog used by this brach was calling getrandom and in some old alpine version it was not present

cataphract: Looks good to me. It's basically what we have for nginx-datadog. You could probably call `syscall`, but this also supports Linux < 3.17

PR #3551, opened 2025-12-28
Use php_log_err() rather than _php_error_log_ex()

Girgias:
### Description

I changed the signature of `_php_error_log()` and removed `_php_error_log_ex()` on php-src@master and thinking of fully removing the exposed API as I can only find one other usage of it outside of dd-trace which is also always calling `_php_error_log()` with the same mode.

So rather than doing effectively a function indirection call `php_log_err()` directly (this is a compat macro on master, but my understanding is you support PHP 7.0 still)

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/loader/dd_library_loader.c
+++ b/loader/dd_library_loader.c
@@ -301,7 +301,7 @@ void ddloader_logv(injected_ext *config, log_level level, const char *format, va
 
     char full[512];
     snprintf(full, sizeof(full), "[dd_library_loader][%s] %s", level_str, msg);
-    _php_error_log(0, full, NULL, NULL);
+    php_log_err(full);
 }
 
 void ddloader_logf(injected_ext *config, log_level level, const char *format, ...) {


realFlowControl:
Thanks @Girgias ❤️ 
I've triggered our [internal GitLab CI job](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-php/-/pipelines/88446157) on this PR and will merge once that's green.

bwoebi:
Hey @Girgias,
This will break on PHP 5 and PHP 7.0. On PHP 7.1+ this is a macro which decays to php_log_err_with_severity, a symbol which did not exist before PHP 7.1.

The problem is that this is a version agnostic zend_extension which we build once - and reuse across all versions (as you can see, it has version checks - and then loads the appropriate ddtrace.so).
(Also, we don't _support_ PHP 5 for ddtrace, but this particular extension is injected regardless of the PHP version, so it just does nothing, but that still requires all symbols to exist.)

If this symbol gets removed on master, the proper fix is using dlsym stuff with disambiguation depending on version number.

PR #3559, opened 2026-01-07
feat(prof): improve time sample accuracy by gathering during allocation samples

morrisonlevi:
### Description

Gathers time samples during allocation samples if `interrupt_count > 0`. This has two benefits:
 1. A slight accuracy improvement.
 2. A slight performance improvement in latency and CPU.

The accuracy benefit comes from the fact that when an interrupt is set, on master it won't be handled until the next VM interrupt. If we're gathering a memory sample and that count is non-zero, that means that we're gathering it sooner than it would have otherwise been handled, so less VM code can change. Notably, this should help with frameless functions (added in PHP 8.4) which allocate.

The performance benefit mainly* comes from combining the two stack walks into one. It is dampened a bit by incurring an extra atomic load, we might revisit the `Ordering` used in the future. When I say a slight performance benefit, here's a comparison to v1.16.0:

| library_version | latency_p99 | cpu% | memory |
|-----------------|-------------|------|---------|
| local | 114.38 ms | 127.8%  | 121.95 MB   |
| 1.16.0 |          114.35 ms |    128.5% |  120.19 MB  |

\* I was going to look at the profiles and post screenshots but for whatever reason I cannot see any frames related to memory allocations in the full host profiler. I may edit this later to show flamegraphs when I can get them.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/tests/correctness/allocation_time_combined.php
+++ b/profiling/tests/correctness/allocation_time_combined.php
@@ -18,7 +18,6 @@ function main() {
         $xs = str_repeat("x", 10_000_000); // 10MB source
         $ys = str_replace("x", "y", $xs); // 10MB allocation in frameless function
         $zs = str_replace("y", "z", $ys); // 10MB allocation in frameless function
-        strlen($zs); // Use the result to prevent optimization
     }
 }
 


profiling/src/allocation/mod.rs
```diff
@@ -74,12 +74,21 @@ impl AllocationProfilingStats {
 #[cold]
 pub fn collect_allocation(len: size_t) {
     if let Some(profiler) = Profiler::get() {
-        // Safety: execute_data was provided by the engine, and the profiler doesn't mutate it.
+        // Check if there's a pending time interrupt that we can handle now
+        // instead of waiting for an interrupt handler. This is slightly more
+        // accurate and efficient, win-win.
+        let interrupt_count = REQUEST_LOCALS
+            .try_with_borrow(|locals| locals.interrupt_count.swap(0, Ordering::SeqCst))
```
morrisonlevi, line 99: Note that `SeqCst` is probably too heavy-handed. Fixing this should be done in another PR as it affects everything which checks this, not just this combined sample.

profiling/src/profiling/uploader.rs
morrisonlevi, line 1: These changes fix a warning you can get on local development about an unused `use super::*;`.

profiling/tests/correctness/allocation_time_combined.php
```diff
@@ -0,0 +1,37 @@
...
+// allocation samples when interrupt_count > 0).
+//
+// Note: Piggybacking is opportunistic and happens when an allocation sample
+// occurs while a time interrupt is pending. PHP checks interrupts frequently
+// (at loop boundaries, function calls, etc.), so the window is narrow. This
+// test just verifies the mechanism works, not that it happens often.
+
+function allocate_large() {
+    // Large allocation to trigger sampling
+    $data = str_repeat("a", 1024 * 200_000); // ~200MB
```
realFlowControl: `str_repeat()` is not a frameless function, `substr()` or `str_replace()` are frameless function for example and do allocate memory.

morrisonlevi: Good idea!

profiling/tests/correctness/allocation_time_combined.php
```diff
@@ -0,0 +1,21 @@
...
+// increasing the likelihood of capturing piggybacked samples.
+
+function main() {
+    $duration = $_ENV["EXECUTION_TIME"] ?? 10;
+    $end = microtime(true) + $duration;
+
+    while (microtime(true) < $end) {
+        // Allocate large strings to trigger allocation sampling
+        $s = str_repeat("x", 10_000_000);
+        strlen($s); // Use the string to prevent optimization
```
realFlowControl: This still uses `str_repeat()` which is not a frameless function, `strlen()` is, but it does not allocate and as such will not trigger a combined sample.
Can we add a text, maybe some Lorem Ipsum (or more nerdy https://future-lives.com/wp-content/uploads/2014/09/TheSentinel.pdf) and replace random words in that text? I'd think using an array as the needle to replace should take time and allocate enough to trigger sampling. Setting `DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE=1` should also help taking more allocation samples and increase the probability of triggering a combined sample.

profiling/tests/correctness/allocation_time_combined.json
realFlowControl, line 1: This entire JSON has nothing to do with how the JSON's look that the https://github.com/DataDog/prof-correctness action accepts. As such prof-correctness seems to not match anything against the pprof which means it does not fail and as such [succeeds](https://github.com/DataDog/dd-trace-php/actions/runs/20798582784/job/59738269559#step:16:16) (which is something we should fix in the upstream action, it should fail if the JSON is in an invalid structure).

Have a look at the other examples, like https://github.com/DataDog/dd-trace-php/blob/4a3a98716843640b4a8d6894da8ef0072500c729/profiling/tests/correctness/timeline.json how the JSON is supposed to look.

profiling/src/profiling/mod.rs
```diff
@@ -1425,6 +1425,22 @@ impl Profiler {
         }
     }
 
+    /// Gets a timestamp for timeline profiling if timeline is enabled.
+    /// Returns NO_TIMESTAMP if timeline is disabled or if getting the time fails.
+    fn get_timeline_timestamp(&self) -> i64 {
+        let system_settings = self.system_settings.load(Ordering::SeqCst);
```
realFlowControl: This just got extracted to a function (which is a nice deduplication, thanks) , but do we need `Ordering::SeqCst`?
I am not against, this comment is not blocking this PR in any way and I did not think too much about it, it is just a bias since the findings in https://github.com/DataDog/dd-trace-php/pull/3505#issuecomment-3612549827, so nothing we need to handle in this PR, just thinking loud.

profiling/src/allocation/mod.rs
realFlowControl: Totally fine with postponing this to another PR, especially that benchmarks show we are slightly better anyway 🎉

.github/workflows/prof_correctness.yml
```diff
@@ -103,6 +103,15 @@ jobs:
           export DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE=1
           php -d extension=$PWD/target/profiler-release/libdatadog_php_profiling.so profiling/tests/correctness/allocations.php
           unset DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE
+          mkdir -p profiling/tests/correctness/allocation_time_combined/
+          export DD_PROFILING_OUTPUT_PPROF=$PWD/profiling/tests/correctness/allocation_time_combined/test.pprof
+          export DD_PROFILING_ALLOCATION_ENABLED=yes
```
realFlowControl, line 124:
```suggestion
          export DD_PROFILING_ALLOCATION_ENABLED=yes
          export DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE=1
```
Just an idea, but by setting the allocation sampling distance to 1, we make sure to take enough samples and have a higher chance of seeing combined samples.

.github/workflows/prof_correctness.yml
```diff
@@ -103,6 +103,15 @@ jobs:
...
+          export DD_PROFILING_WALL_TIME_ENABLED=yes
+          export DD_PROFILING_EXPERIMENTAL_CPU_TIME_ENABLED=no
+          export DD_PROFILING_ENDPOINT_COLLECTION_ENABLED=no
+          export EXECUTION_TIME=5
+          php -d extension=$PWD/target/profiler-release/libdatadog_php_profiling.so profiling/tests/correctness/allocation_time_combined.php
+          unset DD_PROFILING_ALLOCATION_ENABLED DD_PROFILING_WALL_TIME_ENABLED DD_PROFILING_EXPERIMENTAL_CPU_TIME_ENABLED DD_PROFILING_ENDPOINT_COLLECTION_ENABLED EXECUTION_TIME
```
realFlowControl:
```suggestion
          unset DD_PROFILING_ALLOCATION_ENABLED DD_PROFILING_WALL_TIME_ENABLED DD_PROFILING_EXPERIMENTAL_CPU_TIME_ENABLED DD_PROFILING_ENDPOINT_COLLECTION_ENABLED EXECUTION_TIME DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE
```
Just if you accept the change above

realFlowControl (changes requested):
I've created [an upstream PR](https://github.com/DataDog/prof-correctness/pull/83) to fix the empty/malformed expected profile JSON problem that this PR showed. Can you fix the tests, especially the `allocation_time_combined.json` file?

profiling/src/profiling/mod.rs
morrisonlevi: As mentioned in another comment, we should re-examine them in another PR but this was previously using `SeqCst` so I preserved this.

profiling/tests/correctness/allocation_time_combined.json
morrisonlevi: I've sort of fixed it. How can I make this only run for PHP 8.4+? Because the % will be determined by whether it's frameless or not.

profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -0,0 +1,51 @@
...
+      "stack-content": [
+        {
+          "regular_expression": "<?php;main$",
+          "percent": 80,
+          "error_margin": 30
+        },
+        {
+          "regular_expression": "<?php;main;standard\\|str_replace$",
+          "percent": 1.0,
+          "error_margin": 0.9
```
realFlowControl: An error margin of `0.9` is interestingly specific ... and from the prof correctness test ...
```
    pprof_analysis.go:667: Error opening file /github/workspace/profiling/tests/correctness/allocation_time_combined.json: JSON schema validation failed for /github/workspace/profiling/tests/correctness/allocation_time_combined.json:
          - stacks.2.stack-content.1.error_margin: Invalid type. Expected: integer, given: number
```

... not allowed, this needs to be an `integer`, not a `float`.

profiling/tests/correctness/allocation_time_combined.json
realFlowControl: > How can I make this only run for PHP 8.4+?

Not in this file, but in the workflow run itself (the `.github/workflows/prof_correctness.yml`), have a look how it is done for ZTS tests. You should be able to apply the same pattern for PHP 8.4+

But this brings me to a wider topic (one we should probably move to another discussion): `str_repeat()` as a frameless function should have less CPU time on PHP 8.4+ because of it being a frameless function, so depending on how big this effect is, we could "burry" this (but document it) in the `error_margin` or create two JSON files, one to match for PHP < 8.4 and one >= 8.4 with different `percent` values.

Also: we should lower the `error_margin` from `30`.

profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -0,0 +1,51 @@
...
+      ]
+    },
+    {
+      "profile-type": "wall-time",
+      "stack-content": [
+        {
+          "regular_expression": "<?php;main$",
+          "percent": 80,
+          "error_margin": 30
+        },
```
realFlowControl: I guess we are not interested in the `main` function itself.
```suggestion
```
I think we could add the `str_repeat()` function though.

profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -0,0 +1,51 @@
...
+        {
+          "regular_expression": "<?php;main;standard\\|str_replace$",
+          "percent": 40,
+          "error_margin": 30
+        },
+        {
+          "regular_expression": "<?php;main;standard\\|str_repeat$",
+          "percent": 40,
+          "error_margin": 30
+        }
```
realFlowControl, line 17: How did you choose those `percent` values?
Just asking, because I guess that `str_replace()` and `str_repeat()` are the only (let's say at least 99%) functions that do allocations in that test script, so shouldn't it be safe to do:

```suggestion
        {
          "regular_expression": "<?php;main;standard\\|str_replace$",
          "percent": 50,
          "error_margin": 1
        },
        {
          "regular_expression": "<?php;main;standard\\|str_repeat$",
          "percent": 50,
          "error_margin": 1
        }
```

The same might be true for `alloc-samples` bellow.

profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -0,0 +1,51 @@
...
+      "profile-type": "wall-time",
+      "stack-content": [
+        {
+          "regular_expression": "<?php;main$",
+          "percent": 80,
+          "error_margin": 30
+        },
+        {
+          "regular_expression": "<?php;main;standard\\|str_replace$",
+          "percent": 1.0,
```
realFlowControl: This should be an `integer`, not a `float`. I suspect it will fail once the `error_margin` line below is fixed. This is just not failing in the JSON validation, as the `percent` value (and others) are [validated in Go code](https://github.com/DataDog/prof-correctness/blob/401db806aef0eaa6dfc56654c89353e9b027a588/pprof_analysis.go#L148-L170) after the JSON schema validation succeeded.

The value of `1` might be too little as well.

profiling/tests/correctness/allocation_time_combined.json
morrisonlevi: It's because I want a non-zero value, and the true rate is around 0.5-1.0%. Setting a percent of 1 with margin of 1 would allow for 0, which is NOT good.

.github/workflows/prof_correctness.yml
morrisonlevi: Thanks, it was important to add because otherwise the math with sampling distances gets funky.

morrisonlevi:
So it's very easy to see in the log output that the optimization works, you'll see messages like this:

```
[TRACE datadog_php_profiling::profiling] Sent stack sample of 4 frames, 2 labels, 10000032 bytes allocated, 1 allocations, and 12 time interrupts to profiler.
```

Getting the prof-correctness tests right has been quite tricky. Locally, I can see that `alloc-size` is clearly working on PHP 8.4:

Flat | Flat% | Sum% | Cum | Cum% | Name
-- | -- | -- | -- | -- | --
591.28MB | 66.67% | 66.67% | 591.28MB | 66.67% | standard\|str_replace
295.64MB | 33.33% | 100.00% | 295.64MB | 33.33% | standard\|str_repeat
0 | 0.00% | 100.00% | 886.92MB | 100.00% | main
0 | 0.00% | 100.00% | 886.92MB | 100.00% | <?php

And the graph is right, so the path `<?php;main;standard\|str_replace$` should be good:

Screenshot 2026-01-16 at 6 56 23 PM

But when I run prof-correctness in CI, we get this failure:

```
    pprof_analysis.go:647: Analyzing results in /github/workspace/profiling/tests/correctness/allocation_time_combined/test.pprof.1.zst for profile type alloc-size
    pprof_analysis.go:650: Found a profile duration of 10.0 seconds (in test.pprof.1.zst)
    pprof_analysis.go:296: Results stored in  /github/workspace/profiling/tests/correctness/allocation_time_combined/test.pprof.1.json
    pprof_analysis.go:442: Assertion failed: stack '<?php;main;standard\|str_replace$' (labels=[]) should have been 66% +/- 1% of the profile but was 0% with 66% error
    pprof_analysis.go:445: Assertion succeeded: stack '<?php;main;standard\|str_repeat$' (labels=[]) is 33% +/- 1% of the profile (was 33% with 0% error)
```

What could be making this fail?

 - Would having labels cause it to fail? I don't think so, because there are labels and the one for `str_repeat` worked.
 - x86_64 vs aarch64?
 - macOS vs Linux?

profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -0,0 +1,46 @@
...
+        }
+      ]
+    },
+    {
+      "profile-type": "wall-time",
+      "stack-content": [
+        {
+          "regular_expression": "<?php;main;standard\\|str_replace$",
+          "percent": 51,
+          "error_margin": 50
```
morrisonlevi, line 41: You should read this as "any non-zero value is acceptable." This is because the test is focused around memory operations which do not have specific amounts of time, and what we are testing is that `str_replace` will have CPU time attributed to it because of our piggy-backing.

realFlowControl: I'd say it should still have the right amount of CPU time attributed, but I created #3592 for this case to not block this PR longer than necessary.

profiling/tests/correctness/allocation_time_combined.php
```diff
@@ -0,0 +1,25 @@
...
+function main() {
+    $duration = $_ENV["EXECUTION_TIME"] ?? 10;
+    $end = microtime(true) + $duration;
+
+    while (microtime(true) < $end) {
+        // str_replace is frameless in PHP 8.4+ and allocates a new string
+        $xs = str_repeat("x", 10_000_000); // 10MB source
+        $ys = str_replace("x", "y", $xs); // 10MB allocation in frameless function
+        $zs = str_replace("y", "z", $ys); // 10MB allocation in frameless function
+        strlen($zs); // Use the result to prevent optimization
```
realFlowControl:
```suggestion
```

The comment is misleading, the only reason non of this loop body get's replaced by the optimiser is that the `str_repeat()` above [allocates more then 64kb](https://github.com/php/php-src/blob/aeb8524584967bdebffec0e7896ca358a8071f3e/Zend/Optimizer/sccp.c#L779-L786). If you'd "just" repeat `10_000` times, the optimiser would evaluate the entire loop body at compile time and the dead code elimination would remove the entire loop body.

profiling/tests/correctness/allocation_time_combined.json
morrisonlevi: There is no "mathematically" calculated amount of time to attribute because we're doing memory operations, not inherently CPU oriented methods. I'll chime in over there, thanks!

PR #3566, opened 2026-01-12
Adds process_tags to the first span of each tracing payload

dubloom:
### Description

This PR implements this [RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?pli=1&tab=t.0#heading=h.s9l1lctqlg11) for the tracing product.

Add process_tags in APM payload. According to the RFC, the tag `_dd.tags.process` is set in the first first span of each payload.

The set tag are:
- `entrypoint.workdir`. The working dir name, should be the last path segment.
- `entrypoint.name`. The entrypoint of the application. It is the the script name. Set to none if php is executed as an executable.
- `entrypoint.type`. Either `script` or `executable` 
- `entrypoint.basedir`. The base dir is the directory where the called executable resides. Should be the last path segment.
- `runtime.sapi`. The name of the SAPI in which PHP is run. 

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/prof_correctness.yml
+++ b/.github/workflows/prof_correctness.yml
@@ -102,6 +102,10 @@ jobs:
           export DD_PROFILING_OUTPUT_PPROF=$PWD/profiling/tests/correctness/allocations_1byte/test.pprof
           export DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE=1
           php -d extension=$PWD/target/profiler-release/libdatadog_php_profiling.so profiling/tests/correctness/allocations.php
+          mkdir -p profiling/tests/correctness/allocations_1byte_no_zend_alloc/
+          export DD_PROFILING_OUTPUT_PPROF=$PWD/profiling/tests/correctness/allocations_1byte_no_zend_alloc/test.pprof
+          export DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE=1
+          USE_ZEND_ALLOC=0 php -d extension=$PWD/target/profiler-release/libdatadog_php_profiling.so profiling/tests/correctness/allocations.php
           unset DD_PROFILING_ALLOCATION_SAMPLING_DISTANCE
 
       - name: Run ZTS tests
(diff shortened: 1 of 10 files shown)


ext/process_tags.c
```diff
@@ -0,0 +1,228 @@
...
+        return;
+    }
+
+    size_t count = process_tags.count;
+    if (count == process_tags.capacity) {
+        process_tags.capacity *= 2;
+        process_tag_entry_t* bigger_list = realloc(
+            process_tags.tag_list,
+            process_tags.capacity * sizeof(process_tag_entry_t)
+        );
```
bwoebi:
```suggestion
        process_tag_entry_t* bigger_list = perealloc(
            process_tags.tag_list,
            process_tags.capacity * sizeof(process_tag_entry_t),
            1
        );
```
Also at other places, use pemalloc(, 1), and pefree(, 1). (They ultimately call system malloc functions, but defer to PHPs handling for OOM.)

bwoebi: Then you can remove the `if (!bigger_list)` completely.

ext/process_tags.c
```diff
@@ -0,0 +1,228 @@
...
+        process_tags.tag_list = bigger_list;
+    }
+
+    process_tags.tag_list[count].key = tag_key;
+    process_tags.tag_list[count].value = normalized_value;
+    process_tags.count++;
+
+}
+
+static void collect_process_tags(void) {
```
bwoebi, line 119: It is known and acknowledged, that for web requests, where multiple web applications are hosted on a single web server, or have multiple entrypoints, the values will always reflect the first request after process start (which may be distinct between different processes of the same webserver in multiprocessing mode - which is e.g. the default for FPM), given that they're collected in first rinit?
If yes, then it's fine.

dubloom: addressed in [127c555](https://github.com/DataDog/dd-trace-php/pull/3566/commits/127c555a44339cd6cc844f937f8ea67e90613a2e)

ext/process_tags.c
dubloom: addressed in [127c555](https://github.com/DataDog/dd-trace-php/pull/3566/commits/127c555a44339cd6cc844f937f8ea67e90613a2e)

bwoebi (approved):
Looks good :-)

PR #3567, opened 2026-01-12
fix(CI): make the upload script of junit XML file more resilient 

Leiyks:
The goal of this PR is to improve the script used to upload junit XML test file to Datadog.
The standalone binary version of `datadog-ci` is not working when trying to upload from alpine based image.
Therefore the script has been updated to install datadog-ci via npm/npx first before trying to install the standalone binary.

### DIFF
--- a/.gitlab/generate-profiler.php
+++ b/.gitlab/generate-profiler.php
@@ -42,19 +42,19 @@ foreach ($profiler_minor_major_targets as $version) {
     - run_tests_php=$(find $(php-config --prefix) -name run-tests.php) # don't anticipate there being more than one
     - cp -v "${run_tests_php}" tests
     - unset DD_SERVICE; unset DD_ENV
+    - mkdir -p "${CI_PROJECT_DIR}/artifacts/profiler-tests"
 
     - '# NTS'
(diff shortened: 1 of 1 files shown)


realFlowControl (approved):
Nice, that works now also for profiling. The [test suite](https://app.datadoghq.com/ci/test/runs?query=test_level%3Atest%20%40git.repository.id_v2%3A%22github.com%2Fdatadog%2Fdd-trace-php%22%20%40git.commit.sha%3A9818201421fa460dacca6181078facc889a7543a%20%40git.branch%3A%22leiyks%2Ffix-junit-upload%22%20%40test.suite%3Aphp.phpt&_user_click_origin_=github&agg_m=count&agg_m_source=base&agg_t=count&currentTab=performance&eventStack=&fromUser=false&index=citest&start=1767694841104&end=1768299641104&paused=false) for the profiler is `php.phpt` which looks like to be the string `php` followed by the relative path (where `-` and `/` are replaced with `.`). So from my understanding we could move our profiler tests into another dir to have the suite name "corrected".
@DataDog/profiling-php can do this in a separate PR

PR #3571, opened 2026-01-13
test(profiling): add function name length tests

realFlowControl:
### Description

Re-adds the tests that got removed in #3570 and fixes the linking problem. You may review commit by commit.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (approved):
Hey, I added a commit to make a `zend_string_alloc` equivalent helper and then nudge it closer to what the engine really does. Please check it out before merging. Rest looks good to me! We're hitting the rate limit but these were passing before, I can't see how I messed them up.

```
#2 ERROR: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/datadog/dd-appsec-php-ci/manifests/sha256:15a9e23372df35fbb9c587ef7d3ec69be4b134be9742e108a70999f081e75e53: 429 Too Many Requests - Server message: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```

PR #3573, opened 2026-01-13
feat(sidecar): add thread mode as fallback connection for restricted environments

Leiyks:
### Description

Adds a new sidecar connection mode (`DD_TRACE_SIDECAR_CONNECTION_MODE`) that allows the sidecar to run as a thread within the PHP process instead of a separate subprocess. This is useful in restricted environments where spawning subprocesses is not allowed (e.g. some container runtimes, managed hosting).

**New INI setting:** `datadog.trace.sidecar_connection_mode`
- `auto` (default): try subprocess first, fall back to thread if it fails
- `subprocess`: force subprocess mode (previous behavior)
- `thread`: force thread mode (useful in restricted environments; **not compatible with `pcntl_fork()`**)

**Key behaviors:**
- Thread mode uses an abstract Unix socket (Linux) or named pipe (Windows) for IPC between master and worker processes in PHP-FPM
- In FPM, the master process owns the listener thread; worker processes connect to it without spawning their own thread
- The master UID is encoded in the socket name to support cross-user SHM access (e.g. when FPM master runs as root and workers run as `www-data`)
- Orphaned workers promote themselves to master if the original master's thread listener is unavailable
- Fork-safe: thread mode is detected and handles `pcntl_fork()` reconnection gracefully
- Windows support: threaded connection mode works on Windows where subprocess mode was previously the only option

**Tests added:**
- `SidecarThreadModeTest`: multi-request thread mode behavior
- `SidecarThreadModeRootTest`: FPM root+user-switch cross-user SHM scenario (runs in dedicated `fpm-fcgi` CI jobs)
- `.phpt` tests for each connection mode value and `auto` fallback behavior

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/ddtrace.c
```diff
@@ -1612,7 +1615,11 @@ static PHP_MSHUTDOWN_FUNCTION(ddtrace) {
 
     ddtrace_user_req_shutdown();
 
-    ddtrace_sidecar_shutdown();
+    // Only shutdown sidecar in MSHUTDOWN for non-CLI SAPIs
+    // CLI SAPI shuts down in RSHUTDOWN to allow thread joins before ASAN checks
```
bwoebi: How does ASAN influence this? ASAN is supposed to run after MSHUTDOWN as well?
I suspect that is related to the timings and the fact that the libdatadog part doesn't properly handle shutdown currently.
And this should in fact always shutdown the sidecar in MSHUTDOWN.

ext/ddtrace.c
```diff
@@ -2632,6 +2639,21 @@ void dd_internal_handle_fork(void) {
...
+                             current_pid != ddtrace_sidecar_master_pid);
+
+    if (is_child_process && ddtrace_sidecar_active_mode == DD_SIDECAR_CONNECTION_THREAD) {
+        // Clear inherited master listener state (child doesn't own it)
+        ddtrace_ffi_try("Failed clearing inherited listener state",
+                        ddog_sidecar_clear_inherited_listener());
+
+        // Don't try to reconnect in thread mode after fork
+        // Let sidecar stay unavailable
+        LOG(WARN, "Child process after fork with thread mode: sidecar unavailable");
```
bwoebi: Uhm, no? Just connect to the parents process socket? At least attempt to.

ext/handlers_pcntl.c
```diff
@@ -37,6 +37,15 @@ static void dd_prefork() {
 
 static void dd_handle_fork(zval *return_value) {
     if (Z_LVAL_P(return_value) == 0) {
+        // CHILD PROCESS
+
+        // Warn if thread mode is active
+        if (ddtrace_sidecar_active_mode == DD_SIDECAR_CONNECTION_THREAD) {
+            LOG(WARN, "pcntl_fork() detected with thread-based sidecar connection. "
+                      "Thread mode is incompatible with fork and may cause instability. "
+                      "Consider using subprocess mode (DD_TRACE_SIDECAR_CONNECTION_MODE=subprocess).");
```
bwoebi: It is? Why?

tests/ext/sidecar_connection_mode_thread_phpfpm.phpt
```diff
@@ -0,0 +1,243 @@
...
+Thread mode connection with PHP-FPM (manual verification test)
+--SKIPIF--
+<?php
+// This test is currently skipped because automated PHP-FPM thread mode testing
+// in .phpt format is not feasible. The thread listener must be started in the
+// PHP-FPM master process, not in the CLI test script.
+//
+// For manual testing of thread mode with PHP-FPM, see:
+// tests/ext/sidecar_connection_mode_thread_phpfpm_manual.md
+die('skip: Automated PHP-FPM thread mode testing not feasible in .phpt format. See sidecar_connection_mode_thread_phpfpm_manual.md for manual test procedure.');
```
bwoebi: Uh yeah, that one has no place here - the SidecarThreadModeTest ought to be covering this. Just drop this.

components-rs/bytes.rs
```diff
@@ -150,28 +151,28 @@ fn convert_literal_to_bytes_string(string: *const c_char) -> BytesString {
 }
 
 #[no_mangle]
-pub extern "C" fn ddog_set_span_service_zstr(ptr: &mut SpanBytes, str: &mut ZendString) {
+pub extern "C" fn ddog_set_span_service_zstr(ptr: &mut Span<BytesData>, str: &mut ZendString) {
```
bwoebi: Why this change?

ext/sidecar.c
```diff
@@ -178,18 +187,174 @@ static ddog_SidecarTransport *dd_sidecar_connection_factory_ex(bool is_fork) {
...
+    ddog_SidecarTransport *sidecar_transport;
+    if (!ddtrace_ffi_try("Failed connecting to sidecar as worker",
+                        ddog_sidecar_connect_worker((int32_t)ddtrace_sidecar_master_pid, &sidecar_transport))) {
         dd_free_endpoints();
         return NULL;
     }
 
     dd_sidecar_post_connect(&sidecar_transport, is_fork, logpath);
 
+    ddtrace_sidecar_reconnect(&sidecar_transport, dd_sidecar_connection_factory_thread);
```
bwoebi: I don't get this reconnect, I thought we just connected a couple lines above?

ext/sidecar.c
```diff
@@ -216,27 +381,117 @@ bool ddtrace_sidecar_maybe_enable_appsec(bool *appsec_activation, bool *appsec_c
...
+
+        if (!ddog_sidecar_is_master_listener_active(ddtrace_sidecar_master_pid)) {
+            LOG(WARN, "Child process cannot reconnect: parent's listener not active");
+            return;
+        }
+
+        ddtrace_force_new_instance_id();
+
+        // Attempt to connect as a worker to parent's listener
+        ddtrace_sidecar = ddtrace_sidecar_connect_as_worker(true);
```
bwoebi: Shouldn't it just try, and fall back to creating a new master, if it didn't work?
I feel like you cheated around that with calling ddtrace_sidecar_reconnect inside ddtrace_sidecar_connect_as_worker ... but that's really weird :-D

ext/sidecar.c
```diff
@@ -216,27 +381,117 @@ bool ddtrace_sidecar_maybe_enable_appsec(bool *appsec_activation, bool *appsec_c
...
+        LOG(INFO, "Child process reconnected to parent's sidecar listener after fork (child PID=%d, parent=%d)",
+            (int32_t)getpid(), ddtrace_sidecar_master_pid);
     }
+    // Handle subprocess mode fork - reset and spawn new subprocess
+    else if (ddtrace_sidecar_active_mode == DD_SIDECAR_CONNECTION_SUBPROCESS) {
+        ddtrace_force_new_instance_id();
+
+        if (ddtrace_sidecar) {
+            ddog_sidecar_transport_drop(ddtrace_sidecar);
+            ddtrace_sidecar = NULL;
```
bwoebi: This part should run independent of connection mode, right? ddtrace_sidecar merely holds the connection. (not the thread, which we must not drop)

ext/sidecar.c
```diff
@@ -261,8 +516,29 @@ void ddtrace_sidecar_finalize(bool clear_id) {
 }
 
 void ddtrace_sidecar_shutdown(void) {
+#ifndef _WIN32
```
bwoebi: Why not on windows?

bwoebi: This only doesn't need to compare the pids on windows, right? But shutdown obviously still ought to happen.

ext/ddtrace_arginfo.h
```diff
@@ -547,6 +547,9 @@ static void register_ddtrace_symbols(int module_number)
 	REGISTER_LONG_CONSTANT("DDTrace\\DBM_PROPAGATION_DISABLED", DD_TRACE_DBM_PROPAGATION_DISABLED, CONST_PERSISTENT);
 	REGISTER_LONG_CONSTANT("DDTrace\\DBM_PROPAGATION_SERVICE", DD_TRACE_DBM_PROPAGATION_SERVICE, CONST_PERSISTENT);
 	REGISTER_LONG_CONSTANT("DDTrace\\DBM_PROPAGATION_FULL", DD_TRACE_DBM_PROPAGATION_FULL, CONST_PERSISTENT);
+	REGISTER_LONG_CONSTANT("DDTrace\\SIDECAR_CONNECTION_MODE_AUTO", DD_TRACE_SIDECAR_CONNECTION_MODE_AUTO, CONST_PERSISTENT);
+	REGISTER_LONG_CONSTANT("DDTrace\\SIDECAR_CONNECTION_MODE_SUBPROCESS", DD_TRACE_SIDECAR_CONNECTION_MODE_SUBPROCESS, CONST_PERSISTENT);
+	REGISTER_LONG_CONSTANT("DDTrace\\SIDECAR_CONNECTION_MODE_THREAD", DD_TRACE_SIDECAR_CONNECTION_MODE_THREAD, CONST_PERSISTENT);
```
bwoebi: Where does this come from? You know this file is generated from the stubs? I see no change in the stubs, nor do we actually expose any of these values as ints, do we?

ext/sidecar.c
```diff
@@ -216,27 +381,117 @@ bool ddtrace_sidecar_maybe_enable_appsec(bool *appsec_activation, bool *appsec_c
...
+
+    if (mode == DD_TRACE_SIDECAR_CONNECTION_MODE_THREAD) {
+        ddtrace_sidecar_setup_thread_mode(appsec_activation, appsec_config);
+        return;
+    }
+
     ddtrace_set_non_resettable_sidecar_globals();
     ddtrace_set_resettable_sidecar_globals();
 
     ddog_init_remote_config(get_global_DD_INSTRUMENTATION_TELEMETRY_ENABLED(), appsec_activation, appsec_config);
```
bwoebi: Why do you duplicate these couple lines multiple times (into ddtrace_sidecar_setup_thread_mode)? Should be simply at the start of ddtrace_sidecar_setup and never duplicated?

ext/sidecar.c
```diff
@@ -178,18 +187,174 @@ static ddog_SidecarTransport *dd_sidecar_connection_factory_ex(bool is_fork) {
...
+    char logpath[MAX_PATH];
+#else
+    char logpath[MAXPATHLEN];
+#endif
+    int error_fd = atomic_load(&ddtrace_error_log_fd);
+    if (error_fd == -1 || ddtrace_get_fd_path(error_fd, logpath) < 0) {
+        *logpath = 0;
+    }
+
+    ddog_SidecarTransport *sidecar_transport;
```
bwoebi, line 222: Also here, this is an exact copy, can we dedupe this?

bwoebi: Can we have `ddtrace_sidecar_connect(bool as_worker, bool is_fork)` simply?

components-rs/bytes.rs
Leiyks: This was needed when a rebased my libdatadog branch, I guess that's not needed anymore and I'll remove it once I rebase my branches

ext/sidecar.c
Leiyks: Correct, that's dead code I forgot to remove

ext/sidecar.c
Leiyks: The ddtrace_sidecar_reconnect call was just dead code.
But good point, I'll fix the logic here :ok:

ext/sidecar.c
Leiyks: Oh yes, forgot to change this when adding windows support

ext/ddtrace_arginfo.h
Leiyks: dead code as well

ext/sidecar.c
```diff
@@ -198,28 +207,141 @@ static ddog_SidecarTransport *dd_sidecar_connection_factory_ex(bool is_fork) {
...
 
     ddog_SidecarTransport *sidecar_transport;
-    if (!ddtrace_ffi_try("Failed connecting to the sidecar", ddog_sidecar_connect_php(&sidecar_transport, logpath, dd_zend_string_to_CharSlice(get_global_DD_TRACE_LOG_LEVEL()), get_global_DD_INSTRUMENTATION_TELEMETRY_ENABLED(), dd_sidecar_on_reconnect, ddtrace_endpoint))) {
-        dd_free_endpoints();
-        return NULL;
+    if (as_worker) {
+        if (!ddtrace_ffi_try("Failed connecting to sidecar as worker",
+                             ddog_sidecar_connect_worker((int32_t)ddtrace_sidecar_master_pid, &sidecar_transport))) {
+            dd_free_endpoints();
+            return NULL;
```
bwoebi: We need to fall back to creating a master here, I believe. At least if current_pid != master_pid.

bwoebi: And only otherwise print an error.

bwoebi: The specific scenario is: the parent forks and dies. The child lives on ... but it still must submit traces.
(Please add a test for this.)

tests/Sapi/PhpFpm/PhpFpm.php
```diff
@@ -88,10 +96,17 @@ public function __construct($rootPath, $host, $port, array $envs = [], array $in
...
+            $allowRoot = ' --allow-to-run-as-root';
+        }
+
         $cmd = sprintf(
-            'php-fpm -p %s --fpm-config %s -F',
+            'php-fpm -p %s --fpm-config %s -F%s',
             __DIR__,
-            $this->configFile
+            $this->configFile,
+            $allowRoot
```
bwoebi: This change ought to not be necessary I think?
Can you explain this? _If_ there is one case where we want to actually _start_ the webserver as root, we probably still want to end up using it as a different user (via fpm www.conf config) to ensure that permissions are correct? In which case --allow-to-run-as-root should be unneeded.

I.e. the `user` directive under `[www]` section.

ext/sidecar.c
```diff
@@ -246,27 +398,128 @@ bool ddtrace_sidecar_maybe_enable_appsec(bool *appsec_activation, bool *appsec_c
...
+#else
+    ddtrace_sidecar_master_pid = (int32_t)getpid();
+#endif
+
+    zend_long mode = get_global_DD_TRACE_SIDECAR_CONNECTION_MODE();
+
+    if (mode == DD_TRACE_SIDECAR_CONNECTION_MODE_THREAD) {
+#ifndef _WIN32
+        if (geteuid() == 0) {
+            ddog_sidecar_set_shm_open_mode(0644);
```
bwoebi: These files aren't readonly.
Agent info is for example, but limiters aren't (= need to be writeable by child processes).

We really will need to call fchown() with the appropriate user id.

bwoebi: The sidecar also only lazily initializes its shm structures. Hence, we can just grab the uid of the first connecting process and initialize it when accepting the first socket connection.

To transfer this information, provide it via ucred struct (see SCM_CREDENTIALS on unix(7) man page) on the socket.

ext/sidecar.c
```diff
@@ -246,27 +398,128 @@ bool ddtrace_sidecar_maybe_enable_appsec(bool *appsec_activation, bool *appsec_c
...
+bool ddtrace_sidecar_should_enable(bool *appsec_activation, bool *appsec_config) {
+    bool enable_sidecar = ddtrace_sidecar_maybe_enable_appsec(appsec_activation, appsec_config);
+    if (!enable_sidecar) {
+        enable_sidecar = get_global_DD_INSTRUMENTATION_TELEMETRY_ENABLED() ||
+                       get_global_DD_TRACE_SIDECAR_TRACE_SENDER();
+    }
+    return enable_sidecar;
+}
+
 void ddtrace_sidecar_setup(bool appsec_activation, bool appsec_config) {
```
bwoebi, line 409: You do realize that this code is _still_ running in first rinit? The whole premise of the thread stuff is to start the thread in minit. (so that it lives on the parent master process.)

tests/Sapi/PhpFpm/PhpFpm.php
Leiyks: Nevermind this, it was a previous attempt to fix some test. I just forgot to revert it

ext/sidecar.c
Leiyks: yes indeed there was an issue, it's should be fixed now :ok_hand:

.gitlab/build-sidecar.sh
```diff
@@ -13,6 +13,15 @@ set -u
...
 
+# Install automake/libtool required by libdd-libunwind-sys build.rs (autoreconf step)
+if [ "${suffix}" = "-alpine" ]; then
+  apk add --no-cache automake libtool
+elif command -v apt-get &>/dev/null; then
+  apt-get install -y --no-install-recommends automake libtool
+elif command -v yum &>/dev/null; then
+  yum install -y automake libtool
+fi
+
```
bwoebi: Fwiw: the base image now includes those.

profiling/Cargo.toml
```diff
@@ -23,10 +23,10 @@ cpu-time = { version = "1.0" }
...
 crossbeam-channel = { version = "0.5", default-features = false, features = ["std"] }
 http = { version = "1.4" }
-libdd-alloc = {  git = "https://github.com/DataDog/libdatadog", tag = "v27.0.0" }
-libdd-profiling = {  git = "https://github.com/DataDog/libdatadog", tag = "v27.0.0" }
-libdd-common = {  git = "https://github.com/DataDog/libdatadog", tag = "v27.0.0" }
-libdd-library-config-ffi = { path = "../libdatadog/libdd-library-config-ffi" }
+libdd-alloc = {  git = "https://github.com/DataDog/libdatadog", tag = "v28.0.3" }
+libdd-profiling = {  git = "https://github.com/DataDog/libdatadog", tag = "v28.0.3" }
+libdd-common = {  git = "https://github.com/DataDog/libdatadog", tag = "v28.0.3" }
+libdd-library-config-ffi = { git = "https://github.com/DataDog/libdatadog", tag = "v28.0.3" }
```
bwoebi, line 29: Is this actually necessary? Anyway. We'll see soon.

PR #3574, opened 2026-01-13
Reduce CI tests flakiness

Leiyks:
### Description                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                        
  This PR fixes multiple sources of test flakiness in CI by implementing proper retry/wait logic and fixing dependency installation.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                        
  1. Laravel Octane Test Flakiness                                                                                                                                                                                                                                                      
  - Added waitUntilServerRunning() to properly wait for Octane server startup instead of fixed sleep(1)                                                                                                                                                                                 
  - Eliminates race condition causing NULL traces when tests run before server is ready                                                                                                                                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                        
  2. SQLSRV Connection Flakiness                                                                                                                                                                                                                                                        
  - Added retry logic (30 attempts, 500ms delay) for SQL Server connections                                                                                                                                                                                                             
  - Handles variable container startup time

  3. PHP-FPM/Nginx Startup Race Condition                                                                                                                                                                                                                                               
  - Added waitUntilServerRunning() to both PhpFpm and NginxServer classes                                                                                                                                                                                                               
  - Fixed startup order: PHP-FPM starts first (nginx depends on it), then nginx                                                                                                                                                                                                         
  - Eliminates 502 Bad Gateway errors when tests run before backends are ready                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                        
  4. System Tests Docker Hub Authentication                                                                                                                                                                                                                                             
  - Auto-install unzip (via apt-get/apk) when missing to extract Vault CLI                                                                                                                                                                                                              
  - Fixes "unzip: not found" error preventing Docker Hub login                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                        
  5. CentOS Package Verification                                                                                                                                                                                                                                                        
  - Added retry loop (3 attempts) for yum update                                                                                                                                                                                                                                        
  - Handles vault.centos.org reliability issues

### DIFF
--- a/tests/Integrations/SQLSRV/SQLSRVTest.php
+++ b/tests/Integrations/SQLSRV/SQLSRVTest.php
@@ -34,6 +34,43 @@ class SQLSRVTest extends IntegrationTestCase
     {
         parent::ddSetUpBeforeClass();
         self::putenv('DD_SQLSRV_ANALYTICS_ENABLED=true');
+        self::waitForSqlServerReady();
+    }
+
+    private static function waitForSqlServerReady()
+    {
+        $maxAttempts = 30;
+        $attempt = 0;
+        $conn = false;
+
+        while ($attempt < $maxAttempts && $conn === false) {
+            $attempt++;
+            $conn = @sqlsrv_connect(
+                self::$host . ', ' . self::$port,
+                [
+                    'PWD' => self::$password,
+                    'Database' => self::$db,
+                    'UID' => self::$user,
+                    'TrustServerCertificate' => true,
+                ]
+            );
+
+            if ($conn === false) {
+                if ($attempt < $maxAttempts) {
+                    usleep(500000);
+                }
+            }
+        }
+
+        if ($conn === false) {
+            $errors = sqlsrv_errors();
+            $lastError = $errors ? json_encode($errors) : 'Unknown error';
+            throw new \RuntimeException(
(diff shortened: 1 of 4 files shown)


tests/Integrations/SQLSRV/SQLSRVTest.php
```diff
@@ -502,15 +502,39 @@ public function testNoFakeServices()
...
-        );
+        // Retry connection to handle SQL Server container startup time
+        $maxAttempts = 30;
+        $attempt = 0;
+        $conn = false;
+        $lastError = null;
+
+        while ($attempt < $maxAttempts && $conn === false) {
+            $attempt++;
+            $conn = sqlsrv_connect(
```
bwoebi: Won't that mess with some test expectations (i.e. multiple sqlsrv_connect spans instead of one if the first one doesn't succeed)? Doesn't this logic belong into ddSetUpBeforeClass - it just needs to succeed once connecting.

Leiyks: Indeed, nice catch. Fixed it :ok_hand:

PR #3577, opened 2026-01-15
Profiling: refactor AgentEndpoint so it can be const

morrisonlevi:
### Description

This refactors `AgentEndpoint` to use a `Cow<Path>` instead of a `PathBuf` so that we can use `&'static Path`. This avoids an allocation which is a little bit nice, but mainly it's to allow us to make a `const` one: 

```rs
AgentEndpoint::Socket(Cow::Borrowed(AgentEndpoint::DEFAULT_UNIX_SOCKET_PATH))
```

It's a little tricky to make a const `&Path` because `Path::new` isn't const, so it uses some unsafe tricks.  Note that I ran this by miri locally and it was fine, as I expected.

This UDS path is not necessarily the "best" default because it's only partly true: in practice, we use this as the default agent path if it exists, but if it doesn't, then we use `"http://localhost:8126"`. But it has the distinct advantage of being one that can be defined in const contexts.  I plan to use this in another PR to simplify `SystemSettings`.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/config.rs
+++ b/profiling/src/config.rs
@@ -177,7 +177,30 @@ impl SystemSettings {
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum AgentEndpoint {
     Uri(Uri),
-    Socket(PathBuf),
+    Socket(Cow<'static, Path>),
+}
+
+impl AgentEndpoint {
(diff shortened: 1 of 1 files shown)


morrisonlevi:
Yes, that's right. This should make things safer and easier to work with!

PR #3578, opened 2026-01-15
Profiling: fix UB and simplify SystemSettings

morrisonlevi:
### Description

Currently we're using `MaybeUninit` and doing various memory operations on `SYSTEM_SETTINGS`. With a tweak made in #3577, we can simplify this. 

Additionally, this fixes likely undefined behavior on ZTS and possibly even NTS builds as well. There was a mutable reference taken to the `SystemSettings` in `rinit`. Note that we did not _actually_ mutate it, so in practice we were likely fine, but this would have been instant undefined behavior simply to materialize the mutable reference if any other references existed, which in ZTS would have been quite likely (other workers could also be in `rinit`).

Note that there _shouldn't_ be any meaningful changes with regards to the change in the URI default, but there is one thing in particular we should watch out for: the default uri _may_ have changed if we have some bad logic somewhere. I'll call it out with a CR comment.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/config.rs
+++ b/profiling/src/config.rs
@@ -14,13 +14,30 @@ use core::str::FromStr;
 use libc::{c_char, c_int};
 use libdd_common::tag::{parse_tags, Tag};
 pub use libdd_profiling::exporter::Uri;
-use log::{warn, LevelFilter};
+use log::{debug, warn, LevelFilter};
 use std::borrow::Cow;
 use std::ffi::CString;
 use std::path::{Path, PathBuf};
 
+#[derive(Copy, Clone, Debug, Default)]
+pub enum SystemSettingsState {
+    /// Indicates the system settings are not aware of the configuration at
+    /// the moment.
+    #[default]
+    ConfigUnaware,
+
+    /// Indicates the system settings _are_ aware of configuration at the
+    /// moment.
+    ConfigAware,
+
+    /// Expressly disabled, such as a child process post-fork (forks are not
+    /// currently profiled, except certain forks made by SAPIs).
+    Disabled,
+}
+
 #[derive(Clone, Debug)]
 pub struct SystemSettings {
+    pub state: SystemSettingsState,
     pub profiling_enabled: bool,
     pub profiling_experimental_features_enabled: bool,
     pub profiling_endpoint_collection_enabled: bool,
@@ -44,6 +61,7 @@ impl SystemSettings {
     /// Provides "initial" settings, which are all "off"-like values.
     pub const fn initial() -> SystemSettings {
         SystemSettings {
+            state: SystemSettingsState::ConfigUnaware,
             profiling_enabled: false,
             profiling_experimental_features_enabled: false,
             profiling_endpoint_collection_enabled: false,
@@ -73,6 +91,7 @@ impl SystemSettings {
         let trace_agent_url = trace_agent_url();
         let uri = detect_uri_from_config(trace_agent_url, agent_host, trace_agent_port);
         Self {
+            state: SystemSettingsState::ConfigAware,
             profiling_enabled: profiling_enabled(),
             profiling_experimental_features_enabled: profiling_experimental_features_enabled(),
             profiling_endpoint_collection_enabled: profiling_endpoint_collection_enabled(),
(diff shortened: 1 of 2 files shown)


profiling/src/lib.rs
```diff
@@ -449,7 +453,7 @@ impl Default for RequestLocals {
             git_commit_sha: None,
             git_repository_url: None,
             tags: vec![],
-            system_settings: ptr::NonNull::from(INITIAL_SYSTEM_SETTINGS.deref()),
+            system_settings: SystemSettings::get(),
```
morrisonlevi, line 456: Here's an example of a potential behavior divergence. Before, this would have used `INITIAL_SYSTEM_SETTINGS` which would have done `AgentEndpoint::default()`, but now, it will have the value `SystemSettings` uses, which could just be the socket depending on when exactly it was called. More precisely, `AgentEndpoint::default()` checks that the apm.socket exists before using it, and if it doesn't exist, it falls back to the URI.

Now, this shouldn't really matter in any scenarios, because we really should be changing these to _real_ values ASAP. It's something to specifically scrutinize in the PR.

profiling/src/lib.rs
```diff
@@ -323,6 +323,9 @@ extern "C" fn minit(_type: c_int, module_number: c_int) -> ZendResult {
         let _connector = libdd_common::connector::Connector::default();
     }
 
+    // Initialize the lazy lock holding the env var for new origin detection.
+    _ = std::sync::LazyLock::force(&libdd_common::entity_id::DD_EXTERNAL_ENV);
```
morrisonlevi, line 327: This is unrelated to this PR, but it's a safer place for it. Maybe I should make it its own PR but it's 10:35pm for me right now when I noticed this.

profiling/src/profiling/mod.rs
```diff
@@ -904,7 +907,7 @@ impl Profiler {
 
                 let mut timestamp = NO_TIMESTAMP;
                 {
-                    let system_settings = self.system_settings.load(Ordering::SeqCst);
+                    let system_settings = self.system_settings.load(Ordering::Relaxed);
```
danielsn: why is relaxed safe here

profiling/src/config.rs
```diff
@@ -83,15 +92,16 @@ impl SystemSettings {
...
 
 impl SystemSettings {
-    /// # Safety
-    /// Must be called after [first_rinit] and before [shutdown].
-    pub unsafe fn get() -> ptr::NonNull<SystemSettings> {
-        // SAFETY: required by this function's own safety requirements.
-        let addr = unsafe { (*ptr::addr_of_mut!(SYSTEM_SETTINGS)).assume_init_mut() };
-        ptr::NonNull::from(addr)
+    /// Returns the "current" system settings, which are always memory-safe
+    /// but may point to "initial" values rather than the configured ones,
```
danielsn, line 118: should there be an "is_initial" field on the settings for debugging purposes?

profiling/src/profiling/mod.rs
morrisonlevi: Right, good point! I will document that. The answer is that this atomic is used for being Send + Sync, not for it's atomicity really (literally nothing ever changes this to point at a different address).

morrisonlevi: Documented in the new `Profiler::is_timeline_enabled` method 👍🏻

profiling/src/config.rs
morrisonlevi: I've pushed a commit with state tracking, I'll play a little bit with it but don't have time for a lot of experimentation right now.

profiling/src/config.rs
```diff
@@ -14,13 +14,30 @@ use core::str::FromStr;
...
+
+    /// Indicates the system settings _are_ aware of configuration at the
+    /// moment.
+    ConfigAware,
+
+    /// Expressly disabled, such as a child process post-fork (forks are not
+    /// currently profiled, except certain forks made by SAPIs).
+    Disabled,
+}
+
```
realFlowControl, line 37: I like that 👍

PR #3579, opened 2026-01-16
Profiling: fix SystemSettings initialization

morrisonlevi:
### Description

This PR:

- Moves initialization of `SystemSettings` to minit.
- Re-initializes `SystemSettings` on first request to pick up env vars sent by php-fpm/apache (legacy behavior, IMO, kept to preserve behavior).
- Plumbs through `NonZeroU32`/`NonZeroU64 ` so that it's easier to see that the sampling distance cannot be <= 0.
- Adds a `.phpt` test to prevent future regressions.

Builds on #3578.

# Motivation

This fixes a recent regression where `datadog.profiling.allocation_sampling_distance` wasn't being applied. This occurred because we moved the initialization of the profiling stats to `ginit`, and at that point, `SystemSettings` are not initialized, so it picked up the default value.

I hit this regression in my work on PR #3559.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -1542,7 +1542,6 @@ mod tests {
     use crate::{allocation::DEFAULT_ALLOCATION_SAMPLING_INTERVAL, config::AgentEndpoint};
     use libdd_profiling::exporter::Uri;
     use log::LevelFilter;
-    use crate::config::SystemSettingsState;
 
     fn get_frames() -> Vec<ZendFrame> {
         vec![ZendFrame {


profiling/tests/phpt/allocation_sampling_distance.phpt
```diff
@@ -0,0 +1,49 @@
...
+<?php
+if (!extension_loaded('datadog-profiling'))
+    die("skip: test requires datadog-profiling");
+?>
+--INI--
+datadog.profiling.enabled=1
+datadog.profiling.allocation_enabled=1
+datadog.profiling.allocation_sampling_distance=1
+datadog.profiling.log_level=trace
+datadog.profiling.output_pprof=/tmp/profiling-data/pprof
```
realFlowControl: I think this should not be set in the test. CI sets this as an ENV variable to make sure tests run faster and not upload.
```suggestion
```

profiling/tests/phpt/allocation_sampling_distance.phpt
```diff
@@ -0,0 +1,48 @@
...
+//     112 - 24 - 1 = strlen($str)
+//               87 = strlen($str)
+// ... except that the engine does the rounding in the wrong place for
+// str_repeat, so it with our inputs it ends up doing this, with `f` being the
+// function which determines the rounded amount:
+//     1 * 87 + f(24 + 0 + 1)
+//         87 + f(25)
+//         87 + 32 = 119
+// So it over-allocates by 7 bytes, and the log will have 119 bytes.
+// Our regex allows for 112 (correct) and 119 (observed).
```
morrisonlevi, line 41: @danielsn  This little detail cost 1.5 hours of my day. Not now, but it's something we can fix in php-src some day.

PR #3592, opened 2026-01-21
ci(prof): run additional allocation test on all PHP versions

realFlowControl:
### Description

This PR is based on #3559 with the intention to run that new correctness test on all PHP versions and provide a value and error range that matches all PHP versions. Additionally this runs the `allocation_time_combined` test without an adjusted sampling distance for allocations 🎉 

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/prof_correctness.yml
+++ b/.github/workflows/prof_correctness.yml
@@ -47,13 +47,16 @@ jobs:
 
       - name: Build profiler
         run: |
-          echo "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-17 main" | sudo tee -a /etc/apt/sources.list
-          echo "deb-src http://apt.llvm.org/jammy/ llvm-toolchain-jammy-17 main" | sudo tee -a /etc/apt/sources.list
-          curl https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-          curl https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
-          sudo apt remove -y clang-*
+          codename="$(lsb_release -cs)"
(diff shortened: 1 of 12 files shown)


profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -23,12 +23,12 @@
         {
           "regular_expression": "<?php;main;standard\\|str_replace$",
           "percent": 66,
-          "error_margin": 1
+          "error_margin": 3
         },
         {
           "regular_expression": "<?php;main;standard\\|str_repeat$",
           "percent": 33,
-          "error_margin": 1
```
morrisonlevi, line 31: Did you observe 1% to be too narrow?

profiling/tests/correctness/allocation_time_combined.json
```diff
@@ -37,8 +37,8 @@
       "stack-content": [
         {
           "regular_expression": "<?php;main;standard\\|str_replace$",
-          "percent": 51,
-          "error_margin": 50
+          "percent": 70,
+          "error_margin": 30
```
morrisonlevi, line 41: This stack is for wall-time. The idea here is that `str_replace` will use enough CPU that the allocation time spent is negligible, and the replace operation will dominate it.

The variability comes from whether it's piggy-backing on the allocation interrupt (for frameless versions, where we don't get an interrupt on the frameless call) or on older versions without it, we get an interrupt, and it's more accurate.

profiling/tests/correctness/allocation_time_combined.json
realFlowControl: For CPU time yes, not for allocations. Allocations in PHP are exactly predictable. This may only change if there is an upstream change and so far it is exactly the same from PHP 7 to 8

PR #3595, opened 2026-01-21
feat(prof): use a trampoline for FLF functions to intercept timings

bwoebi:
Frameless functions (FLF) do not check the `EG(vm_interrupt)` flag like it would happen usually on internal function returns. This means that from the point of view of the profiler, those functions do not exist, as the engine never checks the interrupt flag and as such never calls the profilers interrupt handler.

This PR adds a trampoline for `aarch64` and `x84_64` to all known frameless functions and a tail call to check the `EG(vm_interrupt)` flag and in case it is raised, call the interrupt handler.

One (acceptable) trade-off is that using these trampolines, `libunwind` in the crash tracker is not able to unwind passed the trampoline. Anyway, the impact is minimal:
- Stack frames from the crash site up to the trampoline are still captured
- Only frames above the trampoline (from caller to main) are lost
- For most crashes, the relevant context is near the crash site, not at the program entry point

The profiler itself is unaffected of this, as we are unwinding the VM stack, following the `execute_data` linked list.

https://datadoghq.atlassian.net/browse/PROF-12085


profiling/src/wall_time.rs
```diff
@@ -137,6 +137,88 @@ pub extern "C" fn ddog_php_prof_interrupt_function(execute_data: *mut zend_execu
...
+            let interrupt_addr = ddog_php_prof_interrupt_function as *const () as i64;
+            #[cfg(target_arch = "aarch64")]
+            dynasm!(assembler
+                ; mov x16, original as i64
+                ; blr x16
+                ; mov x0, 0
+                ; mov x16, interrupt_addr
+                ; blr x16
+                ; ret
+            );
```
morrisonlevi: So basically we load the original function and call it, zero out the return register, then we load the interrupt function and call it. Is that right?  Why is zeroing the return register necessary here? I am no assembly expert, and even less so in aarch64 than x86_64.

morrisonlevi (commented):
I assume this can run afoul of permissions somehow, since we're making executable code at runtime? I'm not well versed here, and yes, if the JIT is enabled you'd have to have that capability anyway, but I'm trying to understand implications of this WIP.

profiling/src/wall_time.rs
bwoebi: x0 is the first arg. But I was confusing myself, obviously it doesn't take NULL (was testing), but EG(current_execute_data). Fixed that :-D

profiling/src/wall_time.rs
```diff
@@ -137,6 +137,94 @@ pub extern "C" fn ddog_php_prof_interrupt_function(execute_data: *mut zend_execu
...
+                ; mov x16, original as u64
+                ; blr x16
+                ; mov x16, interrupt_addr as u64
+                ; br x16  // tail call
+            );
+            #[cfg(target_arch = "x86_64")]
+            dynasm!(assembler
+                ; mov rax, QWORD original as i64
+                ; call rax
+                ; mov rax, QWORD interrupt_addr as i64
```
cataphract: I assume `original` is not returning anything, because you're writing over RAX.

profiling/src/wall_time.rs
```diff
@@ -137,6 +137,94 @@ pub extern "C" fn ddog_php_prof_interrupt_function(execute_data: *mut zend_execu
...
+        use crate::zend;
+
+        pub unsafe fn generate_wrapper(original: *mut c_void) -> *mut c_void {
+            // Calls original function, then calls interrupt function.
+            let mut assembler = Assembler::new().unwrap();
+            let interrupt_addr = ddog_php_prof_icall_trampoline_target as *const ();
+            #[cfg(target_arch = "aarch64")]
+            dynasm!(assembler
+                ; mov x16, original as u64
+                ; blr x16
```
cataphract: this overwrites x30/lr, aren't you gonna lose the original return location of the handler? so when `br x16` returns, it goes back to calling interrupt_addr

profiling/src/wall_time.rs
bwoebi: Correct

profiling/src/wall_time.rs
```diff
@@ -137,6 +137,103 @@ pub extern "C" fn ddog_php_prof_interrupt_function(execute_data: *mut zend_execu
...
+                #[cfg(target_arch = "aarch64")]
+                dynasm!(assembler
+                    ; mov x16, *orig as u64
+                    ; blr x16
+                    ; mov x16, interrupt_addr as u64
+                    ; br x16  // tail call
+                );
+                #[cfg(target_arch = "x86_64")]
+                dynasm!(assembler
+                    ; mov rax, QWORD *orig as i64
```
cataphract, line 209: I think you also need a `sub rsp, 8` to align the stack to 16 bytes and satisfy the abi requirements

profiling/src/wall_time.rs
bwoebi: Ah, right. `call` on x86_64 pushes %eip to the stack, but `blr` doesn't.

profiling/src/wall_time.rs
bwoebi: Right, compilers often handle that by just pushing rbp anyway. Will do.

bwoebi: I need to actually push and pop to restore rsp too, for the final jump.

profiling/src/wall_time.rs
bwoebi: Fixed up, is it correct now?

realFlowControl (approved):
Thanks @bwoebi this is awesome!

profiling/src/wall_time.rs
```diff
@@ -137,6 +137,169 @@ pub extern "C" fn ddog_php_prof_interrupt_function(execute_data: *mut zend_execu
...
+                }
+                #[cfg(target_arch = "x86_64")]
+                dynasm!(assembler
+                    ; push rbp  // align stack
+                    ; mov rax, QWORD *orig as i64
+                    ; call rax
+                    ; pop rbp  // restore stack
+                    ; mov rax, QWORD interrupt_addr as i64
+                    ; jmp rax  // tail call
+                );
```
morrisonlevi, line 214: Regarding this, claude said:

> Potential Issue — frame pointer chain is broken:
> push rbp without mov rbp, rsp saves rbp onto the stack but never updates rbp to point at the new frame. The frame pointer chain has a hole here. DWARF-based unwinding (the normal case) still works fine; only --call-graph fp tooling (e.g. perf record -g fp) would miss this frame. For an observability tool, this is ironic but low-severity. The fix is to add mov rbp, rsp after push rbp.

Claude thinks the aarch64 version keeps a valid frame record.

Based on this info, and if this seems right to you, we should probably update it.

profiling/src/wall_time.rs
```diff
@@ -137,6 +137,169 @@ pub extern "C" fn ddog_php_prof_interrupt_function(execute_data: *mut zend_execu
...
+                        ; stp x29, x30, [sp, -16]! // save link register and allow clobber of x29
+                        ; mov x29, sp // store stack pointer
+                        ; ldr x16, >orig_label
+                        ; blr x16
+                        ; ldp x29, x30, [sp], 16 // restore link register and x29
+                        ; ldr x16, >interrupt_label
+                        ; br x16  // tail call
+                        ; orig_label: ; .qword *orig as i64
+                    );
+                }
```
morrisonlevi, line 205: Claude thinks that there's potential misalignment. They said it probably won't crash in tests because of the defaults of `SCTLR_EL1.A = 0` for both Linux and macOS. I don't see a downside to adding the `.align`, do you?

```
  // Inside the loop — add .align 8 before orig_label                                                                                  
  dynasm!(assembler                                                                                                                    
      ; stp x29, x30, [sp, -16]!                                                                                                       
      ; mov x29, sp                                                                                                                    
      ; ldr x16, >orig_label                                                                                                           
      ; blr x16
      ; ldp x29, x30, [sp], 16                                                                                                         
      ; ldr x16, >interrupt_label                                                                                                      
      ; br x16  // tail call                                                                                                           
      ; .align 8                                                                                                                       
      ; orig_label: ; .qword *orig as i64                                                                                              
  );              

  // After the loop — add .align 8 before interrupt_label                                                                              
  #[cfg(target_arch = "aarch64")]
  dynasm!(assembler                                                                                                                    
      ; .align 8  
      ; interrupt_label: ; .qword interrupt_addr as i64
  );
```

bwoebi: Feel free to add it. Shouldn't do harm.

profiling/src/wall_time.rs
bwoebi: Yeah, it's only relevant if you want to make the frame visible, which we aren't really interested in. We should skip this, it's just overhead. ... I mean the point of flf is to be fast, right (it's on the hot path).

realFlowControl: Jup, I'd say the exact same.

realFlowControl (approved):

PR #3607, opened 2026-01-29
build(profiling): remove once_cell as a dependency

morrisonlevi:
### Description

Replaces once_cell types with ones in the std library (since 1.70). In one case it removes a `Lazy` instead and handles the non-const parts by writing them in `get_module`. Note that `OnceCell` in the stdlib isn't `Sync` so `once_cell::sync::OnceCell` was replaced by `std::sync::OnceLock` in most cases.

My _real_ motivation is to use PHP globals for some things instead of Rust TLS variables. They have been problematic in the memory allocation profiler because of ext/grpc doing unsafe (and wrong, not just unsafe) things. While exploring that, I noticed this cleanup was possible. It's only net new lines of code because there are new comments.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/mod.rs
+++ b/profiling/src/profiling/mod.rs
@@ -31,12 +31,11 @@ use libdd_profiling::exporter::Tag;
 use libdd_profiling::internal::Profile as InternalProfile;
 use log::{debug, info, trace, warn};
 use std::borrow::Cow;
-use std::cell::OnceCell;
 use std::collections::HashMap;
 use std::hash::Hash;
 use std::num::NonZeroI64;
 use std::sync::atomic::{AtomicBool, AtomicPtr, AtomicU32, Ordering};
-use std::sync::{Arc, Barrier};
+use std::sync::{Arc, Barrier, OnceLock};
 use std::thread::JoinHandle;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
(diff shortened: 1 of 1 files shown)


profiling/src/profiling/mod.rs
realFlowControl: I think this should be a `std::sync::OnceLock` instead (became a `std::cell::OnceCell` due to the change above).

In every `RINIT` we call `Profiler::init()`:
https://github.com/DataDog/dd-trace-php/blob/cefbc5b59e252fd1f329c0cb3ac5224b6ec132f7/profiling/src/lib.rs#L704 where we rely on `.get_or_init()` of this `OnceCell`:
https://github.com/DataDog/dd-trace-php/blob/21d2f00064f089eb451df371b58fb056fd348c4c/profiling/src/profiling/mod.rs#L661-L666 which is not thread safe anymore (the comment became wrong). I believe that for ZTS PHP there is a race condition where when two threads "at the same time" go through `RINIT` for the first time, both may race in the `get_or_init()`.

profiling/src/profiling/thread_utils.rs
realFlowControl, line 93: Nice catch, this could have been a `std::cell::OnceCell` from the beginning.

realFlowControl (approved):

profiling/src/profiling/mod.rs
morrisonlevi: Yes, good catch 🫡

PR #3608, opened 2026-01-29
refactor(profiling): use module globals for ZMM state

morrisonlevi:
### Description

- Introduced `ProfilerGlobals` struct which for now just contains `zend_mm_state: Cell<ZendMMState>`.
- Created `module_globals.rs` to centralize PHP globals management.
    - Moved `GLOBALS` (NTS) into module_globals.rs.
    - Moved `GLOBALS_ID_PTR` (ZTS) into module_globals.rs and renamed it `GLOBALS_ID` (because it's not a pointer, it's the thing the pointer points at).
    - Moved `ginit()` and `gshutdown()` lifecycle functions into module_globals.rs, and gave them export names of `ddog_php_prof_{ginit,gshutdown)`.
- Updated `zend::ModuleEntry` to use `ProfilerGlobals` and pointers to module globals. Previously, this was faked just so we could observe the ginit/gshutdown phases but now we have real module globals.
- Moved some tls macros in the allocation profiler into `mod.rs` because they are the same for ge84 and le83.

### Motivation

In #3542 we worked around a compatibility issue with ext/grpc by mimicking the PHP module globals pattern. Why mimic it when we can use it?

For accessing module globals, we do have to translate the C macro `ZEND_MODULE_GLOBALS_BULK` to Rust, though. Here are its expansions:

```
(ZTS) ZEND_MODULE_GLOBALS_BULK(module_name)
→ TSRMG_BULK(module_name##_globals_id, zend_##module_name##_globals *)
→ ((zend_##module_name##_globals *) (*((void ***) tsrm_get_ls_cache()))[TSRM_UNSHUFFLE_RSRC_ID(module_name##_globals_id)])
→ ((zend_##module_name##_globals *) (*((void ***) tsrm_get_ls_cache()))[module_name##_globals_id - 1])

(NTS) ZEND_MODULE_GLOBALS_BULK(module_name)
→ (&module_name##_globals)
```

So for us, `module_name##_globals_id` is just `GLOBALS_ID` and `module_name##_globals` is just `GLOBALS`. The rest matches C directly.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/dockerfiles/verify_packages/installer/utils.sh
+++ b/dockerfiles/verify_packages/installer/utils.sh
@@ -116,7 +116,7 @@ assert_appsec_installed() {
 
 assert_appsec_enabled() {
     output="$(php --ri ddappsec)"
-    if [ -z "${output##*datadog.appsec.enabled => On*}" ]; then
+    if [ -z "${output##*Current state => Enabled*}" ]; then
         echo "---\nOk: ddappsec is enabled\n---\n${output}\n---\n"
     else
         echo "---\nError: Wrong ddappsec should be enabled\n---\n${output}\n---\n"
@@ -126,7 +126,7 @@ assert_appsec_enabled() {
 
 assert_appsec_disabled() {
     output="$(php --ri ddappsec)"
-    if [ -n "${output##*datadog.appsec.enabled => On*}" ]; then
+    if [ -n "${output##*Current state => Enabled*}" ]; then
         echo "---\nOk: ddappsec is not enabled\n---\n${output}\n---\n"
     else
         echo "---\nError: Wrong ddappsec should not be enabled\n---\n${output}\n---\n"


profiling/src/module_globals.rs
```diff
@@ -0,0 +1,111 @@
...
+/// - Must be called by PHP's module shutdown system.
+#[export_name = "ddog_php_prof_gshutdown"]
+pub unsafe extern "C" fn gshutdown(_globals_ptr: *mut c_void) {
+    #[cfg(php_zts)]
+    crate::timeline::timeline_gshutdown();
+
+    // TODO: Florian, do we need this?
+    // let globals = globals_ptr.cast::<ProfilerGlobals>();
+    // (*globals).zend_mm_state = ZendMMState::new();
+
```
realFlowControl, line 112:
```suggestion
```
I think we do not, we did not do so earlier also. Also we unregister our hooks in `RSHUTDOWN`, nothing in allocation profiling touches the global state after `RSHUTDOWN` and PHP itself takes care of freeing the memory, I guess it is fine the way it is.

profiling/src/module_globals.rs
```diff
@@ -0,0 +1,111 @@
...
+/// # Safety
+/// - Must be called by PHP's module initialization system.
+#[export_name = "ddog_php_prof_ginit"]
+pub unsafe extern "C" fn ginit(globals_ptr: *mut c_void) {
+    #[cfg(php_zts)]
+    crate::timeline::timeline_ginit();
+
+    // Initialize ZendMMState in PHP globals.
+    let globals = globals_ptr.cast::<ProfilerGlobals>();
+    (*globals).zend_mm_state = Cell::new(ZendMMState::new());
```
realFlowControl:
```suggestion
    #[cfg(php_zts)]
    {
        // Initialize ZendMMState in PHP globals.
        let globals = globals_ptr.cast::<ProfilerGlobals>();
        (*globals).zend_mm_state = Cell::new(ZendMMState::new());
    }
```
I think we only need to do this in case if PHP ZTS, because otherwise in NTS we would call the `ZendMMState::new()` twice, once above in the 
```rust
pub static mut GLOBALS: ProfilerGlobals = ProfilerGlobals {
    zend_mm_state: Cell::new(ZendMMState::new()),
};
```
and a second time when PHP calls `GINIT`. I think this is no problem after all, just weird doing it twice.

realFlowControl (approved):
We should (in a follow up PR) move all the other "global state" we have in other modules to PHP globals, but so far this is looking good.
Can you do a short run with DoE? I'd not expect this to show any performance problems, but just to be sure and I've added two comments.

PR #3609, opened 2026-01-30
feat(profiling): process_tags on uploader

dubloom:
### Description

This PR implements this [RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?pli=1&tab=t.0#heading=h.s9l1lctqlg11) for Profiling by adding a process tags field to the Uploader from the tracer.

Note that if process_tags are deactivated, the value passed to the profiler will be an empty string. 

Here are an example in datadog: 
image
(see entrypoint.basedir/entrypoint.name fields).

No tests were added at it does not seem to be possible right now (because of the interaction tracer <> profiler)

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/src/profiling/mod.rs
```diff
@@ -716,12 +718,24 @@ impl Profiler {
...
+        // SAFETY: this is set to a noop version if ddtrace wasn't found, and
+        // we're getting the process tags of a PHP thread
+        let process_tags: Option<String> = unsafe {
+            let raw_ptr = datadog_php_profiling_get_process_tags_serialized.unwrap_unchecked()();
+            zai_str_from_zstr(raw_ptr.as_mut())
+                .into_utf8()
+                .ok()
+                .filter(|s| !s.is_empty())
+                .map(|s| s.to_owned())
+        };
```
morrisonlevi, line 732: If the string isn't UTF-8, then this will attach nothing. Should we convert to lossy UTF-8 instead?

dubloom: Tag key are set by us, so we know is utf-8
Tag value are transformed by this method: https://github.com/DataDog/dd-trace-php/blob/f1d36ca6a29e9642175c37323f1e74da2d199028/components-rs/lib.rs#L139.

with the first line being ` let value = tag_value.to_utf8_lossy();`

Therefore (tell me if I'm wrong) but we know the process_tags will be utf-8. If you feel this is better anyway, I can use lossy UTF-8

morrisonlevi: This is okay.

PR #3612, opened 2026-01-31
refactor(profiling): extract Backtrace type

morrisonlevi:
### Description

In a future change, this may hold a refcount for another object, so we need to encapsulate it. The plan is for holding a `ProfilesDictionary` and sending data from it.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/allocation/allocation_ge84.rs
+++ b/profiling/src/allocation/allocation_ge84.rs
@@ -1,7 +1,7 @@
 use crate::allocation::{allocation_profiling_stats_should_collect, collect_allocation};
 use crate::bindings as zend;
 use crate::PROFILER_NAME;
-use core::{cell::Cell, ptr};
+use core::ptr;
 use lazy_static::lazy_static;
 use libc::{c_char, c_int, c_void, size_t};
 use log::{debug, trace, warn};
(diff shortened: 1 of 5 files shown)


realFlowControl:
The `test_collect_stack_sample` test was [not executed](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-php/-/jobs/1391689508), due to a wrong `cfg` attribute:
```
#[cfg(stack_walking_tests)]
// vs.
#[cfg(feature = "stack_walking_tests")]
```
I [fixed that](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-php/-/jobs/1392645730) and also added a stub for the `zend_flf_functions` (with https://github.com/DataDog/dd-trace-php/pull/3612/commits/893a3b433c5cc917133ffbe7ea102da9261433e9) which does not exist in the scope of `cargo test --all-features` as it gets executed in CI .

profiling/src/profiling/stack_walking.rs
```diff
@@ -550,17 +551,18 @@ mod tests {
...
+            assert_eq!(frames[1].function, "function name 002");
+            assert_eq!(frames[1].file, Some("filename-002.php".into()));
+            assert_eq!(frames[1].line, 0);
 
-            assert_eq!(stack[2].function, "function name 001");
-            assert_eq!(stack[2].file, Some("filename-001.php".into()));
-            assert_eq!(stack[2].line, 0);
+            assert_eq!(frames[2].function, "function name 001");
+            assert_eq!(frames[2].file, Some("filename-001.php".into()));
+            assert_eq!(frames[2].line, 0);
```
realFlowControl, line 565: I think that because of the `Deref` implementation for `Backtrace` we might not need these changes, but technically it is fine either way.

PR #3623, opened 2026-02-04
feat(profiling): add heap-live profiling for memory leak detection

realFlowControl:
### Description

Track allocations that survive across profile exports using `heap-live-samples` and `heap-live-size` sample types. Samples are emitted in batches at export time.

Enable via `DD_PROFILING_EXPERIMENTAL_HEAP_LIVE_ENABLED` or `datadog.profiling.experimental_heap_live_enabled` (default disabled), only works when allocation profiling is active.

> [!NOTE]
> This feature is default disabled and if you are running a SAPI like PHP-FPM or mod_php (Apache) that is purely request/response based, you do not get any value out of this. This feature is meant for long running process (CLI workers consuming messages, ReactPHP applications, FrankenPHP Worker Mode, ...).

### Reviewer checklist
- [ ] Test coverage seems ok.

PROF-13688


profiling/src/allocation/mod.rs
```diff
@@ -14,7 +14,17 @@ use log::{debug, trace};
...
+
+/// Fast-path gate for the heap-live tracking code paths.
+///
+/// Mirrors `SystemSettings::profiling_experimental_heap_live_enabled` so the alloc/free/
+/// realloc handlers can short-circuit with a single relaxed atomic load,
+/// avoiding the `Profiler::get()` `OnceLock` load on every `efree`/`erealloc`
+/// when heap-live profiling is disabled (which is the default).
+///
+/// Set in `Profiler::new` and cleared in `Profiler::stop` / `Profiler::kill`.
+pub(crate) static HEAP_LIVE_ENABLED: AtomicBool = AtomicBool::new(false);
```
morrisonlevi: This is memory-model safe but not semantically correct for ZTS builds. At least, not generally. Theoretically we could make it be settable in minit only and then it's shared across all threads as the same value, and then this would be okay.

profiling/src/allocation/mod.rs
```diff
@@ -156,6 +162,15 @@ pub fn collect_allocation(len: size_t) {
     }
 }
 
+/// Called when memory is freed. If this pointer was tracked for live heap,
+/// sends the deallocation sample to cancel out the original allocation.
+#[inline]
+pub fn free_allocation(ptr: *mut c_void) {
```
morrisonlevi: Could we use track/untrack terminology instead? It's a bit more clear when you `untrack_allocation` that it's not actually freeing it when compared to `free_allocation`.

morrisonlevi (commented):
What's the expectation for heap profiles and correlation to traces, spans, and endpoints? Should we even be gathering the local root span id/span id for these?

realFlowControl:
> What's the expectation for heap profiles and correlation to traces, spans, and endpoints? Should we even be gathering the local root span id/span id for these?

I think we should yes. The Heap Live profiles show what triggered that allocation. Also: we are sharing the sample message with the allocation profiler, so once all that information is collected, we already have it anyway (it would be extra work to remove that). What would be nice is if in the backend we could go from a stack trace to the span in the tracer (it is possible the other way round, just not back), but I do not know if that is too much important at all.

So TLDR: I am in favor of keeping that information, also because it is consistent with the allocation profiler.

morrisonlevi (approved):
Code wise, it's looking okay to me. I've still not played much with its functionality. However, it's off by default so at this stage I'm willing to approve.

realFlowControl:
@ajohnston-knak [version `1.22.0`](https://github.com/DataDog/dd-trace-php/releases/tag/1.22.0) has been released bringing a public preview of heap live profiling for PHP.

PR #3627, opened 2026-02-06
Adds process_tags to telemetry payload

dubloom:
This PR implements this [RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?pli=1&tab=t.0#heading=h.s9l1lctqlg11) for telemetry.

Add process_tags to process_tags payload. We should be able to set process_tags when configuring the sidecar (for sidecar reconnect) and we also add a function to be able to set them just after they are acquired

This PR comes with this libdatadog PR: https://github.com/DataDog/libdatadog/pull/1459.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2795,7 +2795,7 @@ dependencies = [
  "libc 0.2.177",
  "maplit",
  "mime 0.3.17",
- "multipart",
+ "multer",
  "nix 0.29.0",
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
Looks good to me, thanks!

PR #3630, opened 2026-02-07
feat(ffe): PHP FFE implementation

leoromanovsky:
## Motivation

Add FFE (Feature Flags & Experimentation) support to dd-trace-php. PHP applications can evaluate feature flags delivered via Remote Config using the same `datadog-ffe` Rust engine used by Ruby and Python. Per the [RFC \"Flag evaluations tracking for APM tracers\"](https://docs.google.com/document/d/1yOdu6FU3Fw-PMhNlSfqawJ5r6f3tSGomjlJVwdtqpZc), we also emit a `feature_flag.evaluations` OTel counter metric on each evaluation so flag usage can be tracked in Datadog.

### Core FFE Implementation
- **Rust FFI layer** (`components-rs/ffe.rs`): C-callable bridge to `datadog-ffe::rules_based` — config store, evaluate, result accessors
- **C extension** (`ext/ddtrace.c`): `ffe_evaluate`, `ffe_has_config`, `ffe_config_changed`, `ffe_load_config` internal functions that marshal PHP arrays to `FfeAttribute` structs
- **Remote Config** (`components-rs/remote_config.rs`): Register `FfeFlags` product + `FfeFlagConfigurationRules` capability; handle add/remove of FFE configs
- **PHP Provider** (`src/DDTrace/FeatureFlags/Provider.php`): Singleton that checks RC config state, calls native evaluate, parses JSON results, reports exposures; enforces `reason=ERROR` for any non-zero error code per OpenFeature spec
- **Exposure pipeline**: LRU dedup cache (65K entries) + batched writer to `/evp_proxy/v2/api/v2/exposures` (1000 event buffer cap)
- **OpenFeature adapter** (`src/DDTrace/OpenFeature/DataDogProvider.php`): Implements `AbstractProvider` for the `open-feature/sdk` composer package
- **Build**: Add `datadog-ffe` to `RUST_FILES` in Makefile for PECL packaging
- **Tests**: LRU cache unit tests, exposure cache unit tests, 220 evaluation correctness tests from JSON fixtures
- **Config**: `DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED` gating via X-macro in `ext/configuration.h`

### OTel Flag Evaluation Metrics (new)
- **`src/DDTrace/FeatureFlags/FlagEvalMetrics.php`**: OTel counter hook that emits `feature_flag.evaluations` (count=1) after each evaluation. Tags: `feature_flag.key`, `feature_flag.provider.name=datadog`, `feature_flag.result.variant`, `feature_flag.result.reason`, `feature_flag.result.allocation_key`, and on error: `error.type` (`type_mismatch` / `parse_error` / `flag_not_found`).
- **Error code mapping corrected**: `errorCodeToTag()` now maps Rust constants correctly: `1=ERROR_TYPE_MISMATCH→type_mismatch`, `2=ERROR_CONFIG_PARSE→parse_error`, `3=ERROR_FLAG_UNRECOGNIZED→flag_not_found` (was transposed in initial implementation).
- **`Provider.php`**: Forces `reason=ERROR` when `error_code != 0`. The Rust layer returns `REASON_DEFAULT` for `FlagUnrecognizedOrDisabled` but the OpenFeature spec requires `ERROR` to be surfaced to callers.

## Decisions

**Evaluation in Rust, not PHP.** All flag evaluation (UFC parsing, targeting rules, shard hashing, allocation resolution) happens in `libdatadog`'s `datadog-ffe` crate via FFI. PHP only handles orchestration (config lifecycle, exposure dedup, HTTP transport). This matches Ruby and Python — no language re-implements evaluation logic.

**Global config behind `Mutex<FfeState>`.** The Rust FFE config is stored in a `lazy_static` global with a `Mutex`. PHP is single-threaded per process, so `RwLock` would be unnecessary complexity.

**Reuses existing RC pipeline.** FFE configs flow through the same sidecar → `ddog_process_remote_configs()` path as APM Tracing and Live Debugger. No new polling mechanism.

**Structured attributes, not JSON blobs.** The C extension converts PHP arrays into `FfeAttribute` structs (typed: string/number/bool) before calling Rust, avoiding JSON encode/decode overhead on the hot path.

**ExposureWriter caps at 1000 events per request.** Matches Ruby and Python. Flush via `register_shutdown_function`.

**FlagEvalMetrics via OTel SDK.** Uses `open-telemetry/api` + OTLP exporter. Recorded after the type-specific evaluation method returns (not inside core `evaluate()`) so type mismatch errors are captured with `reason=error` rather than the intermediate `targeting_match`.

**ERROR reason forced for non-zero error codes.** `FlagUnrecognizedOrDisabled` returns `REASON_DEFAULT` from Rust (the default value was served) but the OpenFeature spec requires `reason=ERROR` whenever an error code is set. `Provider.php` overrides the Rust-reported reason in this case.

### Unit tests
```
Unit tests (local, no extension): 243 tests, 75 assertions, 1 skipped
Unit tests (Docker with extension): 243 tests, 735 assertions — all pass
```

### System tests — all FFE tests pass (PHP 8.0-apache)
```
Scenario: FEATURE_FLAGGING_AND_EXPERIMENTATION
Library: php@8.0-apache

tests/ffe/test_dynamic_evaluation.py ..                                  [ 11%]
tests/ffe/test_exposures.py ...........                                  [ 76%]
tests/ffe/test_flag_eval_metrics.py .....                                [100%]

=============== 17 passed, 1 xfailed in 0:04:12 ================
```

The 1 xfail is `test_cross_request_dedup` — expected because PHP's shared-nothing architecture prevents cross-request LRU state.

### OTel metrics tests (all 5 pass)
```
PASSED Test_FFE_Eval_Metric_Basic        — reason=static, variant=on, allocation_key=default-allocation
PASSED Test_FFE_Eval_Metric_Count        — 5 evaluations → metric count ≥ 5
PASSED Test_FFE_Eval_Metric_Different_Flags — two flags → two metric series
PASSED Test_FFE_Eval_Metric_Error        — reason=error, error.type=flag_not_found
PASSED Test_FFE_Eval_Metric_Type_Mismatch — reason=error, error.type=type_mismatch
```

## LOC (excluding Cargo.lock + JSON fixtures)
| Category | Lines |
|---|---|
| Rust FFI | 323 |
| C extension | 126 |
| PHP source (FFE + metrics) | 973 |
| PHP tests | 559 |
| Build/wiring | 8 |
| **Total** | **1,989** |

## Companion PRs
- **libdatadog**: [DataDog/libdatadog#1532](https://github.com/DataDog/libdatadog/pull/1532)
- **system-tests**: [DataDog/system-tests#6410](https://github.com/DataDog/system-tests/pull/6410)
- **ffe-dogfooding**: [DataDog/ffe-dogfooding#40](https://github.com/DataDog/ffe-dogfooding/pull/40)


components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+/// Holds both the FFE configuration and a "changed" flag atomically behind a
+/// single Mutex. This avoids the race where another thread could observe
+/// `config` updated but `changed` still false (or vice-versa).
+///
+/// A `RwLock` would be more appropriate here (many readers via `ddog_ffe_evaluate`,
+/// rare writer via `store_config`), but PHP is single-threaded per process so
+/// contention is not a practical concern. Keeping a Mutex for simplicity.
+struct FfeState {
+    config: Option<Configuration>,
+    changed: bool,
```
dd-oleksii: minor: not sure how `changed` is used by the other side but having a global `changed` flag seems error-prone. What do you think about replacing it with `version: u64`?

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+/// Check if FFE config has changed since last check.
+/// Resets the changed flag after reading.
+#[no_mangle]
+pub extern "C" fn ddog_ffe_config_changed() -> bool {
+    if let Ok(mut state) = FFE_STATE.lock() {
+        let was_changed = state.changed;
+        state.changed = false;
+        was_changed
```
dd-oleksii: nitpick: not necessarily better but there's a function for this:
```suggestion
        std::mem::replace(&mut state.changed, false)
```

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+const TYPE_OBJECT: i32 = 4;
+
+/// Opaque handle for FFE evaluation results returned to C/PHP.
+pub struct FfeResult {
+    pub value_json: CString,
+    pub variant: Option<CString>,
+    pub allocation_key: Option<CString>,
+    pub reason: i32,
+    pub error_code: i32,
+    pub do_log: bool,
```
dd-oleksii, line 137: minor: shall fields be non-public if it's supposed to be opaque?
```suggestion
    value_json: CString,
    variant: Option<CString>,
    allocation_key: Option<CString>,
    reason: i32,
    error_code: i32,
    do_log: bool,
```

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+#[no_mangle]
+pub extern "C" fn ddog_ffe_evaluate(
+    flag_key: *const c_char,
+    expected_type: i32,
+    targeting_key: *const c_char,
+    attributes: *const FfeAttribute,
+    attributes_count: usize,
+) -> *mut FfeResult {
+    if flag_key.is_null() {
+        return std::ptr::null_mut();
```
dd-oleksii, line 166: minor: consider returning an error result on errors?

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+        TYPE_STRING => ExpectedFlagType::String,
+        TYPE_INTEGER => ExpectedFlagType::Integer,
+        TYPE_FLOAT => ExpectedFlagType::Float,
+        TYPE_BOOLEAN => ExpectedFlagType::Boolean,
+        TYPE_OBJECT => ExpectedFlagType::Object,
+        _ => return std::ptr::null_mut(),
+    };
+
+    // Build targeting key
+    let tk = if targeting_key.is_null() {
```
dd-oleksii, line 183: nitpick:
```suggestion
    let targeting_key = if targeting_key.is_null() {
```

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+        None
+    } else {
+        match unsafe { CStr::from_ptr(targeting_key) }.to_str() {
+            Ok(s) if !s.is_empty() => Some(Str::from(s)),
+            _ => None,
+        }
+    };
+
+    // Build attributes map from the C array
+    let mut attrs = HashMap::new();
```
dd-oleksii, line 193: minor: 
```suggestion
    let mut attrs = HashMap::with_capacity(attributes_count);
```

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+            reason: match a.reason {
+                AssignmentReason::Static => REASON_STATIC,
+                AssignmentReason::TargetingMatch => REASON_TARGETING_MATCH,
+                AssignmentReason::Split => REASON_SPLIT,
+            },
+            error_code: ERROR_NONE,
+            do_log: a.do_log,
+        },
+        Err(err) => {
+            let (error_code, reason) = match &err {
```
dd-oleksii, line 251: nitpick: `reason, error_code` would be a more canonical order
```rust
let (reason, error_code) = match &err {
```

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+                EvaluationError::TypeMismatch { .. } => (ERROR_TYPE_MISMATCH, REASON_ERROR),
+                EvaluationError::ConfigurationParseError => (ERROR_CONFIG_PARSE, REASON_ERROR),
+                EvaluationError::ConfigurationMissing => (ERROR_CONFIG_MISSING, REASON_ERROR),
+                EvaluationError::FlagUnrecognizedOrDisabled => (ERROR_FLAG_UNRECOGNIZED, REASON_DEFAULT),
+                EvaluationError::FlagDisabled => (ERROR_NONE, REASON_DISABLED),
+                EvaluationError::DefaultAllocationNull => (ERROR_NONE, REASON_DEFAULT),
+                _ => (ERROR_GENERAL, REASON_ERROR),
+            };
+            FfeResult {
+                value_json: CString::new("null").unwrap_or_default(),
```
dd-oleksii, line 263: minor: shall it be `std::ptr::null()`? From `ddog_ffe_result_value`, it looks like NULL is a valid value here

components-rs/ddtrace.h
```diff
@@ -61,14 +61,46 @@ uint32_t ddog_get_logs_count(ddog_CharSlice level);
...
+bool ddog_ffe_config_changed(void);
+
+struct FfeResult;
+
+struct FfeAttribute {
+    const char *key;
+    int32_t value_type;       /* 0=string, 1=number, 2=bool */
+    const char *string_value;
+    double number_value;
+    bool bool_value;
```
dd-oleksii: nitpick: can shave some 8 bytes from the struct by reordering fields:
```suggestion
    uint8_t value_type;       /* 0=string, 1=number, 2=bool */
    bool bool_value;
    const char *string_value;
    double number_value;
```
or sort fields from bigger to smaller.

Or better yet, use union:
```c
struct FfeAttribute {
    const char *key;
    uint8_t value_type;
    union {
        const char *string_value;
        double number_value;
        bool bool_value;
    } value;
};
```

components-rs/remote_config.rs
```diff
@@ -101,6 +102,7 @@ pub unsafe extern "C" fn ddog_init_remote_config(
     live_debugging_enabled: bool,
     appsec_activation: bool,
     appsec_config: bool,
+    ffe_enabled: bool,
```
dd-oleksii: minor: unrelated to the core functionality but having a function taking 4 bools is a bit unwieldy. Consider passing a struct, so the caller may use designated initializers (it allows naming fields)

ext/ddtrace.c
```diff
@@ -3008,6 +3009,105 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
+                    array_init(return_value);
+                    val = ddog_ffe_result_value(result);
+                    var = ddog_ffe_result_variant(result);
+                    ak = ddog_ffe_result_allocation_key(result);
+                    if (val) { add_assoc_string(return_value, "value_json", (char *)val); }
+                    else { add_assoc_null(return_value, "value_json"); }
+                    if (var) { add_assoc_string(return_value, "variant", (char *)var); }
+                    else { add_assoc_null(return_value, "variant"); }
+                    if (ak) { add_assoc_string(return_value, "allocation_key", (char *)ak); }
+                    else { add_assoc_null(return_value, "allocation_key"); }
```
dd-oleksii, line 3291: major: why const casts are needed here? is it possible that something modifies our strings?

src/DDTrace/FeatureFlags/ExposureCache.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Parse a composite value string back into [variantKey, allocationKey].
+     */
+    private static function parseValue($value)
+    {
+        $colonPos = strpos($value, ':');
+        if ($colonPos === false) {
+            return [$value, ''];
+        }
```
dd-oleksii: this looks like it should be an error / invalid entry

src/DDTrace/FeatureFlags/ExposureCache.php
```diff
@@ -0,0 +1,117 @@
...
+     */
+    public function add($flagKey, $subjectId, $variantKey, $allocationKey)
+    {
+        $key = self::makeKey($flagKey, $subjectId);
+        $newValue = self::makeValue($variantKey, $allocationKey);
+
+        // Always put (updates LRU position even for duplicates)
+        $oldValue = $this->cache->put($key, $newValue);
+
+        return $oldValue === null || $oldValue !== $newValue;
```
dd-oleksii: nitpick: `null` check is redundant?
```suggestion
        return $oldValue !== $newValue;
```

src/DDTrace/FeatureFlags/ExposureWriter.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Add an exposure event to the buffer.
+     *
+     * @param array $event Exposure event data
+     */
+    public function enqueue(array $event)
+    {
+        if (count($this->buffer) >= self::MAX_BUFFER_SIZE) {
+            $this->droppedEvents++;
+            return;
```
dd-oleksii: minor: shall this trigger a flush?

src/DDTrace/FeatureFlags/ExposureWriter.php
```diff
@@ -0,0 +1,197 @@
...
+        $body = json_encode($payload);
+
+        if (!function_exists('curl_init')) {
+            return;
+        }
+
+        $ch = curl_init($url);
+        if ($ch === false) {
+            return;
+        }
```
dd-oleksii: nitpick: shall we check this earlier? perhaps in constructor — doesn't make sense to collect events if we can't send them

src/DDTrace/FeatureFlags/ExposureWriter.php
```diff
@@ -0,0 +1,197 @@
...
+     */
+    public static function buildEvent(
+        $flagKey,
+        $variantKey,
+        $allocationKey,
+        $targetingKey = null,
+        array $attributes = []
+    ) {
+        return [
+            'timestamp' => (int)(microtime(true) * 1000),
```
dd-oleksii: nitpick: can we pass the timestamp from evaluation into here? otherwise, we might see some confusing exposure events (e.g. timestamp that should have been filtered out by rules)

src/DDTrace/FeatureFlags/ExposureWriter.php
```diff
@@ -0,0 +1,197 @@
...
+     * @return string
+     */
+    private function resolveAgentUrl()
+    {
+        $agentUrl = $this->getConfigValue('DD_TRACE_AGENT_URL', '');
+        if ($agentUrl !== '') {
+            return rtrim($agentUrl, '/');
+        }
+
+        $host = $this->getConfigValue('DD_AGENT_HOST', 'localhost');
```
dd-oleksii: just checking — is this name correct? The other two start with `DD_TRACE_AGENT` and this one is `DD_AGENT`

src/DDTrace/FeatureFlags/Provider.php
```diff
@@ -0,0 +1,301 @@
...
+        5 => 'ERROR',
+    ];
+
+    private static $TYPE_MAP = [
+        'STRING'  => 0,
+        'INTEGER' => 1,
+        'NUMERIC' => 2,
+        'BOOLEAN' => 3,
+        'JSON'    => 4,
+    ];
```
dd-oleksii: minor: does it make sense to keep these translations in C, so we can use enums vs. having to sync maps manually?

src/DDTrace/FeatureFlags/Provider.php
```diff
@@ -0,0 +1,301 @@
...
+    private $shutdownRegistered = false;
+
+    /** @var Provider|null */
+    private static $instance = null;
+
+    public function __construct()
+    {
+        $this->writer = new ExposureWriter();
+        $this->exposureCache = new ExposureCache(65536);
+        $this->enabled = $this->isFeatureFlagEnabled();
```
dd-oleksii: nitpick: misleading name for the function

src/DDTrace/FeatureFlags/Provider.php
```diff
@@ -0,0 +1,301 @@
...
+namespace DDTrace\FeatureFlags;
+
+/**
+ * Datadog Feature Flags and Experimentation (FFE) Provider.
+ *
+ * Evaluates feature flags using the native datadog-ffe engine from libdatadog.
+ * Flag configurations are received via Remote Config through the sidecar.
+ * Reports exposure events to the EVP proxy for analytics.
+ */
+class Provider
```
dd-oleksii: Doesn't look like it actually implements openfeature provider, so the name is misleading

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -0,0 +1,151 @@
...
+        string $flagKey,
+        string $variationType,
+        $defaultValue,
+        ?EvaluationContext $context = null
+    ): ResolutionDetailsInterface {
+        $targetingKey = '';
+        $attributes = [];
+
+        if ($context !== null) {
+            $targetingKey = $context->getTargetingKey() ?? '';
```
dd-oleksii: major: don't default targeting key to empty string. Default value should be `null`

dd-oleksii (approved):
Looks good overall. I think the only major piece is that we shouldn't default targeting key to empty string. oh, and a suspicious const-cast that probably needs some comments/justification

Not sure if that would speed up reviews or not but this PR seems pretty splittable:
- general setup returning default values, OF integration, public API
- evaluation, rust pieces
- exposures, lru cache, etc.
- evaluation metrics

ext/ddtrace.c
```diff
@@ -3008,6 +3009,105 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
+        } else if (FUNCTION_NAME_MATCHES("ffe_has_config")) {
+            RETVAL_BOOL(ddog_ffe_has_config());
+        } else if (FUNCTION_NAME_MATCHES("ffe_config_changed")) {
+            RETVAL_BOOL(ddog_ffe_config_changed());
+        } else if (params_count == 1 && FUNCTION_NAME_MATCHES("ffe_load_config")) {
+            zval *json_zv = ZVAL_VARARG_PARAM(params, 0);
+            if (Z_TYPE_P(json_zv) == IS_STRING) {
+                RETVAL_BOOL(ddog_ffe_load_config(Z_STRVAL_P(json_zv)));
+            }
+        } else if (FUNCTION_NAME_MATCHES("ffe_evaluate") && params_count >= 4) {
```
bwoebi, line 3212: Please add proper documented functions on ddtrace.stub.php for this.

dd_trace_internal_fn is generally meant for testing purposes.

ext/ddtrace.c
bwoebi: `const` was missing on the PHP 7.0 and 7.1 versions - it doesn't touch the string. This makes the API accept it.

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+const TYPE_INTEGER: i32 = 1;
+const TYPE_FLOAT: i32 = 2;
+const TYPE_BOOLEAN: i32 = 3;
+const TYPE_OBJECT: i32 = 4;
+
+/// Opaque handle for FFE evaluation results returned to C/PHP.
+pub struct FfeResult {
+    pub value_json: CString,
+    pub variant: Option<CString>,
+    pub allocation_key: Option<CString>,
```
bwoebi, line 134: Store it as `String`, instead of as CString. This simply costs you an extra conversion for no gain.
There's `libdd_common_ffi::CharSlice` to return these (not-null-byte-delimited strings) to PHP then instead of `*mut c_char`.

ext/ddtrace.c
```diff
@@ -3008,6 +3009,105 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
+                    targeting_key = Z_STRVAL_P(targeting_key_zv);
+                }
+                if (Z_TYPE_P(attrs_zv) == IS_ARRAY) {
+                    HashTable *ht = Z_ARRVAL_P(attrs_zv);
+                    attrs_count = zend_hash_num_elements(ht);
+                    if (attrs_count > 0) {
+                        size_t idx = 0;
+                        zend_string *key;
+                        zval *val;
+                        c_attrs = ecalloc(attrs_count, sizeof(struct FfeAttribute));
```
bwoebi:
```suggestion
                        c_attrs = emalloc(attrs_count * sizeof(struct FfeAttribute));
```
No reason for zeroing here.

ext/ddtrace.c
```diff
@@ -3008,6 +3009,105 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
+                    efree(c_attrs);
+                }
+                if (result) {
+                    const char *val;
+                    const char *var;
+                    const char *ak;
+                    array_init(return_value);
+                    val = ddog_ffe_result_value(result);
+                    var = ddog_ffe_result_variant(result);
+                    ak = ddog_ffe_result_allocation_key(result);
```
bwoebi, line 3285:
```suggestion
                    array_init(return_value);
                    const char *val = ddog_ffe_result_value(result);
                    const char *var = ddog_ffe_result_variant(result);
                    const char *ak = ddog_ffe_result_allocation_key(result);
```
nit

ext/ddtrace.c
bwoebi, line 3291: We don't write our ifs here this way, could you please write it as:

```
if () {
...
} else {
 ...
}
```

ext/ddtrace.c
```diff
@@ -3008,6 +3009,105 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
+                RETVAL_BOOL(ddog_ffe_load_config(Z_STRVAL_P(json_zv)));
+            }
+        } else if (FUNCTION_NAME_MATCHES("ffe_evaluate") && params_count >= 4) {
+            /* ffe_evaluate(flag_key, type_id, targeting_key, attributes) */
+            zval *flag_key_zv = ZVAL_VARARG_PARAM(params, 0);
+            zval *type_zv = ZVAL_VARARG_PARAM(params, 1);
+            zval *targeting_key_zv = ZVAL_VARARG_PARAM(params, 2);
+            zval *attrs_zv = ZVAL_VARARG_PARAM(params, 3);
+            if (Z_TYPE_P(flag_key_zv) == IS_STRING) {
+                /* Declare all variables at top of block for C89/MSVC compatibility */
```
bwoebi, line 3219: We're at C99. (and MSVC supports that). No need.

src/DDTrace/FeatureFlags/ExposureWriter.php
```diff
@@ -0,0 +1,197 @@
...
+    {
+        if (count($this->buffer) >= self::MAX_BUFFER_SIZE) {
+            $this->droppedEvents++;
+            return;
+        }
+        $this->buffer[] = $event;
+    }
+
+    /**
+     * Send all buffered exposure events as a batch to the EVP proxy.
```
bwoebi: Can you route this through the extension all the way back to the sidecar, which shall take care of sending this?

The PHP tracer does not do requests by itself on the main thread (especially not "fire-and-forget" requests).

This will manifest as pure latency on every request on the user side. Very bad thing :-D

Then test_cross_request_dedup would also succeed...

bwoebi: I.e. have ffe_evaluate handle this directly in the rust code. This whole ExposureWriter has no business being in the PHP code.

src/DDTrace/FeatureFlags/FlagEvalMetrics.php
```diff
@@ -0,0 +1,126 @@
...
+     * @return object|null \OpenTelemetry\API\Metrics\CounterInterface
+     */
+    private static function getCounter()
+    {
+        if (self::$initialized) {
+            return self::$counter;
+        }
+        self::$initialized = true;
+
+        if (!function_exists('dd_trace_env_config') || !\dd_trace_env_config('DD_METRICS_OTEL_ENABLED')) {
```
bwoebi: DD_METRICS_OTEL_ENABLED governs the default exporter or the datadog-configured exporter is used. It does not have a fundamental influence on the presence of otel or such.

I don't think this check belongs here.

src/DDTrace/FeatureFlags/Provider.php
```diff
@@ -0,0 +1,301 @@
...
+    {
+        if (function_exists('dd_trace_env_config')) {
+            return (bool)\dd_trace_env_config('DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED');
+        }
+
+        $envVal = getenv('DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED');
+        if ($envVal === false || $envVal === '') {
+            return false;
+        }
+        return strtolower($envVal) === 'true' || $envVal === '1';
```
bwoebi: I think it should default to false here, if dd_trace_env_config is not present. Because that means you also cannot eval it.

tests/phpunit.xml
```diff
@@ -136,6 +136,9 @@
         <testsuite name="unit">
             <directory>./Unit/</directory>
         </testsuite>
+        <testsuite name="featureflags">
+            <directory>./FeatureFlags/</directory>
+        </testsuite>
```
bwoebi, line 145: You also need to ensure this is run in CI (see `.gitlab/generate-tracer.php`)

bwoebi:
Also please don't split this PR. I want to ensure that the whole functionality works properly and shall be included in a single step.

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+use std::collections::HashMap;
+use std::ffi::{c_char, CStr, CString};
+use std::sync::{Arc, Mutex};
+
+/// Holds both the FFE configuration and a "changed" flag atomically behind a
+/// single Mutex. This avoids the race where another thread could observe
+/// `config` updated but `changed` still false (or vice-versa).
+///
+/// A `RwLock` would be more appropriate here (many readers via `ddog_ffe_evaluate`,
+/// rare writer via `store_config`), but PHP is single-threaded per process so
```
realFlowControl: > PHP is single-threaded per process

Thats only true for NTS builds of PHP, not for ZTS builds of PHP

components-rs/ffe.rs
```diff
@@ -0,0 +1,330 @@
...
+/// contention is not a practical concern. Keeping a Mutex for simplicity.
+struct FfeState {
+    config: Option<Configuration>,
+    changed: bool,
+}
+
+lazy_static::lazy_static! {
+    static ref FFE_STATE: Mutex<FfeState> = Mutex::new(FfeState {
```
bwoebi, line 35: Is FFE config truly global across services/envs etc.? Or is it computed depending on the specific target tuple (service, env, version)?
Note that every thread has its own remote config receiver, with its own target tuple.

bwoebi: If it's the latter, you will have to carry the state in DDTRACE_G() (i.e. the thread-local state managed by the PHP runtime).

leoromanovsky:
> I also question the OpenTelemetry metrics a bit. Like, how important are these? The vast majority of our users will _not_ have OpenTelemetry installed or enabled. Submission of OpenTelemetry metrics will also require synchronous flushing of these metrics (especially if these are the only metrics).
> 
> I would encourage aggregating these in tracers directly.

Thanks for asking, some details below:

https://docs.google.com/document/d/1yOdu6FU3Fw-PMhNlSfqawJ5r6f3tSGomjlJVwdtqpZc/edit?tab=t.0#heading=h.yx4xd78f94da

* importance: they are not "core functionality" but extremely useful for understanding how SDKs behave; useful when making changes to flag state
* flush: understood; we will work these into our onboarding documents

leoromanovsky:
> Also please don't split this PR. I want to ensure that the whole functionality works properly and shall be included in a single step.

Thanks! That's my intuition as well for this first landing.

bwoebi:
@leoromanovsky

I mean, in particular this section from the doc:
> The agent metric pipelines team suggested using [OpenTelemetry Metrics](https://opentelemetry.io/docs/specs/otel/metrics/) instead, which does aggregation in tracers so as to not overload the agent.

is at least for dd-trace-php not true (at least not on the scale where it would be needed). dogstatsd metrics however, are actually aggregated internally in the php tracer.

components-rs/ffe.rs
leoromanovsky: Done — `components-rs/ffe.rs:273` uses `std::mem::replace(&mut state.changed, false)`.

components-rs/ffe.rs
leoromanovsky: Done. `FfeResult` fields are stored as `String` / `Option<String>` (`ffe.rs:110-115`). Accessors return `libdd_common_ffi::CharSlice` (`ffe.rs:168-206`) — no CString round-trip.

components-rs/ffe.rs
leoromanovsky: Done — all fields private; access only via `ddog_ffe_result_*` accessors (`ffe.rs:109-116`).

components-rs/ffe.rs
leoromanovsky: Adopted an alternative: pre-eval errors return an `FfeResult::error(code)` sentinel (`ffe.rs:120-130`, `442-464`) instead of null. Callers surface the condition via `ddog_ffe_result_error_code`. Keeps a single return type across the FFI boundary.

components-rs/ffe.rs
leoromanovsky: Done — `ffe.rs:468`.

components-rs/ffe.rs
leoromanovsky: Done — `ffe.rs:377` uses `HashMap::with_capacity(count)`.

components-rs/ffe.rs
leoromanovsky: Refactored into separate `map_reason` / `map_error_code` helpers (`ffe.rs:329-348`); the call site no longer destructures a tuple.

components-rs/ffe.rs
leoromanovsky: Resolved via the L152 change: pre-eval errors now return `FfeResult::error(code)` instead of null (`ffe.rs:442,447,453,459,464`). NULL is no longer part of the contract.

components-rs/ddtrace.h
leoromanovsky: Done — 32-byte layout, fields ordered big→small: `key (8) + string_value (8) + number_value (8) + value_type (4) + bool_value (1)` + 3 pad. Rust `#[repr(C)]` (`ffe.rs:146-158`) and C header (`ddtrace.h:49-55`) kept in sync. Skipped the tagged union — fields are populated by name on both sides, so the flat struct stays readable.

ext/ddtrace.c
leoromanovsky: Per your follow-up reply — `const` was missing on the PHP 7.0 / 7.1 signatures, no mutation happens on our side. Resolving.

ext/ddtrace.c
leoromanovsky: Done. Proper `PHP_FUNCTION(DDTrace_ffe_*)` with documented stubs in `ext/ddtrace.stub.php`; `dd_trace_internal_fn` dispatch removed.

ext/ddtrace.c
leoromanovsky: Done — `ext/ddtrace.c:97` uses `emalloc`.

ext/ddtrace.c
leoromanovsky: Done — `val`, `var`, `ak` at `ext/ddtrace.c:150-152`.

ext/ddtrace.c
leoromanovsky: Done — reformatted to the house style at `ext/ddtrace.c:156-172`.

ext/ddtrace.c
leoromanovsky: Done — loop variables declared at point of first use (`ext/ddtrace.c:99-102`).

src/DDTrace/FeatureFlags/ExposureWriter.php
leoromanovsky: Done. No HTTP from PHP. `ExposureWriter::send()` (`src/OpenFeature/ExposureWriter.php:85-91`) now calls `DDTrace\ffe_send_exposure()` → extension → Rust. The dedup cache (LRU 65536) and batch buffer live in `EXPOSURE_STATE` (`components-rs/ffe.rs:81-98, 556-620`). The sidecar's periodic flush loop calls `ddog_ffe_flush_exposures` (`ffe.rs:638-676`) and ships to the agent EVP proxy. `test_cross_request_dedup` now exercises this path.

src/DDTrace/FeatureFlags/ExposureWriter.php
leoromanovsky: Resolved by the sidecar rearchitecture — flush is owned by the sidecar's periodic loop, not by PHP.

src/DDTrace/FeatureFlags/ExposureWriter.php
leoromanovsky: Resolved — PHP no longer owns the transport, so an agent-reachability check in this class is no longer needed.

src/DDTrace/FeatureFlags/ExposureWriter.php
leoromanovsky: Timestamp is captured inside `buildEventJson()` (`src/OpenFeature/ExposureWriter.php:117`), which runs synchronously at `send()` time, called from `DataDogProvider::resolveViaFfe()` right after the evaluation returns — so the exposure carries the evaluation's timestamp, not the flush timestamp.

src/DDTrace/FeatureFlags/ExposureWriter.php
leoromanovsky: Resolved — the env var was removed along with the PHP-side HTTP transport.

src/DDTrace/FeatureFlags/ExposureCache.php
leoromanovsky: Resolved — `ExposureCache.php` was deleted. Dedup moved to the Rust LRU (`components-rs/ffe.rs:85, 597-608`), so invalid-entry handling is no longer relevant on the PHP side.

src/DDTrace/FeatureFlags/ExposureCache.php
leoromanovsky: Resolved — `ExposureCache.php` was deleted; dedup moved to the Rust LRU.

src/DDTrace/FeatureFlags/Provider.php
leoromanovsky: Resolved — the old `Provider.php` was removed. Replaced by `DataDogProvider` extending `OpenFeature\implementation\provider\AbstractProvider` (`src/OpenFeature/DataDogProvider.php`).

src/DDTrace/FeatureFlags/Provider.php
leoromanovsky: Resolved — `DataDogProvider` now extends `AbstractProvider` and implements the five typed resolvers (`resolveBooleanValue`, etc.) per the OpenFeature spec (`src/OpenFeature/DataDogProvider.php:133-174`).

src/DDTrace/OpenFeature/DataDogProvider.php
leoromanovsky: Done — `EvaluationContextNormalizer::normalize()` preserves `null` (`src/OpenFeature/EvaluationContextNormalizer.php:36-47`); the targeting key is never coerced to `""`.

components-rs/ffe.rs
leoromanovsky: Can ZTS builds be something we follow up with?

bwoebi: No. ZTS support is often structural and I don't recommend squeezing it in (and half the users are probably using ZTS).

components-rs/ffe.rs
```diff
@@ -0,0 +1,592 @@
...
+    }
+}
+
+/// Flush all buffered exposure events as a batched JSON payload.
+///
+/// Returns a JSON string containing the batch payload with service context and all
+/// buffered events. The batch buffer is cleared after flushing.
+///
+/// In production, the sidecar's periodic flush loop calls this function and sends the
+/// result to the agent EVP proxy.
```
bwoebi: "In production"? I don't see any other calls to this except for testing.

So, this "production" thing is not implemented at all? :-D

Your tests work, but it does not _actually_ work.

leoromanovsky: I'm not ready for a re-review on this; was planning on clicking "request review" again but I'll mark it as draft to make it clear.

PR #3634, opened 2026-02-10
test(CI): manually handle git operation for windows jobs

Leiyks:
### Description

Manually handle git operation for windows jobs to have more aggressive git cleanup.
Also disable an online language test that was supposed to be skipped on 7.1


bwoebi (approved):
Looks correct to me.

PR #3639, opened 2026-02-11
test: fix multiple CI flakiness issues

Leiyks:
### Description

Resolves flakiness in multiple test suites:

**Laravel Octane:** Fixed race condition where test callbacks matched spans before controller execution completed. Added resource name validation to ensure full trace capture.

**Git metadata cache:** Added cache invalidation for long-lived worker processes when git directories become invalid or commits change. Improved nginx readiness check with settle time to prevent 502 errors.

**SQL Server integration:** Switched from SQL Server 2022 to 2019 image to avoid kernel 6.7+ compatibility issues causing SIGABRT crashes on container startup.

### DIFF
--- a/.gitlab/generate-common.php
+++ b/.gitlab/generate-common.php
@@ -200,7 +200,7 @@ foreach ($arch_targets as $arch_target) {
       MONGO_INITDB_ROOT_PASSWORD: test
 
   sqlsrv:
-    name: registry.ddbuild.io/images/mirror/sqlserver:2022-latest
+    name: registry.ddbuild.io/images/mirror/sqlserver:2019-CU15-ubuntu-20.04
     alias: sqlsrv-integration
     variables:
       ACCEPT_EULA: Y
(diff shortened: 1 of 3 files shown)


ext/git.c
```diff
@@ -272,10 +272,20 @@ void replace_git_metadata(git_metadata_t *git_metadata, zend_string *commit_sha,
...
+        // Git directory no longer exists - invalidate cached entry
+        zend_hash_del(&DDTRACE_G(git_metadata), cwd);
         return;
     }
     zend_string *commit_sha = get_commit_sha(ZSTR_VAL(git_dir));
 
+    // If we previously had a commit SHA but now can't read it, the git folder became invalid
+    // (e.g., HEAD file was removed). Invalidate the cache entry.
+    if (!commit_sha && git_metadata->property_commit) {
+        zend_hash_del(&DDTRACE_G(git_metadata), cwd);
```
bwoebi: Nitpick, but you also can just put `else if (git_metadata->property_commit) { zend_hash_del(&DDTRACE_G(git_metadata), cwd); }` in 299, without repeated release and return.

PR #3641, opened 2026-02-11
Add debugging lines to integration tests

estringana:
### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
+++ b/appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
@@ -5,9 +5,7 @@ import com.datadog.appsec.php.mock_agent.MsgpackHelper
 import com.datadog.appsec.php.model.Span
 import com.datadog.appsec.php.model.Trace
 import org.junit.jupiter.api.Test
-import org.junit.jupiter.api.Order
 import org.junit.jupiter.api.TestMethodOrder
-import org.junit.jupiter.api.MethodOrderer
 import org.junit.jupiter.params.ParameterizedTest
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
@@ -22,7 +20,6 @@ import org.msgpack.core.MessagePack
 import static com.datadog.appsec.php.test.JsonMatcher.matchesJson
 import static java.net.http.HttpResponse.BodyHandlers.ofString
 import static org.hamcrest.MatcherAssert.assertThat
-@TestMethodOrder(MethodOrderer.OrderAnnotation)
 trait CommonTests {
 
     AppSecContainer getContainer() {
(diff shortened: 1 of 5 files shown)


appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
```diff
@@ -840,6 +843,17 @@ trait CommonTests {
...
+        // - PHP 8.1 non-ZTS: Symfony sends endpoints, making this test invalid
+        // - PHP 7.4 non-ZTS: Laravel sends endpoints, making this test invalid
+        boolean skipBecauseOfSymfony= phpVersion.contains('8.1') && !variant.contains('zts')
+        boolean skipLaravel74 = phpVersion.contains('7.4') && !variant.contains('zts')
+        assumeTrue(!(skipBecauseOfSymfony || skipLaravel74),
+                skipSymfony81NonZts
+                        ? 'Skipping on PHP 8.1 non-ZTS because Symfony sends endpoints'
+                        : (skipLaravel74
+                                ? 'Skipping on PHP 7.4 non-ZTS because of Laravel sends endpoints'
+                                : ''))
```
cataphract: I don't think this makes much sense. Maybe the problem is that this is annotated with `@Order(1)`, but this is not actually a test class, it's a trait. Maybe the including test also has a test annotated with `@Order(1)`, or doesn't have the `@TestMethodOrder` annotation.

Anyway, I think this test can just be removed. It's not testing any observable behaviour. What we care about is whether endpoints are in fact send, not whatever internal state `\DDTrace\are_endpoints_collected()` returns

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Symfony62Tests.groovy
```diff
@@ -158,7 +158,16 @@ class Symfony62Tests {
...
+        def expectedEndpoints = 6
+
+        if (endpoints.size() != expectedEndpoints) {
+            println "Endpoints count mismatch (${endpoints.size()} != ${expectedEndpoints}). Endpoints:\n" +
+                    endpoints.collect { e ->
+                        "- method=${e.method}, path=${e.path}, operationName=${e.operationName}, resourceName=${e.resourceName}"
+                    }.join("\n")
+        }
+
+        assert endpoints.size() == expectedEndpoints
```
cataphract: A better way would be just to change the endpoints class to have a toString() method (either manually or through `@ToString`, if appropriate). That way, when the assertion fails it prints correctly whatever we say

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
estringana: I disagree. We need to test that after adding an endpoint `\DDTrace\are_endpoints_collected()` says so

PR #3648, opened 2026-02-13
feat(prof): add I/O profiling for macOS

realFlowControl:
### Description

This PR:
- adds I/O profiling to macOS
- cleans up some datatypes in I/O profiling's `mod.rs`

#### Why support macOS I/O profiling?

AFAIK we might have a few customers running the profiler on macOS, but the real value is local testability for I/O profiling in our developer environment which is native macOS.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/io/got_macho.rs
+++ b/profiling/src/io/got_macho.rs
@@ -49,10 +49,6 @@
 // - Apple dyld source: https://opensource.apple.com/source/dyld/
 // - Mach-O headers: <mach-o/loader.h>, <mach-o/nlist.h>
 
-// The libc crate deprecates Mach-O types in favor of the mach2 crate, but we only use a few
-// types and don't want to add a dependency just for that.
-// #![allow(deprecated)]
-
 use super::GotSymbolOverwrite;
 use libc::{c_char, c_void};
 use log::{error, trace};


profiling/src/io/got_macho.rs
```diff
@@ -0,0 +1,526 @@
...
+//
+// ## References
+//
+// - Apple Mach-O format: https://developer.apple.com/documentation/kernel/mach_header_64
+// - Apple dyld source: https://opensource.apple.com/source/dyld/
+// - Mach-O headers: <mach-o/loader.h>, <mach-o/nlist.h>
+
+// The libc crate deprecates Mach-O types in favor of the mach2 crate, but we only use a few
+// types and don't want to add a dependency just for that.
+#![allow(deprecated)]
```
morrisonlevi: Why? Is there some undesirable reason to have a macOS-only dependency on mach2?

morrisonlevi (commented):
I have some more reading to do to review the rest of it, but here's some review.

profiling/src/io/got_macho.rs
realFlowControl: The main reason was to keep the dependency surface minimal and avoid extra target-specific dependency/lockfile "overhead" just for this "local-dev" feature. I wanted to keep things "contained", so that being said: there isn’t a functional blocker to using mach2. If you prefer, I can switch this file to mach2 and drop the `#![allow(deprecated)]`

realFlowControl: I added the `mach2` crated (only if compiled for MacOS)

profiling/src/io/got_macho.rs
```diff
@@ -0,0 +1,545 @@
...
+//
+// ## References
+//
+// - Apple Mach-O format: https://developer.apple.com/documentation/kernel/mach_header_64
+// - Apple dyld source: https://opensource.apple.com/source/dyld/
+// - Mach-O headers: <mach-o/loader.h>, <mach-o/nlist.h>
+
+// The libc crate deprecates Mach-O types in favor of the mach2 crate, but we only use a few
+// types and don't want to add a dependency just for that.
+// #![allow(deprecated)]
```
morrisonlevi: I think that these comments need cleaned up since you depend on mach2 now for Apple targets?

morrisonlevi (approved):
I've been trying to review and undersand this with the help of AI. It seems like during first rinit, that every symbol is already resolved. There's no guarantee of this, everything is loaded with LAZY as far as I can tell. So I wonder if all of the code paths are correct or not.

However, this is macOS which is not a prod platform and it seems to work for me locally, so I'm okay with approving it.

profiling/src/io/got_macho.rs
realFlowControl: Ah yes, removed that left-over

PR #3652, opened 2026-02-16
test(profiling): add missing prof correctness tests for I/O profiling

realFlowControl:
### Description

This PR adds missing I/O correctness tests

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/tests/correctness/io.json
morrisonlevi, line 1: An error_margin of 10 is pretty high for this deterministic benchmark, isn't it? Any particular reason it isn't tighter?

realFlowControl: I checked the runs and they all where at "0% error", but if we start adjusting the sampling distance, we might get some errors in the future.
I've tightened the `error_margin` to `2%`. In case we see failures here when adjusting the sampling distance in the future, we can then make a decision of how to handle that.

PR #3653, opened 2026-02-16
Send response headers on meta even without event

estringana:
### Description

Extension should make sure that response `content-type` and `content-length` are always presents on span tags. Even without appsec events

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
+++ b/appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
@@ -241,6 +241,8 @@ trait CommonTests {
         assert span.metrics."_dd.appsec.enabled" == 1.0d
         assert respContentType != null && respContentType.length() > 0
         assert span.meta."http.response.headers.content-type" == respContentType
+        assert span.meta."http.response.headers.content-encoding" == 'foobar'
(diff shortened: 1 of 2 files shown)


appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/CommonTests.groovy
```diff
@@ -225,6 +225,24 @@ trait CommonTests {
...
+            def headerContentType = resp.headers().firstValue('Content-Type')
+            if (headerContentType.isPresent()) {
+                respContentType = headerContentType.get()
+            }
+        }
+
+        Span span = trace.first()
+        assert span.metrics."_dd.appsec.enabled" == 1.0d
+        assert respContentType != null && respContentType.length() > 0
+        assert span.meta."http.response.headers.content-type" == respContentType
```
cataphract, line 243: You could test the other three as well. BTW the PR description says it's content-length and -type that we're adding, but -encoding and -language are sent as well. Is this the intended behavior?

estringana: Yes it is. I'll add a test for the others

estringana: Done

PR #3654, opened 2026-02-17
fix(profiling): store and restore `errno` in I/O profiling wrappers

realFlowControl:
### Description

I/O profiling wraps libc I/O calls (for example `poll`, `recv*`, `send*`, `read`/`write`, `fread`/`fwrite`, `close`) to measure duration and decide whether to collect a sample. That wrapper logic runs after the original syscall returns and includes timing and Poisson sampling-distance updates.

Callers rely on `errno` from the original syscall return path. The additional wrapper work can also touch `errno` (for example error paths in `clock_gettime`, and libm operations used by `rand_distr::Poisson`, such as `exp`/`tan`/`log`-related math). Without preservation, the wrapper can return the syscall result with an unrelated `errno`.

This PR snapshots `errno` immediately after the original syscall and restores it before returning from the wrapper, making the hook transparent to callers.

#### Why now?

We had a customer escalation with I/O profiling enabled where phpredis intermittently reported `RedisException: Connection lost`. In reproduced runs, wrapped syscalls returned with `errno=EAGAIN (11)`, but by wrapper return `errno` had changed to `ERANGE (34)`. That incorrect `errno` drove phpredis into its disconnect path. Restoring `errno` fixes this behavior.

#### Alternatives

An alternative is to defer more work (especially sampling-distance recomputation) to `EG(vm_interrupt)`, similar to wall-time profiling. That could reduce risk in syscall wrappers, but it does not remove the need to preserve `errno` in wrappers, which is why this PR keeps the current architecture. It also would not have fixed this bug, it would only make the race less likely to trigger and harder to diagnose.

### Reviewer checklist
- [ ] Test coverage seems ok.

PROF-13760

### DIFF
--- a/profiling/src/io/mod.rs
+++ b/profiling/src/io/mod.rs
@@ -15,14 +15,38 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Mutex, OnceLock};
 use std::time::Instant;
 
-#[inline]
-unsafe fn get_errno() -> c_int {
-    *libc::__errno_location()
+/// RAII guard that snapshots `errno` on creation and restores it on drop.
+///
+/// I/O profiling wrappers intercept libc calls and run additional logic (timing, sampling) that
+/// can clobber `errno`. Callers of the original syscall expect `errno` to reflect the syscall
+/// result, not our wrapper internals. Create an `ErrnoBackup` right after the original syscall
+/// returns and let it live until the end of the wrapper function, its `Drop` impl will
+/// transparently restore the original `errno` value.
+struct ErrnoBackup {
+    errno: c_int,
+    location: *mut c_int,
 }
 
-#[inline]
-unsafe fn set_errno(errno: c_int) {
-    *libc::__errno_location() = errno;
+impl ErrnoBackup {
+    /// Snapshots the current `errno` value.
+    #[inline]
+    unsafe fn new() -> Self {
+        let location = libc::__errno_location();
+        Self {
+            errno: *location,
+            location,
+        }
+    }
+}
+
+impl Drop for ErrnoBackup {
+    /// Restores `errno` value.
+    #[inline]
+    fn drop(&mut self) {
+        unsafe {
+            *self.location = self.errno;
+        }
+    }
 }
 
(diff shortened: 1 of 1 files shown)


profiling/src/io/mod.rs
```diff
@@ -15,6 +15,16 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Mutex, OnceLock};
 use std::time::Instant;
 
+#[inline]
+unsafe fn get_errno() -> c_int {
+    *libc::__errno_location()
```
bwoebi: No idea if relevant/you care, but this will do two calls which cannot be eliminated.
Maybe you could have `struct ErrnoBackup { errno: c_int; location: mut* c_int; } fn backup_errno() -> ErrnoBackup { unsafe { let location = libc::__errno_location(); ErrnoBackup { errno: *location, location} } }`; and add a trivial drop impl doing `unsafe { *self.location = errno };`. And simply `let _backup_errno = backup_errno();`

realFlowControl: Uh, thats a neat idea, let me change this

realFlowControl: done

profiling/src/io/mod.rs
```diff
@@ -15,6 +15,40 @@ use std::sync::atomic::{AtomicU64, Ordering};
...
+/// transparently restore the original `errno` value.
+struct ErrnoBackup {
+    errno: c_int,
+    location: *mut c_int,
+}
+
+impl ErrnoBackup {
+    /// Snapshots the current `errno` value.
+    #[inline]
+    unsafe fn new() -> Self {
```
morrisonlevi, line 33: It's merged already but is this unsafe? Sure it uses unsafe code but I'm not seeing how this operation is unsafe.

realFlowControl: It would most likely be more correct to not have the function `unsafe`, but wrap the unsafe bits in the code itself, but then it would be just an unsafe block around the entire body or like this:
```rust
    fn new() -> Self {
        let location = unsafe { libc::__errno_location() };
        Self {
            errno: unsafe { *location },
            location,
        }
    }
```
Now that I see it looks okay actually haha ... 😉

PR #3655, opened 2026-02-17
Fix possible race condition leading to a crash upon sidecar reconnect in ZTS

bwoebi:
See also comment added to the code.

### DIFF
--- a/ext/sidecar.c
+++ b/ext/sidecar.c
@@ -144,9 +144,21 @@ static void dd_sidecar_on_reconnect(ddog_SidecarTransport *transport) {
             ddog_CharSlice env_name = dd_zend_string_to_CharSlice(DDTRACE_G(last_env_name));
             ddog_CharSlice version = dd_zend_string_to_CharSlice(DDTRACE_G(last_version));
 
+            ddog_DynamicInstrumentationConfigState dynamic_instrumentation_state;
+#if ZTS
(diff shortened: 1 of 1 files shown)


ext/sidecar.c
```diff
@@ -144,9 +144,21 @@ static void dd_sidecar_on_reconnect(ddog_SidecarTransport *transport) {
             ddog_CharSlice env_name = dd_zend_string_to_CharSlice(DDTRACE_G(last_env_name));
             ddog_CharSlice version = dd_zend_string_to_CharSlice(DDTRACE_G(last_version));
 
+            ddog_DynamicInstrumentationConfigState dynamic_instrumentation_state;
+#if ZTS
+            // With the current architecture of config it's not accessible via the TSRMLS_CACHE and thus may be actually invalid on the current thread.
+            // This is a known issue and will be fixed with refactor of the module_globals usage of config. The current behaviour is not perfect, but has to be considered acceptable.
+            if (zai_config_memoized_entries[DDTRACE_CONFIG_DD_DYNAMIC_INSTRUMENTATION_ENABLED].name_index >= 0) {
```
morrisonlevi, line 151: Should we make `name_index` atomic? I'm not sure exactly when/how this field is updated. The minit and first config things should be fine, not sure about how remote config works.

bwoebi: That value is not supposed to be touched after first rinit anyway. So doesn't matter.

PR #3656, opened 2026-02-17
Block sidecar notification signal while sleeping

bwoebi:
Otherwise users may sleep a lot less than they expect.

### DIFF
--- a/ext/handlers_signal.c
+++ b/ext/handlers_signal.c
@@ -81,6 +81,7 @@ static void dd_handle_signal(zif_handler original_function, INTERNAL_FUNCTION_PA
     x(pcntl_sigwaitinfo) \
     x(sleep) \
     x(usleep) \
+    x(time_nanosleep) \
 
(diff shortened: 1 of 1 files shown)


ext/handlers_signal.c
```diff
@@ -79,6 +79,8 @@ static void dd_handle_signal(zif_handler original_function, INTERNAL_FUNCTION_PA
     x(ldap_connect) \
     x(sqlsrv_connect) \
     x(pcntl_sigwaitinfo) \
+    x(sleep) \
+    x(usleep) \
```
realFlowControl, line 83:
```suggestion
    x(usleep) \
    x(time_sleep_until) \
    x(time_nanosleep) \
```

Not sure, but are these two also affected?

bwoebi: time_nanosleep I actually forgot. Added. time_sleep_until is actually doing a `while (nanosleep() == EINTR)` itself

PR #3658, opened 2026-02-17
Adds process tags to remote config payload

dubloom:
### Description

This PR implements this [RFC](https://docs.google.com/document/d/1AFdLUuVk70i0bJd5335-RxqsvwAV9ovAqcO2z5mEMbA/edit?pli=1&tab=t.0#heading=h.s9l1lctqlg11) for Remote Configuration.

Add process_tags to remote configuration payload.

This PR comes with this libdatadog PR: https://github.com/DataDog/libdatadog/pull/1586

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/tests/integration/gradle/tag_mappings.gradle
+++ b/appsec/tests/integration/gradle/tag_mappings.gradle
@@ -137,5 +137,5 @@ ext.tag_mappings = [
     'nginx-fpm-php-8.5-release-zts': 'sha256:527f2ba273685ff7e3f72a74076fd6d99cbc87d51fb3485fef4dc80a0cfddeee',
     'nginx-fpm-php-8.5-release': 'sha256:fc2eea1ed06dfcf8102bdbfb2596f9a32bbe6493ca83d53370e1438289caa6aa',
     'nginx-fpm-php-8.5-debug': 'sha256:915d0155897e4f42cf1867b93edbb928afcc1c5b3bd3eccdb2d56ac50510cd82',
-	'nginx-fpm-php-8.5-release-musl': 'sha256:73857002765b217e1076adc04f9ebc88505b0fa77fc48e2ad3005594b2f39bc9',
+	'nginx-fpm-php-8.5-release-musl': 'sha256:a1ef09c1bf398049b340ab2a1c40839d828420df5a404831bb7b33a09d433755',
 ]


appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/mock_agent/rem_cfg/RemoteConfigRequest.java
```diff
@@ -218,6 +218,12 @@ public static class TracerInfo {
             @JsonProperty("app_version")
             public String serviceVersion;
 
+            @JsonProperty("process_tags")
+            public List<String> processTags;
+
+            @JsonProperty("container_tags")
+            public String containerTags;
```
Leiyks: Shouldn't it be a List<String> instead ?

ext/sidecar.c
```diff
@@ -124,13 +125,9 @@ void ddtrace_sidecar_update_process_tags(void) {
         return;
     }
 
-    zend_string *process_tags = ddtrace_process_tags_get_serialized();
-    if (!process_tags || ZSTR_LEN(process_tags) == 0) {
-        return;
-    }
```
Leiyks, line 130: Why remove this check ?

appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/mock_agent/rem_cfg/RemoteConfigRequest.java
dubloom: It should indeed, I looked too quickly at the proto, thanks for catching.

Addressed in [982f9bf](https://github.com/DataDog/dd-trace-php/pull/3658/commits/982f9bf20a43a32d2e29d28ae5cdf64dc3673595)

ext/sidecar.c
dubloom: It was firstly done, because we could in theory always send an empty vect (if process_tags are deactivated we would return an empty vec and that is a valid state).

However, that was a nice early optimization, addressed in [982f9bf](https://github.com/DataDog/dd-trace-php/pull/3658/commits/982f9bf20a43a32d2e29d28ae5cdf64dc3673595)

PR #3659, opened 2026-02-17
refactor(profiling): ErrnoBackup::new is safe

morrisonlevi:
### Description

In code review, I noticed `ErrnoBackup::new` was safe, but it was already merged. So I fixed it, and added safety comments. We also verify that it is !Send, because it's not safe to write to another thread's errno, which would happen if we sent `ErrnoBackup` to another thread.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl (approved):
Nice cleanup

PR #3660, opened 2026-02-18
ci: upload system-tests results to Test Optimization and add tracer-release nightly

nccatoni:
### Description

Two changes to the system-tests CI integration:

1. **Upload all system-tests results to Test Optimization** — adds an `after_script` to the `.system_tests` base job that copies JUnit XML files and uploads them via the existing `silent-upload-junit-to-datadog.sh` script.

2. **Add `tracer-release` nightly** — create a new system-tests workflow that uses the official system-tests reusable workflow to run the tracer-release scenario group on master.

This is meant to replace the [system-tests-dashboard nightly](https://github.com/DataDog/system-tests-dashboard/blob/main/.github/workflows/nightly.yml).

### Reviewer checklist
- [x] This is a CI-only change, no code changes.


.gitlab/generate-package.php
```diff
@@ -1223,6 +1223,45 @@
...
+
+      # Upload JUnit results
+      echo "Uploading JUnit results..."
+      /tmp/datadog-ci junit upload \
+        system-tests/logs*/reportJunit.xml \
+        --service system-tests \
+        --env ci \
+        --verbose \
+        --xpath-tag "test.codeowners=/testcase/properties/property[@name='test.codeowners']"
+      echo "=== Upload complete ==="
```
Leiyks: We have a script that basically do all of this in `.gitlab/upload-junit-to-datadog.sh`

Leiyks: You may need to tweak it a bit to add the xpath-tag thing tho

.gitlab/generate-package.php
```diff
@@ -1244,10 +1245,13 @@
       when: always
       paths:
         - .cache/
+  after_script:
+    - DATADOG_API_KEY=$(cat /tmp/.dd-api-key 2>/dev/null) || true
+    - mkdir -p artifacts && for f in system-tests/logs*/reportJunit.xml; do dir=$(basename $(dirname "$f")); cp "$f" "artifacts/reportJunit_${dir}.xml" 2>/dev/null || true; done
```
bwoebi, line 1250: Fetching the API key manually is redundant with the script, isn't it?

nccatoni: The connection to vault fails for the parametric job in the script like in this [run](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-php/-/jobs/1533493564#L3356). From what I understand it might be due to the vault agent getting killed at some point since it only happens for the longest job

bwoebi: that's odd, because it works with other jobs running longer in the main tracer test pipeline.
But I suppose not worth the time debugging that more.

PR #3670, opened 2026-02-24
perf(config): cache sys getenv

morrisonlevi:
### Background

In request initialization aka rinit/activate, calling the system `getenv` when a SAPI env var isn't found is slow enough it shows up in profiles:

image (1)

Note that glibc's `getenv` does a linear scan on the environ and does `strncmp` on the members, so we can definitely do better.

### Description

Caches libc's `getenv` in minit and again in first rinit. Note that php-fpm's `env` directive _actually_ changes the system env var, it doesn't include it in the fastcgi environment, so that's why we have to grab it again on first rinit. This change is also more correct: system environment variables map to a system INI setting, which shouldn't change at runtime, so if they do, we shouldn't really respect it.

We also make another optimization as part of this: we borrow memory from the config when possible for the decoder. Examples include otel normalization when the result is a static string like "0" or "1", as well as when using system environment variables.

Parts of zai config had to be refactored. The internals were very brittle to changes. This PR refines the tests and hopefully improves documentation as well so that it's a bit less so.

Overall, we shaved off roughly 40-50% of the time spent in `zai_config_rinit`. Here's an example diff from the profiling comparison view:

Screenshot 2026-03-19 at 3 15 49 PM

Note that you can see `sapi_getenv` in the flamegraph now where you couldn't before. However, we still come out ahead.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/ddtrace.c
+++ b/ext/ddtrace.c
@@ -1615,7 +1615,6 @@ static PHP_MSHUTDOWN_FUNCTION(ddtrace) {
 
     ddtrace_sidecar_shutdown();
 
-    ddtrace_live_debugger_mshutdown();
     ddtrace_process_tags_mshutdown();
 
 #if PHP_VERSION_ID >= 80000 && PHP_VERSION_ID < 80100
--- a/ext/ddtrace.h
+++ b/ext/ddtrace.h
@@ -41,6 +41,13 @@ typedef struct ddtrace_span_event ddtrace_span_event;
 typedef struct ddtrace_exception_span_event ddtrace_exception_span_event;
 typedef struct ddtrace_git_metadata ddtrace_git_metadata;
 
+typedef struct dd_refcounted_linked dd_refcounted_linked;
+
+typedef struct {
+    zend_arena *arena;
+    dd_refcounted_linked *ephemerals;
+} dd_capture_arena;
+
 extern datadog_php_sapi ddtrace_active_sapi;
 
 extern ddog_CharSlice php_version_rt;
@@ -147,8 +154,7 @@ ZEND_BEGIN_MODULE_GLOBALS(ddtrace)
     ddog_AgentRemoteConfigReader *agent_config_reader;
     ddog_RemoteConfigState *remote_config_state;
     ddog_AgentInfoReader *agent_info_reader;
-    zend_arena *debugger_capture_arena;
-    HashTable debugger_capture_ephemerals;
+    dd_capture_arena debugger_capture_arena;
     ddog_Vec_DebuggerPayload exception_debugger_buffer;
     HashTable active_rc_hooks;
     HashTable *agent_rate_by_service;
--- a/ext/exception_serialize.c
+++ b/ext/exception_serialize.c
@@ -91,7 +91,7 @@ static void ddtrace_capture_string_value(zend_string *str, struct ddog_CaptureVa
     if (!value->not_captured_reason.len) {
         value->value = (ddog_CharSlice) {.ptr = ZSTR_VAL(str), .len = ZSTR_LEN(str)};
         if (value->value.len > config->max_length) {
-            char *integer = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena), 20);
+            char *integer = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena).arena, 20);
             int len = sprintf(integer, "%" PRIuPTR, value->value.len);
             value->size = (ddog_CharSlice) {.ptr = integer, .len = len};
             value->value.len = config->max_length;
@@ -103,7 +103,7 @@ static void ddtrace_capture_string_value(zend_string *str, struct ddog_CaptureVa
 static void ddtrace_capture_long_value(zend_long num, struct ddog_CaptureValue *value) {
     value->type = DDOG_CHARSLICE_C("int");
     if (!value->not_captured_reason.len) {
-        char *integer = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena), 20);
+        char *integer = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena).arena, 20);
         int len = sprintf(integer, ZEND_LONG_FMT, num);
         value->value = (ddog_CharSlice) {.ptr = integer, .len = len};
     }
@@ -133,7 +133,7 @@ void ddtrace_create_capture_value(zval *zv, struct ddog_CaptureValue *value, con
         case IS_DOUBLE: {
             value->type = DDOG_CHARSLICE_C("float");
             if (!value->not_captured_reason.len) {
-                char *num = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena), 20);
+                char *num = zend_arena_alloc(&DDTRACE_G(debugger_capture_arena).arena, 20);
                 php_gcvt(Z_DVAL_P(zv), (int) EG(precision), '.', 'E', num);
                 value->value = (ddog_CharSlice) {.ptr = num, .len = strlen(num)};
             }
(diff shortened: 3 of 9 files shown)


zend_abstract_interface/config/config.c
```diff
@@ -237,6 +294,13 @@ void zai_config_first_time_rinit(bool in_request) {
     (void)in_request;
 #endif
 
+    if (in_request) {
```
bwoebi: We only need it for FPM, right? So let's add a sapi_module check here?
Saving some minor time on CLI scripts.

zend_abstract_interface/config/config.c
```diff
@@ -12,11 +12,66 @@ HashTable zai_config_name_map = {0};
...
+    if (!buf.ptr || !buf.len) return false;
+    if (id >= zai_config_memoized_entries_count) return false;
+    zai_config_memoized_entry *memoized = &zai_config_memoized_entries[id];
+    if (name_index >= memoized->names_count) return false;
+
+    zai_config_cached_env_value *cached = &zai_config_cached_env_values[id][name_index];
+    if (!cached->ptr) return false;
+    if (cached->len >= buf.len) return false;
+
+    memcpy(buf.ptr, cached->ptr, cached->len + 1);
```
bwoebi: There's not really an intrinsic need to copy. The copying in zai_getenv_ex is also unnecessary, but here we very clearly don't need it (but yeah, you'll have to duplicate branches then).
I'd personally return a simple zai_str instead of zai_env_buffer.

zend_abstract_interface/config/config.c
```diff
@@ -12,11 +12,66 @@ HashTable zai_config_name_map = {0};
...
-    // But in general allmost all configurations shall only be accessed after first RINIT. (the trivial getter will
-    return zai_getenv_ex(name, buf, true) == ZAI_ENV_SUCCESS;
+typedef struct {
+    // ptr == NULL means env var was unset at MINIT.
+    // ptr != NULL with len == 0 means env var was set to an empty string.
+    char *ptr;
+    size_t len;
+} zai_config_cached_env_value;
+
+static zai_config_cached_env_value zai_config_cached_env_values[ZAI_CONFIG_ENTRIES_COUNT_MAX][ZAI_CONFIG_NAMES_COUNT_MAX];
```
bwoebi:
```suggestion
static zai_string zai_config_cached_env_values[ZAI_CONFIG_ENTRIES_COUNT_MAX][ZAI_CONFIG_NAMES_COUNT_MAX];
```
Simply?

morrisonlevi: zai_string isn't supposed to have a null pointer, right? IIRC that's why I did this. This distinguishes between unset and null. But there is a `zai_option_str`, and I think using a `zai_option_str` except actually owning it is better than putting `null` into `zai_string`. So I'll experiment with that.

zend_abstract_interface/config/config.c
morrisonlevi: I'm not sure what litespeed does, and IIRC we did have a customer using it. How about a not-CLI SAPI check instead? That way web SAPIs stay consistent, but it's skipped on CLI where it's unnecessary.

morrisonlevi: I think it's reasonably clearer now. Patch is smaller, at least xD

zend_abstract_interface/config/config.c
```diff
@@ -12,11 +12,68 @@ HashTable zai_config_name_map = {0};
...
+        }
+    }
+}
+
+static bool zai_config_get_env_value(zai_str name, zai_config_id id, uint8_t name_index, zai_env_buffer *buf) {
     // TODO Handle other return codes
-    // We want to explicitly allow pre-RINIT access to env vars here. So that callers can have an early view at config.
-    // But in general allmost all configurations shall only be accessed after first RINIT. (the trivial getter will
-    return zai_getenv_ex(name, buf, true) == ZAI_ENV_SUCCESS;
+    if (zai_getenv(name, buf) == ZAI_ENV_SUCCESS) {
```
bwoebi:
```suggestion
    if (zai_getenv_ex(name, buf, false, false) == ZAI_ENV_SUCCESS) {
```
? Because you defined zai_getenv() as always accessing the process env.

zend_abstract_interface/config/config.c
bwoebi: A non cli sapi check would be fine by me too. Just let's not eat the cost for cli.

zend_abstract_interface/config/config.c
morrisonlevi: `zai_config_find_and_set_value` is called only by `zai_config_first_time_rinit`. At this time, we do want to get the process environment, yes?

zend_abstract_interface/config/config.c
```diff
@@ -12,55 +13,120 @@ HashTable zai_config_name_map = {0};
...
+            zai_option_str val = zai_sys_getenv(name);
+            if (zai_option_str_is_none(val)) {
+                continue;
+            }
+
+            size_t len = val.len;
+            // +1 to hold the null terminator.
+            char *dst = pemalloc(len + 1, 1);
+            // +1 to include the null terminator in the copy.
+            memcpy(dst, val.ptr, len + 1);
```
bwoebi:
```suggestion
            char *dst = zend_strndup(val.ptr, len);
```

morrisonlevi: Forgot this API exists. Current version locally looks like this:

```c
static void zai_config_sys_env_cache(void) {
    for (zai_config_id i = 0; i < zai_config_memoized_entries_count; i++) {
        zai_config_memoized_entry *m = &zai_config_memoized_entries[i];
        for (uint8_t n = 0; n < m->names_count; n++) {
            zai_str name = ZAI_STR_NEW(m->names[n].ptr, m->names[n].len);
            zai_option_str val = zai_sys_getenv(name);
            zai_config_cached_sys_env[i][n] = zai_option_str_is_some(val)
                ? zend_strndup(val.ptr, val.len) : NULL;
        }
    }
}
```

zend_abstract_interface/config/config_ini.c
```diff
@@ -427,42 +427,47 @@ void zai_config_ini_rinit(void) {
...
-
     for (uint16_t i = 0; i < zai_config_memoized_entries_count; ++i) {
+        ZAI_ENV_BUFFER_INIT(buf, ZAI_ENV_MAX_BUFSIZ);
         zai_config_memoized_entry *memoized = &zai_config_memoized_entries[i];
         if (memoized->ini_change == zai_config_system_ini_change) {
             continue;
         }
 
-        // makes only sense to update INIs once, avoid rereading env unnecessarily
-        if (!env_to_ini_name || !memoized->original_on_modify) {
```
morrisonlevi, line 439: This "optimization" to skip work seems to break behaviors. Removing it means we introduce more `sapi_getenv` calls, but it still comes out ahead performance-wise.

zend_abstract_interface/config/config_ini.c
```diff
@@ -427,42 +427,47 @@ void zai_config_ini_rinit(void) {
...
+                && zai_config_process_runtime_env(memoized, buf, in_startup, i, name_index)) {
+                memoized->name_index = ZAI_CONFIG_ORIGIN_LOCAL_STABLE;
+                memoized->config_id = (zai_str) ZAI_STR_FROM_ZSTR(entry->config_id);
                 goto next_entry;
             }
         }
 
+        if (memoized->env_config_fallback && memoized->env_config_fallback(&buf, false) && zai_config_process_runtime_env(memoized, buf, in_startup, i, 0)) {
+            goto next_entry;
+        }
```
bwoebi: Moving this out of the loop seems right!

zend_abstract_interface/config/config_ini.c
bwoebi: Interesting. Hm.

bwoebi: Is it due to the interaction with env_config_fallback?

morrisonlevi: Bob and I hopped on a call. The test setup is complex (by necessity if you want it to be real and not synthetic) and there were some quite stupid code in there I overlooked (dangers of using AI). With the corrected setup, it seems it might be able to be re-enabled.

morrisonlevi: Present on latest version, tests are passing.

bwoebi (approved):
Looks good to me, nice job Levi :-)

PR #3671, opened 2026-02-24
ci: adjust CPU and memory settings

morrisonlevi:
### Description

K8s is scheduling based on  REQUEST, but allowed to run up to LIMIT. So if a job requests 3 GiB of memory, it's allowed to go up to 4 GiB. On a whole-node basis, if enough jobs do this then node can run out of memory and start using OOM-kill. Note that for CPU, the limit is not enforced unless the request is equal to the limit (but I'm not sure that's actually happening either, but it's what it's _supposed_ to be doing).

These specific job families failed for me in PRs, which is why I've adjusted these specific jobs. Note that in the case of appsec, it did _not_ reach its own limit, but rather the node it was on became full.

I am increasing the limits by feedback from a coworker who supports our CI infrastructure.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.gitlab/generate-profiler.php
+++ b/.gitlab/generate-profiler.php
@@ -39,6 +39,7 @@ foreach ($profiler_minor_major_targets as $version) {
     - if [ -f /sbin/apk ] && [ $(uname -m) = "aarch64" ]; then ln -sf ../lib/llvm17/bin/clang /usr/bin/clang; fi
 
     - cd profiling
+    - 'echo "nproc: $(nproc)"'
     - 'echo "KUBERNETES_CPU_REQUEST: ${KUBERNETES_CPU_REQUEST:-<unset>}"'
     - |
       if [ -n "${KUBERNETES_CPU_REQUEST:-}" ]; then


cataphract (approved):

.gitlab/generate-profiler.php
```diff
@@ -38,6 +39,13 @@
     - if [ -f /sbin/apk ] && [ $(uname -m) = "aarch64" ]; then ln -sf ../lib/llvm17/bin/clang /usr/bin/clang; fi
 
     - cd profiling
+    - 'echo "nproc: $(nproc)"'
```
morrisonlevi, line 42: Note that this reports a quite large number, such as `66`. This means we should be setting the number of works for parallel systems and not let them discover it!

.gitlab/generate-profiler.php
```diff
@@ -18,9 +18,10 @@
   tags: [ "arch:${ARCH}" ]
   image: registry.ddbuild.io/images/mirror/datadog/dd-trace-ci:${IMAGE_PREFIX}${PHP_MAJOR_MINOR}${IMAGE_SUFFIX}
   variables:
-    KUBERNETES_CPU_REQUEST: 5
-    KUBERNETES_MEMORY_REQUEST: 3Gi
-    KUBERNETES_MEMORY_LIMIT: 4Gi
+    KUBERNETES_CPU_REQUEST: 3
+    KUBERNETES_CPU_LIMIT: 3
+    KUBERNETES_MEMORY_REQUEST: 6Gi
+    KUBERNETES_MEMORY_LIMIT: 6Gi
```
morrisonlevi, line 24: These numbers look good in one of dashboards. Note that we are really memory bound and for parts of it we are only engaging a single core so there's not a lot of value in increasing CPUs at the moment.

5 GiB is too small, I've observed 4.9 GiB in practice.

PR #3677, opened 2026-02-26
fix(profiling): respect ZAI_CONFIG_ORIGIN_DEFAULT for agent endpoint detection

Leiyks:
### Description

Since `a3409cd78`, `DD_AGENT_HOST` and `DD_TRACE_AGENT_PORT` have non-empty defaults (`"localhost"` / `8126`). The PHP tracer already filters these with `ZAI_CONFIG_ORIGIN_DEFAULT` checks, but the Rust profiling code was not updated — causing `agent_host()` and `trace_agent_port()` to return `Some(...)` unconditionally and bypass the Unix socket fallback that SSI environments rely on.

Fix: adds `ddog_php_prof_config_is_set_by_user()` (checks `name_index != ZAI_CONFIG_ORIGIN_DEFAULT`) and gates both functions on it.

Also pins `cargo install cbindgen` to `--version 0.29.0 --locked` to avoid `getrandom v0.4.1` pulling in `edition2024` support, which requires Cargo ≥ 1.85 (container is on 1.84.1).

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -451,7 +451,7 @@ generate_cbindgen: cbindgen_binary # Regenerate components-rs/ddtrace.h componen
 
 cbindgen_binary:
 	if ! command -v cbindgen &> /dev/null; then \
-		cargo install cbindgen; \
+		cargo install cbindgen --version 0.29.0 --locked; \
 	fi
 
 EXT_DIR:=/opt/datadog-php


morrisonlevi (approved):
package-oci failures are unrelated. Merging.

PR #3679, opened 2026-02-26
ci: fix package-oci failure in parent pipeline

Leiyks:
## Root Cause

Commit a3409cd78 included `one-pipeline.locked.yml` in the parent pipeline to get the supported-configurations validation templates. This accidentally also brought in `package-oci` and its downstream jobs, which have no access to `package loader` artifacts in the parent pipeline.

`package-oci` is meant to run only in the **child pipeline** (via `package-trigger`) where `generate-package.php` provides the correct `needs: [package loader]` override.

## Fix

- Set `SKIP_SHARED_PIPELINE: "true"` globally to suppress shared-pipeline jobs in the parent pipeline.
- Override to `"false"` in `package-trigger` so the child pipeline still runs them correctly.
- Explicitly suppress `requirements_json_test` since it doesn't check `SKIP_SHARED_PIPELINE`.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -9,6 +9,11 @@ variables:
   GIT_SUBMODULE_STRATEGY: recursive
   # Only clone libdatadog submodule by default
   GIT_SUBMODULE_PATHS: libdatadog
+  # Suppress shared-pipeline jobs (package-oci and its downstream) in the
+  # parent pipeline. They run in the child pipeline (package-gen.yml) where
+  # package loader artifacts are available. Override to "false" in
+  # package-trigger so the child pipeline runs them correctly.
+  SKIP_SHARED_PIPELINE: "true"
   RELIABILITY_ENV_BRANCH:
     value: "master"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -98,6 +98,13 @@ package-trigger:
     NIGHTLY_BUILD: $NIGHTLY_BUILD
     RELIABILITY_ENV_BRANCH: $RELIABILITY_ENV_BRANCH
 
+# package-oci runs in the child pipeline (package-gen.yml) with the correct
+# needs: [package loader] override. The parent pipeline doesn't have package
+# loader artifacts, so we suppress it here.
+package-oci:
+  rules:
+    - when: never
```
morrisonlevi: There's a `package-oci` job in a child pipeline though, will this suppress that one too?

Leiyks: Indeed, it even blocked the pipeline completely, found another fix which actually works

PR #3680, opened 2026-02-27
Support parsing partial json in appsec

cataphract:
### Description

Backport partial json parsing from API10 branch.

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/src/extension/json_truncated_parser.cpp
```diff
@@ -0,0 +1,407 @@
+// Unless explicitly stated otherwise all files in this repository are
+// dual-licensed under the Apache-2.0 License or BSD-3-Clause License.
+//
+// This product includes software developed at Datadog
+// (https://www.datadoghq.com/). Copyright 2021 Datadog, Inc.
+
+#include "json_truncated_parser.h"
```
estringana, line 7: Why have you decided to go for a CPP implementation on the exension?

appsec/src/helper/.gitignore
```diff
@@ -1 +0,0 @@
-/version.hpp
```
estringana, line 1: Why is this changed?

cataphract: https://github.com/DataDog/dd-trace-php/commit/3df4e17f08#diff-d7c6e792306074c39c93c98658584807ea65a8c68bbce5c3d353e3afc0a87e55L1-L19 changed the location where version.hpp was generated (no more on the source tree), so this ignore is not needed anymore

appsec/src/extension/json_truncated_parser.cpp
cataphract: It was dictated by the choice of library (rapidjson). The C libraries with SAX-like, event-driven API have not seen updates in more than a decade.

PR #3691, opened 2026-03-06
fix: enforce span limit in curl_multi_exec and DDTrace\start_span code paths

Leiyks:
## Summary

`DD_TRACE_SPANS_LIMIT` was correctly enforced in the hook system (`uhook.c`, `uhook_legacy.c`, `uhook_attributes.c`) but bypassed in the `curl_multi_exec` code paths, causing unbounded span creation and OOM when curl multi is called repeatedly (e.g. AWS SDK SQS workloads on ECS Fargate).

Fixes APMS-18744.

### Root cause

- **`dd_multi_inject_headers()`** (`ext/handlers_curl.c`): PHP 8 curl multi creates one `DDTRACE_INTERNAL_SPAN` per handle on every `curl_multi_exec` call without checking the limit. With ~100k SQS calls (each using curl multi internally), this produced 100k+ unconstrained spans — the primary cause of the reported OOM.
- **`dd_inject_distributed_tracing_headers_multi()`** (`ext/handlers_curl_php7.c`): same issue on PHP 7.

### Fix

Add `!ddtrace_tracer_is_limited()` guards to both curl multi paths. When the span limit is reached, the paths fall through to header injection only (no span created) — distributed tracing propagation is preserved, only span tracking is skipped.

### DIFF
--- a/ext/ddtrace.c
+++ b/ext/ddtrace.c
@@ -3214,7 +3214,7 @@ static inline void dd_start_span(INTERNAL_FUNCTION_PARAMETERS) {
 
     ddtrace_span_data *span;
 
-    if (get_DD_TRACE_ENABLED() && !ddtrace_tracer_is_limited()) {
+    if (get_DD_TRACE_ENABLED()) {
         span = ddtrace_open_span(DDTRACE_USER_SPAN);
     } else {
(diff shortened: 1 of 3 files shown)


bwoebi (approved):
I think we should do this, yes.

PR #3692, opened 2026-03-06
ci(prof): fix "argument unused during compilation" warning

realFlowControl:
### Description

When compiling the profiler we see a lot of these lines:

```
warning: datadog-php-profiling@0.0.0: clang: warning: argument unused during compilation: '-fuse-ld=lld' [-Wunused-command-line-argument]
```

This stems from this line in `build.rs`:

https://github.com/DataDog/dd-trace-php/blob/0d2b2ffe3ff8c19c521abacfeba57d589a51fe9c/profiling/build.rs#L173

Afaik the `build.rs` is "only" used to compile and not for linking step, so this `-fuse-ld=lld` might not belong there anyway?

Additionally I am updating llvm to 19 for the prof-correctness job, just because it matches the `rustc -Vv` LLVM version:

```
$ rustc -vV   
rustc 1.84.1 (e71f9a9a9 2025-01-27)
binary: rustc
commit-hash: e71f9a9a98b0faf423844bf0ba7438f29dc27d58
commit-date: 2025-01-27
host: aarch64-apple-darwin
release: 1.84.1
LLVM version: 19.1.5
```

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/build.rs
+++ b/profiling/build.rs
@@ -170,7 +170,6 @@ fn build_zend_php_ffis(
                 .chain([Path::new("../zend_abstract_interface")])
                 .chain([Path::new("../")]),
         )
-        .flag_if_supported("-fuse-ld=lld")
         .flag_if_supported("-std=c11")
         .flag_if_supported("-std=c17");
     #[cfg(feature = "test")]
--- a/.github/workflows/prof_correctness.yml
+++ b/.github/workflows/prof_correctness.yml
@@ -47,13 +47,16 @@ jobs:
 
       - name: Build profiler
         run: |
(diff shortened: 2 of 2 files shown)


profiling/build.rs
```diff
@@ -170,7 +170,6 @@ fn build_zend_php_ffis(
                 .chain([Path::new("../zend_abstract_interface")])
                 .chain([Path::new("../")]),
         )
-        .flag_if_supported("-fuse-ld=lld")
```
morrisonlevi, line 173: Yes, I think this is correct. It should be added to rustflags or `.cargo/config.toml`, not the cc crate.

PR #3698, opened 2026-03-10
build(profiling): allow list of undefined symbols on macOS

morrisonlevi:
### Description

Uses a list of symbols on macOS to allow to be undefined. They are undefined because they are provided by the PHP engine, which does not provide a libzend.so or similar for us to link to. They are provided by the SAPI that our profiler is linked into.

Before this patch, we used linker flags `-undefined dynamic_lookup`, set in `profiling/.cargo/config.toml`. This allows all undefined symbols to pass through without issue. In several different branches, I experimented with using the [`no_panic` crate](https://docs.rs/no-panic/latest/no_panic/):

> If the function does panic (or the compiler fails to prove that the function cannot panic), the program fails to compile with a linker error that identifies the function name.

The problem is that this technique isn't compatible with allowing undefined symbols unconditionally!

Additionally, the list of symbols serves as an audit list.

### Reviewer checklist
- [x] Test coverage seems ok.

### DIFF
--- a/profiling/build.rs
+++ b/profiling/build.rs
@@ -54,6 +54,7 @@ fn main() {
     cfg_php_major_version(vernum);
     cfg_php_feature_flags(vernum);
     cfg_zts();
+    apple_linker_flags();
 }
 
 fn php_config_vernum() -> u64 {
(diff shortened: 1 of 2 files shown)


profiling/build.rs
```diff
@@ -460,3 +461,157 @@ int main() {
...
+/// which produces more unresolved symbols--we fall back to
+/// `-undefined dynamic_lookup` for debug builds.
+///
+/// To regenerate the symbol list after adding new PHP/C API calls:
+///   1. Edit the profile != release check to always be true e.g. if true {
+///   2. Build a release: `cargo build -p datadog-php-profiling --release`
+///   3. Extract symbols:
+///      `nm -u target/release/libdatadog_php_profiling.dylib | sort -u |
+///       grep -v '^$\|^ERROR\|dyld_stub' | sed 's/^/-Wl,-U,/' | tr '\n' ' '`
+///   4. Update the ALLOWED_UNDEFINED_SYMBOLS list below.
```
realFlowControl, line 476: Special thanks for this documentation 🎉

PR #3708, opened 2026-03-18
Adds container tags support for DBM

dubloom:
This PR implements the container tags support for DBM as described in this [RFC](https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M/edit?pli=1&tab=t.0).

Tracers communicate their containerID through a request http header that can be back propagate by the agent. This feature is implemented in the agent in this libdatadog PR: https://github.com/DataDog/libdatadog/pull/1700.

We use this container tags hash alongside process tags to compute a hash called `base_hash`. This base_hash is injected in DBM comments and DBM spans. 

Note that the base_hash should be computed as soon as process tags are available because if the tracer is not run within a container, it will never receive a container_tags_hash.

### DIFF
--- a/appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/docker/AppSecContainer.groovy
+++ b/appsec/tests/integration/src/main/groovy/com/datadog/appsec/php/docker/AppSecContainer.groovy
@@ -90,6 +90,7 @@ class AppSecContainer<SELF extends AppSecContainer<SELF>> extends GenericContain
         withEnv 'DD_TRACE_DEBUG', '1'
         withEnv 'DD_AUTOLOAD_NO_COMPILE', 'true' // must be exactly 'true'
         withEnv 'DD_TRACE_GIT_METADATA_ENABLED', '0'
+        withEnv 'DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED', '0'
         withEnv 'DD_INSTRUMENTATION_TELEMETRY_ENABLED', '1'
         // very verbose:
         withEnv '_DD_DEBUG_SIDECAR_LOG_METHOD', 'file:///tmp/logs/sidecar.log'


ext/fnv.c
```diff
@@ -0,0 +1,12 @@
+#include "fnv.h"
+
+uint64_t dd_fnv1_64(const unsigned char *data, size_t len) {
```
bwoebi: Can you move this into the rust code maybe? (components-rs or libdatadog)

ext/process_tags.c
```diff
@@ -168,6 +200,39 @@ static int cmp_process_tag_by_key(const void *tag1, const void* tag2) {
...
+        hash_value = dd_fnv1_64((unsigned char *)ZSTR_VAL(process_tags.serialized), ZSTR_LEN(process_tags.serialized));
+    }
+
+    zend_string *hash_value_str = strpprintf(0, "%" PRIu64, hash_value);
+    if (!hash_value_str) {
+        return;
+    }
+
+    process_tags.base_hash = zend_string_init(ZSTR_VAL(hash_value_str), ZSTR_LEN(hash_value_str), 1);
+    zend_string_release(hash_value_str);
```
bwoebi: You can use `smart_str_append_printf()` to directly write to a smart_str allocated via `smart_string_alloc(0, 21, 1)`, avoiding the intermittend string.

ext/agent_info.c
```diff
@@ -20,3 +21,18 @@ void ddtrace_agent_info_rinit() {
...
+
+void ddtrace_get_container_tags_hash(void) {
+    if (DDTRACE_G(agent_info_reader)) {
+        bool changed;
+        ddog_CharSlice hash = ddog_get_agent_info_container_tags_hash(
+            DDTRACE_G(agent_info_reader),
+            &changed
+        );
+        if (hash.len > 0) {
+            zend_string *hash_str = zend_string_init(hash.ptr, hash.len, 0);
```
bwoebi: Why not simply allocate a persistent string directly?

dubloom: addressed in [1317582](https://github.com/DataDog/dd-trace-php/pull/3708/commits/1317582a3f205c90946e9fb295e347eb93842a6e)

ext/fnv.c
dubloom: addressed in [1317582](https://github.com/DataDog/dd-trace-php/pull/3708/commits/1317582a3f205c90946e9fb295e347eb93842a6e)

ext/process_tags.c
dubloom: addressed in [1317582](https://github.com/DataDog/dd-trace-php/pull/3708/commits/1317582a3f205c90946e9fb295e347eb93842a6e)

PR #3710, opened 2026-03-19
Implement client-side trace stats computation

bm1549:
### Description

Activates Client-Side Trace Stats Computation via libdatadog's SpanConcentrator.

When `DD_TRACE_STATS_COMPUTATION_ENABLED=true`:
- Traces are routed through libdatadog's TraceExporter with stats enabled
- SpanConcentrator computes stats with 10-second buckets
- Stats sent to `/v0.6/stats` endpoint
- Top-level spans marked with `_dd.top_level` metric
- `Datadog-Client-Computed-Stats` and `Datadog-Client-Computed-Top-Level` headers sent

Default: disabled.

Part of cross-tracer CSS v1.3.0 alignment effort.

**Changed files:**
- `ext/configuration.h` — new `DD_TRACE_STATS_COMPUTATION_ENABLED` config flag
- `ext/sidecar.c` — `ddtrace_trace_exporter_init()` initializes TraceExporter when enabled
- `ext/auto_flush.c` — routes traces through TraceExporter when stats computation is enabled
- `ext/coms.c` — skips legacy HTTP flush when stats computation handles sending
- `ext/serializer.c` — sets `Datadog-Client-Computed-Stats` and `Datadog-Client-Computed-Top-Level` headers
- `components-rs/` — Rust bindings for `ddog_TraceExporter` and stats config
- `metadata/supported-configurations.json` — added `DD_TRACE_STATS_COMPUTATION_ENABLED` entry
- `tests/ext/` — three new phpt tests covering config, disabled-by-default, and top-level marking

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/ext/client_side_stats_config.phpt
+++ b/tests/ext/client_side_stats_config.phpt
@@ -12,5 +12,5 @@ echo dd_trace_env_config("DD_TRACE_STATS_COMPUTATION_ENABLED") ? "true" : "false
 echo "\n";
 
 ?>
---EXPECTF--
-DD_TRACE_STATS_COMPUTATION_ENABLED=true%A
+--EXPECT--
+DD_TRACE_STATS_COMPUTATION_ENABLED=true


components-rs/trace_exporter.h
```diff
@@ -0,0 +1,166 @@
+/**
```
bm1549, line 1: 🤖🇱🇺 Where does this come from? You know `cbindgen.toml` now includes `libdd-data-pipeline-ffi` — cbindgen will generate these types. And the upstream `cbindgen.toml` for that crate renames `ExporterError` → `ddog_TraceExporterError`, `ExporterErrorCode` → `ddog_TraceExporterErrorCode`, `ExporterResponse` → `ddog_TraceExporterResponse`. Your manual header uses `ddog_ExporterError` throughout — wrong names when the generated header lands. This file should not exist.

ext/auto_flush.c
```diff
@@ -46,7 +46,44 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
...
+                if (serialized_trace.len > limit) {
+                    LOG(ERROR, "Agent request payload of %zu bytes exceeds configured %zu byte limit; dropping request",
+                        serialized_trace.len, limit);
+                    success = false;
+                } else {
+                    ddog_ByteSlice trace_data = {
+                        .ptr = (const uint8_t *)serialized_trace.ptr,
+                        .len = serialized_trace.len,
+                    };
+                    ddog_ExporterError *err = ddog_trace_exporter_send(
```
bm1549, line 68: 🤖🇱🇺 `ddog_trace_exporter_send` calls `runtime.block_on(...)` internally — this is a synchronous, blocking network call. The PHP tracer does not do requests by itself on the main thread. This will manifest as pure latency on every request on the user side. Very bad thing :-D

Also: `trace_count=1` always — is the SpanConcentrator expecting all traces for the request together?

ext/auto_flush.c
```diff
@@ -56,8 +93,8 @@ ZEND_RESULT_CODE ddtrace_flush_tracer(bool force_on_startup, bool collect_cycles
                     .lang_vendor = DDOG_CHARSLICE_C_BARE(""),
                     .tracer_version = DDOG_CHARSLICE_C_BARE(PHP_DDTRACE_VERSION),
                     .lang_version = php_version_rt,
-                    .client_computed_top_level = false,
-                    .client_computed_stats = !get_global_DD_APM_TRACING_ENABLED(),
+                    .client_computed_top_level = get_global_DD_TRACE_STATS_COMPUTATION_ENABLED(),
+                    .client_computed_stats = get_global_DD_TRACE_STATS_COMPUTATION_ENABLED() || !get_global_DD_APM_TRACING_ENABLED(),
```
bm1549, line 97: 🤖🇱🇺 If `DD_TRACE_STATS_COMPUTATION_ENABLED=true` but `ddtrace_trace_exporter == NULL` (init failed), we fall through to this sidecar path and tell it `client_computed_stats = true` — but the client didn't compute any stats. Should only set this when `ddtrace_trace_exporter != NULL`.

ext/sidecar.c
```diff
@@ -246,6 +248,80 @@ bool ddtrace_sidecar_maybe_enable_appsec(bool *appsec_activation, bool *appsec_c
...
+    if (err) {
+        LOG(ERROR, "Failed to set trace exporter URL: %s", err->msg);
+        ddog_trace_exporter_error_free(err);
+        ddog_trace_exporter_config_free(config);
+        free(url);
+        return;
+    }
+    free(url);
+
+    ddog_trace_exporter_config_set_language(config, DDOG_CHARSLICE_C("php"));
```
bm1549, line 277: 🤖🇱🇺 All of these return `ddog_ExporterError *` — dropped on the floor. Memory leak on error.

ext/sidecar.c
```diff
@@ -25,6 +26,7 @@ ZEND_EXTERN_MODULE_GLOBALS(ddtrace);
 ddog_Endpoint *ddtrace_endpoint;
 ddog_Endpoint *dogstatsd_endpoint; // always set when ddtrace_endpoint is set
 struct ddog_InstanceId *ddtrace_sidecar_instance_id;
+ddog_TraceExporter *ddtrace_trace_exporter;
```
bm1549, line 29: 🤖🇱🇺 The TraceExporter creates a tokio multi-thread runtime with worker threads (`worker_threads(1)`). After `fork()` in FPM, those threads are gone in the child but this handle remains. The child will deadlock or crash trying to `block_on` on a dead runtime.

Need to re-initialize the exporter post-fork, similar to the `ddtrace_sidecar_ensure_active` pattern. The specific scenario: the master initializes the exporter, forks workers — each worker needs a fresh exporter with a fresh runtime. (Please add a test for this.)

PR #3711, opened 2026-03-19
ExecSolib strategy: make ddtrace.so directly executable

cataphract:
Introduce the ExecSolib spawn strategy by embedding an ELF entry point (_dd_solib_start) into ddtrace.so itself.

ddtrace.so becomes pie executable and runs without the dynamic linker. After self-relocation:

* it loads the trampoline into memory, but doesn't execute it yet
* it copies ddtrace.so (/proc/self/exe) into a memfd and massages it so its loading by the dynamic linker without php doesn't fail.
* replaces in the command line all occurrences to ddtrace.so to /proc/self/fd/<mem fd>, so that dlopen from the trampoline will use the massaged version
* loads the dynamic linker into memory
* jumps into the dynamic linker after massaging the auxiliary vector so that ld.so executes the loaded trampoline

tested on glibc and musl on linux aarch64

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/solib_bootstrap.c
```diff
@@ -0,0 +1,1182 @@
...
+// sections of the memfd via pwrite:
+//
+//   Ste 1: .dynamic section:
+//     neutralize DT_NEEDED (no PHP deps loaded), DT_INIT/FINI (no PHP init
+//     called), DT_VERNEED/VERSYM (no version checking), DT_BIND_NOW / DF flags
+//     (lazy PLT ok)
+//
+//   Step 2: .dynsym table:
+//     every STB_GLOBAL/SHN_UNDEF symbol → STB_WEAK, so unresolved PHP symbols
+//     silently resolve to NULL rather than aborting dlopen
```
bwoebi, line 836: Is there a fundamental blocker to just applying these changes directly to the `.so` after _build_? Rather than doing this at sidecar startup. If we did that we wouldn't even need this memfd at all, and a lot less logic - just redirecting main then?

Or would just this fully neutralize/break the normal extension too?

cataphract: Making the imports weak I think wouldn't break anything, though it could result in detecting problems with missing symbols later, perhaps in paths we're not testing (this is already a problem with lazy loading of symbols in glibc -- not musl, but here the null pointer dereference may make the problem less obvious).

The initializers may be a bigger problem, though that can be worked around by calling something similar lazily/later.

Ultimately, though, I think the split binary with ssi_entry.c is a much cleaner solution by avoiding linking the extension and sidecar into the same binary.

bwoebi: At least ddtrace.so has no meaningful initializers at all (I see only `__gmon_start__@plt`, but that's irrelevant).

bwoebi: The _only_ thing is really missing symbols, which would manifest as null pointer derefs, right?
(But we really shouldn't have missing symbols in the first place, because why would we, unless something went wrong at compilation - and we only need to weaken PHP's symbols, none other, where we anyway build against a very specific ABI)

cataphract: right:

```
$ cat weak_test.c
#include <stdio.h>

void __attribute__((weak)) missing_func(void);

int main(void) {
    fprintf(stderr, "Address of missing_func: %p\n", (void *)missing_func);

    if (missing_func) {
        missing_func();
    } else {
        fprintf(stderr, "missing_func is NULL — guarded call skipped\n");
    }

    fprintf(stderr, "Calling missing_func unconditionally...\n");
    missing_func();
    fprintf(stderr, "Survived (unexpected)\n");
    return 0;
}
$ gcc -O0 weak_test.c
$ ./a.out
Address of missing_func: (nil)
missing_func is NULL — guarded call skipped
Calling missing_func unconditionally...
Segmentation fault         (core dumped) ./a.out
```

cataphract:
Abandoned in favor of https://github.com/DataDog/dd-trace-php/pull/3869

PR #3712, opened 2026-03-19
refactor(profiling): shrink max file and function name

morrisonlevi:
### Description

The maximum length is now 16,383, aka `(1 << 14) - 1`, or one less than 16 KiB. This has nice properties:

 1. It always encodes to 1 or 2 bytes in protobuf's varint encoding.
 2. It's a single byte "wasted" to use a 16 KiB buffer, which is a multiple of common page sizes, which could be nice for a reusable buffer.

It's also plenty long--the UI is not going to nicely show function or file names that are this long either.

This doesn't check all strings, just the function and file names. I am thinking about enforcing it for all strings but this needs more discussion, such as for an exception message or a fatal error message.

I am hoping to use this restriction in future work, so shipping this change ahead of time can partly derisk it. We shouldn't generally lose any behavior here either, as we don't drop samples or other behaviors, we just substitute with `"[suspiciously large string]"`.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/src/profiling/stack_walking.rs
+++ b/profiling/src/profiling/stack_walking.rs
@@ -16,10 +16,13 @@ use crate::bindings::{
 const COW_PHP_OPEN_TAG: Cow<str> = Cow::Borrowed("<?php");
 const COW_TRUNCATED: Cow<str> = Cow::Borrowed("[truncated]");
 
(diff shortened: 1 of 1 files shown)


bwoebi (approved):
Looks correct to me

PR #3716, opened 2026-03-20
Expand appsec integrations to Laminas Framework

estringana:
### Add AAP support to Laminas framework

This PR introduces AAP support to the Laminas framework, enabling improved observability and authentication-related event tracking.

#### Features added
- **Provide `http.route`**  
  Captures and exposes the resolved route for incoming requests.

- **Collect endpoints at startup**  
  Gathers and registers available endpoints during application bootstrap.

- **Login success event**  
  Emits an event when a user successfully authenticates.

- **Login failure event**  
  Emits an event when authentication fails.

- **Authenticated request events**  
  Tracks requests performed by authenticated users.

- **Collect path parameters**  
  Extracts and includes route path parameters in the collected data.

#### Limitations / Missing features
- **Signup events**  
  Not implemented because Laminas does not provide a built-in signup mechanism. This functionality is application-specific and must be implemented by developers if needed.

### Reviewer checklist
- [ ] Test coverage seems ok.


cataphract:
@codex review

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
```diff
@@ -0,0 +1,176 @@
...
+@Testcontainers
+@EnabledIf('isExpectedVersion')
+@TestMethodOrder(MethodOrderer.OrderAnnotation)
+class Laminas33Tests {
+
+    /**
+     * Laminas MVC 3.3.x supports PHP 7.3–8.1 per composer constraints in www/laminas33.
+     */
+    static boolean expectedVersion =
+            ['7.3', '7.4', '8.0', '8.1'].contains(getPhpVersion()) && !getVariant().contains('zts')
```
cataphract, line 36: Shouldn't we test with a more recent version? This seems very old. I fyou want you can test the oldest supported and the newest supported (make a base class / trait with the tests and then extend in two separate test classes)

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+            return $base . $child;
+        }
+
+        if ($route instanceof \Laminas\Router\Http\TreeRouteStack && $hasChild) {
+            return self::httpRouteTemplateFromNamedRouteStack($route, $segments[1]);
+        }
+
+        if ($hasChild) {
+            return self::httpRouteTemplateFromMatchedRoute($route, null);
+        }
```
cataphract, line 1188: Wouldn't this cause the route name to be set multiple times? For the individual parts and for the high level router? I think you should not call this method unless it's the top level.

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
```diff
@@ -0,0 +1,176 @@
...
+
+import static com.datadog.appsec.php.integration.TestParams.getPhpVersion
+import static com.datadog.appsec.php.integration.TestParams.getVariant
+import static java.net.http.HttpResponse.BodyHandlers.ofInputStream
+import static java.net.http.HttpResponse.BodyHandlers.ofString
+
+@Testcontainers
+@EnabledIf('isExpectedVersion')
+@TestMethodOrder(MethodOrderer.OrderAnnotation)
+class Laminas33Tests {
```
cataphract, line 30: You don't have tests for interesting routers. Try to test the full hierarchy:

```
  RouteInterface                        (single method: match())
  └── RouteStackInterface               (adds: addRoute, addRoutes, removeRoute, setRoutes)
      └── SimpleRouteStack              (concrete; adds getRoutes() + PriorityList storage)
          └── TreeRouteStack            (adds path-offset matching, defaultParams injection)
              ├── Part                  (a "node" in the tree: base route + lazy children)
              └── Chain                 (concatenates multiple route stages sequentially)

  HttpRouteInterface extends RouteInterface
      ├── Literal                       (exact string match)
      ├── Segment                       (`:param` placeholders, optional parts)
      ├── Regex                         (regex match with named captures)
      ├── Hostname                      (matches the Host header)
      ├── Method                        (matches HTTP verb)
      ├── Scheme                        (matches http/https)
      ├── Wildcard                      (key/value pairs after a prefix)
      └── Placeholder                   (always matches, no constraints)
```

For instance, I can tell that chain doesn't work. This test fails the final assertion

```groovy
    @Test
    @Order(8)
    void 'nested Part and Chain routes produce correct http route'() {
        HttpRequest nestedReq = container.buildReq('/resource/42/99').GET().build()
        Trace nestedTrace = container.traceFromRequest(nestedReq, ofString()) { HttpResponse<String> resp ->
            assert resp.statusCode() == 200
        }
        assert nestedTrace.first().meta.'http.route' == '/resource/:resourceId/:subId'

        HttpRequest chainReq = container.buildReq('/chain/abc').GET().build()
        Trace chainTrace = container.traceFromRequest(chainReq, ofString()) { HttpResponse<String> resp ->
            assert resp.statusCode() == 200
        }
        assert chainTrace.first().meta.'http.route' == '/chain/:chainId'
    }
```

with

```
diff --git a/appsec/tests/integration/src/test/www/laminas33/module/Application/config/module.config.php b/appsec/tests/integration/src/test/www/laminas33/module/Application/config/module.config.php
index ee2635397..c9c91f285 100644
--- a/appsec/tests/integration/src/test/www/laminas33/module/Application/config/module.config.php
+++ b/appsec/tests/integration/src/test/www/laminas33/module/Application/config/module.config.php
@@ -56,6 +56,60 @@ return [
                     ],
                 ],
             ],
+            'nested_resource' => [
+                'type' => Literal::class,
+                'options' => [
+                    'route' => '/resource',
+                    'defaults' => [
+                        'controller' => DynamicPathController::class,
+                        'action' => 'index',
+                    ],
+                ],
+                'may_terminate' => true,
+                'child_routes' => [
+                    'item' => [
+                        'type' => Segment::class,
+                        'options' => [
+                            'route' => '/:resourceId',
+                            'defaults' => [
+                                'controller' => DynamicPathController::class,
+                                'action' => 'index',
+                            ],
+                        ],
+                        'may_terminate' => true,
+                        'child_routes' => [
+                            'sub' => [
+                                'type' => Segment::class,
+                                'options' => [
+                                    'route' => '/:subId',
+                                    'defaults' => [
+                                        'controller' => DynamicPathController::class,
+                                        'action' => 'index',
+                                    ],
+                                ],
+                            ],
+                        ],
+                    ],
+                ],
+            ],
+            'chained_resource' => [
+                'type' => Literal::class,
+                'options' => [
+                    'route' => '/chain',
+                ],
+                'chain_routes' => [
+                    [
+                        'type' => Segment::class,
+                        'options' => [
+                            'route' => '/:chainId',
+                            'defaults' => [
+                                'controller' => DynamicPathController::class,
+                                'action' => 'index',
+                            ],
+                        ],
+                    ],
+                ],
+            ],
         ],
     ],
     'controllers' => [
```

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
```diff
@@ -0,0 +1,176 @@
...
+            assert re.statusCode() == 403
+            assert re.body().toLowerCase().contains('blocked')
+        }
+
+        Span span = trace.first()
+        assert span.metrics.'_dd.appsec.enabled' == 1.0d
+        assert span.metrics.'_dd.appsec.waf.duration' > 0.0d
+        assert span.meta.'_dd.appsec.event_rules.version' != ''
+        assert span.meta.'appsec.blocked' == 'true'
+    }
```
cataphract, line 250: no tests for http.route

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -274,9 +274,56 @@ static function (SpanData $span) use ($controller, $action) {
...
+
+                // Push path params to appsec
+                if (function_exists('\datadog\appsec\push_addresses')) {
+                    $params = $routeMatch->getParams();
+                    // Filter out the framework-specific params (controller, action)
+                    $pathParams = array_diff_key($params, array_flip(['controller', 'action']));
+                    if (count($pathParams) > 0) {
+                        \datadog\appsec\push_addresses(["server.request.path_params" => $pathParams]);
+                    }
+                }
```
cataphract: This should fire only for the top level router, otherwise you're pushing to the WAF multiple times.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -274,9 +274,56 @@ static function (SpanData $span) use ($controller, $action) {
...
+                    if ($httpRoute !== null && $httpRoute !== '') {
+                        $rootSpan->meta[Tag::HTTP_ROUTE] = $httpRoute;
+                    }
+                }
+
+                // Push path params to appsec
+                if (function_exists('\datadog\appsec\push_addresses')) {
+                    $params = $routeMatch->getParams();
+                    // Filter out the framework-specific params (controller, action)
+                    $pathParams = array_diff_key($params, array_flip(['controller', 'action']));
```
cataphract: This filtering looks insufficient (see __NAMESPACE__, __CONTROLLER__, middleware). You should also filter non-strings.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+            null,
+            static function (HookData $hook) {
+                $result = $hook->returned;
+
+                if (!$result instanceof \Laminas\Authentication\Result) {
+                    return;
+                }
+
+                if ($result->isValid()) {
+                    self::trackUserLoginSuccess($result, $hook->args[0] ?? null);
```
cataphract: Make sure you test this in the integration tests with `authenticate` being called both with null and an adapter instance for both the success and the failure cases. Claude flagged this as problematic, but it was too much in the weeds for me.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+                    $userLogin,
+                    $userExists,
+                    []
+                );
+            }
+        );
+
+        hook_method(
+            'Laminas\Authentication\AuthenticationService',
+            'getIdentity',
```
cataphract, line 714: You prob should dedup these calls to avoid calling track_authenticated_user_event_automated multiple times per request. For instance, you call this method yourself from another hook.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+                continue;
+            }
+            $seen[$key] = true;
+            $out[] = $row;
+        }
+
+        return $out;
+    }
+
+    private static function walkRouteStackCollectEndpointRows(
```
cataphract, line 1127: Part children are only populated on first match(). The onRoute hook fires after one path through the route tree is matched, so only the subtree along the matched path is materialized. You need to force this materialization.

See the in the other comment when I added extra routes with child routes. Those are not found by the `Endpoints are sent` test.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+        $rootRouter,
+        $currentStack,
+        string $namePrefix,
+        array &$rows
+    ): void {
```
cataphract, line 1132: This does not seem to catch the `Chain` router I added. The code only recurses into `Part`. Note that Chain is  fundamentally different from Part in this context - its children are sequential stages that together form a single endpoint URL, not separate sibling  endpoints.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+        foreach ($currentStack->getRoutes() as $name => $route) {
+            $qualifiedName = $namePrefix === '' ? (string) $name : $namePrefix . '/' . $name;
+            $path = self::httpRouteTemplateFromNamedRouteStack($rootRouter, $qualifiedName);
+            if ($path !== null && $path !== '') {
+                $method = 'GET';
```
cataphract: This hardcoded GET is not correct. Make sure the integration tests include non-GET endpoints.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+        return $out;
+    }
+
+    private static function walkRouteStackCollectEndpointRows(
+        $rootRouter,
+        $currentStack,
+        string $namePrefix,
+        array &$rows
+    ): void {
+        foreach ($currentStack->getRoutes() as $name => $route) {
```
cataphract: The code guards on instanceof RouteStackInterface but then calls->getRoutes(). getRoutes() is defined only on the concrete SimpleRouteStack

cataphract (changes requested):

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
estringana: Done

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
estringana: Done

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
estringana: Done

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Laminas auth gives no a ideal method to hook which ensures it won't be called twice during a request. It's a very light service.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Done

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Sorted

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Nice catch. Sorted

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Done

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: It's all behind an adapter not null check

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Solved

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: sorted

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +642,531 @@ static function ($This, $scope, $args) {
...
+
+                if ($adapter && method_exists($adapter, 'getIdentity')) {
+                    $userLogin = $adapter->getIdentity();
+                }
+
+                $userExists = ($code === \Laminas\Authentication\Result::FAILURE_CREDENTIAL_INVALID);
+
+                \datadog\appsec\track_user_login_failure_event_automated(
+                    $userLogin,
+                    $userLogin,
```
cataphract: I believe user_id should NOT be populated unless both a) the user exists and b) login actually doubles as user id. For instance, the symfony integration sends null in user_id.

Additionally, `AutomatedLoginEventsTestSuite::testUserLoginFailureEvent` only asserts `userLogin == $email`, not the userId

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +642,531 @@ static function ($This, $scope, $args) {
...
+        if (\DDTrace\are_endpoints_collected()) {
+            return;
+        }
+        if (!($rootRouter instanceof \Laminas\Router\SimpleRouteStack)) {
+            return;
+        }
+        $endpoints = self::collectLaminasRouteEndpointRows($rootRouter, $rootRouter, '');
+        foreach ($endpoints as $row) {
+            \DDTrace\add_endpoint($row['path'], 'http.request', $row['resourceName'], $row['method']);
+        }
```
cataphract, line 1018: The new `\DDTrace\flush_endpoints()` is not called, unlike in Symfony/Laravel

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -274,9 +274,69 @@ static function (SpanData $span) use ($controller, $action) {
...
+                    && !($this instanceof \Laminas\Router\Http\Part)
+                    && $routeName !== null
+                    && $routeName !== ''
+                ) {
+                    $params = $routeMatch->getParams();
+                    $pathParams = array_diff_key(
+                        $params,
+                        array_flip([
+                            'controller',
+                            'action',
```
cataphract, line 305: At least `action` can be user-controlled. In Laminas, segment routes can capture it straight from the URL. The default config that ships with laminas-mvc-skeleton has:
```php
'application' => [
    'type' => Segment::class,
    'options' => [
        'route' => '/application[/:action]',
        ...
    ],
],
```

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
```diff
@@ -0,0 +1,206 @@
...
+@Testcontainers
+@EnabledIf('isExpectedVersion')
+@TestMethodOrder(MethodOrderer.OrderAnnotation)
+class Laminas33Tests {
+
+    /**
+     * Laminas MVC 3.3.x supports PHP 7.3–8.1 per composer constraints in www/laminas33.
+     */
+    static boolean expectedVersion =
+            ['7.3', '7.4', '8.0', '8.1', '8.2', '8.3', '8.4'].contains(getPhpVersion()) && !getVariant().contains('zts')
```
cataphract: This does not match the comment

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +642,531 @@ static function ($This, $scope, $args) {
...
+        if (!($stack instanceof \Laminas\Router\SimpleRouteStack)) {
+            return null;
+        }
+
+        return $stack->getRoute($name);
+    }
+
+    private static function laminasMaterializePartChildRoutes(\Laminas\Router\Http\Part $part): void
+    {
+        $rp = new \ReflectionProperty(\Laminas\Router\Http\Part::class, 'childRoutes');
```
cataphract: prefer to use the instance based api, because it's more resilient to where `childRoutes` is defined.

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
cataphract: I mean a more recent _Laminas_ version. The 3.3.x branch looks very old (last update 6 years ago)

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
cataphract: I see no tests for Regex, Hostname, Scheme, Wildcard, Placeholder

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
cataphract: I think middleware and non-string values still leak

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
cataphract: `track_authenticated_user_event_automated` calls the WAF. There's nothing light about it, esp. for something that will trigger for all authenticated requests. You need to implement some sort of deduping

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
cataphract: The point is that you're not capturing the data in that case:

```php
$authAdapter->setIdentity($email);
$authAdapter->setCredential(md5($password));
$this->authService->setAdapter($authAdapter);
$result = $this->authService->authenticate();
```

You're just passing null. You should fetch the adapter from the auth service in that case.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +629,447 @@ static function ($This, $scope, $args) {
...
+                if (!function_exists('\datadog\appsec\track_user_login_failure_event_automated')) {
+                    return;
+                }
+
+                $code = $result->getCode();
+                $adapter = $hook->args[0] ?? null;
+                $userLogin = null;
+
+                if ($adapter && method_exists($adapter, 'getIdentity')) {
+                    $userLogin = $adapter->getIdentity();
```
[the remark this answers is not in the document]
cataphract: This is the same as I raised below

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +642,531 @@ static function ($This, $scope, $args) {
...
+        if (is_object($matchedRoute)) {
+            if (method_exists($matchedRoute, 'getSpec')) {
+                $routeSpec = $matchedRoute->getSpec();
+                if (is_string($routeSpec) && $routeSpec !== '') {
+                    return $routeSpec;
+                }
+            }
+            
+            if (
+                method_exists($matchedRoute, 'getRoute')
```
cataphract, line 896: In my tests, this route didn't show in the catalog:

```php
            'tenant_with_profile' => [
                'type' => Hostname::class,
                'options' => [
                    'route' => ':tenant.example.com',
                ],
                'may_terminate' => false,
                'child_routes' => [
                    'profile' => [
                        'type' => Literal::class,
                        'options' => [
                            'route' => '/profile',
                            'defaults' => [
                                'controller' => DynamicPathController::class,
                                'action' => 'index',
                            ],
                        ],
                    ],
                ],
            ],
```

try adding an integration test for it. To test actually hitting it, rather than the catalog, you also need `-Djdk.httpclient.allowRestrictedHeaders=Host`

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +642,531 @@ static function ($This, $scope, $args) {
...
+        $chainRoutes = $rp->getValue($chain);
+        if ($chainRoutes !== null) {
+            $chain->addRoutes($chainRoutes);
+            $rp->setValue($chain, null);
+        }
+
+        return $chain->getRoutes();
+    }
+
+    private static function extractHttpVerbFromRoute($route): string
```
cataphract: This doesn't support multiple verbs:

```php
'multi_verb_test' => [
    'type' => Literal::class,
    'options' => [
        'route' => '/multi-verb',
        'defaults' => [
            'controller' => DynamicPathController::class,
            'action' => 'index',
        ],
    ],
    'may_terminate' => false,
    'child_routes' => [
        'read' => [
            'type' => Method::class,
            'options' => ['verb' => 'GET,HEAD,OPTIONS'],
        ],
        'write' => [
            'type' => Method::class,
            'options' => ['verb' => 'POST,PUT'],
        ],
    ],
],
```

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -274,9 +274,69 @@ static function (SpanData $span) use ($controller, $action) {
                 }
                 $rootSpan->meta['laminas.route.name'] = $routeName;
                 $rootSpan->meta['laminas.route.action'] = "$controller@$action";
+
+                if (
+                    $this instanceof \Laminas\Router\Http\TreeRouteStack
+                    && $routeName !== null
+                    && $routeName !== ''
+                ) {
```
cataphract, line 287: This prob should have

```php
 && !($this instanceof \Laminas\Router\Http\Part
```

like on line 292. Or better, just eval the condition once (`$isTopLevelRouterMatch = ...`).

cataphract (changes requested):

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Can you expand your point on this comment? Action should be sent as part of params

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Nice catch. Done

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Nice catch. This is not coped with on Laravel neither

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: It works now

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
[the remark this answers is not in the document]
estringana: Sorted

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: I have adapted this to send user login and id

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: I think I got it. I have just added a route like this and test it

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Done

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
estringana: I added those routes for tests

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Done. I did it on tags.c. If during the same request it calls more than 1 to track_authenticated_user_event_automated with same id, it won't propagate the call to the helper

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
estringana: Reverted. We can do in another ticket anohter version of Laminas. I used this one which is the one tested on apm

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laminas33Tests.groovy
estringana: We can do this in another ticket

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: I am filtering out more params now.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: This has been fixed and tested

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -582,10 +646,542 @@ static function ($This, $scope, $args) {
...
+    }
+
+    private static function extractHttpVerbFromRoute($route): array
+    {
+        if ($route instanceof \Laminas\Router\Http\Chain) {
+            foreach (self::laminasGetChainRoutes($route) as $chainRoute) {
+                $methods = self::extractHttpVerbFromRoute($chainRoute);
+                if (sizeof($methods) == 0 && $methods[0] !== 'GET') {
+                    return $methods;
+                }
```
cataphract: This doesn't look correct. If size is 0, `$methods[0]` is undefined.

appsec/src/extension/tags.c
```diff
@@ -356,9 +357,15 @@ void dd_tags_rinit(void)
...
-    _auth_user_event_automated_last_user_id = NULL;
+    // Just in case (e.g. rinit without a prior rshutdown in edge paths)...
+    if (_event_user_id) {
+        zend_string_release(_event_user_id);
+        _event_user_id = NULL;
+    }
+    if (_auth_user_event_automated_last_user_id) {
+        zend_string_release(_auth_user_event_automated_last_user_id);
+        _auth_user_event_automated_last_user_id = NULL;
+    }
```
cataphract: This releases are not correct because the strings are not persistent. Either make them persistent or set it to NULL unconditionally.

estringana: I nulled them as they were

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
estringana: Yeah, sorted

cataphract (approved):

PR #3721, opened 2026-03-24
Fix _dd.p.ksr scientific notation for very small sampling rates

bm1549:
### Description

Fix `_dd.p.ksr` span tag formatting for very small sampling rates. Rates below 0.001 were formatted using C's `%.6g` format which outputs scientific notation (e.g. `0.000001` → `"1e-06"`). Changed to explicit integer-level rounding via `floor(rate * 1e6 + 0.5)` and `%.6f` formatting to always produce decimal notation with up to 6 decimal digits, trailing zeros stripped.

Fixes APMAPI-1869

**Related PRs:**
- dd-trace-py: https://github.com/DataDog/dd-trace-py/pull/17086
- dd-trace-rb: https://github.com/DataDog/dd-trace-rb/pull/5497
- dd-trace-js: https://github.com/DataDog/dd-trace-js/pull/7846
- system-tests: https://github.com/DataDog/system-tests/pull/6466

### Testing

- Updated test description in `tests/ext/priority_sampling/028-ksr-tag-formatting.phpt`
- Added new test `tests/ext/priority_sampling/029-ksr-tag-small-rate-formatting.phpt` verifying:
  - Rate 0.000001 → `"0.000001"` (decimal notation, not `"1e-06"`)
  - No scientific notation characters in output

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/ext/priority_sampling/030-ksr-tag-rounding-boundary.phpt
+++ b/tests/ext/priority_sampling/030-ksr-tag-rounding-boundary.phpt
@@ -1,0 +1,19 @@
+--TEST--
+_dd.p.ksr rounds small rates correctly at 6-decimal boundary
+--ENV--
+DD_TRACE_SAMPLING_RULES=[{"sample_rate": 0.00000051}]
+DD_TRACE_GENERATE_ROOT_SPAN=1
+--FILE--
+<?php
+$root = \DDTrace\root_span();
+
+\DDTrace\get_priority_sampling();
+
+$ksr = isset($root->meta["_dd.p.ksr"]) ? $root->meta["_dd.p.ksr"] : "-";
+echo "_dd.p.ksr = ", $ksr, "\n";
+// 0.00000051 rounds up to 0.000001 at 6 decimal places
+echo "no_sci_notation = ", (strpos($ksr, 'e') === false) ? "true" : "false", "\n";
+?>
+--EXPECT--
+_dd.p.ksr = 0.000001
+no_sci_notation = true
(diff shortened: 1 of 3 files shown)


ext/priority_sampling/priority_sampling.c
```diff
@@ -65,8 +66,20 @@ static void dd_update_decision_maker_tag(ddtrace_root_span_data *root_span,
 static void dd_update_knuth_sampling_rate_tag(ddtrace_root_span_data *root_span, double sample_rate) {
     zend_array *meta = ddtrace_property_array(&root_span->property_meta);
 
+    // Round to 6 decimal places at integer level to avoid IEEE 754 precision issues,
+    // then format with fixed-point notation (never scientific notation).
+    double rounded = floor(sample_rate * 1e6 + 0.5) / 1e6;
```
bwoebi, line 71: Are you sure?
snprintf() should round itself.

Doing this should covert `0.0000001` to `0.000001` instead of `0`

bwoebi: ~Or maybe that's just PHP and not snprintf.~ - a quick google lets me find https://stackoverflow.com/a/24120950.

bm1549: 0.0000001 (1e-7) should go to 0 since we only want 6 decimal places

bwoebi: Ah, but the rounding is implementation defined, so actually libc defined.

bwoebi: Ah, that's why floor(). Got it.

bm1549: The test cases are chosen deliberately to highlight the edge cases I wanted to cover

```
0.000001   -> 0.000001
0.00000051 -> 0.000001
0.0000001  -> 0

bwoebi: Yes, I just was sort of missing the floor and just seeing the 0.5, and confused. You could have used ap_php_snprintf(), which has a fixed behaviour and you use the rounding as described there, but it's good as is.

PR #3722, opened 2026-03-24
ci(prof): update profiling ASAN runs

realFlowControl:
### Description

This PR:
- adds PHP 8.5 to ASAN tests
- fixes cache usage
- Updates `checkout` and `cache` action versions
- switch to use Datadog private runners to run ASAN not only on amd64 but also aarch64
- instruct `run-tests.php` to run tests in parallel
- don't reinstall nightly Rust, but use the nightly that is baked into the bookworm image

... bringing down CI time from ~10 minutes to ~5 minutes for the ASAN jobs.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/prof_asan.yml
+++ b/.github/workflows/prof_asan.yml
@@ -62,7 +62,6 @@ jobs:
           set -eux
           switch-php nts-asan
           cd profiling/tests
-          jobs=$(getconf _NPROCESSORS_ONLN)
           cp -v $(php-config --prefix)/lib/php/build/run-tests.php .
           export DD_PROFILING_OUTPUT_PPROF=/tmp/pprof
-          php run-tests.php -j"${jobs}" --show-diff --asan -d extension=datadog-profiling.so phpt
+          php run-tests.php -j$(nproc) --show-diff --asan -d extension=datadog-profiling.so phpt


.github/workflows/prof_asan.yml
```diff
@@ -5,65 +5,64 @@ on:
...
-            ~/.cargo/git/db/
-            /tmp/build-cargo/
-          key: ${{ runner.os }}-cargo-asan-${{ hashFiles('**/Cargo.lock') }}
-
       - name: Run phpt tests
         run: |
           set -eux
           switch-php nts-asan
           cd profiling/tests
+          jobs=$(getconf _NPROCESSORS_ONLN)
```
morrisonlevi: Is this actually what we want? For instance, on gitlab runners depending on your QOS it's definitely _not_ what you want, it will report a large number even though you've only requested 4 CPUs.

realFlowControl: It is working correct though for the GitHub Actions where we are using this right now.
We could also just hard code `8` in this case, because the amd64 and aarch64 runners both have 8 cores currently.

realFlowControl: Do you know how `nproc` behaves in the Gitlab runners with regards to requested CPU cores?

morrisonlevi: `nproc` is also wrong, this is why we use variables like `KUBERNETES_CPU_REQUEST`. See `"profiling tests"` in `.gitlab/generate-profiler.php`.

realFlowControl: Found it https://github.com/DataDog/dd-trace-php/blob/a1bb0382c570a3e3d8e2c118eb5d95e146a7df3c/.gitlab/generate-profiler.php#L20-L22

I guess we do not have `KUBERNETES_CPU_REQUEST` in GitHub Actions, so if we do not want the `getconf` there, we could switch to `nproc` or hard code?

morrisonlevi: Is there a reason you reached for `getconf _NPROCESSORS_ONLN` instead of `nproc`? I don't have an objection, just curious. I can't seem to find equivalent variables for GHA, so perhaps `nproc` or similar is the best we can do.

realFlowControl: No, no real reason. The chain of events was:
- "it should be called `ncores` or something", but that did not work
- I search in the [GitLab CI yaml](https://github.com/DataDog/dd-trace-php/blob/a1bb0382c570a3e3d8e2c118eb5d95e146a7df3c/.gitlab/generate-profiler.php#L60) but we are not running tests in parallel there
- so I fired up a google search and it showed me `getconf _NPROCESSORS_ONLN` and it worked
- then I discovered it was `nproc` but I left it anyway because it was correct

PR #3724, opened 2026-03-24
Fix SSI crashing on apache reload; add SSI int tests for appsec

cataphract:
### Description

* Adding SSI tests to the appsec integration tests
* Fix crash that was found after apache reload

### Reviewer checklist
- [ ] Test coverage seems ok.


loader/dd_library_loader.c
```diff
@@ -974,6 +994,22 @@ static int ddloader_zend_extension_startup(zend_extension *ext) {
...
 }
 
+static void ddloader_zend_extension_shutdown(zend_extension *ext) {
+    UNUSED(ext);
+    /* Close injected extension handles first (ddtrace.so, ddappsec.so, etc.),
+     * then libddtrace_php.so which they may depend on. */
+    for (unsigned int i = 0; i < sizeof(ddloader_injected_ext_config) / sizeof(ddloader_injected_ext_config[0]); ++i) {
+        if (ddloader_injected_ext_config[i].so_handle) {
+            DL_UNLOAD(ddloader_injected_ext_config[i].so_handle);
+            ddloader_injected_ext_config[i].so_handle = NULL;
```
bwoebi, line 1004: That's odd. Isn't ddtrace_shutdown going to be executed _after_ the loader extension shutdown? zend_extension is destroyed in order of registration according to `zend_shutdown_extensions`.

cataphract: Yeah, unfortunately I think this latest iteration is not crashing just by accident. The mshutdown of ddtrace doesn't crash because ddtrace.so is still loaded even after the dlclose, but of course that's just a variant on the initial bug. I can reproduce the crash on 8.3-debug

PR #3725, opened 2026-03-24
Move appsec communication to sidecar

cataphract:
* **Remove the separate unix socket used by the appsec helper; instead, appsec messages now go through the sidecar protocol**. This added some significant overhead, so there is also a change on libdatadog to avoid having doing extra copies by splitting the the client and server representation of messages. The performance is still a bit worse than the old implementation. Part of it explained by message overhead, another part is that messages received are not processed inline, instead, the model of one tokio task per client is kept; while before this task was also responsible for receiving its own data, from its own socket, now there is additional scheduling/moving data around.

* **Build the appsec helper are part of sidecar**. There is no separate .so for the helper. Instead, it's all built together. This allowed removing some FFI glue and unifying the logging -- now the helper, by default, logs to the sidecar log. Because I didn't want to move the helper code to libdatadog, there is a new sidecar crate in this repository that just wraps the initialization made in the libdatadog crate and registers the appsec backend (except on windows, where appsec is not supported).
  There remains no production unified musl+glibc binary for the helper. However, for testing, we build a portable sidecar in SSI mode (i.e. libdatadog_php.so is portable), that could became the foundation for a future portable sidecar -- and, later, ddtrace.so itself.

* **Remove the legacy C++ helper**.

For more information, see also the individual commits.


appsec/src/extension/commands_helpers.c
```diff
@@ -251,30 +233,30 @@ static inline void _omsg_destroy(dd_omsg *nonnull omsg)
...
 // incoming
 static ATTR_WARN_UNUSED dd_result _imsg_recv(
-    dd_imsg *nonnull imsg, dd_conn *nonnull conn)
+    dd_imsg *nonnull imsg, dd_conn *nonnull conn, zend_llist *nonnull iovecs)
 {
-    mlog(dd_log_debug, "Will receive response from helper");
+    mlog(dd_log_debug, "Will exchange message with helper");
 
-    dd_result res = dd_conn_recv(conn, &imsg->_data, &imsg->_size);
+    // dd_result res = dd_conn_roundtripv(conn, iovecs, &imsg->_data,
```
estringana: Should this comment go?

estringana (approved):

appsec/src/extension/ddappsec.c
```diff
@@ -160,22 +160,91 @@ static int ddappsec_startup(zend_extension *extension)
...
+#    else
+    bool is_main_thread = count == 0;
+#    endif
+    if (!is_main_thread) {
+#    if defined(__linux__)
+        extern void *__dso_handle;
+        // adds a dependency on glibc 2.18
+        extern int __cxa_thread_atexit_impl(
+            void (*func)(void *), void *arg, void *dso_handle);
+        __cxa_thread_atexit_impl(_tshutdown_handler, NULL, __dso_handle);
```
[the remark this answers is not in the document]
cataphract: centos 7 security support ended two years ago, and ZTS is not that popular on linux anyway, esp. before frankenphp. So I'll accept the incompatibility with glibc 2.17

appsec/helper-rust/src/lib.rs
```diff
@@ -1,258 +1,93 @@
...
-        log::error!("Failed to ensure uniqueness: {}", e);
-        return 1;
-    }
-    let lock = lock.unwrap();
-
     init_waf_logging(&config);
 
     if let Err(e) = rc_notify::resolve_symbols() {
-        log::error!(
+        crate::error!(
```
bwoebi: Is this still needed? I thought we're dropping all runtime resolving?

appsec/helper-rust/src/client/sidecar_msg.rs
```diff
@@ -0,0 +1,674 @@
...
+    }
+}
+
+async fn on_message_impl(
+    session_id: &[u8],
+    client_id: ClientId,
+    command: Vec<u8>,
+) -> anyhow::Result<(ClientId, HelperResponse)> {
+    let (request_tx, resolved_id) = sender_for_client(ClientKey {
+        session_id: session_id.to_vec(),
```
bwoebi, line 153: Why are you carrying around the session_id as a Vec<u8> instead of String? And then converting... back via `let session_id_prod = { || String::from_utf8_lossy(session_id).to_string() };`? Like in `ddappsec_helper::on_message(session_id.as_bytes()`

components-rs/Cargo.toml
```diff
@@ -56,6 +59,11 @@ libc = "0.2"
 bincode = { version = "1.3.3" }
 hashbrown = "0.15"
 
+# ddtrace-sidecar only adds the appsec backend registration around libdatadog's daemon entrypoint,
+# which is a unix-only concern. On Windows `run_sidecar` uses `datadog_sidecar` directly instead.
+[target.'cfg(unix)'.dependencies]
+ddtrace-sidecar = { path = "../sidecar" }
```
bwoebi: Why is this its own crate now? Sure, it's unix only, but why it's own crate?

If we're compiling appsec and tracer together, we probably should rather go ahead with the approach from #4120, just applied to appsec as well.

appsec/src/extension/entity_body.c
```diff
@@ -79,11 +79,12 @@ void dd_entity_body_startup(void)
 
 void dd_entity_body_shutdown(void) { dd_xml_parser_shutdown(); }
 
-void dd_entity_body_gshutdown(void)
+void dd_entity_body_tshutdown(void)
```
bwoebi, line 82: What's the motivation behind the naming convention change?
gshutdown is php thread shutdown.

Is _tshutdown_ meant to say that this is also called on non-php-threads or something? Otherwise please leave this.

appsec/src/extension/helper_process.c
```diff
@@ -105,11 +86,37 @@ void dd_helper_shutdown(void)
...
+    if (!dd_conn_connected(&globals->conn)) {
+        mlog(dd_log_debug, "Not connected; nothing to do in %s", __func__);
+        goto exit;
+    }
+
+    // we assume we're running before the ddtrace shutdown, so we're able to
+    // send the goodbye
+    mlog_g(dd_log_debug, "Sending client shutdown goodbye on gshutdown");
+    struct client_shutdown_data data = {.clean = true, .error = NULL};
+    int res = dd_client_shutdown(&globals->conn, &data);
```
bwoebi, line 107: I don't particularly like this. I would just rely on the sidecar side to signal the connection closed to the appsec.

We can defer this until I have fixed up the session vs process data on the sidecar side though.

appsec/helper-rust/glibc_compat.c
bwoebi, line 1: We don't wish to retain that?

components-rs/sidecar.rs
```diff
@@ -42,53 +49,155 @@ pub static mut DDOG_PHP_FUNCTION: *const u8 = std::ptr::null();
...
+
+#[cfg(unix)]
+#[no_mangle]
+pub unsafe extern "C" fn datadog_crasht_init_with_sidecar(
+    ffi_config: libdd_crashtracker_ffi::Config<'_>,
+    metadata: libdd_crashtracker_ffi::Metadata<'_>,
+    transport: *mut SidecarTransport,
+    sidecar_master_pid: i32,
+) -> ffi::VoidResult {
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| -> anyhow::Result<()> {
```
bwoebi: No point in catch_unwind, we compile with abort.

ext/sidecar.c
```diff
@@ -99,7 +99,30 @@ DATADOG_PUBLIC uint64_t datadog_get_sidecar_queue_id(void) {
...
 
+#ifdef ZTS
+DATADOG_PUBLIC ddog_SidecarTransport **ddtrace_get_sidecar_transport(void *tsrm_ls) {
+    if (tsrm_ls) {
+        void *saved = TSRMLS_CACHE;
+        TSRMLS_CACHE = tsrm_ls;
+        ddog_SidecarTransport **result = &TSRMG_STATIC(
+            datadog_globals_id, zend_datadog_globals *, sidecar);
+        TSRMLS_CACHE = saved;
+        return result;
```
bwoebi:
```suggestion
        void *TSRMLS_CACHE = tsrm_ls;
        return &DATADOG_G(sidecar);
```
just shadow the global

compile_rust.sh
```diff
@@ -11,14 +11,52 @@ case "${host_os}" in
...
+if test -n "$RUST_TOOLCHAIN"; then
+  set -- "+$RUST_TOOLCHAIN" "$@"
 fi
 
-SIDECAR_VERSION=$(cat ../VERSION) RUSTFLAGS="$RUSTFLAGS" "${DDTRACE_CARGO:-cargo}" build $(test "${PROFILE:-debug}" = "debug" || echo --profile "$PROFILE") "$@"
+SIDECAR_VERSION=$(cat ../VERSION) RUSTFLAGS="$RUSTFLAGS" "${DDTRACE_CARGO:-cargo}" "$@"
+cargo_status=$?
+if test "$cargo_status" -ne 0; then
+  exit "$cargo_status"
+fi
```
bwoebi: what does that do? shell scripts generally return the last exit code of the file anyway?

appsec/helper-rust/src/client/sidecar_msg.rs
cataphract: I do represent it as vector throughout once it crosses to helper-rust, in fact helper-rust has `type SessionId = Vec<u8>;`. You're right that originally it comes as a &str from sidecar, but once it crosses the border it's always Vec<u8> (or &[u8]). And the original conversion from &str to &[u8] is actually only a transmutation, so effectively a no-op.

The conversion back to string happens only when formatting it for error messages.

I could, I suppose, represent in helper-rust it as a String instead, and nothing would be lost (String is basically a newtype over Vec<u8>).

components-rs/Cargo.toml
cataphract: The crate exists to bridge libdatadog and helper-rust, which exist in separate repositories. It prepares the AppSecBackend factory that allows sidecar to call into Appsec and creates an alternative entrypoint to the binary for registering the factory (`ddtrace_sidecar_entry_point`) that wraps the original `ddog_daemon_entry_point`.

appsec/src/extension/entity_body.c
cataphract: I've decided call `tshutdown` to the functions that are supposed to be called on thread destruction and leave `gshutdown` for the functions responsible to destroy php module globals, i.e., functions called from the module GSHUTDOWN handler.

The reason for this is that the GSHUTDOWN handler is not appropriate for destroying actual thread-local variables (__thread / C11 _Thread_local), since it may run _after_ the thread has exited.

appsec/src/extension/helper_process.c
cataphract: Some context here: with our own socket, we used to distinguish, for metrics and logging purposes, unclean connection exits from the extension (in the middle of a request or even in the middle of a message, which could happen due to crash or simply because the extension received data it deemed invalid and decided to restart the connection to recover) from clean ones.

But now we don't have any control or visibility over the socket itself -- the extension can't "disconnect" to restart the interaction, so this was the scheme I came up with. It has the advantage the extension can also give a reason for the "disconnection".

appsec/helper-rust/glibc_compat.c
cataphract: Now that we're built as part of sidecar, there's no separate musl-built-glibc-runnable build of the rust-helper to have. I do have experimental musl-built-glibc-runnable builds of sidecar and ddtrace on appsec/tests/integration, but we can discuss that later. It's not used in the CI jobs to produce release artifacts anyway.

compile_rust.sh
cataphract: nothing, I guess there used to be more stuff later

appsec/helper-rust/src/lib.rs
cataphract: yes, I missed these ones

appsec/src/extension/helper_process.c
bwoebi: It's fine for now. We should consolidate this later with relying on ClearQueueId.(i.e. a disconnect after CleanQueueId is assumed clean)

appsec/helper-rust/src/client/sidecar_msg.rs
bwoebi: I know, I just found it an odd choice to store it distinctly (the main usefulness is the Display/Debug impls of String; with &[u8] you just get a vec of ints).

.gitlab/benchmarks.yml
```diff
@@ -232,11 +232,6 @@ check-big-regressions:
 
 .macrobenchmarks:
   stage: benchmarks
-  needs:
-    - job: "package extension: [amd64, x86_64-unknown-linux-gnu]"
-      artifacts: true
-    - job: datadog-setup.php
-      artifacts: true
```
bwoebi, line 239: Why drop this? You don't see it fail because that job doesn't run on your branch, but it needs the artifacts.

cataphract: Fixed

bwoebi (approved):

PR #3733, opened 2026-03-26
Improve dd_library_loader dlclosing

cataphract:
### Description

Improve dd_library_loader dlclosing

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/loader/dd_library_loader.c
+++ b/loader/dd_library_loader.c
@@ -998,15 +998,24 @@ static int ddloader_zend_extension_startup(zend_extension *ext) {
 
 static void ddloader_zend_extension_shutdown(zend_extension *ext) {
     UNUSED(ext);
-    /* Close injected extension handles first (ddtrace.so, ddappsec.so, etc.),
-     * then libddtrace_php.so which they may depend on. */
     for (unsigned int i = 0; i < sizeof(ddloader_injected_ext_config) / sizeof(ddloader_injected_ext_config[0]); ++i) {
-        if (ddloader_injected_ext_config[i].so_handle) {
-            DL_UNLOAD(ddloader_injected_ext_config[i].so_handle);
-            ddloader_injected_ext_config[i].so_handle = NULL;
+        // Set the handle on the zend_extension so that zend_extension_dtor()
+        // will call DL_UNLOAD after this shutdown callback runs. Zend extension
+        // shutdown callbacks are run in the same order (not reverse) as they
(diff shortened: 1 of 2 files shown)


loader/dd_library_loader.c
```diff
@@ -459,31 +422,32 @@ static void ddloader_telemetryf(telemetry_reason reason, injected_ext *config, c
     }
 
     pid_t loader_pid = getpid();
+    // Use double-fork to avoid needing a background reaper thread.
+    // The intermediate child exits immediately and is reaped synchronously
+    // by the parent via waitpid(). The grandchild is reparented to init,
+    // which reaps it automatically after it calls execv().
```
bwoebi: I think I forgot to mention this in the comment :-( - but we have users with "misbehaving" PID 1 scripts in docker containers, which do not reap and accumulate zombies.

That's why we don't double fork. So no, don't just replace that, we have to actually reap this ourselves.

loader/dd_library_loader.c
```diff
@@ -996,15 +960,24 @@ static int ddloader_zend_extension_startup(zend_extension *ext) {
...
     for (unsigned int i = 0; i < sizeof(ddloader_injected_ext_config) / sizeof(ddloader_injected_ext_config[0]); ++i) {
-        if (ddloader_injected_ext_config[i].so_handle) {
-            DL_UNLOAD(ddloader_injected_ext_config[i].so_handle);
-            ddloader_injected_ext_config[i].so_handle = NULL;
+        // Set the handle on the zend_extension so that zend_extension_dtor()
+        // will call DL_UNLOAD after this shutdown callback runs. Zend extension
+        // shutdown callbacks are run in the same order (not reverse) as they
+        // were loaded. So the callbacks for ddtrace/appsec/profiling will run
+        // AFTER this callback.
+        __auto_type ext_config = &ddloader_injected_ext_config[i];
```
bwoebi: I think we need a lint against __auto_type :-P

In all honesty, I just don't like it in C. It's easy to accidentally alias stuff to the wrong type in C - I think always writing the type is definitely superior.

loader/dd_library_loader.c
bwoebi: Because yes, double fork is the obvious solution to this problem, just not an acceptable one here.

loader/dd_library_loader.c
cataphract: well, we can stop using it once we move to C23, because `auto` has been repurposed there :p

PR #3759, opened 2026-03-31
fix(config): stop leak and avoid buffer reuse

### LINKED ISSUE #3729: bug: malloc_consolidate heap corruption with 1.17.0 on FrankenPHP
### Bug report

After upgrading from dd-trace-php 1.16.0 to 1.17.0, we see glibc heap corruption in our FrankenPHP/Laravel Octane (long-running ZTS) production environment:
```
  malloc_consolidate(): unaligned fastbin chunk detected
  memory allocation of 3420891154821048684 bytes failed
```
The process crashes and requires a restart. Intermittent, affects ~20% of pods under production load.
We confirmed by comparing two Docker images identical in every way except dd-trace-php version:

Environment (identical between both images):
  - PHP 8.4.19 (ZTS, Zend Memory Manager disabled)
  - Debian 13 (trixie), x86_64
  - FrankenPHP/Caddy 2.11.2 (serversideup/php:8.4-frankenphp)
  - grpc ext 1.78.0
  - redis ext 6.3.0

Results:
  - dd-trace-php 1.16.0 - stable under production load
  - dd-trace-php 1.17.0 - heap corruption under load

Profiling config: allocation profiling enabled, timeline enabled, CPU time profiling enabled.

**Pinning back to 1.16.0 resolved the issue immediately.**

### PHP version

PHP 8.4.19 (ZTS) built Mar 16 2026

1.17.0

### Installed extensions

Installed extensions: Core, ctype, curl, datadog-profiling, date, ddappsec, ddtrace, dom, fileinfo, filter, grpc, hash, iconv, json, libxml, mbstring, mysqlnd, openssl, pcntl, pcre, PDO, pdo_mysql, pdo_pgsql, pdo_sqlite, Phar, posix, random, readline, redis, Reflection, session, SimpleXML, sodium, SPL, sqlite3, standard, tokenizer, xml, xmlreader, xmlwriter, Zend OPcache, zip, zlib

### Output of `phpinfo()`

_No response_

### Upgrading from

1.16.0

morrisonlevi:
### Description

The leak is rare, it's happening on mshutdown, so unless you are spamming graceful restarts, this doesn't affect much, but of course, we should fix it anyway. Introduced by me in 1.17.0.

The buffer changes should fix #3729. I was able to make a phpt using a failed decoder with Claude's help.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/ext/otel_propagators_empty_dd_propagation_style.phpt
+++ b/tests/ext/otel_propagators_empty_dd_propagation_style.phpt
@@ -1,0 +1,32 @@
+--TEST--
+OTEL_PROPAGATORS fallback uses correct buffer when DD_TRACE_PROPAGATION_STYLE fails to decode
+--DESCRIPTION--
+Regression test for a heap buffer overflow in zai_config_find_and_set_value.
+
+When DD_TRACE_PROPAGATION_STYLE is set to a value that fails to decode (e.g. a
+bare comma, which is all-separators and produces an empty set), the sys env cache
+stores a small persistent allocation for the raw value. The decode failure leaves
+value.len == 0, so the OTEL fallback is triggered. On the unfixed code, buf.ptr
+is still aliased to that small allocation; ddtrace_conf_otel_propagators then
+writes up to 30 bytes into it via memcpy, causing a heap-buffer-overflow.
+
+Under ASAN, the unfixed code crashes the process during MINIT before any PHP
+code runs; the test captures no output and fails. The fix uses a fresh 32 KB
+buffer for every fallback call, preventing the overflow.
+
+Note: an empty string (DD_TRACE_PROPAGATION_STYLE=) cannot be used here because
+PHP's proc_open silently drops env-array entries with empty-string values. A bare
+comma (",") is non-empty so proc_open passes it, but SET_LOWERCASE decode rejects
+it (all-separator input produces zero set elements), triggering the same
+fallback+overflow path.
+--SKIPIF--
+<?php if (!extension_loaded('ddtrace')) die('skip: ddtrace extension required'); ?>
+--ENV--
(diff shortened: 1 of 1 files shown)


zend_abstract_interface/config/config.c
```diff
@@ -99,9 +99,11 @@ static void zai_config_find_and_set_value(zai_config_memoized_entry *memoized, z
             break;
         }
     }
-    if (!value.len && memoized->env_config_fallback && memoized->env_config_fallback(&buf, true)) {
-        zai_config_process_env(memoized, buf, &value);
+    ZAI_ENV_BUFFER_INIT(fallback_buf, ZAI_ENV_MAX_BUFSIZ);
+    if (!value.len && memoized->env_config_fallback && memoized->env_config_fallback(&fallback_buf, true)) {
+        zai_config_process_env(memoized, fallback_buf, &value);
         name_index = ZAI_CONFIG_ORIGIN_MODIFIED;
+        buf = fallback_buf;
```
bwoebi, line 106: This does a 32 KB memcpy, doesn't it?

bwoebi: If you want to do something along these lines, it should be `char *env = buf;` and later `env = fallback_buf`;

morrisonlevi: > This does a 32 KB memcpy, doesn't it?

I just woke up but I don't think so. It's an `zai_env_buffer`, which is just a pointer + len to a hidden buffer. We just assign the ptr + len to the buf.

```c
typedef struct zai_env_buffer_s {
    size_t len;
    char *ptr;
} zai_env_buffer;

#define ZAI_ENV_BUFFER_INIT(name, size) \
    char name##_storage[size];          \
    zai_env_buffer name = {size, name##_storage}
```

PR #3761, opened 2026-04-01
fix(laravel): use route URI instead of unnamed_route in resource name

Leiyks:
## Summary

When `DD_HTTP_SERVER_ROUTE_BASED_NAMING` is enabled (default), routes defined without a `->name()` call produced resource names like:

```
App\Http\Controllers\API\v1\Apps\AppsController@getStripeConnectionToken unnamed_route
```

For unnamed routes, the `unnamed_route` suffix is now replaced with the actual route URI:

```
App\Http\Controllers\API\v1\Apps\AppsController@getStripeConnectionToken stripe/connection
```

Named routes are fully unaffected.

Fixes APMS-18982.

## Root cause

`LaravelIntegration::findRoute` sets the resource to `{action} {routeName}`. For routes defined with the old string syntax (`Route::get('path', 'Controller@method')`) or any route without `->name()`, `normalizeRouteName()` returns the constant `'unnamed_route'`. There was no fallback — the literal string ended up in the resource.


bwoebi (approved):
Looks good, nice fix.

PR #3764, opened 2026-04-01
Batch endpoint collection & remove Wordpress Endpoint collection

bwoebi:
Endpoint collection is also called multiple times in symfony in close succession (before the sidecar actually is done processing).

This leads to quite a bit of overhead.


ext/ddtrace.c
Leiyks, line 2870: Shouldn't we also add `!DDTRACE_G(telemetry_buffer)` in this condition as well for consistency ?

PR #3770, opened 2026-04-02
Use a sidecar connection per PHP thread

bwoebi:
While the sidecar has been a process global for the longest time, it's also unfair with respect to limits and not compatible with the newly introduced outbox.

Having a connection per thread ensures that every thread can have up to 100 queued items. It also reduces contention on the SidecarTransport, which was indeed adding quite a bit of latency due to threads accumulating in futex_wait for very short requests under moderate load.

Moving the background sender stats to its own thread in preparation of dropping the Mutex on the SidecarTransport fully.

### DIFF
--- a/ext/sidecar.c
+++ b/ext/sidecar.c
@@ -1,6 +1,5 @@
 #include <php.h>
 #include <main/SAPI.h>
-#include <stdatomic.h>
 #include "ddtrace.h"
 #include "auto_flush.h"
 #include "compat_string.h"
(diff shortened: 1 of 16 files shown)


ext/signals.c
```diff
@@ -403,7 +406,7 @@ static void dd_sigint_sigterm_handler(int sig, siginfo_t *si, void *uc) {
     memcpy(&dd_signal_data.si, si, sizeof(*si));
     dd_signal_data.uc = uc;
 
-    if (ddtrace_sidecar) {
+    if (ddtrace_sidecar_for_signal) {
```
Leiyks, line 406:
```suggestion
    if (atomic_load(&ddtrace_sidecar_for_signal)) {
```

ext/sidecar.c
Leiyks, line 540:
```suggestion
#endif

atomic_store(&ddtrace_sidecar_for_signal, NULL);

if (ddtrace_sidecar_active_mode == DD_SIDECAR_CONNECTION_THREAD &&
```

Probably best to clear this here instead of later to avoid any case of use-after-free don't you think ? 🤔

ext/sidecar.c
```diff
@@ -879,19 +864,28 @@ void ddtrace_sidecar_rinit(void) {
...
 }
 
+void ddtrace_sidecar_gshutdown(void) {
+    if (DDTRACE_G(sidecar)) {
+        // Drain any accumulated background-sender metrics before the transport goes away.
+        ddtrace_telemetry_flush_bgs_metrics_final();
+        ddog_sidecar_transport_drop(DDTRACE_G(sidecar));
+        DDTRACE_G(sidecar) = NULL;
+    }
+}
```
Leiyks, line 885: Shoudn't this also clear `ddtrace_sidecar_for_signal` ?

bwoebi: Done.

ext/signals.c
bwoebi: Not atomic anymore.

ext/sidecar.c
bwoebi: Done

Leiyks (approved):

PR #3771, opened 2026-04-03
build: update and shrink build images, migrate to clang 19

morrisonlevi:
### Description

A few key changes:
 - Update the PHP versions to the latest. Sort out various dependency pins and patches.
 - Update clang from 17 to 19. Going to clang 19 matches Rust 1.84, but libdatadog recently introduced a compiler flag which was introduced in clang 19 (though anciently in GCC). libdatadog has since [been patched](https://github.com/DataDog/libdatadog/pull/1791), but it's a good idea to upgrade anyway.
     - This doesn't port all of appsec's tests, because going from Clang 17 -> 19 requires quite some changes, needs a dedicated PR.
 - Shrink the CentOS 7 base image from 5.12 GB to 2.72 GB on disk (arm64) and from 1,820 MB to 793 over the wire (again arm64). This comes from a variety of sources including:
     - Dropping some kernel and firmware updates (these are container images, the host is the thing that needs those, not our container).
     - Deleting sources of some dev tools like cmake. Also delete some help docs for cmake, as well as ccmake and cpack.
     - Moving devtoolset-9 to being part of the LLVM layer, and removing it in that same layer. It's not needed after LLVM builds. 
     - Drops protobuf. This used to be a dependency for the profiler, but it hasn't been one for quite some time.

### Reviewer checklist
- [ ] Test coverage seems ok.


.gitlab/ci-images.yml
```diff
@@ -63,7 +63,7 @@ Alpine:
         - 7.0-alpine
   script:
     - cd dockerfiles/ci/alpine_compile_extension
-    - docker login -u "$CI_REGISTRY_USER" -p "$CI_REGISTRY_TOKEN" $CI_REGISTRY
+    - echo "$CI_REGISTRY_TOKEN" | docker login -u "$CI_REGISTRY_USER" --password-stdin "$CI_REGISTRY"
```
danielsn, line 66: Is this to prevent tokens from leaking into the logs?

dockerfiles/ci/centos/7/base.Dockerfile
```diff
@@ -7,6 +7,9 @@ RUN set -eux; \
     sed -i s/^#.*baseurl=http/baseurl=http/g /etc/yum.repos.d/*.repo; \
     sed -i s/^mirrorlist=http/#mirrorlist=http/g /etc/yum.repos.d/*.repo; \
     echo 'ip_resolve = IPv4' >>/etc/yum.conf; \
+# kernel and linux-firmware are useless in containers (host kernel is used);
+# excluding them globally prevents ~183 MB of waste from being pulled in as side-effects.
```
danielsn, line 11: nice

dockerfiles/ci/bookworm/php-8.5/0001-Disable-preserve_none-under-ASan.patch
```diff
@@ -0,0 +1,32 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
danielsn, line 1: Is there a test for this patch, and a test to see when this patch is no longer necessary?

.gitlab/ci-images.yml
morrisonlevi: Something like that, something complained here, claude recommended this as more secure _anyway_ and it seemed plausible enough while being harmless if wrong.

dockerfiles/ci/bookworm/php-8.5/0001-Disable-preserve_none-under-ASan.patch
morrisonlevi: We cannot build the PHP 8.5 CI image without this patch, the build will fail. I'll need to make an upstream PR, I have not done this yet, just discovered this edge late last night or this morning (can't quite remember which).

morrisonlevi: PR is open: https://github.com/php/php-src/pull/21676

dockerfiles/ci/bookworm/php-8.5/0001-Disable-preserve_none-under-ASan.patch
```diff
@@ -0,0 +1,32 @@
...
+--- a/Zend/zend_portability.h
++++ b/Zend/zend_portability.h
+@@ -333,7 +333,12 @@ char *alloca();
+ #endif
+ 
+ #ifdef HAVE_PRESERVE_NONE
+-# define ZEND_PRESERVE_NONE __attribute__((preserve_none))
++# if __has_feature(address_sanitizer) || defined(__SANITIZE_ADDRESS__)
++/* preserve_none + ASan crashes clang 19+ on x86-64, see llvm/llvm-project#95928 */
++#  define ZEND_PRESERVE_NONE
```
bwoebi, line 26: Bummer, that's what enables the tail call VM. Which is something probably worth testing under asan for interceptors.

Maybe we should go straight for newer clang versions where that's fixed.

bwoebi (approved):
Overall looks great, thanks

dockerfiles/ci/bookworm/php-8.5/0001-Disable-preserve_none-under-ASan.patch
morrisonlevi: See my upstream PHP PR: https://github.com/php/php-src/pull/21676  because LLVM itself disabled it in main just 1 week ago: https://github.com/llvm/llvm-project/pull/190001.

PR #3772, opened 2026-04-03
fix(ci): fix Valgrind false positives and CentOS 7 aarch64 profiling build

Leiyks:
### Valgrind false positives
- **`tests/ext/base_service.phpt`**: add `--SKIPIF--` to skip under Valgrind (`USE_TRACKED_ALLOC=1`). The test has a memory leak on PHP 7.0 that causes `test_extension_ci: [7.0]` to fail via the leaked-test check in the Makefile.
- **`Makefile`**: pass `DD_SPAWN_WORKER_STABLE_TRAMPOLINE=1` in the Valgrind test command. When set, the trampoline defers its self-deletion until after the entry function returns, so Valgrind can read DWARF symbols without getting `"connection to image failed"`. PHP's `run-tests.php` marks any test with a non-empty `.mem` file as `LEAKED`, so this was causing false CI failures.
- **`.gitlab/collect_artifacts.sh`**: collect `*.mem` files alongside `*.diff` so Valgrind leak output is available in CI artifacts.
- **`libdatadog` submodule bump** to pick up the `spawn_worker` trampoline fix (see [libdatadog#1844](https://github.com/DataDog/libdatadog/pull/1844)).

### CentOS 7 aarch64 profiling build
- **`.gitlab/build-profiler.sh`** and **`.gitlab/generate-profiler.php`**: set `BINDGEN_EXTRA_CLANG_ARGS="-I$(clang --print-resource-dir)/include"` on CentOS 7 aarch64. The custom-built LLVM 17 in the CI image doesn't put its resource directory on the default include path, causing bindgen to fail with `fatal error: 'stddef.h' file not found`. Mirrors the existing Alpine aarch64 workaround.

## Test plan
- [ ] `test_extension_ci: [7.0]` Valgrind pass no longer reports leaked test
- [ ] `profiling tests` / `compile profiler extension` on CentOS 7 aarch64 pass


bwoebi (approved):
Looks good! ... Updating libdatadog module first.

PR #3778, opened 2026-04-07
Claude instructions to reproduce/troubleshoot CI failures

cataphract:
### Description

Add instructions so that claude can monitor CI pipelines, download artifacts, reproduce and troubleshoot CI failures locally.

These were mostly generated through several iterations where claude is asked to execute jobs by reading the instructions and suggest changes to the files and then by manually reviewing the suggestions.

In the end, the instructions don't need to be 100% accurate, and claude fails to closely follow the instructions 100% of the time anyway; the point is that the instructions and the helper scripts allow it to mimic the CI jobs locally in a reasonable amount of time, without excessive trial and error.

### Reviewer checklist
- [ ] Test coverage seems ok.


.claude/ci/build-slim-package.py
```diff
@@ -0,0 +1,305 @@
+#!/usr/bin/env -S uv run --script
```
bwoebi: This is essentially doing the same than `tooling/bin/build-debug-artifact`?
Only that it seemingly doesn't take care of caching the build, i.e. full rebuild every time? - I see, dockerh does that. Odd.

.claude/ci/system-tests.md
```diff
@@ -0,0 +1,375 @@
...
+practice you only need the version matching your weblog — **use the
+[slim build](#1c-slim-build--one-php-version-only)** and stub the
+rest with empty files. Section 1c has a complete copy-paste recipe
+(build + stubs + tarball assembly) for PHP 8.2 NTS.
+
+`generate-final-artifact.sh` does a hard `cp` for every PHP API ×
+variant. Empty stub files satisfy this for versions you don't build.
+Profiling can also be fully stubbed. See
+[section 1c](#1c-slim-build--one-php-version-only) for all stub
+commands.
```
bwoebi: Why is this talking about generate-final-atrifact at all? seems like noise, you always want a local build?

.claude/ci/build-slim-package.py
cataphract: I just looked at it, and if we want to use `tooling/bin/build-debug-artifact`instead, it needs to be updated to use the centos build. Because otherwise the bookworm images are too recent for system-tests

.claude/ci/system-tests.md
cataphract: Yeah, this has leftovers from a previous strategy. I'll clean that up

.claude/ci/build-slim-package.py
bwoebi: Well, I've been using `tooling/bin/build-debug-artifact` all the time successfully for system tests. Did you actually try it? (`./tooling/bin/build-debug-artifact gnu-aarch64-8.2-nts /Users/bob.weinand/system-tests/binaries` is what I've been using mostly)

.claude/ci/download-artifacts
```diff
@@ -0,0 +1,390 @@
+#!/usr/bin/env -S uv run --script
```
bwoebi: Two questions:
a) Can we move this under tooling/bin ?
b) Would it be reasonable to have it as PHP script instead of python?

.claude/ci/github-actions-profiler.md
```diff
@@ -0,0 +1,300 @@
...
+For a debug (unoptimized) build:
+
+```bash
+cargo rustc --features=trigger_time_sample --profile dev --crate-type=cdylib
+```
+
+Output: `target/debug/libdatadog_php_profiling.so` (~144 MB vs ~20 MB for profiler-release).
+Use the same `php -d extension=...` command, just point to the debug path.
+
+## ZTS tests -- parallel PECL extension
```
bwoebi, line 214: https://github.com/DataDog/dd-trace-php/blob/master/dockerfiles/ci/bookworm/build-extensions.sh#L171-L175
What's this about?

.claude/debugging-system-tests.md
```diff
@@ -0,0 +1,235 @@
...
+                print(f, tags)
+"
+```
+
+PHP/tracer logs:
+
+```
+logs/docker/weblog/logs/php_error.log
+logs/docker/weblog/logs/tracer.log
+```
```
bwoebi, line 238: Something I noticed is that it sometimes adds debug info, and then becomes frustrated because it doesn't  show up.
fprintf(stderr), php_log_err(E_NOTICE) etc.

Basically, it has the tracer logs, so it should simply use `LOG(ERROR, <printf compatible args>)` in tracer code or `error!()` in profiler code, and it'll definitely show up.

.claude/general.md
```diff
@@ -0,0 +1,65 @@
...
+investigated, you MUST use the following template:
+
+```
+The user presented this problem:
+
+> (fill in user query here)
+
+I inspected the source code, and I found the following:
+
+> (fill in your findinds, with references to source files and lines)
```
bwoebi:
```suggestion
> (fill in your findings, with references to source files and lines)
```

.claude/ci/build-slim-package.py
cataphract: I'll try, but the images for apache-mod-... are the appsec-ci images, which are based on bullseye

.claude/ci/download-artifacts
cataphract: sure

.claude/ci/github-actions-profiler.md
cataphract: yeah, once it moved to use the ci bookworm images this no longer became necessary. I'll reword.

.claude/ci/build-slim-package.py
cataphract: yeah, fails:

  [08-Apr-2026 15:02:37 UTC] PHP Warning:  PHP Startup: Unable to load dynamic library 'ddtrace.so'
  (tried: .../ddtrace.so (/lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found
  (required by .../ddtrace.so)), ...

bwoebi: I wasn't aware that we were using different base images for system tests :-(

bwoebi: I see, I was working with `php-fpm-*` images.

bwoebi: We really ought to unify images between appsec and tracer. (Ideally by just putting everything missing into bookworm images)

.claude/debugging-system-tests.md
cataphract: I've tested this and added a section explainging debug methods. Interestinglt, `error!` doesn't print anything either with the current config in system-tests.

.claude/ci/build-slim-package.py
cataphract: I've changed it to build on centos, like on ci.

.claude/ci/windows-tests.md
```diff
@@ -0,0 +1,88 @@
...
+| `package extension windows` | package | Assembles Windows DLLs into `dbgsym.tar.gz` release archive |
+| `verify windows` | package | Installs packaged extension via Chocolatey, runs `verify_windows.ps1` smoke tests |
+
+Runner: `windows-v2:2019`
+Image: `registry.ddbuild.io/images/mirror/datadog/dd-trace-ci:php-{ver}_windows`
+Matrix: PHP 7.2--8.5 (versions where `version_compare($v, "7.2", ">=")`)
+
+## What It Tests
+
+`windows test_c` starts `httpbin-windows` and `php-request-replayer-2.0-windows`
```
bwoebi: Can we try avoiding info which becomes trivially stale, e.g. say `Matrix: PHP 7.2+`, `php-request-replayer-*-windows` etc.? (also in the other files)

cataphract: Done in 9c628cba2

bwoebi (approved):
Let's merge this, we can always refine later.

PR #3779, opened 2026-04-07
Again improve symfony http.route resolution

cataphract:
Turns out it's quite feasible to resolve at runtime the route path from the route name using files in the Symfony cache dir. Not only that, but that strategy was already being used in EndpointCatalog.php.

This fixes a performance issue -- avoid the deserializaiton and possible transfer from Redis of the full route collection cache on every request.

The routing data is in PHP files and with opcache it's basically just one big compile-time variable.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/Integrations/Symfony/V3_3/CommonScenariosTest.php
+++ b/tests/Integrations/Symfony/V3_3/CommonScenariosTest.php
@@ -52,7 +52,7 @@ class CommonScenariosTest extends WebFrameworkTestCase
                     )->withExactTags([
                         'symfony.route.action' => 'AppBundle\Controller\CommonScenariosController@simpleAction',
                         'symfony.route.name' => 'simple',
-                        'http.route' => '/simple',
+
                         'http.method' => 'GET',
                         'http.url' => 'http://localhost/simple?key=value&<redacted>',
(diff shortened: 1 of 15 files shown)


tests/Integrations/Symfony/V3_3/CommonScenariosTest.php
```diff
@@ -52,7 +52,7 @@ public function provideSpecs()
                     )->withExactTags([
                         'symfony.route.action' => 'AppBundle\Controller\CommonScenariosController@simpleAction',
                         'symfony.route.name' => 'simple',
-                        'http.route' => '/simple',
```
estringana, line 55: Why `http.route` is not on tests any more?

cataphract: I disabled it for old versions of Symfony. There, we don't have the files in the cache dir and falling back on getRouteCollection() is too slow.

estringana (changes requested):
Can you please remove the code coming from `glopes/claude-ci` branche since it's not on this pr where that should be reviewed plz?

estringana (approved):

Issue #2629, opened 2024-04-17
[Bug]: CLI PHP Processes do not stop and with Apache (mod_php) the connection is closed unexpectedly 


elzekool:
### Bug report

On our platform we see two behaviors that seem related to the DataDog PHP tracing agent:

- PHP Processed execute but then don't stop (e.g. keep running)
- Apache kills connections unexpectedly

Both issues only arrise sporadicly and we did not find an exact trigger that makes it hard to provide details on the second issue. We could connect with one stuck process with GDB and extracted the following stacktrace:

root@juju-machine-7:/home/ubuntu# strace -p 1333711
strace: Process 1333711 attached
futex(0x7f9fa43899d0, FUTEX_WAIT, 1333735, NULL

#0  __pthread_clockjoin_ex (threadid=140323631699712, thread_return=thread_return@entry=0x0, clockid=clockid@entry=0, abstime=abstime@entry=0x0, block=block@entry=true) at pthread_join_common.c:145
#1  0x00007f9fb0d1e9f3 in __pthread_join (threadid=<optimized out>, thread_return=thread_return@entry=0x0) at pthread_join.c:24
#2  0x00007f9face1f39a in ddtrace_coms_flush_shutdown_writer_synchronous () at /home/circleci/datadog/tmp/build_extension/ext/coms.c:1237
#3  0x00007f9face1f42f in ddtrace_coms_flush_shutdown_writer_synchronous () at /home/circleci/datadog/tmp/build_extension/ext/coms.c:1197
#4  0x00007f9face157b2 in zm_shutdown_ddtrace (type=<optimized out>, module_number=25) at /home/circleci/datadog/tmp/build_extension/ext/ddtrace.c:897
#5  0x00005566fde21e1f in ?? ()
#6  0x00005566fde1ba50 in ?? ()
#7  0x00005566fde2d639 in zend_hash_graceful_reverse_destroy ()
#8  0x00005566fde1c9e7 in ?? ()
#9  0x00005566fddbc6e2 in php_module_shutdown ()
#10 0x00005566fdc82a09 in ?? ()
#11 0x00007f9fb0d5c083 in __libc_start_main (main=0x5566fdc825e0, argc=3, argv=0x7fff3af15a38, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fff3af15a28) at ../csu/libc-start.c:308
#12 0x00005566fdc831be in _start ()

I could trace this to the following function call:
[...]

PR #3782, opened 2026-04-10
fix: prevent CLI/cron process hang on shutdown with unreachable agent

masak1yu:
## Summary

Fixes the issue where PHP CLI/cron processes hang indefinitely on shutdown when the Datadog agent is unreachable. Related to #2629.

### Root Cause

Since v1.4.0, `DD_TRACE_CLI_ENABLED` defaults to `true` (previously `false`). This means the background writer thread is now started for CLI/cron processes. When the agent is unreachable, the shutdown sequence hangs:

1. `ddtrace_coms_flush_shutdown_writer_synchronous()` waits for the writer thread (with `DD_TRACE_SHUTDOWN_TIMEOUT`)
2. On timeout, `pthread_cancel()` is called
3. **`pthread_join()` is called with NO timeout** — hangs if `pthread_cancel` fails to take effect

### Why v1.0.0 worked but v1.4.0+ doesn't

| Config | v1.0.0 | v1.4.0+ |
|--------|--------|---------|
| `DD_TRACE_CLI_ENABLED` | `false` (default) | **`true`** (default) |
| Writer thread for CLI | Not started | **Started** |
| Shutdown hang risk | None | **Yes, if agent unreachable** |

### Changes

1. **Bounded `pthread_join` (`ext/coms.c`)**: After `pthread_cancel`, wait a bounded grace period (`DD_TRACE_SHUTDOWN_TIMEOUT`) for the writer thread to stop. If it doesn't respond, call `pthread_detach` instead of hanging on `pthread_join` forever.

2. **Early retry abort (`ext/coms.c`)**: Check `shutdown_when_idle` flag in the curl retry loop so the writer thread exits promptly when shutdown is pending.

3. **Regression test**: Added `.phpt` test that verifies CLI shutdown completes with an unreachable agent (`192.0.2.1` TEST-NET-1).

### Before / After

**Before (without fix):**
```
$ DD_TRACE_CLI_ENABLED=1 DD_AGENT_HOST=192.0.2.1 DD_TRACE_SIDECAR_TRACE_SENDER=0 \
  php -r "DDTrace\start_span(); DDTrace\close_span();"
# Process hangs indefinitely at shutdown
# strace shows: futex(0x..., FUTEX_WAIT, ...) — stuck in pthread_join
```

**After (with fix):**
```
$ DD_TRACE_CLI_ENABLED=1 DD_AGENT_HOST=192.0.2.1 DD_TRACE_SIDECAR_TRACE_SENDER=0 \
  DD_TRACE_SHUTDOWN_TIMEOUT=2000 \
  php -r "DDTrace\start_span(); DDTrace\close_span();"
# Process exits within ~2-3 seconds (bounded by DD_TRACE_SHUTDOWN_TIMEOUT)
```

## Test plan

- [ ] New `.phpt` test passes: `tests/ext/background-sender/cli_shutdown_no_hang_unreachable_agent.phpt`
- [ ] Existing background-sender tests still pass
- [ ] Existing PCNTL `testDoesNoHangAtShutdownWhenEnabled` / `testDoesNoHangAtShutdownWhenDisabled` tests still pass
- [ ] Manual verification: CLI process exits cleanly when agent is unreachable
- [ ] Verify no regression in trace delivery when agent IS reachable


masak1yu:
## Verification Results

Performed a controlled before/after comparison using Docker to demonstrate the fix.

### Environment

- **Base image**: `datadog/dd-trace-ci:php-8.1_bookworm-7` (PHP 8.1, Debian Bookworm)
- **Host**: macOS (Docker Desktop)
- **Build**: Full `make` from source for each branch

### Test Conditions

| Parameter | Value | Purpose |
|---|---|---|
| `datadog.agent_host` | `192.0.2.1` (TEST-NET-1) | Unreachable address that silently drops packets |
| `datadog.trace.cli_enabled` | `1` | Enable CLI tracing (the default since v1.4.0) |
| `datadog.trace.sidecar_trace_sender` | `0` | Force traditional background writer path |
| `datadog.trace.agent_connect_timeout` | `3000` (3s) | Connect timeout per attempt |
| `datadog.trace.agent_retries` | `5` | 5 retries = 6 total attempts |
| `datadog.trace.shutdown_timeout` | `30000` (30s) | High value so main thread waits for writer to finish on its own |
| `datadog.instrumentation_telemetry_enabled` | `0` | Isolate to writer thread behavior |
| `datadog.remote_config_enabled` | `0` | Isolate to writer thread behavior |
| `datadog.crashtracking_enabled` | `0` | Isolate to writer thread behavior |

### Test Script

A minimal PHP batch job:
```php
<?php
DDTrace\start_span();
echo "batch job: processing...\n";
usleep(100000); // 100ms of "work"
DDTrace\close_span();
echo "batch job: done\n";
```

### Methodology

1. Built Docker image from `datadog/dd-trace-ci:php-8.1_bookworm-7`
2. Copied **master** (unfixed) source into a Docker volume (`cli-hang-before`)
3. Copied **fix branch** source into a separate Docker volume (`cli-hang-after`)
4. Ran each container with identical PHP ini settings pointing to the unreachable agent
5. Measured wall-clock time from PHP invocation to process exit

### Results

| Branch | Status | Elapsed Time |
|---|---|---|
| **master** (unfixed) | OK | **18,161 ms (~18s)** |
| **fix/cli-shutdown-hang-2629** (fixed) | OK | **174 ms (~0.2s)** |

### Analysis

- **BEFORE (master)**: The writer thread's curl retry loop does not check the shutdown flag between retries. All 6 attempts are made (6 × 3s connect timeout = ~18s), blocking process exit.
- **AFTER (fix branch)**: The retry loop now checks `shutdown_when_idle` and breaks after the first failed attempt. The bounded `pthread_join` with `pthread_detach` fallback ensures the process never hangs indefinitely. Result: **~104x faster shutdown**.

This confirms the fix resolves the CLI/cron hang described in #2629.

ext/coms.c
```diff
@@ -972,6 +972,10 @@ static void _dd_curl_send_stack(struct _writer_loop_data_t *writer, ddtrace_coms
     int retries = MAX(get_global_DD_TRACE_AGENT_RETRIES(), 0) + 1;
     CURLcode res = CURLE_UNSUPPORTED_PROTOCOL; // Set a default value to avoid compiler warning
     for (int retry = 0; retry < retries; retry++) {
+        // Abort retries early when shutdown is pending to avoid delaying process exit
+        if (atomic_load(&writer->shutdown_when_idle)) {
+            break;
+        }
```
bwoebi: Shouldn't it try at least once? I.e. move this part to right before `continue;`?

bwoebi: Because well, as it turns out (and as I suspected), in quite some tests it fails to submit the one trace at all, even with a reachable agent.

ext/coms.c
```diff
@@ -1488,9 +1492,32 @@ bool ddtrace_coms_flush_shutdown_writer_synchronous(void) {
...
     if (should_join && !_dd_has_pid_changed()) {
-        // when timeout was not reached and we haven't forked (without restarting thread)
-        // this ensures situation when join is safe from being deadlocked
-        pthread_join(writer->thread->self, NULL);
+        // After pthread_cancel (if timeout was reached), verify the thread has actually
+        // stopped before calling pthread_join. pthread_cancel may fail to take effect if
+        // the thread is blocked in a non-cancellable operation (e.g. certain curl states
+        // or platform-specific pthread_cancel limitations). An unbounded pthread_join in
+        // that case causes the CLI/cron process to hang indefinitely (see #2629).
+        if (atomic_load(&writer->running) || atomic_load(&writer->starting_up)) {
```
bwoebi, line 1501: This part makes sense and looks good to me!

masak1yu:
@bwoebi Thanks for the review! You're right — the `shutdown_when_idle` check at the top of the retry loop was preventing even the first send attempt.

I've moved the check to after `curl_easy_perform` fails (right before `continue;`), so that:
- At least one send attempt is always made → no trace loss when the agent is reachable
- Retries are still skipped during shutdown → fast exit when the agent is unreachable

See commit 6b6cd8e.

bwoebi (approved):
Looks good now, thanks for the improvement!

PR #3791, opened 2026-04-15
Handle APM_MULTI_CONFIG, fixup missing exception replay features

bwoebi:
And fix tests to match libdatadog changes.


ext/live_debugger.c
```diff
@@ -1150,6 +1262,58 @@ static ddog_VoidCollection dd_eval_try_enumerate(void *ctx, const void *zvp) {
...
                 return collection;
             }
+            // For Traversable objects implemented in C (e.g. SplFixedArray, ArrayObject),
+            // Exclude generators and user-defined Iterator/IteratorAggregate implementations (arbitrary side effects).
+            if (Z_OBJCE_P(zv)->get_iterator && Z_OBJCE_P(zv)->type != ZEND_USER_CLASS && Z_OBJCE_P(zv) != zend_ce_generator) {
+                zai_sandbox sandbox;
+                zai_sandbox_open(&sandbox);
+
+                // Collect key-value pairs into two parallel zval arrays on the eval arena. Hard-limited for perf reasons.
+                zend_long max_items = MAX(1024, eval_ctx->config->max_collection_size);
```
Leiyks, line 1272:
```suggestion
                zend_long max_items = MIN(1024, eval_ctx->config->max_collection_size);
```

Isn't this supposed to be MIN in order to be actually limiting ?

metadata/supported-configurations.json
```diff
@@ -106,7 +106,7 @@
       {
         "implementation": "B",
         "type": "boolean",
-        "default": "true"
+        "default": "false"
```
Leiyks, line 109: Is this intended ?

Leiyks (approved):

ext/live_debugger.c
bwoebi: No, MAX is correct - it should be minimum 1024, but at least as much as configured.

metadata/supported-configurations.json
bwoebi: Yes.

ext/live_debugger.c
Leiyks: Ok, I was a bit mislead by the comment.
Maybe you can make it clearer that it is supposed to be a lower and not an upper limit in the comment perhaps

bwoebi: Done

PR #3824, opened 2026-04-24
fix(live-debugger): skip probe installation when sidecar is unavailable

Leiyks:
## Problem

`DDTRACE_G(sidecar)` can be NULL when RC callbacks fire probe installation (RC is initialized before `ddtrace_sidecar_rinit()`), and also when the sidecar fails to connect entirely.

Additionally, `ddog_process_remote_configs()` is called in `ddtrace_sidecar_submit_root_span_data_direct` (added by c46963fc) on every span submission when the service/env hasn't changed. For class-targeted probes (e.g. `Delayed::foo`) that are in RECEIVED state — hook installed but class not yet loaded — the RC state machine sees them as unconfirmed and calls `add_probe` again for the same probe ID in the same request. This results in a spurious second RECEIVED diagnostic in `debugger_span_probe_class.phpt`.

## Fix

1. **Remove all NULL sidecar guards** from `ddog_send_debugger_diagnostics` call sites (`dd_probe_resolved`, `dd_probe_mark_active`, `dd_init_live_debugger_probe`). A NULL sidecar at probe installation time is a lifecycle bug, not an expected condition — surface it rather than silently no-op.

2. **Deduplicate probe installation** in `dd_init_live_debugger_probe`: before installing a new hook, check if a probe with the same ID is already in `active_rc_hooks`. If it is, skip re-installation and return the existing hook ID. This is not losing information — the probe config is already tracked; the RC state machine is simply retrying an already-pending probe.

### DIFF
--- a/ext/live_debugger.c
+++ b/ext/live_debugger.c
@@ -154,10 +154,7 @@ static void dd_probe_resolved(void *data, bool found) {
         def->probe.status_msg = DDOG_CHARSLICE_C("Method does not exist on the given class");
         def->probe.status_exception = DDOG_CHARSLICE_C("METHOD_NOT_FOUND");
     }
-    ddog_SidecarTransport *sidecar = DDTRACE_G(sidecar) ? DDTRACE_G(sidecar) : ddtrace_sidecar_for_signal;
-    if (sidecar) {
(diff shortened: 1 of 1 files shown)


bwoebi (changes requested):
This points to a deeper problem.
This testsuite never should have `DDTRACE_G(sidecar) == NULL`. How does that happen?

We never should install debugger probes before initializing the sidecar. And we probably should also skip installing them if connecting to the sidecar failed.

PR #3827, opened 2026-04-24
ci: fix exit 75 retry with FF_ENABLE_BASH_EXIT_CODE_CHECK and FF_USE_NEW_BASH_EVAL_STRATEGY

Leiyks:
## Root cause

GitLab Runner does not reliably propagate exit codes from external shell scripts (like `wait-for-service-ready.sh`) to the job's exit code. Without the fix, a `before_script` calling `.gitlab/wait-for-service-ready.sh` that exits 75 would be seen by the runner as exit code 1, so `retry: exit_codes: [75]` never matched and kafka jobs were never retried on transient infra failures.

This is addressed by two official GitLab Runner [feature flags](https://docs.gitlab.com/runner/configuration/feature-flags/) (both disabled by default):
- [`FF_ENABLE_BASH_EXIT_CODE_CHECK`](https://docs.gitlab.com/runner/configuration/feature-flags/): check exit code after each command rather than relying solely on `set -e`
- [`FF_USE_NEW_BASH_EVAL_STRATEGY`](https://docs.gitlab.com/runner/configuration/feature-flags/): run the bash `eval` call in a subshell for proper exit code detection

## Fix

- **`generate-common.php`**: enable both feature flags globally via a top-level `variables:` block so all generated child pipelines inherit them
- **`generate-common.php`**: keep `exit_codes: [75]` in the `default:` retry block (confirmed working once exit codes propagate correctly)
- **`generate-tracer.php`**: remove all explicit per-job/per-template `retry_on_infra_failure()` and `retry_on_script_and_infra_failure()` overrides — `default:` handles everything

## Side effect fix

`FF_ENABLE_BASH_EXIT_CODE_CHECK` adds an explicit exit code check after every YAML script step. This broke `command -v switch-php && switch-php` in `generate-profiler.php`: when `switch-php` is absent, `command -v` exits 1, the `&&` short-circuits, and the runner treats the non-zero exit as a job failure — even though the intent was "run only if present".

Fixed by rewriting as `if command -v switch-php > /dev/null 2>&1; then switch-php ...; fi`, whose false branch always exits 0.

## Validation

Confirmed with live CI runs:
- Jobs exiting 1 are **not** retried
- Jobs exiting 75 (via `wait-for-service-ready.sh`) **are** retried

### DIFF
--- a/.gitlab/generate-tracer.php
+++ b/.gitlab/generate-tracer.php
@@ -230,6 +230,7 @@ foreach ($asan_minor_major_targets as $major_minor):
           - PHP_MAJOR_MINOR: "<?= $major_minor ?>"
             ARCH: "<?= $arch ?>"
       artifacts: true
+  retry: 2
   variables:
     WAIT_FOR: test-agent:9126
     KUBERNETES_CPU_REQUEST: 6
@@ -506,6 +507,17 @@ foreach ($all_minor_major_targets as $major_minor):
 <?php endif; ?>
 <?php sidecar_logs(); ?>
   timeout: 40m
+  retry:
+    max: 2
+    when:
+      - script_failure
+      - unknown_failure
+      - data_integrity_failure
+      - runner_system_failure
+      - scheduler_failure
+      - api_failure
+      - stuck_or_timeout_failure
+      - job_execution_timeout
   script:
     - make install_all
     - export XFAIL_LIST="dockerfiles/ci/xfail_tests/${PHP_MAJOR_MINOR}.list"
--- a/.gitlab/generate-tracer.php
+++ b/.gitlab/generate-tracer.php
@@ -662,14 +662,6 @@ foreach ($services as $part => $service) {
     CI_DEBUG_SERVICES: "true"
   retry:
     max: 2
-    when:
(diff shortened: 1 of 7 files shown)


.gitlab/generate-tracer.php
```diff
@@ -615,7 +615,7 @@ function before_script_steps($with_docker_auth = false) {
 <?php if ($sapi): ?>
     DD_TRACE_TEST_SAPI: "<?= $sapi ?>"
 <?php endif; ?>
-<?php if (str_contains($target, "kafk")): ?>
+<?php if (str_contains($target, "kafka")): ?>
```
bwoebi, line 618: Does this fix something? I'm certain this was not a typo

Leiyks: Not really I was just suprised by the typo and fixed it, I don't know if it was intended at all :thinking:

bwoebi: Ah, the reason why this was done is obsolete now. Okay, I accept that.

bwoebi (approved):
Looks good now.
Exotic.

PR #3829, opened 2026-04-24
Allow dumping shm data directly for inspection

bwoebi:
This allows directly inspecting and analyzing the contents of data stored in SHM for debugging/testing purposes.

### DIFF
--- a/components-rs/agent_info.rs
+++ b/components-rs/agent_info.rs
@@ -11,7 +11,23 @@
 use crate::stats::apply_concentrator_config;
 use datadog_sidecar::service::agent_info::AgentInfoReader;
 use libdd_common_ffi::slice::CharSlice;
+use libdd_data_pipeline::agent_info::schema::AgentInfoStruct;
 use std::ffi::c_char;
+use std::ffi::CString;
+
+fn info_to_concentrator_config(info: &AgentInfoStruct) {
+    apply_concentrator_config(
+        info.peer_tags.as_deref().unwrap_or(&[]).to_owned(),
+        info.span_kinds_stats_computed.as_deref().unwrap_or(&[]).to_owned(),
+        info.filter_tags.as_ref().and_then(|f| f.require.as_deref()).unwrap_or(&[]).to_owned(),
+        info.filter_tags.as_ref().and_then(|f| f.reject.as_deref()).unwrap_or(&[]).to_owned(),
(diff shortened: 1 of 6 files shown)


components-rs/agent_info.rs
```diff
@@ -49,18 +65,31 @@ pub unsafe extern "C" fn ddog_apply_agent_info(
...
+/// Returns NULL if no info has been read yet.
+/// The returned pointer must be freed with `ddog_agent_info_json_free`.
+#[no_mangle]
+pub unsafe extern "C" fn ddog_agent_info_as_json(reader: &mut AgentInfoReader) -> *mut c_char {
+    let (changed, info) = reader.read();
+    if let Some(info) = info {
+        if changed {
+            info_to_concentrator_config(info);
         }
+        CString::new(serde_json::to_string(info).unwrap()).unwrap().into_raw()
```
Leiyks, line 83: Shouldn't we have proper error handling here ?

ext/ddtrace.c
```diff
@@ -3136,13 +3136,56 @@ PHP_FUNCTION(dd_trace_internal_fn) {
...
                 usleep(10000); // 10ms
                 waited += 10;
             }
             RETVAL_BOOL(ddog_is_agent_info_ready());
+        } else if (FUNCTION_NAME_MATCHES("get_loaded_remote_configs")) {
+            // Returns a PHP array mapping loaded RC config IDs to their content summary.
+            // e.g. ["datadog/2/LIVE_DEBUGGING/logProbe_log.../config" => ["type"=>"probe","id"=>"log..."]]
+            if (DDTRACE_G(remote_config_state)) {
+                char *rc_json = ddog_remote_config_get_loaded_configs(DDTRACE_G(remote_config_state));
+                if (zai_json_decode_assoc_safe(return_value, rc_json, strlen(rc_json), 128, false) != SUCCESS) {
```
Leiyks, line 3149: Should check if `rc_json` is null

bwoebi: Why? ddog_remote_config_get_loaded_configs never returns null?

components-rs/agent_info.rs
bwoebi: Well, that json was previously properly deserialized, then I very much hope that it can be serialized again :-D
If not there's something seriously broken.

ext/ddtrace.c
Leiyks: oh yes indeed, that's my bad

Leiyks (approved):

PR #3831, opened 2026-04-26
fix(tracing): intern process_tags.serialized to avoid ZTS refcount race

morrisonlevi:
SCP-11166

### Description

Should help with #3729.

ext/process_tags.c stores a single process-wide zend_string at first-request init (zend_string_init persistent=1) and shares it across all ZTS threads. Every span serialization on a first-span path passes that pointer to Rust, which calls back into PHP's zend_string_addref/zend_string_release through ddog_init_span_func to manage a refcount on its Bytes wrapper. Those ops are non-atomic ++/-- on a single shared field; under FrankenPHP + PHP 8.4 ZTS with concurrent worker threads at load, the races produce torn refcount values, premature frees, and use-after-free reads of process_tags.serialized during convert_zend_to_bytes_string -> String::from_utf8_lossy, surfacing as SIGSEGV / SIGABRT / "memory allocation of <huge> bytes failed" / glibc "double free or corruption".

Mark the string IS_STR_INTERNED (= GC_IMMUTABLE) on creation so the refcount helpers see it as immutable and skip the ++/--. Clear the flag in clear_process_tags before the release so the persistent allocation is actually freed at module shutdown.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/ddtrace.c
+++ b/ext/ddtrace.c
@@ -1415,6 +1415,25 @@ void ddtrace_init_known_strings(void) {
 }
 #endif
 
+// Refcount helpers wired into Rust via ddog_init_span_func. They no-op on
+// interned (= GC_IMMUTABLE) zend_strings so process-wide shared strings
+// don't get racing non-atomic ++/-- on their refcount when fed into
+// convert_to_bytes from multiple ZTS threads concurrently. Naming mirrors
+// Zend's GC_TRY_ADDREF / GC_TRY_DELREF convention. Note we call the full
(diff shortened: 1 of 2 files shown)


ext/ddtrace.c
bwoebi, line 1: The changes here should be unnecessary. zend_string_addref and zend_string_release check themselves for interned already.

bwoebi (approved):
Removed the redundant interned checks, looks fine now.

PR #3832, opened 2026-04-26
fix(zai-config): dup INI values per ZTS thread instead of refcounting

morrisonlevi:
SCP-1166

### Description

Relates to #3729.

In ZTS mode, `zai_config_ini_rinit` was assigning `zend_string_copy(source->value)` to each thread's per-thread INI entry. This was a refcount-bump of the process-wide master zend_string stored in `zai_config_memoized_entries`. PHP's refcount operations are not atomic. With many ZTS workers each running rinit concurrently, the competing ++/-- on the shared refcount produced torn values, premature frees, and use-after-free reads, observed as `zend_mm_heap` corrupted or crashes under heavy FrankenPHP load, particularly on aarch64.

### Reviewer checklist
- [ ] Test coverage seems ok.


zend_abstract_interface/config/config_ini.c
```diff
@@ -399,12 +399,25 @@ void zai_config_ini_rinit(void) {
...
+                    // per thread keeps the refcount private and always 1.
                     if (ini->modified) {
                         if (ini->orig_value == ini->value) {
-                            ini->value = source->value;
+                            // ini->value and ini->orig_value alias the same
+                            // allocation (single owning ref). Repoint
+                            // ini->value to its own dup *before* we release
+                            // orig_value below; otherwise the release would
+                            // free the alloc and leave ini->value dangling.
+                            ini->value = zend_string_dup(source->value, 1);
```
bwoebi: This line feels wrong. You're replacing an assignment (no addref) by a dup (has a refcount).

bwoebi:
Alternatively, we might just intern the persistent zai config ini values (`source->value`) - they're never going to change after first_rinit here? Would that work?

bwoebi (approved):
This seems stable now.

PR #3845, opened 2026-04-28
fix: crashtracker runtime stack infinite loop

gyuheon0h:
### Description

There was a crash report with the runtime stack unwinding infinite looping: https://app.datadoghq.com/logs?query=-%40org_id%3A603589%20%40error.is_crash%3Atrue%20%40org_id%3A1000626888&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice%2C%40org_id&event=AwAAAZ3U6zFHIybQJAAAABhBWjNVNnpGSEFBQVF1X283ZzRmcEh3QUEAAAAkMDE5ZGQ0ZWUtNTMzNC00Mzk4LTliNzktOGU5YTQ5NjZhZDhjAANM0A&fromUser=true&index=instrumentation-telemetry-data&link_source=monitor_notif&messageDisplay=inline&refresh_mode=sliding&storage=flex_tier&stream_sort=desc&viz=stream&from_ts=1747059580431&to_ts=1747145980431&live=true

Here is what the raw stack looks like

```
    [
           {
              "column": 4294967295,
              "file": "some-repeated-file-name.php",
              "function": "someRepeatedFunc",
              "line": 157,
              "type_name": "<corrupted scope>"
            },
            {
              "column": 4294967295,
              "file": "some-repeated-file-name.php",
              "function": "someRepeatedFunc",
              "line": 157,
              "type_name": "<corrupted scope>"
            },
            {
              "column": 4294967295,
              "file": "some-repeated-file-name.php",
              "function": "someRepeatedFunc",
              "line": 157,
              "type_name": "<corrupted scope>"
            },
            ....
            {
              "column": 4294967295,
              "file": "some-repeated-file-name.php",
              "function": "someRepeatedFunc",
              "line": 157,
              "type_name": "<corrupted scope>"
            },
            {
              "column": 4294967295,
              "file": "some-repeated-file-name.php",
              "function": "someRepeatedFunc",
              "line": 157,
              "type_name": "<corrupted scope>"
            }
          ]
```
          
          
Couple of things here. 
1. `type_name` is `corrupted scope'
    - this means that this line:
 ```
 frame.type_name = func->common.scope ? zai_is_mapped(&func->common.scope->name, sizeof(zend_string *)) ? dd_validate_zstr(func->common.scope->name) : DDOG_CHARSLICE_C("<corrupted scope>") : DDOG_CHARSLICE_C("");
```
   triggered

2. The column number is 4294967295. This is unrealistic. The reason this is so is because its `-1` in `uint32`.  2^32-1 = 4294967295

Then, this points to this section of code
```
if (!zai_is_mapped(opline, sizeof(zend_op))) {
                frame.column = -1;
                EMIT(&frame);
                continue;
            }
```
being the culprit. `frame.column` is set to -1, and the frame does not proceed, since we don't perform `call = call->prev_execute_data;`

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/ext/crashtracking_frames.c
+++ b/ext/crashtracking_frames.c
@@ -95,6 +95,7 @@ static void dd_frames_callback(void (*emit_frame)(const ddog_crasht_RuntimeStack
             if (!zai_is_mapped(opline, sizeof(zend_op))) {
                 frame.column = -1;
                 EMIT(&frame);
+                call = call->prev_execute_data;
                 continue;
             }
 
--- a/ext/crashtracking_frames.c
+++ b/ext/crashtracking_frames.c
@@ -93,7 +93,7 @@ static void dd_frames_callback(void (*emit_frame)(const ddog_crasht_RuntimeStack
 
             const zend_op *opline = call->opline;
             if (!zai_is_mapped(opline, sizeof(zend_op))) {
-                frame.column = -1;
+                frame.column = 0;
                 EMIT(&frame);
                 call = call->prev_execute_data;
                 continue;
@@ -105,7 +105,7 @@ static void dd_frames_callback(void (*emit_frame)(const ddog_crasht_RuntimeStack
                     if (zai_is_mapped(op, sizeof(zend_op))) {
                         frame.line = op->lineno;
                     } else {
-                        frame.column = -2;
+                        frame.column = 0;
                     }
                 } else {
                     frame.line = func->op_array.line_end;
--- a/ext/crashtracking_frames.c
+++ b/ext/crashtracking_frames.c
@@ -93,7 +93,7 @@ static void dd_frames_callback(void (*emit_frame)(const ddog_crasht_RuntimeStack
 
             const zend_op *opline = call->opline;
             if (!zai_is_mapped(opline, sizeof(zend_op))) {
-                frame.column = 0;
+                frame.column = -1;
                 EMIT(&frame);
                 call = call->prev_execute_data;
                 continue;
@@ -105,7 +105,7 @@ static void dd_frames_callback(void (*emit_frame)(const ddog_crasht_RuntimeStack
                     if (zai_is_mapped(op, sizeof(zend_op))) {
                         frame.line = op->lineno;
                     } else {
-                        frame.column = 0;
+                        frame.column = -2;
                     }
                 } else {
                     frame.line = func->op_array.line_end;

### CODE REVIEW

**Comment by gyuheon0h** [comment]
* **#3845** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-php/3845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-php/3845?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

File: ext/crashtracking_frames.c [hunk 1]
```diff
@@ -93,8 +93,9 @@ static void dd_frames_callback(void (*emit_frame)(const ddog_crasht_RuntimeStack
 
             const zend_op *opline = call->opline;
             if (!zai_is_mapped(opline, sizeof(zend_op))) {
-                frame.column = -1;
+                frame.column = 0;
                 EMIT(&frame);
+                call = call->prev_execute_data;
```
bwoebi, line 98: This is valid, yeah.

bwoebi:
@gyuheon0h I was intentionally using the negative numbers as markers.

Unless there was a problem with the fact that they wrap?

gyuheon0h:
> @gyuheon0h I was intentionally using the negative numbers as markers.
> 
> Unless there was a problem with the fact that they wrap?

@bwoebi Not particularly, but theyre noisy IMO. Libdatadog also says to send 0 for unknown values, but up to you 🤷

ext/crashtracking_frames.c
morrisonlevi: In light of the other investigation... we cannot be certain `call->` is valid xD

bwoebi:
Well, PHP already uses `0` itself when unknown. This is to distinguish.

And sure, they're noisy, so you actually see "there's corruption"- that's good thing, right?

Could you please revert that part? Thanks :-)

gyuheon0h:
> Well, PHP already uses `0` itself when unknown. This is to distinguish.
> 
> And sure, they're noisy, so you actually see "there's corruption"- that's good thing, right?
> 
> Could you please revert that part? Thanks :-)

@bwoebi done, thanks :) 👍

ext/crashtracking_frames.c
bwoebi: This is checked in line 57.

bwoebi (approved):

PR #3853, opened 2026-05-01
Fix SpanStack active reference corruption

morrisonlevi:
### Description

This PR is stacked on #3851 so that needs to merge first.

This test demonstrates that the active span can be corrupted by turning it into a reference. This may be the cause of a customer crash identified by a crash report.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Thanks for the reproducer!

PR #3713, opened 2026-03-19
Add standalone datadog-ipc-helper binary

cataphract:
Extract the sidecar into a dedicated `datadog-ipc-helper` binary (new `components-rs/sidecar-bin` crate) that is located at runtime alongside the `ddtrace.so` path, removing the `php_shared_build` cfg flag and the old `php_sidecar_mockgen` crate entirely.

Build script gains platform specific RUSTFLAGS for the sidecar binary: musl disables static CRT (needed for dlopen of the AppSec helper) and mac os switches fat LTO to thin LTO to work around problems with the llvm mac os linker.

### Reviewer checklist
- [ ] Test coverage seems ok.
- [ ] Appropriate labels assigned.

PR #3869, opened 2026-05-08
Spawn the sidecar via linker instead of trampoline

bwoebi:
Alternative to #3711 and #3713

### DIFF
--- a/loader/tests/functional/test_configuration_telemetry.php
+++ b/loader/tests/functional/test_configuration_telemetry.php
@@ -16,7 +16,7 @@ $output = runCLI('-r "echo \'foo\'; dd_trace_internal_fn(\'finalize_telemetry\')
 assertMatchesFormat($output, '%A"loaded_by_ssi":true%s%A');
 
 // Let time to write the telemetry log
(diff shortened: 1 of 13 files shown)


loader/tests/functional/test_configuration_telemetry.php
```diff
@@ -16,7 +16,7 @@
 assertMatchesFormat($output, '%A"loaded_by_ssi":true%s%A');
 
 // Let time to write the telemetry log
-usleep(10000);
+usleep(300000);
```
Leiyks, line 19: Is this normal ? Why such a bump in value is needed

Leiyks (approved):

loader/tests/functional/test_configuration_telemetry.php
bwoebi: I think some occasional recent loader test flakiness from that. We have other tests, which also have 300ms here, and which do actually not flake.

Ultimately, this test waits on the sidecar to fully initialize and submit this telemetry. This isn't strictly related to this PR.

PR #3874, opened 2026-05-12
fix(tracing): use-after-realloc in ddtrace_format_tracestate

morrisonlevi:
[PROF-14628](https://datadoghq.atlassian.net/browse/PROF-14628)

### Description

There is a logic error in `ddtrace_format_tracestate` that can theoretically result in a stale pointer read caused by a realloc.

I found this by investigating a customer crash which has `smart_str_append` in the stack. I am not sure it is _this_ one.

### Reviewer checklist
- [ ] Test coverage seems ok.

[PROF-14628]: https://datadoghq.atlassian.net/browse/PROF-14628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ext/handlers_http.h
+++ b/ext/handlers_http.h
@@ -24,12 +24,15 @@ static inline zend_string *ddtrace_format_tracestate(zend_string *tracestate, ui
         smart_str_appends(&str, "o:");
         size_t origin_offset = ZSTR_LEN(str.s);
         smart_str_append(&str, origin);
-        for (signed char *cur = (signed char *)ZSTR_VAL(str.s) + origin_offset,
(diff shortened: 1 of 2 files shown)


bwoebi (approved):
The extra `{` indent level is odd, but logic is right.

morrisonlevi:
The extra block scope is to ensure that the temporaries are not used later on, because the whole issue with the bug is the pointer being used after it has been invalidated, so we didn't want their scope to live any longer than necessary.

PR #3898, opened 2026-05-21
Fix memory corruption of INIs in ZTS builds

### LINKED ISSUE #3896: [Bug]: `datadog-profiling` 1.19.x/1.20.0 aborts FrankenPHP with "Out of memory" when `opcache.preload` is set
### Bug report

With `dd-trace 1.19.0`–`1.20.0` installed via `datadog-setup.php --enable-profiling`, FrankenPHP exits within ~1s of startup with a bare `Out of memory` on stdout whenever `opcache.preload` is configured — even when the preload file is just `<?php`. The crash happens during PHP module init, before any request is served. With `DD_PROFILING_LOG_LEVEL=trace`, the last log line is always a `datadog_php_profiling::timeline` "Compile file" event for the preload. Pinning back to `1.18.0` makes the same image boot cleanly.

## Environment

- Base image: `dunglas/frankenphp:1.12.3-php8.5.6-bookworm`
- Install: `datadog-setup.php --php-bin=all --enable-profiling`
- Trigger: `opcache.preload=<any .php file>` + `DD_PROFILING_LOG_LEVEL=trace`

> This is also reproducible in our Symfony app without setting any debugging `DD_x` vars.

## Underlying error

An earlier capture printed the Rust allocator panic before the terse `Out of memory`:

```
memory allocation of 7218835291860648052 bytes failed
```

≈ 6.5 EB — a corrupted/uninitialised size inside the profiler's timeline allocator.

Sample reproducer:

```Dockerfile
# Reproducer for dd-trace 1.19.x/1.20.0 "Out of memory" on FrankenPHP.
# Build: docker build --build-arg DD_VERSION=1.20.0 -t ddoom .
# Run:   docker run --rm -e DD_TRACE_DEBUG=1 -e DD_PROFILING_LOG_LEVEL=trace ddoom

ARG DD_VERSION=1.20.0
FROM dunglas/frankenphp:1.12.3-php8.5.6-bookworm
ARG DD_VERSION
USER root

RUN install-php-extensions opcache \
 && curl -fsSL "https://github.com/DataDog/dd-trace-php/releases/download/${DD_VERSION}/datadog-setup.php" \
      | php -- --php-bin=all --enable-profiling

RUN mkdir -p /app/public \
 && echo '<?php phpinfo();' > /app/public/index.php \
 && echo '<?php' > /app/preload.php \
 && printf '%s\n' \
      "opcache.enable=1" "opcache.enable_cli=1" \
      "opcache.preload=/app/preload.php" "opcache.preload_user=www-data" \
    > /usr/local/etc/php/conf.d/zzz-app.ini \
 && chown -R www-data:www-data /app /data/caddy /config/caddy

USER www-data
WORKDIR /app
ENV APP_ENV=prod
```
```shell
#!/usr/bin/env bash
# Builds the Dockerfile for dd-trace 1.18.0 and 1.20.0, runs each container
# with profiling debug envs, and reports whether it crashed.
# Containers are left around for inspection: `docker logs <name>`,
# `docker inspect <name>`, etc. Remove them with `docker rm -f <name>`.

set -euo pipefail
cd "$(dirname "$0")"

check() {
    local v="$1" name="ddoom-${1//./-}"
    echo "── dd-trace $v ──"

    echo "→ building image"
    local img; img=$(docker build -q --build-arg "DD_VERSION=$v" .)

    echo "→ starting container $name with DD_TRACE_DEBUG=1, DD_PROFILING_LOG_LEVEL=trace"
    docker rm -f "$name" >/dev/null 2>&1 || true
    docker run -d --name "$name" \
        -e DD_TRACE_DEBUG=1 -e DD_PROFILING_LOG_LEVEL=trace \
        "$img" >/dev/null

    echo "→ waiting 5s, then inspecting"
    sleep 5
    local st ec
    st=$(docker inspect "$name" --format '{{.State.Status}}')
    ec=$(docker inspect "$name" --format '{{.State.ExitCode}}')
    echo "  status=$st exit=$ec"

    if docker logs "$name" 2>&1 | grep -q "^Out of memory"; then
        echo "  ✗ crashed: \"Out of memory\" in stdout"
        echo "  last profiler trace:"
        docker logs "$name" 2>&1 | grep datadog_php_profiling | tail -1 | sed 's/^/    /'
    else
        echo "  ✓ no \"Out of memory\" in stdout"
    fi
    echo "  inspect with: docker logs $name | tail -20"
    echo
}

check 1.18.0
check 1.20.0
```

### PHP version

8.5.6

1.20.0

### Installed extensions

_No response_

### Output of `phpinfo()`

_No response_

### Upgrading from

_No response_

bwoebi:
The preloading skip was lost during the last refactor; putting it back.

Fixes #3896.

### DIFF
--- a/zend_abstract_interface/config/config_ini.c
+++ b/zend_abstract_interface/config/config_ini.c
@@ -407,7 +407,6 @@ void zai_config_ini_rinit(void) {
                     // On ZTS INIs must be not shared between threads (otherwise: refcount race conditions). Hence we dup them rather than just copy.
                     ZEND_ASSERT(ini && source);
                     ZEND_ASSERT(ini != source && "ZTS INI sync must not run against the global INI table");
-                    ZEND_ASSERT(ini->value != source->value && "ZTS INI values must not be shared before syncing");
                     if (ini->modified) {
                         bool identical_orig = ini->orig_value == ini->value;
                         zend_string_release(ini->orig_value);
--- a/zend_abstract_interface/config/config_ini.c
+++ b/zend_abstract_interface/config/config_ini.c
@@ -396,7 +396,7 @@ void zai_config_ini_rinit(void) {
 
 #if ZTS
     // Skip during preloading, in that case EG(ini_directives) is the actual source of truth (NTS-like)
-    if (env_to_ini_name && !inis_synchronized && zai_config_first_rinit_done) {
+    if (env_to_ini_name && !inis_synchronized && zai_config_first_rinit_done && !in_startup) {
         for (uint16_t i = 0; i < zai_config_memoized_entries_count; ++i) {
             zai_config_memoized_entry *memoized = &zai_config_memoized_entries[i];
             if (!memoized->original_on_modify) {
--- a/zend_abstract_interface/config/config_ini.c
+++ b/zend_abstract_interface/config/config_ini.c
@@ -405,6 +405,9 @@ void zai_config_ini_rinit(void) {
                     zend_ini_entry *source = memoized->ini_entries[n],
                                    *ini = zend_hash_find_ptr(EG(ini_directives), source->name);
                     // On ZTS INIs must be not shared between threads (otherwise: refcount race conditions). Hence we dup them rather than just copy.
+                    ZEND_ASSERT(ini && source);
+                    ZEND_ASSERT(ini != source && "ZTS INI sync must not run against the global INI table");
+                    ZEND_ASSERT(ini->value != source->value && "ZTS INI values must not be shared before syncing");
                     if (ini->modified) {
                         bool identical_orig = ini->orig_value == ini->value;
                         zend_string_release(ini->orig_value);


morrisonlevi (commented):
We should be able to write a test for this, yes?

bwoebi:
Yeah, probably an asan tea test would catch it.

zend_abstract_interface/config/config_ini.c
realFlowControl, line 407:
```suggestion
                    // On ZTS INIs must be not shared between threads (otherwise: refcount race conditions). Hence we dup them rather than just copy.
                    ZEND_ASSERT(ini && source);
                    ZEND_ASSERT(ini != source && "ZTS INI sync must not run against the global INI table");
                    ZEND_ASSERT(ini->value != source->value && "ZTS INI values must not be shared before syncing");
```

bwoebi: That's right.

realFlowControl:
@bwoebi / @morrisonlevi I have a PR at https://github.com/DataDog/dd-trace-php/pull/3907 to add a test, but the preloading test in PHP profiling also fails. We could merge the PR, but actually we either need ZTS ASAN NDEBUG builds or support DEBUG builds in the profiler. WDYT?

realFlowControl (approved):

realFlowControl:
Whoops @bwoebi could you have a look at the failing ASAN tests?

realFlowControl:
@morrisonlevi see https://github.com/DataDog/dd-trace-php/actions/runs/26288603294/job/77382502552?pr=3908 for a DEBUG ZTS ASAN run, it shows this issue 🎉

bwoebi:
@realFlowControl They *can* be the same value, if the value happens to be interned. In that case, zend_string_dup (in copy_ini_entry) will not actually duplicate the string.

zend_abstract_interface/config/config_ini.c
```diff
@@ -405,6 +405,9 @@ void zai_config_ini_rinit(void) {
                     zend_ini_entry *source = memoized->ini_entries[n],
                                    *ini = zend_hash_find_ptr(EG(ini_directives), source->name);
                     // On ZTS INIs must be not shared between threads (otherwise: refcount race conditions). Hence we dup them rather than just copy.
+                    ZEND_ASSERT(ini && source);
+                    ZEND_ASSERT(ini != source && "ZTS INI sync must not run against the global INI table");
+                    ZEND_ASSERT(ini->value != source->value && "ZTS INI values must not be shared before syncing");
```
bwoebi:
```suggestion
```
Let's just remove it, as long as we don't run against the same source ini, it will be fine.

PR #3906, opened 2026-05-22
feat(ffe): add runtime-backed PHP feature flag evaluation

leoromanovsky:
## Motivation

This PR adds real server-side PHP feature flag evaluation backed by Remote Config and the `libdatadog` native evaluator. PHP 7 gets a Datadog API, PHP 8 can use the optional OpenFeature bridge, and both paths use the same live runtime evaluator.

This is intentionally the evaluation layer only. Exposure delivery, exposure caching, and evaluation metrics land in the later hook, exposure, and metrics PRs.

Shared planning doc: https://docs.google.com/document/d/1NvMfTpZWLBlFmEFNjdnlMyeVpy5l7KD8qujGFco6w2w/edit?tab=t.0

## Decisions

- Production clients/providers read only from the live Remote Config-backed native runtime.
- The root package remains PHP 7 installable; the OpenFeature SDK is used only in PHP 8 test/app environments and is not a root dependency.
- `productionRuntime=false` remains a temporary guardrail while the hook, exposure delivery, and evaluation metrics layers are incomplete. The final production-ready PHP FFE stack should not ship with that warning enabled.
- The PHP/OpenFeature bridge passes a targeting key plus flat primitive attributes (`bool`, `int`, `float`, `string`) into the native evaluator. Nested arrays, objects, and null attribute values are dropped at the boundary.
- `DDTrace\Testing\ffe_load_config` exists only for local/canonical fixture tests.

## Where this PR fits in the stack

This is the **bottom layer** of the 4-PR stack. #3909 adds the shared hook, while #3910 (EVP exposures) and #3911 (OTLP metrics) build on top of that hook.

pr3906-runtime-evaluation-stack-position

## Where this PR fits in the target system

This PR contributes the in-PHP evaluation surface: `UserCode` -> `OpenFeature Client` (PHP 8) / `DDTrace FeatureFlags Client` (PHP 7 + 8), the `DDTrace OpenFeature DataDogProvider`, the `NativeEvaluator` FFI bridge into libdatadog, and the Remote Config client for the `FFE_FLAGS` product.

pr3906-runtime-evaluation-system-scope

## Changes

- Imports the `libdatadog` FFE evaluator path and keeps PHP user-facing APIs as thin adapters over the native runtime.
- Wires PHP 7-safe `DDTrace\FeatureFlags` evaluation through live Remote Config.
- Adds a PHP 8-only `DDTrace\OpenFeature` provider bridge.
- Adds the canonical `DataDog/ffe-system-test-data` submodule and a native-runtime PHPT loop over `ufc-config.json` and `evaluation-cases/*.json`.

## Not Included

- Exposure writer, exposure cache, or EVP forwarding.
- Exposure flush/dedup tests.
- Flag-evaluation metric emission.

## Validation

- `php vendor/bin/phpunit --config phpunit.xml tests/api/Unit/FeatureFlags tests/OpenFeature/DataDogProviderTest.php`: 29 tests / 87 assertions.
- `make test_featureflags`: 8 tests / 29 assertions.
- `MAX_TEST_PARALLELISM=1 TESTS=tests/ext/ffe/native_bridge_evaluate.phpt make test_c`: 1/1 passed.
- `MAX_TEST_PARALLELISM=1 TESTS=tests/ext/ffe/system_test_data_evaluate.phpt make test_c`: 1/1 passed.
- `make test_internal_api_randomized`: passed.

## Dogfooding Validation

See DataDog/ffe-dogfooding#68.

Ran the PHP 7 and PHP 8 servers locally; observed correct evaluations.

Screenshot 2026-05-22 at 12 57 05 PM


ext/autoload_php_files.c
```diff
@@ -32,7 +32,9 @@ static zend_class_entry *(*dd_prev_autoloader)(zend_string *name, zend_string *l
 static zend_bool dd_api_is_preloaded = false;
 static zend_bool dd_otel_is_preloaded = false;
 static zend_bool dd_legacy_tracer_is_preloaded = false;
+static zend_bool dd_openfeature_is_preloaded = false;
 #endif
+static zend_bool dd_openfeature_is_loaded = false;
```
morrisonlevi: You already pushed a fix for this because I gave the review on slack, but wanted to post here too:

On ZTS without preloading, this will race across requests. The fix is probably using a PHP module global which is initialized and re-initialized with the value of dd_openfeature_is_preloaded on the start of each request.

components-rs/ffe.rs
```diff
@@ -0,0 +1,445 @@
...
+    pub variant: Option<CString>,
+    pub allocation_key: Option<CString>,
+    pub reason: i32,
+    pub error_code: i32,
+    pub do_log: bool,
+}
+
+#[repr(C)]
+pub struct FfeAttribute {
+    pub key: *const c_char,
```
bwoebi: Avoid using c_char, prefer `libdd_common_ffi::CharSlice`. This has a bit nicer ergonomics, avoids string length recalculation and the C side has helpers to convert from zend_string* to CharSlice..

ext/ddtrace.c
```diff
@@ -2951,6 +2951,139 @@ PHP_FUNCTION(DDTrace_flush_endpoints) {
...
+    size_t idx = 0;
+    zend_string *key;
+    zval *value;
+    struct ddog_FfeResult *result;
+    const char *value_json;
+    const char *variant;
+    const char *allocation_key;
+
+    ZEND_PARSE_PARAMETERS_START(4, 4)
+        Z_PARAM_STRING(flag_key, flag_key_len)
```
bwoebi: Intead of unused length, use Z_PARAM_STR (and convert zend_strings to CharSlice) (in all added functions here)

components-rs/ffe.rs
```diff
@@ -0,0 +1,445 @@
...
+const TYPE_STRING: i32 = 0;
+const TYPE_INTEGER: i32 = 1;
+const TYPE_FLOAT: i32 = 2;
+const TYPE_BOOLEAN: i32 = 3;
+const TYPE_OBJECT: i32 = 4;
+
+pub struct FfeResult {
+    pub value_json: CString,
+    pub variant: Option<CString>,
+    pub allocation_key: Option<CString>,
```
bwoebi: If you want to reduce cleanup and copying and make the ddog_ffe_result_* functions unnecessary, you could make this struct `#[repr(C)]` and use `crate::bytes::OwnedZendString`. You'll get direct `zend_string*` pointers, which the caller can simply move out of the struct when building the result.

Simpler code and less overhead at once.

ext/ddtrace.stub.php
```diff
@@ -845,6 +845,38 @@ function add_endpoint(string $path, string $operation_name, string $resource_nam
      * Call this once after batching all add_endpoint() calls.
      */
     function flush_endpoints(): void {}
+
+    /**
+     * Evaluate a feature flag using the stored UFC configuration.
+     *
+     * @param string $flagKey The flag key to evaluate.
+     * @param int $expectedType The expected flag type (0=string, 1=int, 2=float, 3=bool, 4=object).
```
bwoebi: There should be probably a bunch of constants:

```
const FFE_STRING = 0;
const FFE_INT = 1;
...
```
and use it in your `NativeEvaluator::typeId()` method

```suggestion
     * @param int $expectedType The expected flag type (one of the DDTrace\FFE_* constants).
```

then.

ext/ddtrace.c
```diff
@@ -2951,6 +2951,139 @@ PHP_FUNCTION(DDTrace_flush_endpoints) {
...
+                default:
+                    continue;
+            }
+
+            idx++;
+        } ZEND_HASH_FOREACH_END();
+        attrs_count = idx;
+    }
+
+    ddtrace_process_remote_config_now();
```
bwoebi: What's that for? The tracer gets notified (via unix signal) whenever new RC updates are incoming. You don't need to poll for that.

src/api/FeatureFlags/Client.php
```diff
@@ -0,0 +1,201 @@
...
+
+final class Client
+{
+    private $evaluator;
+    private $warningEmitter;
+    private $warnedAboutNonProductionRuntime = false;
+
+    private function __construct(
+        Evaluator $evaluator,
+        WarningEmitter $warningEmitter
```
bwoebi: I'd reuse the `DDTrace\Log\LoggerInterface` here - if you want a logger using trigger_error by default, add a TriggerErrorLogger to src/api/log, which uses trigger_error().

That implementation has also a wrapper to be compatible with the wider PHP logger ecosystem.

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -0,0 +1,211 @@
...
+
+final class DataDogProvider extends AbstractProvider
+{
+    protected static string $NAME = 'Datadog';
+
+    private FeatureFlagsClient $client;
+    private WarningEmitter $warningEmitter;
+    private bool $warnedAboutNonProductionRuntime = false;
+
+    public function __construct()
```
bwoebi: At least the logger should be configurable here in the constructor.

src/api/FeatureFlags/Internal/ResultMapper.php
```diff
@@ -0,0 +1,316 @@
+<?php
+
+namespace DDTrace\FeatureFlags\Internal;
+
+use DDTrace\FeatureFlags\EvaluationDetails;
+use DDTrace\FeatureFlags\EvaluationErrorCode;
+use DDTrace\FeatureFlags\EvaluationReason;
+use DDTrace\FeatureFlags\EvaluationType;
+
+final class ResultMapper
```
bwoebi, line 10: There is not really a purpose of this being a standalone class, is there? It could just be all private functions on NativeEvaluator, right?

src/api/FeatureFlags/Internal/NativeEvaluator.php
```diff
@@ -0,0 +1,134 @@
...
+            }
+        }
+
+        return $normalized;
+    }
+
+    private function withProviderState(array $rawResult)
+    {
+        $hasConfig = $this->remoteConfig->hasConfig();
+        $configVersion = $this->remoteConfig->configVersion();
```
bwoebi: if isAvailable is true (DDTrace\ffe_evaluate existing - precondition for this function being called) it should be safe to assume that the other functions also exist (ffe_has_config and ffe_config_version).

```suggestion
        $hasConfig = \DDTrace\ffe_has_config();
        $configVersion = \DDTrace\ffe_config_version();
```

No need for RemoteConfigClient.php.

src/api/FeatureFlags/Internal/NativeEvaluator.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public function evaluate(
+        $flagKey,
+        $expectedType,
+        $defaultValue,
+        $targetingKey = null,
+        array $attributes = array()
+    ) {
+        if (!self::isAvailable()) {
```
bwoebi: Why not make the constructor private and rely on function createOrUnavailable having checked availability in the first place (I also think that method should just be called `create()`)?
Making this branch fully redundant.

src/api/FeatureFlags/Internal/UnavailableEvaluator.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace DDTrace\FeatureFlags\Internal;
+
+use DDTrace\FeatureFlags\EvaluationDetails;
+use DDTrace\FeatureFlags\EvaluationErrorCode;
+use DDTrace\FeatureFlags\EvaluationReason;
+
+final class UnavailableEvaluator implements Evaluator
+{
+    const WARNING_MESSAGE = 'Datadog-backed PHP feature flag evaluation is not fully enabled yet. '
```
bwoebi: This message - the `yet`, sounds sort of like it'll become available at some point. Yet once you call `Client::create()`, you'll have an UnavailableEvaluator for the whole lifetime of the Client, regardless of whether it becomes available later.

This may not be optimal, or the message is at least a bit confusing.

components-rs/ffe.rs
```diff
@@ -0,0 +1,333 @@
...
+        Err(error) => {
+            let (error_code, reason) = match &error {
+                EvaluationError::TypeMismatch { .. } => (ERROR_TYPE_MISMATCH, REASON_ERROR),
+                EvaluationError::ConfigurationParseError => (ERROR_CONFIG_PARSE, REASON_ERROR),
+                EvaluationError::ConfigurationMissing => (ERROR_CONFIG_MISSING, REASON_ERROR),
+                EvaluationError::FlagUnrecognizedOrDisabled => {
+                    (ERROR_FLAG_UNRECOGNIZED, REASON_DEFAULT)
+                }
+                EvaluationError::FlagDisabled => (ERROR_NONE, REASON_DISABLED),
+                EvaluationError::DefaultAllocationNull => (ERROR_NONE, REASON_DEFAULT),
```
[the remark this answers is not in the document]
leoromanovsky: Handled by ResultMapper::isDefaultReturn() plus unit coverage for disabled and default-null results. Those now return the caller default with DISABLED/DEFAULT reason instead of going through value coercion and becoming TYPE_MISMATCH.

src/api/FeatureFlags/Internal/ResultMapper.php
leoromanovsky: Keeping this as a separate internal class intentionally. It isolates the raw bridge-result contract from NativeEvaluator, and lets the mapper unit tests cover disabled/default/null/error coercion without requiring native functions or Remote Config state. NativeEvaluator is now limited to input normalization, the native call, and provider-state attachment.

components-rs/ffe.rs
leoromanovsky: Agreed that this would be a cleaner/lower-copy result path. I left the result ownership shape unchanged in this review pass and applied the lower-risk FFI input cleanup with CharSlice now. Moving result materialization to OwnedZendString changes ownership across the C/Rust/PHP boundary, so I would rather do that as a focused follow-up if we see measurable overhead after M1.

bwoebi: @leoromanovsky I'd like to see this regardless of overhead (which will marginally improve, but it's not the main point), because it's probably 50+ lines less code, making the feature simpler.

leoromanovsky: Done.

leoromanovsky:
Feedback from @bwoebi addressed; retest with local build against ffe-dogfooding server with satisfactory results.

components-rs/ffe.rs
```diff
@@ -0,0 +1,453 @@
...
+
+fn result_from_assignment(assignment: Result<ffe::Assignment, EvaluationError>) -> FfeResult {
+    match assignment {
+        Ok(assignment) => FfeResult {
+            value_json: Some(string_to_zend_string(assignment_value_to_json(
+                &assignment.value,
+            ))),
+            variant: Some(string_to_zend_string(
+                assignment.variation_key.as_str().to_string(),
+            )),
```
bwoebi:
```suggestion
            variant: Some(assignment.variation_key.as_str().into()),
```
should work, same for allocation_key and value_json in error branch

ext/remote_config.c
```diff
@@ -64,6 +64,17 @@ void ddtrace_check_for_new_config_now(void) {
     }
 }
 
+void ddtrace_process_remote_config_now(void) {
```
bwoebi: Leftover now

src/api/FeatureFlags/Internal/ResultMapper.php
bwoebi: Okay.

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -0,0 +1,207 @@
...
+
+    private FeatureFlagsClient $client;
+    private LoggerInterface $warningLogger;
+    private bool $warnedAboutNonProductionRuntime = false;
+
+    public function __construct(?FeatureFlagsClient $client = null, ?LoggerInterface $logger = null)
+    {
+        $this->client = $client ?: FeatureFlagsClient::createWithDependencies(
+            null,
+            new NullLogger(LogLevel::EMERGENCY)
```
bwoebi: Is this intentional that it doesn't pass $this->warningLogger here?

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -0,0 +1,207 @@
...
+        );
+        $this->warningLogger = $logger ?: new TriggerErrorLogger();
+    }
+
+    /**
+     * @internal Tests and Datadog-owned bridge adapters only.
+     */
+    public static function createWithDependencies(
```
bwoebi: redundant with constructor.

What you probably want is:

```
    public function __construct(?LoggerInterface $logger = null) // the public API, only has $logger
    {
        $this->warningLogger = $logger ?: new TriggerErrorLogger();
        $this->client = FeatureFlagsClient::createWithDependencies(
            null,
            $this->warningLogger
        );
    }

    public static function createWithDependencies(?FeatureFlagsClient $client = null, ?LoggerInterface $logger = null): self { // the internal test function, can swap the FeatureFlagsClient impl.
        $instance = new self($logger);
        if ($logger) {
            $instance->logger = $logger;
        }
    }
```

Or, honestly, it's only for testing, right? Strip createWithDependencies fully from the API and assign it in the test:
$provider = new DataDogProvider($logger);
(fn() => $this->client = new RecordingLogger)->call($provider); // bypass private state directly in the test.

That would be my favourite approach.

src/api/FeatureFlags/Client.php
```diff
@@ -0,0 +1,201 @@
...
+
+    public static function create()
+    {
+        return self::createWithDependencies();
+    }
+
+    /**
+     * @internal Tests and Datadog-owned bridge adapters only.
+     */
+    public static function createWithDependencies(
```
bwoebi: Same as for DataDogProvider, you probably should make the __construct public API, remove Evaluator as arg (logger becomes optional arg), and overwrite it in tests via Closure::call().

.github/dependabot.yml
```diff
@@ -5,6 +5,13 @@
 
 version: 2
 updates:
+  - package-ecosystem: "gitsubmodule"
```
bwoebi, line 8: This is global, I don't want that for other submodules. If you can restrict it to the ffe tests submodule, fine by me.

ext/ddtrace.c
```diff
@@ -2951,6 +2951,131 @@ PHP_FUNCTION(DDTrace_flush_endpoints) {
...
+        Z_PARAM_STR(flag_key)
+        Z_PARAM_LONG(type_id_zl)
+        Z_PARAM_STR_OR_NULL(targeting_key)
+        Z_PARAM_ARRAY(attrs_zv)
+    ZEND_PARSE_PARAMETERS_END();
+
+    type_id = (int32_t) type_id_zl;
+    if (targeting_key != NULL) {
+        targeting_key_slice = dd_zend_string_to_CharSlice(targeting_key);
+    }
```
bwoebi: nit: dd_zend_string_to_CharSlice handles NULL itself already, if is redundant.

ext/ddtrace.stub.php
```diff
@@ -845,6 +875,38 @@ function add_endpoint(string $path, string $operation_name, string $resource_nam
...
+     *
+     * @param string $flagKey The flag key to evaluate.
+     * @param int $expectedType One of the DDTrace\FFE_* constants.
+     * @param string|null $targetingKey The targeting key for evaluation context.
+     * @param array $attributes Flat key-value map of evaluation context attributes (string keys, primitive values).
+     * @return array|null Associative array with keys: value_json, variant, allocation_key, reason, error_code, do_log. Null only if evaluation engine is unavailable.
+     *
+     * @internal Used by the Datadog feature flag client.
+     */
+    function ffe_evaluate(string $flagKey, int $expectedType, ?string $targetingKey, array $attributes): ?array {}
```
bwoebi: minor suggestion, but feel free to ignore:

add
```
class FfeResult {
    public ?string $valueJson = null;
    public ?string $variant = null;
    public ?string $allocationKey = null;
    public int $reason;
    public int $errorCode;
    public bool $doLog;
}
```

and return that instead of array. Constructing/accessing objects is faster than arrays too, and better automcomplete etc. But not necessary.

src/api/FeatureFlags/Internal/ResultMapper.php
```diff
@@ -0,0 +1,316 @@
...
+     * @return EvaluationDetails
+     */
+    public function map($rawResult, $expectedType, $defaultValue)
+    {
+        if (!EvaluationType::isValid($expectedType)) {
+            throw new \InvalidArgumentException('Unknown feature flag value type: ' . (string) $expectedType);
+        }
+
+        if ($rawResult instanceof EvaluationDetails) {
+            return $rawResult;
```
bwoebi: Is this branch ever true?

.github/dependabot.yml
leoromanovsky: addressed by scoping to ffe fixtures data

src/api/FeatureFlags/Client.php
```diff
@@ -0,0 +1,181 @@
...
+
+        $this->evaluator = NativeEvaluator::create();
+        $this->logger = $logger ?: new TriggerErrorLogger();
+    }
+
+    public static function create()
+    {
+        return new self();
+    }
+
```
bwoebi:
```suggestion
```

Redundant leftover method? And remove testCreateBuildsDefaultRemoteConfigBackedClient

src/api/FeatureFlags/EvaluationDetails.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->reason = $reason;
+        $this->variant = $variant;
+        $this->errorCode = $errorCode;
+        $this->errorMessage = $errorMessage;
+        $this->flagMetadata = $flagMetadata;
+        $this->exposureData = $exposureData;
+        $this->providerState = $providerState;
+    }
+
+    public function getValue()
```
bwoebi, line 65: given that these getters are public API, can you please add `@return` phpdoc annotations to them?

bwoebi: the getters in Client.php probably too.

ext/ddtrace.c
```diff
@@ -2951,6 +2953,125 @@ PHP_FUNCTION(DDTrace_flush_endpoints) {
...
+        Z_PARAM_ARRAY(attrs_zv)
+    ZEND_PARSE_PARAMETERS_END();
+
+    type_id = (int32_t) type_id_zl;
+    attributes = Z_ARRVAL_P(attrs_zv);
+    attrs_count = zend_hash_num_elements(attributes);
+
+    if (attrs_count > 0) {
+        c_attrs = ecalloc(attrs_count, sizeof(struct ddog_FfeAttribute));
+        ZEND_HASH_FOREACH_STR_KEY_VAL(attributes, key, value) {
```
bwoebi: One question, just for confirmation / awareness:

In PHP any string key representable as an integer gets forcibly converted to integer in an array.
This iteration will skip all integer values.

So, I assume, it's impossible to have an attribute named `"1234"` or any other number? Otherwise you'll have to handle that edge case.

src/api/FeatureFlags/Client.php
leoromanovsky: Thanks! Doing a proactive scan for other redundant methods.

morrisonlevi (approved):
Looks good to me.

PR #3907, opened 2026-05-22
test(profiling): ASAN test for #3898

realFlowControl:
### Description

Adds a test for #3898 / #3896

This does not yet trigger in CI and that is because we do not run profiler tests against ZTS ASAN because the ZTS ASAN images we prebuild only have DEBUG ZTS ASAN and no NDEBUG version, but the profiler does not build against DEBUG PHP builds.

I hacked this a bit locally and not only does this newly added test fail, but also the other preloading test:

```
  Number of tests : 40
  Tests skipped   : 11
  Tests failed    : 2
  Tests passed    : 27

  Failed tests:

  [profiling] Profiling should only be enabled after preloading has happened
  [profiling] ASAN regression: ZTS preloading must not duplicate INI values from themselves

  ASAN caught the expected issue:

  ERROR: AddressSanitizer: heap-use-after-free
  READ of size 4
  #0 zend_string_dup .../Zend/zend_string.h:217
  #1 zai_config_ini_rinit .../zend_abstract_interface/config/config_ini.c:429
  ...
  freed by thread T0 here:
  #1 zend_string_release .../Zend/zend_string.h:339
  #2 zai_config_ini_rinit .../zend_abstract_interface/config/config_ini.c:428
```

### DIFF
--- a/profiling/tests/phpt/preload_ini_zts_asan.phpt
+++ b/profiling/tests/phpt/preload_ini_zts_asan.phpt
@@ -1,0 +1,34 @@
+--TEST--
+[profiling] ASAN regression: ZTS preloading must not duplicate INI values from themselves
+--DESCRIPTION--
+During preloading in ZTS builds, EG(ini_directives) is still the global INI
+table. The ZTS INI synchronization step must be skipped in that state. If it is
+not skipped, it releases zend_ini_entry.value and then duplicates the same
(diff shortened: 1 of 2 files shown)


realFlowControl:
We already have a test that fails, it is just not executed, as we where missing ZTS ASAN tests, but https://github.com/DataDog/dd-trace-php/pull/3908 adds these

PR #3908, opened 2026-05-22
feat(profiler): Support PHP DEBUG builds for the profiler

realFlowControl:
### Description

This adds PHP `DEBUG` build support to the profiler. It is not meant to be shipped as of know, we just need this in order to run ASAN tests for the Profiler in PHP ZTS builds, as the docker images we are using for our tests do only have a ZTS ASAN version that is also DEBUG.

This PR's CI already shows, that adding this would have shown us the bug reported as #3896 in our own CI early.

Plan:
- merge #3898 into `master`
- rebase this PR on `master` so we have the fix
- review and merge this PR into `master` 

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.github/workflows/prof_asan.yml
+++ b/.github/workflows/prof_asan.yml
@@ -5,10 +5,12 @@ on:
 
 jobs:
   prof-asan:
+    name: PHP ${{ matrix.php-version }} ${{ matrix.php-build }} (${{ matrix.runner }})
     runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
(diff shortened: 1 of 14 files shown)


bwoebi (approved):
Thank you! I always hated that there were no profiler debug builds!

morrisonlevi (approved):
Looks good! Well, looks _terrible_ but that's because I hate the debug API. Really wish it had the same API/ABI and everything was drop-in compatible (so you can have a debug extension on prod PHP and so on). But that's a tangent!

PR #3909, opened 2026-05-22
Add FFE evaluation completion hook

leoromanovsky:
## Motivation

EVP exposure delivery and OTLP evaluation metrics both need the same internal post-evaluation signal. This PR adds that shared hook layer on top of the runtime evaluation work in #3906.

It does not add exposure transport, caching, batching, OpenFeature SDK hook behavior, or metric export behavior.

Planning/reference doc: https://docs.google.com/document/d/1NvMfTpZWLBlFmEFNjdnlMyeVpy5l7KD8qujGFco6w2w/edit?tab=t.0

## Decisions

- The hook fires after `Client::evaluate()` produces `EvaluationDetails` for evaluated success/default/error results.
- Caller validation failures before evaluation, such as invalid flag keys or invalid default types, do not enter the hook.
- Hook exceptions are swallowed; telemetry side effects must never alter returned flag values.
- The hook is internal and dependency-injected through `createWithDependencies()` only.
- The hook lives in the Datadog feature-flag client layer so PHP 7 API and PHP 8 OpenFeature bridge paths share the same post-evaluation signal.
- `DefaultEvaluationCompletedHook`, `CompositeEvaluationCompletedHook`, and OpenFeature SDK hook behavior land with their first consumers in #3910 / #3911.

## Where this PR fits in the stack

This is the **internal hook** layer. Runtime evaluations (#3906) sit beneath it; EVP exposures (#3910) and OTLP metrics (#3911) are independent siblings directly on top of it.

pr3909-hook-layer-stack-position

## Where this PR fits in the target system

This PR contributes only the internal DD Client post-evaluation envelope and callback point. Downstream writers, sidecar FFIs, libdatadog sidecar forwarding, and Agent / OTLP backends land in #3910 / #3911 / DataDog/libdatadog#2026.

pr3909-hook-layer-system-scope

## Changes

- Adds an internal `EvaluationCompleted` envelope with flag key, value/default types, targeting context, result value, reason, variant, error fields, allocation key, and `doLog`.
- Adds `EvaluationCompletedHook` plus a no-op implementation.
- Invokes the hook from `DDTrace\FeatureFlags\Client::evaluate()` after evaluation details are produced.
- Adds unit coverage for the Datadog PHP API path and the PHP 8 OpenFeature provider bridge path.


leoromanovsky:
Folded into #3910 per architectural review — keeping the M2/M3 base seam separate from its first consumer added review burden without removing it (no in-PR consumer is hard to justify). PR #3910 (M2 exposures) now lands the shared evaluation-completed envelope/hook and its first consumer in a single PR. The branch `leo.romanovsky/m2-m3-evaluation-completed-base` is preserved for traceability.

PR #3910, opened 2026-05-22
Add FFE exposure emission

leoromanovsky:
## Motivation

PHP FFE exposure delivery needs to follow the native runtime architecture from the shared design doc: https://docs.google.com/document/d/1NvMfTpZWLBlFmEFNjdnlMyeVpy5l7KD8qujGFco6w2w/edit?tab=t.0

DataDog/dd-trace-php#3906 gives PHP a real Remote Config-backed evaluator. This PR adds the exposure path on top of that runtime without routing product exposure delivery through PHP writer or transport code. The goal is to record `doLog=true` evaluations cheaply during native evaluation, avoid I/O on the evaluation path, and let the shared sidecar own cross-request exposure deduplication.

## Changes

This PR is stacked on DataDog/dd-trace-php#3906 and points the `libdatadog` submodule at DataDog/libdatadog#2026 (`8be471fbc`), which is exposure-only. It records exposure candidates from the native `DDTrace\ffe_evaluate` result when `doLog`, `allocationKey`, and `variant` are present, buffers those candidates in request-local native memory, and flushes them during request shutdown to the sidecar.

The result FFI shape was also simplified per review feedback: Rust now exposes nullable raw `zend_string*` pointers for `FfeResult` strings. C can move those strings directly into PHP result properties and exposure recording without the extra result accessor/cleanup machinery.

## Decisions

Exposure logging remains separate from evaluation metrics. This PR intentionally expects EVP exposures and no OTLP metric stream; DataDog/dd-trace-php#3911 owns metric delivery and DataDog/libdatadog#2052 owns the metric-only sidecar work.

No sidecar or network I/O happens in `ddog_ffe_evaluate`; request shutdown is the flush point. The sidecar owns exposure cache persistence across requests and threads.

## Validation

Focused PHP test on this branch after the `libdatadog` bump:

```sh
cd /Users/leo.romanovsky/go/src/github.com/DataDog/dd-trace-php-ffe-runtime-first
MAX_TEST_PARALLELISM=1 make test_c TESTS=tests/ext/ffe/native_bridge_evaluate.phpt
```

Result: `1/1 passed`.

Local ffe-dogfooding validation against branch `leo.romanovsky/m2-ffe-exposures` now pushed as `416749dd3`, with `libdatadog` submodule `8be471fbc` and `ffe-dogfooding` main `d8e2b540`:

```sh
DD_TRACE_PHP_PATH=/Users/leo.romanovsky/go/src/github.com/DataDog/dd-trace-php-ffe-runtime-first \
DD_TRACE_PHP_FORCE_REBUILD=1 \
DD_TRACE_PHP_BUILD_JOBS=8 \
docker-compose -f docker-compose.yml -f local/docker-compose.php.yml \
  up -d --build --force-recreate app-php7 app-php8-openfeature evaluator

FLAG_KEY=ffe-dogfooding-string-flag \
SUBJECT_ID=php-exposure-validator-1780007366 \
REQUESTS=5 \
WAIT_SECONDS=60 \
EXPECT_OTLP=0 \
EXPECT_EVP=1 \
scripts/validate-local-php-telemetry.sh
```

Result:

```text
OTLP metrics not observed: php7=0 php8-openfeature=0 flag=ffe-dogfooding-string-flag
EVP exposure cache OK: php7=1 php8-openfeature=1 flag=ffe-dogfooding-string-flag subject=php-exposure-validator-1780007366 requests=5
```

Dashboard responded at `http://localhost:8080/dashboard`; both PHP apps reported `PROVIDER_READY` with `hasConfig=true`. A live spot check returned `variant_2` and `evaluation_reason=SPLIT` for PHP 7 and PHP 8/OpenFeature. Mock intake retained exactly two exposure entries for the subject, one for `ffe-dogfooding-php7` and one for `ffe-dogfooding-php8-openfeature`, with allocation key `e88251cd-6b34-43ad-b6fa-fc9c25594088`.

Local system-tests validation through DataDog/system-tests#7031:

```sh
cd /Users/leo.romanovsky/go/src/github.com/DataDog/dd-trace-php-ffe-runtime-first
./tooling/bin/build-debug-artifact gnu-aarch64-8.2-nts \
  /Users/leo.romanovsky/go/src/github.com/DataDog/system-tests-pr7031/binaries

cd /Users/leo.romanovsky/go/src/github.com/DataDog/system-tests-pr7031
DOCKER_CONFIG=/tmp/system-tests-docker-config-nocredsstore \
SYSTEM_TEST_BUILD_TIMEOUT=1800 \
./build.sh --library php --weblog-variant php-fpm-8.2

DOCKER_CONFIG=/tmp/system-tests-docker-config-nocredsstore \
TEST_LIBRARY=php \
WEBLOG_VARIANT=php-fpm-8.2 \
./run.sh +l php FEATURE_FLAGGING_AND_EXPERIMENTATION tests/ffe/test_exposures.py
```

Result: `11 passed in 78.22s`.

Related PRs: DataDog/dd-trace-php#3906, DataDog/libdatadog#2026, DataDog/system-tests#7031.


tabgok (approved):
Approving from a conceptual standpoint for IDM.

Still needs code-review/approval from PHP devs

leoromanovsky:
Known CI note: this PHP SDK PR currently has an expected CI failure from a Rust toolchain version mismatch between the newer libdatadog dependency and the dd-trace-php build environment. Please continue review on the code/design; this PR will not be merged until the trace team resolves the Rust version mismatch in the PHP build images/tooling.

leoromanovsky:
Local validation note for the Rust blocker: ffe-dogfooding was validated through its local PHP runtime mode, not the dd-trace-php CI image. The local PHP dogfooding Dockerfiles install Rust with `rustup`, and `scripts/build-local-php.sh` builds the mounted dd-trace-php checkout with `RUSTC_BOOTSTRAP=1 make -j$DD_TRACE_PHP_BUILD_JOBS`. That local path gets around the stale Rust toolchain currently present in PHP CI, so it proves runtime behavior but does not remove the CI merge blocker.

ext/ddtrace.stub.php
```diff
@@ -1061,6 +1061,13 @@ function add_span_flag(\DDTrace\SpanData $span, int $flag): void {}
      */
     function handle_fork(): void {}
 
+    /**
+     * Flushes native FFE exposure batches for integration tests.
+     *
+     * @internal
+     */
+    function flush_ffe_exposures(): bool {}
```
bwoebi: Could you please move this under the `DDTrace\Testing` namespace? Internal is for being used in our src/, not tests/.

ext/sidecar.c
```diff
@@ -675,6 +721,96 @@ void ddtrace_sidecar_dogstatsd_count(zend_string *metric, zend_long value, zval
...
 }
 
+void ddtrace_ffe_record_exposure(
+    const char *flag_key,
+    size_t flag_key_len,
+    const char *targeting_key,
+    size_t targeting_key_len,
+    zend_string *subject_attributes_json,
+    const char *allocation_key,
+    const char *variant
```
bwoebi: What's the motivation of splitting into char*+len, just to recreate zend_string later? Pass zend_string directly and use `zend_string_copy` to incref.

ext/ddtrace.c
```diff
@@ -1932,6 +1936,8 @@ static PHP_RSHUTDOWN_FUNCTION(ddtrace) {
         ddtrace_rshutdown_remote_config();
     }
 
+    ddtrace_ffe_flush_exposures();
```
bwoebi: It is known and intended that for long-running CLI scripts, exposures will only be flushed at the request end, right?

ext/sidecar.c
```diff
@@ -28,6 +29,51 @@ ddog_Endpoint *ddtrace_endpoint;
...
+    zend_string *subject_id;
+    zend_string *subject_attributes_json;
+    zend_string *allocation_key;
+    zend_string *variant;
+} ddtrace_ffe_exposure;
+
+static void ddtrace_ffe_release_exposure(ddtrace_ffe_exposure *exposure) {
+    if (!exposure) {
+        return;
+    }
```
bwoebi: unreachable branch

ext/sidecar.c
```diff
@@ -28,6 +29,51 @@ ddog_Endpoint *ddtrace_endpoint;
...
+    if (exposure->subject_id) {
+        zend_string_release(exposure->subject_id);
+    }
+    if (exposure->subject_attributes_json) {
+        zend_string_release(exposure->subject_attributes_json);
+    }
+    if (exposure->allocation_key) {
+        zend_string_release(exposure->allocation_key);
+    }
+    if (exposure->variant) {
```
bwoebi: No branches needed, these pointers are always set.

ext/ddtrace.c
leoromanovsky: Yes, that is intentional for this PR. The exposure buffer is request-local native memory and is flushed at RSHUTDOWN. For PHP-FPM, it flushes once per request; for CLI, the script lifetime is the request lifetime.

The tradeoff is intentional: evaluation stays I/O-free, and the sidecar receives/deduplicates exposures only at flush time. The caveat is long-running CLI scripts: exposures are delayed until shutdown, and the current in-process buffer is capped at 1000 entries. If we decide long-running CLI needs better behavior, I would handle that as a follow-up with an explicit periodic/threshold flush path rather than adding sidecar I/O to evaluation.

bwoebi: My intuitive approach would be a (fixed size) ring buffer via shared memory (a few pages of shm , which the sidecar polls (e.g. 1x/sec). This would also naturally limit the amount of exposures per second, which the sidecar has to process (hard limit on overhead esssentially).

leoromanovsky:
Revalidated locally; exposures are being received.

ext/sidecar.c
```diff
@@ -27,6 +28,38 @@ ddog_Endpoint *datadog_endpoint;
 ddog_Endpoint *dogstatsd_endpoint; // always set when datadog_endpoint is set
 struct ddog_InstanceId *datadog_sidecar_instance_id;
 
+#define DATADOG_FFE_EXPOSURE_BUFFER_LIMIT 1000
```
bwoebi: Let's create a `tracer/ffe.c` file, instead of stuffing this here. This is a tracer concern, not a (future) common extension concern.

Not that obvious yet, but I've just done the split! Thanks!

tracer/functions.c
```diff
@@ -1649,6 +1686,7 @@ PHP_FUNCTION(DDTrace_ffe_evaluate) {
     ZEND_PARSE_PARAMETERS_END();
 
     type_id = (int32_t) type_id_zl;
+    ddtrace_ffe_refresh_remote_config();
```
bwoebi: why have you reintroduced that? RC is still getting pushed during runtime, no need for polling.

leoromanovsky: Got re-added in the merge conflict by accident; removing in both branches.

bwoebi (approved):
Looks good to me now!

PR #3911, opened 2026-05-22
Add FFE evaluation metrics

leoromanovsky:
## Motivation

PHP FFE evaluation metrics need to follow the native runtime architecture from the shared design doc: https://docs.google.com/document/d/1NvMfTpZWLBlFmEFNjdnlMyeVpy5l7KD8qujGFco6w2w/edit?tab=t.0

DataDog/dd-trace-php#3906 gave PHP a real Remote Config-backed evaluator. This PR adds metric delivery without a PHP OTLP writer, encoder, or transport. PHP records a typed final evaluation result; native/libdatadog sidecar code owns buffering, aggregation, encoding, and OTLP submission.

Local system-test validation also exposed a runtime bug in the evaluator config store: RC could be delivered and ACKed on one native thread while a PHP request evaluated on another thread and still saw `provider_not_ready`. The UFC config now lives in synchronized process-wide Rust state instead of `thread_local!`, so RC-loaded config is visible to later evaluations in the same process. This is the right shape for PHP-FPM/ZTS because the stored value is Rust `Configuration` data, not PHP request-local state.

## Changes

This PR targets current `master` and advances the `libdatadog` submodule to `091c98df6` from DataDog/libdatadog#2076. The companion libdatadog change lets the sidecar session own the OTLP metrics endpoint and preserves request-replayer test behavior through libdatadog endpoint handling.

Native `DDTrace\ffe_evaluate` records evaluation metrics by default. The PHP 8 OpenFeature path disables automatic native recording for the raw evaluation call and then records the final OpenFeature-aware outcome once through the typed `EvaluationMetricRecorder` adapter.

FFE metric buffering and flushing live in `tracer/ffe.c` / `tracer/ffe.h`, matching the tracer/common-extension split requested in review. The flush path sends metric batches to the sidecar without passing an endpoint per flush; the sidecar session is configured once with the resolved OTLP metrics endpoint.

Endpoint resolution now goes through `ext/otel_config.*`. Unix `DD_TRACE_AGENT_URL` values are normalized through libdatadog's UDS helper so `/var/run/datadog/apm.socket` becomes a valid `unix://<hex-socket-path>/v1/metrics` endpoint.

The FFE evaluator config store in `components-rs/ffe.rs` is now process-wide `LazyLock<RwLock<...>>` state. This fixes the observed RC ACK vs evaluation-thread split and adds a regression test asserting config loaded on one thread is visible on another.

## Decisions

Metric delivery remains separate from exposure logging. This PR intentionally expects OTLP evaluation metrics and no EVP exposure stream; DataDog/dd-trace-php#3910 owns exposure delivery and DataDog/libdatadog#2026 owns the exposure-only sidecar work.

The PHP layer is a typed adapter only. It does not aggregate metrics, encode OTLP, or submit HTTP payloads. Users do not need to call anything extra; metrics are recorded by the Datadog provider/evaluator path.

The evaluator config is intentionally not request-local or PHP module-global. FFE config comes from RC and must persist across requests and be visible across native/PHP threads in the same process. The shared state is protected by Rust `RwLock` and stores no PHP `zval`, `zend_string`, or request-owned memory.

## Validation

Code style:

```sh
git diff --check
```

Result: passed locally.

Focused Rust regression tests:

```sh
RUSTC_BOOTSTRAP=1 cargo test -p datadog-php ffe::tests -- --nocapture
```

Result: `2 passed`; includes `configuration_state_is_process_wide` and `empty_targeting_key_is_not_dropped`.

ZTS build proof:

```sh
./tooling/bin/build-debug-artifact gnu-aarch64-8.2-zts <output-dir>
```

Result: passed locally; produced `dd-library-php-1.20.0-aarch64-linux-gnu.tar.gz` from the ZTS target.

Local system-tests validation through DataDog/system-tests#7033:

```sh
# Build local debug artifact. The local VERSION was temporarily stamped
# as 1.21.0-dev so system-tests#7033 activates the PHP metric tests.
./tooling/bin/build-debug-artifact gnu-aarch64-8.2-nts <system-tests>/binaries

SYSTEM_TEST_BUILD_TIMEOUT=1800 ./build.sh php -w php-fpm-8.2

TEST_LIBRARY=php \
WEBLOG_VARIANT=php-fpm-8.2 \
./run.sh +l php FEATURE_FLAGGING_AND_EXPERIMENTATION tests/ffe/test_flag_eval_metrics.py
```

Before the process-wide FFE config fix, the active run failed `12/17` with emitted metrics tagged as `error.type:provider_not_ready` even though RC delivery was ACKed. After the fix, result was:

```text
17 passed in 81.31s
```

FFE dogfooding transport proof:

```sh
cd ffe-dogfooding
FLAG_KEY=php-metrics-dogfood-1780520626 \
SUBJECT_ID=php-metrics-validator-1780520626 \
REQUESTS=5 \
WAIT_SECONDS=120 \
EXPECT_OTLP=1 \
EXPECT_EVP=0 \
scripts/validate-local-php-telemetry.sh
```

Result:

```text
OTLP metrics OK: php7=5 php8-openfeature=5 flag=php-metrics-dogfood-1780520626
EVP exposures not observed: php7=0 php8-openfeature=0 flag=php-metrics-dogfood-1780520626 subject=php-metrics-validator-1780520626
```

Dogfooding did not prove live non-default RC values in this run because the local agent was rejecting the configured key with `API Key invalid (403 response)`. The system-tests run above is the config-backed validation evidence for this PR.

Related PRs: DataDog/dd-trace-php#3906, DataDog/dd-trace-php#3910, DataDog/libdatadog#2026, DataDog/libdatadog#2076, DataDog/system-tests#7033.


src/api/FeatureFlags/Internal/Metric/OtlpHttpMetricTransport.php
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        return 'http://' . $host . ':' . self::DEFAULT_OTLP_PORT . '/v1/metrics';
+    }
+
+    private function post($endpoint, $payload)
+    {
+        $parts = parse_url($endpoint);
+        if (!is_array($parts) || !isset($parts['scheme'], $parts['host']) || strtolower($parts['scheme']) !== 'http') {
+            return false;
```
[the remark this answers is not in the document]
leoromanovsky: @datadog fix this

leoromanovsky:
Known CI note: this PHP SDK PR currently has an expected CI failure from a Rust toolchain version mismatch between the newer libdatadog dependency and the dd-trace-php build environment. Please continue review on the code/design; this PR will not be merged until the trace team resolves the Rust version mismatch in the PHP build images/tooling.

leoromanovsky:
Local validation note for the Rust blocker: ffe-dogfooding was validated through its local PHP runtime mode, not the dd-trace-php CI image. The local PHP dogfooding Dockerfiles install Rust with `rustup`, and `scripts/build-local-php.sh` builds the mounted dd-trace-php checkout with `RUSTC_BOOTSTRAP=1 make -j$DD_TRACE_PHP_BUILD_JOBS`. That local path gets around the stale Rust toolchain currently present in PHP CI, so it proves runtime behavior but does not remove the CI merge blocker.

leoromanovsky:
Merged latest main and libdatadog; resolved conflicts. Refactored according to new organization and re-ran local e2e test.

morrisonlevi (commented):
Unix Domain Sockets (UDS) don't work correctly because the URI authority is not hex-encoded. This is required because the authority "/var/run/datadog/apm.socket" contains slashes, but it's part of the authority, not the URI path. So the final URI should look something like `unix://2f7661722f72756e2f64617461646f672f61706d2e736f636b6574/v1/metrics` for that example "/var/run/datadog/apm.socket" domain socket.

libdatadog and components-rs have various helpers. Maybe one of these will help:

 - `ddog_endpoint_from_url`
 - `datadog_parse_agent_url`

I don't know the FFI bits as well, since I mostly live in the Rust bindings personally.

You should be able to write a test for this with the request replayer, I believe it supports UDS these days.

tracer/ffe.c
```diff
@@ -22,6 +34,270 @@ typedef struct {
...
+        ZSTR_LEN(agent_host)
+    );
+    if (agent_scheme) {
+        zend_string_release(agent_scheme);
+    }
+    return endpoint;
+}
+
+static zend_string *datadog_ffe_otlp_metrics_endpoint(void) {
+    const char *metrics_endpoint = getenv("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT");
```
bwoebi: Any access to otel config must go through configuration.h/otel_config.c.

src/api/FeatureFlags/Internal/Metric/EvaluationMetricRecorder.php
```diff
@@ -0,0 +1,75 @@
...
+        if (function_exists('dd_trace_env_config')) {
+            return \dd_trace_env_config('DD_METRICS_OTEL_ENABLED') === true;
+        }
+
+        $value = getenv('DD_METRICS_OTEL_ENABLED');
+        if ($value === false) {
+            return false;
+        }
+
+        return in_array(strtolower((string) $value), array('1', 'true', 'yes', 'on'), true);
```
bwoebi:
```suggestion
        return \dd_trace_env_config('DD_METRICS_OTEL_ENABLED') === true;
```
The tracer is always loaded when this code is reached. Avoid guards and direct getenv here.

src/api/FeatureFlags/Internal/NativeEvaluator.php
```diff
@@ -9,24 +9,26 @@ final class NativeEvaluator implements Evaluator
...
 
-    private function __construct($mapper = null)
+    private function __construct($mapper = null, $recordMetrics = true)
     {
         if ($mapper !== null && !$mapper instanceof ResultMapper) {
             throw new \InvalidArgumentException('Expected a ResultMapper instance');
         }
 
         $this->mapper = $mapper ?: new ResultMapper();
+        $this->recordMetrics = (bool) $recordMetrics;
```
bwoebi:
```suggestion
    private function __construct($mapper = null, bool $recordMetrics = true)
    {
        if ($mapper !== null && !$mapper instanceof ResultMapper) {
            throw new \InvalidArgumentException('Expected a ResultMapper instance');
        }

        $this->mapper = $mapper ?: new ResultMapper();
        $this->recordMetrics = $recordMetrics;
```

tracer/ffe.c
leoromanovsky: Thanks, yeah, I had a bad feeling about that one but hadn't discovered the canonical pattern; appreciate you taking a close look.

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -80,6 +110,8 @@ private function resolve(
         ?EvaluationContext $context
     ): ResolutionDetailsInterface {
         $details = $this->evaluate($flagKey, $expectedType, $defaultValue, $this->normalizeContext($context));
+        $this->warnIfNonProductionRuntime($details);
+        $this->recordEvaluationMetric($flagKey, $details);
```
sameerank, line 119: The other SDKs have been using the `finally` hook to record evaluations. E.g.

Ruby: https://github.com/DataDog/dd-trace-rb/blob/6cf558f959f5edff37b6a243d25754687032efc9/lib/datadog/open_feature/hooks/flag_eval_hook.rb#L29
Go: https://github.com/DataDog/dd-trace-go/blob/f7dde0ce6bcea5417dc6663cf8b089d8e5555129/openfeature/flageval_metrics.go#L61
Java: https://github.com/DataDog/dd-trace-java/blob/81bb30521d96519d510b0446c5ca3eefb631e0b9/products/feature-flagging/feature-flagging-api/src/main/java/datadog/trace/api/openfeature/FlagEvalHook.java#L36

Explanation for why we use finally in JS: https://github.com/DataDog/dd-trace-js/blob/41789b40877c2072f8022fd3d854749741103d26/packages/dd-trace/src/openfeature/eval-metrics-hook.js#L14-L20

And I see that our reason for not doing that here is that `ResolutionDetails` aren't included in OpenFeature's implementation https://github.com/open-feature/php-sdk/blob/af1de424cbe9dcaae0c8b17d92754ffcd28d5f2f/src/implementation/hooks/HookExecutor.php#L81-L88 so it's not possible to use the `finally` hook here.

It took a bit of cross-referencing to understand why this differed from the other SDKs, so worth adding a comment explaining why we do this differently

Otherwise I think this is a good approach given the limitations 👍 and hopefully at some point we can update the `^2.1` version constraint

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -100,6 +132,27 @@ private function resolve(
...
+            $details->getReason(),
+            $details->getErrorCode(),
+            $this->allocationKey($details)
+        ));
+    }
+
+    private function allocationKey(EvaluationDetails $details): ?string
+    {
+        $exposure = $details->getExposureData();
+        if (!is_array($exposure) || !isset($exposure['allocationKey']) || !is_string($exposure['allocationKey'])) {
```
sameerank: I'm recalling that in the other SDKs we wanted to indicate that this was an internal key in the name https://github.com/DataDog/dd-trace-rb/pull/5599#discussion_r3169157112

But I guess that's not possible here without a concept of `flagMetadata`

tracer/ffe.c
```diff
@@ -22,6 +33,133 @@ typedef struct {
...
+        DATADOG_G(ffe_metric_buffer_cap) = new_cap;
+    }
+
+    datadog_ffe_metric *buffer = (datadog_ffe_metric *) DATADOG_G(ffe_metric_buffer);
+    datadog_ffe_metric *metric = &buffer[DATADOG_G(ffe_metric_buffer_len)++];
+    metric->flag_key = zend_string_init(flag_key, flag_key_len, 0);
+    metric->variant = zend_string_init(variant ? variant : "", variant ? variant_len : 0, 0);
+    metric->reason = zend_string_init(reason ? reason : "", reason ? reason_len : 0, 0);
+    metric->error_type = zend_string_init(error_type ? error_type : "", error_type ? error_type_len : 0, 0);
+    metric->allocation_key = zend_string_init(allocation_key ? allocation_key : "", allocation_key ? allocation_key_len : 0, 0);
```
sameerank: So on a successful eval, this sends `error_type:""` and `allocation_key:""`, but [the sidecar omits them in `ffe_metric_from_ffi`](https://github.com/DataDog/libdatadog/blob/582bee193e6f4688fa546ca0b3e3b382de672538/datadog-sidecar-ffi/src/lib.rs#L1303), which matches what the other SDKs do

Libdatadog integrations are a bit opaque so this looked like an accident and took cross-referencing to realize that it wasn't. A comment would be helpful here too

sameerank (approved):
Looked at this mainly checking for parity with the other server SDKs, and I think overall it looks right. Nice work!

bwoebi (approved):
Looks like all non-flaky tests pass now

PR #3912, opened 2026-05-23
Split the tracer parts off from ext directory

bwoebi:
Move all the things.
Drop a few obsolete things.
Change nearly no code.

- git data gets collected in ext/, but git object is collected in tracer/
- the sidecar is now always spawned, unconditionally


ext/sidecar.c
```diff
@@ -656,279 +634,205 @@ void ddtrace_sidecar_push_tags(ddog_Vec_Tag *vec, zval *tags) {
...
+        datadog_sidecar_push_tag(vec, dd_zend_string_to_CharSlice(key), dd_zend_string_to_CharSlice(Z_STR(value_str)));
         zend_string_release(Z_STR(value_str));
     }
     ZEND_HASH_FOREACH_END();
 }
 
-void ddtrace_sidecar_dogstatsd_count(zend_string *metric, zend_long value, zval *tags) {
-    if (!DDTRACE_G(sidecar) || !get_DD_INTEGRATION_METRICS_ENABLED()) {
+void datadog_sidecar_dogstatsd_count(zend_string *metric, zend_long value, zval *tags) {
+    if (!DATADOG_G(sidecar)) {
```
Leiyks, line 645: is the drop of `DD_INTEGRATION_METRICS_ENABLED` intentional ?

tracer/tracer_api.h
```diff
@@ -0,0 +1,52 @@
...
+#else
+static inline void ddtrace_populate_span_data(ddtrace_span_data *span, zend_string **service, zend_string **env, zend_string **version) {
+    *service = NULL;
+    *env = NULL;
+    *version = NULL;
+}
+#endif
+extern ddog_LiveDebuggerSetup ddtrace_live_debugger_setup;
+
+ddtrace_span_data *ddtrace_active_span(void);
```
Leiyks: Shoudn't this one be in a `#ifdef DDTRACE` guard as well ?

Leiyks (approved):
lgtm 👌

ext/sidecar.c
bwoebi: yes. I have no idea why it was there in the first place (and `DD_INTEGRATION_METRICS_ENABLED` lives in tracer config, rather than ext/ config).

That function was (originally) introduced for OpenAI integration, but that integration already checks this.

tracer/tracer_api.h
bwoebi: Agree, let's fully ifdef this, gives compile warnings rather than runtime link failures.

PR #3915, opened 2026-05-24
fix: correct unit mismatch in php.compilation.total_time_ms

### LINKED ISSUE #3914: php.compilation.total_time_ms reports microseconds instead of milliseconds since v0.98.0
## Description

Since PR #2230 (v0.98.0), the span metric `php.compilation.total_time_ms` reports values in **microseconds** rather than **milliseconds**, inflating reported values by 1000x.

## Root cause

PR #2230 replaced the `_get_microseconds()` helper (which used `clock_gettime` and returned microseconds) with `zend_hrtime()` (which returns nanoseconds) in `ext/engine_hooks.c`, but did not update the accumulator or the serializer divisor:

```c
// Before (correct): _get_microseconds() returned microseconds
DDTRACE_G(compile_time_microseconds) += (int64_t)(_get_microseconds() - start);

// After (incorrect): zend_hrtime() returns nanoseconds, stored as-is
DDTRACE_G(compile_time_microseconds) += (int64_t)(zend_hrtime() - start);
```

The serializer in `ext/serializer.c` divides by `1000.` expecting microseconds → milliseconds, but now receives nanoseconds, producing microseconds labeled as milliseconds.

Notably, the same PR correctly handled this conversion in `circuit_breaker.c`:
```c
return zend_hrtime() / 1000;  // correctly converts nanoseconds → microseconds
```

## Evidence

A request with a total duration of ~4,350 ms reported `php.compilation.total_time_ms` of ~28,346 — impossible if in milliseconds (exceeds request duration), but perfectly reasonable as ~28.3 ms in microseconds.

## PHP version

All versions (the bug is in the tracer, not PHP)

## Tracer version

\>= 0.98.0

## Installed extensions

N/A — affects the tracer itself

## Output of phpinfo()

N/A

dortort:
## Description

PR #2230 (v0.98.0) replaced `_get_microseconds()` (which used `clock_gettime` and returned microseconds) with `zend_hrtime()` (which returns nanoseconds) in `ext/engine_hooks.c`, but did not update the accumulator. The serializer in `ext/serializer.c` divides by `1000.` expecting microseconds → milliseconds, so since v0.98.0 the metric `php.compilation.total_time_ms` has been reporting **microseconds labeled as milliseconds** — a 1000x inflation.

The same PR correctly handled this conversion in `circuit_breaker.c` (`return zend_hrtime() / 1000;`). This change applies the same pattern to the compile time accumulator.

### Evidence

A request with a total duration of ~4,350 ms reported `php.compilation.total_time_ms` of ~28,346 — impossible if in milliseconds (exceeds request duration), but consistent with ~28.3 ms when interpreted as microseconds.

Fixes #3914

### DIFF
--- a/ext/engine_hooks.c
+++ b/ext/engine_hooks.c
@@ -77,7 +77,7 @@ static zend_op_array *_dd_compile_file(zend_file_handle *file_handle, int type) 
     zend_op_array *res;
     uint64_t start = zend_hrtime();
     res = _prev_compile_file(file_handle, type);
-    DDTRACE_G(compile_time_microseconds) += (int64_t)(zend_hrtime() - start);
+    DDTRACE_G(compile_time_microseconds) += (int64_t)((zend_hrtime() - start) / 1000);
     return res;
 }
 


bwoebi (approved):
Thank you!

PR #3920, opened 2026-05-27
Fix void post hook overwriting retval in OTel hook polyfill

Leiyks:
## Summary

- `dd_uhook_end` in `ext/hook/uhook_otel.c` masked the post closure's return type with `IS_VOID` (the raw type code, `14`) instead of `MAY_BE_VOID` (the bitmask flag `1 << 14` used inside `PURE_MASK`). The AND between a bitmask and a raw type code is always zero, so the "skip retval replacement for `:void` post hooks" branch never fired and every `:void` post hook silently overwrote the function's actual return value with the closure's implicit-null return.
- Add a `.phpt` regression covering both branches (`:void` preserves retval, typed-non-void overrides retval).

## How the bug surfaced

dd-trace polyfills `\OpenTelemetry\Instrumentation\hook` and forces `extension_loaded("opentelemetry")` to true (`ext/hook/uhook_otel.c:269`). Packages like `opentelemetry-auto-symfony` register hooks against this polyfill even when the real OTel C extension is absent.

`SymfonyInstrumentation::register()` installs a `:void` post hook on `Symfony\Component\HttpKernel\HttpKernel::handle`. With the buggy mask, that `:void` closure's implicit null return was copied into the real Response retval. The outer `Symfony\Component\HttpKernel\Kernel::handle`'s `:Response` runtime check then fired:

```
TypeError: Symfony\Component\HttpKernel\Kernel::handle(): Return value must be of type
Symfony\Component\HttpFoundation\Response, null returned
```

Not Symfony-version-specific. Triggers for any `:void` post hook on any function with a non-void return type.

### DIFF
--- a/ext/hook/uhook_otel.c
+++ b/ext/hook/uhook_otel.c
@@ -187,7 +187,7 @@ static void dd_uhook_end(zend_ulong invocation, zend_execute_data *execute_data,
 
     if (!Z_ISUNDEF(rv)) {
         const zend_function *func = zend_get_closure_method_def(def->end);
-        if (func->common.fn_flags & ZEND_ACC_HAS_RETURN_TYPE && (ZEND_TYPE_PURE_MASK(func->common.arg_info[-1].type) & IS_VOID) == 0) {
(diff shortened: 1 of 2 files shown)


bwoebi:
Fix is right, the test needs fixing :-)

bwoebi (approved):

bwoebi:
Nice find!

PR #3925, opened 2026-05-27
feat(appsec): collect security-testing headers on HTTP entry spans

christophe-papazian:
APPSEC-67550

Unconditionally collect `x-datadog-endpoint-scan` and `x-datadog-security-test` HTTP request headers as `http.request.headers.*` span tags on the service entry span, independent of `DD_TRACE_HEADER_TAGS` and AppSec enablement. Also forwarded to the inferred proxy span.

Covers all PHP SAPIs (Apache/fpm, CLI, FrankenPHP, RoadRunner) via the C extension and Swoole via its PHP integration.

RFC: [Security Testing: Trace Attribution for Inventory Enrichment and Pollution Prevention](https://docs.google.com/document/d/1uR4QQvU8pItEV2zFqr3-L6jO2jxzmvLrFTX_yyvqIOA)

## Test plan
- `tests/ext/root_span_security_testing_headers.phpt` — headers captured, independent of `DD_TRACE_HEADER_TAGS`
- `tests/ext/root_span_security_testing_headers_absent.phpt` — tags absent when headers not sent
- `tests/ext/inferred_proxy/security_headers_forwarded.phpt` — both spans carry the tags
- `tests/Integrations/Swoole/SecurityTestingHeadersTest.php` — Swoole path: present + absent


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Changelog for older versions can be found in our [release page](https://github.c
 
 ## Tracer
 ### Added
+- Unconditionally collect `x-datadog-endpoint-scan` and `x-datadog-security-test` HTTP request headers as span tags on HTTP entry spans, independent of `DD_TRACE_HEADER_TAGS` and AppSec enablement (APPSEC-62412)
```
estringana: We usually do this at releasing time. Make sure the description of your PR is clear and it will be added here when deployed

tests/ext/root_span_security_testing_headers.phpt
```diff
@@ -0,0 +1,21 @@
+--TEST--
+Security-testing headers are collected unconditionally on the root span
+--ENV--
+DD_TRACE_AUTO_FLUSH_ENABLED=0
+DD_TRACE_GENERATE_ROOT_SPAN=0
+DD_TRACE_HEADER_TAGS=
+HTTP_X_DATADOG_ENDPOINT_SCAN=endpoint-scan-uuid
+HTTP_X_DATADOG_SECURITY_TEST=security-test-uuid
+--GET--
+foo=bar
```
estringana: I don't think you need this `foo=bar` on the tests, do you?

tracer/serializer.c
```diff
@@ -693,6 +693,30 @@ static void dd_set_entrypoint_root_span_props(struct superglob_equiv *data, ddtr
     }
 
     if (data->server) {
+        // Unconditionally collect security-testing headers (APPSEC-62412)
+        static const struct {
```
estringana: I would try to reuse as much as possible the code inside dd_add_header_to_meta. Right now it's checking DD_TRACE_HEADER_TAGS which is not needed for this but all other aldd header related code you can reuse it. Also you can add a macro like this to avoid the duplication 
```
#define DD_UNCONDITIONAL_SERVER_HEADER(server_key, tag) \
    { server_key, sizeof(server_key) - 1, tag, sizeof(tag) - 1 }

static const struct { ... } sec_headers[] = {
    DD_UNCONDITIONAL_SERVER_HEADER("HTTP_X_DATADOG_ENDPOINT_SCAN",
                                   "http.request.headers.x-datadog-endpoint-scan"),
    ...
};
```

ext/serializer.c
```diff
@@ -1846,6 +1870,8 @@ ddog_SpanBytes *ddtrace_serialize_span_to_rust_span(ddtrace_span_data *span, ddo
         transfer_meta_data(rust_span, serialized_inferred_span, "_dd.p.dm", true);
         transfer_meta_data(rust_span, serialized_inferred_span, "_dd.p.ksr", false);
         transfer_meta_data(rust_span, serialized_inferred_span, "_dd.p.tid", true);
+        transfer_meta_data(rust_span, serialized_inferred_span, "http.request.headers.x-datadog-endpoint-scan", false);
```
estringana: You could move this `http.request.headers.x-datadog-endpoint-scan` to a constant since it's used in a few places as well. Same with the other

CHANGELOG.md
christophe-papazian: Done, removed. The PR description already starts with APPSEC-67550 and describes the change.

tests/ext/root_span_security_testing_headers.phpt
christophe-papazian: You're right, removed.

tracer/serializer.c
christophe-papazian: Done — added the `DD_UNCONDITIONAL_SERVER_HEADER` macro exactly as suggested, and also moved the tag-name strings to `DD_TAG_HTTP_REQH_ENDPOINT_SCAN` / `DD_TAG_HTTP_REQH_SECURITY_TEST` constants (see the follow-up comment). Both are used in the struct initializer and the `transfer_meta_data` calls.

One note on the tags.c placement: the RFC requires collection to be unconditional regardless of whether AppSec is enabled, so we kept the `serializer.c` block (tracer, always loaded) and also added the headers to `_relevant_basic_headers` in `tags.c` as you suggested — the two paths compose cleanly since `zend_hash_str_add_new` in the tracer runs at serialization time, and AppSec's `_try_add_tag` (which uses `zend_hash_add`) is a no-op if the key already exists.

ext/serializer.c
christophe-papazian: Done — added `DD_TAG_HTTP_REQH_ENDPOINT_SCAN` and `DD_TAG_HTTP_REQH_SECURITY_TEST` at the top of `serializer.c`. They are now used in both the struct initializer (via the macro) and the `transfer_meta_data` calls.

appsec/src/extension/tags.c
```diff
@@ -305,6 +305,8 @@ static void _init_relevant_headers(void)
     ADD_RELEVANT_BASIC_HEADER("content-type");
     ADD_RELEVANT_BASIC_HEADER("user-agent");
     ADD_RELEVANT_BASIC_HEADER("accept");
+    ADD_RELEVANT_BASIC_HEADER("x-datadog-endpoint-scan");
```
estringana: Since this tags should be reported regardless of appsec present or not. I don't think this is required

christophe-papazian: Agreed, removed. The `serializer.c` block covers it unconditionally for all deployments, so the AppSec-specific addition is redundant.

ext/serializer.c
```diff
@@ -693,6 +696,32 @@ static void dd_set_entrypoint_root_span_props(struct superglob_equiv *data, ddtr
     }
 
     if (data->server) {
+        // Security-testing headers (APPSEC-62412): collected unconditionally
+        // here so they are present even when the AppSec extension is not loaded.
+        // The AppSec extension also collects them via _relevant_basic_headers
+        // in appsec/src/extension/tags.c.
```
cataphract: Update this comment; there's no change in appsec/**/tags.c

tracer/serializer.c
```diff
@@ -1848,6 +1877,8 @@ ddog_SpanBytes *ddtrace_serialize_span_to_rust_span(ddtrace_span_data *span, ddo
         transfer_meta_data(rust_span, serialized_inferred_span, "_dd.p.dm", true);
         transfer_meta_data(rust_span, serialized_inferred_span, "_dd.p.ksr", false);
         transfer_meta_data(rust_span, serialized_inferred_span, "_dd.p.tid", true);
+        transfer_meta_data(rust_span, serialized_inferred_span, DD_TAG_HTTP_REQH_ENDPOINT_SCAN, false);
+        transfer_meta_data(rust_span, serialized_inferred_span, DD_TAG_HTTP_REQH_SECURITY_TEST, false);
```
cataphract, line 1841: Are you sure abou this? You need the tags in both spans?

tracer/serializer.c
```diff
@@ -693,6 +696,32 @@ static void dd_set_entrypoint_root_span_props(struct superglob_equiv *data, ddtr
...
+        for (size_t i = 0; i < sizeof(sec_headers) / sizeof(*sec_headers); i++) {
+            zval *hval = zend_hash_str_find(data->server, sec_headers[i].server_key, sec_headers[i].server_len);
+            if (hval) {
+                ZVAL_DEREF(hval);
+                if (Z_TYPE_P(hval) == IS_STRING) {
+                    zval zv;
+                    ZVAL_STR_COPY(&zv, Z_STR_P(hval));
+                    zend_hash_str_add_new(meta, sec_headers[i].tag, sec_headers[i].tag_len, &zv);
+                }
+            }
```
cataphract: I imagine this will be sent only in a very small amount of requests. Yet, you're adding overhead in every request looking for them.

Ideally you'd add it to the hastable returned by get_DD_TRACE_HEADER_TAGS() or equivalent, though I think this would require a startup change after zai_config_minit() modyfying `zai_config_memoized_entries[DDTRACE_CONFIG_DD_TRACE_HEADER_TAGS].decoded_value` and another change in `ddtrace_alter_DD_TRACE_HEADER_TAGS`, so it might no be worth the complexity.

Anyway, give it a thought.

ext/serializer.c
christophe-papazian: Fixed — removed the stale reference to `appsec/src/extension/tags.c` since that change was reverted.

tracer/serializer.c
christophe-papazian: Yes, intentional. The RFC does say attributes must not be *propagated downstream*, but the inferred proxy span represents an upstream API gateway (e.g., AWS API Gateway) sitting in front of the PHP app — not a downstream service. The api-endpoint-reducer also processes inferred spans for inventory purposes, so tagging it avoids the same scan/test traffic pollution problem there. All other tracer implementations (Java, JS, .NET) also forward these headers to the inferred proxy span.

tracer/serializer.c
christophe-papazian: Good point. Two `zend_hash_str_find` calls per request is O(1) each and minimal in practice, but I understand the concern.

The `DD_TRACE_HEADER_TAGS` approach would mean piggybacking on the existing `HTTP_*` loop that already iterates all `$_SERVER` keys — so overhead-wise it'd be zero *extra* cost for the absent case, but the loop already runs for every HTTP request anyway. The downside is that it conflates user-configurable header tags with RFC-mandated ones, and as you note it requires non-trivial changes to startup and `ddtrace_alter_DD_TRACE_HEADER_TAGS`.

I'll leave the current approach as-is given the tradeoff, but happy to revisit if you feel strongly about it.

estringana:
LGTM. @bwoebi do you mind to review it as well?

tracer/serializer.c
bwoebi: I agree, DD_TRACE_HEADER_TAGS is a better approach, it also scales with any future webserver integration.

christophe-papazian: Changes done, please review again

bwoebi (approved):
Simplified the impl a bit, we can merge this as-is now.

PR #3926, opened 2026-05-27
Bump Rust to 1.87, Clang/LLVM to 20

morrisonlevi:
### Description

This bumps some dependencies:
 - Rust: 1.84 -> 1.87
 - Clang/LLVM: 17/19 -> 20

And also shrinks the bookworm images slightly:
 - Moves chown earlier.
 - Avoids installing the rust HTML docs.
 
 Note that appsec is likely to require significant upgrades as they have not updated since Clang/LLVM 17. I'm working through them but expect it to take a while.

#### Misc Notes

 - I updated gdb in bookworm because the newer compiler choked on the older one.
 - The variable `DDCI_CONFIGURE_OTEL_EXPORTER: "true"` is used when debugging docker buildx jobs, the job can somehow link to the buildx runner. I don't know how to do this, but the folks in our CI & Infra support folks do.

### Reviewer checklist
- [ ] Test coverage seems ok.


.gitlab/ci-images.yml
```diff
@@ -90,40 +93,198 @@ Bookworm:
...
+        Image = "datadog/dd-trace-ci:php-7.3_windows"
+        Dockerfile = "Dockerfile"
+        BuildArgs = @("phpVersion=7.3.33", "vsVersion=vc15", "phpTarGzUrl=https://www.php.net/distributions/php-7.3.33.tar.gz", "phpSha256Hash=9a369c32c6f52036b0a890f290327f148a1904ee66aa56e2c9a7546da6525ec8")
+      }
+      "php-7.2" = @{
+        Image = "datadog/dd-trace-ci:php-7.2_windows"
+        Dockerfile = "Dockerfile"
+        BuildArgs = @("phpVersion=7.2.34", "vsVersion=vc15", "phpTarGzUrl=https://www.php.net/distributions/php-7.2.34.tar.gz", "phpSha256Hash=8b2777c741e83f188d3ca6d8e98ece7264acafee86787298fae57e05d0dddc78")
+      }
+      "php-7.1" = @{
```
leoromanovsky: I noticed 7.0 and 7.1 are dropped from testing above. I assume we are dropping support; if so are these needed?

.gitlab/ci-images.yml
```diff
@@ -92,38 +99,198 @@ Bookworm:
...
+      if ([string]::IsNullOrWhiteSpace($target)) { continue }
+      if (-not $targets.ContainsKey($target)) {
+        Write-Host "ERROR: unknown Windows CI image target $target."
+        exit 1
+      }
+
+      $config = $targets[$target]
+      $image = $config["Image"]
+      $dockerfile = $config["Dockerfile"]
+      $buildCommand = @("build", "--pull", "--no-cache", "-t", "$image", "-f", "$dockerfile")
```
bwoebi: Using docker-compose did not work? (not available or something?)
Disappointing to repeat the definitions here.

morrisonlevi: Correct, no "docker compose" nor "docker-compose" at all, I checked both.

dockerfiles/ci/windows/vc14.Dockerfile
bwoebi, line 1: We can drop this too (vc14 is 7.0/7.1 only), and the vs14_buildtools.exe file too.

bwoebi (approved):
Thanks for taking care of this.

leoromanovsky (approved):
lgtm from ffe; looks like you are keeping all php version support ✔️

.gitlab/ci-images.yml
bwoebi: https://docs.docker.com/compose/install/standalone#on-windows-server that's apparently the instructions to manually install

Trying.

.gitlab/ci-images.yml
morrisonlevi: I'll double check that I fixed them all, but we are not dropping any version support in this PR. What we are intentionally dropping is building 7.0/7.1 _for Windows_ which isn't supported today, but we had various infrastructure around for it. So we are dropping that, because I wasted a nontrivial amount of time trying to get it to work, not realizing we didn't need it! I'm still pulling at the strings to delete everything related to 7.0/7.1, thanks for noticing this one!

However, an AI agent got too eager and deleted support from other platforms too and it slipped past initial review. I've at least partially fixed it, but I'll make sure that Alpine, CentOS, and our Bookworm images all have 7.0/7.1 support before merging.

leoromanovsky: Good call!

PR #3940, opened 2026-05-29
Add dynamic_service DBM propagation mode

amarziali:
### Description

Introduces a new DBM propagation mode value, `dynamic_service`, as a convenience alias for the combination of `DD_DBM_PROPAGATION_MODE=service` and `DD_DBM_INJECT_SQL_BASEHASH=true`. With a single setting, users get service-level SQL comment propagation (`dddbs`, `ddps`, etc.) plus the process-tags base hash injected as `ddsh` in the SQL comment and `_dd.propagated_hash` on the span, without needing to manage two separate variables.

**Behavior:**
- Same as `service` mode: no `traceparent` header, no `_dd.dbm_trace_injected` span tag
- Base hash (`ddsh` / `_dd.propagated_hash`) is always injected regardless of `DD_DBM_INJECT_SQL_BASEHASH`

Please note that this mode will replace `service` on the long term (and hence eventually also the `DD_DBM_INJECT_SQL_BASEHASH` will be deprecated/removed at same time)

### Reviewer checklist
- [ ] Test coverage seems ok.


src/DDTrace/Integrations/DatabaseIntegrationHelper.php
bwoebi, line 43: When full is not possible, shouldn't it fall back to `DBM_PROPAGATION_DYNAMIC_SERVICE` now?

amarziali: this is a good question indeed. I think that right now is more a product choice - when that will be validated it will be the default hence will happen. I can ask btw to know if we can do beforehand

bwoebi (approved):
Looks good to me!

raphaelgavache (approved):

Issue #3917, opened 2026-05-26
bug: PHP-FPM worker crashes with heap corruption in _dd_curl_reset_headers (1.18.0)


masak1yu:
### Summary

PHP-FPM worker processes crash with SIGABRT due to heap corruption in the background writer thread (\`_dd_writer_loop\`) with dd-trace-php 1.18.0. Each crash causes in-flight HTTP requests to be dropped with a 504 Gateway Timeout from the upstream reverse proxy.

### Environment

- **dd-trace-php version:** 1.18.0
- **PHP:** 7.4 NTS, PHP-FPM (standard process manager, not FrankenPHP / not ZTS)
- **OS:** Linux x86_64 (glibc)
- **libcurl:** system libcurl (\`/lib64/libcurl.so.4\`)
- **Rollback version that resolved the issue:** 1.0.0

### Symptom

PHP-FPM worker processes terminate with \`SIGABRT\` under production load. The crash is triggered by glibc detecting heap corruption (\`malloc_printerr\`) when the background writer thread calls \`curl_slist_free_all\` on the stored HTTP header list (\`writer->headers\`).

From the upstream reverse proxy (nginx), the crashing worker produces a \`504 Gateway Timeout\` (\`upstream timed out, 110: Connection timed out\`), interrupting request processing mid-flight.

### Coredump stack trace

```
#0  0x00007f59ee354690 in raise () from /lib64/libpthread.so.0
#1  0x00007f59dfa617b0 in libdd_crashtracker::collector::signal_handler_manager::chain_signal_handler ()
    at libdatadog/libdd-crashtracker/src/collector/signal_handler_manager.rs:125
#2  libdd_crashtracker::collector::crash_handler::handle_posix_sigaction () at libdatadog/libdd-crashtracker/src/collector/crash_handler.rs:209
#3  <signal handler called>
#4  0x00007f59ee594ae0 in raise () from /lib64/libc.so.6
#5  0x00007f59ee595f88 in abort () from /lib64/libc.so.6
#6  0x00007f59ee5d4b94 in __libc_message () from /lib64/libc.so.6
#7  0x00007f59ee5da80a in malloc_printerr () from /lib64/libc.so.6
#8  0x00007f59ee5dad16 in munmap_chunk () from /lib64/libc.so.6
#9  0x00007f59eb30021d in curl_slist_free_all () from /lib64/libcurl.so.4
[...]

PR #3945, opened 2026-06-03
fix(tracer): use atomic_store for writer->headers assignment

realFlowControl:
### Description

`writer->headers` is declared `_Atomic` but was being stored via a plain non-atomic assignment in `dd_curl_set_headers`, while `dd_curl_reset_headers` reads it with `atomic_exchange`. This replaces the plain write with `atomic_store` to match.

Note: the root cause of the crash in #3917 is not fully identified/fixed with this PR

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
I surely can approve this, but atomic_store only matters when it's sequenced relative to other operations.

This shouldn't fix anything at all.

PR #3948, opened 2026-06-03
feat(svc-src): emit _dd.svc_src per Service Override Source Attribution RFC

Leiyks:
Implements [Service Override Source Attribution](https://docs.google.com/document/d/1c47iSTWxIOHMHfZTF2nT9xfyQaIBP9KJvI9sRn5SvpM) for the PHP tracer.

Every span whose service was *actively chosen* (vs. the tracer's default) now carries a `_dd.svc_src` meta tag identifying who chose it. The tag is cleared on default-service spans (DD_SERVICE-set or auto-resolved) per the RFC.

## Source values

| Source | Value | Set by |
|---|---|---|
| Integration override | `<integration_name>` | `Integration::handleInternalSpanServiceName` (Curl, Guzzle, PDO, Mysqli, SQLSRV, MongoDB, Memcache(d), Predis, PHPRedis, Eloquent, AMQP, …) |
| Config split-by-X | `opt.http_client_split_by_domain` | Curl, Guzzle, Psr18, Ratchet |
| | `opt.db_client_split_by_instance` | PDO, Mysqli, SQLSRV |
| | `opt.redis_client_split_by_host` | PHPRedis |
| Framework root-span | `<integration_name>` (only when DD_SERVICE not user-set) | new `Integration::tagFrameworkServiceSource` helper — Laravel, LaravelQueue, Drupal, Laminas, Lumen, Slim, SymfonyMessenger, Roadrunner, Ratchet (web), CakePHP |
| Manual API | `m` | serializer.c fallback — span's service differs from root's and no other layer tagged svc_src |
| Default (DD_SERVICE / auto) | *not set* | per RFC |

Child spans inherit `_dd.svc_src` from their parent at span-creation time (serializer.c).

## Tests

- 3 new `.phpt` tests covering the contract (`svc_src_default_cleared`, `svc_src_inheritance`, `svc_src_manual_override`)
- Updated affected integration-test expectations across the matrix


tracer/serializer.c
```diff
@@ -739,6 +739,16 @@ void ddtrace_inherit_span_properties(ddtrace_span_data *span, ddtrace_span_data
...
     zend_array *parent_meta = ddtrace_property_array(&parent->property_meta);
 
+    // Per RFC "Service Override Source Attribution", inherit _dd.svc_src
+    // from parent at span creation time.
+    zval *parent_svc_src = zend_hash_str_find(parent_meta, ZEND_STRL("_dd.svc_src"));
+    if (parent_svc_src) {
+        zend_array *child_meta = ddtrace_property_array(&span->property_meta);
+        zval copy;
+        ZVAL_COPY(&copy, parent_svc_src);
+        zend_hash_str_update(child_meta, ZEND_STRL("_dd.svc_src"), &copy);
```
bwoebi:
```suggestion
        Z_TRY_ADDREF_P(parent_svc_src);
        zend_hash_str_update(child_meta, ZEND_STRL("_dd.svc_src"), parent_svc_src);
```
No copying needed

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -158,6 +158,7 @@ static function (SpanData $span) use ($className, $methodName) {
                         $span->resource = $className . '@' . $methodName;
                         $span->type = Type::WEB_SERVLET;
                         $span->service = \ddtrace_config_app_name('laminas');
+                        Integration::tagFrameworkServiceSource($span, 'laminas');
                         $span->meta[Tag::COMPONENT] = 'laminas';
```
bwoebi: These three lines are actually the combination of operations being always the same for any framework integration.

Maybe the `Integration::tagFrameworkServiceSource` should actually be `Integration::setComponentMetadata($span, $component)` and take care of all three operations.

tracer/serializer.c
```diff
@@ -1766,6 +1776,13 @@ ddog_SpanBytes *ddtrace_serialize_span_to_rust_span(ddtrace_span_data *span, ddo
     }
     if (!is_inferred_span && !zend_string_equals_ci(Z_STR(prop_service_as_string), Z_STR(prop_root_service_as_string))) {
         ddog_add_str_span_meta_zstr(rust_span, "_dd.base_service", Z_STR_P(&prop_root_service_as_string));
+
+        // Per RFC "Service Override Source Attribution": when the service was
+        // overridden but no integration/config layer tagged _dd.svc_src, the
+        // override came from the manual API.
+        if (!zend_hash_str_exists(meta, ZEND_STRL("_dd.svc_src"))) {
+            ddog_add_str_span_meta_str(rust_span, "_dd.svc_src", "m");
```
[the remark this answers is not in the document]
bwoebi: This is the wrong place? You seem to unconditionally set, instead of only when it was manually overridden, by _other means than DD_SERVICE_.

This piece of code belongs into `ddtrace_span_data_readonly`.

tracer/functions.c
```diff
@@ -502,6 +502,17 @@ static zval *ddtrace_span_data_readonly(zend_object *object, zend_string *member
                 ZVAL_EMPTY_STRING(&span->property_version);
             }
         }
+
+        if (!is_top_level_root && span->root != NULL && Z_TYPE_P(value) == IS_STRING
+            && Z_TYPE(span->root->property_service) == IS_STRING
+            && !zend_string_equals(Z_STR_P(value), Z_STR(span->root->property_service))) {
+            zend_array *meta = ddtrace_property_array(&span->property_meta);
+            if (!zend_hash_str_exists(meta, ZEND_STRL("_dd.svc_src"))) {
```
bwoebi: So, depending on the place you set the service name, manually, a pre-existing integration svc_src won't be overwritten? Seems wrong to me.

bwoebi: Remove the exists check and order the assignment of _dd.svc_src _after_ asssigning ->service, to immediately overwrite it... Otherwise don't assign ->service at all.

bwoebi: We also don't _really_ need `ddtrace_config_app_name` anymore. This function essentially just returns it's input if DD_SERVICE is empty, which you're anyway checking with `if (!\dd_trace_env_config('DD_SERVICE')) {`.

bwoebi: I.e.
```
        if (!\dd_trace_env_config('DD_SERVICE')) {
            $span->service = $component;
            $span->meta['_dd.svc_src'] = $service ?? $component;
        }
```
should be enough (including dropping the exists check here).

bwoebi:
Actually, the failing tests raise a good point:
what about the _default_ service names `web.request` and `<entrypoint-name>` respectively? We don't really have an integration name for those, but we need to assign something to `_dd.svc_src` for these, to distinguish from DD_SERVICE, maybe `web` and `cli` for these "integrations"?

tracer/functions.c
```diff
@@ -502,6 +502,15 @@ static zval *ddtrace_span_data_readonly(zend_object *object, zend_string *member
                 ZVAL_EMPTY_STRING(&span->property_version);
             }
         }
+
+        if (!is_top_level_root && span->root != NULL && Z_TYPE_P(value) == IS_STRING
+            && Z_TYPE(span->root->property_service) == IS_STRING
+            && !zend_string_equals(Z_STR_P(value), Z_STR(span->root->property_service))) {
```
bwoebi: why this check? What is the root span service important here for?

tracer/hook/uhook_attributes.c
```diff
@@ -117,6 +117,10 @@ static void dd_fill_span_data(dd_uhook_def *def, ddtrace_span_data *span) {
         zval *service = &span->property_service;
         zval_ptr_dtor(service);
         ZVAL_STR_COPY(service, def->service);
+        zend_array *meta = ddtrace_property_array(&span->property_meta);
```
bwoebi: Needs a check for is_identical before overwriting.

bwoebi:
I assume _dd.svc_src also needs to be transferred to inferred spans?

bwoebi (approved):

PR #3960, opened 2026-06-05
Rollback unboxing from 1f7e90ca

bwoebi:
This became quite unergonomic with the downcast, but looks like there was a reason for the boxing.

Otherwise ShmLimiter references weird addresses.


morrisonlevi (approved):
Looks good to me as long as tests pass, pretty mechanical.

PR #3961, opened 2026-06-06
Exclude OTLP header and Datadog key configs from telemetry

bm1549:
### Description

Stops configuration telemetry from reporting `DD_API_KEY` or OTLP exporter header values.

`DD_API_KEY` now uses a sensitivity flag carried through ZAI config metadata. The OTLP header configs remain in supported-config metadata with `sensitive: true`, but are rejected at the PHP whitelist and native ingestion boundaries. `DD_TRACE_ENABLED` remains omitted separately because its shutdown-time value is not meaningful.

The metadata generator now reads sensitivity from all config headers and uses comment/string-aware parsing. Its CI self-test covers parser edge cases. Runtime coverage checks the native bypass path and a future OTLP traces header variant, with safe configs as positive controls.

### Reviewer checklist
- [x] Test coverage seems ok.


ext/telemetry.c
```diff
@@ -98,8 +98,7 @@ void datadog_telemetry_finalize() {
 #if ZTS
             ini = zend_hash_find_ptr(EG(ini_directives), ini->name);
 #endif
-            if (cfg->names[0].len != sizeof("DD_TRACE_ENABLED") - 1
-                || memcmp(cfg->names[0].ptr, "DD_TRACE_ENABLED", sizeof("DD_TRACE_ENABLED") - 1) != 0) { // DD_TRACE_ENABLED is meaningless: always off at rshutdown
+            if (!cfg->sensitive) {
```
bm1549: While DD_TRACE_ENABLED isn't sensitive, it can be handled in the same way

bwoebi (approved):
I've rewrote the code to be much simpler, I think we can now merge that.

PR #3964, opened 2026-06-08
fix(tracing): sandbox should save/restore jit_trace_num

morrisonlevi:
### Description

Fixes a PHP 8 tracing-JIT crash after sandboxed hook bailouts.

AppSec/RASP can trigger a `zend_bailout()` when it blocks a request during a hooked call. That bailout skips PHP’s normal `EG(jit_trace_num)` restore in `zend_call_function()`. ddtrace’s sandbox restored `current_execute_data`, but not `jit_trace_num`, leaving the JIT with a stale trace id.

A later JIT side exit could then read exit metadata from the wrong trace and crash.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
LGTM assuming CI passes.

PR #3965, opened 2026-06-09
fix(crasht): metadata correctness

morrisonlevi:
### Description

This fixes some small issues with the metadata:

1. If the JIT buffer is 0, then opcache was wrongfully marked as disabled. If it's 0, then _JIT_ is disabled, not opcache.
2. System INI doesn't mean what I thought it did at the time I wrote it, which is that values at startup are consistent. Unfortunately PHP can treat Apache INIs as system settings, for example. So we don't skip over most settings now. Just know that they might be overwritten.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi:
[ASAN Opcache tests: [8.5]](https://gitlab.ddbuild.io/datadog/apm-reliability/dd-trace-php/builds/1753825289) failure is on master, not related. I am looking into that.

realFlowControl (approved):
Thanks

bwoebi (approved):

PR #3970, opened 2026-06-10
feat: OTel context sharing

morrisonlevi:
### Description

This adds OTel context sharing using libdatadog. It adds it to both tracing (producer) and profiling (consumer).

This is mostly AI-written with a little human review as I went. It is not ready for merge, but it did pass some local end-to-end experimentation.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (changes requested):
I think the approach proposed by this PR is flawed / more expensive than needed.

The PR tries to synchronize state between otel context and root span.
The otel context is, at it's core, a pointer to some struct.

All we need is making the otel context part of the root_span_data struct.
Any updates to the `trace_id` / the roots span id -> update the local root span id / trace id part of the context, any updates to `active` -> update the span_id of the context.
And if there's no root span data, we just write NULL to the TLS variable.

I'm very not fond of "synchronizing" with otel context, any updates to the otel context should in my opinion simply happen in-place. There's also no reason for a "root span override".
It feels also way heavier than it should be. Updating a span id due to open/close/drop should be one 64 bit write. That's it.
Switching the span stack should cause one pointer write of the span stacks root span otel context to the TLS pointer. That's it, too.

Further, this PR at the very least forgot to handle ddtrace_update_root_id_properties().

bwoebi:
I also don't understand why this functionality depends on libdatadog. The code should be IMHO _simpler and less LoC_ if we just manage this manually.

tracer/otel_context.h
```diff
@@ -1,10 +1,45 @@
...
+#error "Unsupported byte order"
+#endif
+    memcpy(dest, &be_value, sizeof(be_value));
+}
+
+static inline void ddtrace_store_otel_thread_context_span_id(ddog_ThreadContextRecord *ctx, uint64_t span_id) {
+    uint8_t span_id_bytes[8];
+    ddtrace_write_otel_context_u64_be(span_id_bytes, span_id);
+
+    // OTEP 4947 only guarantees 2-byte record alignment, so span_id is bytes, not a uint64_t.
```
bwoebi: Uh? OTEP guarantees that _other processes_ have to guarantee 2-byte record alignment.
_We_ do align our data structure on 8 bytes and can trivially do an atomic store to `(uint64_t *)ctx->span_id`. (effectively, but you can make it explicit in span.h with an aligning attribute)

tracer/span.h
```diff
@@ -141,6 +145,9 @@ struct ddtrace_root_span_data {
     zval property_trace_id;
     zval property_git_metadata;
     zval property_inferred_span;
+#ifdef __linux__
+    ddog_ThreadContextRecord otel_context;
+#endif
```
bwoebi: You must place additional struct fields before ddtrace_span_data, containing `zend_object`. Otherwise you'll trivially crash with `class Foo extends DDTrace\SpanData { public $foo; }`

tracer/otel_context.c
```diff
@@ -0,0 +1,142 @@
...
+#endif
+
+ZEND_EXTERN_MODULE_GLOBALS(datadog);
+
+#ifdef __linux__
+_Static_assert(sizeof(ddog_ThreadContextRecord) == 640, "unexpected OTel thread context record size");
+_Static_assert(_Alignof(ddog_ThreadContextRecord) == 2, "unexpected OTel thread context record alignment");
+_Static_assert(offsetof(ddog_ThreadContextRecord, span_id) == 16, "unexpected OTel span id offset");
+_Static_assert(offsetof(ddog_ThreadContextRecord, attrs_data_size) % _Alignof(uint16_t) == 0, "unaligned OTel attrs data size");
+
```
bwoebi, line 20:
```suggestion
_Static_assert(offsetof(ddtrace_root_span_data, ddtrace_root_span_data) % 8 == 0, "unexpected root_span_data alignment");

```
And now go ahead with atomic span updates

tracer/otel_context.c
```diff
@@ -0,0 +1,236 @@
...
+void ddtrace_otel_update_trace_id(ddtrace_root_span_data *root) {
+    if (!root) {
+        return;
+    }
+
+    datadog_otel_thr_ctx_rec *record = &root->otel_context;
+    if (!record->attrs_data_size) {
+        return;
+    }
```
bwoebi: Some of the functions guard against this. Why?

cataphract: ddtrace_otel_attach_stack is what initializes the record. The idea is that it's pointless to change a record that was never attached as the thread local context (and therefore is not initialized)

bwoebi: That's certainly pointless, but nearly every span is going to be attached, right?
It doesn't do harm and the check is just extra overhead.

cataphract: I've partially initialized it, though not the span id or the atributes as those are more appropriate to initialize upon attachment.

cataphract:
The failure in dd-gitlab/check-big-regressions is a direct consequence of this change (7 root spans x 1000 revs x 640 bytes = 4 480 000 bytes = 4.49 MB). This is by design

profiling/src/php_ffi.c
```diff
@@ -7,14 +7,154 @@
...
+
+    zend_module_entry *module;
+    ZEND_HASH_FOREACH_PTR(&module_registry, module) {
+        if (module && module->handle) {
+            tls_symbol = DL_FETCH_SYMBOL(module->handle, DATADOG_PHP_PROFILING_OTEL_TLS_SYMBOL);
+            if (tls_symbol) {
+                return tls_symbol;
+            }
+        }
+    } ZEND_HASH_FOREACH_END();
```
bwoebi: Is that necessary? In what case would the symbol _not_ be availble via dlsym(RTLD_DEFAULT), but available via direct DL_FETCH_SYMBOL?

profiling/src/php_ffi.c
```diff
@@ -7,14 +7,154 @@
...
+    /*
+     * The writer is stopped while this signal-handler-style reader runs on the same thread, so
+     * valid only has to guard against compiler reordering. The writer fences before/after field
+     * updates; this reader fences after observing valid=1 so field reads stay after the guard.
+     */
+    if (!record || record->valid != 1) {
+        return false;
+    }
+
+    atomic_signal_fence(memory_order_acquire);
```
bwoebi: This is not a "signal-handler-style reader". It runs synchronously with the code in vm_interrupt and similar safe places. Doesn't need a fence either, just a simple validity check.

tracer/ddtrace.c
```diff
@@ -148,20 +150,32 @@ static inline void dd_alter_prop(size_t prop_offset, zval *old_value, zval *new_
 
 bool datadog_alter_dd_service(zval *old_value, zval *new_value, zend_string *new_str) {
     dd_alter_prop(XtOffsetOf(ddtrace_span_properties, property_service), old_value, new_value, new_str);
+    ddtrace_span_stack *stack = DDTRACE_G(active_stack);
+    if (stack && stack->root_span) {
+        ddtrace_otel_update_attribute_values(stack->root_span);
+    }
```
bwoebi, line 153: extract that snippet into dd_alter_prop, it's only used for service/env/version anyway.

ext/datadog.c
```diff
@@ -152,11 +155,37 @@ static void datadog_shutdown(zend_extension *extension) {
...
+    if (!get_DD_TRACE_REPORT_HOSTNAME()) {
+        return DDOG_CHARSLICE_C("");
+    }
+
+    if (ZSTR_LEN(get_DD_HOSTNAME())) {
+        return dd_zend_string_to_CharSlice(get_DD_HOSTNAME());
+    }
+
+    // Match tracer/serializer.c hostname publishing: DD_HOSTNAME wins, then gethostname().
+#ifdef __linux__
```
bwoebi: Did you mean to put this `#ifdef` to the top of the function? There's no point in half the function.

bwoebi:
A call to datadog_publish_otel_process_context() is missing in `datadog_internal_handle_fork()`.

compile_rust.sh
```diff
@@ -13,6 +13,33 @@ case "${host_os}" in
     ;;
 esac
 
+# GCC < 9 doesn't support -fuse-ld=lld (emitted by libdd-otel-thread-ctx-ffi/build.rs).
+# Intercept CC calls and replace -fuse-ld=lld with -B<dir> where <dir>/ld -> ld.lld.
```
bwoebi: Do we still need this, or is this obsolete now that we define our own struct?

bwoebi: I think the *only* thing we import from there is the `otel_thread_ctx_v1` symbol itself. Is there a reason we can't simply put it into our `.c` file? We don't need that asm-way to declare it, but can simply declare the __thread symbol directly, right? That just was because it was scoped from rust.

profiling/src/php_ffi.c
cataphract: I don't think it's needed. PHP always loads with RTLD_GLOBAL and our only call without it in `loader/dd_library_loader.c` is only used to increment the refcount on glibc

profiling/src/php_ffi.c
cataphract: The thing is the profiler is not the only consumer of the otel thread local info.

The agent will also read it on kprobes/uprobes, for instance.

bwoebi: ... yes? This code here is just relevant to the profiler. The _agent_, i.e. an external reader hooking at an arbitrary point in time obviously needs that. But the profiler does not.

profiling/src/php_ffi.c
bwoebi: Yep, exactly.

profiling/src/php_ffi.c
cataphract: you're right, i didn't pay attention to the context

profiling/src/php_ffi.c
cataphract: left only `dlsym(RTLD_DEFAULT, DATADOG_PHP_PROFILING_OTEL_TLS_SYMBOL);`

profiling/src/php_ffi.c
cataphract: removed comment and compiler fence

ext/datadog.c
cataphract: the idea was to call gethostname only on linux, but i've now matched what's tracer/serializer.c, let's hope it doesn't blow up on windows where the function is apparently slightly different and depends on a separate header

compile_rust.sh
cataphract: I removed the dependency on libdd-otel-thread-ctx-ffi / libdd-otel-thread-ctx.

cataphract:
Dep on libdatadog: https://github.com/DataDog/libdatadog/pull/2176

ext/datadog.c
```diff
@@ -733,7 +739,11 @@ static PHP_MINFO_FUNCTION(datadog) {
 
 void datadog_internal_handle_fork(void) {
     // CHILD PROCESS
-    datadog_sidecar_handle_fork();
+    bool runtime_id_changed = false;
+    datadog_sidecar_handle_fork(&runtime_id_changed);
+    if (runtime_id_changed) {
+        datadog_publish_configured_otel_process_context();
+    }
```
bwoebi: You need to do it unconditionally, the libdatadog code uses `libc::MFD_CLOEXEC`.

ext/sidecar.c
```diff
@@ -435,15 +435,18 @@ void datadog_sidecar_minit(void) {
...
+
     ddog_RemoteConfigFlags flags = {0};
     bool enable_sidecar = datadog_sidecar_should_enable(&flags);
 
     if (!enable_sidecar) {
         return;
     }
 
+    *runtime_id_changed = datadog_sidecar_instance_id != NULL;
     datadog_force_new_instance_id();
```
bwoebi, line 472: Feel free to move this to the start of the function so that it happens unconditionally.

ext/datadog.c
bwoebi: If you want to make it dependent on datadog_sidecar_instance_id (not sure why you would), please check this in datadog_publish_configured_otel_process_context, not here.

yannham:
FYI: there's a bug in the libdatadog implementation of publishing the threadlocal attributes to the process context, where thread-context-related attributes are stored in the wrong place. https://github.com/DataDog/libdatadog/pull/2167 fixes it and is being merged, but until you can depend on a version that includes it, an external context reader might not be able to properly resolve the thread context records.

ivoanjo (commented):
Since this is dependent on https://github.com/DataDog/libdatadog/pull/2176 I suggest we ship that one first.

Having said that, I think it would make sense to split this PR into a few different bits? 3k lines of C + groovy + python + PHP + rust +  extras is going to be really hard to review and land IMO?

cataphract:
Signatures in Levi's commits were being shown as unverified, so I resigned also commits my key and put myself as author of those commits.

ivoanjo:
I guess https://github.com/DataDog/dd-trace-php/pull/4077 is a replacement for this PR. Should we maybe close this PR?

PR #3976, opened 2026-06-11
Make the SpanStack happier with cloning operations

bwoebi:
More stuff to make sure we don't leave invalid state behind with random operations (like if a span/stack is stored in some property of a class being deeply cloned or whatever).


morrisonlevi (approved):
Okay this looks like it might _actually_ be the crash customers are hitting, rather than just "random" technical issues that the AI has been finding!

PR #3984, opened 2026-06-14
feat(openfeature): emit server-side EVP flagevaluation

leoromanovsky:
## Motivation

Customers need PHP services to report server-side feature-flag evaluation counts through the same backend contract as the other SDKs, including both the PHP 7 extension path and PHP 8 OpenFeature path. This contribution adds native PHP EVP `flagevaluation` delivery through the shared libdatadog sidecar path while preserving existing OTel metric and exposure behavior, giving APM a backend-verifiable rollout signal for approval.

## Justification

The 5 MiB limit is unlikely for small applications, but it is reachable at the scale this rollout is designed to support. Using compact JSON estimates against a 5,242,880-byte body limit: a minimal degraded row is about 137 bytes, so about 37,991 rows fit; a small full row is about 277 bytes, so about 18,859 rows fit; a normal full row with 10 context attributes is about 588 bytes, so about 8,901 rows fit; a max bounded-context row with 256 256-character fields is about 68,341 bytes, so only about 76 rows fit.

The target scale is 2,500 flags x 50 full-fidelity buckets, or 125,000 rows. Even small full rows can exceed 35 MiB at that scale, so byte splitting is a real compliance requirement rather than only defensive hardening. Existing backpressure bounds the queue and aggregate cardinality, but it does not bound the final encoded POST body; async posting avoids blocking customer evaluations but a single oversized request can still be rejected with 413.

## Changes

- Adds native PHP EVP flagevaluation aggregation using backend-visible dimensions: flag key, variant key, allocation key, runtime-default state, error message, real targeting-rule key when present, targeting key, and bounded context.
- Prunes evaluation context before it participates in aggregation keys or queued snapshots.
- Keeps OpenFeature `reason` out of EVP payloads and aggregation keys.
- Preserves visible variant/allocation/error dimensions when overflow folds into degraded events.
- Flushes the aggregated flagevaluation batch before exposure and OTel metric sidecar actions at request shutdown.
- Splits native-to-sidecar flagevaluation batches into bounded 512-event IPC chunks.
- Bumps `libdatadog` to the companion flagevaluation sidecar delivery head.
- Consumes the libdatadog sidecar follow-up that splits final EVP POST bodies by encoded uncompressed JSON bytes under the 5 MiB limit, degrades oversized full rows, and drops/logs only if a degraded row still cannot fit.

## Decisions

- EVP cardinality is defined only by fields the worker receives; `reason` is not a hidden aggregate key.
- Degraded events omit targeting key and context but keep the visible dimensions needed for backend counts.
- PHP keeps the shared sidecar delivery architecture instead of adding a direct EVP HTTP writer to the extension.
- System-test validation covers the production per-flag degradation threshold and exercises overflow past the full-fidelity cap.
- Payload splitting happens in the libdatadog sidecar flagevaluation flusher where the final uncompressed EVP POST body exists; PHP keeps count-based IPC chunking as a secondary guard, not the EVP byte-limit mechanism.
- This PR remains draft until the cross-SDK rollout and review pass are complete.

```mermaid
flowchart TD
  A[PHP extension flushes aggregated rows] --> B[send bounded IPC chunks to sidecar]
  B --> C[sidecar coalesces and serializes candidate JSON]
  C --> D{POST body <= 5 MiB?}
  D -- yes --> E[post asynchronously through Agent EVP proxy]
  D -- no --> F{single full row can degrade?}
  F -- yes --> G[omit targeting_key and context]
  G --> C
  F -- no --> H[drop, log, count]
```

### Payload Limit Follow-Up

- PHP commit: `37176c9b4`
- `libdatadog` submodule pointer now references `46734bc1e`, which contains the sidecar EVP byte-splitting/degrade/drop fix.
- No PHP source changed in this follow-up; the behavior change is consumed through the libdatadog submodule pointer.
- Companion libdatadog validation in DataDog/libdatadog#2117: `cargo nextest run -p datadog-sidecar ffe_flagevaluation_flusher`, `cargo check -p datadog-sidecar`, `cargo +nightly-2026-02-08 fmt --all -- --check`, and `cargo +stable clippy -p datadog-sidecar --all-targets --all-features -- -D warnings` passed.

### Dogfooding App

- `ffe-dogfooding` `app-php7` and `app-php8-openfeature` were run with local PHP artifacts and the companion libdatadog sidecar changes.
- Evaluated `ffe-dogfooding-string-flag` through PHP 7 with public-safe targeting keys:
  - `php7-evp-agent-20260622T2252-alpha`
  - `php7-evp-agent-20260622T2252-bravo`
  - `php7-evp-agent-20260622T2252-charlie`
- Evaluated `ffe-dogfooding-string-flag` through PHP 8 OpenFeature with public-safe targeting keys:
  - `php8of-evp-agent-20260622T2255-alpha`
  - `php8of-evp-agent-20260622T2255-bravo`
  - `php8of-evp-agent-20260622T2255-charlie`
- App-side result: all six evaluations returned `variant_2`.

### System Tests

- Companion draft PR: https://github.com/DataDog/system-tests/pull/7187

### Staging End-To-End

- Dogfooding ran without the local mock-intake EVP tee/proxy, so the Agent sent EVP traffic through the normal backend path.
- Retriever staging query returned 6 `flagevaluation` rows for the exact PHP 7 and PHP 8 OpenFeature targeting keys above.
- Each row had `flag.key=ffe-dogfooding-string-flag`, `variant.key=variant_2`, `allocation.key=allocation-override-392dd7c149f8`, and `evaluation_count=1`.


src/api/FeatureFlags/Internal/Metric/EvaluationMetricRecorder.php
```diff
@@ -59,6 +59,23 @@ public function record(EvaluationMetric $metric)
...
     }
 
+    /**
+     * Returns true when the EVP flagevaluation emission killswitch
+     * DD_FLAGGING_EVALUATION_COUNTS_ENABLED is not explicitly disabled.
+     * Default: enabled.
+     *
+     * @internal Killswitch only — does not control the OTel metric path.
+     */
+    public static function isEvpEnabled()
```
bwoebi, line 69: Is this intentionally not called anywhere?

components-rs/ffe.rs
```diff
@@ -90,6 +103,546 @@ const TYPE_FLOAT: i32 = 2;
...
+            flag_evaluations: chunk.to_vec(),
+        })
+        .collect()
+}
+
+/// Flush aggregated EVP flag evaluation events to the sidecar.
+///
+/// # Safety
+/// All pointer parameters must be valid non-null pointers to live objects.
+/// `service`, `env`, `version` must be valid UTF-8 `CharSlice` values.
```
bwoebi, line 606:
```suggestion
```
No, you convert it lossily, no such guarantee needed (and should not either)

components-rs/ffe.rs
```diff
@@ -90,6 +103,546 @@ const TYPE_FLOAT: i32 = 2;
...
+    queue_id: &QueueId,
+    counters: AggregatorTelemetryCounters,
+) {
+    let actions = flag_evaluation_telemetry_actions(counters);
+    if actions.is_empty() {
+        return;
+    }
+
+    register_flag_evaluation_count_metric(transport, FLAG_EVALUATION_ROWS_DROPPED_METRIC);
+    register_flag_evaluation_count_metric(transport, FLAG_EVALUATION_ROWS_DEGRADED_METRIC);
```
bwoebi, line 583: You're re-registering this over and over. Registering this once per thread is sufficient.

PR #3988, opened 2026-06-16
fix(profiling): `free()` always untracking

realFlowControl:
### Description

Since https://github.com/DataDog/dd-trace-php/pull/3623 got merged, our overhead rose which is due to an oversight: even when heap live profiling is disabled, the `free()` path would always got the untrack allocation route:
image

This PR makes sure to inject the NO-OP `free()` implementation in `zend_mm_set_custom_handlers_ex()`/`zend_mm_set_custom_handlers()` when heap live is disabled and only the untracking `free()` version when heap live is enabled.

### Reviewer checklist
- [ ] Test coverage seems ok.

PROF-13688

### DIFF
--- a/profiling/src/allocation/allocation_ge84.rs
+++ b/profiling/src/allocation/allocation_ge84.rs
@@ -108,7 +108,7 @@ pub fn first_rinit_should_disable_due_to_jit() -> bool {
         && *JIT_ENABLED
 }
 
-pub fn alloc_prof_rinit() {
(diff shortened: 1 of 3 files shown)


morrisonlevi (approved):
Check if the same issue can happen for realloc's free path, but otherwise, approved.

realFlowControl:
It can and I enabled auto-merge on this PR ...
I'll open a follow up PR

PR #3990, opened 2026-06-16
refactor(profiling): remove lazy_static, optimize Sapi/RefCellExt

realFlowControl:
## Summary

[Ponytail](https://github.com/DietrichGebert/ponytail) audit cleanup for the PHP profiler.

## Changes

1. Replace `lazy_static` with `std::sync::LazyLock` (MSRV 1.87+)
2. Add `#[inline]` to `RefCellExt` trait methods:
   I am not 100% sure about this, it is definitely not wrong, but I'd guess the compiler would inline this magically anyway?
4. Simplify `Sapi::from_name()`

Powered by Kimi K2.5 and pi

You can review commit by commit (each step above is a commit)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1369,7 +1369,6 @@ dependencies = [
  "dynasmrt",
  "env_logger 0.11.6",
  "http",
- "lazy_static",
  "libc 0.2.177",
  "libdd-alloc",
  "libdd-common",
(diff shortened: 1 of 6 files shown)


morrisonlevi (approved):
Approved but note that the inline attributes on generic methods are pointless: all (or maybe nearly all?) generic methods are inlined as part of how they work.

PR #3993, opened 2026-06-16
fix(profiling): skip realloc heap-live untracking when disabled

realFlowControl:
### Description

Follow up PR to #3988

This PR removes the remaining heap-live overhead from allocation profiling when heap-live is disabled.

`free()` already selects a no-untrack handler for heap-live-disabled requests. `realloc()` now does the same: request init installs a tracking realloc handler only when heap-live is enabled, otherwise it installs a no-untrack handler that keeps allocation sampling but skips `untrack_allocation(prev_ptr)`.

The tracking and no-untrack realloc paths are separate implementations, so the hot path does not pay a heap-live flag branch.

PROF-13688


morrisonlevi (approved):
Looks good, thanks for following up!

PR #3996, opened 2026-06-16
feat(FeatureFlags): FFE APM feature-flag span enrichment (experimental, gated)

leoromanovsky:
# feat(FeatureFlags): FFE APM feature-flag span enrichment

> ⚠️ Experimental, opt-in, gated behind `DD_EXPERIMENTAL_FLAGGING_PROVIDER_SPAN_ENRICHMENT_ENABLED` (off by default).

## Summary

Adds Feature Flag Events (FFE) span enrichment to the feature-flag integration. When feature
flags are evaluated, the evaluation metadata is attached to the root APM span so APM customers
can filter traces and errors by active flag variant, and the FFE/Experimentation platform can
correlate spans with experiments. The wire format matches the merged reference implementation
(`dd-trace-js#8343`) so backend/Trino decode is identical.

## How it works

1. A flag is evaluated (via the OpenFeature `DataDogProvider` or the native `DDTrace\FeatureFlags\Client`).
2. Each evaluation is accumulated inline into a single **shared, request-scoped registry** bound to the current root span.
3. At root-span close, the accumulated state is encoded and written as `ffe_*` tags on the root's `meta`.

## Configuration

Opt-in, off by default:

```bash
DD_EXPERIMENTAL_FLAGGING_PROVIDER_SPAN_ENRICHMENT_ENABLED=true
```

This is distinct from `DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED`.

## Span tags added

| Tag | Description | Format |
|-----|-------------|--------|
| `ffe_flags_enc` | All evaluated flag serial IDs | base64 delta-varint |
| `ffe_subjects_enc` | Subject → flags mapping (when `doLog=true`) | JSON `{ sha256(key): encodedIds }` |
| `ffe_runtime_defaults` | Fallback values for flags not in UFC | JSON `{ flagKey: value }` |

Limits: 200 serial IDs, 10 subjects, 20 experiments/subject, 5 runtime defaults, 64 chars/runtime-default value (UTF-8-safe truncation).

## Changes

- **Gate + config**: add `DD_EXPERIMENTAL_FLAGGING_PROVIDER_SPAN_ENRICHMENT_ENABLED` (`ext/configuration.h`), off by default; thread the split `serial_id` Rust → C → PHP mapper (`components-rs/ffe.rs`, `components-rs/common.h`, `tracer/functions.c`, `ResultMapper.php`). Because the C ABI cannot represent `Option<i32>`, presence is carried separately via `has_serial_id` — consumers must gate on it and never treat `serial_id == 0` as absent.
- **Codec + accumulator** (`SpanEnrichmentAccumulator.php`): delta-varint serial IDs + SHA256-hashed subject keys, frozen limits, and Node-parity value stringification (floats, `JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES`, UTF-8-safe truncation).
- **Shared registry** (`SpanEnrichmentRegistry.php`): a single request-scoped accumulator that ALL PHP evaluation paths route through, so evaluations under one root span **aggregate** rather than **overwrite** (the native staging API replaces the request-global tag slots wholesale). Centralizes the root-span lifecycle: at most one root-close reset per root, and a reset on any root-span boundary transition / dropped root / request shutdown so nothing leaks across spans or requests.
- **`SpanEnrichmentBinder`** (`SpanEnrichmentBinder.php`): a thin, gate-gated adapter that binds both the OpenFeature `DataDogProvider` and the native `DDTrace\FeatureFlags\Client` to the shared registry, so non-OpenFeature consumers are enriched too.
- **Native staging + flush** (`tracer/ffe.c`/`.h`, `tracer/ddtrace_globals.h`, `tracer/span.c`, `tracer/ddtrace.c`): request-local `zend_string` slots staged from PHP via a new internal `DDTrace\Internal\set_ffe_span_enrichment_tags()`; flushed onto the root's `meta` at root-span close **before the sampling decision** (so span-level sampling rules matching `ffe_*` can see them), and cleared on dropped root / request shutdown.
- **Non-creating root accessor** (`DDTrace\Internal\peek_root_span_id()`, `tracer/functions.c`): returns the active root's own `span_id` **without** calling `dd_ensure_root_span()`, so detecting a root-span boundary while merely evaluating a flag never creates an autoroot span. Uses `span_id` (never reused in-process) rather than a recyclable object handle to avoid identity aliasing.
- **Test build helper** (`config.m4`, `tracer/functions.c`): standard CI/test/package builds now define `-DDD_TEST_HELPERS=1`, which compiles in a test-only `await_ffe_config` `dd_trace_internal_fn` helper that actively pumps Remote Config (can block up to 5s) so long-running CLI test servers (system-tests parametric app / ffe-dogfooding) can deterministically wait for the pushed UFC. A hardened production build can drop the flag to compile these surfaces out entirely.
- **Tests**: accumulator + registry unit tests; `.phpt` ext tests for the native bridge, serial-id passthrough, and the non-creating `peek_root_span_id`.

## Decisions

- **Inline accumulation (not a `finally` hook)**: PHP OpenFeature does not pass `ResolutionDetails` to `finally` hooks, so enrichment is accumulated inline on the evaluation path.
- **Single shared registry (aggregate, not overwrite)**: the native staging API replaces the request-global tag slots wholesale, so per-provider/per-client accumulators would clobber each other under one root span. All paths route through one shared, request-scoped accumulator so the staged tag set is the union of all evaluations on the active root.
- **No idle per-span overhead when the gate is off** — no binder is constructed, nothing is accumulated/staged, and the close-span flush is a cheap early-return.
- **Lifecycle**: accumulator reset on the root-span boundary, on root close, on dropped root, and at request shutdown.
- `ffe_*` are bare tag names on span `meta` (not `_dd.`-prefixed); subject keys are SHA256 hashes emitted only when logging is authorized.

## Known limitations

- **Single active root at a time.** The registry tracks exactly one active root via a flat `rootId`/accumulator pair, and the native staging bridge mirrors that with one set of request-global slots. When two roots are genuinely open concurrently (fibers, multiple span stacks) and evaluations interleave, switching the active root discards the other root's accumulated data and flushing can attach staged tags to the wrong root. A proper fix needs per-root-keyed storage on **both** sides (PHP accumulators keyed by root id AND native slots keyed the same way) — a native wire-contract-adjacent change, deferred as a follow-up given the feature is experimental and gated off by default.

### FFE dogfooding app

Validated live against the `ffe-dogfooding` app via a `trace-intake` tee-proxy that captures the raw `/v0.4/traces` payload and decodes the `ffe_*` tags. Flag `ffe-dogfooding-string-flag` (serial `2312`):

- **Gate ON** — the root span (`web.request`, auto-instrumented web SAPI) carried `ffe_flags_enc` decoding to serial `[2312]` plus a SHA256-hashed `ffe_subjects_enc` → `[2312]`.
- **Gate OFF** — span flushed with zero `ffe_*` tags.

Aggregation, unicode/object runtime defaults, and codec parity (`ZAgUAg==` → `[100,108,128,130]`) were additionally exercised end-to-end through the real OpenFeature provider path and confirmed against `scripts/decode_ffe_span_tags.py`.

### Local system-tests run

Ran the frozen `system-tests` parametric suite (`tests/parametric/test_ffe/test_span_enrichment.py`, unchanged) against this branch's tracer (PHP 8.2 NTS, C extension built from source for `aarch64-linux-gnu`):

```
TEST_LIBRARY=php ./run.sh PARAMETRIC -k span_enrichment
Library: php@1.21.0
============================= 18 passed in 49.99s ==============================
```

All 18 cases pass — `ffe_flags_enc` aggregates serial IDs across evaluations and propagates from child spans to the root; `ffe_subjects_enc` carries SHA256-hashed targeting keys gated on `doLog`; `ffe_runtime_defaults` is added for not-found flags with 64-char truncation; and all frozen limits are enforced. The `SpanEnrichmentBinder` change was required so the native `DDTrace\FeatureFlags\Client` path (used by the parametric server) is enriched. The system-tests enablement (parametric `server.php` + `manifests/php.yml`) is a separate draft PR against `DataDog/system-tests`.


tracer/span.c
```diff
@@ -976,6 +977,13 @@ void ddtrace_close_top_span_without_stack_swap(ddtrace_span_data *span) {
     if (span->std.ce == ddtrace_ce_root_span_data) {
         ddtrace_root_span_data *root = ROOTSPANDATA(&span->std);
         LOG(SPAN_TRACE, "Closing root span: trace_id=%s, span_id=%" PRIu64, Z_STRVAL(root->property_trace_id), span->span_id);
+
+        // APM feature-flag span enrichment (PHP-01): when the root span closes,
+        // flush any staged ffe_* tags into its meta. This is gated and a no-op
+        // (cheap early-return) when the feature is off or nothing was staged
+        // (DG-005). The accumulator is request-scoped and cleared here so it
+        // never leaks across requests.
+        ddtrace_ffe_flush_span_enrichment_tags(ddtrace_property_array(&root->property_meta));
```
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 265422ffd — `ddtrace_drop_span()` now calls `ddtrace_ffe_clear_span_enrichment_tags()` on the root-drop branch so staged tags can't leak onto the next root's close. Added a regression test (`tests/ext/ffe/root_drop_clears_staged_tags.phpt`) simulating a genuine drop via `try_drop_span`.

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
```diff
@@ -0,0 +1,307 @@
...
+                return $value;
+            }
+            return mb_substr($value, 0, $maxLength, 'UTF-8');
+        }
+
+        if (strlen($value) <= $maxLength) {
+            return $value;
+        }
+
+        $truncated = substr($value, 0, $maxLength);
```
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 265422ffd — extracted the fallback into `truncateUtf8ByteFallback()`, which now walks codepoint-by-codepoint via UTF-8 leading-byte length instead of cutting at raw bytes. Added `testTruncateUtf8ByteFallbackCountsCharactersNotBytes` covering ASCII/CJK/emoji cases.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        $root = \DDTrace\root_span();
+        if ($root === null) {
+            return;
+        }
+
+        $root->onClose[] = static function () use ($reset) {
+            $reset();
+        };
```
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 265422ffd — the reset closure is now prepended (`prependOnCloseReset()`), not appended, to `$root->onClose`, so it always runs last under tracer/span.c's reverse-iteration order. Added `testPrependOnCloseResetRunsLastUnderReverseIteration` to lock in the ordering.

pavlokhrebto:
@codex review

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,331 @@
...
+     * @param string|null $targetingKey
+     */
+    public function accumulate($flagKey, $details, $targetingKey)
+    {
+        try {
+            $this->resetForRootBoundary();
+
+            if ($this->accumulator === null) {
+                $this->accumulator = new SpanEnrichmentAccumulator();
+            }
```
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 92f1b6ffe: `accumulate()` now returns early right after `resetForRootBoundary()` when `$this->rootId` is null, so an evaluation with no active root is never allocated/staged. Added `testAccumulateWithoutAnActiveRootIsSkippedRatherThanStaged` in SpanEnrichmentRegistryTest.php.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,331 @@
...
+     * only the active root span's evaluations and never leaks across spans /
+     * requests. Fires on ANY transition (new root, or losing the active root) so
+     * a dropped/abandoned root -- which never runs its $onClose handler -- cannot
+     * leak into the next root or request.
+     */
+    private function resetForRootBoundary()
+    {
+        $rootId = $this->currentRootSpanId();
+        if ($rootId === $this->rootId) {
+            return;
```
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 92f1b6ffe: switched the identity source in `peek_root_span_id()` (tracer/functions.c) from the zend object handle to the root span's own `span_id` (random 64-bit, never reused within the process), so a recycled handle from a dropped root can no longer alias a later root. Updated `currentRootSpanId()`'s docblock and `tests/ext/ffe/peek_root_span_id_non_creating.phpt` to assert stability/uniqueness instead of the old (now-incorrect) spl_object_id equivalence.

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
```diff
@@ -0,0 +1,332 @@
...
+            // length so the 64-char truncation can cut mid-escape-sequence.
+            $encoded = json_encode($value, JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
+            $valueStr = $encoded === false ? '' : $encoded;
+        } elseif (is_bool($value)) {
+            // Match the Node String(boolean) form: "true"/"false".
+            $valueStr = $value ? 'true' : 'false';
+        } elseif ($value === null) {
+            $valueStr = 'null';
+        } else {
+            $valueStr = (string) $value;
```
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 92f1b6ffe: added `stringifyFloat()` in SpanEnrichmentAccumulator.php to match Node's `String(number)` — lowercase `e` exponents, `-0` renders as `"0"`, no trailing `.0` before an exponent, and `NaN`/`Infinity`/`-Infinity` for those special values. Added corresponding cases to `nodeStringValueRenderings()` in SpanEnrichmentAccumulatorTest.php.

tracer/span.c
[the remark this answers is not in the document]
pavlokhrebto: Fixed in 92f1b6ffe: moved the FFE tag flush in tracer/span.c so it runs for root spans before `ddtrace_decide_on_closed_span_sampling()`, so sampling rules that match on `ffe_*` tags can now see them at decision time.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        if ($this->accumulator !== null) {
+            $this->accumulator->clear();
+        }
+        $this->resetStaging();
+        $this->rootId = $rootId;
```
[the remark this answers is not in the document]
pavlokhrebto: Acknowledged as a known limitation rather than fixed here — this registry (and the native staging bridge in tracer/ffe.c) track exactly one active root at a time by design. Fixing concurrent roots properly needs per-root-keyed storage on both the PHP accumulator side and the native request-global slots, which is a native change beyond a PHP-only patch. Documented this explicitly in the class docblock and in `resetForRootBoundary()`'s docblock (commit 92f1b6ffe) so it's tracked as a deliberate deferral rather than an oversight, given the feature is experimental and gated off by default. Happy to file a follow-up issue if useful.

components-rs/common.h
```diff
@@ -484,6 +484,8 @@ typedef struct ddog_FfeResult {
   _zend_string * allocation_key;
   int32_t reason;
   int32_t error_code;
+  int64_t serial_id;
```
typotter: 👍

components-rs/ffe.rs
```diff
@@ -97,6 +97,14 @@ pub struct FfeResult {
     pub allocation_key: MaybeOwnedZendString,
     pub reason: i32,
     pub error_code: i32,
+    // serial_id is the selected split's serial id, carried for APM span
```
typotter, line 100: had no idea the i32 would have such knock-on complications. :(

components-rs/ffe.rs
```diff
@@ -97,6 +97,14 @@ pub struct FfeResult {
...
+    // enrichment (ffe_flags_enc). The source field is Option<i32>; since the
+    // C ABI cannot represent Option<i32> as a plain field, we surface the
+    // presence separately via has_serial_id. Consumers MUST gate on
+    // has_serial_id (the Pattern B "missing variant => default" semantic) and
+    // never treat serial_id == 0 as "absent".
```
typotter, line 105: 👍

src/api/FeatureFlags/Internal/ResultMapper.php
```diff
@@ -274,6 +274,15 @@ private function exposureData($rawResult)
             $exposureData['doLog'] = (bool) $this->read($rawResult, array('do_log', 'doLog'), false);
         }
 
+        // serialId is the selected split's serial id, surfaced from the native
+        // bridge for APM span enrichment. It is only present when the native
+        // result actually carried one; a null/absent value must be left out so
+        // downstream consumers can treat "no serialId" as a runtime default.
```
typotter: misleading comment: "no serialId" is a transitional state for any variant assigned by the waterfall and doesn't, in and of its own absence, indicate the runtime default was used.

src/api/FeatureFlags/Client.php
```diff
@@ -20,6 +22,19 @@ public function __construct($logger = null)
 
         $this->evaluator = NativeEvaluator::create();
         $this->logger = $logger ?: new TriggerErrorLogger();
+        // DG-004/DG-005: the native Client does NOT go through the OpenFeature
```
typotter: still plan to fix?

src/api/FeatureFlags/Internal/ResultMapper.php
pavlokhrebto: Good catch, fixed -- reworded to not conflate absence of `serialId` with "runtime default was used." Absence alone just means the native result carried no serial id (a waterfall-assigned variant can also have none); `SpanEnrichmentRegistry::accumulate()` is what actually decides "runtime default," and it requires both no serialId AND no variant.

src/api/FeatureFlags/Client.php
pavlokhrebto: Yes -- fixed. These require_once calls were redundant: SpanEnrichmentBinder/Accumulator/Registry live in the same PSR-4 autoloaded namespace as Client itself (src/api/FeatureFlags/, DDTrace\\ -> ./src/api/), so whatever loaded Client.php already makes them autoload-resolvable on first reference -- no explicit require needed, and it means gate-off no longer eagerly loads SpanEnrichmentBinder.php either.

components-rs/ffe.rs
```diff
@@ -97,6 +97,14 @@ pub struct FfeResult {
...
+    pub serial_id: i64,
+    pub has_serial_id: bool,
```
bwoebi: I'm confused, why is serial_id i64? Yes, we need an extra bool flag to represent the option, but serial_id can still be i32?

tracer/ddtrace.stub.php
```diff
@@ -1092,6 +1093,29 @@ function record_ffe_evaluation_metric(string $flagKey, ?string $variant, ?string
...
+     * only calls it when the gate is on (DG-004 inline accumulation / DG-005).
+     *
+     * @internal
+     */
+    function set_ffe_span_enrichment_tags(?string $flagsEnc, ?string $subjectsEnc, ?string $runtimeDefaults): void {}
+
+    /**
+     * Identity (the root span's own span_id) of the active root span, or null
+     * when there is none. Unlike \DDTrace\root_span(), this NEVER creates a root span as a
+     * side effect (it does not call dd_ensure_root_span()). Used by APM
```
bwoebi: This should not be relevant. Outside of some testing scenarios, the root span should not be removable in the first place. Just use `\DDTrace\root_span()->id`. And for the tests, there exists `TracerTestTrait::isolateTracer()` to avoid tests conflicting with each other.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+     *
+     * KNOWN LIMITATION: this is where the single-active-root assumption bites.
+     * If root A is still open when root B becomes active (concurrent roots via
+     * fibers / multiple span stacks), this discards A's accumulated data on
+     * the A->B transition, and switching back to A later starts it fresh
+     * rather than resuming. See the class docblock.
+     */
+    private function resetForRootBoundary()
+    {
+        $rootId = $this->currentRootSpanId();
```
bwoebi:
```suggestion
        $rootSpan = \DDTrace\root_span();
        $rootId = $rootSpan ? $rootSpan->id : null;
```
That's totally fine.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+     * registry reads the active root span and binds to its $onClose directly.
+     *
+     * @param callable|null $rootIdResolver function(): ?int
+     * @param callable|null $rootCloseScheduler function(int $rootId, callable $reset): void
+     */
+    public function setRootSpanSeams($rootIdResolver, $rootCloseScheduler)
+    {
+        $this->rootIdResolver = $rootIdResolver;
+        $this->rootCloseScheduler = $rootCloseScheduler;
+    }
```
bwoebi: Why do we have to emulate root spans in testing rather than using real root spans? (using `ini_set("datadog.trace.generate_root_span", "0");` in the test class to avoid the autoroot span, then manually start some root spans as needed. Should simplify logic in the class code here a bit.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+                $this->rootCloseBoundRootId = null;
+            }
+        };
+
+        if ($this->rootCloseScheduler !== null) {
+            \call_user_func($this->rootCloseScheduler, $rootId, $reset);
+            return;
+        }
+
+        if (!\function_exists('DDTrace\\root_span')) {
```
bwoebi: General note on function_exists checks: Please avoid them, unless the code is actually reachable through the proper API surface with the tracer extension not loaded. (Most of the FFE stuff is already gated via `UnavailableEvaluator` in `NativeEvaluator::create()`, preventing this code from being reached at all, except through direct calls, which users aren"t supposed to do.)

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+            return;
+        }
+
+        $root = \DDTrace\root_span();
+        if ($root === null) {
+            return;
+        }
+
+        $root->onClose = self::prependOnCloseReset($root->onClose, $reset);
```
bwoebi:
```suggestion
        // $onClose is walked in reverse registration order, explicitly try running first.
        array_unshift($root->onClose, $reset);
```
RootSpanData->onClose is typed as array, no need for checks.

Not sure why prependOnCloseReset does an extra wrapping of $reset in a Closure - that's unneeded.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+     * retention the review flagged).
+     *
+     * @param int $rootId
+     */
+    private function scheduleResetOnRootClose($rootId)
+    {
+        if ($this->rootCloseBoundRootId === $rootId) {
+            return;
+        }
+        $this->rootCloseBoundRootId = $rootId;
```
bwoebi: Why does this function have an extra guard, when function is always executed in line with resetForRootBoundary, and nowhere else, and resetForRootBoundary has effectively the same guard?

The call to `$this->scheduleResetOnRootClose($rootId);` could be replaced by:

```
        array_unshift($rootSpan->onClose, function () use ($rootId) {
            if ($this->rootId === $rootId) {
                if ($this->accumulator !== null) {
                    $this->accumulator->clear();
                }
                $this->rootId = null;
            }
        });
```
And that's it.

The code gets harder to parse with a lot of redundant checks spread over multiple functions.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+
+            if ($this->accumulator === null) {
+                $this->accumulator = new SpanEnrichmentAccumulator();
+            }
+
+            $exposure = $details->getExposureData();
+            $serialId = is_array($exposure) && array_key_exists(self::SERIAL_ID_METADATA_KEY, $exposure)
+                ? $exposure[self::SERIAL_ID_METADATA_KEY]
+                : null;
+            $doLog = is_array($exposure) && !empty($exposure[self::DO_LOG_METADATA_KEY]);
```
bwoebi: Is $exposure ever not an array?

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+            } else {
+                $variant = $details->getVariant();
+                if ($variant === null || $variant === '') {
+                    $this->accumulator->addDefault((string) $flagKey, $details->getValue());
+                }
+            }
+
+            $this->stage();
+        } catch (\Throwable $e) {
+            // Never let span enrichment break flag evaluation.
```
bwoebi: Shouldn't this be logged? (possibly do the try-catch at the call site in Client.php where the logger is present)

src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -111,12 +130,26 @@ private function resolve(
...
         // final provider result.
         $this->recordEvaluationMetric($flagKey, $details);
+        // DG-004: PHP has no finally hook with ResolutionDetails, so APM span
+        // enrichment is accumulated INLINE here, on the same code path and from
+        // the same $details, immediately after the metrics hook. Skipped entirely
+        // with the gate off (no binder was constructed); when on, it feeds the
+        // SHARED registry so this provider's tags aggregate with any other
+        // evaluation path active on the same root span.
+        if ($this->spanEnrichmentBinder !== null) {
+            $this->spanEnrichmentBinder->accumulate(
```
bwoebi: Why is spanEnrichment duplicated across evaluate() callers (i.e. here and in Client.php) rather than being part of the NativeEvaluator::evaluate() call, which has all the context needed (and which also properly guards against extension being loaded as well)?

src/api/FeatureFlags/SpanEnrichmentBinder.php
bwoebi, line 1: What's the purpose of this class?

From the comments I suppose it's an historical artifact and can actually be removed now - `gateEnabled` can be moved to `SpanEnrichmentRegistry`, and the static accumulate function simply called directly.

tracer/ddtrace.stub.php
```diff
@@ -1092,6 +1093,29 @@ function record_ffe_evaluation_metric(string $flagKey, ?string $variant, ?string
...
+     * Stage the encoded APM feature-flag span enrichment tags in native
+     * request-local memory. The values are written onto the root span's meta
+     * when the root span closes and cleared afterwards (request-scoped). Passing
+     * null/empty for a tag clears that slot. This is gated by
+     * DD_EXPERIMENTAL_FLAGGING_PROVIDER_SPAN_ENRICHMENT_ENABLED; the provider
+     * only calls it when the gate is on (DG-004 inline accumulation / DG-005).
+     *
+     * @internal
+     */
+    function set_ffe_span_enrichment_tags(?string $flagsEnc, ?string $subjectsEnc, ?string $runtimeDefaults): void {}
```
bwoebi: What's the point of this function, compared to trivially assigning the values on `\DDTrace\root_span()->meta[<the three keys>] = <value>;` in PHP code directly and unsetting if null? There's no reason that these tags must only be added at the very end to the root span, they can just live there.

From what I see there's not even any special internal state across root spans and such nor internal code writing to this, just simple assign-and-clear.

tracer/functions.c
```diff
@@ -2114,6 +2121,35 @@ PHP_FUNCTION(dd_trace_internal_fn) {
                 waited += 10;
             }
             RETVAL_BOOL(ddog_is_agent_info_ready());
+#ifdef DD_TEST_HELPERS
```
bwoebi:
```suggestion
```
Drop this guard, dd_trace_internal_fn is basically only test/debug helpers, but we keep them. Just like you unconditionally set DD_TEST_HELPERS in config.m4.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,366 @@
...
+ * native staging bridge (tracer/ffe.c set_ffe_span_enrichment_tags) mirrors
+ * that with a single set of request-global slots. When two roots are
+ * genuinely open concurrently (fibers, multiple span stacks) and evaluations
+ * interleave across them, switching the active root discards the other
+ * root's accumulated data, and flushing can attach staged tags to the wrong
+ * root. Fixing this properly requires per-root-keyed storage on BOTH sides
+ * (PHP accumulators keyed by root id, AND the native slots keyed the same
+ * way) -- a native wire-contract-adjacent change, not a PHP-only patch.
+ * Deferred as a follow-up given the feature is experimental and gated off by
+ * default.
```
bwoebi: If you drop the native part (as I suggested in the comment to the stub file), this becomes simpler, and you can just tie the accumulator to the rootspan via `ObjectKVStore`, and the logic doesn't even need $rootId at all, but can just become `$accumulator = ObjectKVStore::get($rootSpan, "ffe_accumulator"); if (!$accumulator) { ObjectKVStore::get($rootSpan, "ffe_accumulator", $accumulator = new SpanEnrichmentAccumulator); } /* do some logic to update the accumulator */`. This will be trivially compatible with multiple root spans active at the same time.

bwoebi: Further, the current code calls `$this->stage()` unconditionally on every accumulate() call. Which really isn't efficient. You could expand `if (!$accumulator) { \array_unshift($rootSpan->onClose, static function ($rootSpan) use ($accumulator) { $accumulator->stage($rootSpan); }); }` (move `stage` function to accumulator then).

bwoebi: And at that point this whole SpanEnrichmentRegistry becomes trivial, doesn't need to be a singleton, but it can be a single static `accumulate` function without any other internal storage.

bwoebi (changes requested):
I've been reviewing this and slowly getting understanding of the architecture created here. I think this implementation is a lot more complex than is justified for what it does.

Please check the comments in reverse order (newest first) when fixing, a lot of the earlier comments should be obsolete, if just refactoring the SpanEnrichmentRegistry as suggested to be very straightforward.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,111 @@
...
+                    $accumulator->addDefault((string) $flagKey, $details->getValue());
+                }
+            }
+
+            // Write (overwrite) the encoded union onto the root span's meta. Safe
+            // to do on every evaluation: toSpanTags() re-encodes the accumulator's
+            // full state, so the latest write is always the complete union.
+            foreach ($accumulator->toSpanTags() as $key => $value) {
+                $root->meta[$key] = $value;
+            }
```
bwoebi: Yes, it's safe, but do we actually want to always compute it (which seems rather expensive to me?)

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,111 @@
...
+final class SpanEnrichmentRegistry
+{
+    const ACCUMULATOR_KEY = 'ffe_accumulator';
+    const SERIAL_ID_METADATA_KEY = 'serialId';
+    const DO_LOG_METADATA_KEY = 'doLog';
+    const CONFIG_KEY = 'DD_EXPERIMENTAL_FLAGGING_PROVIDER_SPAN_ENRICHMENT_ENABLED';
+
+    /**
+     * Whether the experimental span-enrichment gate is on. Process-level env
+     * config; cannot toggle mid-request.
```
bwoebi:
```suggestion
     * Whether the experimental span-enrichment gate is on.
```
It can technically be toggled mid-request, but it's not a case we should care about.

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
```diff
@@ -0,0 +1,357 @@
...
+    /**
+     * Reset all accumulated state. Called after the tags are flushed onto the
+     * root span so a reused accumulator never leaks across spans/requests.
+     */
+    public function clear()
+    {
+        $this->serialIds = array();
+        $this->subjects = array();
+        $this->defaults = array();
+    }
```
bwoebi:
```suggestion
```
obsolete now

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
```diff
@@ -0,0 +1,357 @@
...
+    /** @var array<string, string> flagKey => stringified default value. */
+    private $defaults = array();
+
+    /**
+     * Record a serial id seen during evaluation. Deduped via a set; dropped
+     * (with no error) once the frozen cap is reached.
+     */
+    public function addSerialId($id)
+    {
+        $id = (int) $id;
```
bwoebi: You do put a cast a few times below the function start in this class, why not put it in the signature like `function addSerialId(int $id)`? PHP will even cast it for you if not an int currently.

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
```diff
@@ -0,0 +1,357 @@
...
+     * (e.g. CJK, emoji) far below $maxLength chars. Kept as its own method so
+     * it is directly testable regardless of whether ext-mbstring is loaded in
+     * the environment running the tests.
+     */
+    private function truncateUtf8ByteFallback($value, $maxLength)
+    {
+        if (strlen($value) <= $maxLength) {
+            return $value;
+        }
+
```
bwoebi, line 335: Instead of manual counting `return preg_match("(.{,$maxLength})u", $value, $m) ? $m[0]  : "";` would work too, assuming the input is UTF-8.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            // Write (overwrite) the encoded union onto the root span's meta. Safe
+            // to do on every evaluation: toSpanTags() re-encodes the accumulator's
+            // full state, so the latest write is always the complete union.
+            foreach ($accumulator->toSpanTags() as $key => $value) {
+                $root->meta[$key] = $value;
+            }
+        } catch (\Throwable $e) {
+            // Enrichment must never break flag evaluation.
```
bwoebi, line 110: I think that comment landed under outdated now, but just wondering, intentional that it's not even logged?

pavlokhrebto:
## Summary of changes since the review

Thanks @bwoebi — the "more complex than justified" call was right. Reworked the architecture along the lines you suggested and then addressed the follow-up comments. Overview:

### Architecture: state on the span, not a native staging layer
- Span-enrichment state now lives **on the root span** via `ObjectKVStore` (the PHP analogue of the `WeakMap`/`WeakKeyDictionary` the other SDKs use), and the encoded `ffe_*` tags are written **directly to `root_span()->meta`**.
- Removed the entire native staging layer: `DDTrace\Internal\set_ffe_span_enrichment_tags()`, the request-global staging slots, and the span-close flush / dropped-root / rshutdown clear hooks. `ffe.c` / `span.c` / `ddtrace.c` are back to their master baseline.
- Removed `DDTrace\Internal\peek_root_span_id()` — uses `\DDTrace\root_span()` directly.
- Removed `SpanEnrichmentBinder` and the `SpanEnrichmentRegistry` singleton + manual root-id/reset lifecycle; it's now a stateless helper (`gateEnabled()` + `record()`).
- Enrichment is recorded once in `NativeEvaluator::evaluate()` (the shared choke point, only reachable with the extension loaded) instead of duplicated in `Client` and `DataDogProvider`; dropped the `function_exists` guards.
- Dropped the `#ifdef DD_TEST_HELPERS` guard around `await_ffe_config` (kept, like the other internal helpers).

### Concurrent roots now work
Keying the accumulator on the span object means concurrently-open roots (fibers / multiple span stacks) each keep their own tags — the "known limitation" I'd previously documented is gone, with a `.phpt` covering it.

### serial_id width
Narrowed `FfeResult.serial_id` from `i64` to `i32` — the libdatadog source is `Option<i32>`.

### Follow-up polish (your second pass)
- Deleted the now-unused `clear()`; dropped the redundant `is_array($exposure)` guards (`getExposureData()` is typed `array`).
- Simplified UTF-8 truncation to `preg_match('/\A.{0,N}/su', …)` (byte-walk kept only as an invalid-UTF-8 fallback).
- Moved the casts into typed signatures (`addSerialId(int $id)`, etc.).
- Added a `dirty` flag so `record()` only re-encodes/rewrites the tags when an evaluation actually changed state.

### Tests
Rewrote the tests to drive **real** root spans (aggregation onto meta, per-root isolation across concurrent stacks, gate-off no-op); the frozen codec keeps its unit suite.

Split into reviewable commits (PHP core → native removal → tests → comment cleanup → review follow-ups).

src/api/FeatureFlags/SpanEnrichmentRegistry.php
pavlokhrebto: Intentional — enrichment must never surface into flag evaluation, and there is no logger wired into this static path; happy to add a debug-level log if you would prefer.

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
pavlokhrebto: Done — switched to `preg_match("/\A.{0,N}/su", …)`, keeping the byte-walk only as a fallback for invalid UTF-8.

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
pavlokhrebto: Done — moved the casts into the signatures (`addSerialId(int $id)`, `addSubject(string, int)`, `addDefault(string, …)`).

src/api/FeatureFlags/SpanEnrichmentAccumulator.php
pavlokhrebto: Done — removed the now-unused `clear()` (and its test).

src/api/FeatureFlags/SpanEnrichmentRegistry.php
pavlokhrebto: Done — trimmed the docblock as suggested.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
pavlokhrebto: Good call — added a `dirty` flag so `record()` only re-encodes and rewrites the tags when an evaluation actually changed the accumulated state.

src/api/FeatureFlags/SpanEnrichmentRegistry.php
pavlokhrebto: No — `getExposureData()` is typed `array`, so I dropped the redundant `is_array()` guards.

tests/ext/ffe/span_enrichment_concurrent_roots.phpt
```diff
@@ -0,0 +1,65 @@
+--TEST--
+FFE span enrichment: concurrently-open root spans each keep their own tags
+--SKIPIF--
+<?php
+if (getenv('PHP_PEAR_RUNTESTS') === '1') {
+    die('skip: the src/ PHP API is not shipped in the PECL test package');
```
bwoebi: tests needing src should be run with phpunit. Luckily phpunit can run .phpt tests as well, you'll just have to move it into the OpenFeature folder (I believe?).

bwoebi (approved):
Looks good to me now!

bwoebi:
You can ignore the test failures, and merge, they're unrelated.

PR #3998, opened 2026-06-17
refactor(profiling): remove locks and simplify PROFILER_NAME_STR/PROFILER_VERSION_STR

morrisonlevi:
### Description

In a recent PR we refactored to use LazyLocks instead of `lazy_static` macros. While reviewing that, I suspected these could remove the lazy lock altogether and be made const. It's true, we can, but it was not as "idiomatic" as I expected. Nonetheless, I think it's an improvement.

Note that if we panic, it will be at build time, as these are const contexts.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl (approved):

realFlowControl:
The failing UBSAN is fixed in upstream https://github.com/krakjoe/parallel/pull/381

PR #4004, opened 2026-06-22
fix(ci): PHP profiler language tests YAML invalid

realFlowControl:
### Description

The last cleanup commit (failing if profiling is not loaded in PHP Language Tests) in #3364  actually broke the YAML and it failed in complete silence. Turns out: if there is a : sign somewhere in the list, it is not a array of strings in YAML anymore, but a map ...

On a personal note: YAML was a mistake

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Requiring no quoting for values in yaml was indeed a mistake

PR #4002, opened 2026-06-22
ci: publish CI images to internal registry first

realFlowControl:
### Description

When building CI docker images, this PR changes the process to:
- publish images to `registry.ddbuild.io` (Datadog internal container registry)
- use those images directly for GitLab Jobs (they are authenticated anyway)
- use the `public-images` downstream job to magically sync those images to Docker Hub for usage with GitHub CI and external contributors
- CI images jobs are generated by reading from the `docker-compose.yml` + `.env` files
- Windows build + publish jobs are now generated from `docker-compose.yml` too (previously hand-written)

#### So how do I build images?

- kick off the `ci-images` job in the GitLab CI Pipeline
- follow the link to the downstream job and kick off the builds you need
- after those are finished, kick off the publish jobs for those new images you need synced to Docker Hub

#### Wins

- no logging in to Docker Hub to get a PAT
- no manually starting a GitLab CI run anymore (with that PAT)
- no manual syncing public Docker Hub images to our internal registry (well running a script and making a PR and finding someone to approve and wait for them to be synced)
- the job list is generated from the compose files, so adding/removing a PHP version is a one-line change with no pipeline edits
- publish is independent of building, so Docker Hub can be (re)provisioned from existing internal images without a rebuild

### Status

All images (including Windows) have been rebuild using this branch and pushed to `registry.ddbuild.io`, the `*bookworm_9` images have been synced to Docker Hub, once approved I'll sync the other images as well.

### Reviewer checklist
- [ ] Test coverage seems ok.
- [ ] Appropriate labels assigned.

PR #4011, opened 2026-06-24
fix(ci): `apt-get update` before `apt install` in docker-image jobs

realFlowControl:
### Description

While working on #4002 all `randomized tests` job suddenly started failing. This is because of the apt index is stale on the image and we are not updating it before running `apt install -y php git make curl`, so that fails with:

```
Err … libxml2 2.9.14+dfsg-1.3ubuntu3.7  404 Not Found
E: Failed to fetch …/libxml2_2.9.14%2bdfsg-1.3ubuntu3.7_amd64.deb  404  Not Found
E: Unable to fetch some archives, maybe run apt-get update?
```

This also fails on `master` since today: https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-php/-/jobs/1799170178

This PR adds `apt-get update` before each of the three `apt install` steps in `.gitlab/generate-package.php`. The other apt-based jobs in this file already do this.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.gitlab/generate-package.php
+++ b/.gitlab/generate-package.php
@@ -798,6 +798,7 @@ endforeach;
     DOCKER_COMPOSE_DOWNLOAD_NAME: docker-compose-linux-x86_64
   before_script:
 <?php dockerhub_login() ?>
+    - apt-get update
     - apt install -y php git make curl
     - curl -L --fail https://github.com/docker/compose/releases/download/v2.36.0/${DOCKER_COMPOSE_DOWNLOAD_NAME} -o /usr/local/bin/docker-compose
     - chmod +x /usr/local/bin/docker-compose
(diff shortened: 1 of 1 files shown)


Leiyks (approved):

bwoebi:
Alternatively, we could just have rebuilt these images, but okay :-D

realFlowControl:
We have an aging todo in there to build a proper image that has php, make and git pre-installed:

https://github.com/DataDog/dd-trace-php/blob/8b0c1e6666b95e969da8d02f373c2ba58e73becb/.gitlab/generate-package.php#L792

The image used here is "just" an ubuntu iamge. I figured it'll be a little less bad than it is currently with adding `apt-get upadte`. And this way the todo comment stays valid :-P

PR #4013, opened 2026-06-24
ci: publish docker image for system tests

MilanGarnier:
### Description

Adds a CI job at `dd-trace-php/.gitlab/generate-package.php` which pushes the build to github's container registry (similarly to what already exists for prod builds). This is necessary to enable `LIBRARY_TARGET_BRANCH` to work for the PHP target in `utils/scripts/load-binary.sh`. (mimicking what already exists for dd-trace-dotnet)

The endgoal is to have https://github.com/DataDog/system-tests/pull/7209 picking up the right branch

### Reviewer checklist


.gitlab/generate-package.php
```diff
@@ -1515,6 +1515,39 @@
...
+"publish docker image for system tests":
+  stage: release
+  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/docker:29.4.0-noble
+  tags: [ "docker-in-docker:amd64" ]
+  rules:
+    - when: on_success
+  needs:
+    - job: "datadog-setup.php"
+      artifacts: true
+    - job: "package extension: [x86_64, x86_64-unknown-linux-gnu]"
```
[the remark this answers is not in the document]
MilanGarnier: Fixed — changed `[x86_64, x86_64-unknown-linux-gnu]` to `[amd64, x86_64-unknown-linux-gnu]` to match the job name generated by `$platform['arch']`.

realFlowControl:
I just recently learned that https://github.com/DataDog/public-images exists, could this do the syncing of the image for us?

MilanGarnier:
@realFlowControl I didn't know `publish-images` either, but from what I understand, the tool seems to be centered aroung publishing releases across various sources, whereas what I want to do here is to push once a build, exclusively for system-tests to use. (but I'm really not an expert here)

.gitlab/generate-package.php
```diff
@@ -1518,6 +1518,51 @@
...
+      artifacts: true
+    - job: "datadog-setup.php"
+      artifacts: true
+    - job: "package extension: [amd64, x86_64-unknown-linux-gnu]"
+      artifacts: true
+  variables:
+    GIT_STRATEGY: none
+  script: |
+    set -e
+    IMAGE="ghcr.io/datadog/dd-trace-php/dd-library-php:${CI_COMMIT_REF_SLUG}"
```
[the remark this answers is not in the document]
MilanGarnier: relevant but ran against an outdated commit, fixed in system-tests

bwoebi (approved):
So, I don't mind this. I just don't know whether there is any problem in having that gigabyte sized dockerfile produced with every commit.
This will quickly occupy terabytes.
Do we have to pay for that? Must we clean up old images eventually? Or is this already handled?

MilanGarnier:
> So, I don't mind this. I just don't know whether there is any problem in having that gigabyte sized dockerfile produced with every commit. This will quickly occupy terabytes. Do we have to pay for that? Must we clean up old images eventually? Or is this already handled?

@bwoebi Those are very good questions. I investigated and I made a few improvements to adjust the scope of this PR:
- I ended up adding a workflow responsible for deleting the last uploaded image when branches are closed. It mimics the way it currently works in the dotnet tracer, except that it's not hidden behind a PR tag. (https://github.com/DataDog/dd-trace-dotnet/blob/e5a508263cdf2131bc183d201866491f8479c948/.github/workflows/delete-pr-image.yml)
- Since we push images with a given tags may already be present on ghcr, The publishing job checks and tries to delete the image present with that given tag before uploading. This should prevent piling up untagged images
- I disabled the publishing on the default branch, since it already gets published differently, the goal of this PR being to run system tests on feature branches in the first place

realFlowControl:
Hey @MilanGarnier, just a heads up: I merged #4002 and you might see conflicts in your PR

.github/workflows/delete-pr-image.yml
```diff
@@ -0,0 +1,31 @@
+name: "Delete Docker image on PR close"
```
bwoebi, line 1: If you tie this to PRs, you probably also should restrict publishing docker images to branches having an associated PR as well.

MilanGarnier: I added a test in the publish job, the job will exit with a warning if it runs on a branch without an associated PR
Maybe an `exit 0` in that case can be desired, but I guess a warning makes it clearer that running the system tests on a feature branch will require an open PR and not only having the remote branch exist. 

[Gate upload to having a PR open](https://github.com/DataDog/dd-trace-php/pull/4013/commits/47446346670a9480b79f690b9f9b06c883912998)

bwoebi: Looks good (with exit 3)

PR #4017, opened 2026-06-25
chore: cleanup bits php logo

realFlowControl:
### Description

The Bits vs. PHP logo should use the same background colour as the official PHP logo which is #777BB3

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Beautiful. :-)

PR #4024, opened 2026-06-30
fix(live-debugger): retire removed probe configs to prevent use-after-free

Leiyks:
## Problem

`test_extension_ci: [7.1]` (the **valgrind** master job) fails with a `LEAKED TEST SUMMARY` on `tests/ext/live-debugger/debugger_enable_dynamic_config.phpt`. The `.mem` artifact shows it is **not a leak but a use-after-free**:

```
Invalid read of size 1
  ... probe.id.to_utf8_lossy()                    send_data.rs:523
  ddog_send_debugger_diagnostics                  remote_config.rs:786
  dd_probe_mark_active                            tracer/live_debugger.c:226
  dd_span_probe_begin                             tracer/live_debugger.c:239
 Address ... is 0 bytes inside a block of size 1 free'd
  drop String -> Probe -> ... -> Box<(RemoteConfigParsed, MaybeShmLimiter)>
  ddog_process_remote_configs                     components-rs/remote_config.rs:445  (LiveDebugger Remove)
```

The freed block is **size 1** — the 1-byte probe id `"1"` the test generates.

## Root cause

The C side (`tracer/live_debugger.c:162`, `def->probe = *probe`) keeps a shallow copy of the FFI `ddog_Probe`, whose `ddog_CharSlice` fields (`id`, capture config, shm limiter, …) **borrow** strings owned by the boxed `RemoteConfigParsed` stored in `live_debugger.active`. The `Box<>` exists *specifically* to give that data a stable heap address for these borrows (per its own comment).

On a LiveDebugger **Remove** (and on an `Entry::Occupied` re-add) the box was dropped immediately, freeing those strings. But `zai_hook_remove` can **defer** the actual hook teardown, so a still-installed probe can fire afterwards and read the freed strings via `ddog_send_debugger_diagnostics(&def->probe, …)` → UAF.

Latent since the feature landed (`f87446c4d`, 2024-10); only the valgrind job observes it because plain runs read the freed-but-intact bytes and pass. It is timing-gated, which is why it surfaces in the full shuffled valgrind suite rather than in isolation.

## Fix

Retire removed/replaced boxes into a new `live_debugger.retired` vec instead of dropping them, and free them in `ddog_rshutdown_remote_config` — by which point all probe hooks are torn down and no PHP user code can fire a probe. `retired` is cleared every request, so it stays bounded.

## Verification

- Builds on PHP 7.1 with the master-pinned libdatadog (`6a6d4a535`).
- The previously-leaking test passes **10/10 under valgrind** with the fix; non-regressing.
- Note: the UAF needs full-suite (shared request-replayer + shuffle) timing and does not reproduce in isolation, so the authoritative signal is the `test_extension_ci: [7.1]` valgrind job on this branch.

### DIFF
--- a/components-rs/datadog.h
+++ b/components-rs/datadog.h
@@ -187,6 +187,8 @@ void ddog_set_dynamic_instrumentation_enabled(struct ddog_RemoteConfigState *rem
 
 void ddog_rshutdown_remote_config(struct ddog_RemoteConfigState *remote_config);
 
+void ddog_live_debugger_free_retired(struct ddog_RemoteConfigState *remote_config);
+
 void ddog_shutdown_remote_config(struct ddog_RemoteConfigState*);
 
 void ddog_log_debugger_data(const struct ddog_Vec_DebuggerPayload *payloads);
--- a/components-rs/remote_config.rs
+++ b/components-rs/remote_config.rs
@@ -111,6 +111,13 @@ pub struct LiveDebuggerCallbacks {
 pub struct LiveDebuggerState {
     pub spans_map: HashMap<String, i64>,
     pub active: HashMap<String, Box<(RemoteConfigParsed, MaybeShmLimiter)>>, // Box<> for stable heap address!
+    // Parsed configs removed (or replaced) from `active` whose probe hooks may still be
+    // installed: the C side stores shallow `ddog_Probe` copies that borrow `CharSlice`s
+    // (probe id, capture config, shm limiter, ...) into these boxes, and `zai_hook_remove`
+    // can defer the actual teardown of those hooks past this point. Dropping the box here
+    // would free the borrowed strings while a deferred hook can still fire and read them
+    // (use-after-free). Keep them alive until request shutdown, when all hooks are gone.
+    pub retired: Vec<Box<(RemoteConfigParsed, MaybeShmLimiter)>>,
     pub config_id: String,
     pub allow_dfa: Option<Regex>,
     pub deny_dfa: Option<Regex>,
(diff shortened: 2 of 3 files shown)


bwoebi:
My problem with this PR is that it misidentifies the root cause:
There's no reason why the begin hook should be called in the first place on an already removed hook. This should not be possible - zai_hook_continue carefully uses the ht iterators API to that extent and checks for negative hook id, which signals removal.

And end hooks are anyway guarded with `def->removed` where needed.

I don't see where ddog_process_remote_configs was called from the shortened trace, so that gives me no hint.

Leiyks:
Thanks — you're right, and it reframes the bug. Here's the full trace you were missing.

All 4 invalid reads are the same 1-byte block (the probe id `"1"`), read while building/serializing the diagnostics payload:

**Read**
```
probe.id.to_utf8_lossy()                 send_data.rs:523
ddog_send_debugger_diagnostics           remote_config.rs:786
dd_probe_mark_active                     live_debugger.c:226
dd_span_probe_begin                      live_debugger.c:239
zai_hook_continue                        hook.c:1053    // the hook->begin(...) call
zai_interceptor_ext_nop_handler_no_prev  interceptor.c:186
```
**Freed by**
```
drop Box<(RemoteConfigParsed, MaybeShmLimiter)>
ddog_process_remote_configs              remote_config.rs:445   // LiveDebugger Remove arm
dd_vm_interrupt                          remote_config.c:29
```

So the caller is `dd_vm_interrupt` — the interrupt-driven RC poll processing a LiveDebugger **Remove**, which drops the box owning the probe strings.

You're right that `zai_hook_continue` skips `hook->id < 0`, so `begin` can't fire on the removed hook. That's exactly the point: the hook firing here has `id >= 0` — it was **never removed** — yet its backing `active` box was freed by a Remove. The only way both hold is that the probe had **more than one installed zai hook**, with `spans_map` tracking only the last: `remove_config` tears down the tracked hook and frees the box, while an earlier, untracked hook stays live and keeps borrowing it (`def->probe` is a shallow copy whose `CharSlice`s point into that box).

The orphaning comes from `spans_map.insert(probe_id, hook_id)` overwriting a prior entry **without removing the prior hook**, in both install paths:
- `apply_config` (set_probe at remote_config.rs:494 + the following `spans_map.insert`)
- the di-enable reinstall in `ddog_set_dynamic_instrumentation_enabled` (set_probe at remote_config.rs:736 + `spans_map.insert`)

Neither checks whether `spans_map[probe_id]` already holds a hook, so any sequence that installs the same probe twice against the same `active` box leaks the first hook; the tracked one is later removed on Remove (freeing the box) and the leaked one dangles.

That makes the real defect the hook **tracking**, not the box lifetime — this PR's box-retirement only masks the memory error (the orphan would still fire `begin` and emit bogus EMITTING diagnostics for a probe that's meant to be gone). The proper fix looks like: remove/dedup the existing hook before overwriting `spans_map[probe_id]` (or track all hook ids per probe so removal tears every one down).

I haven't yet pinned the exact double-install sequence for this test (it doesn't reproduce in isolation — the Remove only appears in the full shuffled valgrind suite, and I hit local env friction trying to log it). Does the orphan-hook framing match your read? If so I'll rework the PR toward the install/removal-tracking fix (and confirm with targeted logging) rather than the box-retirement approach.

bwoebi:
Installing the same probe twice should not be possible either given that RemoteConfigManager::fetch_update() should never emit an add without a previous removal for any given path.

This really needs a reproducer.

PR #4025, opened 2026-06-30
Propagate agent configured timeout to sidecar

bwoebi:
And make two RC tests more resilient to race conditions.

Increase the timeout in tests to avoid genuine timeouts from request-replayer overload to happen.


Leiyks (approved):

Leiyks (approved):
looks good to me, let's see if it definitely fixes our issues

PR #4030, opened 2026-07-02
fix(curl): enforce DD_TRACE_SPANS_LIMIT for curl_multi_exec parent spans (APMS-19944)

Leiyks:
## What

Skip creating the `curl_multi_exec` parent span when the tracer is already limited (`DD_TRACE_SPANS_LIMIT` reached / memory pressure).

## Why (APMS-19944)

Follow-up to #3691 (APMS-18744). That PR limited the *child* header-injection spans but left the **parent** `curl_multi_exec` span unbounded.

The parent span is opened via the user-facing `DDTrace\start_span()` (`CurlIntegration.php`), which #3691 deliberately left exempt from the span limit (guarding it broke `testTracerFlushedWhenSpanLimitExceeded` and Guzzle's `isolateTracer()`). Distributed-header injection then flags that span `DDTRACE_SPAN_FLAG_NOT_DROPPABLE` (`handlers_http.h`), so the post-hook's `try_drop_span()` can never reclaim it.

In long-running CLI workloads that repeatedly call `curl_multi_exec` (e.g. AWS SDK SQS `sendMessage` batches), one un-droppable parent span accumulates **per call**, growing memory 1:1 with calls until **OOM**.

Reported: PHP 8.2, ECS Fargate ARM64, reproduced on 1.17.0 and 1.22.0 — 10,000 iterations → ~10,500 spans in one trace.

## Fix

Gate the *internal integration* use of `start_span()` with `dd_trace_tracer_is_limited()`, without touching the user-facing API (so the tests behind the #3691 revert stay green). When limited, the C handler still injects distributed tracing headers against the active span, so propagation is preserved; child request spans are already limited.
 
Added regression test `tests/ext/integrations/curl/curl_multi_exec_span_limit.phpt` — PASS with the fix, FAIL (`LEAK`) without it.

Fixes: APMS-19944

### DIFF
--- a/src/DDTrace/Integrations/Curl/CurlIntegration.php
+++ b/src/DDTrace/Integrations/Curl/CurlIntegration.php
@@ -100,7 +100,7 @@ final class CurlIntegration extends Integration
         ]);
 
         \DDTrace\install_hook('curl_multi_exec', static function (HookData $hook) {
-            if (\count($hook->args) < 2) {
+            if (\count($hook->args) < 2 || \dd_trace_tracer_is_limited()) {
                 return;
             }
             $data = null;
@@ -111,10 +111,6 @@ final class CurlIntegration extends Integration
             }
             if ($data) {
                 $hook->data = $data;
-                return;
-            }
-
-            if (\dd_trace_tracer_is_limited()) {
                 return;
             }
 
--- a/src/DDTrace/Integrations/Curl/CurlIntegration.php
+++ b/src/DDTrace/Integrations/Curl/CurlIntegration.php
@@ -114,6 +114,10 @@ final class CurlIntegration extends Integration
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/DDTrace/Integrations/Curl/CurlIntegration.php
```diff
@@ -114,6 +114,10 @@ public static function init(): int
                 return;
             }
 
+            if (\dd_trace_tracer_is_limited()) {
```
bwoebi: Move it earlier to avoid the end hook (hook->data check)?

Leiyks: Done in 06bcdf29a — moved the `dd_trace_tracer_is_limited()` check to the top

bwoebi (approved):

PR #4036, opened 2026-07-08
fix(live-debugger): fix use-after-free, tags leak, and Windows 7.x crash on probe removal

Leiyks:
Fixes live-debugger probe-lifetime bugs surfaced by a flaky `LEAKED` on `debugger_enable_dynamic_config.phpt`, plus a Windows-only crash the new regression test exposed.

## Fixes
1. **Use-after-free (original failure).** `spans_map` was keyed inconsistently — `probe.id` on apply/remove vs `config_id` on the DI-reinstall path. Two configs sharing a `probe.id` orphaned a hook whose parsed-config box was then freed, so its next fire read the freed `probe.id`. Now keyed by `config_id` everywhere. Reproduced on PHP 7.1 & 8.3 under valgrind.
2. **`tags` leak.** `probe.into()` allocates a `CharSliceVec` for `tags` that `dd_probe_dtor` never freed. Added `ddog_drop_probe` (consumes the FFI probe by value) so drop glue frees `tags` + nested allocations in one place.
3. **PHP 7.x Windows crash (0xC0000005).** MSVC 2017 if-converted the inline ternary building `zai_hook_remove`'s scope arg into an unconditional `ZSTR_LEN(def->scope)`, a NULL deref for global functions. Build the `zai_str` views with explicit `if`-guards. PHP 8.x (MSVC 2019) / Linux unaffected.

Also drops the stale `active_live_debugger_hooks` entry on removal so a later lookup of the same id can't use-after-free the freed `def`.

## Test & verification
`tests/ext/live-debugger/debugger_remove_shared_probe_id.phpt` installs two configs sharing a `probe.id` (distinct tags), removes both without firing, then calls the target — exercising the UAF, the tags allocation, and the Windows removal path. Green across the full `test_c` matrix: **windows 7.2–8.5** and **ASAN (Linux) 7.4–8.5**; valgrind clean (0 bytes lost, no double-free).


bwoebi (approved):
I'm still not persuaded that probe_ids can effectively be shared, but it won't break anything. So let's do that.

PR #4038, opened 2026-07-10
chore(profiling): clean up code and update Rust dependencies

realFlowControl:
### Description

This PR cleans up the profiler implementation and updates its direct Rust dependencies. It removes unused code and dependencies, reduces repeated test and build-script plumbing, and updates dependencies in focused commits.

The main runtime improvements are the upgrade to `rustc-hash` 2.1.3, which provides a faster and better-finalized hash algorithm with improved string hashing for stack interning, live heap tracking, and the I/O file descriptor cache, and the upgrade to `rand` 0.9.4 plus `rand_distr` 0.5.1, which uses the rewritten large-lambda Poisson sampler. In a local two-million-sample release benchmark, Poisson sampling improved from 121.44 ns to 33.23 ns at lambda 100 and from approximately 114 ns to 22 ns at the I/O and allocation lambdas, a 3.7–5.3x speedup. The Poisson state grows by 48 bytes per sampler, approximately 480 bytes per PHP thread (1 for NTS, n for ZTS).

The `dynasmrt` 5.0.0 update also picks up internal AArch64 instruction-cache handling and an upstream fix for an illegal-instruction crash on Apple AArch64, directly covering the generated PHP frameless-call trampolines.

Remaining updates are maintenance and compatibility upgrades; Criterion and UUID are intentionally unchanged because their latest versions would respectively conflict with `criterion-perf-events` and add another `getrandom` generation without improving profiler code.

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Nice cleanup. I like the merging of zts+debug checks :-D

PR #4043, opened 2026-07-15
fix: report failed capture-expression evaluations instead of dropping

p-datadog:
## Motivation

When a capture expression fails to evaluate, we currently drop the error (`ddog_evaluation_error_drop(result.error); continue;` in `dd_log_probe_add_capture_fields`), so the customer sees neither a value nor any error signal. Every other tracer surfaces failed capture expressions, and our own metric-probe path already reports value-evaluation errors — this is the one spot that silently omits them.

## Change

Replace the drop with `dd_submit_probe_eval_error_snapshot(&def->parent.probe, result.error)`, the same call the metric-probe and template/condition paths use. The loop still `continue`s, so remaining expressions are unaffected.

## Test

`tests/ext/live-debugger/debugger_capture_expression_error.phpt`: a log probe whose capture expression cannot evaluate (getmember on an int) now emits an `evaluationErrors` entry instead of producing no output. I couldn't run it locally (needs the dev-env request-replayer + built extension; the test SKIPIFs without a dev env) — relying on CI.

## Notes

- Part of a chain to enable PHP capture expressions end-to-end. The companion rate-limit change is in libdatadog (DataDog/libdatadog#2241); once vendored, PHP capture-expression probes also default to the snapshot rate. This PR is the tracer-side error-reporting piece and stands on its own.
- Draft pending the rest of the chain (dd-source matrix + web-ui version gate).

### DIFF
--- a/tests/ext/live-debugger/debugger_capture_expression_error.phpt
+++ b/tests/ext/live-debugger/debugger_capture_expression_error.phpt
@@ -1,0 +1,66 @@
+--TEST--
+A failed capture-expression evaluation is reported, not silently dropped
+--SKIPIF--
+<?php include __DIR__ . '/../includes/skipif_no_dev_env.inc'; ?>
+--ENV--
+DD_AGENT_HOST=request-replayer
+DD_TRACE_AGENT_PORT=80
+DD_TRACE_GENERATE_ROOT_SPAN=0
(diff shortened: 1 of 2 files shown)


bwoebi:
@p-datadog Didn't we want to rate-limit this?

Issue #2311, opened 2023-10-11
[Bug]: PHPRedis infinite flushing loop


B-Galati:
### Bug report

Hello,

When PHPRedis integration is enabled it keeps flushing while my worker is totally idled.

For example:

```
$ php bin/console -n messenger:consume -q events --time-limit=240 --memory-limit=128M
[2023-10-11T09:53:24+00:00] [ddtrace] [debug] - Attempting integrations load; note: some integrations are only loaded on first usage
[2023-10-11T09:53:24+00:00] [ddtrace] [debug] - Loaded integration web
[2023-10-11T09:53:25+00:00] [ddtrace] [debug] - Loaded integration DDTrace\Integrations\Logs\LogsIntegration
[2023-10-11T09:53:25+00:00] [ddtrace] [debug] - Loaded integration DDTrace\Integrations\PHPRedis\PHPRedisIntegration
Notifying profiler of finished local root span.
Flushing trace of size 1 to send-queue for http://datadog-agent.infra:8126
Notifying profiler of finished local root span.
Flushing trace of size 1 to send-queue for http://datadog-agent.infra:8126
Notifying profiler of finished local root span.
Flushing trace of size 1 to send-queue for http://datadog-agent.infra:8126
Notifying profiler of finished local root span.
Flushing trace of size 1 to send-queue for http://datadog-agent.infra:8126
```

The message `Flushing trace...` keeps appearing every 1 second I would say.

Is it by design or something is wrong?

### PHP version

8.2

### Tracer or profiler version

0.92.2

B-Galati:
@PROFeNoM thank you but I don't understand. I am using Symfony, not Laravel and there are no integrations that relate to these Frameworks.

PROFeNoM:
All my apologies, I read too quickly and overlooked the fact that you weren't using Laravel Horizon, but symfony messenger.

I'm not familiar with Symfony Messenger, but looking at their source code, it indeed seems like the consumer [waits](https://github.com/symfony/messenger/blob/641bde9896f8f4399426a65a0925c09d05519bfa/Command/ConsumeMessagesCommand.php#L82) 1s before checking for new messages, as you observed.

Assuming you are using a redis receiver, I would then indeed believe that the current behavior is expected.

B-Galati:
We are using RabbitMQ but that's not impossible it's calling Redis for some reasons.

It may be because the container is restarted between each messages but I need to dig further.

Not a bug with Datadog in any case!

Thanks!

PR #4046, opened 2026-07-17
WIP: feat(tracer): align C span code with v1-compatible stub API (v2 groundwork)

Leiyks:
Migrate the tracer to the V1 Efficient Trace Payload protocol.

- **Sidecar path (PHP 8.3+): v1-only.** Spans build directly into libdatadog's native v1 model via one `dd_span_sink` finalization body (no v04 intermediate): promoted `env`/`version`/`component`/`span.kind`, chunk-level 128-bit `trace_id` + `sampling_mechanism`, unified typed `attributes` (incl. `meta_struct` as bytes), native `span_links`/`span_events`.
- **Old-agent compatibility:** when the agent `/info` doesn't advertise `/v1.0/traces`, the sidecar transcodes v1→v0.4 (existing libdatadog encoder). Isolated + removable — dropping v0.4 later = delete the `/info` check + the transcode call.
- **In-process background-sender (PHP ≤8.2): stays v0.4**, transcoding v0.4→v1 on send behind the same removable check.
- Userland introspection (`dd_trace_serialize_closed_spans`) returns the v1 shape uniformly on all versions; `DD_TRACE_AGENT_PROTOCOL_VERSION` gate removed.
- Test harness: request-replayer gained a PHP v1 decoder + advertises `/v1.0/traces`; `tests/ext` expectations updated to v1 (603 passing).

Depends on libdatadog #2311. Follow-ups: native array/map AnyValue attributes (currently JSON-string), `process_tags` as a payload attribute, sidecar 404 fail-closed hardening.


tracer/serializer.c
```diff
@@ -932,7 +930,493 @@ static void dd_serialize_json(zend_array *arr, smart_str *buf, int options) {
...
+    } ZEND_HASH_FOREACH_END();
+    dd_serialize_json(Z_ARRVAL(tmp), buf, 0);
+    zval_ptr_dtor(&tmp);
+}
+
+// --- Span finalization sink (v0.4 vs native V1) ---
+// ddtrace_serialize_span_to_rust_span runs one finalization body and routes every field/meta/metrics
+// write through a dd_span_sink. On the in-process (<=8.2) path the sink targets a V0.4 ddog_SpanBytes
+// (also read back by functions.c introspection); on the sidecar path it targets the native V1 builder
+// chunk/span directly, with no V0.4 intermediate. The V1 routing promotes env/version/component/
```
bwoebi: No, we want v1 only, and have libdatadog convert back from v1 to v0.4 if needed, after creating the native spans.
Completely get rid of any v04 stuff in serializer. That's for auto_flush.c.

tracer/serializer.c
```diff
@@ -932,7 +930,493 @@ static void dd_serialize_json(zend_array *arr, smart_str *buf, int options) {
...
+        // v0.4 form is "-N"; the mechanism is the trailing unsigned integer.
+        const char *p = value.ptr; size_t n = value.len;
+        if (n && *p == '-') { p++; n--; }
+        uint32_t mech = 0;
+        for (size_t i = 0; i < n; i++) { if (p[i] < '0' || p[i] > '9') { mech = 0; break; } mech = mech * 10 + (uint32_t)(p[i] - '0'); }
+        ddog_v1_set_chunk_sampling_mechanism(b, s->chunk, mech);
+        return true;
+    }
+    // 128-bit trace-id high half is carried by the chunk trace id, not a span attribute.
+    if (DD_CS_EQ(key, "_dd.p.tid")) { return true; }
```
bwoebi: users aren't really supposed to touch that stuff at all. No conversion needed.

tracer/serializer.c
```diff
@@ -932,7 +930,493 @@ static void dd_serialize_json(zend_array *arr, smart_str *buf, int options) {
...
+
+// Route a meta (string) key to its V1 destination. Returns true when the key is consumed as a
+// promoted span field or a chunk-level attribute (and therefore must NOT be added to the span
+// attribute map). Mirrors the V0.4->V1 routing that dd_v1_convert_span used to apply post-hoc.
+static bool dd_v1_route_meta(dd_span_sink *s, ddog_CharSlice key, ddog_CharSlice value) {
+    ddog_TracerPayloadV1Builder *b = s->builder;
+    if (DD_CS_EQ(key, "env")) { ddog_v1_set_span_env(b, s->chunk, s->span, value); return true; }
+    if (DD_CS_EQ(key, "version")) { ddog_v1_set_span_version(b, s->chunk, s->span, value); return true; }
+    if (DD_CS_EQ(key, "component")) { ddog_v1_set_span_component(b, s->chunk, s->span, value); return true; }
+    if (DD_CS_EQ(key, "span.kind")) { ddog_v1_set_span_kind(b, s->chunk, s->span, dd_span_kind_meta_to_otel(value)); return true; }
```
bwoebi: We should check the values on the meta of the PHP span directly, discard them from meta (via `_find`+`_del`) and have the actual serialization only look at taking the values from the span properties, bypassing meta fully, avoiding individual checks during iteration.

components-rs/bytes.rs
```diff
@@ -152,192 +154,657 @@ fn convert_literal_to_bytes_string(string: *const c_char) -> BytesString {
...
+
+// ------------------- Chunk / span / link / event creation -------------------
+
+/// Appends a chunk carrying the 128-bit trace id (high/low halves), returning its index.
+#[no_mangle]
+pub extern "C" fn ddog_new_chunk(
+    builder: &mut TracerPayloadV1Builder,
+    trace_id_high: u64,
+    trace_id_low: u64,
+) -> usize {
```
bwoebi: Why the indirection and returning `usize` instead of a pointer to the span?
A direct pointer should be much faster rather than always fetching payload -> chunk -> span -> link via indices - up to three levels of indirection. And then it breaks down with recursive attributes anyway (a capability which isn't even exposed currently, but we'll want to make use of it for arrays - arrays no longer shall be flattened but be translated to proper nested attributes).
The only caveat is that one must not add new chunks/spans while we operate on a span, but that's not a problem, because we just don't do that (with the exception of adding the inferred span, which has - in the code on master - an existing refetch.

tracer/serializer.c
```diff
@@ -925,14 +923,212 @@ void ddtrace_set_root_span_properties(ddtrace_root_span_data *span) {
...
+// set, removes it from the source. On V1 the unified attribute map subsumes meta and metrics in one op.
+void transfer_span_attr(dd_span_sink *src, dd_span_sink *dst, const char *key, bool delete_source) {
+    ddog_transfer_span_attr(src->builder, src->chunk, src->span, dst->span, key, delete_source);
+}
+void transfer_span_metric(dd_span_sink *src, dd_span_sink *dst, const char *key, bool delete_source) {
+    ddog_transfer_span_attr(src->builder, src->chunk, src->span, dst->span, key, delete_source);
+}
+
+// Adds a string-valued V1 attribute from a zval: arrays/objects are JSON-encoded (the attribute FFI
+// has no array/object variant, so JSON avoids a lossy "Array" cast). `add_call` is the target FFI.
```
bwoebi, line 996: Where does it get this idea of JSON encoding from?

tracer/serializer.c
```diff
@@ -1542,35 +1894,46 @@ ddog_SpanBytes *ddtrace_serialize_span_to_rust_span(ddtrace_span_data *span, ddo
...
-        return NULL;
+        return (dd_span_sink){0};
     }
 
-    uintptr_t rust_span_index = ddog_get_trace_size(trace);
-    bool is_first_span = rust_span_index == 0;
-    ddog_SpanBytes *rust_span = ddog_trace_new_span(trace);
+    // The span is built directly into the native V1 builder chunk/span; every field/meta/metrics
+    // write below routes through the sink. The chunk carries the 128-bit trace id (set at creation).
+    dd_span_sink sink = {0};
```
bwoebi: Not sure why dd_span_sink still exists. `.chunk`/`->chunk` is at a glance not used outside this function. And for everything else you can just pass span directly around.

tracer/serializer.c
```diff
@@ -925,79 +923,493 @@ void ddtrace_set_root_span_properties(ddtrace_root_span_data *span) {
...
+            if (dest_is_map) ddog_attr_map_put_str(dest, key, v); else ddog_attr_list_push_str(dest, v);
+            break;
+        }
+        case IS_ARRAY: {
+            zend_array *arr = Z_ARR_P(value);
+            bool child_is_list = zend_array_is_list(arr);
+            struct ddog_AttrBuilder *child = child_is_list
+                ? (dest_is_map ? ddog_attr_map_open_list(dest, key) : ddog_attr_list_open_list(dest))
+                : (dest_is_map ? ddog_attr_map_open_map(dest, key) : ddog_attr_list_open_map(dest));
             GC_PROTECT_RECURSION(arr);
```
bwoebi: The actual *check* for recursion is missing (add test)

tracer/serializer.c
```diff
@@ -925,79 +923,493 @@ void ddtrace_set_root_span_properties(ddtrace_root_span_data *span) {
...
+// and nested arrays of those. `dd_attr_native_safe` gates on exactly that; anything else (null,
+// non-finite / scientific-notation floats, objects — which json_encode may route through
+// JsonSerializable / property mangling —, recursive arrays) keeps the old json-string path, which
+// is byte-identical by construction. Unlike the SPAN native path (which stringifies leaves to
+// mirror the dotted-key flatten), these leaves preserve the PHP scalar TYPE to match json_encode.
+
+// json_encode prints doubles in decimal notation for |x| in [1e-4, 1e17) (and 0); Rust's f64
+// Display matches it there. Outside that window json_encode switches to exponential, which Rust
+// does not reproduce, so such values fall back to the json-string path.
+static bool dd_double_json_native_safe(double d) {
```
bwoebi: drop that

tracer/ddtrace.stub.php
bwoebi: mark them @deprecated

PR #4061, opened 2026-07-24
Fix memory leak on roadrunner/frankenphp/etc due to appsec telemetry

cataphract:
### Description

During "user request mode" (one of FrankenPHP's modes, roadrunner, etc.), the telemetry buffer would grow at constant rate of 300 bytes/req, never being drained, draining would only happen at rshutdown

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
Looks right!

PR #4062, opened 2026-07-24
feat: OTel process and thread contexts

morrisonlevi:
### Description

This implements OTel process and thread contexts described here:

- [OTEP 4719 Process Context: Sharing Resource Attributes with External Readers](https://github.com/open-telemetry/opentelemetry-specification/blob/2e90a4df976ba2609577f3308ad27188a2619853/oteps/profiles/4719-process-ctx.md)
- [OTEP 4947 Thread Context: Sharing Thread-Level Information with External Readers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/profiles/4947-thread-ctx.md)

This allows out-of-process readers like an EBPF profiler on Linux to get information from the tracer such as span and trace ids (and more).

This ALSO adds support for W3C trace flags in more places. This part needs specific review too, but I added it because it's part of these OTEPs to use it.

### Platforms

To avoid having significantly different APIs on Linux vs macOS/Windows, this PR implements process and thread contexts on macOS and Windows too. They lack discovery and out-of-process readers: it's just and internal detail and is non-standard. There are still some divergences to make this work (they use different thread local storage mechanisms) but at the high level they are quite similar.

The profiler has been ported to use process and thread contexts when available on all platforms. 

### Notes

I need help to implement and verify the Windows portion.

Since PHP can have different service, env, version, etc on every request, we do not store these as resource attributes in the process context. Instead they are stored as thread attributes. I've opened a [draft PR for our EBPF profiler fork](https://github.com/DataDog/opentelemetry-ebpf-profiler/pull/98) to understand this pattern, and I've verified locally that it works. 

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/datadog.c
```diff
@@ -581,6 +604,22 @@ static PHP_RINIT_FUNCTION(datadog) {
     // Things that should only run on the first RINIT after each minit.
     pthread_once(&dd_rinit_once_control, dd_rinit_once);
 
+    // Module globals are thread-local in ZTS builds, so each thread owns and publishes its fixed
+    // mapping. In NTS this updates the single process mapping on each request.
```
bwoebi, line 608: This isn't #if !ZTS guarded?

global hostname/service/env/version/process tags should be published in minit in zts probably to have the baseline?

ext/datadog.c
```diff
@@ -581,6 +604,22 @@ static PHP_RINIT_FUNCTION(datadog) {
...
+    if (DATADOG_G(otel_process_context)) {
+        zend_string *process_tags = datadog_process_tags_get_serialized();
+        bool published = datadog_otel_process_context_publish(
+            DATADOG_G(otel_process_context),
+            dd_zend_string_to_CharSlice(get_DD_HOSTNAME()),
```
bwoebi, line 613: Is get_DD_HOSTNAME() is correct? I mean, on spans we check `if (ZSTR_LEN(get_DD_HOSTNAME())) { get_DD_HOSTNAME() } else { gethostname() }` essentially. Do we need that logic here as well, or only custom defined hostname relevant?

components-rs/lib.rs
```diff
@@ -54,6 +60,299 @@ pub static mut datadog_formatted_root_session_id: [u8; 36] = [0u8; 36];
 #[allow(non_upper_case_globals)]
 pub static mut datadog_formatted_parent_session_id: [u8; 36] = [0u8; 36];
 
+const PHP_OTEL_PROCESS_CONTEXT_SIZE: usize = 16 * 1024;
```
bwoebi, line 63: Let's create a dedicated `.rs` file to clean this up a bit.

PR #4067, opened 2026-07-27
Preserve ci-images manual job availability on development branches (docs and generator)

realFlowControl:
### Motivation

- Ensure the `ci-images` manual job and generated image-build jobs remain runnable from development branches so image changes can be built and validated before merge.
- Prevent maintainers from accidentally restricting those manual jobs with `CI_COMMIT_REF_PROTECTED` or `CI_DEFAULT_BRANCH` rules.
- Clarify that these pipelines run only in the private Datadog GitLab instance and that access to manual jobs is governed by project membership and settings.

### Description

- Add explanatory comment to `.gitlab-ci.yml` above the `ci-images` job explaining why it must remain available on development branches and not be restricted by protected-ref rules.
- Add the same guidance to the generated CI images generator `.gitlab/generate-ci-images.php` so generated child pipelines include the note in their header and template snippets.
- Update `dockerfiles/ci/README.md` with an important admonition block describing that image-build jobs are intentionally available from development branches and how access is controlled.

### Testing

- No automated tests were run because the changes are non-functional documentation and generator comments only.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a675b084460832696759c07c6b8dda9)

The sad reality about this PR: Codex keeps stumbling over this over and over and pings to me to fix a problem that does not exist. Hopefully, with some documentation, it stops wasting my time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -94,6 +94,12 @@ appsec-trigger:
 # Manual maintenance pipeline that (re)builds the CI Docker images. Generated
 # from dockerfiles/ci/*/docker-compose.yml + .env so versions live in one place.
 # No strategy: depend — the parent must not wait on these manual jobs.
+#
+# This job MUST remain available on development branches: image changes are
+# built and validated from their branch before merge. Do not restrict it with
+# CI_COMMIT_REF_PROTECTED or CI_DEFAULT_BRANCH. This pipeline runs only in the
+# private Datadog GitLab; external pull requests cannot trigger it. Access to
+# manual jobs is controlled by that GitLab project's membership and settings.
 ci-images:
   stage: ci-build
   rules:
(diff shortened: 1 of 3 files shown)


bwoebi:
I'm slightly confused. Why would we have even get that idea of restricting jobs here? Was there some issue where that broke?

realFlowControl:
> I'm slightly confused. Why would we have even get that idea of restricting jobs here? Was there some issue where that broke?

Likewise and I am not really happy with these comments as well.
Look at https://datadoghq.atlassian.net/browse/APMSP-3852
This is the second time I got pinged with a high severity issue on the same situation. I want it to stop, but apperantly Codex is not aware that our GitLab-CI jobs are not triggered by 3rd party contributors and also those jobs are manual jobs.

Not 100% sure what the correct solution is.

bwoebi:
I would just discard these jira tickets. We have to accept that AI-produced output can be garbage and just ignored.

PR #4068, opened 2026-07-27
perf(profiling): avoid FFI call on allocation hot path

realFlowControl:
### Description

Read `executor_globals.current_execute_data` directly on NTS builds instead of calling through the C FFI wrapper for every allocation and reallocation. Keep the existing wrapper on ZTS builds.

A 60-second macOS sample reduced `get_current_execute_data` self time from 0.412% (207/50,210 main-thread samples) to 0.002% (1/49,737). Known allocation-hook self time fell from 4.184% to 3.601%, a 13.9% relative reduction. Repeated end-to-end throughput remained within system scheduling noise.

### Reviewer checklist
- [ ] Test coverage seems ok.

https://datadoghq.atlassian.net/browse/PROF-15505


bwoebi (approved):
Looks good. Now we need to also inline the ZTS path :-)

PR #4071, opened 2026-07-27
fix(feature-flags): make evaluation logging non-throwing

leoromanovsky:
## Motivation

A customer found that successful OpenFeature evaluations returned the code default in Laravel. Laravel converts the Datadog warning into `ErrorException`. OpenFeature catches that exception as `GENERAL` and discards the evaluated value.

Existing FFE system tests called the native Datadog client. That path bypassed the OpenFeature provider, its logger boundary, and framework error handling. [DataDog/system-tests#7393](https://github.com/DataDog/system-tests/pull/7393) makes OpenFeature the default `/ffe` path and adds that path to Laravel and Symfony. PHP 8 must return the configured value without an error. A runtime without the provider must return `PROVIDER_NOT_READY`. This contract makes future regressions fail in the framework path.

[PR #3906](https://github.com/DataDog/dd-trace-php/pull/3906) added this warning as a temporary guardrail. [The reviewer asked the implementation to reuse `LoggerInterface`](https://github.com/DataDog/dd-trace-php/pull/3906#discussion_r3308276556) and suggested `TriggerErrorLogger` only if `trigger_error()` remained the default. Exposure delivery and evaluation metrics are now complete, so the guardrail and its logger are obsolete.

## Changes and Decisions

Successful native evaluations now report `productionRuntime=true` and `reason=ready`. Only `PROVIDER_NOT_READY` results produce the not-ready warning.

The change removes `TriggerErrorLogger` and uses the existing tracer logger. Both feature-flag entry points wrap their logger with `NonThrowingLogger`. This boundary prevents current and future logger failures from changing evaluation results. The provider retains `LoggerInterface` and uses `datadogLogger` because the OpenFeature parent class owns `$logger`.

## Validation

[DataDog/system-tests#7393](https://github.com/DataDog/system-tests/pull/7393) passed locally with system-tests commit `0dad5d48bec63cffdd206b724013d26f497c0c8c` and tracer commit `bf360ef80839f483179e37215536d1713116167c`:

- PHP-FPM 8.2: 8 passed in 52.33 seconds.
- Laravel 11: 1 passed in 34.21 seconds.
- Symfony 7: 1 passed in 34.15 seconds.


src/DDTrace/OpenFeature/DataDogProvider.php
```diff
@@ -30,7 +31,7 @@ final class DataDogProvider extends AbstractProvider
     protected static string $NAME = 'Datadog';
 
     private Evaluator $evaluator;
-    private LoggerInterface $warningLogger;
+    private NonThrowingLogger $safeLogger;
```
leoromanovsky: * shouldn't this interface still be the same?
* why is it even called "warning logger" instead of "logger" to begin with?

src/api/FeatureFlags/Client.php
```diff
@@ -19,7 +21,7 @@ public function __construct($logger = null)
         }
 
         $this->evaluator = NativeEvaluator::create();
-        $this->logger = $logger ?: new TriggerErrorLogger();
+        $this->logger = new NonThrowingLogger($logger ?: GlobalLogger::get());
```
pavlokhrebto, line 24: Behavior change worth calling out: `Logger::get()` returns `NullLogger(EMERGENCY)` unless `DD_TRACE_DEBUG` is set, so the not-ready warning is now effectively **silent by default** (previously `TriggerErrorLogger` always emitted `E_USER_WARNING`). 

As far as i remember this is intended, but please note in the PR that not-ready state is now surfaced via exposures/metrics rather than a userland warning. Same applies to the equivalent line in `DataDogProvider`.

src/api/FeatureFlags/Internal/NativeEvaluator.php
```diff
@@ -103,9 +103,9 @@ private function withProviderState($rawResult)
             'ready' => $hasConfig,
             'hasConfig' => $hasConfig,
             'configVersion' => $configVersion,
-            'productionRuntime' => false,
+            'productionRuntime' => true,
```
pavlokhrebto, line 106: After this PR nothing reads `productionRuntime` for logic (the old `warnIfNonProductionRuntime` checks are removed); it is only surfaced as metadata / asserted in the .phpt. Two small things: (a) consider `=> $hasConfig` instead of a constant `true` — a config-missing result (`reason='configuration_missing'` -> `PROVIDER_NOT_READY`) currently reports `productionRuntime=true` alongside a not-ready error, which is contradictory; (b) a one-line comment that it is informational-only would help the next reader.

src/api/Log/NonThrowingLogger.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace DDTrace\Log;
+
+/**
+ * Prevents logger failures from escaping into application code.
```
pavlokhrebto, line 6: Nice boundary. Note it swallows all Throwables silently on every method, so a genuinely broken/misconfigured underlying logger becomes invisible. Acceptable trade-off here (and I would not add an error_log fallback — recursion risk), just flagging the blind spot.

tests/ext/ffe/system_test_data_evaluate.phpt
```diff
@@ -153,6 +155,16 @@ function run_fixture_case($client, $fileName, $index, array $case, array &$failu
         $context
     );
 
+    $providerState = $details->getProviderState();
+    if (($providerState['productionRuntime'] ?? null) !== true) {
```
pavlokhrebto, line 159: This asserts the `hasConfig=true` path (`productionRuntime=true`/`reason=ready`). The `hasConfig=false` native branch — config-missing -> `PROVIDER_NOT_READY` with a returned value, which is exactly the scenario the warning re-keying governs — is not covered end-to-end. Consider a case that exercises it to lock the contract (the unit tests cover `UnavailableEvaluator`'s PROVIDER_NOT_READY, but not native config-missing).

pavlokhrebto (commented):
Reviewed the diff — solid, well-tested fix for a real footgun (trigger_error -> ErrorException -> OpenFeature discards the value). A few non-blocking notes inline. The two I would most want confirmed before approval: (1) the not-ready warning is now effectively silent by default (NullLogger unless DD_TRACE_DEBUG), and (2) light end-to-end coverage of the native config-missing branch. No interaction with the merged span-enrichment work.

pavlokhrebto (approved):

tabgok (approved):
looks good from the IDM perspective

bwoebi (approved):

chasdevs (commented):
Missed the boat on approval, but that is pretty wild.

PR #4074, opened 2026-07-28
perf(prof): speed up interrupt_count checks

realFlowControl:
### Description

Speed up `interrupt_count` checks by making it `Relaxed` and moving it to the `ProfilerGlobals` instead of using `REQUEST_LOCALS`. The biggest change comes from avoiding `REQUEST_LOCALS`, which in this case is totally safe because `interrupt_count` is atomic.

Notably, part of the interrupt handler checks can be a bit hot in certain cases where we don't expect there to be an interrupt very often:
 - PHP <=8.3: `execute_internal` wrapper
 - PHP 8.4+: frameless functions 

So we want the "0 interrupt count" check to be as fast and as early as possible.

This also removes the `profiling_enabled` check of the interrupt function. This is unnecessary because:

 1. We do not set interrupts when disabled to begin with, generally. `trigger_time_sample` can still trigger but this is test only.
 2. For edge cases having a pending interrupt going into a fork, the child will call `Profiler::kill()` which will cause `Profiler::get()` to return None.
 
 These changes were inspired by and directly derived from iterating on Florian's original proposal (see Original description below). My (Levi) own benchmarks are similar to his (I hope), here is throughput:

 - Mean: 22,773,158/s → 24,067,515/s (+5.68%)
 - Median: 22,810,705/s → 23,913,084/s (+4.83%)
 - 95% Confidence interval: +1.09% to +10.44%

<details>
<summary>Original description</summary>

Profiler interrupt checks run after internal, traced, and frameless calls even though a profiler interrupt is usually not pending. Export an `unlikely` entry point that reads the profiler-owned per-thread `interrupt_count` with `Ordering::Relaxed` and returns immediately when it is zero; only a pending interrupt reaches the authoritative `SeqCst` swap and stack collection. The regular Zend VM interrupt callback remains unchanged.

Move `interrupt_count` from `REQUEST_LOCALS` to PHP module globals so the hot path avoids borrowing request locals. The PHP 8.3-and-older internal-function hook and tracer hook call sites use the new entry point; PHP 8.4+ frameless handling uses the same relaxed precheck. Tracer symbol lookup falls back to the old entry point for compatibility with older profiler versions.

### Benchmark

Fresh release builds from `master` (`24c420794`) and this branch (`32abe928a`), using six balanced 60-second runs per binary on PHP 8.3 ZTS, macOS ARM64:

- Mean throughput: 32,155,879/s to 34,645,984/s (**+7.74%**)
- Median throughput: 32,308,450/s to 35,053,828/s (**+8.50%**)
- Every paired run improved: +5.72% to +9.10%
- Fresh 70-second native samples reduced interrupt entry-point top-of-stack samples from 4,335 to 1,096 (**-74.7% relative**)

Wall-time correctness on a four-second CPU/sleep workload remained stable:

- Master: 4.00s total; `sleep()`: 2.01s (50.37%)
- Candidate: 4.01s total; `sleep()`: 2.012s (50.16%)

</details>

### Reviewer checklist
- [ ] Test coverage seems ok.

https://datadoghq.atlassian.net/browse/PROF-15506


morrisonlevi (commented):
Early on in `ddog_php_prof_interrupt_function` we check the interrupt count being greater-than 0. I am surprised this helped so much. I'm going to play with it a bit locally.

morrisonlevi:
I could not reproduce such a significant gain! Florian, will you pull my changes and see what those give on your benchmarks? My changes performed more consistently for me locally, and keep the same spirit (fast return if there isn't a pending interrupt).

The reason `ddog_php_prof_interrupt_function_unlikely` is faster for its callers than `ddog_php_prof_interrupt_function` is because it uses a `load(Relaxed)` of the `interrupt_count`. This avoids a write synchronization that the `store(0, SeqCst)` version incurs _even_ if there wasn't an interrupt. The `EG(vm_interrupt)` check you did would do the same, but we don't actually care about `EG(vm_interrupt)`, only `interrupt_count`.

I also move the interrupt_count into ProfilerGlobals and access SystemSettings through the global so we don't have to use REQUEST_LOCALS at all on the hot path here.

 Nine pinned 60-second runs each, PHP 8.3 ZTS on my benchmark:

 - Mean throughput: 23,174,141/s → 24,435,029/s (+5.44%)
 - Median throughput: 23,200,257/s → 24,290,987/s (+4.70%)
 - Paired improvement: +5.51%
 - 95% CI: +2.08% to +8.95%

profiling/src/wall_time.rs
```diff
@@ -109,42 +110,78 @@ static mut PREV_INTERRUPT_FUNCTION: Option<VmInterruptFn> = None;
...
+    // more expensive atomic read-modify-write in `interrupt_function` on
+    // the idle path. It is only a fast-path hint: `interrupt_function` does
+    // the authoritative swap. If another consumer clears the count between
+    // the load and swap, the swap simply observes zero; if a producer adds
+    // an interrupt after this load observes zero, the count remains pending
+    // and the corresponding VM interrupt will provide another opportunity
+    // to handle it.
+    if globals.interrupt_count.load(Ordering::Relaxed) == 0 {
+        return;
+    }
```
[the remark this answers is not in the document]
morrisonlevi: True, but not particularly consequential. What if the interrupt happened literally the next time unit _after_ an authoritative atomic operation? It will get delayed until the next check. In the worst case, this will delay to the next VM interrupt. There's not much of a difference between these two cases--it is non-deterministic by design.

With that said, I have moved it to a `.swap(0, Relaxed)` which is an atomic read-modify-write.

morrisonlevi (approved):
I approve of this PR--I basically rewrote it! Florian, please review. I made sure to preserve your commit history. Without your work here this wouldn't haven't happened. Also, some of the TSRM stuff was pulled from another one of your PRs, so you do have code in here still despite the reverted commits!

profiling/src/wall_time.rs
```diff
@@ -109,31 +109,29 @@ static mut PREV_INTERRUPT_FUNCTION: Option<VmInterruptFn> = None;
...
-            return;
-        }
-
-        if let Some(profiler) = Profiler::get() {
-            // Safety: execute_data was provided by the engine, and the profiler doesn't mutate it.
-            profiler.collect_time(execute_data, interrupt_count);
-        }
-    });
+    // SAFETY: interrupt callbacks run while the current PHP thread's module globals are valid.
+    let atomic_count = unsafe { &(*module_globals::get_profiler_globals()).interrupt_count };
```
realFlowControl, line 113: Just thinking loud: We do not follow `fork()` currently and by not checking on `profiling_enabled` anymore, could we end up in a situation where after `fork()` the child has a raised `EG(vm_interrupt)` and we have an `interrupt_count` > 0?

Actually this is not new, it is like this on `master` already and OTOH we call `collect_time_if_enabled()` which does `Profiler::get()` which returns `None` in a forked child.

So I think we are good.

profiling/src/module_globals.rs
```diff
@@ -51,6 +63,27 @@ mod zts {
     }
 }
 
+#[cfg(php_zts)]
+#[inline]
+pub unsafe fn get_tsrm_ls_cache() -> *mut c_void {
+    zts::get_ls_cache()
+}
```
realFlowControl, line 70: NIT: This has only one call site, could we just fold it into the `get_profiler_globals_from_cache(zts::get_ls_cache())`?

profiling/src/module_globals.rs
```diff
@@ -51,6 +63,27 @@ mod zts {
...
+
+#[cfg(php_zts)]
+#[inline]
+pub unsafe fn get_tsrm_resource_from_cache(ls_cache: *mut c_void, id: i32) -> *mut c_void {
+    zts::tsrmg_bulk(ls_cache, id)
+}
```
realFlowControl, line 76: NIT: Also has just one call site and could be folded into `get_profiler_globals_from_cache()`.

realFlowControl (commented):
I can't formally approve this PR as the author 😉
But here we are: ✅

About the two nitpicks: I am pretty sure the compiler will just inline them, so feel free to ignore. One upside (that I am not sure about) could be that in a debug build we might actually not inline and have the symbol available.

profiling/src/wall_time.rs
morrisonlevi: Yes, it's theoretically possible to have a pending interrupt going into a fork. On master there are two checks, the `profiling_enabled` check and also the `Profiler::get()` check. This branch removes the `profiling_enabled` check but it still has `Profiler::get()` which returns `None`. In such cases, the forked thread's module globals are still memory-safe to access, so everything is fine here.

realFlowControl: Thanks for confirming

PR #4077, opened 2026-07-29
feat: publish Linux OTel contexts

morrisonlevi:
[PROF-15487](https://datadoghq.atlassian.net/browse/PROF-15487)

### Description

Publish standard OTel process and thread contexts from the tracer and make the profiler consume them for runtime identity and effective service metadata. Handle span, stack, Fiber, configuration, and fork lifecycle changes while retaining the legacy non-Linux path.

WIP for appsec.

### Reviewer checklist
- [ ] Test coverage seems ok.

[PROF-15487]: https://datadoghq.atlassian.net/browse/PROF-15487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


profiling/src/process_context/linux.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    pub(crate) fn reset(&mut self) {
+        *self = Self::new();
+    }
+
+    /// Reads Process Context on this PHP thread's first request. A failed read
+    /// installs an empty context so later requests do not repeat discovery.
+    pub(crate) fn initialize(&mut self) {
+        if self.context.is_some() {
+            return;
```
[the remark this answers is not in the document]
morrisonlevi: We never write service, env, and version to the process context. There's no stale identity.

profiling/src/profiling/mod.rs
```diff
@@ -1738,26 +1824,64 @@ impl Profiler {
...
+                    environment: context.environment,
+                    version: context.version,
+                },
+            ),
+            ThreadContextRead::Inactive => (
+                unsafe { libc::pthread_self() as i64 },
+                0,
+                0,
+                ProcessIdentity::default(),
+            ),
```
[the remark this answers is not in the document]
morrisonlevi: The tracer and profiler are meant to be upgraded in tandem. They are not designed to be upgraded piecemeal and in the future this will be impossible when the extensions are merged.

components-rs/Cargo.toml
```diff
@@ -56,7 +56,12 @@ libc = "0.2"
 bincode = { version = "1.3.3" }
 hashbrown = "0.15"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+libdd-library-config = { path = "../libdatadog/libdd-library-config", default-features = false, features = ["otel-thread-ctx"] }
+libdd-otel-thread-ctx = { path = "../libdatadog/libdd-otel-thread-ctx" }
```
bwoebi: Do I see it correctly that the sole usage of this crate is essentially pinning the struct? Including some build.rs hacks?

Wouldn't it be simpler to just declare `otel_thread_ctx_v1` without `extern` in tracer/otel_context.c and be done with it?

ext/otel_context.c
```diff
@@ -0,0 +1,31 @@
...
+    if (gethostname(detected_hostname, HOST_NAME_MAX) == 0) {
+        hostname = (ddog_CharSlice){
+            .ptr = detected_hostname,
+            .len = strnlen(detected_hostname, HOST_NAME_MAX),
+        };
+    }
+
+    zend_string *process_tags = datadog_process_tags_get_serialized();
+    datadog_publish_otel_process_context(
+        (ddog_CharSlice){.ptr = (const char *)runtime_id, .len = sizeof(runtime_id)},
```
bwoebi: Why shuffle the runtime-id around? Just directly generate in the rust code of datadog_publish_otel_process_context.

bwoebi: Same for ddtrace_get_container_id.

profiling/src/process_context/linux.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    fn extra_string(&self, offset: Option<usize>) -> Option<&str> {
+        offset
+            .and_then(|offset| self.context.extra_attributes.get(offset))
+            .and_then(string_value)
+    }
+
+    fn identity(&self) -> ProcessIdentity {
+        ProcessIdentity {
```
bwoebi: Why not `&'a self -> ProcessIdentity<'a>` and reference `&'a str` in ProcessIdentity struct? For the other *_string fns here you do the same.

bwoebi: Oh I see you have `ProcessIdentifyRef` doing that - the getter should probably just return the Ref, remapping should be a `From<>` impl or such.

profiling/src/profiling/mod.rs
```diff
@@ -166,6 +170,85 @@ pub struct Label {
     pub value: LabelValue,
 }
 
+struct SampleLabels {
+    labels: Vec<Label>,
+    #[cfg(target_os = "linux")]
```
bwoebi: Why is this linux-only? I thought we'd fallback to the dlsym read on non-linux?

profiling/src/process_context/linux.rs
```diff
@@ -0,0 +1,583 @@
...
+    fn decode_with_refresh(
+        &mut self,
+        attributes: &[u8],
+        defaults: ProcessIdentityRef<'_>,
+    ) -> ThreadContext {
+        let (decoded, unknown_index) = self.decode_thread_attributes(attributes, defaults);
+        if !unknown_index {
+            return decoded;
+        }
+
```
bwoebi: that seems redundant to me, you anyway try decode_thread_attributes on caller site before calling decode_with_refresh. (and this code is not like it would need to handle race conditions - it's on same thread.)

bwoebi: I'd propose inlining the remainder of this function at caller site to make it more obvious and remove this function.

tracer/otel_context.c
```diff
@@ -0,0 +1,222 @@
...
+
+    ddtrace_otel_record_set_trace_flags(&root->otel_context, root);
+}
+
+void ddtrace_otel_update_attribute_values(ddtrace_root_span_data *root) {
+    if (!root || !get_DD_TRACE_ENABLED()) {
+        return;
+    }
+
+    if (++DDTRACE_G(otel_context_attributes_generation) == 0) {
```
bwoebi: Is this ever -1, or what is this checking for?

tracer/ddtrace.c
```diff
@@ -692,6 +706,13 @@ void ddtrace_internal_handle_postfork() {
 }
 
 void ddtrace_internal_handle_fork() {
+#ifdef __linux__
+    ddtrace_span_stack *stack = DDTRACE_G(active_stack);
+    if (stack && stack->root_span) {
+        ddtrace_otel_update_attribute_values(stack->root_span);
+    }
```
bwoebi:
```suggestion
```
Just detach, what's that for? the ddtrace_free_span_stacks a bit later will unconditionally free all in-flight spans.

tracer/distributed_tracing_headers.c
```diff
@@ -349,15 +353,18 @@ static ddtrace_distributed_tracing_result ddtrace_read_distributed_tracing_ids_t
...
 
-        zend_string_release(traceparent);
-
         if ((!trace_id.low && !trace_id.high) || !parent_id) {
+            zend_string_release(traceparent);
             return result;
         }
 
+        result.trace_flags = (uint8_t)ddtrace_parse_hex_span_id_str(tracedata->trace_flags, 2)
+            & DDTRACE_TRACE_FLAGS_SUPPORTED;
```
bwoebi: Moving this line a bit higher doesn't do harm, just trace_id is used for some conditions, which is why it must happen afterwards.

tracer/functions.c
```diff
@@ -423,6 +428,11 @@ static zend_object *ddtrace_span_stack_clone_obj(zend_object *old_obj) {
 }
 
 static void ddtrace_span_data_free_storage(zend_object *object) {
+#ifdef __linux__
+    if (object->ce == ddtrace_ce_root_span_data) {
+        ddtrace_detach_otel_thread_context_for_root(ROOTSPANDATA(object));
+    }
+#endif
```
bwoebi: I need to read more code first, but this definitely is not the correct place for this. This code is called for any freed RootSpanData, regardless of whether it's active (in fact when it's reaching this code, it's should not be active anymore).

tracer/functions.c
```diff
@@ -579,15 +596,34 @@ static zval *ddtrace_root_span_data_write(zend_object *object, zend_string *memb
...
         cache_slot = NULL;
     }
 
 #if PHP_VERSION_ID >= 70400
     zval *ret = ddtrace_span_data_readonly(object, member, value, cache_slot);
 #else
     ddtrace_span_data_readonly(object, member, value, cache_slot);
+#endif
+    if (trace_id_changed) {
+        span->trace_flags = trace_id_random ? DDTRACE_TRACE_FLAG_RANDOM : 0;
```
bwoebi: Doesn't this overwrite sampled? Shouldn't it be more like `span->trace_flags = (span->trace_flags & ~DDTRACE_TRACE_FLAG_RANDOM) | (trace_id_random ? DDTRACE_TRACE_FLAG_RANDOM : 0);`?

profiling/src/profiling/mod.rs
morrisonlevi: Yes, we can use `None` for the fields in `ProcessIdentify` when using just a `ddtrace_profiling_context` from the dlsym on non-Linux.

tracer/otel_context.c
morrisonlevi: I think the AI was just overly concerned about wrap-arounds?

tracer/otel_context.c
```diff
@@ -0,0 +1,222 @@
...
+void ddtrace_otel_attach_stack(ddtrace_span_stack *stack) {
+    if (!stack || !stack->root_span || !stack->active) {
+        ddtrace_otel_detach();
+        return;
+    }
+
+    ddtrace_root_span_data *root = stack->root_span;
+    if (UNEXPECTED(root->otel_context_attributes_generation != DDTRACE_G(otel_context_attributes_generation))) {
+        ddtrace_otel_refresh_attribute_values(root);
+    }
```
bwoebi: This is the sole place otel_context_attributes_generation gets consumed.
But why? Why do we need to call ddtrace_otel_refresh_attribute_values() in this case?

morrisonlevi: The generation check is necessary. It is a lazy invalidation mechanism for the OTel records embedded in multiple root spans.

 Each span stack/root owns a separate cached Thread Context record. Nested/manual stacks and suspended fibers can therefore hold records that are not currently attached. Service, environment, or version changes
 can affect multiple records because non-entrypoint roots inherit metadata from the entrypoint root.

 On a metadata change:

 1. The active record is refreshed.
 2. The global generation is incremented.
 3. Other records remain stale while inactive.
 4. When one is attached again, the generation mismatch causes it to refresh.

 Removing the refresh from `ddtrace_otel_attach_stack()` would expose stale metadata after switching stacks. This is covered by `tests/ext/otel_thread_context_stack_switch.phpt`, particularly switching back to
` mainRoot` after updating its metadata while `otherRoot` is active.

-----

Probably should leave a code comment here then.

tracer/span.h
```diff
@@ -121,7 +124,15 @@ static inline ddtrace_inferred_span_data *INFERRED_SPANDATA(zend_object *obj) {
...
     datadog_trace_id trace_id;
     uint64_t parent_id;
+    uint8_t trace_flags;
     ddtrace_rule_result sampling_rule;
+#ifdef __linux__
+    datadog_otel_thr_ctx_rec otel_context;
+    uint64_t otel_context_attributes_generation;
+    // Address of otel_thread_ctx_v1 for this root's thread, resolved at root creation to avoid a dynamic TLS lookup
+    // on every stack switch.
+    void **otel_context_slot;
```
bwoebi, line 137: Is this actually that expensive? Any DDTRACE_G() lookup will basically do the same, I hope the __thread dispatch table is in CPU cache at all times.
I would just remove this.

tracer/otel_context.c
```diff
@@ -0,0 +1,222 @@
...
+
+    record->attrs_data[0] = DDTRACE_OTEL_ATTR_LOCAL_ROOT_SPAN_ID;
+    record->attrs_data[1] = 16;
+    for (uint8_t i = 0; i < 16; ++i) {
+        record->attrs_data[2 + i] = hex_digits[(root->span_id >> ((15 - i) * 4)) & 0xf];
+    }
+
+    zend_string *service = NULL, *env = NULL, *version = NULL;
+    ddtrace_span_data *source = ddtrace_otel_attr_source_span(root);
+    datadog_populate_target_data_with_defaults(source, &service, &env, &version, get_DD_SERVICE(), get_DD_ENV(), get_DD_VERSION());
```
bwoebi: Why do we call this here?

tracer/span.h
morrisonlevi: I actually benchmarked this before making the commit. I had the AI regenerate assembly to show the difference:

-----
 The optimization was introduced by e864a7b6b. I reproduced the relevant assembly using a default-visible global-dynamic TLS symbol in an LLD-linked shared library.

 ##### AMD64

 Direct TLS assignment:

 ```asm
   attach_direct:
       push   %rbx
       mov    %rdi, %rbx
       lea    otel_thread_ctx_v1@TLSGD(%rip), %rdi
       call   __tls_get_addr@plt
       mov    %rbx, (%rax)
       pop    %rbx
       ret
 ```

 Cached slot assignment:

 ```asm
   attach_cached:
       mov    %rsi, (%rdi)
       ret
 ```

 ##### ARM64

 Direct TLS assignment:

 ```asm
   attach_direct:
       stp     x29, x30, [sp, #-16]!
       mov     x29, sp
       mov     x8, x0
       adrp    x0, <TLS descriptor page>
       ldr     x1, [x0, <resolver offset>]
       add     x0, x0, <descriptor offset>
       blr     x1
       mrs     x9, tpidr_el0
       str     x8, [x9, x0]
       ldp     x29, x30, [sp], #16
       ret
 ```

 Cached slot assignment:

 ```asm
   attach_cached:
       str     x1, [x0]
       ret
 ```

morrisonlevi: On "regular" code this will not matter at all, but it might for fiber heavy code.

tracer/otel_context.c
```diff
@@ -0,0 +1,222 @@
...
+
+    size_t offset = DDTRACE_OTEL_LOCAL_ROOT_SPAN_ID_ATTR_SIZE;
+    offset = ddtrace_otel_record_write_attr_zstr(record, offset, DDTRACE_OTEL_ATTR_SERVICE_NAME, ddtrace_otel_attr_zstr(service));
+    offset = ddtrace_otel_record_write_attr_zstr(record, offset, DDTRACE_OTEL_ATTR_DEPLOYMENT_ENVIRONMENT_NAME, ddtrace_otel_attr_zstr(env));
+    offset = ddtrace_otel_record_write_attr_zstr(record, offset, DDTRACE_OTEL_ATTR_SERVICE_VERSION, ddtrace_otel_attr_zstr(version));
+
+    char thread_id[32];
+    int thread_id_len = snprintf(thread_id, sizeof(thread_id), "%llu", (unsigned long long)syscall(SYS_gettid));
```
bwoebi, line 207: should we cache the tid vs doing a syscall on the write?

tracer/otel_context.c
bwoebi: But then it should only increment then when the actual entrypoint root span is updated, and not any root span in ddtrace_otel_update_attribute_values.

tracer/otel_context.c
morrisonlevi: It's needed but it can be simplified to use this helper `datadog_populate_target_data` which already exists:

```c
static inline void datadog_populate_target_data(ddtrace_span_data *span, zend_string **service, zend_string **env, zend_string **version) {
    datadog_populate_target_data_with_defaults(span, service, env, version, get_DD_SERVICE(), get_DD_ENV(), get_DD_VERSION());
}
```

AI explanation below.

-----

This call resolves the effective Universal Service Monitoring identity written into the fixed-size OTel Thread Context record:

 ```c
   datadog_populate_target_data_with_defaults(...)
 ```

 It is needed because:

 - service.name, deployment environment, and service version may come from span properties or tracer configuration.
 - A nested local root should inherit these attributes from the entrypoint root rather than publishing its own integration service. This is why ddtrace_otel_attr_source_span() is used.
 - There may be no applicable entrypoint span, in which case configured/default values are required.
 - The record must contain copied bytes because an external profiler cannot safely dereference Zend strings.

 This behavior is visible in otel_thread_context_stack_switch.phpt: otherRoot has other-service, but its Thread Context intentionally publishes main-service.

tracer/otel_context.c
bwoebi: Also, should we attach simply the actual entrypoint root span to root_span_data as pointer? And compare the local generation vs the generation on the entrypoint root span? Instead of having a thread global generation counter, which invalidates much more than it needs to.

tracer/span.c
```diff
@@ -1085,6 +1125,13 @@ void ddtrace_drop_span(ddtrace_span_data *span) {
     } else {
         ZVAL_NULL(&stack->property_active);
     }
+#ifdef __linux__
+    if (stack->active) {
+        ddtrace_update_otel_thread_context_span_id(SPANDATA(stack->active));
+    } else {
+        ddtrace_otel_detach();
+    }
+#endif
```
bwoebi: This code should not be executed when a ddtrace_switch_span_stack happens later, otherwise it will be transiently detached before it switches to a parent stack.

```
    if (&stack->root_span->span == span) {
        ddtrace_switch_span_stack(stack->parent_stack);
        stack->root_span = NULL;
    } else if (!stack->active || SPANDATA(stack->active)->stack != stack) {
        dd_close_entry_span_of_stack(stack);
    } else {
#ifdef __linux__
        if (stack->active) {
            ddtrace_update_otel_thread_context_span_id(SPANDATA(stack->active));
        } else {
            ddtrace_otel_detach();
        }
#endif
    }
```

like this. Closing suffers from the same issue. dd_close_entry_span_of_stack also needs a detach if it's not switching spans.

tracer/functions.c
bwoebi: I think this can just be outright removed. Existing code handles this already.

components-rs/lib.rs
```diff
@@ -90,6 +90,58 @@ pub extern "C" fn datadog_format_runtime_id(buf: &mut [u8; 36]) {
...
+    // resource keys must be omitted from the process-wide context. The keys remain discoverable
+    // through threadlocal.attribute_key_map.
+    if let Some(resource) = context.resource.as_mut() {
+        resource.attributes.retain(|attribute| {
+            !matches!(
+                attribute.key.as_str(),
+                "service.name" | "service.version" | "deployment.environment.name"
+            )
+        });
+    }
```
cataphract: would be better to change libdatadog not to add these when it's not provided

appsec/cmake/ddtrace.cmake
```diff
@@ -26,9 +26,9 @@ add_custom_target(libdatadog_stamp
...
 set(EXPORTS_FILE "${CMAKE_BINARY_DIR}/ddtrace_exports.version")
 add_custom_target(ddtrace_exports
-    COMMAND bash -c "{ echo -e '{\\nglobal:'; sed 's/$/;/' '${CMAKE_SOURCE_DIR}'/../datadog.sym; echo -e 'local:\\n*;\\n};'; } > '${EXPORTS_FILE}'"
+    COMMAND bash -c "{ echo -e '{\\nglobal:'; sed 's/$/;/' '${CMAKE_SOURCE_DIR}'/../datadog-linux.sym; echo -e 'local:\\n*;\\n};'; } > '${EXPORTS_FILE}'"
     BYPRODUCTS ${EXPORTS_FILE}
-    DEPENDS ${CMAKE_SOURCE_DIR}/../datadog.sym
+    DEPENDS ${CMAKE_SOURCE_DIR}/../datadog-linux.sym
     VERBATIM
 )
 elseif(APPLE)
```
cataphract, line 40: I don't think this will build on mac os as is, as it appears to include linux only sources.

And the linux part doesn't add `-mtls-dialect=gnu2`.

But not critical as I promised the cmake ddtrace build would not be a maintenance burden :p

ext/sidecar.c
```diff
@@ -469,7 +469,11 @@ void datadog_sidecar_handle_fork(void) {
         return;
     }
 
-    datadog_force_new_instance_id();
+    if (datadog_sidecar_instance_id) {
+        ddog_sidecar_instanceId_drop(datadog_sidecar_instance_id);
+        datadog_sidecar_instance_id = NULL;
+    }
+    dd_set_resettable_sidecar_globals();
```
cataphract: After this, `datadog_force_new_instance_id` becomes dead code (only called from a test function). Why not just make `datadog_force_new_instance_id` unconditional and call it in the beginning of `datadog_internal_handle_fork` ?

profiling/src/process_context/linux.rs
```diff
@@ -0,0 +1,565 @@
...
+        if self.context.is_some() {
+            return;
+        }
+        if self.refresh().is_err() {
+            self.context = Some(CachedProcessContext::default());
+        }
+    }
+
+    fn refresh(&mut self) -> std::io::Result<()> {
+        let result = ProcessContextSelfReader::new().and_then(|reader| reader.read());
```
cataphract, line 214: It should not be necessary to create a new `ProcessContextSelfReader` unless there was a fork or unpublish + publish. To do so, incurs the penalty of having to do discovery again, and calling `pipe2` again.

profiling/src/process_context/linux.rs
```diff
@@ -0,0 +1,565 @@
...
+            return decoded;
+        }
+        drop(cache);
+
+        let Ok(mut cache) = cell.try_borrow_mut() else {
+            return decoded;
+        };
+        if cache.refresh().is_err() {
+            return decoded;
+        }
```
cataphract, line 549: If this starts failsing we presumably keep getting `unknown_index == true`, and we try again, with discovery and everything. I don't think it should retry indefinitely.

profiling/src/process_context/linux.rs
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    let attributes = unsafe {
+        std::slice::from_raw_parts(record.add(THREAD_CONTEXT_HEADER_SIZE), attributes_size)
+    };
+    let mut context = with_cache(|cell| {
+        let Ok(cache) = cell.try_borrow() else {
+            return ThreadContext::default();
+        };
+        let (decoded, unknown_index) = cache.decode_thread_attributes(attributes, defaults);
```
cataphract, line 538: Have you considering caching also the decoded `ThreadContext`. Presumably the cache would be hit most of the time.

profiling/src/profiling/mod.rs
```diff
@@ -166,6 +171,82 @@ pub struct Label {
...
+    }
+}
+
+fn effective_profile_tags(base: Arc<Vec<Tag>>, identity: &ProcessIdentity) -> Arc<Vec<Tag>> {
+    let mut updated = None;
+    apply_profile_tag_override(&base, &mut updated, "service", identity.service.as_deref());
+    apply_profile_tag_override(&base, &mut updated, "env", identity.environment.as_deref());
+    apply_profile_tag_override(&base, &mut updated, "version", identity.version.as_deref());
+    updated.map(Arc::new).unwrap_or(base)
+}
```
morrisonlevi: TODO: I think the performance of this function is going to be a bit poor. Should probably not emit the service/env/version tags and "merge" things at a higher level, rather than updating tags.

profiling/src/process_context/linux.rs
morrisonlevi: It's not particularly performance sensitive. There are two call sites:

 1. In RINIT, it's called from `initialize` if there isn't a cached process context already. You can think of this as once per thread.
 2. If there is an unknown thread attribute key, then we refresh the process context. In practice this should not happen, all of our keys are static at the moment. The exception is after a fork, as it gets invalidated during the pre-fork (this means that the parent also is invalidated).
 
Doing it this way means we don't hold open 2 file descriptors per thread. The thing cannot be shared across threads either.

In other words, aside from bugs, it's already almost optimal. If customers have fork-heavy code, we may want to revise this pre-fork part though.

profiling/src/process_context/linux.rs
morrisonlevi: I addressed this in https://github.com/DataDog/dd-trace-php/pull/4077/commits/2c1a93d6a722c095fb141a29ed30430c768273fa, give it a look!

ext/otel_context.c
```diff
@@ -0,0 +1,23 @@
...
+
+#include "datadog.h"
+#include "ffi_utils.h"
+#include "process_tags.h"
+#include <components-rs/datadog.h>
+
+void datadog_otel_process_context_publish(void) {
+    char detected_hostname[HOST_NAME_MAX + 1] = {0};
+    ddog_CharSlice hostname = {0};
+    if (gethostname(detected_hostname, HOST_NAME_MAX) == 0) {
```
cataphract: >  The gethostname() function returns the standard host name for the current processor, as previously set by sethostname().  The namelen argument specifies the size of the name array.  The returned name is null-terminated, unless insufficient space is provided.
> Host names are limited in length to {sysconf(_SC_HOST_NAME_MAX)} characters, not including the trailing null, currently 255.

So you should pass `sizeof(detected_hostname)` / `HOST_NAME_MAX + 1` to `gethostname`.

profiling/src/php_ffi.c
```diff
@@ -166,6 +186,9 @@ void datadog_php_profiling_startup(zend_extension *extension) {
     for (const zend_llist_element *item = list->head; item; item = item->next) {
         const zend_extension *maybe_ddtrace = (zend_extension *)item->data;
         if (maybe_ddtrace != extension && is_ddtrace_extension(maybe_ddtrace)) {
+#ifdef __linux__
+            datadog_php_profiling_ddtrace_handle = maybe_ddtrace->handle;
```
cataphract: I don't think the ssi loader sets this, which would cause `datadog_php_profiling_get_otel_thread_context` to return early.

profiling/src/process_context/linux.rs
cataphract: Actually the decoding is not too bad, and in the absence of a version fetching and validating the cache would probably not be worth it. Still, since this is in the hot path, it would be better to avoid allocations altogether here and return

```
  struct ThreadContext<'a> {
      service: Option<&'a str>,
      environment: Option<&'a str>,
      version: Option<&'a str>,
      // numeric fields...
  }
  ```

(To satisfy the borrow checker, you also need to lifetime-parameterize the callers and change `crate::process_context::with_thread_context()` to take a `do_with_thread_context(tcr: ThreadContextRead<'_>)` rather than just returning it.)

And if I read the code correctly you also recalculate the effective tags (more allocations) if there thread-local overrides. You could also have a thread-local cache for this that recalculate only if there are changes in the base/overrides

Comment (cataphract) [reply profiling/src/process_context/linux.rs:538]: Diff: https://gist.github.com/cataphract/bbf58dd0d8a959c0ee4eace88e14e27e

File: ext/otel_context.c (hunk 25 above)
Comment (morrisonlevi) [reply ext/otel_context.c]: I also pushed the hostname stuff into `datadog_publish_otel_process_context` so there's just 1 FFI function instead of two with the same name just transposed:

- `datadog_publish_otel_process_context`
- `datadog_otel_process_context_publish`

File: profiling/src/php_ffi.c (hunk 26 above)
Comment (morrisonlevi) [reply profiling/src/php_ffi.c]: It's worse than I realized: the profiler should query based on process and thread context availability in the whole process, not just based on if the tracer was loaded, to handle OTel SDK as well. So I've reworked this code.

File: profiling/src/profiling/mod.rs (hunk 24 above)
Comment (bwoebi) [reply profiling/src/profiling/mod.rs]: Gustavos https://gist.github.com/cataphract/bbf58dd0d8a959c0ee4eace88e14e27e solves this as well.

File: appsec/cmake/ddtrace.cmake (hunk 19 above)
Comment (morrisonlevi) [reply appsec/cmake/ddtrace.cmake:40]: What command in what images do you use to build this stuff?

File: components-rs/lib.rs (hunk 18 above)
Comment (morrisonlevi) [reply components-rs/lib.rs]: PR for it: DataDog/libdatadog#2337

File: appsec/cmake/ddtrace.cmake (hunk 19 above)
Comment (cataphract) [reply appsec/cmake/ddtrace.cmake:40]:
```
mkdir -p appsec/Debug && cd appsec/Debug
cmake -DCMAKE_BUILD_TYPE=Debug -DDD_APPSEC_DDTRACE_ALT=ON ..
make -j ddtrace
```

File: profiling/src/profiling/mod.rs (hunk 24 above)
Comment (morrisonlevi) [reply profiling/src/profiling/mod.rs]: The performance difference of this patch was not very strong for me. I tried another version which is a bit more code but creates a specialized UnifiedServiceTagging struct so there aren't 3 allocations, just one, and by working tag/labels a bit. Showed 8.5% improvement on a local benchmark.

File: appsec/cmake/ddtrace.cmake (hunk 19 above)
Comment (morrisonlevi) [reply appsec/cmake/ddtrace.cmake:40]: Think I've fixed this!

File: tracer/otel_context.c [hunk 27]
```diff
@@ -0,0 +1,228 @@
...
+    datadog_otel_thr_ctx_rec *record = &root->otel_context;
+    ddtrace_otel_record_set_trace_id(record, root->trace_id);
+    ddtrace_otel_record_store_span_id(record, root->span_id);
+    ddtrace_otel_record_set_trace_flags(record, root);
+    ddtrace_root_span_data *source = ddtrace_otel_attr_source_root(root);
+    ddtrace_otel_record_set_attrs(record, root, source);
+    ddtrace_root_span_data *generation_source = source ? source : root;
+    root->otel_context_attributes_source_generation = generation_source->otel_context_attributes_generation;
+    atomic_store_explicit(&record->valid, 1, memory_order_relaxed);
+}
```
ivoanjo, line 65: Minor: Is the assumption here that prior to `ddtrace_otel_init_root_span` getting called, `&record->valid` was always 0 so we don't need the `ddtrace_otel_record_begin_update()` or equivalent?

tracer/otel_context.c
```diff
@@ -0,0 +1,228 @@
...
+    root->otel_context_slot = &otel_thread_ctx_v1;
+    datadog_otel_thr_ctx_rec *record = &root->otel_context;
+    ddtrace_otel_record_set_trace_id(record, root->trace_id);
+    ddtrace_otel_record_store_span_id(record, root->span_id);
+    ddtrace_otel_record_set_trace_flags(record, root);
+    ddtrace_root_span_data *source = ddtrace_otel_attr_source_root(root);
+    ddtrace_otel_record_set_attrs(record, root, source);
+    ddtrace_root_span_data *generation_source = source ? source : root;
+    root->otel_context_attributes_source_generation = generation_source->otel_context_attributes_generation;
+    atomic_store_explicit(&record->valid, 1, memory_order_relaxed);
```
ivoanjo: I'm not sure this is correct -- this should be `memory_order_release` not `memory_order_relaxed` because otherwise the compiler is allowed to reorder things above (setting the data) before valid is set?

(Probably calling `ddtrace_otel_record_end_update()` instead is the way to fix this?)

tracer/otel_context.c
```diff
@@ -0,0 +1,228 @@
...
+    if (UNEXPECTED(root->otel_context_attributes_source_generation != generation_source->otel_context_attributes_generation)) {
+        ddtrace_otel_refresh_attribute_values(root);
+    }
+    ddtrace_otel_attach(root);
+}
+
+void ddtrace_otel_detach(void) {
+    atomic_signal_fence(memory_order_release);
+    otel_thread_ctx_v1 = NULL;
+}
```
ivoanjo, line 137: I think these two lines should be switched? In particular, the barrier right now is not particularly protecting `otel_thread_ctx_v1` -- in theory I think the compiler would be able to inline `ddtrace_otel_detach` into the caller and then reorder the write to `otel_thread_ctx_v1` to later?

tracer/otel_context.c
```diff
@@ -0,0 +1,228 @@
...
+
+static void ddtrace_otel_record_begin_update(datadog_otel_thr_ctx_rec *record) {
+    atomic_store_explicit(&record->valid, 0, memory_order_relaxed);
+    atomic_signal_fence(memory_order_acquire);
+}
+
+static void ddtrace_otel_record_end_update(datadog_otel_thr_ctx_rec *record) {
+    atomic_signal_fence(memory_order_release);
+    atomic_store_explicit(&record->valid, 1, memory_order_relaxed);
+}
```
ivoanjo, line 147: Same note as for `ddtrace_otel_detach` -- the signal fence should come last for similar reasons

tracer/otel_context.h
```diff
@@ -0,0 +1,63 @@
...
+typedef struct ddtrace_root_span_data ddtrace_root_span_data;
+typedef struct ddtrace_span_stack ddtrace_span_stack;
+
+#define DATADOG_PHP_PROFILING_OTEL_ATTRS_DATA_SIZE 612
+
+typedef struct {
+    uint64_t trace_id[2];
+    _Atomic(uint64_t) span_id;
+    _Atomic(uint8_t) valid;
+    _Atomic(uint8_t) trace_flags;
```
ivoanjo, line 28: Minor: Maybe this is useful for the PHP profiler, so it's probably ok to keep; just wanted to point out that `span_id` and `trace_flags` don't actually need to be atomic, (e.g. similar to how trace_id and the others also don't need).

`valid` is the only one that needs to be treated specially, since in practice it's kind of a crappy lock -- 0 means lock is taken by writer, 1 means it's free for reading ;)

ivoanjo (commented):
I'm a big n00b on the codebase, yet I gave a pass on the OTel thread context bits, hopefully this helps!

tracer/otel_context.c
morrisonlevi: Yes, it's `ecalloc`'d. At the same time, we can just use `ddtrace_otel_record_end_update`, no big deal, I'll do that.

tracer/otel_context.h
morrisonlevi: These remain atomic because sometimes all we do is flip a single field. Since they are small enough, they will not tear in practice. Cheaper/faster when switching fibers which happens often for code that does this, or for trace propagation, etc.

tracer/otel_context.c
morrisonlevi: I intentionally kept the release signal fence before the pointer store. This matches OTEP 4947 and libdatadog’s `ThreadContext::attach()`, which performs `compiler_fence(Ordering::Release)` followed by a relaxed TLS swap.

tracer/span.c
```diff
@@ -958,6 +999,11 @@ void ddtrace_close_top_span_without_stack_swap(ddtrace_span_data *span) {
     } else {
         ZVAL_NULL(&stack->property_active);
     }
+#ifdef __linux__
+    if (stack == DDTRACE_G(active_stack) && stack->active && SPANDATA(stack->active)->stack == stack) {
+        ddtrace_update_otel_thread_context_span_id(SPANDATA(stack->active));
+    }
+#endif
```
bwoebi, line 1007: I think the `stack == DDTRACE_G(active_stack)` is wrong, because all it does is updating the spans id. If there's a later swap to this closed spans stack, it'll be stale.
And let's mirror it like `dd_drop_span`, i.e. 

```
    if (!stack->active || SPANDATA(stack->active)->stack != stack) {
        dd_close_entry_span_of_stack(stack);
#ifdef __linux__
    } else {
        ddtrace_update_otel_thread_context_span_id(SPANDATA(stack->active));
#endif
    }
```

tracer/span.c
```diff
@@ -22,15 +22,27 @@
...
 
 #define USE_REALTIME_CLOCK 0
 #define USE_MONOTONIC_CLOCK 1
 
 ZEND_EXTERN_MODULE_GLOBALS(datadog);
 
+#ifdef __linux__
+static inline void ddtrace_update_otel_thread_context_span_id(ddtrace_span_data *span) {
+    if (span && span->root) {
```
bwoebi: this should rather be a ZEND_ASSERT() than an if.
except `ddtrace_open_span` (which definitely has a valid span), all other callers use `SPANDATA(stack->active)` which does pointer mainpulation.
And an active span without a `root` is a broken span and would segfault otherwhere.

tracer/span.c
morrisonlevi: I believe I've fixed this, check the current code and refer to our conversation in Slack.

bwoebi (approved):
Logic looks right to me now.

PR #4084, opened 2026-07-30
feat(profiling): add build compatibility with PHP 8.6

realFlowControl:
### Description

Add PHP 8.6 build compatibility in the profiler by:
- removing the `ZEND_RESULT_CODE` alias
- including `Zend/zend_compile.h` for the `ZEND_INTERNAL_FUNCTION` const
- the C compilation mode changes from strict C11/C17 to GNU11/GNU17 because PHP 8.6's `ZEND_CONTAINER_OF` uses `typeof`, which is a GNU extension before C23, using the GNU dialect keeps the existing C standard baseline without requiring C23-capable toolchains
- Documents and locally allows the version-dependent `zend_ini_entry.modified` cast: it converts PHP 8.6's `bool` to `u8` and is only redundant for PHP 8.5 and older

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/profiling/build.rs
+++ b/profiling/build.rs
@@ -176,8 +176,8 @@ fn build_zend_php_ffis(
                 .chain([Path::new("../zend_abstract_interface")])
                 .chain([Path::new("../")]),
         )
-        .flag_if_supported("-std=c11")
-        .flag_if_supported("-std=c17");
+        .flag_if_supported("-std=gnu11")
+        .flag_if_supported("-std=gnu17");
     #[cfg(feature = "test")]
     build.define("CFG_TEST", "1");
     build.compile("php_ffi");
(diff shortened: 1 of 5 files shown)


bwoebi (approved):
Looks good!

morrisonlevi (commented):
Looks good to me as well! (Yes, I know it's merged ^_^)

remicollet:
Notice: max supported version not changed in package.xml

bwoebi:
We generally wait with that until RC phase, because by then the internal API is generally stable enough that it will not change.
Do you prefer another approach?

remicollet:
> We generally wait with that until RC phase, because by then the internal API is generally stable enough that it will not change. Do you prefer another approach?

That's fine for me
Only strange to see "Compatibility with PHP 8.6" in the Changelog ;)

PR #4086, opened 2026-08-03
build: remove RUSTC_BOOSTRAP=1/nightly requirements

morrisonlevi:
### Description

We haven't needed these things since a bit.

My motivation now is that codex tooling regularly trips up on this. When I told codex about it, it pointed out to me these requirements were actually unused now. I'm not sure if that's macOS specific or not, so we should be cautious about CI results.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl (approved):

bwoebi (approved):
Indeed, the main usage of this was #[linkage(extern_weak)], but we no longer need this since we handle this stuff at link time now.

PR #4087, opened 2026-08-03
Avoid leaking patched module_entry across loader shutdown

bwoebi:
This prevents crashes in SSI if the loaded library is not actually unloaded.


Leiyks (approved):
Lgtm

PR #4097, opened 2026-08-06
ci: split valgrind extension tests and stop retrying real failures

bengl:
### Description

Three CI cost/latency fixes in the tracer pipeline, found while looking at
this repo's CI times in CI Visibility.

**1. `test_extension_ci` ran the extension `.phpt` suite twice in one job.**
`make test_extension_ci` did a normal pass and then a valgrind leak-check
pass sequentially in a single job. The valgrind pass is roughly an order of
magnitude slower, and the pair routinely blew the job's 120m timeout —
throwing away the normal pass's results along with it. Over 14 days this job
class recorded **2440 failures against 3403 successes**, with `[8.0]` (481
errors / 301 successes) and `[7.4]` (321 / 260) worst affected; a large share
ended in `Job failed: execution took longer than 2h0m0s`.

Now split into `test_extension_ci_normal` / `test_extension_ci_valgrind` make
targets, run as two CI jobs. The normal job runs its pass twice, at
@bwoebi's request: shared state between `.phpt` tests only surfaces on a
second pass over the same sandbox.

**2. `MAX_TEST_PARALLELISM` didn't match the CPU request.** The job requested
12 CPUs but ran 4 test workers. On PHP >= 7.4 the normal job now uses 12.
The pods also requested CPU only — no CPU limit, no memory request — so they
were Burstable, and `ci.ddbuild.io/enforce-static-cpus=true` only pins
exclusive cores for Guaranteed QoS. Both jobs now set
`KUBERNETES_CPU_LIMIT: 12` and `KUBERNETES_MEMORY_REQUEST`/`_LIMIT: 8Gi`,
matching the existing `PHP Language Tests` precedent, so the annotation
takes effect and `-j12` gets real cores on the normal job.

**3. `ASAN test_c` overrode `default.retry` to re-add `script_failure`.**
It used the bare `retry: 2` shorthand, which retries *every* failure reason,
so genuine test failures were re-run up to 3x. The job can't block a merge —
it's in `flaky-jobs.txt` — so the retries bought no signal. It now inherits
the repo-wide infrastructure-only `default.retry` from `generate-common.php`,
which also gains it the `exit_codes: [75, 128]` transient-network retry the
bare shorthand didn't have.

`PHP Language Tests` also overrides `default.retry` with a `when:` list
including `script_failure`. That override is kept, at @bwoebi's request.

#### Timeouts are version-conditional on purpose

`RUN_TESTS_IS_PARALLEL` gates `-j` on PHP >= 7.4, so 7.0–7.3 run the suite
serially and `MAX_TEST_PARALLELISM` is dead config there. Measured durations
for successful runs over 14 days (both passes combined):

| PHP | pc95 | max |
|---|---|---|
| 7.0–7.3 | 82–89m | 91–99m |
| 7.4–8.5 | 29–31m | 32–39m |

So `>= 7.4` normal jobs get `45m` + parallelism 12; `< 7.4` keeps `120m` with
no parallelism override. Cutting the pre-7.4 budget without per-pass timing
would risk timeouts that `default.retry` then pays for three times over.

#### Behaviour explicitly preserved

- **Merge gate unchanged.** Both job names keep the `test_extension_ci:`
  prefix, so they still match the `test_extension_ci:*` glob in
  `flaky-jobs.txt`. Simulated `merge-gate.sh`'s bash glob matching over all
  generated job names: **245 gating jobs before and after**, and all 22
  `test_extension_ci` jobs non-gating.
- **`make test_extension_ci` still runs both passes**, via recursive
  `$(MAKE)` rather than prerequisites, so `make -jN` can't run them
  concurrently over the same `.phpt` sandbox.
- The valgrind recipe keeps `TEST_PHP_OUTPUT`, `DD_SPAWN_WORKER_STABLE_TRAMPOLINE`,
  `DD_TRACE_AGENT_TIMEOUT`, the `tests/ext/valgrind` `PATH` shim and the
  trailing `! grep -e '^LEAKED TEST SUMMARY'` check verbatim.

#### Verification

- `php -l` clean on all `.gitlab/generate-*.php`; all five generators run.
- Generated YAML for appsec / profiler / shared / package is **byte-identical**
  to master; only the tracer pipeline changed.
- Tracer pipeline: exactly **11 jobs added, 0 removed**; exactly 25 existing
  jobs changed (14 `ASAN test_c` retry-only, 11 `test_extension_ci`). The 11
  remaining `- script_failure` entries are all `PHP Language Tests`, kept
  deliberately.
- Generated YAML parses; every `timeout` emits at job level in both
  conditional branches.

### Reviewer checklist
- [ ] Test coverage seems ok.


.gitlab/generate-tracer.php
```diff
@@ -237,7 +237,11 @@ function before_script_steps($with_docker_auth = false) {
           - PHP_MAJOR_MINOR: "<?= $major_minor ?>"
             ARCH: "<?= $arch ?>"
       artifacts: true
-  retry: 2
+  # No `retry:` override: inherit the repo-wide `default.retry` from
```
brettlangdon: this comment probably isn't needed long term / only really applies as context to the PR.

.gitlab/generate-tracer.php
```diff
@@ -534,17 +597,10 @@ function before_script_steps($with_docker_auth = false) {
...
-    when:
-      - script_failure
-      - unknown_failure
-      - data_integrity_failure
-      - runner_system_failure
-      - scheduler_failure
-      - api_failure
-      - stuck_or_timeout_failure
-      - job_execution_timeout
+  # No `retry:` override: inherit the repo-wide `default.retry` from
```
brettlangdon: same here

.gitlab/generate-tracer.php
```diff
@@ -355,6 +359,16 @@ function before_script_steps($with_docker_auth = false) {
 <?php
 foreach ($all_minor_major_targets as $major_minor):
 ?>
+<?php /*
+ * The extension .phpt suite runs as two jobs, not two passes of one job. The
```
brettlangdon: Same here, I think this comment can be simplified as saying to parallelize them separately since valgrind takes a long time.

we don't need all the context you have here

.gitlab/generate-tracer.php
```diff
@@ -355,6 +354,8 @@ function before_script_steps($with_docker_auth = false) {
 <?php
 foreach ($all_minor_major_targets as $major_minor):
 ?>
+<?php /* Normal and valgrind passes run as separate jobs: valgrind is far
+   slower, so pairing them in one job blew its timeout. */ ?>
```
brettlangdon:
```suggestion
   slower, so run in parallel. */ ?>
```

brettlangdon (approved):
nit, otherwise lgtm

.gitlab/generate-tracer.php
```diff
@@ -369,13 +370,46 @@ function before_script_steps($with_docker_auth = false) {
...
+    KUBERNETES_POD_ANNOTATIONS_1: "ci.ddbuild.io/enforce-static-cpus=true"
+<?php if (version_compare($major_minor, "7.4", ">=")): ?>
+  timeout: 45m
+<?php else: ?>
+  # run-tests.php only gets -j on PHP >= 7.4 (RUN_TESTS_IS_PARALLEL in the
+  # Makefile), so these versions run serially and need the larger budget.
+  timeout: 120m
+<?php endif; ?>
+  script:
+    - make test_extension_ci_normal
```
bm1549, line 388: Request from @bwoebi is to run this job twice since it uncovers shared state problems 

```suggestion
    - make test_extension_ci_normal
    - make test_extension_ci_normal
```

.gitlab/generate-tracer.php
```diff
@@ -534,17 +568,6 @@ function before_script_steps($with_docker_auth = false) {
     DD_INSTRUMENTATION_TELEMETRY_ENABLED: 0
 <?php endif; ?>
   timeout: 40m
```
bm1549, line 571: @bwoebi asked to re-add these retries

```suggestion
  timeout: 40m
  retry:
    max: 2
    when:
      - script_failure
      - unknown_failure
      - data_integrity_failure
      - runner_system_failure
      - scheduler_failure
      - api_failure
      - stuck_or_timeout_failure
      - job_execution_timeout
```

.claude/ci/tracer-unit-tests.md
```diff
@@ -549,14 +550,22 @@ make test_opcache
...
+  with the normal pass. `make test_extension_ci` still runs both
+  serially for local use.
+
+- **Retries are infrastructure-only.** `test_extension_ci`,
+  `ASAN test_c`, and `PHP Language Tests` inherit `default.retry` from
+  `.gitlab/generate-common.php`, which retries runner/API/timeout
+  failures but *not* `script_failure`. A genuine test failure is not
+  retried — retrying it tripled compute without changing the outcome.
+  These job classes are listed in `flaky-jobs.txt`, so their failures
+  are already non-gating.
```
bm1549: Removed the infra-only retry at @bwoebi 's request

```suggestion
```

.claude/ci/tracer-unit-tests.md
```diff
@@ -549,14 +550,15 @@ make test_opcache
...
-  calls to run them under valgrind. This makes the job significantly
-  slower and is specific to CI.
+- **The valgrind pass is a separate job.** `make test_extension_ci_normal`
+  runs the suite normally; `make test_extension_ci_valgrind` re-runs it
+  under valgrind (`run-tests.php -m`) for leak checking. Both prepend
+  `tests/ext/valgrind` to `$PATH` — the shim there intercepts `valgrind`
+  invocations to add the suppressions file. The valgrind pass is roughly
+  an order of magnitude slower, which is why it no longer shares a job
+  with the normal pass. `make test_extension_ci` still runs both
+  serially for local use.
```
bm1549, line 561:
```suggestion

- **`PHP Language Tests` has retry:2 in CI.** These tests are
  inherently flaky due to timing-sensitive PHP runtime tests. The CI
  job retries up to 2 times on script failure.
```

Leiyks (approved):

PR #4101, opened 2026-08-10
Implement normalised http routes

estringana:
### Description

Add `_dd.appsec.normalized_route` tag to HTTP framework integrations (Laravel, Slim, Symfony, Laminas, CakePHP, Yii, CodeIgniter, WordPress) per RFC-1103. The tag exposes a normalized form of the matched route, stripping concrete parameter values and framework-specific syntax (regexconstraints, optional markers) into a canonical `{param}` notation suitable for security analysis.

### Reviewer checklist
- [ ] Test coverage seems ok.


src/DDTrace/Util/RouteNormalizer.php
```diff
@@ -0,0 +1,469 @@
...
+     */
+    public static function normalizeFromSymfony(string $path)
+    {
+        return self::normalizeBraceRoute($path, []);
+    }
+
+    /**
+     * Normalize a Laminas route template.
+     *
+     * Laminas uses :param for dynamic parameters and [...] for optional sections.
```
bwoebi, line 61: According to https://docs.laminas.dev/laminas-router/routing/ the :colon syntax is not the only one and they also have e.g. `Laminas\Router\Http\Regex`.

bwoebi:
Do you have a rough estimate on the overhead of the route normalization? It seems expensive.

Laminas is obviously just slow because it does runtime route matching, but other stacks like Slim or Symfony rely on precompiled routes.
E.g. for Symfony we could make it part of the logic used in #3676. Please check if other frameworks have similar, acceptable solutions. Or consider some caching in the C code. (a LRU cache of route to normalized route with e.g. 500 routes max)

src/DDTrace/Util/RouteNormalizer.php
estringana: Done

tabgok:
Approving from the stance of IDM, due to the new field being introduced

tabgok (approved):

tracer/routing_cache.c
```diff
@@ -0,0 +1,59 @@
+#include "routing_cache.h"
+#include "ddtrace.h"
+
+static HashTable ddtrace_rcache;
```
bwoebi: This needs to be either `TSRM_TLS` or requires mutexes.

This code, as is, is not thread-safe at all.

estringana: Solved, can you re-review plz

tests/ext/routing_cache/cache_capacity_eviction.phpt
```diff
@@ -0,0 +1,26 @@
+--TEST--
+DDTrace\routing_cache evicts the oldest inserted entry when capacity (500) is exceeded
```
cataphract, line 2: This file is misnamed. This is not an LRU cache, it's a fixed capacity FIFO cache.

src/DDTrace/Integrations/CakePHP/CakePHPIntegration.php
```diff
@@ -71,7 +71,19 @@ public static function init(): int
 
             $rootSpan = \DDTrace\root_span();
             if ($rootSpan !== null) {
-                $rootSpan->meta[Tag::HTTP_ROUTE] = $app->template;
+                $template = $app->template;
+                $rootSpan->meta[Tag::HTTP_ROUTE] = $template;
+                $cacheKey = $template;
+                $normalizedRoute = \DDTrace\routing_cache_get($cacheKey);
```
cataphract: This applies to all integrations. The RFC says to add _dd.appsec.normalized_route only when API Security is enabled and a span already has http.route (technically, it doesn't say to add "only" those conditions are present, but that's the most common-sense reading).

src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
```diff
@@ -442,20 +442,35 @@ static function() {
...
-                /** @var ContainerInterface $container */
-                $container = self::$kernel->getContainer();
-                $path = EndpointCatalog::pathForRoute($route_name, $container);
-
-                // Try with locale suffix (Symfony i18n routing convention)
-                if ($path === null) {
-                    $locale = $request->attributes->get('_locale');
-                    if ($locale !== null) {
-                        $path = EndpointCatalog::pathForRoute($route_name . '.' . $locale, $container);
+                $cacheKey = $route_name;
```
cataphract: Also in other frameworks: the cache key needs to be richer otherwise, for optional parameters, it caches only the first materialization of route that was found.

cataphract (changes requested):
It seems there are many gaps in framework support. I asked AI for failing counterexamples against the spec and got many:

https://github.com/DataDog/dd-trace-php/compare/estringana/normalise-http-route...glopes/normalise-http-route-failing?expand=1

I already removed a few that were obviously invalid (like, changing the routing type at runtime and complained about cache staleness), but the rest seem mostly valid

estringana:
@cataphract pr is ready back to you with all your comments addressed

src/DDTrace/Util/RouteNormalizer.php
```diff
@@ -581,4 +502,62 @@ public static function encodeParamName(string $name): string
...
+        $urlIdx  = 0;
+
+        foreach ($templateSegments as $seg) {
+            if (preg_match('/^\{([^}?:]+)\}$/', $seg, $m)) {
+                // Whole-segment param — present if there is a URL segment at this position
+                if ($urlIdx < count($urlSegments)) {
+                    $matched[$m[1]] = $urlSegments[$urlIdx];
+                }
+                $urlIdx++;
+            } elseif (preg_match('/\{/', $seg)) {
```
cataphract: unnecessary usage of regex

src/DDTrace/Integrations/Laravel/LaravelIntegration.php
```diff
@@ -753,4 +788,91 @@ public static function normalizeRouteName($routeName)
...
+                if ($urlIdx < count($urlSegs) && $urlSegs[$urlIdx] !== '') {
+                    preg_match_all('/\{([^}?:]+)(\?)?\}/', $seg, $pm, PREG_SET_ORDER);
+                    $paramNames  = array_map(static function($m) { return $m[1]; }, $pm);
+                    $staticParts = preg_split('/\{[^}]+\}/', $seg);
+                    $n           = count($paramNames);
+                    $urlSeg      = $urlSegs[$urlIdx];
+
+                    for ($k = $n; $k >= 1; $k--) {
+                        $regexBody = '';
+                        for ($ri = 0; $ri < $k; $ri++) {
```
cataphract: **Infer optional participation with Laravel’s compiled matcher**

This does not follow the RFC’s request-specific optional-shape rule for mixed segments: it replaces every parameter with `(.+)`, which ignores Laravel `where()` requirements. The subsequent comparison against `$allParams` cannot recover capture provenance because Laravel has already merged `->defaults()` into that map. Example of such a failure:

```
route:       Regex /normalized-regex-chain/%name%.%ext% + Literal /view
requirement: ext=pdf|json
default:     ext=txt
request:     /normalized-regex-chain/report.txt/view
captures:    name=report.txt, ext absent
post-default route params: name=report.txt, ext=txt
RFC-compliant: omit the tag, or emit /normalized-regex-chain/{name}/view
actual:      /normalized-regex-chain/{name+ext}/view
```

Suggestion: replace this parser with the framework matcher Laravel used to bind the route. It already preserves requirements and optional-group participation:

```php
private static function laravelUrlMatchedParams($route, $request, array $allParams): array
{
    $matches = [];
    $path = '/' . ltrim($request->decodedPath(), '/');
    if (preg_match($route->getCompiled()->getRegex(), $path, $matches) !== 1) {
        return $allParams;
    }

    $matched = [];
    foreach ($route->parameterNames() as $name) {
        if (isset($matches[$name]) && $matches[$name] !== ''
            && key_exists($name, $allParams)) {
            $matched[$name] = $allParams[$name];
        }
    }
    return $matched;
}
```

src/DDTrace/Integrations/Symfony/SymfonyIntegration.php
```diff
@@ -456,6 +456,51 @@ static function() {
...
+                                        }
+                                    }
+                                }
+                            }
+                            unset($_r, $_route, $_compiled, $_regex, $_rxm, $_k, $_v);
+                        }
+                        if ($matchedParams === null) {
+                            $matchedParams = \DDTrace\Util\RouteNormalizer::inferSymfonyRouteParams($path, $request->getPathInfo());
+                        }
+                        $cacheKey = $route_name . '|' . implode(',', array_keys($matchedParams));
```
cataphract: **[Low] Cache the template lookup, not capture participation**

`$matchedParams` is request-specific, so the regex replay must remain before this result-cache lookup. The deterministic part is `EndpointCatalog::pathForRoute()`, which currently checks and requires the generated route file, then converts its tokens on every request.

A minimal version using the existing persistent string cache is:

```php
private static function cachedPathForRoute($routeName, ContainerInterface $container)
{
    $cacheDir = $container->hasParameter("kernel.cache_dir")
        ? $container->getParameter("kernel.cache_dir") : null;
    if (!is_string($cacheDir) || $cacheDir === "") {
        return EndpointCatalog::pathForRoute($routeName, $container);
    }

    $key = "symfony:template:" . $cacheDir . "\0" . $routeName;
    $path = \DDTrace\routing_cache_get($key);
    if ($path !== false) {
        return $path;
    }

    $path = EndpointCatalog::pathForRoute($routeName, $container);
    if ($path !== null) {
        \DDTrace\routing_cache_set($key, $path);
    }
    return $path;
}
```

Use this helper for both the base and locale-suffixed lookup; keep the existing compiled-regex replay to calculate `$matchedParams`. The shared routing cache is only 500 entries, so a dedicated bounded template cache may be preferable if template entries would evict too many normalized-route results.

cataphract: **[Medium] Key the cached result by its template and actual captures**

`$route_name` is only used to find the route. The cached value is a normalized *path shape*, which depends on the resolved `$path` and on which named captures participated in this request. Both are available at this point.

```php
$capturedNames = array_keys($matchedParams);
sort($capturedNames, SORT_STRING);
$cacheKey = "symfony:" . $path . "\0" . implode(",", $capturedNames);
```

This keeps `/files/{name}.{ext?}` with only `name` separate from the shape where both `name` and `ext` were captured, and it keeps locale-resolved templates such as `/products/{id}` and `/produits/{id}` separate. The compiled-regex result is still what determines `$matchedParams`; this change only makes the result-cache identity match the result being stored.

src/DDTrace/Util/RouteNormalizer.php
```diff
@@ -0,0 +1,682 @@
...
+                    if ($n > 0) {
+                        $urlSeg      = $urlSegments[$urlIdx];
+                        $staticParts = preg_split('/\{[^}]+\}/', $seg);
+                        // Try with k params (k = n, n-1, ..., 1). Drop params from the right
+                        // until the URL segment matches. This handles optional trailing captures
+                        // that were injected as route defaults but absent from the URL.
+                        for ($k = $n; $k >= 1; $k--) {
+                            $regexBody = '';
+                            for ($i = 0; $i < $k; $i++) {
+                                $regexBody .= preg_quote($staticParts[$i], '/') . '(.+)';
```
cataphract: **Do not infer dynamic participation without the framework matcher**

This helper is not an RFC-compliant fallback for dynamic optional components. The RFC requires optional parameter participation to come from framework match metadata or an exact replay of the compiled matcher; this synthesized `(.+)` expression ignores route requirements, optional-capture semantics, and unmatched positions. It can therefore report a parameter as present even when the framework rejected that capture and supplied a default instead.

The inference logic here also fails in specific examples, which I can provide if you want.

Please make “accurate participation unavailable” explicit (for example, return `null`) and have dynamic-route callers omit `_dd.appsec.normalized_route` in that case. URL-only best effort should be limited to purely static optional components, as allowed by the RFC.

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -284,6 +284,95 @@ static function (SpanData $span) use ($controller, $action) {
...
+                                // Bracket-optional route: encode which optional colon-params
+                                // have values that appear in the URL path (position > 0 to
+                                // avoid matching the mandatory route prefix with a default).
+                                // Also include static-only optional sections ([/draft] etc.)
+                                // so absent/present shapes get distinct cache keys.
+                                preg_match_all('/:([a-zA-Z_][a-zA-Z0-9_-]*)/', $httpRoute, $_pm);
+                                $_present = [];
+                                foreach ($_pm[1] as $_p) {
+                                    if (isset($allParams[$_p]) && $urlPath !== null &&
+                                        strpos($urlPath, '/' . (string)$allParams[$_p]) > 0) {
```
cataphract: **Match Laminas optionals with the compiled Segment matcher**

This cache is bounded, but the optional parameter participation test is not equivalent to Laminas routing. `$allParams` comes from `RouteMatch::getParams()`, where captured parameters and route defaults have already been merged. Searching those values and static fragments anywhere in the URL loses both capture provenance and route position.

Two concrete false positives are:

```
template: /prefix/:id[/:value]
URL:      /prefix/foo
matcher:  id=foo; value=foo comes only from the route default
expected: /prefix/{id}
actual:   /prefix/{id}/{value}
```

```
template: /prefix/:id[/foo]
URL:      /prefix/foo
matcher:  id=foo; optional static /foo absent
expected: /prefix/{id}
actual:   /prefix/{id}/foo
```

Nested static groups also collide in the cache:

```
template: /archive[/draft[/preview]]
first URL:  /archive       -> key /archive# -> caches /archive
second URL: /archive/draft -> same key, because the one-pass bracket regex
                              sees only the inner [/preview]
expected: /archive/draft
actual:   /archive
```

For dynamic optionals, the RFC requires participation to come from framework match metadata or an exact replay of the framework matcher. A defaulted value occurring somewhere in the URL is not evidence that its capture participated.

I tested replacing this heuristic with a replay of Laminas `Segment::match()` before defaults are merged. The important part is to use the Segment's compiled regex and `paramMap`, retain capture offsets, and walk the Segment's parsed `parts` at those exact offsets to recover the expanded route shape:

```php
public static function inferLaminasSegmentMatch(
    \Laminas\Router\Http\Segment $route,
    string $urlPath
) {
    $routeData = \Closure::bind(
        static function ($segment) {
            return [
                $segment->regex,
                $segment->paramMap,
                $segment->parts,
                $segment->translationKeys,
            ];
        },
        null,
        \Laminas\Router\Http\Segment::class
    )($route);

    if (!is_array($routeData) || count($routeData) !== 4 || !empty($routeData[3])) {
        return null;
    }

    $matches = [];
    if (@preg_match(
        '(^' . $routeData[0] . '$)',
        $urlPath,
        $matches,
        PREG_OFFSET_CAPTURE
    ) !== 1) {
        return null;
    }

    $captures = [];
    $params = [];
    foreach ($routeData[1] as $group => $name) {
        if (!isset($matches[$group][0]) || $matches[$group][0] === '') {
            continue;
        }
        $captures[$name][] = [
            'raw' => $matches[$group][0],
            'offset' => $matches[$group][1],
        ];
        $params[$name] = rawurldecode($matches[$group][0]);
    }

    $states = [[
        'offset' => 0,
        'template' => '',
        'capture_indexes' => [],
    ]];
    $states = self::matchLaminasSegmentParts(
        $routeData[2],
        $urlPath,
        $captures,
        $states
    );

    foreach ($states as $state) {
        if ($state['offset'] !== strlen($urlPath)) {
            continue;
        }
        foreach ($captures as $name => $values) {
            if (($state['capture_indexes'][$name] ?? 0) !== count($values)) {
                continue 2;
            }
        }
        return [
            'template' => $state['template'],
            'params' => $params,
        ];
    }

    return null;
}
```

`matchLaminasSegmentParts()` advances literal parts only at the current URL offset, consumes each parameter only at its matcher-reported capture offset, and branches structurally for each optional part. That preserves nested optional groups and repeated values without putting raw request values into the cache key. The call site then uses the exact expanded template as the bounded shape key and normalizes with only the pre-default captured parameters:

```php
$segmentMatch = self::inferLaminasSegmentMatch($leafRoute, $urlPath);
$cacheKey = $httpRoute . '#' . $segmentMatch['template'];

$normalizedRoute = RouteNormalizer::normalizeFromLaminas(
    $segmentMatch['template'],
    $segmentMatch['params'],
    null,
    null
);
```

src/DDTrace/Integrations/Laminas/LaminasIntegration.php
```diff
@@ -284,6 +284,95 @@ static function (SpanData $span) use ($controller, $action) {
...
+                                        unset($_routeRegex, $_rxm, $_k, $_v);
+                                    }
+                                    unset($_leafRoute);
+                                } catch (\Throwable $_ex) {
+                                    unset($_leafRoute, $_routeRegex, $_rxm, $_k, $_v, $_ex);
+                                    $urlMatchedFromRegex = null;
+                                }
+                                $_braceTemp = preg_replace('/%([a-zA-Z_][a-zA-Z0-9_]*)%/', '{$1}', $httpRoute);
+                                $_urlMatchedKeys = array_keys(
+                                    $urlMatchedFromRegex ?? \DDTrace\Util\RouteNormalizer::inferSymfonyRouteParams($_braceTemp, $urlPath)
```
cataphract: **[High] Do not fall back to URL-only parameter inference**

`inferSymfonyRouteParams()` must not be used here to guess dynamic optional participation. RFC-1103 says: “Tracers must use the framework's matched-parameter map,” “must not infer presence from the URL alone,” and must “fall back to omitting the tag rather than emitting an inaccurate value.”

When `inferLaminasRegexMatch()` cannot replay a specific route composition, this fallback discards matcher constraints and reconstructs captures with generic `(.+)` expressions. That can mark defaulted parameters as URL captures and produce a plausible but incorrect normalized route.

Please either support the specific route pattern by replaying its actual Laminas matcher while preserving pre-default capture participation, or treat participation as unavailable and do not set `_dd.appsec.normalized_route` for that request. A constraint-free inferred shape is not an RFC-compliant fallback.

cataphract: **Do not use constraint-free URL inference as a fallback**

As mentioned in other comments, RFC-1103 says: “Tracers must use the framework's matched-parameter map,” “must not infer presence from the URL alone,” and must “fall back to omitting the tag rather than emitting an inaccurate value.” `inferSymfonyRouteParams()` does the prohibited URL-only inference with generic `(.+)` captures, so it loses matcher constraints and cannot distinguish an absent capture from a route default.

```text
rule:     /x/{name}.{ext?}; ext=pdf|json; default ext=txt
request:  /x/report.txt
expected: /x/{name} (or no tag)
actual:   /x/{name+ext}
```

```text
rule:     Regex /normalized-regex-chain/%name%.%ext% + Literal /view; ext=pdf|json; default ext=txt
request:  /normalized-regex-chain/report.txt/view
expected: /normalized-regex-chain/{name}/view (or no tag)
actual:   /normalized-regex-chain/{name+ext}/view
```

In both cases the framework captures the entire `report.txt` value as `name`; `ext` does not participate and is added only from the default. Please support each specific pattern by replaying its actual matcher before defaults are merged, or omit `_dd.appsec.normalized_route`.

src/DDTrace/Util/RouteNormalizer.php
```diff
@@ -0,0 +1,682 @@
...
+     * WordPress route matching uses regex rules like "^blog/([^/]+)/?$".
+     * Named parameters are not available; placeholders param1, param2, … are used.
+     *
+     * @param string      $matchedRule Value of $wp->matched_rule
+     * @param string|null $urlPath     Value of $wp->request; used to detect which
+     *                                 optional capture groups actually participated
+     *                                 in the match, so phantom segments are not emitted.
+     * @return string|null
+     */
+    public static function normalizeFromWordPress(string $matchedRule, $urlPath = null)
```
cataphract: **Use PCRE's match result instead of parsing the rewrite expression**

The cache now distinguishes capture participation, but the route plan feeding it is still produced by a partial PCRE parser. That parser misclassifies valid syntax:

```text
^normalized-quoted-literal/\Qfile.json\E$
expected: /normalized-quoted-literal/file.json
actual:   /normalized-quoted-literal/Qfile.jsonE

^normalized-quoted-captures/(?'first'[^/]+)-(?'second'[^/]+)$
expected: /normalized-quoted-captures/{first+second}
actual:   /normalized-quoted-captures/{param1}
```

PHP's match result already provides the authoritative numeric captures, named aliases, participation, and byte offsets. We can build the request-specific route from that result instead of extracting literals and capture names from the regex source. Before retaining concrete gaps as constants, use a conservative rejection filter: if the rule can consume variable text outside a capture (for example an uncaptured character class, repetition, or backreference, say `^foo/[a-z]+/bar`), return `null` and omit the tag. Fixed literal alternatives and optional literals like `red|blue` remain bounded.

cataphract (commented):
The mostly unreviewed implementation of the suggestions is at https://github.com/DataDog/dd-trace-php/compare/82db9a3fd0bb94a4961d2ef9a53eb7799928db59...glopes/normalise-http-route

PR #4102, opened 2026-08-11
fix(stats): scan all peer tag keys sent by the agent

ichinaski:
### Description

`DDTRACE_MAX_PEER_TAGS` is meant to bound how many peer tags a single span collects, but it was also truncating the key list received from the agent's `/info` before lookup. That list is sorted alphabetically and is already longer than the cap, so keys sorting last (`out.host`, `peer.hostname`, `peer.service`, `server.address`) never matched.

It became visible when the agent added `db.system.name` to its semantic registry: the key sorts at index 16 and pushed `network.destination.name` from 31 to 32, out of the scanned window. Guzzle client spans carry only that key, so client-side stats reported no peer tags at all and system-tests `Test_Peer_Tags` started failing on every PHP PR.

The existing `client_side_stats_peer_tags.phpt` now advertises a key list longer than the cap with a matching key at each end. It fails without the fix (`out.host` dropped) and passes with it. Verified against system-tests `TRACE_STATS_COMPUTATION` on `datadog/agent-dev:master-py3` with a locally built package.

[APMSP-3911](https://datadoghq.atlassian.net/browse/APMSP-3911)

### Reviewer checklist
- [ ] Test coverage seems ok.

[APMSP-3911]: https://datadoghq.atlassian.net/browse/APMSP-3911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tracer/span_stats.c
+++ b/tracer/span_stats.c
@@ -349,10 +349,6 @@ static void ddtrace_span_concentrator_feed_cb(const ddog_SpanConcentrator *c, vo
         size_t peer_tag_keys_count = 0;
         const ddog_CharSlice *peer_tag_keys = ddog_span_concentrator_peer_tag_keys(c, &peer_tag_keys_count);
         if (peer_tag_keys_count > 0 && peer_tag_keys) {
-            // All configured keys are looked up: the cap bounds how many tags we *collect*, not how
-            // many keys we scan. The agent's key list is sorted alphabetically and already longer
-            // than DDTRACE_MAX_PEER_TAGS, so capping the scan would silently ignore whichever keys
-            // happen to sort last (e.g. out.host, peer.hostname, server.address).
             for (size_t i = 0; i < peer_tag_keys_count && actual_peer_tags < DDTRACE_MAX_PEER_TAGS; i++) {
                 const ddog_CharSlice *k = &peer_tag_keys[i];
                 zval *val = zend_hash_str_find(pre->meta, k->ptr, k->len);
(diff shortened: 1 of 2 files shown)


tracer/span_stats.c
```diff
@@ -346,11 +348,12 @@ static void ddtrace_span_concentrator_feed_cb(const ddog_SpanConcentrator *c, vo
...
         const ddog_CharSlice *peer_tag_keys = ddog_span_concentrator_peer_tag_keys(c, &peer_tag_keys_count);
-        if (peer_tag_keys_count > DDTRACE_MAX_PEER_TAGS) {
-            peer_tag_keys_count = DDTRACE_MAX_PEER_TAGS;
-        }
         if (peer_tag_keys_count > 0 && peer_tag_keys) {
-            for (size_t i = 0; i < peer_tag_keys_count; i++) {
+            // All configured keys are looked up: the cap bounds how many tags we *collect*, not how
+            // many keys we scan. The agent's key list is sorted alphabetically and already longer
+            // than DDTRACE_MAX_PEER_TAGS, so capping the scan would silently ignore whichever keys
+            // happen to sort last (e.g. out.host, peer.hostname, server.address).
```
bwoebi:
```suggestion
```

bwoebi (approved):
Thanks, I misunderstood the config when I implemented it.

PR #4106, opened 2026-08-13
ci: keep request replayer responsive under parallel tests

bm1549:
### Description

[High-level change and expected CI impact report](https://chonk.us1.prod.dog/v2/get/chonk-uploads-prod/brian.marks/6f855a69-f90c-4da2-83aa-ea69e9d6dab8)

Stacked on #4097.

`test_extension_ci` runs as many as 12 PHPT workers against one single-threaded request-replayer. If its worker gets stuck processing a request, `/replay` also blocks. The test process then waits on agent requests until the job reaches its two-hour timeout.

Run request-replayer with 16 PHP CLI workers. Its file-backed queues use read-modify-write operations, so each session now has a file lock around state changes. MessagePack decoding stays outside the lock. Requests for different sessions can still run at the same time, including the un-tokened PHPT path.

The image is built and signed by CI (`Services build` in the `ci-images` pipeline) with `src/` baked in, so no runtime source injection is needed. UDP metrics runs under a restart loop. It publishes an immutable versioned tag (`php-request-replayer-3.0`), replacing the manual Docker Hub push and the `DataDog/images` mirror digest bump.

This is separate from #4097 (now merged): that PR splits normal and valgrind passes and raises test parallelism; this PR keeps request-replayer responsive under that parallel workload.

#### Verification

- The exact CI image retained 200 concurrent requests and passed replay, metrics, stats, remote-config, agent-info, clear, and next-response checks.
- Killing the UDP metrics process produced a replacement process, which accepted and replayed a tagged metric.
- The test harness passed 12 consecutive runs in the exact `generate-templates` image with one CPU and 256 MiB. Child logs use files rather than undrained pipes and are printed if the test fails.
- PHP and shell syntax checks pass; `git diff --check` is clean.

### Reviewer checklist
- [ ] Test coverage seems ok.


.gitlab/test-request-replayer-concurrency.php
bwoebi, line 1: I don't consider this test worth it. It does a simple flock - if you forget this on new endpoints or something, it won't catch it.

.gitlab/install-request-replayer-source.sh
bwoebi, line 1: This is to be replaced by a proper docker container rebuild - seems temporary. :-D

.gitlab/generate-common.php
```diff
@@ -227,11 +227,28 @@ function windows_git_setup_with_packages() {
       KUBERNETES_SERVICE_MEMORY_LIMIT: 512Mi
 
   request-replayer:
-    name: registry.ddbuild.io/images/mirror/datadog/dd-trace-ci:php-request-replayer-2.0
+    name: registry.ddbuild.io/ci/dd-trace-php/dd-trace-ci:php-request-replayer-3.0
```
morrisonlevi: If we're changing the name/url, why don't we do something actually sensible like `dd-trace-php/request-replayer:3.0`?

dockerfiles/services/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.7'
+
```
morrisonlevi:
```suggestion
```

We should delete the version. The top-level version key was deprecated when Compose merged into the unified Compose Specification (aka Compose v2). 

Also, the file should be renamed to `compose.yaml`, which is the first filename compose will look for. FYI the order is:
 1. `compose.yaml`
 2. `compose.yml`
 3. `docker-compose.yaml`
 4. `docker-compose.yml`

I understand not changing existing files as much because that might cause churn or break things, which is annoying, but for new stuff this isn't a problem.

.gitlab/test-request-replayer-concurrency.php
Leiyks: Removed it

.gitlab/install-request-replayer-source.sh
Leiyks: Added proper docker container rebuild mechanism

.gitlab/generate-common.php
Leiyks: Sure, I don't mind changing

.gitlab/generate-common.php
```diff
@@ -227,11 +227,28 @@ function windows_git_setup_with_packages() {
...
+        # own spawn races under PHP_CLI_SERVER_WORKERS.
+        (
+          n=0
+          while true; do
+            n=$((n+1))
+            if [ "$n" -gt 20 ]; then echo "metricsserver: giving up after $n starts" >&2; exit 1; fi
+            php /var/www/metricsserver.php || echo "metricsserver: exited (start $n), restarting" >&2
+            sleep 1
+          done
+        ) &
```
bwoebi: Can we bring it back and use a flock() to avoid racing it? Would love if request-replayer were self-contained here.

bwoebi: (I know we have REQUEST_REPLAYER_METRICS_SERVER_MANAGED ... but yea...)

bwoebi (approved):
Looks good to me, and tests seem to work

PR #4115, opened 2026-08-18
fix: vsnprintf off-by-one errors

morrisonlevi:
### Description

Fixes some off-by-one errors when using `vsnprintf`.

`vsnprintf` returns the required character count excluding the null terminator. To write an N-sized message, you need an N+1 (or larger) sized buffer. If you don't, then `vsnprintf` will truncate the message. Since this is usually used with logs, this is usually not that consequential, and in some cases the length is correctly tracked so it's truly harmless.

BUT in appsec's `_emit_error`, the buffer was sized `len` and `vsnprintf` was told `len + 1`, so there's a genuine heap-buffer-overflow there.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/appsec/src/extension/request_abort.c
+++ b/appsec/src/extension/request_abort.c
@@ -585,7 +585,7 @@ static void _emit_error(const char *format, ...)
     char *msg = NULL;
     bool free_msg = false;
     if (len >= (int)sizeof(buf)) {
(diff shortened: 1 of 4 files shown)


bwoebi (approved):
Looks right to me. Quite the footgun :-D

PR #4120, opened 2026-08-19
refactor: consolidate Rust extension workspace

morrisonlevi:
### Description

This is a step on the road to having a common extension between the tracer and profiling. The high level idea is to still have separate `ddtrace.so` and `datadog-profiling.so` artifacts, but they are now buildable through the top-level `config.m4` and `Cargo.toml`. Meaning, we have reduce the number of crates in the system.

The `config.m4` file sets up `CARGO_TARGET_DIR` by default to be `target-profiling/`, by the way. Since we can build different libraries from the same crate, we would otherwise end up clobbering the other if you built them both without renaming in between. I don't like this but it seemed the best for now, it will go away when we _actually_ combine the .so files.

### Reviewer checklist
- [ ] Test coverage seems ok.


.gitlab/generate-profiler.php
```diff
@@ -59,15 +59,15 @@
     - '# NTS'
     - '# Use if/then instead of `command -v switch-php && switch-php` — the && form exits 1 when switch-php is absent, which FF_ENABLE_BASH_EXIT_CODE_CHECK treats as a job failure'
     - if command -v switch-php > /dev/null 2>&1; then switch-php "${PHP_MAJOR_MINOR}"; fi
-    - cargo build --profile profiler-release --all-features
-    - (cd ../; TEST_PHP_JUNIT="${CI_PROJECT_DIR}/artifacts/profiler-tests/nts-results.xml" php profiling/tests/run-tests.php -d "extension=/mnt/ramdisk/cargo/profiler-release/libdatadog_php_profiling.so" --show-diff -g "FAIL,XFAIL,BORK,WARN,LEAK,XLEAK,SKIP" "profiling/tests/phpt")
+    - (cd ..; phpize && DDTRACE_PROFILING_FEATURES="debug_stats,stack_walking_tests,test,tracing,tracing-subscriber,trigger_time_sample" ./configure --disable-ddtrace-tracer --enable-ddtrace-profiling && make -j$(nproc))
```
bwoebi: Can we have `--with-ddtrace-features=` instead of an env var?

components-rs/lib.rs
```diff
@@ -1,35 +1,86 @@
-#![allow(static_mut_refs)] // remove with move to Rust 2024 edition
-
+#![allow(static_mut_refs)]
+// remove with move to Rust 2024 edition
```
bwoebi: why did you put that on the line after?

components-rs/tracer.rs
bwoebi, line 1: Can we move these to the tracer/ dir please? (`tracer/mod.rs`) when we do this anyway?

components-rs/components.rs
bwoebi, line 1: What's this file for? Why is that not simply part of lib.rs?

ext/configuration.c
```diff
@@ -14,6 +14,28 @@ ZEND_EXTERN_MODULE_GLOBALS(datadog);
 
 #include <tracer/configuration_dependencies.h>
 
+#ifndef DDTRACE
```
bwoebi, line 17: Right, I missed that.
But it still needs to submit to the sidecar, i.e.
```
    if (DATADOG_G(request_initialized)) {
        ddtrace_sidecar_submit_span_data_direct(&DATADOG_G(sidecar), NULL, new_str, get_DD_ENV(), get_DD_VERSION());
    }
```
needs to be preserved for non-tracer build.

ext/remote_config.c
```diff
@@ -83,6 +87,7 @@ static void dd_sigvtalarm_handler(int signal, siginfo_t *siginfo, void *ctx) {
     ZEND_HASH_FOREACH_PTR(&datadog_tls_bases, TSRMLS_CACHE) {
 #endif
     // On Linux the signal gets delivered to the thread that set the timer, so we don't need to iterate all threads
+#ifdef DDTRACE
```
bwoebi: why the extra #ifdef? shouldn't the whole block from lines 79 to 110 be in #ifdef DDTRACE?

bwoebi: (honestly, it should be moved to the tracer/ dir directly)

profiling/src/profiler/mod.rs
bwoebi, line 1: You really want `profiling::profiling`? Maybe do like `profiling::profiler` or such?

morrisonlevi: Well, no, of course not. It's going from `datadog_php_profiling::profiling::` to `crate::profiling::profiling::` and sure that sucks but it's just a mechanical rename. I'd like to restructure more things too but the PR is already uncomfortably big. I see your point though so I'll think of something to rename this one!

morrisonlevi: I just went with profiler in this case.

ext/remote_config.c
morrisonlevi: Resolved, I think? Take a look

bwoebi (approved):
Looks good now :-)

PR #4134, opened 2026-08-25
Fix gitlab pipeline issues

randomanderson:
### Description

This PR fixes a few failures with the gitlab pipeline that affected 1.24.0 and 1.24.1.

* A couple jobs were using unsigned images that were getting blocked now that signature enforcement has rolled out. Rather than find updated versions of these esoteric images, I replaced them with generic images.
* `ci_authenticated_job.yml` no longer exists and contains an unsigned image
* `REPO_NOTIFICATION_CHANNEL` was unset so release failure notifications when unseen by the PHP team.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (commented):
Assuming the image substitutions actually work, this is acceptable to me. I have no idea why `registry.ddbuild.io/ci/async-profiler-build:v71888475-datadog-ci` is being used for  `"upload SSI debug symbols"` anyway.

.gitlab/generate-package.php
```diff
@@ -1739,7 +1741,7 @@ function appsec_image_from_tag_mapping(string $tag): string
       else
         echo "UPSTREAM_TRACER_VERSION=$(<VERSION)" > upstream.env
       fi
-    - mv packages/dd-library-php-*-x86_64-linux-gnu.tar.gz dd-library-php-x86_64-linux-gnu.tar.gz
+    - cp packages/dd-library-php-*-x86_64-linux-gnu.tar.gz dd-library-php-x86_64-linux-gnu.tar.gz
```
bwoebi, line 1744: What's this for? These are gigantic files and changing this back to `mv` would be one of the first things I would do if I were to optimize this job probably.

randomanderson: The user in the new image isn't root so it can't move files from the Gitlab cache which are owned by root.

Looking at the job, the `cp` command takes ~1 second to complete so its not a bottleneck.

bwoebi: Alright then

bwoebi (approved):

PR #4135, opened 2026-08-25
feat(ffe): send the split serial id on exposure events [EX-3425]

danyal002:
### Description

Sends the split serial id on the exposure events that `dd-trace-php` emits. The exposures intake uses the serial id to find the holdout an allocation comes from. The compiler rewrites a holdout into an ordinary allocation before an SDK receives it, so the serial id is the only link back to it.

The serial id travels from the evaluation result, through the exposure buffer, into the sidecar FFI struct. Serial ids are zero-based per organization, so `0` is a real value and cannot signal absence. The value therefore carries a separate presence flag, matching the existing `FfeResult` convention in this file.

| File | Change |
|---|---|
| `tracer/functions.c` | pass `result.serial_id` and `result.has_serial_id` into the recorder |
| `tracer/ffe.c` | hold both on the buffered exposure, set both on `ddog_FfeExposure` |
| `tracer/ffe.h` | recorder signature |
| `libdatadog` | advance to `0c0c60b96` |
| `components-rs/common.h`, `components-rs/sidecar.h`, `Cargo.lock` | regenerated |

Serialization, the wire key, and exposure deduplication live in libdatadog and are done there: DataDog/libdatadog#2402. That PR also fixed the deduplication cache, which compared only allocation key and variant, so a serial id that appeared or changed on an otherwise unchanged assignment was suppressed.

The submodule bump is required, not incidental: `ddog_FfeExposure` gains its two fields in `0c0c60b96`, so the header and the compiled Rust struct must move together.

`components-rs/sidecar.h` and `Cargo.lock` are generated output, regenerated here with `make cbindgen`. Their churn is accumulated drift rather than a consequence of this change. The checked-in headers were last regenerated in 018f21296, and the submodule pointer has advanced twice since then (#4118, #4123) without a regeneration, so this run sweeps up everything in between. Concretely, `sidecar.h` gains two additive `ddog_sidecar_send_trace_v1_*` declarations from 69df7ed9b, which nothing in this repo calls, and `Cargo.lock` gains the `zrip` crates. Reviewing `tracer/` plus the three added lines in `common.h` covers the whole behavioural change.

This overlaps DataDog/dd-trace-php#4125, which bumps to a commit that predates `0c0c60b96`; whichever lands second needs a rebase.

No validation is applied to the value. The flag configuration layer owns that.

### Testing

PHP 8.3 debug on `bookworm`, submodule at `0c0c60b96`:

- `make -j4 all` links `ffe.o` and `functions.o` with no new warnings
- `make test_c TESTS=tests/ext/ffe` passes 6/6, none skipped
- an evaluation on a split declaring `serialId: 0` reports `serialId=0` with `doLog=true`, a split omitting it reports `null`, and `flush_ffe_exposures()` hands the batch to the sidecar

No test here covers the buffer-to-FFI hop. The exposure buffer is not observable from PHP, and no test in this repo asserts an exposure reaching the agent. Cross-language coverage comes from system-tests, where `Test_FFE_Exposure_Serial_Id` and the three serial-id caching classes already exist; `manifests/php.yml` gets flipped off `missing_feature (EX-3425)` in a follow-up once this ships.

### Reviewer checklist
- [ ] Test coverage seems ok.

*This description was generated by Claude.*

### DIFF
--- a/tracer/ffe.c
+++ b/tracer/ffe.c
@@ -31,6 +31,8 @@ typedef struct {
     zend_string *subject_attributes_json;
     zend_string *allocation_key;
     zend_string *variant;
+    int32_t serial_id;
+    bool has_serial_id;
 } dd_ffe_exposure;
 
 static void dd_ffe_release_metric(dd_ffe_metric *metric) {
@@ -168,7 +170,9 @@ void ddtrace_ffe_record_exposure(
     zend_string *targeting_key,
     zend_string *subject_attributes_json,
     zend_string *allocation_key,
-    zend_string *variant
+    zend_string *variant,
+    int32_t serial_id,
+    bool has_serial_id
 ) {
     if (ZSTR_LEN(flag_key) == 0 || ZSTR_LEN(variant) == 0) {
         return;
@@ -200,6 +204,8 @@ void ddtrace_ffe_record_exposure(
     exposure->subject_attributes_json = zend_string_copy(subject_attributes_json);
     exposure->allocation_key = zend_string_copy(allocation_key);
     exposure->variant = zend_string_copy(variant);
+    exposure->serial_id = serial_id;
+    exposure->has_serial_id = has_serial_id;
 }
 
 bool ddtrace_ffe_flush_exposures(void) {
@@ -224,6 +230,8 @@ bool ddtrace_ffe_flush_exposures(void) {
             .subject_attributes_json = dd_zend_string_to_CharSlice(buffer[i].subject_attributes_json),
             .allocation_key = dd_zend_string_to_CharSlice(buffer[i].allocation_key),
             .variant = dd_zend_string_to_CharSlice(buffer[i].variant),
+            .serial_id = buffer[i].serial_id,
+            .has_serial_id = buffer[i].has_serial_id,
         };
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2922,6 +2922,7 @@ dependencies = [
  "tracing",
  "uuid",
  "web-time",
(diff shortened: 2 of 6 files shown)


tracer/ffe.c
```diff
@@ -224,6 +230,8 @@ bool ddtrace_ffe_flush_exposures(void) {
             .subject_attributes_json = dd_zend_string_to_CharSlice(buffer[i].subject_attributes_json),
             .allocation_key = dd_zend_string_to_CharSlice(buffer[i].allocation_key),
             .variant = dd_zend_string_to_CharSlice(buffer[i].variant),
+            .serial_id = buffer[i].serial_id,
```
sameerank, line 233: Non-blocking: could we add a focused test for this C buffer-to-FFI mapping, especially distinguishing `serial_id = 0` with `has_serial_id = true` from an absent serial ID? Go has an [analogous event-to-buffer test](https://github.com/DataDog/dd-trace-go/blob/2b8a95175e47a1e560a541206bd46f49e0b0b974/openfeature/exposure_test.go#L169-L247). Fine to defer if enabling the planned system tests will exercise this exact path.

sameerank (approved):

bwoebi (approved):

tracer/ffe.c
danyal002: So the system test does test this exact path: `system-tests/tests/ffe/test_exposures_datadog_agent.py::Test_FFE_Exposure_Caching_Serial_Id_Appears`. I also added similar tests in the Rust sidecar for this sdk.

PR #4138, opened 2026-08-26
fix(installer): set explicit modes instead of inheriting the umask

Leiyks:
### Description

`datadog-setup.php` never set explicit modes on anything it installed, so
permissions fell out of whatever umask the invoking user happened to have.

**Symptom.** A customer on Arch Linux/arm64 ran the installer as root with
`umask 0027`. `/usr/lib/php/modules/ddtrace.so` and the
`/opt/datadog/dd-library/` tree came out with no read bit for "other".
php-fpm workers run as a different user (`http`), so loading the extension
failed with `EACCES` — which surfaced confusingly as:

```
Failed connecting to the sidecar (subprocess mode): Connection refused (os error 111)
```

The spawned sidecar subprocess couldn't load the extension. The customer
worked around it by hand with `chmod a+r`, having had to do the same earlier
for `/opt/datadog/dd-library/`.

The INI file was affected the same way, and that case is worse: a `0640`
`98-ddtrace.ini` means PHP can't read `extension=ddtrace.so` at all, so the
tracer just vanishes with *no* permission error anywhere. That is likely why
this took a while to diagnose.

### What changed

- Explicit modes on everything installed, independent of umask: `0644` for
  data files (`.so`, `.ini`, PHP sources), `0755` for directories and
  executables.
- Re-running the installer now repairs a previously-broken install, including
  a stale restrictive INI.
- Extensions are chmod'd at their temp name *before* the atomic rename, so the
  live `.so` path is never briefly unreadable.
- No `chown`. Non-root installs into a user-owned prefix are unaffected.

The policy is **chmod what we own, warn about what we don't**:

- Directories the installer did not create are never chmod'd — silently
  widening a user-supplied `--install-dir` isn't acceptable — it warns with an
  actionable command instead.
- Files the installer does not own are never chmod'd. Only INI drop-ins in a
  scan directory qualify (whatever name was adopted). A `php.ini`, which
  `--ini` can point at, is the user's and may deliberately not be
  world-readable since it can hold credentials; those get a warning.

### Testing

New `dockerfiles/verify_packages/installer/test_install_umask_permissions.sh`
— the package-verification harness previously had no permission assertions at
all. It installs under `umask 0027`, asserts modes in both affected locations,
and finishes on the real symptom: `php -m` as another user must list ddtrace.
It fails on the pre-fix installer and passes after.

### DIFF
--- a/datadog-setup.php
+++ b/datadog-setup.php
@@ -30,6 +30,14 @@ const OPT_INI_SETTING = 'd';
 // Release version is set while generating the final release files
 const RELEASE_VERSION = '@release_version@';
 
+// Modes set explicitly, so that the install does not depend on the caller's umask
+const MODE_FILE = 0644;
+const MODE_EXECUTABLE = 0755;
+const MODE_DIR = 0755;
+
+// The INI file the installer creates in the INI scan directory
+const DEFAULT_INI_FILE_NAME = '98-ddtrace.ini';
+
 // phpcs:disable Generic.Files.LineLength.TooLong
 // For testing purposes, we need an alternate repo where we can push bundles that includes changes that we are
 // trying to test, as the previously released versions would not have those changes.
@@ -351,6 +359,7 @@ function cmd_config_set(array $options)
                         }
                         continue;
                     } else {
+                        set_mode($iniFile, MODE_FILE);
                         echo "Success.\n";
                     }
                 }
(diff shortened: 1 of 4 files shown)


datadog-setup.php
```diff
@@ -925,6 +939,240 @@ function find_main_ini_files(array $phpProperties)
...
+ * skipped, as chmod would follow them outside of the tree.
+ *
+ * `$dataOnly` marks a tree with no runnable file, so that an exec bit recorded
+ * in the package - a few of the shipped PHP sources carry one - is dropped.
+ *
+ * @param string $path
+ * @param bool $dataOnly
+ * @return void
+ */
+function set_mode_recursive($path, $dataOnly = false)
```
bwoebi: What's $dataOnly for? As far as I know, we don't install executables at all. (only shared objects)

bwoebi (approved):
Looks goo, but never add the executable bit to files please.

PR #4144, opened 2026-08-27
ci: reuse system-tests runner image

bm1549:
### Description

> [Link](https://chonk.us1.prod.dog/v2/get/chonk-uploads-prod/brian.marks/75ba650a-aa10-422e-8c04-467f96416d87) (AppGate required)

System-test jobs currently spend about 4m18s on recursive submodules and about 5 minutes installing Docker and Python tooling. This moves them to the upstream system-tests runner image, disables dd-trace-php submodules for these jobs, reuses the image virtualenv, and builds only the PHP weblog.

Each pipeline resolves `system-tests/main` once in `prepare code` and passes that exact SHA to every system-test job through a dotenv artifact. This keeps upstream source drift between pipelines without letting parallel jobs within one pipeline use different revisions.

The runner stays content-addressed at `ci-runner:75a57a4b6391`. Each job performs an exact shallow checkout, then runs the upstream `build_ci_image.py --check-only` check before fetching Docker Hub or Datadog credentials. A change to the upstream runner Dockerfile or `requirements.txt` fails early until this repository advances the image tag. Parent YAML defaults are moved into the generated child pipeline so the system-test override cannot be outranked by trigger variables.

Expected impact: about 9m18s less wall time per executing job, roughly 2h38m of executor time across 17 routine expansions, or 6h31m when all 42 expansions run. Shallow-checkout savings are not included.

Tested with:

- Fresh `system-tests/main` checkout and source/image hash match
- Generated-pipeline contract for main resolution, dotenv propagation, exact checkout, and image validation
- Runner tool, virtualenv, default, alternate-weblog, tracer-release, and parametric probes
- Generated package YAML: 30 definitions / 42 expansions with unchanged matrices
- Trigger precedence, artifacts, failure propagation, PHP lint, and `git diff --check`

### Reviewer checklist
- [ ] Test coverage seems ok.


.gitlab/generate-package.php
```diff
@@ -1286,34 +1289,23 @@ function appsec_image_from_tag_mapping(string $tag): string
...
-      mkdir -p /etc/apt/keyrings
-      curl -fsSL https://download.docker.com/linux/debian/gpg -o /etc/apt/keyrings/docker.asc
-      echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.asc] https://download.docker.com/linux/debian $(. /etc/os-release && echo "$VERSION_CODENAME") stable" > /etc/apt/sources.list.d/docker.list
-      apt-get update -o dir::state::lists="$APT_CACHE/lists"
-      apt-get install -y --no-install-recommends -o dir::state::lists="$APT_CACHE/lists" -o dir::cache::archives="$APT_CACHE/archives" docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
-
-      # Install Python dependencies
-      pip install -U pip virtualenv
+      set -e
+      readonly PINNED_SYSTEM_TESTS_COMMIT=f16fe9ed13a45ca4b2bbb9bc55c643347b6c611f
```
bwoebi: Where does this come from? We would like to not pin this, or is this somehow tied to the runner image?

bwoebi: I mean, that would require us to repeatedly bump it, if we wanted the newest system-tests right?

bm1549: The agent decided to do this since it hates unpinned dependencies, even though I asked it to stay on latest

I've since updated it to follow main

PR #4145, opened 2026-08-27
fix(drupal): hook the theme engine service and guarantee render hook removal

Leiyks:
# Description

Fixes APMS-20395.

On Drupal >= 11.3, `drupal.template.file` was never set on any `drupal.theme.render` span, and hooks accumulated on `twig_render_template` until `DD_TRACE_HOOK_LIMIT`. A customer measured the tag present on **0 of 61,160** render spans.

**Root cause.** Drupal 11.3.0 switched `ThemeManager::render()` to a theme-engine *service* (`[$engine, 'renderTemplate']`), falling back to the global `{engine}_render_template()` only when no service is registered. Core ships a Twig service, so the global is never called — but it is still *defined*, because `ThemeInitialization` unconditionally includes `twig.engine`. So `function_exists()` returned true, the integration installed a hook on every render, the hook never fired, and it was never removed.

A second defect, on **every** Drupal version: `render()` can return early (theme hook not found, exception) without calling the render function. Those hooks leaked and then fired on *later* renders, overwriting unrelated spans' tags — measured 19 spans mis-tagged in a 30-render test.

# Changes

- Hook `ThemeEngineInterface::renderTemplate` once at `init()` — covers Twig, contrib engines, and module templates in one install. The service path passes the template path without its extension, so `.html.twig` is re-appended for Twig to keep the tag value identical to pre-11.3.
- Install the legacy `{engine}_render_template` hook once per engine instead of once per render, seeded on `twig_render_template` at `init()`. This also fixes a pre-existing gap: core falls back to `twig_render_template` when `{engine}_render_template` is absent, which was previously untagged.
- Create the `drupal.theme.render` span from `install_hook` + `$hook->span()` rather than `trace_method`. `$hook->span()` is per-invocation, so the frame's own span is known directly — no `active_span()` lookup and no guessing which frame a tag belongs to.
- Bail early when `dd_trace_tracer_is_limited()`. Unlike `trace_method`, `install_hook` callbacks are not limit-gated, so without this the whole body ran on every render past the span limit (measured 200x on a 200-render probe).

# Reviewer notes

- `allowNestedHook()` is not needed: it escapes reentrancy initiated *from inside a callback*, not the hooked function recursing. `install_hook` holds `running` only across the callback, so it is already recursive — measured 7/7 begins on a nested render chain, versus 1/1 for `trace_method` with `recurse => false`.
- Tag attribution is by frame, not by span name — an ancestor render span carries the same name under recursion, so a name check would mis-attribute a nested template to its parent.

### DIFF
--- a/src/DDTrace/Integrations/Drupal/DrupalIntegration.php
+++ b/src/DDTrace/Integrations/Drupal/DrupalIntegration.php
@@ -259,8 +259,11 @@ class DrupalIntegration extends Integration
             [
                 'recurse' => true,
                 'prehook' => function (SpanData $span, $args) use (&$renderHookIds) {
-                    // Reset first, so a stale entry left by a dropped span can never be reused.
+                    // A dropped span skips the posthook, so its hook may still be installed.
                     $spanKey = \spl_object_hash($span);
+                    if (!empty($renderHookIds[$spanKey])) {
+                        remove_hook($renderHookIds[$spanKey]);
+                    }
                     $renderHookIds[$spanKey] = 0;
 
                     $span->name = 'drupal.theme.render';
--- a/src/DDTrace/Integrations/Drupal/DrupalIntegration.php
+++ b/src/DDTrace/Integrations/Drupal/DrupalIntegration.php
@@ -228,14 +228,12 @@ class DrupalIntegration extends Integration
             }
         );
 
-        // The span of the ThemeManager::render frame currently executing, or null when that
-        // frame has none of its own. An ancestor render span carries the same name under
-        // 'recurse' => true, so the tag target can only be matched by identity.
+        // The span of the executing ThemeManager::render frame, or null when it has none of its own.
+        // Ancestors share its name under 'recurse' => true, so it can only be matched by identity.
         $renderSpan = null;
 
         $tagTemplateFile = static function ($file) use (&$renderSpan) {
-            // A nested render whose span was dropped leaves active_span() on its ancestor,
-            // which would otherwise take the inner template.
+            // A nested render whose span was dropped leaves active_span() on an ancestor.
             if ($renderSpan && $renderSpan === active_span()) {
                 $renderSpan->meta['drupal.template.file'] = $file;
             }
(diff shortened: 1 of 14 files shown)


src/DDTrace/Integrations/Drupal/DrupalIntegration.php
```diff
@@ -228,12 +228,44 @@ static function ($registry, $scope, $args) {
...
+
         trace_method(
             'Drupal\Core\Theme\ThemeManager',
             'render',
             [
                 'recurse' => true,
-                'prehook' => function (SpanData $span, $args) {
+                'prehook' => function (SpanData $span, $args) use (&$renderHookIds) {
+                    // A dropped span skips the posthook, so its hook may still be installed.
+                    $spanKey = \spl_object_hash($span);
```
bwoebi: spl_object_hash is deprecated since PHP 8.6. You can simply use spl_object_id.
You might want to simply use install_hook() to transfer the hook id via $hook->data from begin to end.

src/DDTrace/Integrations/Drupal/DrupalIntegration.php
```diff
@@ -339,6 +349,47 @@ static function (HookData $hook) use ($span) {
             ]
         );
 
+        // Drupal <= 11.2 renders through the {engine}_render_template() global; unlike the tracing
+        // posthook, this end hook also runs for a dropped or span-limited render.
+
+        // Must follow the trace_method: begin hooks run in install order, so active_span() is our own span.
+        install_hook(
+            'Drupal\Core\Theme\ThemeManager::render',
```
bwoebi: Why did you not move over the existing trace_method as well?

Also, doesn't this need a `$hook->allowNestedHook()`, to mirror the recursiveness?

Leiyks: allowNestedHook is a no-op here since it only matters when the hooked function is called from inside its own callback, which we are not doing

For the existing trace_method there is no reason, just updated it 👍

bwoebi (approved):
Looks right to me now! Let's merge this.

PR #4146, opened 2026-08-27
fix(ci): clean sessions between test runs

bm1549:
### Description

The live-debugger span probe tests set request-replayer session tokens in PHPT `--INI--` sections. PHP's test runner does not apply those per-test INI values to `--CLEAN--`, so cleanup cleared the unscoped session and left remote-config state behind. CI runs the extension suite twice, and the second pass could load stale probes before the test body reset the session.

Move the two affected tokens into `--ENV--` so the test and its cleanup process use the same scoped session.

Validation:
- 20 consecutive PHP 8.0 repetitions of both tests in parallel: 40/40 passed
- complete PHP 8.0 live-debugger suite with 12 workers: 14/14 passed

### Reviewer checklist
- [ ] Test coverage seems ok.


bwoebi (approved):
This is absolutely correct. Ugh, thanks!

leoromanovsky (approved):
Thank you for iterating on this

PR #4149, opened 2026-08-28
Unify test frameworks. Step 01 Laravel8

estringana:
### Why this PR exists

Now when we add appsec coverage to a new framework, we maintain two separate copies of it:

- tests/Frameworks/<Framework>/Version_X/ — used by tracer integration tests
- appsec/tests/integration/src/test/www/<framework>/ — a duplicate used by appsec integration tests

This made sense historically but we are in the same repository now moving into a more unified experience.

Due to this duplication, whenever we have to test something in addition to duplicate the framework, we have to duplicate the endpoints. Like it happened with all authentication features.

#### The proposal

Retire the appsec-side duplicates and point the appsec integration test container at the canonical copy under `tests/Frameworks/`. Any routes or fixtures that appsec tests need get added directly to the shared framework. The tracer tests don't  care about the extra routes, and appsec gets a richer, always-current environment for free.

####  This PR: Laravel 8x as the first step

- Removes the standalone `appsec/tests/integration/src/test/www/laravel8x/` copy
- Adds the appsec-specific routes and fixtures (login, register, dynamic-path, endpoint discovery) to `tests/Frameworks/Laravel/Version_8_x/`
- Updates `Laravel8xTests.groovy` to mount the shared path and target PHP 8.1 (matching the version the tracer already tests against)
- All tracer `test_web_laravel_8x jobs` and all appsec integration test jobs pass

#### Next steps
- Symfony 6.2
- Laminas MVC 3.3
- WordPress 6.1
- Roadrunner
  
### Not doing
- llm
- payment
- frankenphp

### Reviewer checklist
- [ ] Test coverage seems ok.


appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laravel8xTests.groovy
```diff
@@ -209,12 +208,11 @@ class Laravel8xTests {
             endpoints.size() > 0
         })
 
-        assert endpoints.size() == 6
+        assert endpoints.size() == 27
```
Leiyks, line 211: Probably best to just assert the number of endpoint is not 0 here ?
just in case someone add a new route later on

tests/Frameworks/Laravel/Version_8_x/initialize.sh
```diff
@@ -0,0 +1,34 @@
...
+cd /var/www
+
+export DD_TRACE_CLI_ENABLED=false
+
+# Clear stale bootstrap cache built with dev dependencies (e.g. facade/ignition).
+# Without this, artisan fails with "Class not found" for removed dev providers.
+rm -f /var/www/bootstrap/cache/*.php
+
+# The vendor/ directory is pre-installed on the host and mounted via the overlay.
+# Run composer install to regenerate the autoloader for the container's PHP version.
```
Leiyks: Not true anymore with those changes no ?

Leiyks (approved):
lgtm overall 👍

appsec/tests/integration/src/test/groovy/com/datadog/appsec/php/integration/Laravel8xTests.groovy
estringana: I thought about it but since we are testing the routes are extracted properly, I'd keep it.

tests/Frameworks/Laravel/Version_8_x/initialize.sh
estringana: I amended a bit the comments

cataphract (approved):

PR #4151, opened 2026-09-01
[PROF-15825] Add socket I/O profiling correctness test

realFlowControl:
### Description

Adds a self-contained profiler correctness test for socket reads and writes. It uses an ephemeral loopback TCP port, waits for readable data with `stream_select()`, and performs 8 KiB writes from two call sites in a 1:2 ratio. The expectations verify the 33%/66% stack split and the calculated total of `(4096 + 8192) * 8192 = 100663296` bytes with a 1% margin for both socket reads and writes. The prof-correctness analyzer passes against a locally generated profile.

[PROF-15825](https://datadoghq.atlassian.net/browse/PROF-15825)

### Reviewer checklist
- [ ] Test coverage seems ok.

[PROF-15825]: https://datadoghq.atlassian.net/browse/PROF-15825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


morrisonlevi (approved):
Error margins might be a bit too tight, if we start flaking I think we would be fine to relax a little.

PR #4159, opened 2026-09-04
fix(profiling): IO/allocation sampling intervals

realFlowControl:
### Description

I/O and allocation sampling used Poisson-distributed distances, but upscaling assumes exponential distances with inclusion probability `p = 1 - exp(-size / interval)`. Both PHP allocator implementations share the allocation sampler.

This can overestimate totals by about 50%. For 110 KiB operations at the 100 KiB interval, the expected sampling probability is about 67%, but the old I/O sampler picks nearly every operation while upscaling still multiplies by about 1.50. Allocations near the 4 MiB interval have the same problem.

Use one shared exponential interval sampler for I/O and allocation profiling and retain the existing NTS/ZTS RNG choices. Round distances up to whole units with a minimum of one, and sample I/O when the counter reaches zero. Seeded allocation tests cover the default interval and tiny sampling distances; an I/O regression test covers the exact boundary.

Stabilize the I/O correctness tests with larger workloads, 6% byte margins, and 3-percentage-point socket-share margins. The larger workloads produce about 15,000 samples per metric, with roughly 0.8% relative sampling noise.

Strengthen the allocation correctness workload to 512 fixed iterations and check its actual totals: 18,874,368,000 payload bytes and 2,048 allocations. Use 6% total margins and 3-percentage-point share margins. The combined allocation/time test runs at least 128 iterations with 6-percentage-point allocation-share margins.

Keep the generator stack test's allocation-byte share assertions unchanged and remove its unstable allocation-count share assertions. Rare small allocation samples can dominate upscaled counts even when generator frames are correct. Allocation-count accuracy remains covered by the dedicated allocation test.

The old allocation percentage-only tests passed despite inflated totals. With the strengthened test, master reports about 13.78% excess bytes and 17.62% excess allocations and fails 100/100 runs.

Validated with PHP 8.5 NTS and the pinned dependencies: all 29 Rust unit tests pass. The tiny-allocation and I/O boundary regressions both fail without the rounding and boundary fixes.

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (approved):
I think `AllocationProfilingStats` and `IOProfilingStats` should get merged. Leaving this for a future PR though, as it's a bit invasive.

PR #4160, opened 2026-09-04
fix(profiling): use exponential allocation sampling

realFlowControl:
### Description

Apply the sampling fix from #4159 to allocation profiling. The sampler used Poisson-distributed distances, but upscaling assumes exponential distances with inclusion probability `p = 1 - exp(-size / interval)`. Allocations near the 4 MiB interval can be overestimated by about 50%. Both PHP allocator implementations share this sampler.

Use exponential distances, retain the existing NTS/ZTS RNGs, and add a seeded regression test. Change the allocation correctness workload to 512 fixed iterations and check its actual totals: 18,874,368,000 payload bytes and 2,048 allocations. Use 6% total margins and 3-percentage-point share margins. The combined allocation/time test runs at least 128 iterations with 6-point allocation-share margins. Test notes document the binomial sampling-noise estimates and integer percentage truncation.

The old percentage-only tests passed on master despite inflated totals. With the strengthened allocation test, master reports about 13.78% excess bytes and 17.62% excess allocations and fails 100/100 runs.

### Reviewer checklist
- [ ] Test coverage seems ok.


realFlowControl:
Merged into #4159.

PR #4164, opened 2026-09-07
Fix SidecarFeaturesDisabledTests regression from sidecar appsec integration

estringana:
## Summary

After #3725 embedded appsec into the sidecar, `SidecarFeaturesDisabledTests > appsec is disabled and sidecar is not launched` started failing intermittently.

The test was killing the sidecar with `kill -9` and then restarting Apache. The problem: existing Apache workers hold a reconnect callback (registered in `ddog_sidecar_connect_php`). When the sidecar is killed, the reconnect mechanism immediately spawns a new sidecar subprocess — before Apache restarts. That new subprocess becomes an orphan (its parent workers die during the restart), and `pgrep -f datadog-ipc-helper` finds it, failing the assertion.

Fix: restart Apache first so the old workers — and their reconnect callbacks — are torn down cleanly. Only then kill any remaining sidecar. Since the new workers start with `datadog.appsec.enabled=false`, they don't connect to the sidecar and no reconnect will be triggered.


cataphract (commented):
I think it would be preferable to make the shutdown messages not try to restart sidecar. So, do not retry in that case.

cataphract (approved):

Issue #4163, opened 2026-09-07
[Bug]: FrankenPHP still exits 137 with no shutdown logs on musl/amd64 — #3651 reproduces on 1.25.0


cattz:
### Bug report

[#3651](https://github.com/DataDog/dd-trace-php/issues/3651) is closed as fixed in 1.17.0 by
[#3662](https://github.com/DataDog/dd-trace-php/pull/3662). On **musl/amd64** the same failure is
still present, and reproduces identically on every version we tested including **1.25.0**.

The symptom differs from the original report in one important way: there is **no
`fatal error: non-Go code disabled sigaltstack` message**. The process simply never shuts down.
Caddy's `signal.Notify` never fires, `trapSignalsPosix` never runs, and not a single shutdown log
line is emitted — not even `shutting down apps, then terminating`, which Caddy writes *before* any
draining begins. The container runs to the end of its termination grace period and is SIGKILLed
(exit 137).

Because #3651 was closed and the loud error is gone, this reads as "fixed" from the outside. It
cost us about two weeks: we misdiagnosed it as a Caddy `grace_period` sizing problem and shipped
two changes that could not possibly have worked, before finding the real cause.

### Environment

- dd-trace-php **1.25.0, 1.24.2, 1.17.0 and 1.16.0** — all identical (see matrix)
- FrankenPHP **1.12.4**, PHP **8.3 ZTS**, `dunglas/frankenphp:1.12.4-php8.3-alpine`
- **Alpine / musl**, extension `ddtrace-20230831-alpine-zts.so`, **amd64**
- `ENTRYPOINT ["tini", "--", "/docker-entrypoint.sh"]`, entrypoint ends in `exec frankenphp run …`
- Production: EKS Auto Mode, Bottlerocket 2026.8.26, containerd 2.2.7, kubelet v1.34.9, amd64 nodes

### Version matrix

Same base image, tracer enabled, one HTTP request served, then `docker stop -t 30`:

| dd-trace-php | stop time | exit code | Caddy shutdown log lines |
|---|---|---|---|
| 1.25.0 (latest) | 30s | 137 | 0 |
| 1.24.2 | 31s | 137 | 0 |
| **1.17.0 (the #3662 release)** | **30s** | **137** | **0** |
| 1.16.0 (pre-fix) | 31s | 137 | 0 |

1.17.0 behaves exactly like 1.16.0 here, so this is **not a regression introduced after the fix** —
[...]

PR #4167, opened 2026-09-07
Fix #4163: broken frankenphp shutdown signal on musl

bwoebi:
That's ugly with the manual assembly trampolines, but seems it's needed. Boo musl, boo.

Also making sure that we *actually* run the frankenphp tests, which are skipped under NTS...

### DIFF
--- a/.gitlab/generate-package.php
+++ b/.gitlab/generate-package.php
@@ -2,7 +2,7 @@
 
 include "generate-common.php";
 
-const FRANKENPHP_ALPINE_PHP_VERSION = "8.5";
+const FRANKENPHP_ALPINE_PHP_VERSION = "8.3.12";
 
 $build_platforms = [
     [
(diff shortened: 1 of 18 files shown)


tabgok (approved):
approved from IDM

morrisonlevi (approved):
I am incredibly skeptical of the assembly, which I did not actually review as my assembly skills are far too weak for such a long function (for me). I'm also skeptical of the technique. Musl blocks `clone(2)` for safety reasons, and we're brute forcing it anyway.

Nonetheless, these are used on `SIGINT` and `SIGTERM` handlers which are going to shutdown the process eventually _anyway_ so if we get this wrong... yes, it's not great, but a crash there is at least constrained to something that was going to shutdown already. Additionally, there are tests in the PR which do seem to demonstrate it to at least be minimally safe (it's not immediately blowing anything up). 

The only other idea I have is to delay the signal handling until a "safe point" such as a combination of PHP lifecycle handlers and vm_interrupt, similar to profiling. The problem is that if you are between requests, then the next safe point is effectively indefinite, so this would block the `SIGINT` from doing its job.

PR #4171, opened 2026-09-08
ci: fix Debian 11 EOL apt breakage, add trixie coverage

Leiyks:
Debian 11 (bullseye) is EOL and its apt repos broke, taking master red: in pipeline 135955234, 66 of 67 failed jobs were bullseye. `bookworm` is the control — 22/22 bookworm `verify debian` cells pass while 22/22 bullseye cells fail on identical scripts, only `$IMAGE` differing.

Two independent failures, so relaxing apt date checks alone is not enough: `bullseye-security` `InRelease` expired 2026-09-07T21:13:05Z, and the `deb11uN` pool objects are incompletely served. Over 40 random objects: `deb.debian.org` 13/40, `security.debian.org` 14/40, `snapshot.debian.org@20260901T000000Z` 40/40; `archive.debian.org` has no `bullseye-security` at all. Snapshot is the only complete source, and the pin is an end state, not a point in the past — `bullseye-security` is terminally frozen and `snapshot@20260901` is byte-identical (`cmp`) to the final live index, so there is no date to maintain.

Debian 11 stays in the matrix as a supported customer install target; `debian:trixie-slim` is added alongside.

### DIFF
--- a/.gitlab/generate-package.php
+++ b/.gitlab/generate-package.php
@@ -1277,7 +1277,7 @@ endforeach;
 
 .system_tests:
   stage: verify
-  image: registry.ddbuild.io/images/mirror/python:3.12-slim-bullseye
+  image: registry.ddbuild.io/images/mirror/python:3.12-slim-bookworm
   tags: [ "docker-in-docker:amd64" ]
   variables:
     TEST_LIBRARY: php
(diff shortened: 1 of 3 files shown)


bwoebi (approved):
Looks good to me, thanks for salvaging our CI :-D

PR #4177, opened 2026-09-09
feat(otel): otel sampling support in distributed tracing

vpellan:
### Description

Add support for OTel sampling in distributed tracing, by reading and emitting ot.th and ot.rv tags. These tags will be forwarded to services using OTel SDKs, that should forward them to the backend.

Distributed tracing will work between DD-instrumented services and OTel-instrumented services, downstream OTel services will be able to forward the th field it to the collector and backend, which will calculate metrics correctly

Locally passes https://github.com/DataDog/system-tests/pull/7518

### Reviewer checklist
- [ ] Test coverage seems ok.


tracer/priority_sampling/priority_sampling.c
```diff
@@ -389,6 +389,11 @@ static void dd_decide_on_sampling(ddtrace_root_span_data *span) {
                              &limit_zv);
     }
 
+    span->otel_sampling_rate = sample_rate;
+    span->otel_sampling_decision = mechanism == DD_MECHANISM_MANUAL || mechanism == DD_MECHANISM_ASM || limited || sample_rate <= 0
+        ? DDTRACE_OTEL_SAMPLING_DECISION_NON_PROBABILITY
+        : DDTRACE_OTEL_SAMPLING_DECISION_PROBABILITY;
```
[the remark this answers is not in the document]
vpellan: The RFC says the limiter changes attribution only when it converts a keep into a drop.

tracer/otel_sampling.c
```diff
@@ -0,0 +1,528 @@
...
+  char generated_threshold[14];
+
+  if (decision == DDTRACE_OTEL_SAMPLING_DECISION_NON_PROBABILITY) {
+    fields.threshold = NULL;
+    fields.threshold_len = 0;
+  } else if (decision == DDTRACE_OTEL_SAMPLING_DECISION_PROBABILITY &&
+             !fields.random_value && !fields.threshold && sample_rate > 0) {
+    ddtrace_otel_generate_fields(&fields, generated_random_value,
+                                 generated_threshold, trace_id,
+                                 sampling_priority, sample_rate);
```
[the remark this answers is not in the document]
vpellan: [7d5421d](https://github.com/DataDog/dd-trace-php/pull/4177/commits/7d5421d727b0c60b8116bd84d44da0554584f2af)

tracer/otel_sampling.c
```diff
@@ -0,0 +1,528 @@
...
+
+  smart_str limited = {0};
+  size_t member_count = 0;
+  if (members.datadog_member) {
+    ddtrace_otel_append_member(&limited, members.datadog_member,
+                               members.datadog_member_len, &member_count);
+  }
+  if (members.otel_member) {
+    ddtrace_otel_append_member(&limited, members.otel_member,
+                               members.otel_member_len, &member_count);
```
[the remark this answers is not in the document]
vpellan: [6841cfb](https://github.com/DataDog/dd-trace-php/pull/4177/commits/6841cfb7e71bdd06900df59a5ca4ce75152b4ae0)

tracer/distributed_tracing_headers.c
```diff
@@ -464,14 +465,19 @@ static ddtrace_distributed_tracing_result ddtrace_read_distributed_tracing_ids_t
                 }
 
                 last_comma = *ptr == ',';
-                // preserve only up to 31 vendor specific values, excluding our own
-                if (last_comma && ++commas == 30) {
+                // W3C Trace Context permits at most 32 list-members. Owned members
+                // are removed or normalized after this pass, then protected when
+                // the outbound header is rebuilt.
+                if (last_comma && ++commas == 32) {
```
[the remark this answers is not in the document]
vpellan: [cbed965](https://github.com/DataDog/dd-trace-php/pull/4177/commits/cbed965e8b5632b1b4f6f88f9d55e107d8399404)

tracer/distributed_tracing_headers.c
```diff
@@ -464,14 +465,19 @@ static ddtrace_distributed_tracing_result ddtrace_read_distributed_tracing_ids_t
...
                     --persist;
                     break;
                 }
             }
             *persist = 0; // and zero-terminate it
             ZSTR_LEN(result.tracestate) = persist - ZSTR_VAL(result.tracestate);
+            zend_string *normalized_tracestate = ddtrace_otel_sampling_update_tracestate(
+                result.tracestate, trace_id.low, result.priority_sampling, DDTRACE_OTEL_SAMPLING_DECISION_INHERITED, 0);
```
bwoebi: It's odd to me to do a two-step normalization here. There's no reason why the otel parsing cannot be inlined in the loop above, producing a single normalization pass.

bwoebi: `result.tracestate` is supposed to contain the values untouched by us - i.e. other vendors, which get appended (as it should - new values should come first), `ddtrace_otel_sampling_update_tracestate` however manipulates the string in place every time it touches it.
Let's please not store individual data (like the otel rv and th) inside the string after parsing. Store it here on `result`, make it properly available on the span (for inspection / manual changing / interaction with otel sdk), and rebuild it alongside the dd= tags in ddtrace_format_tracestate.

bwoebi: It should also help keeping the implementation tidier, keeping scanning separate from formatting.

bwoebi: Also, in fact, you're storing `otel_sampling_rate` on the span, but never storing the parsed values there, which strikes me as odd.
Because, in fact, the `otel_sampling_rate` seems to always be equal to the datadog sampling rate (making it redundant as is??)

tracer/otel_sampling.c
```diff
@@ -0,0 +1,567 @@
...
+  if (separator_len) {
+    smart_str_appendc(result, ';');
+  }
+  smart_str_appendl(result, field, field_len);
+}
+
+static uint64_t ddtrace_otel_threshold_for(double sample_rate) {
+  if (sample_rate >= 1) {
+    return 0;
+  }
```
bwoebi, line 100: Redundant condition with the if just below?

tracer/otel_sampling.c
```diff
@@ -0,0 +1,567 @@
...
+                                         char threshold[14], uint64_t trace_id,
+                                         zend_long sampling_priority,
+                                         double sample_rate) {
+  uint64_t threshold_value = ddtrace_otel_threshold_for(sample_rate);
+  uint64_t random_value_int = ddtrace_otel_reconcile_random_value(
+      ddtrace_otel_derive_random_value(trace_id), threshold_value,
+      sampling_priority > 0);
+
+  ddtrace_otel_encode_56_bit_hex(random_value_int, random_value);
+  ddtrace_otel_encode_56_bit_hex(threshold_value, threshold);
```
bwoebi: If you'd print it directly to the smart_str, you could use `smart_str_append_printf(&result, "rv:%014x", random_value)` instead of reimplementing hex-printing. (and then remove trailing zeroes on the threshold)

tracer/distributed_tracing_headers.c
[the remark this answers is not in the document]
bwoebi: This doesn't really need to be increased to 32, we just should handle "ot" as well and not increase commas for the "ot" case.

tracer/distributed_tracing_headers.c
vpellan: Fixed in 89ab3d46d:
- it now parses and removes `ot` during the existing normalization pass.
- parsed OTel fields are stored separately on the trace/span and rebuilt during formatting.
- extraction now parses state, while injection handles formatting.
- parsed values are stored on the span and the redundant otel_sampling_rate was removed.

tracer/otel_sampling.c
vpellan: Yes, it was redundant and has been removed in 189d449da

tracer/otel_sampling.c
vpellan: Done in af34e5782; sampling fields are now formatted directly with smart_str_append_printf

tracer/distributed_tracing_headers.c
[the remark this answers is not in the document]
vpellan: Fixed in 53e88b181; dd and ot are excluded from the 30 vendor-member count so both owned slots remain available.

tracer/otel_sampling.h
```diff
@@ -0,0 +1,27 @@
...
+#include <php.h>
+
+#define DDTRACE_OTEL_MAX_VALUE_LEN 256
+
+typedef struct {
+    uint64_t random_value;
+    uint64_t threshold;
+    size_t unknown_fields_len;
+    uint8_t random_value_len;
+    uint8_t threshold_len;
```
bwoebi: I don't think these two fields make sense. You're effectively using them as sentinel values - and otherwise they're hardcoded to 14.
Why not put a sentinel value into random_value/threshold instead? (e.g. `(uint64_t)-1`).

bwoebi: Except for threshold len, which you're computing, but at parse time ... i.e. if anything would change it at runtime, it breaks.

tracer/otel_sampling.c
```diff
@@ -0,0 +1,345 @@
...
+    smart_str_0(&vendors);
+    return vendors.s;
+  }
+  return zend_string_init("", 0, 0);
+}
+
+void ddtrace_otel_sampling_append_to_tracestate(smart_str* tracestate, const ddtrace_otel_sampling_state* state) {
+  smart_str value = {0};
+  if (state->random_value_len) {
+    smart_str_append_printf(&value, "rv:%0*" PRIx64, state->random_value_len, state->random_value);
```
bwoebi: This is always 14, per spec. so 
```suggestion
    smart_str_append_printf(&value, "rv:%014" PRIx64, state->random_value);
```

tracer/otel_sampling.c
```diff
@@ -0,0 +1,345 @@
...
+  }
+  if (state->threshold_len) {
+    if (value.s) {
+      smart_str_appendc(&value, ';');
+    }
+    smart_str_append_printf(&value, "th:%0*" PRIx64, state->threshold_len, state->threshold);
```
bwoebi:
```suggestion
    if (state->threshold) {
        smart_str_append_printf(&value, "th:%014" PRIx64, state->threshold);
        while (ZSTR_VAL(value.s)[ZSTR_LEN(value.s) - 1] == '0') {
            --ZSTR_LEN(value.s);
        }
    } else {
        smart_str_appends(&value, "th:0");
    }
```

tracer/otel_sampling.c
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    member = member_end == end ? end : member_end + 1;
+  }
+
+  if (vendors.s) {
+    smart_str_0(&vendors);
+    return vendors.s;
+  }
+  return zend_string_init("", 0, 0);
```
bwoebi:
```suggestion
  return ZSTR_EMPTY_ALLOC();
```

tracer/otel_sampling.c
bwoebi: Did this commit somehow not land on this branch?

PR #4179, opened 2026-09-10
feat!: combined tracer+profiler in ddtrace.so

morrisonlevi:
### Description

This PR removes `datadog-profiling.so` and moves the profiler into `ddtrace.so`. This can affect customer code if they are checking `extension_loaded("datadog-profiling")` or similar. It can also affect build systems and management if it moves or alters INI entries, such as making assumptions that not loading `datadog-profiling.so` will mean the profiler isn't loaded. The `datadog-setup.php` script tries to understand some of these things and disable the profiler, but it's not guaranteed.

#### Motivation

 - Theoretically, there's a size benefit. This has not been realized yet, and actually things have gotten bigger. Personally, I am not sure if that should block merging this; this is a large PR that will have a lot of conflicts, so we may wish to merge it even if we regress, and work on size later. But honestly this was one of the drivers, if we fail to meet this, then we've failed, and maybe we shouldn't ship this!
 - Less cross-object dependencies, as the tracer and profiler are now in the same `.so`.
 - TODO
 
 #### Testing
 
This adds a task to `Loader test on <arch> libc` jobs which ensures that we have a `wall-time` sample with the expected stack for SSI. It tries to avoid dependencies and workflows which can introduce flakiness, so it uses the version of Python built into the image, uses libzstd rather than a Python package for it, the profile is written locally to disk, and so on. It piggy-backs onto the loader test instead of a new job because it takes 2+ minutes to pull images and git repos and just mere seconds to run the test.

### Reviewer checklist
- [ ] Test coverage seems ok.


ext/configuration_tags.c
```diff
@@ -0,0 +1,75 @@
+// Copyright 2026-Present Datadog, Inc. https://www.datadoghq.com/
+// SPDX-License-Identifier: Apache-2.0
+
```
bwoebi, line 3: Btw. we have a default license for the repo and don't add these individually.

ext/configuration_shared.h
bwoebi, line 1: We definitely need to find a way to not do this.
Fine for the initial draft to test it, but this is not something I'm willing to accept. Also, like "preserve its historical ID ordering" wtf :-D

bwoebi:
Can we ensure that there's at least one loader test with a proper stack walk producing an usable profile with SSI?

ext/configuration_shared.h
morrisonlevi: Okay, I'll await your experiments that you'd said you'd perform for this (from slack conversation).

PR #4180, opened 2026-09-10
fix(appsec): musl ZTS load + ZTS thread-exit teardown; ci: coverage disk envelope

Leiyks:
### Description

One commit per change; they share no code and can be reviewed separately.

**`fix(appsec)`: weakly reference `__cxa_thread_atexit_impl`.** musl exports no such symbol (it is a glibc 2.18 internal) and resolves relocations eagerly, so the strong reference made `dlopen("ddappsec.so")` fail outright. Every ZTS build emits it, so the failure set is `musl ∩ ZTS` — **already shipped in 1.25.0 and 1.25.1, where AppSec is therefore fully non-functional on Alpine/FrankenPHP ZTS, so this needs to be in the next release.**

**`fix(appsec)`: pthread fallback for the thread-exit destructor.** Without that symbol nothing tears non-main threads down, so GINIT falls back to a `pthread_key_create` destructor and a `ZEND_TLS` flag lets whichever of GSHUTDOWN-on-the-owning-thread or thread-exit comes first do the work. That ordering also fixes a segfault reproducible on unmodified master (`_tshutdown_handler` → `dd_log_level` → `zai_config_get_value` after PHP freed the thread's config; `pecl parallel`, 3 `Runtime`s then `close()`), which the old `registered_thread_local_dtor` handshake caused by leaving thread-exit as the only teardown path on glibc ZTS.

**`ci`: disk envelope for `helper-rust integration coverage`.** The job never set `DOCKER_LOOPBACK_SIZE`, so it ran on the DinD default of ~20G; the added `df` probes measure the volume going 3.2G → 19.9G across the build, leaving ~0.1G of headroom there, hence the alternation between `No space left on device` and `Bus error`. Raised to 50G, and the job now passes.

**Test infrastructure (separate author).** Portable `dd_library_loader` in the appsec integration tests, a runtime `gnu_get_libc_version` probe replacing the compile-time `__MUSL__` so one loader build serves both libcs, and an Apache ZTS musl AppSec test image so this combination is covered at all; a final commit drops the now-dead `__MUSL__` detection.

**`installer tests` stays red.** It installs the *published* tarball for `$(cat VERSION)` and its `needs` supply only `*-unknown-linux-gnu` bundles, so the musl-ZTS sub-test keeps pulling 1.25.1 whatever master contains; only a release carrying the first fix clears it.


appsec/src/extension/ddappsec.c
```diff
@@ -195,15 +195,19 @@ static PHP_GINIT_FUNCTION(ddappsec)
...
-            void (*func)(void *), void *arg, void *dso_handle);
-        __cxa_thread_atexit_impl(_tshutdown_handler, NULL, __dso_handle);
+        // Weak because musl exports no __cxa_thread_atexit_impl (it is a glibc
+        // 2.18 internal); a strong reference makes dlopen() of this DSO fail.
+        extern int __cxa_thread_atexit_impl(void (*func)(void *), void *arg,
+            void *dso_handle) __attribute__((weak));
+        if (__cxa_thread_atexit_impl) {
+            __cxa_thread_atexit_impl(_tshutdown_handler, NULL, __dso_handle);
+            ddappsec_globals->registered_thread_local_dtor = true;
+        }
```
cataphract, line 236: Isn't this just going to leak memory? I think we should provide a fallback implementation if `__cxa_thread_atexit_impl`, like libc++ and stdlibc++ do for `__cxa_thread_atexit` when `__cxa_thread_atexit_impl` is unavailable (using pthread facilities).

cataphract:
I like it now. Next week I'll also add an appsec integration tsts against musl zts so this doesn't recur.

appsec/src/extension/ddappsec.c
Leiyks: You are right, pushed a proper fix.
`helper-rust-integration integration coverage` job is still red tho, I am checking

cataphract (approved):

PR #4182, opened 2026-09-10
fix appsec int tests fail on frankenphp

cataphract:
### Description

After the appsec image upgrade, the builds of ddappsec.so started targeting a version of glibc too recent for the frankenphp image.

* In appsec integration tests, build ddappsec.so portably. No separate musl/glibc builds now, like already happened for the tracer and libdatadog/components-rs
* Do away with elf patching which used to remove the libc dependency. The new toolchain image adds a libc dependency that need not be removed
* Fix some task input mismatches in build.gradle

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (approved):
Looks good to Claude, and it does fix the tests it aims to. I don't quite grasp it, but the way it makes it portable is instead of patching away glibc symbols, it builds them on musl to begin with. Something like that.

cataphract:
@morrisonlevi 

> Looks good to Claude, and it does fix the tests it aims to. I don't quite grasp it, but the way it makes it portable is instead of patching away glibc symbols, it builds them on musl to begin with. Something like that.

Sorry, I gave a truncated explanation, but to be clear what fixes it is to build ddappsec against a patched musl. With some previous toolchains, doing so would leave a NEEDED entry in the ELF for the musl shared library's explicit name, which would have to be patched away with patchelf in order to load it on glibc. However, the new toolchain doesn't require this anymore for reasons not worth going into. So I removed the step that would do the patching as well.

PR #4183, opened 2026-09-10
ci: use HTTPS for Alpine package repositories

morrisonlevi:
### Description

We started enforcing some egress restrictions in our CI. Older alpine images use `http` rather than `https` by default in their `apk` commands. This PR replaces `http://` with `https://` in `/etc/apk/repositories`. Let's see if this does the trick, as we should use `https://`!

This also fixes `Prepare code` which installs `unzip` which ends up using `http` in Debian packages. In this case we fix it by moving the base image from `php:8.2-cli` to `composer:2`. Note that `prepare code` (lowercase) already uses `composer:2` to run `make generate`. These jobs are in different pipelines but they're doing the same thing in those pipelines.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/.gitlab/generate-package.php
+++ b/.gitlab/generate-package.php
@@ -1106,6 +1106,10 @@ endforeach;
 <?php dockerhub_login() ?>
     - mkdir build
     - mv packages build
+    # Older Alpine images bake HTTP repository URLs into /etc/apk/repositories.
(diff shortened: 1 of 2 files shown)


morrisonlevi:
Didn't work for some reason, they just failed with IO error instead.

PR #4186, opened 2026-09-11
test: reduce sidecar testing flakes

morrisonlevi:
### Description

This fixes two different sidecar flakes.

1.  `inWebServer()` retries the entire server lifecycle when no traces are received, but `testComposerInteroperabilityWhenNoInitHook` expects no traces. It then kills a healthy PHP server and immediately restarts it on the same port. With telemetry enabled, the extension can start a sidecar that briefly outlives the forcibly stopped PHP process and may retain the server socket. The unnecessary retry then fails with Server never came up. Fixed by setting `retries=0` for this test. There are no other tests I could find that also assert that no traces are made.
2. Fixes a race in the `agent_sampling_sidecar.phpt` caused by matching stale shared-memory data from earlier or parallel test processes. Each invocation now uses unique sampling configuration markers, ensuring
 it waits for its own sidecar updates. Failures also report concise, actionable messages instead of dumping large shared-memory buffers.

### Reviewer checklist
- [ ] Test coverage seems ok.

### DIFF
--- a/tests/ext/background-sender/agent_sampling_sidecar.phpt
+++ b/tests/ext/background-sender/agent_sampling_sidecar.phpt
@@ -18,6 +18,9 @@ datadog.trace.agent_test_session_token=background-sender/agent_sampling_sidecar
 <?php
 include __DIR__ . '/../includes/request_replayer.inc';
 
+$contents = [];
+$filename = null;
+
 // Wait until the sidecar has published this test invocation's response. Each
 // invocation uses random markers so a repeated or parallel run cannot match a
 // stale shared-memory file left by an earlier process.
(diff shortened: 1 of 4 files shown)


tests/ext/background-sender/agent_sampling_sidecar.phpt
bwoebi, line 1: Is there a specific reason you dropped the debuggability of this test? Otherwise probably fine.

morrisonlevi: I guess you consider hexdumps more debuggable than messages? xD
I can revert that aspect, sure.

bwoebi: Why not both :-P

bwoebi (approved):

PR #4191, opened 2026-09-14
chore(agents): add starter dd-trace-php overrides for dd-apm-sdk-review

robertomonteromiguel:
### What does this PR do?

Adds the **dd-trace-php layer** on top of the shared skill copy.

- `.agents/dd-apm-sdk-review-overrides/` — one starter PHP reviewer fact: userland follows PSR-2 / `composer lint` (fix with `composer fix-lint`), as written in `CONTRIBUTING.md` § "PHP linting"
- `AGENTS.md` — short hook (this repo had none): local agents run the skill; Codex reads `review-without-harness.md`
- `.llm-validation/` — first suite in this repo, one case (`php-conventions-psr2`), plus the GitLab `"llm validation"` include

Stacked on the core-copy PR (#4190). Shared skill source: [dd-apm-sdk-review-core#1](https://github.com/DataDog/dd-apm-sdk-review-core/pull/1).

### Motivation

This is the half that PHP reviewers should actually read. The repo had no agent hook or eval gate yet; this lands the minimum so the skill is usable.

### Additional Notes

**How to review**

- Start here: `.agents/dd-apm-sdk-review-overrides/`
- Then: the `AGENTS.md` Review Guidelines block
- Then: `.llm-validation/` if you care about the gate
- Skip `.agents/skills/dd-apm-sdk-review/` — that is the parent PR / core repo

Merge with the parent skill-copy PR, not as a standalone.

### Reviewer checklist
- [ ] Test coverage seems ok.

Made with [Cursor](https://cursor.com)


robertomonteromiguel:
@codex review

robertomonteromiguel:
@codex review

robertomonteromiguel:
@codex review

.gitlab-ci.yml
```diff
@@ -17,6 +17,24 @@ variables:
...
+    LLMVAL_PLATFORM_PROJECT: "ddoghq/llm-validation-platform"
+    LLMVAL_PLATFORM_REF: *llmval_platform_sha
+    LLMVAL_BASE_REF: "master"
+    # Repo-wide default is recursive (libdatadog, etc.). JS/Java skip submodules
+    # on this job; inherit that so the gate does not clone GitHub before run.sh.
+    GIT_SUBMODULE_STRATEGY: none
+    # dd-trace-php's fabric-proxy 403s NodeSource/npm/dot.net; JS/Java SAs allow
+    # them. Bypass the proxy for the hosts run.sh needs (CIEXE-714 annotation).
+    # dot.net 301s to builds.dotnet.microsoft.com; the install script also uses aka.ms.
+    KUBERNETES_POD_ANNOTATIONS_NODESOURCE: "beta.fabric.datadoghq.com/no-proxy-additions=deb.nodesource.com,registry.npmjs.org,dot.net,builds.dotnet.microsoft.com,aka.ms"
```
bwoebi: Wrong place, these rules should be in https://github.com/ddoghq/dd-source/blob/main/domains/devex/ci/gitlab/apps/gitlab-runner-egress-acl/config/k8s/values/projects/dd-trace-php.yaml.

.github/CODEOWNERS
```diff
@@ -1,6 +1,11 @@
 # Default owners
 * @DataDog/apm-php @DataDog/apm-php-core
 
+# Agent review skill and eval suite (explicit so MergeGate shows an owner)
```
bwoebi: Mergegate doesn't work with the `*` rule above?

.agents/skills/dd-apm-sdk-review/SKILL.md
```diff
@@ -70,11 +75,24 @@ fi
...
-  # No PR yet, or gh failed to resolve one (e.g. a stacked branch with no PR open):
-  # do NOT silently fall back to origin/master. Stop and ask the human/agent to
-  # confirm the actual merge target before computing any diff or running reviewers.
-  echo "Could not resolve a PR base branch — what is the actual merge target for this branch (e.g. a parent feature branch on a stacked PR)?"
-  exit 1
+  # No PR yet (or gh could not resolve one). Infer the repo default branch so
+  # a local-only, not-yet-pushed branch still reviews. Never @{u}. Print the
+  # inferred target — a stacked branch against a parent feature can correct
+  # it. Only ask when none of those refs exist locally.
+  infer_remote="${DD_REMOTE:-origin}"
```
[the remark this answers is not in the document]
robertomonteromiguel: Fixed upstream in core `5b5b08d` and recopied here. No DataDog remote and no PR → ask; no fork-`origin` fallback.

AGENTS.md
```diff
@@ -0,0 +1,12 @@
+# AGENTS.md - dd-trace-php
+
+## Review Guidelines
+
+**Local agent with a skill harness:** Run the [dd-apm-sdk-review](./.agents/skills/dd-apm-sdk-review/) skill on demand when asked. It is not required before every push. If any
+`P0` issues are reported, you must either fix them or get explicit authorization from the human you
+are working with and record the unresolved finding verbatim in the PR description. `P1` and `P2`
```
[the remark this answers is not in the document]
robertomonteromiguel: Fixed. AGENTS.md (and the mirrored skill) now say to record location and class of issue only — never secret values, tokens, credentials, or exploit details. The Security lens stays out; Codex security owns that lane.

.agents/skills/dd-apm-sdk-review/reviewers/report-template.md
```diff
@@ -14,8 +14,14 @@ A reviewer that could not do its job reports `NOT VERIFIED (<reason>)` for its a
 
 ## Report format
 
+The **gate line is the first thing a human must see** — same line as the title, or the line immediately under it. Do not bury `DO NOT PUSH` / `WAITING ON HUMAN` / `READY TO PUSH` under empty headings.
+
+**Omit every section that has nothing to say.** An empty `## P0`, `## P1`, `## P2`, `## Not verified`, `## Checked and fine`, or `## Coverage gaps` heading is noise. If the only thing you would print is a placeholder, drop the section.
+
+Findings stay above the `---`. On `APPROVE_WITH_COMMENTS`, the P1/P2 list **and** the fix-or-dismiss question go **above** the `---`, right under the gate line — not after Rule files / Related skills.
```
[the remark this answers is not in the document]
robertomonteromiguel: Won't fix. The report title *is* the gate line. `Verdict` / `Target` are metadata under it, not a second gate.

robertomonteromiguel:
@codex review

.github/CODEOWNERS
robertomonteromiguel: Agreed — `*` already assigns `@DataDog/apm-php` and `@DataDog/apm-php-core`. Dropped the three duplicate lines. If MergeGate needs an explicit path after all, say so and we can put them back.

.gitlab-ci.yml
robertomonteromiguel: Agreed — leaving the annotation here for now. Opened a draft against the project ACL: https://github.com/ddoghq/dd-source/pull/93032 (`deb.nodesource.com`, `dot.net`, `builds.dotnet.microsoft.com`, `aka.ms`; `registry.npmjs.org` was already allowed). Once that lands we can drop `KUBERNETES_POD_ANNOTATIONS_NODESOURCE` from this PR.

bwoebi: Approved over there

robertomonteromiguel:
@codex review

.gitlab-ci.yml
robertomonteromiguel: Next 403 after Node/.NET: `dotnet restore` → `api.nuget.org` (NU1301, fabric-proxy 403). Added that host (and `nuget.org`) to the temporary annotation here, and to the ACL draft: https://github.com/ddoghq/dd-source/pull/93032. Still planning to drop `KUBERNETES_POD_ANNOTATIONS_NODESOURCE` once the ACL is deployed.

.agents/dd-apm-sdk-review-overrides/reviewers/conventions.md
```diff
@@ -0,0 +1,13 @@
+Override for `reviewers/conventions.md` (in the core skill folder) — read that file first, then this.
+
+# Conventions — dd-trace-php specifics
+
+This file starts with one confirmed pattern and should grow. Do not treat it as exhaustive.
+
+The source of truth for *how* style is checked is [`phpcs.xml`](../../../phpcs.xml) via `composer lint` / `composer fix-lint`. [`CONTRIBUTING.md`](../../../CONTRIBUTING.md) § "PHP linting" still names PSR-2; the ruleset that command actually runs is [PSR-12](https://www.php-fig.org/psr/psr-12/). When those disagree, `phpcs.xml` wins.
```
bwoebi, line 7: We need to drop this phpcs.xml file. It's very outdated and usused and I've seen agents pick it up.
The only linting we use is `composer ci-lint`.

bwoebi (changes requested):
I did a pass over the generic guidelines and how they apply to this repository, before approving on this.
Given that we're supposed to not comment in #4190, I'll comment here about missing overrides:

### `conventions.md`
- `- **Commit and PR hygiene** as this repo requires — the exact title format, label rules, and template are in .agents/dd-apm-sdk-review-overrides/reviewers/conventions.md if it exists, otherwise AGENTS.md.` - no such thing.
- `- **Config options.** Is a new option registered through this repo's own registration path, named per its conventions, documented, and given telemetry where the repo does that? This repo's exact registration steps are in .agents/dd-apm-sdk-review-overrides/reviewers/conventions.md if it exists — do not restate them from memory. Whether bypassing the registry rises to P0 is the design reviewer's call (it judges the architectural impact); report a bypass you find here as at least a P1 naming/registration gap.` - Isn't configuration public API? Doesn't that rather point to `design.md`?

### `correctness.md`
- `- **Async and ordering.** A callback, promise, or event assumed to fire in an order the runtime does not guarantee;
  a race between two paths touching the same state.` - the PHP threads are assumed to be isolated. Unless accessing global state, there must not be superfluous mutexes on only effectively thread locally accessible state. Conversely, check new global state, whether it's properly mutexed (obviously stuff running in minit or mshutdown is single-threaded and doesn't need a mutex), and if it is, check whether it risks being contended when hundreds of php threads run handling requests in the same process.

### `design.md`

- Extension public API are the `*.stub.php` files, PHP public API is in `src/api`. And probably configuration as defined in `configuration.h`. But note that some functions (internal functions are marked such) and configs can obviously be internal and broken at will.
- `- **Direction of dependencies.** Does the change introduce a cycle, or make a lower layer depend on a higher one?` - cross dependencies are normal, in particular not between the C and rust code. The only soft layer we have is ext/ -> product directories (have as little dependencies from ext/ on product dirs as possible). The hard layers (layers which are impossible to cross without a lot of effort) are between extension API and our PHP code in src/ and components -> zend_abstract_interface -> other extension code.
- `- **Extension points.** If this is an integration/plugin/instrumentation, does it use the repo's standard extension mechanism rather than a bespoke hook?` - I disagree on that. Use whatever fits best. We do have hooking functions, but in fact, we encourage using bespoke hooks if possible and non-invasive. (It just in most cases is not possible.)
- `**Configuration surface.**` - `configuration.h` files or `otel_config.c` for otel stuff.
- `- **Error strategy.** Does the change match the repo's convention for tracer failures (fail-soft, log-and-continue, never break the app)? A new hard throw on a customer path is a P0 finding.` - A hard throw on direct public API calls is expected, in particular for bad input values. (But not for internal state depending behaviour).
- Keep in mind that libdatadog is used as a submodule and may be taken into consideration.
- Hot path aggregation logic should be done in the extension / in a shared memory buffer. Aggregation across php threads and processes shall be done sidecar side. If operations can be batched, they should be and done on sidecar side
- The extension must never do network I/O itself, except through the sidecar.
- Avoid sending big buffers of data to the sidecar, that's expensive to read and write from (except as fallback).

### `maintainability.md`

- `**Intent is recoverable.** Can a reader tell *why* this code exists, not just what it does? Non-obvious decisions, workarounds, and version-specific hacks need a comment naming the reason. A magic constant with no explanation is a P1.` - magic constants are very common (sizing a buffer, comparing php versions, powers of two for bit operations), and "version specific" behaviour is also typical for the code base, like some stuff just being different between versions, not everything e.g. stuff in compatibility.h would need a comment.
- `- **Function and file size.** Does a new function do one thing? Was an already long function made longer instead of split?` - we don't mind about long functions, functions are mostly about code reuse and separation of concerns, not function size. The compiler can better reason within single functions. A function can be as small or big as needed to fulfil its task.
- `- **Test quality.** Do new tests assert behavior or implementation details? Are they deterministic (no sleeps, no wall-clock dependence, no network, no ordering assumptions)? A flaky new test is a P1.` - a lot of tests are dependent on local network and wall-clock to some degree. In particular tests asserting anything in framework tests or generally testing the end-to-end functionalities are going to send stuff to the request-replayer. This is within the boundaries of how testing works here. None of that is a problem. What should be asserted instead is that tests depending on the request-replayer are properly isolated. (phpt tests using `datadog.trace.agent_test_session_token` for example).
- `- **Error handling and observability.** When this fails in production, will the logs say what happened and where? Silent catch/swallow blocks that drop context are a P1; ones that swallow a real failure mode are P0.` - wrong thing to check for in C code. We don't have errors propagating that way.
- `- **Dead code and leftovers.** Commented-out code, unused parameters, debug prints, `TODO` without a ticket reference, stale docs left describing the old behavior.` - we sometimes add TODO as a signal "this behaviour is this way, if this needs improvement, here's this TODO". We don't add TODO as "this should be fixed immediately / needs to be scheduled". TODO is perfectly fine.
- `**Release notes / changelog.**` - we have an AI skill compiling release notes at release time, there's no such thing.

### `performance.md`
- `- **flag-with-confidence** — the cost is *mechanism-determined* and visible in the code: allocation, boxing, copying, unbounded growth, a native/FFI crossing. State it plainly.` - Rust FFI crossing is cheap! Native (i.e. extension API) function calling is also generally cheaper than doing it in plain PHP code! copying is mostly fine, deep copying is not (like creating a rust string from a charslice can be expensive for big strings / if done often).
- `| **SEV-2** | Startup latency | Eager loading, init, transformation |` Startup latency is just as bad as response time for CLI processes. Same severity category.
- `- Fix: defer to background/writer thread or task, lazy-compute, batch.` - never spawn new threads in the extension code, only ever on the sidecar side.
- `6. native-boundary-crossing` - as said, just drop that completely. (why is this even in the universal guidelines, isn't that language specific how expensive it is?)

---

Please have a look on how this can be integrated into overrides, or the generic core instructions adjusted appropriately. I feel like some of those shouldn't be even needing to be overridden, but be dropped from the generic reviewer instructions alltogether and be added in language specific ways.

robertomonteromiguel:
> I did a pass over the generic guidelines and how they apply to this repository, before approving on this. Given that we're supposed to not comment in #4190, I'll comment here about missing overrides:
> 
> ### `conventions.md`
> * `- **Commit and PR hygiene** as this repo requires — the exact title format, label rules, and template are in .agents/dd-apm-sdk-review-overrides/reviewers/conventions.md if it exists, otherwise AGENTS.md.` - no such thing.
> * `- **Config options.** Is a new option registered through this repo's own registration path, named per its conventions, documented, and given telemetry where the repo does that? This repo's exact registration steps are in .agents/dd-apm-sdk-review-overrides/reviewers/conventions.md if it exists — do not restate them from memory. Whether bypassing the registry rises to P0 is the design reviewer's call (it judges the architectural impact); report a bypass you find here as at least a P1 naming/registration gap.` - Isn't configuration public API? Doesn't that rather point to `design.md`?
> 
> ### `correctness.md`
> * `- **Async and ordering.** A callback, promise, or event assumed to fire in an order the runtime does not guarantee; a race between two paths touching the same state.` - the PHP threads are assumed to be isolated. Unless accessing global state, there must not be superfluous mutexes on only effectively thread locally accessible state. Conversely, check new global state, whether it's properly mutexed (obviously stuff running in minit or mshutdown is single-threaded and doesn't need a mutex), and if it is, check whether it risks being contended when hundreds of php threads run handling requests in the same process.
> 
> ### `design.md`
> * Extension public API are the `*.stub.php` files, PHP public API is in `src/api`. And probably configuration as defined in `configuration.h`. But note that some functions (internal functions are marked such) and configs can obviously be internal and broken at will.
> * `- **Direction of dependencies.** Does the change introduce a cycle, or make a lower layer depend on a higher one?` - cross dependencies are normal, in particular not between the C and rust code. The only soft layer we have is ext/ -> product directories (have as little dependencies from ext/ on product dirs as possible). The hard layers (layers which are impossible to cross without a lot of effort) are between extension API and our PHP code in src/ and components -> zend_abstract_interface -> other extension code.
> * `- **Extension points.** If this is an integration/plugin/instrumentation, does it use the repo's standard extension mechanism rather than a bespoke hook?` - I disagree on that. Use whatever fits best. We do have hooking functions, but in fact, we encourage using bespoke hooks if possible and non-invasive. (It just in most cases is not possible.)
> * `**Configuration surface.**` - `configuration.h` files or `otel_config.c` for otel stuff.
> * `- **Error strategy.** Does the change match the repo's convention for tracer failures (fail-soft, log-and-continue, never break the app)? A new hard throw on a customer path is a P0 finding.` - A hard throw on direct public API calls is expected, in particular for bad input values. (But not for internal state depending behaviour).
> * Keep in mind that libdatadog is used as a submodule and may be taken into consideration.
> * Hot path aggregation logic should be done in the extension / in a shared memory buffer. Aggregation across php threads and processes shall be done sidecar side. If operations can be batched, they should be and done on sidecar side
> * The extension must never do network I/O itself, except through the sidecar.
> * Avoid sending big buffers of data to the sidecar, that's expensive to read and write from (except as fallback).
> 
> ### `maintainability.md`
> * `**Intent is recoverable.** Can a reader tell *why* this code exists, not just what it does? Non-obvious decisions, workarounds, and version-specific hacks need a comment naming the reason. A magic constant with no explanation is a P1.` - magic constants are very common (sizing a buffer, comparing php versions, powers of two for bit operations), and "version specific" behaviour is also typical for the code base, like some stuff just being different between versions, not everything e.g. stuff in compatibility.h would need a comment.
> * `- **Function and file size.** Does a new function do one thing? Was an already long function made longer instead of split?` - we don't mind about long functions, functions are mostly about code reuse and separation of concerns, not function size. The compiler can better reason within single functions. A function can be as small or big as needed to fulfil its task.
> * `- **Test quality.** Do new tests assert behavior or implementation details? Are they deterministic (no sleeps, no wall-clock dependence, no network, no ordering assumptions)? A flaky new test is a P1.` - a lot of tests are dependent on local network and wall-clock to some degree. In particular tests asserting anything in framework tests or generally testing the end-to-end functionalities are going to send stuff to the request-replayer. This is within the boundaries of how testing works here. None of that is a problem. What should be asserted instead is that tests depending on the request-replayer are properly isolated. (phpt tests using `datadog.trace.agent_test_session_token` for example).
> * `- **Error handling and observability.** When this fails in production, will the logs say what happened and where? Silent catch/swallow blocks that drop context are a P1; ones that swallow a real failure mode are P0.` - wrong thing to check for in C code. We don't have errors propagating that way.
> * `- **Dead code and leftovers.** Commented-out code, unused parameters, debug prints, `TODO` without a ticket reference, stale docs left describing the old behavior.` - we sometimes add TODO as a signal "this behaviour is this way, if this needs improvement, here's this TODO". We don't add TODO as "this should be fixed immediately / needs to be scheduled". TODO is perfectly fine.
> * `**Release notes / changelog.**` - we have an AI skill compiling release notes at release time, there's no such thing.
> 
> ### `performance.md`
> * `- **flag-with-confidence** — the cost is *mechanism-determined* and visible in the code: allocation, boxing, copying, unbounded growth, a native/FFI crossing. State it plainly.` - Rust FFI crossing is cheap! Native (i.e. extension API) function calling is also generally cheaper than doing it in plain PHP code! copying is mostly fine, deep copying is not (like creating a rust string from a charslice can be expensive for big strings / if done often).
> * `| **SEV-2** | Startup latency | Eager loading, init, transformation |` Startup latency is just as bad as response time for CLI processes. Same severity category.
> * `- Fix: defer to background/writer thread or task, lazy-compute, batch.` - never spawn new threads in the extension code, only ever on the sidecar side.
> * `6. native-boundary-crossing` - as said, just drop that completely. (why is this even in the universal guidelines, isn't that language specific how expensive it is?)
> 
> Please have a look on how this can be integrated into overrides, or the generic core instructions adjusted appropriately. I feel like some of those shouldn't be even needing to be overridden, but be dropped from the generic reviewer instructions alltogether and be added in language specific ways.

Thank you for taking the time to review this—I really appreciate your feedback. Since these comments relate to the shared core instructions, could you please move them to the [[core PR](https://github.com/DataDog/dd-apm-sdk-review-core/pull/1)](https://github.com/DataDog/dd-apm-sdk-review-core/pull/1)? That will make the discussion visible to all teams and ensure any agreed changes are applied consistently across repositories.

.agents/dd-apm-sdk-review-overrides/reviewers/conventions.md
bwoebi: We dropped it now on master.

PR #4193, opened 2026-09-15
test(profiling): avoid I/O sample queue overflow

realFlowControl:
### Description

Write to a regular temporary file in the I/O correctness test and rewind it after every operation. This slows the synthetic write burst enough for the profiler consumer to drain its queue while keeping the file bounded to 10 KiB. `tmpfile()` also handles cleanup automatically.

### Why

The I/O upscaling correctness workload writes 163,840 small chunks to `/dev/null`. This can produce sampled events faster than the profiler's 100-message channel can consume them. The profiler intentionally uses non-blocking sends, so samples are dropped with `sending on a full channel` and the write total is underestimated.

This caused [PHP 8.1 ZTS to report 8.3% too few write bytes](https://github.com/DataDog/dd-trace-php/actions/runs/34907535322/job/104187568978?pr=4187) and [PHP 8.4 NTS to report 12.2% too few write bytes](https://github.com/DataDog/dd-trace-php/actions/runs/34907535322/job/104187568821?pr=4187).

The read totals passed because reading from `/dev/zero` is slower than reusing one buffer and writing it to `/dev/null`. Accounting for the queue-full warnings brings both write totals within the expected sampling noise, so this is queue overload rather than a problem with the exponential sampler introduced in #4159.

### Reviewer checklist
- [ ] Test coverage seems ok.


profiling/tests/correctness/io_upscaling.php
```diff
@@ -7,6 +7,7 @@
 function write_operations($handle, $data) {
     for ($i = 0; $i < OPERATION_COUNT; $i++) {
         fwrite($handle, $data);
+        rewind($handle);
```
morrisonlevi, line 10: This will also act as a speed limiter, as it slows us down to perform the syscall to seek (and remember the data is now backed by a file, it's a "real" operation). No problem with that, just pointing it out.

profiling/tests/correctness/io_upscaling.php
```diff
@@ -17,7 +18,7 @@ function read_operations($handle) {
 }
 
 function main() {
-    $write_handle = fopen('/dev/null', 'wb');
+    $write_handle = tmpfile();
```
morrisonlevi, line 21: If this returns `false` for failure, then `write_operations` will throw `TypeError`, so the error handling is robust enough (although the error may be initially confusing).

morrisonlevi (approved):
Approved. I think we should also optimize the cost of each sample as we've discussed in #4194 and in Slack.

We may also consider increasing the channel capacity a little bit. We did not increase it when we added I/O profiling, and an increased channel will increase the amount of samples that can be preserved whenever the background thread is unscheduled. It won't help much the background thread is being scheduled but simply isn't keeping up, though, which is why we should optimize the cost of creating/handling each sample. Optimizing that helps other profile types as well.

PR #4197, opened 2026-09-15
Use __cxa_atexit to avoid crash in atexit after unloading

bwoebi:
atexit() is truly at program end, but the tracer may have been unloaded by then. Using __cxa_atexit with our DSO triggers it at DSO unload time (or at exit, if not unloaded (yet)).


morrisonlevi (approved):
Approved. This is a bug we've been discussing on Slack that has been the cause of system tests failing for about a week.

PR #4204, opened 2026-09-16
style: fix ZAI_ASSERT/ZAI_ASSUME -Wunused-value warnings under GCC

morrisonlevi:
### Description

`ZAI_ASSUME` checked `defined(__GCC__)`, but GCC defines `__GNUC__`. Oops. So it was falling back to `(true, true)` and warning.

Once fixed, it warned for a similar reason, because now it expands to `__builtin_expect(...) ? __builtin_unreachable() : true`. So the `__extension({ })` syntax suppresses this. Have to do some other stuff to make it C+ compatible too.

### Reviewer checklist
- [ ] Test coverage seems ok.


zend_abstract_interface/zai_assert/zai_assert.h
```diff
@@ -39,22 +39,38 @@
...
  * ZAI_ASSERT is like ZEND_ASSERT and C assert that it will expand into a valid
  * expression which returns true (if it fails, it will not return at all).
+ *
+ * On GCC/Clang this is a statement expression rather than a plain comma
+ * expression: with a comma expression, the trailing `true` is an unused
+ * constant when ZAI_ASSERT is used as a bare statement (its most common
+ * use), and GCC warns on it (-Wunused-value). Statement expressions are a
+ * GNU extension unavailable on MSVC, so Windows (which only builds the
+ * tracer, not the profiler) keeps the plain comma-expression form; MSVC
+ * does not raise the equivalent warning there.
```
bwoebi:
```suggestion
 * Prevent -Wunused-value on GCC/clang, but use the comma operator on MSVC which doesn't support __extension__.
```

That's all what it wants to say. Please tidy such comments up!

bwoebi (approved):

PR #4207, opened 2026-09-17
test: exempt known broken AppSec musl ZTS releases

morrisonlevi:
### Description

The appsec installer tests will forever fail on Alpine ZTS releases for 1.25.x because they have a dependency on `__cxa_thread_atexit_impl` which is not provided. Skip them so we don't fail them in every PR until 1.25.2 or 1.26.0 are made.

### Reviewer checklist
- [ ] Test coverage seems ok.


cataphract (approved):
Approved. But why are we running tests against releases on every commit?

morrisonlevi:
I had the same question! What this is really testing is that the `datadog-setup.php` script can actually work with GitHub artifacts, which we obviously don't have for the current release. I had to ask Bob about this. Bob thinks that the fix should go a different way but he said it would be easier to open a PR than to explain it. So I am awaiting that PR.

PR #4212, opened 2026-09-17
test: remove appsec's nightly toolchain

morrisonlevi:
### Description

When I rebuilt these images this week, I didn't realize they were going to pick up an unpinned nightly. It's not a "problem" per-se, but it does create a bunch of warnings in the logs, such as this one in `helper-rust integration coverage`:

```
warning: use of deprecated method `std::sync::atomic::Atomic::<u64>::fetch_update`: renamed to `try_update` for consistency
  --> appsec/helper-rust/src/service/metrics.rs:22:14
   |
22 |             .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |s| {
   |              ^^^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
help: replace the use of the deprecated method
   |
22 -             .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |s| {
22 +             .try_update(Ordering::Relaxed, Ordering::Relaxed, |s| {
   |
```

But it's not actionable, as `.try_update` doesn't stabilize until `1.99` and our MSRV is only `1.91`.

~Version `2025-10-31` is near `1.91.1` and is the same version used elsewhere in our images. Pinning it removes a bunch of warnings.~ We don't need nightly, we'll just set `RUSTC_BOOTSTRAP=1`.

### Reviewer checklist
- [ ] Test coverage seems ok.


cataphract:
IIRC nightly is only to rebuild std. Can't we use stable for building, but set RUSTC_BOOTSTRAP=1? That way we wouldn't even need to install the nightly toolchain.

cataphract (commented):
I've removed the nightly toolchain. Images built and tags updated here.

cataphract (approved):
I've removed the nightly toolchain. Images built and tags updated here.

PR #4213, opened 2026-09-17
Publish process context during MINIT

cataphract:
Publish in the parent so the agent can reuse inferred TLS offsets in worker forks. See https://github.com/DataDog/datadog-agent/blob/main/pkg/security/ebpf/c/include/helpers/span.h#L57-L64

### Reviewer checklist
- [ ] Test coverage seems ok.


morrisonlevi (approved):
Looks fine to me, but I don't understand the full implications. Could you explain a bit more?

And obviously we need to get the tests working:

```
========DIFF========
001+ Server failed to start: Invalid address: 127.0.0.1:0
001- One Process Context mapping before any request: bool(true)
========DONE========
FAIL Linux OTel Process Context is published during MINIT [tmp/build_extension/tests/ext/otel_process_context_minit.phpt] 
```

cataphract:
> Looks fine to me, but I don't understand the full implications. Could you explain a bit more?

@morrisonlevi 
The agent has an optimization where forked processes (that have not execed) can have the otel tls layout parameters copied from the parent: https://github.com/DataDog/datadog-agent/blob/1d8c717e05c12eef13a1e91b97c1f32357182e38/pkg/security/ebpf/c/include/helpers/span_otel.h#L91-L103 and its calller https://github.com/DataDog/datadog-agent/blob/main/pkg/security/ebpf/c/include/helpers/span.h#L57-L64

For this to work, the parent needs to already have published the process context. So I'm doing that on MINIT, which is what is run pre-fork on fpm.

It also has the advantage that it reduces the window where we're already processing requests but the agent has not yet resolved the offsets. This is currently a problem in the system-tests, where we ahave to introduce a delay in the beginning of the request to give time to the user-space part of the resolution to take place.

PR #4219, opened 2026-09-19
Misc CI fixes

cataphract:
### Description

For the most part still unreviewed AI changes. Wait to leave draft state for review

### Reviewer checklist
- [ ] Test coverage seems ok.


libdatadog
cataphract, line 1: Note: rate_limiter.rs seems to have [other problems](https://gist.github.com/cataphract/7c2e980da83d4f776df370451e047709) that need to be fixed.

ext/remote_config.c
```diff
@@ -131,7 +143,13 @@ void datadog_minit_remote_config(void) {
 }
 
 void datadog_mshutdown_remote_config(void) {
-#ifndef _WIN32
+#ifdef _WIN32
+    if (DATADOG_G(sidecar)) {
+        datadog_sidecar_clear_reconnect_fn(&DATADOG_G(sidecar));
```
bwoebi, line 157: Can we please split this out of the PR?
This, including the associated libdatadog changes adds a #[blocking] to the mshutdown, which may introduce a significant wait with a busy sidecar on shutdown. That's something we want to avoid in non-thread mode sidecar.
A better mechanism would be introducing a Mutex handle on windows, which is held by either side:
- the sidecar guards its calls to `CreateRemoteThread(datadog_set_all_thread_vm_interrupt)` with it. If the mutex is already held (by PHP that is, as it's not the sidecar holding it), the sidecar will skip calling `CreateRemoteThread`.
- the extension at the end of mshutdown, effectively preventing the sidecar from calling it.

Said mutex handle can be provided over shm during connect.

.gitlab/generate-shared.php
```diff
@@ -27,7 +27,7 @@
     - if [ -d "/opt/catch2" ]; then export CMAKE_PREFIX_PATH=/opt/catch2; fi
     - mkdir -p tmp/build_php_components_asan && cd tmp/build_php_components_asan
     - cmake $([ -f "/etc/debian_version" ] && echo "-DCMAKE_TOOLCHAIN_FILE=../../cmake/asan.cmake") -DCMAKE_BUILD_TYPE=Debug -DDATADOG_PHP_TESTING=ON ../../components
-    - make -j all
+    - bash "${CI_PROJECT_DIR}/.gitlab/run-make-with-infra-retry.sh" -j all
```
bwoebi: You really think simple retries will help... or rather, like, do we understand why the linker even crashes? (or is it something trivial like race conditions in the linker, which a retry obviously makes work?)

bwoebi:
> build: update libdatadog sidecar deadlock fix = https://github.com/DataDog/libdatadog/compare/0e8c8db3028a3526f5e6affeaa749194bbb9c500...d3be0178a28b223a5c62a41dc9ca9640f9fc8103

Seems like the wrong fix to me, can we instead hold the lock the other way round on telemetry shutdown? TelemetryCachedEntry anyway is `Arc<Mutex<Option<TelemetryCachedClient>>>` - so we can just take() the Option, and then drop the lock. Then acquire the TelemetryCachedSet lock and check whether that entry is still None, and if yes, drop it.

Saves us from duplicating the map every time we want to flush.

.gitlab/generate-shared.php
cataphract: I'm going to remove this one. IIRC it's only one that I "fixed" without managing to reproduce it. And I tried many thousands of links. Maybe it was cosmic ray.

loader/dd_library_loader.c
```diff
@@ -1038,6 +1035,10 @@ static int ddloader_zend_extension_startup(zend_extension *ext) {
 
 static void ddloader_zend_extension_shutdown(zend_extension *ext) {
     UNUSED(ext);
+    // Zend still owns the loader and injected extension handles here. Wait for
+    // every loader thread before any of those handles can be released.
+    ddloader_join_reapers();
```
bwoebi, line 1040: The whole point of that construction is to _not wait for telemetry sending on shortlived cli processes_.
... and now you synchronize on it.

cataphract: Yes, I'm aware of this, which is why i reached to you. However, the thread, in its current form where it calls dlclose, cannot be allowed to still run when the atexit handlers run. Do you have any ideas?

bwoebi: You could add a global which prevents that thread from calling dlclose upon exiting. Like if the process is anyway in shutting down, no point in dlclosing then.

cataphract: Hmm how would that avoid the race? Even assuming that I could reliably register an atexit destructor to set the variable before the other atexit destructors (perhaps by registering it as late as possible, since they run in lifo order), I don't think that closes the race:

Thread:
Read variable (false)
Main:
Set variable true
Thread:
Dlclose 
Main:
Run the destructor

cataphract: I guess it could work with a mutex if I held it in the thread during the dlclose and the check before, but I'm a bit weary of whether that could deadlock with other glibc locks

bwoebi: Yes, a mutex was my thought as well. Not sure which other locks it would deadlock with though.

cataphract: That are two problems with this direction:

* It's difficult to register an atexit destructor that runs before all others, because C++ lazily intiialized sdtatic variables can register theirs as late as they're accessed for the first time,
* More importantly, where would I put the executable code for the atexit destructor that did:
  ```
  static void block_unload(void) {
      pthread_mutex_lock(&unload_mtx);
      shutting_down = true;
      pthread_mutex_unlock(&unload_mtx);
  }
  ```

  The two variables could be put on the stack, but here we're unloading the very module where we could place the function.

TLS destructors also run before atexit destructors, however those would prevent the ejection of the respective module.

ext/remote_config.c
cataphract: See https://github.com/DataDog/libdatadog/pull/2569

loader/dd_library_loader.c
cataphract: Opened the fix at https://github.com/DataDog/dd-trace-php/pull/4229 and rebased on that.

